Issue #6, opened 2020-05-09
write unit tests for num instructions in vm_num.go


mathetake:
https://github.com/mathetake/gasm/blob/52edbd17a5a15da3f201baffe011f9f562167606/wasm/vm_num_test.go#L3

PR #8, opened 2020-05-14
add some tests for num instructions 

ilmanzo:
about issue #6 : I started writing some tests , let me know what you think.

I have a question about testing "signed" operations; how do you manage it when the stack can contain only uint64 values ?

### DIFF
--- a/wasm/vm_num_test.go
+++ b/wasm/vm_num_test.go
@@ -1,44 +1,94 @@
 package wasm
 
 import (
-	"fmt"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
 )
 
-func TestNumOps(t *testing.T) {
+type NumTestSuite struct {
+	suite.Suite
+	vm *VirtualMachine
+}
 
+func (suite *NumTestSuite) SetupTest() {
+	suite.vm = &VirtualMachine{
+		OperandStack: NewVirtualMachineOperandStack(),
+	}
+}
+
+func (suite *NumTestSuite) Testi32eqz() {
+	suite.vm.OperandStack.Push(0)
+	i32eqz(suite.vm)
+	suite.Equal(uint64(1), suite.vm.OperandStack.Pop())
+}
+
+func (suite *NumTestSuite) Testi32ne() {
 	var testTable = []struct {
-		input [2]int                // numbers to push on the stack
-		op    func(*VirtualMachine) // operation to apply
-		desc  string                // string description of the operation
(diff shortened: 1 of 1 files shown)


mathetake:
@ilmanzo thank you for this 🎖️ 

> I have a question about testing "signed" operations; how do you manage it when the stack can contain only uint64 values ?

Maybe you would better take a look at this link here: https://stackoverflow.com/a/50815961 

basically all type of numbers are stored in uint64 with its corresponding bit representation

wasm/vm_num_test.go
```diff
@@ -1,3 +1,40 @@
...
+func Test_i32eq(t *testing.T) {
+	vm := buildvm()
+	vm.OperandStack.Push(uint64(3))
+	vm.OperandStack.Push(uint64(3))
+	i32eq(vm)
+	assert.Equal(t, uint32(1), uint32(vm.OperandStack.Pop()))
+}
+
+func Test_i32ne(t *testing.T) {
+	vm := buildvm()
```
mathetake, line 31: can you rewrite tests with multiple cases in the Table Drive style? https://github.com/golang/go/wiki/TableDrivenTests

ilmanzo:
thanks for the suggestion; I've refactored the tests with a table-driven style.

wasm/vm_num_test.go
```diff
@@ -1,3 +1,44 @@
...
 
-// fixme
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNumOps(t *testing.T) {
```
mathetake, line 10: thank you for trying grouping all instruction together but I don't think this is the way to go because there's several types of num operations (https://webassembly.github.io/spec/core/syntax/instructions.html#numeric-instructions).

I want separated unit tests for each instruction and each is written in table driven manner

ilmanzo:
ok I splitted each test in own function; (now using testify/suite) but each test have its own table . What do you think of this approach ?

wasm/vm_num_test.go
```diff
@@ -1,3 +1,94 @@
...
+	vm *VirtualMachine
+}
+
+func (suite *NumTestSuite) SetupTest() {
+	suite.vm = &VirtualMachine{
+		OperandStack: NewVirtualMachineOperandStack(),
+	}
+}
+
+func (suite *NumTestSuite) Testi32eqz() {
```
mathetake, line 20: should we add the case where the expected value is 1?

ilmanzo: ok

PR #9, opened 2020-05-18
add RuntimeData field to VirtualMachine

zhiqiangxu:
This `RuntimeData` can be used to persist runtime status  for `wasm`.

### DIFF
--- a/wasm/vm.go
+++ b/wasm/vm.go
@@ -20,6 +20,7 @@ type (
 		Globals       []uint64
 
 		OperandStack *VirtualMachineOperandStack
+		RuntimeData  interface{}
 	}
 
 	NativeFunctionContext struct {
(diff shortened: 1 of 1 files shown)


wasm/vm.go
```diff
@@ -20,6 +20,7 @@ type (
 		Globals       []uint64
 
 		OperandStack *VirtualMachineOperandStack
+		RuntimeData  interface{}
```
mathetake, line 23: could you please add the explanation of what this for?

zhiqiangxu: Added some comment.

zhiqiangxu:
> @zhiqiangxu thank you for this!
> 
> I think this would be helpful in some cases. Just for curious, do you have any specific example or use case of this field? I think we would better add some examples in this PR's description for future users
> 
> Thanks!

This field essentially attaches a piece of dedicated data to `VirtualMachine` so that no extra lookup is required, I have a similar field in my previous project: https://github.com/zhiqiangxu/qrpc/blob/master/frame.go#L15

PR #21, opened 2021-03-11
Fix table index bug

Gaboose:
Hi @mathetake! First of all, nice work with this project!

I tried running a rust-compiled hello world program with the `wasm32-wasi` target, started adding the required methods into the `wasi` package, and was pretty excited to get it working, but then kept getting a `function signature mismatch` panic from the `callIndirect` function in `vm_call.go`. It took me quite a while to debug, and I got to understand the wasm binary format way better because of it, but finally found the culprit:

There's a classic bug in the code that populates the table index space. This loop:
```go
for i, b := range elem.Init {
	table[i+offset] = &b
}
```
Takes the pointer of the temporary variable, which is reused on every iteration. The fix is just to use `table[i+offset] = &elem.Init[i]`.

Hope you see this pull request, and merge it! I want to contribute some more code to the `wasi` package afterwards.

### DIFF
--- a/wasm/module.go
+++ b/wasm/module.go
@@ -121,40 +121,48 @@ func (m *Module) buildIndexSpaces(externModules map[string]*Module) error {
 
 func (m *Module) resolveImports(externModules map[string]*Module) error {
 	for _, is := range m.SecImports {
-		em, ok := externModules[is.Module]
-		if !ok {
-			return fmt.Errorf("failed to resolve import of module name %s", is.Module)
-		}
-
-		es, ok := em.SecExports[is.Name]
-		if !ok {
-			return fmt.Errorf("%s not exported in module %s", is.Name, is.Module)
-		}
-
-		if is.Desc.Kind != es.Desc.Kind {
-			return fmt.Errorf("type mismatch on export: got %#x but want %#x", es.Desc.Kind, is.Desc.Kind)
-		}
-		switch is.Desc.Kind {
-		case 0x00: // function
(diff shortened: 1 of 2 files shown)


wasm/module.go
```diff
@@ -318,13 +326,13 @@ func (m *Module) buildTableIndexSpace() error {
 		if size > len(table) {
 			next := make([]*uint32, size)
 			copy(next, table)
-			for i, b := range elem.Init {
-				next[i+offset] = &b
+			for i := range elem.Init {
+				next[i+offset] = &elem.Init[i]
```
mathetake, line 330:
```suggestion
			for _, elm := range elem.Init {
				next[i+offset] = &elm
```

wasm/module.go
```diff
@@ -318,13 +326,13 @@ func (m *Module) buildTableIndexSpace() error {
...
 			}
 			m.IndexSpace.Table[elem.TableIndex] = next
 		} else {
-			for i, b := range elem.Init {
-				table[i+offset] = &b
+			for i := range elem.Init {
+				table[i+offset] = &elem.Init[i]
```
mathetake, line 335:
```suggestion
			for i, elm := range elem.Init {
				table[i+offset] = &elm
```

mathetake (commented):
@Gaboose thank you so much! Actually I have been aware of this Go-newbie bug since #16 is submitted, but never got to have a time to fix til now. 

LGTM ~modulo some nits.~ Thank you!

PR #27, opened 2021-10-01
Add Wasm Specification tests (v1.0).

mathetake:
This PR will add all the test suites to verify that the VM complies with the semantics specified by Wasm specification 1.0, and do some refactoring after my hiatus from this project.

v1.1 will come later:D.


mathetake:
The purpose of this also include building the base line of performance and allow me to do tons of optimizations I can think of.

PR #37, opened 2021-10-27
build: go 1.16 support by adding +build comments alongside go:build ones

bobby-stripe:
I believe this is the way to enable both go1.16 and go1.17+ support.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/tetratelabs/wazero
 
-go 1.17
+go 1.16
 
 require github.com/stretchr/testify v1.5.1
 
--- a/go.mod
+++ b/go.mod
@@ -3,9 +3,3 @@ module github.com/tetratelabs/wazero
 go 1.16
 
 require github.com/stretchr/testify v1.5.1
-
-require (
-	github.com/davecgh/go-spew v1.1.0 // indirect
-	github.com/pmezard/go-difflib v1.0.0 // indirect
-	gopkg.in/yaml.v2 v2.2.2 // indirect
-)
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
I rebased this since main drifted a bit

mathetake (approved):

codefromthecrypt:
Thanks for the help @bobby-stripe!

codefromthecrypt:
#46 adds CI for go 1.16. A later PR will reduce build flags to reduce compilation drift.

bobby-stripe:
thanks for merging this!

codefromthecrypt:
analysis on pros and cons of the last remaining build flag https://github.com/tetratelabs/wazero/pull/52

bobby-stripe:
@codefromthecrypt yup! we are

bobby-stripe:
happy to have this rolled back if its a burden

codefromthecrypt:
mainly we are thinking about how to handle trailing versions, as 1.18 is coming soon! I might preemptively roll back 1.16 and see if anyone else screams 😎 thanks for the feedback!

codefromthecrypt:
before I pull the trigger on removing 1.16.. @pkedy @pims @cretz @wuhuizuo @wdvxdr1123 can you mention if you require 1.16 and/or how close you are to needing 1.18? Currently, we don't test with 1.18 betas. It would be convenient to only test two versions of go, when it comes out (1.17 and 1.18).

pims:
I’m using 1.17 and no immediate plan to upgrade to 1.18.

codefromthecrypt:
I think @pkedy is busy, but considering dapr set floor level to go 1.17 I think we should be able to proceed. there are a lot of perf advantages to this.

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/317 will move to floor 1.17

PR #41, opened 2021-11-03
Introduce wazero's IR and add its compiler/interpreter

mathetake:
This commit introduces the wazero's IR (called `wazeroir` ) as a building block for future JIT compilation engine. 
Notably this adds `wazeroir` package and the implementation of the IR compiler, and also 
its direct interpreter.

The wazeroir is inspired by microwasm format (a.k.a. LightbeamIR), previously used
in the lightbeam JIT compiler in Wasmtime, though it is not specified and only exists
in [the previous codebase of wasmtime](https://github.com/bytecodealliance/wasmtime/blob/v0.29.0/crates/lightbeam/src/microwasm.rs).
The main difference from microwasm is that it doesn't have block operations
in order to simplify the implementation.

The wazeroir direct interpreter passes 100% of Wasm specification tests, and fully compatible just like 
`naivevm` interpreter.

Here's the example output of wazero IR translation from the Wasm function of factorial 

Wast:

```wat
  (func $fac (param i64) (result i64)
    (if (result i64) (i64.eqz (local.get 0))
      (then (i64.const 1))
      (else
        (i64.mul
          (local.get 0)
          (call $fac (i64.sub (local.get 0) (i64.const 1)))
        )
      )
    )
  )
```

wazeroir:

```
.entrypoint:
	pick 0
	i64.eqz
	br_if .L2, .L2_else
.L2:
	i64.const 1
	br .L2_cont
.L2_else:
	pick 0
	pick 1
	i64.const 1
	i64.sub
	call 24
	i64.mul
	br .L2_cont
.L2_cont:
	drop 1..1
	br .return
```


wasm/microwasm/operations.go
```diff
@@ -0,0 +1,591 @@
+package microwasm
```
codefromthecrypt, line 1: comment somewhere in this file a permalink to the spec for things basically following it. This helps later on updates (ex 1.0 to 1.1 and so forth) as well to understand what is spec-driven vs choice (ex no need to debate naming on things in the spec)

wasm/microwasm/operations.go
```diff
@@ -0,0 +1,591 @@
...
+	return OperationKindEq
+}
+
+type OperationNe SignLessInt
+
+func (o *OperationNe) Kind() OperationKind {
+	return OperationKindNe
+}
+
+type OperationEqz SignLessInt
```
codefromthecrypt, line 359: if at the top of the file, there's a link to the overall spec, I don't mind adding a commit to backfill comments per type. They would be like

```
// OperationEqz implements X ex foo = bar
// See https:// direct path to OperationEqz
```

we can assume that unlike GitHub code links spec links are permalinks even if sometimes they mess that up, it is usually mechanical to fix. GitHub code links are less ideal

codefromthecrypt (commented):
meta comment which I can help with if you point me to where to look

wasm/microwasm/compiler.go
```diff
@@ -0,0 +1,503 @@
...
+	}
+	for _, t := range f.Signature.InputTypes {
+		emitLocal(t)
+	}
+	for _, t := range f.LocalTypes {
+		emitLocal(t)
+	}
+
+	// Insert the function control frame.
+	returns := make([]SignlessType, len(f.Signature.ReturnTypes))
```
mathetake, line 94:
```suggestion
	returns := make([]SignlessType, 0, len(f.Signature.ReturnTypes))
```

mathetake:
progress: 90% done for translation, 0% for direct interpreter

mathetake:
progress: 100% done for translation/compilation, 20% for direct interpreter

mathetake:
progress: 100% done for translation/compilation, 100% for direct interpreter

now debugging the implementation and the followings tests are not yet passed by the interpreter:
```
align.wast.wast
float_exprs.wast
labels.wast
left-to-right.wast
memory.wast
start.wast
```

mathetake:
6 cases in  `float_exprs.wast` are the only things to burn! 🔥

mathetake:
Passed all the spec tests!

wasm/microwasm/interpreter.go
```diff
@@ -295,7 +293,7 @@ func (it *interpreter) lowerMicrowasm(f *wasm.FunctionInstance,
 			op.b1 = byte(o.Type)
 			op.us = make([]uint64, 2)
 			op.us[0] = uint64(o.Arg.Alignment)
-			op.us[0] = uint64(o.Arg.Offest)
+			op.us[1] = uint64(o.Arg.Offest)
```
mathetake, line 296: hehe this was the culprit 😄

wasm/spectests/util.go
```diff
@@ -0,0 +1,200 @@
+package spectests
```
mathetake, line 1: note this is just moved from spec_test.go so we can reuse it in wazeroir/compier_test.go

wasm/spectests/spec_test.go
```diff
@@ -202,173 +30,190 @@ func TestSpecification(t *testing.T) {
...
-								assert.Equal(t, len(exps), len(vals), msg) &&
-								assert.Equal(t, len(exps), len(types), msg) {
-								for i, exp := range exps {
-									assertValueEq(t, vals[i], exp, types[i], msg)
+			for _, tc := range []struct {
+				name   string
+				engine wasm.Engine
+			}{
+				{engine: naivevm.NewEngine(), name: "naivevm"},
+				{engine: wazeroir.NewEngine(), name: "wazeroir_interpreter"},
```
mathetake, line 43: now we have two interpreter which are 100% compatible with the spec!

codefromthecrypt (commented):
started review. I pushed a few commits vs having you slave to complete them. if you feel strongly against, revert! I'll continue reviewing also

wasm/spectests/util.go
```diff
@@ -0,0 +1,175 @@
...
+		// TODO:
+	case "assert_unlinkable":
+		// TODO:
+	case "assert_uninstantiable":
+		// TODO:
+	}
+	return "{" + msg + "}"
+}
+
+func (c Command) GetAssertReturnArgs() []uint64 {
```
codefromthecrypt, line 108: maybe comment somewhere why all return vals are coerced to uint64.. is this because it is the widest type and we reduce bits as needed upon evaluation?

codefromthecrypt:
I pulled some text from your PR desc into a RATIONALE.md. This will help as we make modifications, to have a place to store important "why"s

RATIONALE.md
```diff
@@ -0,0 +1,13 @@
+# Notable rationale of wazero
+
+## wazeroir
+wazero's intermediate representation (IR) is called `wazeroir`. Compiling into an IR provides us a faster interpreter
+and a building block for a future JIT compilation engine. Both of these help answer demands for a more performant
+runtime vs interpreting Wasm directly (the `naivevm` interpreter).
+
+### Intermediate Representation (IR) design
+`wazeroir`'s initial design borrowed heavily from the defunct `microwasm` format (a.k.a. LightbeamIR). Notably,
+`wazeroir` doesn't have block operations: this simplifies the implementation.
```
codefromthecrypt, line 10: does this have any impact? For example, what do we lose without block operations? Is it just for humans reading the IR? or is there a functionality or perf degrade resulting?

wasm/store.go
```diff
@@ -447,20 +447,26 @@ func (s *Store) buildFunctionInstances(module *Module, target *ModuleInstance) (
...
-		}
-
 		target.Functions = append(target.Functions, f)
 		s.Functions = append(s.Functions, f)
 	}
+
+	// We compile functions after successfully finished analyzing all functions.
+	// This is not only because we want to do early feedback on malicious binaries,
+	// but also during the compilation phase, the compilers have to see all the possible
+	// functions' signatures in the module instance.
```
codefromthecrypt, line 464: grammar

```suggestion
	// function signatures in the module instance.
```

RATIONALE.md
mathetake: I think nothing to loose -- it is just a convenience layer for removing unused blocks in lightbeam code, but I realized that it's redundant, and we could achieve that without block operations (though I haven't done it yet).

wasm/engine.go
```diff
@@ -6,3 +6,13 @@ type Engine interface {
 	Call(f *FunctionInstance, args ...uint64) (returns []uint64, err error)
 	Compile(f *FunctionInstance) error
 }
+
+type NopEngine struct{}
```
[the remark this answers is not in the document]
mathetake: I think now we could remove this NopEngine -- I just used this in compiler_test.go but all the objectives there are achieved by spec_test.go so I will remove this!

wasm/spectests/util.go
mathetake: this is not only the place where we have all in uint64 -- Wasm only has int32,int64(unsigned),float32,float64 and we have to represent them in one type and that's why we use uint64 for all types here and elsewhere.

.gitignore
```diff
@@ -0,0 +1 @@
+/.idea
```
mathetake, line 1: I would prefer you would put this in ~/.gitignore as this is not generated by this project :D @codefromthecrypt

wasm/wazeroir/compiler.go
```diff
@@ -0,0 +1,1480 @@
...
+			break operatorSwitch
+		}
+
+		// Create a new frame -- entering this block.
+		frame := &controlFrame{
+			frameID:          c.nextID(),
+			originalStackLen: len(c.stack),
+			kind:             controlFrameKindBlockWithoutContinuationLabel,
+		}
+		frame.asBranchTarget()
```
mathetake, line 200: seems unnecessary 😄 
```suggestion
```

wasm/wazeroir/interpreter.go
```diff
@@ -0,0 +1,1502 @@
...
+			}
+		case OperationKindDrop:
+			{
+				it.drop(op.rs[0])
+				frame.pc++
+			}
+		case OperationKindSelect:
+			{
+				c := it.pop()
+				v2 := it.pop()
```
codefromthecrypt, line 601: do we check length anywhere? ex `requireLen(4, it)` sometimes in parsing it is easier to evaluate preconditions so you can pop freely later.

wasm/wazeroir/interpreter.go
```diff
@@ -0,0 +1,1502 @@
...
+				switch SignLessType(op.b1) {
+				case SignLessTypeI32, SignLessTypeI64:
+					v2, v1 := it.pop(), it.pop()
+					b = v1 != v2
+				case SignLessTypeF32:
+					v2, v1 := it.pop(), it.pop()
+					b = math.Float32frombits(uint32(v2)) != math.Float32frombits(uint32(v1))
+				case SignLessTypeF64:
+					v2, v1 := it.pop(), it.pop()
+					b = math.Float64frombits(v2) != math.Float64frombits(v1)
```
codefromthecrypt, line 768: fromBits here and elsewhere

wasm/wazeroir/interpreter.go
```diff
@@ -0,0 +1,1502 @@
...
+}
+
+func (it *interpreter) callNativeFunc(f *interpreterFunction) {
+	frame := &interpreterFrame{f: f}
+	memoryInst := frame.f.moduleInstance.Memory
+	globals := frame.f.moduleInstance.Globals
+	tables := f.moduleInstance.Tables
+	it.pushFrame(frame)
+	bodyLen := uint64(len(frame.f.body))
+	for frame.pc < bodyLen {
```
codefromthecrypt, line 538: add a TODO or mental note to comment either on the constants or each case block what the operation implies in terms of how many if any args are popped and what if anything we expect to push, if there are any expected handled errors (ex overflow returns zero value or whatever). Main idea isn't to copy the spec, just a one max 2 line overview per non-trivial op

wasm/wazeroir/interpreter.go
```diff
@@ -0,0 +1,1502 @@
...
+		case OperationKindStore32:
+			{
+				val := uint32(it.pop())
+				base := op.us[1] + it.pop()
+				binary.LittleEndian.PutUint32(memoryInst.Buffer[base:], val)
+				frame.pc++
+			}
+		case OperationKindMemorySize:
+			{
+				v := uint64(int32(uint64(len(frame.f.moduleInstance.Memory.Buffer)) / wasm.PageSize))
```
codefromthecrypt, line 709: comment why the casting in an out is going on

wasm/wazeroir/signature.go
```diff
@@ -0,0 +1,376 @@
...
+		in:  []SignLessType{SignLessTypeF64, SignLessTypeF64},
+		out: []SignLessType{SignLessTypeF64},
+	}
+	signature_UnknownUnkownI32_Unknown = &signature{
+		in:  []SignLessType{SignLessTypeUnknown, SignLessTypeUnknown, SignLessTypeI32},
+		out: []SignLessType{SignLessTypeUnknown},
+	}
+)
+
+// wasmOptcodeSignature returns the signature of given Wasm optcode.
```
codefromthecrypt, line 145: maybe comment how this is intended to be used? maybe this is for validation?

wasm/wazeroir/signature.go
```diff
@@ -0,0 +1,376 @@
...
+		return signature_None_I32, nil
+	case wasm.OptCodeI64Const:
+		return signature_None_I64, nil
+	case wasm.OptCodeF32Const:
+		return signature_None_F32, nil
+	case wasm.OptCodeF64Const:
+		return signature_None_F64, nil
+	case wasm.OptCodeI32eqz:
+		return signature_I32_I32, nil
+	case wasm.OptCodeI32eq, wasm.OptCodeI32ne, wasm.OptCodeI32lts,
```
codefromthecrypt, line 266: I'm guessing that due to number of ops we aren't creating constants of each op including their signature.. if were not burdensome I might just because code navigation could easily link to the signature which could be helpful elsewhere when we are thinking about the signature implications of an op. For example, here we can do a reference search and then look in the case statement to figure it out. That's good also, though more indirect vs having the signature a field of the op

wasm/wazeroir/wazeroir.go
```diff
@@ -0,0 +1,12 @@
...
+// The wazeroir is inspired by microwasm format (a.k.a. LightbeamIR), previously used
+// in the lightbeam JIT compiler in Wasmtime, though it is not specified and only exists
+// in the previous codebase of wasmtime
+// e.g. https://github.com/bytecodealliance/wasmtime/blob/v0.29.0/crates/lightbeam/src/microwasm.rs
+//
+// The main difference from microwasm is that it doesn't have block operations
+// in order to simplify the implementation.
+// This package also has an implementation of direct interpreter of wazeroir which
+// implements the wasm.Engine interface, meaning that it can be used as an engine
+// of wazero runtime.
```
codefromthecrypt, line 11: you can put a code note here to look at the RATIONALE.md for a more complete description. that would allow a reduction of lines here

codefromthecrypt (approved):
As everything are nit, style or things that can happen after clicking 👍

epic job and good work!

wasm/store.go
```diff
@@ -447,20 +447,26 @@ func (s *Store) buildFunctionInstances(module *Module, target *ModuleInstance) (
...
-			return rollbackFuncs, fmt.Errorf("compilation failed at index %d/%d: %v", codeIndex, len(module.FunctionSection), err)
-		}
-
 		target.Functions = append(target.Functions, f)
 		s.Functions = append(s.Functions, f)
 	}
+
+	// We compile functions after successfully finished analyzing all functions.
+	// This is not only because we want to do early feedback on malicious binaries,
+	// but also during the compilation phase, the compilers have to see all the possible
```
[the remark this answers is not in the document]
mathetake: oh yeah the spectest fails when we revert this :D

wasm/wazeroir/compiler.go
```diff
@@ -0,0 +1,1480 @@
...
+	f      *wasm.FunctionInstance
+	result []Operation
+}
+
+func (c *compiler) stackDump() string {
+	strs := make([]string, 0, len(c.stack))
+	for _, s := range c.stack {
+		strs = append(strs, s.String())
+	}
+	return "[" + strings.Join(strs, ", ") + "]"
```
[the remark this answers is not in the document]
mathetake: oh yeah but in that case you have to eliminate the last "," from the tail before appending "]" 😄  that's why intentionally used strings.Join and +.

wasm/spectests/util.go
codefromthecrypt: would it hurt to use a type for []uint64 where the intent is the same? That would allow consolidation of the intent which is to be the widest type to handle all arg types.

.gitignore
codefromthecrypt: the main risk of leaving out well known IDE paths is that pull requests from external devs end up including them. while I can do this, it is generally friendly to ignore widely used IDE files.

codefromthecrypt: done anyway, just that's why!

mathetake: hehe thanks, I understand your rationale! :D

codefromthecrypt: yeah because we will have to tell a potential noob this in various operating system dialects.

```bash
$ git config --global core.excludesFile '~/.gitignore'
$ echo .idea >> ~/.gitignore
```

but it maybe won't come up so we can wait and see!

wasm/wazeroir/interpreter.go
```diff
@@ -0,0 +1,1502 @@
...
+		case *OperationFConvertFromI:
+			op.b1 = byte(o.InputType)
+			op.b2 = byte(o.OutputType)
+		case *OperationF32DemoteFromF64:
+		case *OperationF64PromoteFromF32:
+		case *OperationI32ReinterpretFromF32,
+			*OperationI64ReinterpretFromF64,
+			*OperationF32ReinterpretFromI32,
+			*OperationF64ReinterpretFromI64:
+			// Reinterpret ops are essentially nop for interpreter mode.
```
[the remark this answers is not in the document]
mathetake: because we treat all values as uint64, and the reinterpret is only used at module validation phase to check type soundness of all the operations.

wasm/wazeroir/interpreter.go
mathetake: the bit size is different from the above :D

wasm/wazeroir/interpreter.go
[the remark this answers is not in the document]
codefromthecrypt: sorry I meant comment why :D

wasm/wazeroir/interpreter.go
mathetake: as I commented https://github.com/tetratelabs/wazero/blob/microwasm/wasm/wazeroir/compiler.go#L64-L67, the stack operations are always valid at this runtime phase -- operations are already statically checked at module validation phase before compilation : https://github.com/tetratelabs/wazero/blob/microwasm/wasm/store.go#L445

wasm/wazeroir/interpreter.go
codefromthecrypt: nm that's a case format inconsistency in the stdlib!

wasm/wazeroir/compiler.go
```diff
@@ -0,0 +1,1480 @@
...
+		)
+	case wasm.OptCodeLocalSet:
+		if index == nil {
+			return fmt.Errorf("index does not exist for local.set")
+		}
+		depth := c.localDepth(*index)
+		c.emit(
+			// +1 because we already manipulated the stack before
+			// called localDepth ^^.
+			&OperationSwap{Depth: depth + 1},
```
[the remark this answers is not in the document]
mathetake: this is also covered in spectest :D

wasm/wazeroir/compiler.go
```diff
@@ -0,0 +1,1480 @@
...
+		c.emit(
+			&OperationSelect{},
+		)
+	case wasm.OptCodeLocalGet:
+		if index == nil {
+			return fmt.Errorf("index does not exist for local.get")
+		}
+		depth := c.localDepth(*index)
+		c.emit(
+			// -1 because we already manipulated the stack before
```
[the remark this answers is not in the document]
mathetake: same -- covered in spectests :D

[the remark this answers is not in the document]
codefromthecrypt: good to hear the spectests are good at edge cases!

mathetake:
@codefromthecrypt thanks for the review on this massive PR :D

PR #45, opened 2021-11-15
Ensures examples build

codefromthecrypt:
This adds CI for example code TinyGo -> Wasm. Differences are ignored because compilation is not idempotent.


codefromthecrypt:
PS this completes Makefile coverage!

mathetake (approved):

mathetake:
Thank you!

PR #46, opened 2021-11-16
Ensures CI tests with Go 1.16

codefromthecrypt:
This ensures we temporarily support go 1.16 per #37

Notably, this does not attempt to run linters or other optional tasks as we use go 1.17 features such as `go run` against a versioned tag.


mathetake (approved):
Thanks!🥇

PR #47, opened 2021-11-16
Removes option to disable stack overflow guard

codefromthecrypt:
This removes risk of compilation drift by removing the option to disable stack overflow protection.

### DIFF
--- a/wasm/buildoptions/stackoverflow.go
+++ b/wasm/buildoptions/stackoverflow.go
@@ -1,0 +1,7 @@
+package buildoptions
+
+// CallStackHeightLimit is the maximum WebAssembly call stack height. This allows wazero to raise
+// wasm.ErrCallStackOverflow instead of overflowing the Go runtime.
+//
+// The default value should suffice for most use cases. Those wishing to change this can via `go build -ldflags`.
+var CallStackHeightLimit = 2000
--- a/wasm/naivevm/vm_stack.go
+++ b/wasm/naivevm/vm_stack.go
@@ -139,14 +139,15 @@ func (s *frameStack) pop() *frame {
 }
 
 func (s *frameStack) push(val *frame) {
-	if s.sp+1 == len(s.stack) {
-		if callStackHeightLimit <= s.sp {
+	newSp := s.sp + 1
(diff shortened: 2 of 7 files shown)


wasm/naivevm/vm_stack.go
```diff
@@ -137,12 +137,10 @@ func (s *frameStack) pop() *frame {
 }
 
 func (s *frameStack) push(val *frame) {
-	if buildoptions.CheckCallStackOverflow {
+	if s.sp+1 == len(s.stack) {
 		if buildoptions.CallStackHeightLimit <= s.sp {
```
codefromthecrypt, line 141: I moved this inside so that it is only hit when actually growing

wasm/buildoptions/stackoverflow.go
```diff
@@ -0,0 +1,7 @@
+package buildoptions
+
+// CallStackHeightLimit is the maximum WebAssembly call stack height. This allows wazero to raise
```
codefromthecrypt, line 3: if this doc is inaccurate help me make it better!

wasm/wazeroir/interpreter.go
```diff
@@ -70,10 +70,8 @@ func (it *interpreter) drop(r *InclusiveRange) {
 }
 
 func (it *interpreter) pushFrame(frame *interpreterFrame) {
-	if buildoptions.CheckCallStackOverflow {
-		if buildoptions.CallStackHeightLimit <= len(it.frames)-1 {
-			panic(wasm.ErrCallStackOverflow)
-		}
+	if buildoptions.CallStackHeightLimit <= len(it.frames)-1 {
```
mathetake, line 73: While there let's reduce the cost of each operation😄
```suggestion
	if buildoptions.CallStackHeightLimit <= len(it.frames) {
```

mathetake (approved):
Thanks!

codefromthecrypt:
I owe vm_stack_test.go push test backfill prior to merge

mathetake (approved):
Thanks!!!!

codefromthecrypt:
Thanks for the help!

PR #48, opened 2021-11-16
Fix cursive VM calls with naivevm runtime.

mathetake:
This commit fixes a bug that exists in naivevm implementation
around stack frames. Notably this fixes the usage of 
recursive VM calls (i.e. calling VM functions inside of the 
host function implementations). In order to test the behavior,
we enhance the host_func.go example so that we use 
the recursive VM calls. 

ps currently we only run examples with naivevm, but
we might want to run them on both engines (naivevm
and wazeroir).


codefromthecrypt (approved):
good to cover this case, as I don't think it is really an edge case!

PR #49, opened 2021-11-16
Add basic benchmarks among engines.

mathetake:
This commit adds the basic benchmark to compare the performance of 
different engine implementations (currently naivevm vs wazeroir interpreter).
Notably this proves that wazeroir interpreter drastically outperforms the 
naivevm by the order of magnitudes. 

The follow up commits will add more variety of cases.

____

```
goos: linux
goarch: amd64
pkg: github.com/tetratelabs/wazero/bench
cpu: AMD Ryzen 9 3950X 16-Core Processor
BenchmarkEngines
BenchmarkEngines/naivevm
BenchmarkEngines/naivevm/base64_5_per_exec
BenchmarkEngines/naivevm/base64_5_per_exec-32         	  399853	      2542 ns/op	     848 B/op	      37 allocs/op
BenchmarkEngines/naivevm/base64_100_per_exec
BenchmarkEngines/naivevm/base64_100_per_exec-32       	  470588	      2443 ns/op	     848 B/op	      37 allocs/op
BenchmarkEngines/naivevm/base64_10000_per_exec
BenchmarkEngines/naivevm/base64_10000_per_exec-32     	  458335	      2532 ns/op	     848 B/op	      37 allocs/op
BenchmarkEngines/naivevm/fibo_for_5
BenchmarkEngines/naivevm/fibo_for_5-32                	   55761	     22410 ns/op	    9224 B/op	     380 allocs/op
BenchmarkEngines/naivevm/fibo_for_10
BenchmarkEngines/naivevm/fibo_for_10-32               	    3958	    266217 ns/op	  108937 B/op	    4511 allocs/op
BenchmarkEngines/naivevm/fibo_for_20
BenchmarkEngines/naivevm/fibo_for_20-32               	      36	  32989618 ns/op	13474027 B/op	  558218 allocs/op
BenchmarkEngines/wazeroir_interpreter
BenchmarkEngines/wazeroir_interpreter/base64_5_per_exec
BenchmarkEngines/wazeroir_interpreter/base64_5_per_exec-32         	 3180720	       365.5 ns/op	     104 B/op	       5 allocs/op
BenchmarkEngines/wazeroir_interpreter/base64_100_per_exec
BenchmarkEngines/wazeroir_interpreter/base64_100_per_exec-32       	 3351330	       360.7 ns/op	     104 B/op	       5 allocs/op
BenchmarkEngines/wazeroir_interpreter/base64_10000_per_exec
[...]

PR #50, opened 2021-11-17
Use wazeroir in examples.

mathetake:
As per #49, this commit uses wazeroir in examples to showcase that it is the default engine we recommend.


mathetake:
merging this as this is trivial!

PR #53, opened 2021-11-18
Execute _start func before benches.

mathetake:
Previously, we didn't execute _start function which is in charge of
in-Wasm memory state initialization.

That's why we faced the memory related error with `-gc=conservative` 
which is the default validator.

Also this commit revives the file system example which was temporarily removed.


examples/host_func_test.go
```diff
@@ -25,7 +25,7 @@ func Test_hostFunc(t *testing.T) {
 
 	// Host-side implementation of get_random_string on Wasm import.
 	getRandomString := func(ctx *wasm.HostFunctionCallContext, retBufPtr uint32, retBufSize uint32) {
-		const bufferSize = 10
+		const bufferSize = 10000
```
mathetake, line 28: This makes it necessary for Wasm program to grow memory space, and the growing memory fails when  we don't execute _start function beforehand. That's it!

examples/file_system_test.go
```diff
@@ -0,0 +1,71 @@
+package examples
```
mathetake, line 1: revived the file_system test!

bench/bench_test.go
```diff
@@ -15,41 +15,50 @@ import (
...
+	mod, err := wasm.DecodeModule((buf))
+	if err != nil {
+		panic(err)
+	}
+	err = store.Instantiate(mod, "test")
+	if err != nil {
+		panic(err)
+	}
+
+	// We assume that TinyGo binary expose "_start" symbol
```
codefromthecrypt, line 44: is there an api to check existence of a function? if so, we could still do this, and work if we switch the wasm implementation.

examples/file_system_test.go
codefromthecrypt: welcome back

examples/host_func_test.go
codefromthecrypt: ok so this fails nicer in missing `_start` in other words, right?

```suggestion
		const bufferSize = 10000 // force memory space grow to ensure eager failures on missing setup 
```

examples/host_func_test.go
```diff
@@ -54,6 +54,13 @@ func Test_hostFunc(t *testing.T) {
 	err = store.Instantiate(mod, "test")
 	require.NoError(t, err)
 
+	// We assume that TinyGo binary expose "_start" symbol
+	// to initialize the memory state.
+	// Meaning that TinyGo binary is "WASI command":
+	// https://github.com/WebAssembly/WASI/blob/main/design/application-abi.md
+	_, _, err = store.CallFunction("test", "_start")
```
codefromthecrypt, line 61: same suggestion as above, if we can detect presence of a function

codefromthecrypt (approved):
Looks good only one suggestion, but not sure it is possible

bench/bench_test.go
mathetake: we don't have such API but the CallFunction returns error in these case. I would create specific error for these cases `ErrFunctionNotFound`, but leave it for other PRs!

mathetake: https://github.com/tetratelabs/wazero/blob/main/wasm/store.go#L162-L170

codefromthecrypt: I will raise a PR for `HasFunction` or similar!

PR #54, opened 2021-11-18
wazeroir: emit backtrace in runtime error message.

mathetake:
This commit adds the parser of the "name" custom section[1],
and starts emitting the backtrace of the program  into the runtime error message
by leveraging the function name informations stored in "name" custom section.

Note that this is the minimal implementation -- the trace doesn't contain the 
original Wasm instruction address nor the source code info. The former
can be implemented easily but the latter requires us to implement the Wasm 
DWARF format parser of `.debug_*` custom sections[2].

[1] https://webassembly.github.io/spec/core/appendix/custom.html#name-section
[2] https://yurydelendik.github.io/webassembly-dwarf/

The following is the output from examples/trap_test.go:

```
=== RUN   Test_trap
panic: causing panic!!!!!!!!!!
wasm runtime error: unreachable
wasm backtrace:
	0: runtime._panic
	1: main.three
	2: main.two
	3: main.one
	4: cause_panic
```

___

As a comparison wasmtime emits like this:

```
wasmtime run examples/wasm/trap.wasm --invoke cause_panic
panic: causing panic!!!!!!!!!!
Error: failed to run main module `examples/wasm/trap.wasm`

Caused by:
    0: failed to invoke `cause_panic`
    1: wasm trap: unreachable
       wasm backtrace:
           0:  0x39d - <unknown>!runtime._panic
           1:  0x657 - <unknown>!main.three
           2:  0x64d - <unknown>!main.two
           3:  0x643 - <unknown>!main.one
           4:  0x639 - <unknown>!cause_panic
       note: run with `WASMTIME_BACKTRACE_DETAILS=1` environment variable to display more information
```


wasm/module.go
```diff
@@ -78,3 +80,61 @@ func DecodeModule(binary []byte) (*Module, error) {
...
+	r := bytes.NewReader(namesec)
+	for {
+		id, err := r.ReadByte()
+		if err != nil {
+			return nil, fmt.Errorf("failed to read subsection ID: %w", err)
+		}
+
+		size, _, err := leb128.DecodeUint32(r)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read the size of subsection: %w", err)
```
codefromthecrypt, line 99: maybe add the subsection ID into the error message? I would use code suggestion, but not sure the numeric encoding norms (ex if hex or otherwise)

wasm/module.go
```diff
@@ -78,3 +80,61 @@ func DecodeModule(binary []byte) (*Module, error) {
...
+		}
+
+		if id == 1 {
+			// ID = 1 is the function name subsection.
+			break
+		} else {
+			// Skip other subsections.
+			_, err := r.Seek(int64(size), io.SeekCurrent)
+			if err != nil {
+				return nil, fmt.Errorf("failed to seek: %w", err)
```
codefromthecrypt, line 109: maybe error should be like the other.. `"failed to skip subsection %s: %w", id, err`

wasm/store.go
```diff
@@ -416,13 +416,21 @@ func (s *Store) buildFunctionInstances(module *Module, target *ModuleInstance) (
...
 	functionDeclarations = append(functionDeclarations, module.FunctionSection...)
 	for _, g := range module.GlobalSection {
 		globalDecalarations = append(globalDecalarations, g.Type)
 	}
 	memoryDeclarations = append(memoryDeclarations, module.MemorySection...)
 	tableDeclarations = append(tableDeclarations, module.TableSection...)
 
+	functionNames, _ := module.GetFunctionNames()
+	if functionNames == nil {
+		// We cannot guarantiee the existence of "name" custom section
```
codefromthecrypt, line 429: also add why "name" could be missing? Ex not in standard yet vs not in the standard we support vs will always be custom. If always custom, some mention of de'facto use since we are supporting it here.

```suggestion
		// We cannot guarantee the existence of "name" custom section
```

wasm/store.go
```diff
@@ -431,7 +439,18 @@ func (s *Store) buildFunctionInstances(module *Module, target *ModuleInstance) (
...
 		}
 
+		// Get the function name if it exists.
+		// Please note that we plus with importedFunctionNum as
+		// the index in the function name section is for all the declarations,
+		// meaning that index starts with imported functions, and then
+		// the function defined in the module comes.
+		name, ok := functionNames[uint32(codeIndex)+uint32(importedFunctionNum)]
+		if !ok {
+			name = "unknown"
```
codefromthecrypt, line 449: if empty isn't a valid function name, let's use that vs "unknown"

wasm/wazeroir/interpreter.go
```diff
@@ -153,7 +153,7 @@ func (it *interpreter) Compile(f *wasm.FunctionInstance) error {
 // Lowers the wazeroir operations to interpreter friendly struct.
 func (it *interpreter) lowerIROps(f *wasm.FunctionInstance,
 	ops []Operation) (*interpreterFunction, error) {
-	ret := &interpreterFunction{moduleInstance: f.ModuleInstance, signature: f.Signature}
+	ret := &interpreterFunction{funcInstantance: f, signature: f.Signature}
```
codefromthecrypt, line 156: spelling

wasm/store.go
```diff
@@ -431,7 +439,18 @@ func (s *Store) buildFunctionInstances(module *Module, target *ModuleInstance) (
 			return rollbackFuncs, fmt.Errorf("code index out of range")
 		}
 
+		// Get the function name if it exists.
+		// Please note that we plus with importedFunctionNum as
```
codefromthecrypt, line 443: maybe simplify to "The function index includes imports, skip them"

wasm/module.go
```diff
@@ -78,3 +80,61 @@ func DecodeModule(binary []byte) (*Module, error) {
 	}
 	return ret, nil
 }
+
+func (m *Module) GetFunctionNames() (map[uint32]string, error) {
```
codefromthecrypt, line 84: this one screams "unit test me", especially with realistic inputs

examples/trap_test.go
```diff
@@ -28,4 +29,5 @@ func Test_trap(t *testing.T) {
 
 	_, _, err = store.CallFunction("test", "cause_panic")
 	require.Error(t, err)
+	fmt.Println(err)
```
codefromthecrypt, line 32: why not `require.Equal`?

codefromthecrypt (approved):
Ack some of my comments imply work possibly more than the code here, but I have to be Jiminy Cricket.

wasm/store.go
mathetake: will add the following: 

```
// That is because the custom section is optional
// in the sepc and some compiler optimize it out to reduce
// the binary size.
```

wasm/store.go
mathetake: empty is valid but this block means that `functionNames` map is empty (i.e. name custom section is removed out), so I think I will keep this as-is!

wasm/wazeroir/interpreter.go
mathetake: oh my 😥

wasm/module.go
mathetake: added!

mathetake:
@codefromthecrypt thanks for the help!!

examples/trap_test.go
```diff
@@ -28,4 +28,13 @@ func Test_trap(t *testing.T) {
 
 	_, _, err = store.CallFunction("test", "cause_panic")
 	require.Error(t, err)
+
+	const expErrMsg = `wasm runtime error: unreachable
```
codefromthecrypt, line 32: I :heart: this

codefromthecrypt:
great work

PR #55, opened 2021-11-19
Update TinyGo for examples builds.

mathetake:
0.21.0 has been released 10hrs ago!


codefromthecrypt:
> 0.21.0 has been released 10hrs ago!

what took you so long? 😉

PR #56, opened 2021-11-19
WIP: Adds user-level interface for WebAssembly 1.0 features

codefromthecrypt:
This adds a user-level interface narrowed to features supported in 1.0.
The implementation is not optimized to perform better in this case, but
it could. The main point of this is to feel out the API prior to doing
too much work.

In the future, we'd hide most apis currently exposed as internal. This
starts the discussion of the minimum viable interface for those invoking
functions, without requiring 1 impl per version of Wasm.

### DIFF
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/tetratelabs/wazero/wasi"
 	"github.com/tetratelabs/wazero/wasm"
 	"github.com/tetratelabs/wazero/wasm/naivevm"
+	"github.com/tetratelabs/wazero/wasm/v1_0"
 	"github.com/tetratelabs/wazero/wasm/wazeroir"
 )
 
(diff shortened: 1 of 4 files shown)


bench/bench_test.go
```diff
@@ -11,19 +11,28 @@ import (
...
+	"github.com/tetratelabs/wazero/wasm/v1_0"
 	"github.com/tetratelabs/wazero/wasm/wazeroir"
 )
 
 func BenchmarkEngines(b *testing.B) {
 	b.Run("naivevm", func(b *testing.B) {
 		store := newStore(naivevm.NewEngine())
 		setUpStore(store)
-		runAllBenches(b, store)
+		if m, ok := v1_0.NewModule(store, "test"); !ok {
```
codefromthecrypt, line 22: I used this in bench just to use it (so there's a diff to look at)

wasm/v1_0/wasm.go
```diff
@@ -0,0 +1,149 @@
...
+	for _, t := range f.f.Signature.InputTypes {
+		result = append(result, f.convertValueType(t))
+	}
+	return result
+}
+
+func (f *function) convertValueType(t wasm.ValueType) ValueType {
+	var r ValueType
+	switch t {
+	case wasm.ValueTypeI32:
```
codefromthecrypt, line 118: intentionally masking the universe of value types so that we can control our (probably internal) list independent of api views

codefromthecrypt (commented):
notes!

wasm/v1_0/wasm.go
```diff
@@ -0,0 +1,149 @@
...
+	case 0:
+		return 0, false
+	case 1:
+		return f.convertValueType(f.f.Signature.ReturnTypes[0]), true
+	default:
+		panic(fmt.Sprintf("invalid function %s/%s: more than one value type is unexpected in Wasm 1.0", f.m.n, f.n))
+	}
+}
+
+func (f *function) Call(parameters ...uint64) (uint64, error) {
```
codefromthecrypt, line 143: we might consider renaming call to "invoke" as it is the language the spec uses.. https://webassembly.github.io/spec/core/exec/modules.html#exec-invocation

mathetake:
I believe this is superseded by #238

PR #57, opened 2021-11-22
examples: panic on errors in Wasm.

mathetake:
This commit makes sure that in-Wasm errors are detected as panics in examples.


codefromthecrypt (approved):
If I understand properly, panic is good because there's currently no way to portably represent an error in Wasm.

On the other hand, panic will show the backtrace due to a recent change, and that will help people understand how panics (unhandled exceptions) work in wasm!

mathetake:
exactly!

PR #62, opened 2021-11-29
Adds header decoding tests and simplifies

codefromthecrypt:
This simplifies the header checks as they are ASCII so can be compared easier.

This also defocuses internal code from `io.Reader` as it is unimportant as `wasm.Reader` doesn't need to be exported: since it doesn't need to be exported, functions that require `io.Reader` will already fail compilation upon not implemented at the call site.

### DIFF
--- a/wasm/module.go
+++ b/wasm/module.go
@@ -8,50 +8,47 @@ import (
 	"github.com/tetratelabs/wazero/wasm/leb128"
 )
 
-var (
-	magic   = []byte{0x00, 0x61, 0x73, 0x6D}
-	version = []byte{0x01, 0x00, 0x00, 0x00}
+const (
+	magic   = "\x00asm"
+	version = "\x01\x00\x00\x00"
 )
 
-type Reader struct {
+type reader struct {
 	binary []byte
 	read   int
 	buffer *bytes.Buffer
 }
 
-func (r *Reader) Read(p []byte) (n int, err error) {
+func (r *reader) Read(p []byte) (n int, err error) {
 	n, err = r.buffer.Read(p)
 	r.read += n
 	return
 }
 
-var _ io.Reader = &Reader{}
-
-type (
-	// Static binary representations.
-	Module struct {
-		TypeSection     []*FunctionType
-		ImportSection   []*ImportSegment
-		FunctionSection []uint32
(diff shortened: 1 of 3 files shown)


wasm/module.go
```diff
@@ -8,50 +8,47 @@ import (
...
-	version = []byte{0x01, 0x00, 0x00, 0x00}
+const (
+	magic   = "\x00asm"
+	version = "\x01\x00\x00\x00"
 )
 
-type Reader struct {
+type reader struct {
 	binary []byte
 	read   int
```
codefromthecrypt, line 18: ex this field is read directly. if we didn't implement `io.Reader` calls like `io.ReadFull` would fail. IOTW, we don't need to double-check an internal struct (IMHO).

wasm/module.go
```diff
@@ -8,50 +8,47 @@ import (
...
-		ExportSection   map[string]*ExportSegment
-		StartSection    *uint32
-		ElementSection  []*ElementSegment
-		CodeSection     []*CodeSegment
-		DataSection     []*DataSegment
-		CustomSections  map[string][]byte
-	}
-)
+// Module is a WebAssembly binary representation.
+// See https://www.w3.org/TR/wasm-core-1/#modules%E2%91%A8
```
mathetake, line 29: ❤️

wasm/module_test.go
```diff
@@ -6,6 +6,45 @@ import (
...
+		},
+		{
+			name:        "wrong version",
+			input:       []byte("\x00asm\x01\x00\x00\x01"),
+			expectedErr: "invalid version header",
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc // pin! see https://github.com/kyoh86/scopelint for why
```
mathetake, line 34: I think we don't necessarily add this comment everywhere :D
```suggestion
		tc := tc
```

mathetake (approved):
Thanks!

wasm/module_test.go
codefromthecrypt: ok if no comment, I'll switch slightly to ensure compile fail.

codefromthecrypt:
thanks for the help review and the open mind @mathetake! each tiny PR was a glitch in my head synced to disk

PR #63, opened 2021-12-02
Adds skeleton lexer and benchmarks for text format

codefromthecrypt:
This adds an incomplete lexer for the Web Assembly text format. This is
left incomplete for the purpose of a smaller review. The code may change
as we add a parser and complete lexing floating point, but probably not
too much.

This also adds benchmarks as it is very easy to hurt performance unless
you compare before and after every change.

One design note is this uses a single loop instead of an iterator api more
common in lexers (ex call .nextToken etc). This is to save allocations and
the need to define yet another intermediate type.

For example, lexing the fibonnaci example is allocation free (at the
moment anyway!). It is important to note that the comparison against
Wat2Wasm isn't fair because they don't expose the lexing component.
Later, when we parse it will become more fair. It would also be ideal to
compare against wasm-tools somehow, but I've not found a C bridge to
that (rust project).

```
/private/var/folders/vd/1cf8zdb1721f4z5rjggy8bp40000gn/T/GoLand/___BenchmarkLex_in_github_com_tetratelabs_wazero_wasm_wat.test -test.v -test.paniconexit0 -test.bench ^\QBenchmarkLex\E$ -test.run ^$
goos: darwin
goarch: amd64
pkg: github.com/tetratelabs/wazero/wasm/wat
BenchmarkLex
BenchmarkLex/example_vs_utf8.ValidString
BenchmarkLex/example_vs_utf8.ValidString-16         	 2007007	       588.0 ns/op	     704 B/op	       1 allocs/op
BenchmarkLex/example_vs_wasmtime.Wat2Wasm
BenchmarkLex/example_vs_wasmtime.Wat2Wasm-16        	   45301	     26035 ns/op	     848 B/op	       3 allocs/op
BenchmarkLex/example
BenchmarkLex/example-16                             	 1000000	      1025 ns/op	       0 B/op	       0 allocs/op
```


codefromthecrypt:
Switched lexer to a incremental parse loop. This will allow us to more cheaply handle postion, column, etc and also avoid allocations when it comes to integrating with a real parser.

codefromthecrypt:
here are some early results of the general logic. There will be some optimizations left as there are some escapes noted in `go build -gcflags -m wasm/wat/*.go` which may not be great. Meanwhile, I'll move along functionality wise

```
BenchmarkLex
BenchmarkLex/base
BenchmarkLex/base-16         	67579126	        15.85 ns/op
BenchmarkLex/whitespace_chars
BenchmarkLex/whitespace_chars-16         	61862394	        19.48 ns/op
BenchmarkLex/unicode_line_comment
BenchmarkLex/unicode_line_comment-16     	26326196	        45.67 ns/op
BenchmarkLex/unicode_block_comment
BenchmarkLex/unicode_block_comment-16    	20848278	        52.55 ns/op
```

codefromthecrypt:
changed the example to use fibonacci and rejigged things to ease test writing. there's now some visible tech debt on token types that have similar branches. I'll work on that probably a lot of it will get better with a bitflag table to identify the token type from the current first byte

codefromthecrypt:
today I reworked logic to eliminate some (not all) tech debt and work on performance. I'm happy on performance for now as reached an emotional threshold of 1us to lex the example.

wasm/wat/lexer.go
```diff
@@ -0,0 +1,239 @@
...
+		col = col + 1                              // the current character is at least one byte long
+		if b1 == ' ' || b1 == '\t' || b1 == '\r' { // fast path ASCII whitespace
+			continue // next whitespace
+		}
+
+		// Handle comments, noting they can be nested
+		switch b1 {
+		case '(':
+			peekPos := p + 1
+			if peekPos < length && source[peekPos] == ';' { // block comment
```
mathetake, line 55: what if `peekPos < length` doesn't satisfy? better reporting parse error?

wasm/wat/lexer.go
```diff
@@ -0,0 +1,239 @@
...
+				p = peekPos // continue after "(;"
+				col = col + 1
+				blockCommentLevel = blockCommentLevel + 1
+				continue
+			}
+		case ';': // possible line comment or block comment end
+			peekPos := p + 1
+			if peekPos < length {
```
mathetake, line 63: same what if peekPos < length doesn't satisfy? better reporting parse error? :D

wasm/wat/lexer.go
```diff
@@ -0,0 +1,239 @@
...
+
+				if b2 == ';' { // line comment
+					// Start after ";;" and run until the end. Note UTF-8 (multi-byte) characters are allowed.
+					peekPos = peekPos + 1
+					col = col + 1
+
+					for peekPos < length {
+						peeked := source[peekPos]
+						if peeked == '\n' {
+							break // EOL bookkeeping will proceed on the next iteration
```
mathetake, line 80: maybe better use the label break as we are in the multiple nested loop here for readability!

wasm/wat/lexer.go
```diff
@@ -0,0 +1,239 @@
...
+
+						col = col + 1
+						s := utf8Size[peeked] // While unlikely, it is possible the byte peeked is invalid unicode
+						if s == 0 {
+							return fmt.Errorf("%d:%d found an invalid byte in line comment: 0x%x", line, col, peeked)
+						}
+						peekPos = peekPos + s
+					}
+
+					p = peekPos - 1 // at the '\n'
```
mathetake, line 91: maybe add the same comment for why -1 (the for loop will + 1") below?

wasm/wat/lexer_test.go
```diff
@@ -0,0 +1,541 @@
...
+	"testing"
+	"unicode/utf8"
+
+	"github.com/bytecodealliance/wasmtime-go"
+	"github.com/stretchr/testify/require"
+)
+
+// exampleWat was at one time in the wasmtime repo under cranelift. We added a unicode comment for fun!
+const exampleWat = `(module
+  ;; 私たちはフィボナッチ数列を使います。何故ならみんなやってるからです。
```
mathetake, line 15: 🤣

wasm/wat/token.go
```diff
@@ -0,0 +1,217 @@
...
+	// tokenReserved is a sequence of idChar characters which are neither a tokenId nor a tokenString.
+	//
+	// For example, '0$y' is a tokenReserved, because it doesn't start with a letter or '$'.
+	//
+	// See https://www.w3.org/TR/wasm-core-1/#text-reserved
+	tokenReserved
+)
+
+// tokenNames is index-coordinated with tokenType
+var tokenNames = []string{
```
mathetake, line 94: nit 
```suggestion
var tokenNames = [...]string{
```

wasm/wat/token.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	return
+}
+
+func _idChar(ch byte) bool {
+	switch ch {
+	case '!':
+		fallthrough
+	case '#':
+		fallthrough
```
mathetake, line 164: I think we could reduce the # of lines like this:
```suggestion
	case '!', '#', .... :
		return true
```

wasm/wat/lexer.go
```diff
@@ -0,0 +1,239 @@
...
+				}
+				peekPos = peekPos + s
+			}
+
+			if !hitQuote {
+				return fmt.Errorf("%d:%d expected end quote", line, col)
+			}
+
+			// set the position to after the quote
+			p = peekPos
```
mathetake, line 179: As the `peekPos` is *on* the quote :D
```suggestion
			p = peekPos
			// set the position to after the quote
```

mathetake (approved):
brilliant! 🏅 I mostly understand the logics so feel free to merge it :D (Assuming that there are proper test cases!)

wasm/wat/lexer.go
codefromthecrypt: '(' is a valid token, so peekPos being not enough or peeked char being something besides ';' will result in '('.

wasm/wat/lexer.go
codefromthecrypt: it will error because ';' is an invalid token. I think what's confusing you perhaps is that we are only handling whitespace indicators in this section. Anything that passes down will become a tokenization pass or fail (and ';' would fail).

It is a bit easier to keep the responsibility separate (whitespace including comment handling) from tokenization.

wasm/wat/lexer.go
codefromthecrypt: ok sure!

wasm/wat/lexer.go
codefromthecrypt: sure

wasm/wat/token.go
codefromthecrypt: oh cool. my go is better when you teach me things like this!

wasm/wat/token.go
codefromthecrypt: ah right I'll do that.

wasm/wat/lexer.go
codefromthecrypt: sounds like I need a comment

codefromthecrypt:
I thought about @mathetake's comments last night and in the middle of it :D: he's right. error reporting should win vs responsibility separation. anything that nests is particularly tricky to understand. The part that confuses this lexing is that parens are used both in block comments and in S expressions. Right now, we are only counting nesting in block comments because we are required to for unicode skipping to work. The better solution I think vs comments is to enforce nesting in normal parens, also. This brings coherence to the two places we are nesting as they need similar error reporting.

Ex. now at the lexing layer, we allow ")(" which is valid tokenization, but invalid code. The error for ")(" will be very similar to the same broken block comment ";)(;". If we will move paren tokenization up into the same place that handles white space, we reduce questions about the overlap in characters and also can tighten reporting. Counting and enforcing parens costs a little code, but we'd have to do it at the parsing layer if not here. So, it ends up at worst a wash, performance wise.

wasm/wat/lexer.go
codefromthecrypt: fixed

wasm/wat/lexer.go
codefromthecrypt: fixed

codefromthecrypt:
@mathetake thanks for the pointers, PTAL at the latest commit if you have a minute 🙇

mathetake (approved):
nice!

wasm/wat/token.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// tokenId is a sequence of idChar characters prefixed by a '$':
+	//
+	// For example, in the below, '$y' is an id:
+	//		local.get $y
+	//		i32.const 6
+	//		i32.lt_s
+	//
+	// See https://www.w3.org/TR/wasm-core-1/#text-id
+	tokenId
```
mathetake, line 77: maybe 
```suggestion
	tokenID
```

codefromthecrypt:
Thanks for the help @mathetake!

Issue #65, opened 2021-12-08
Tracking issue for the baseline JIT engine for amd64 target


mathetake:
## Baseline JIT engine
- Stack hight checks (https://github.com/tetratelabs/wazero/pull/68)
- Implement all the Instructions
    - https://github.com/tetratelabs/wazero/pull/70
    - https://github.com/tetratelabs/wazero/pull/71
    - https://github.com/tetratelabs/wazero/pull/73
    - https://github.com/tetratelabs/wazero/pull/81
    - https://github.com/tetratelabs/wazero/pull/82
    - https://github.com/tetratelabs/wazero/pull/85
    - https://github.com/tetratelabs/wazero/pull/86
    - https://github.com/tetratelabs/wazero/pull/89
    - https://github.com/tetratelabs/wazero/pull/93
    - https://github.com/tetratelabs/wazero/pull/96
    - https://github.com/tetratelabs/wazero/pull/97
    - https://github.com/tetratelabs/wazero/pull/99
    - https://github.com/tetratelabs/wazero/pull/103
    - https://github.com/tetratelabs/wazero/pull/111
    - https://github.com/tetratelabs/wazero/pull/113
    - https://github.com/tetratelabs/wazero/pull/116
    - https://github.com/tetratelabs/wazero/pull/118
    - https://github.com/tetratelabs/wazero/pull/119
    - https://github.com/tetratelabs/wazero/pull/121
    - https://github.com/tetratelabs/wazero/pull/124
    - https://github.com/tetratelabs/wazero/pull/126
    - https://github.com/tetratelabs/wazero/pull/132
    - https://github.com/tetratelabs/wazero/pull/141 for br_table
 
- basic stack trace: https://github.com/tetratelabs/wazero/pull/70 
- Pass spectests
    - https://github.com/tetratelabs/wazero/pull/127
    - https://github.com/tetratelabs/wazero/pull/141

## Post baseline
- achieve function calls without going back to Go code (`engine.exec`)
- Adds tests and better documentations on mmap system calls (See https://github.com/tetratelabs/wazero/pull/60#discussion_r765449928)
- Bunch of optimizations! (not limited to single pass)

PR #68, opened 2021-12-09
jit: grow stack automatically. 

mathetake:
This commit resolves one TODO for growing stack automatically 
when the engine enters new JITed function frames and sees the necessity
to grow the stack.

part of #65.

### DIFF
--- a/wasm/jit/engine.go
+++ b/wasm/jit/engine.go
@@ -269,8 +269,7 @@ func (e *engine) maybeGrowStack(requiredHeight uint64) {
 		// This case we need to grow the stack as the empty slots
 		// are not able to store all the stack items.
 		// So we grow the stack with the new len = currentLen*2+required.
-		required := requiredHeight - remained
-		newStack := make([]uint64, currentLen*2+(required))
+		newStack := make([]uint64, currentLen*2+(requiredHeight))
 		top := e.currentBaseStackPointer + e.currentStackPointer
 		copy(newStack[:top], e.stack[:top])
 		e.stack = newStack
--- a/wasm/jit/engine.go
+++ b/wasm/jit/engine.go
@@ -268,7 +268,7 @@ func (e *engine) maybeGrowStack(maxStackPointer uint64) {
 	if maxStackPointer > remained {
 		// This case we need to grow the stack as the empty slots
 		// are not able to store all the stack items.
-		// So we grow the stack with the new len = currentLen*2+required.
+		// So we grow the stack with the new len = currentLen*2+maxStackPointer.
 		newStack := make([]uint64, currentLen*2+(maxStackPointer))
 		top := e.currentBaseStackPointer + e.currentStackPointer
 		copy(newStack[:top], e.stack[:top])
--- a/wasm/jit/engine.go
+++ b/wasm/jit/engine.go
@@ -160,7 +160,7 @@ func NewEngine() wasm.Engine {
 	return newEngine()
 }
 
-const initialStackSize = 100
+const initialStackSize = 1024
 
 func newEngine() *engine {
 	e := &engine{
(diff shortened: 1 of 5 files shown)


wasm/jit/engine.go
```diff
@@ -251,16 +251,31 @@ type compiledWasmFunction struct {
...
+		// This case we need to grow the stack as the empty slots
+		// are not able to store all the stack items.
+		// So we grow the stack with the new len = currentLen*2+required.
+		required := requiredHeight - remained
+		newStack := make([]uint64, currentLen*2+(required))
+		top := e.currentBaseStackPointer + e.currentStackPointer
+		copy(newStack[:top], e.stack[:top])
+		e.stack = newStack
+	}
+	// TODO: maybe better think about how to shrink the stack as well.
```
codefromthecrypt, line 278: good note.. it may be hard to decide this without keeping stats of stack height

wasm/jit/jit_value_location_amd64.go
```diff
@@ -155,6 +156,9 @@ func (s *valueLocationStack) push(loc *valueLocation) {
 	} else {
 		s.stack[s.sp] = loc
 	}
+	if s.sp > s.maxHeight {
```
codefromthecrypt, line 159: max is a word usually implying a boundary, but this feels like it will change any time `sp` changes. for that reason, I feel it is useful to add a comment like `// allow the go allocated stack to grow to accommodate the stack pointer`

even if we comment, though I'd still feel like it isn't so much a `max` as a length thing, as it is adjusted whenever it needs to be... wdyt?

codefromthecrypt (commented):
while maxtStackHeight (or maxStackHeight ;))  is indeed the maximum it can grow, it feels more like a size/length thing vs a max thing, as it is dynamically adjusted.

Ex if I have an array backed list, I might not call the size of the array max even if it is true that's the maximum (until it changes!)

wasm/jit/engine.go
```diff
@@ -251,16 +251,31 @@ type compiledWasmFunction struct {
...
+// for the next function frame execution.
+func (e *engine) maybeGrowStack(requiredHeight uint64) {
+	currentLen := uint64(len(e.stack))
+	remained := currentLen - e.currentBaseStackPointer
+	if requiredHeight > remained {
+		// This case we need to grow the stack as the empty slots
+		// are not able to store all the stack items.
+		// So we grow the stack with the new len = currentLen*2+required.
+		required := requiredHeight - remained
+		newStack := make([]uint64, currentLen*2+(required))
```
codefromthecrypt, line 273: hard to do math in my head, but will `currentLen*2+(required)` always be less than `requiredHeight`? if not, feels even less like a max :D

wasm/jit/jit_value_location_amd64.go
codefromthecrypt, line 159: unrelated but would this condition trigger when `s.sp >= uint64(len(s.stack))` if so, maybe move the resetting `s.maxHeight = s.sp` to there? unless we are expecting shrinking?

mathetake: nope, we have to track s.sp whenever we `push`, meaning we advance sp😄  `s.sp >= uint64(len(s.stack)` satisfies when we try pushing more items than the current slice can accommodate.

wasm/jit/engine.go
mathetake: right, fixed: https://github.com/tetratelabs/wazero/pull/68/commits/49c0763bfac7bf66ff7a4f6549c6e59e61fb37f7 :D

wasm/jit/jit_value_location_amd64.go
mathetake: it's not about the length of slice here, but about the stack pointer. So I will rename `maxHeight` to `maxStackPointer`  to clarify!

codefromthecrypt: 💯

mathetake: https://github.com/tetratelabs/wazero/pull/68/commits/31611a5160d80b58483a08618af219830922b84c done! Thanks for the comment!

wasm/jit/jit_value_location_amd64.go
```diff
@@ -104,9 +104,10 @@ func newValueLocationStack() *valueLocationStack {
 }
 
 type valueLocationStack struct {
```
mathetake, line 106: I think I should add comments on this struct

codefromthecrypt: are you expecting me to say "no you shouldn't!"?😎

wasm/jit/engine.go
```diff
@@ -251,16 +251,30 @@ type compiledWasmFunction struct {
 	codeInitialAddress uintptr
 	// The same purpose as codeInitialAddress, but for memory.Buffer.
 	memoryAddress uintptr
+	// The max of the stack pointer this function can reach.
```
codefromthecrypt, line 254: This will help in code navigation a bit as I think what is confusing is that setting this is more like scheduling it to grow vs synchronously doing it.

```suggestion
	// The max of the stack pointer this function can reach. Lazily applied via maybeGrowStack.
```

codefromthecrypt (approved):
looking good

mathetake:
I will put the comments on the struct and then merge this :D

wasm/jit/jit_value_location_amd64.go
mathetake: https://github.com/tetratelabs/wazero/pull/68/commits/86b0105859364caa0340ed61a82a99c557923cb7 here you go!

mathetake:
@codefromthecrypt thanks for the reviews! 💯 🏅 🙏

PR #69, opened 2021-12-10
Ignore Wasm spectest files from GitHub language statistics

mathetake:


codefromthecrypt (approved):
hah good point

mathetake:
Now the appearance becomes what wazero is really about :D

Screen Shot 2021-12-10 at 10 27 46

Issue #58, opened 2021-11-23
Implement DWARF parser for better backtraces


mathetake:
## Background 

LLVM-based compilers for Wasm, for examples C/C++, Rust, Zig, TinyGo (virtually 100% of viable languages),
emit DWARF information into `.debug_*` custom sections.  The following is the sections contained in a TinyGo binary:
```
$ wasm-objdump main.wasm -h

main.go.wasm:	file format wasm 0x1

Sections:

     Type start=0x0000000b end=0x00000158 (size=0x0000014d) count: 42
   Import start=0x0000015b end=0x000003df (size=0x00000284) count: 18
 Function start=0x000003e2 end=0x000004e8 (size=0x00000106) count: 260
    Table start=0x000004ea end=0x000004ef (size=0x00000005) count: 1
   Memory start=0x000004f1 end=0x000004f4 (size=0x00000003) count: 1
   Global start=0x000004f6 end=0x000004fe (size=0x00000008) count: 1
   Export start=0x00000501 end=0x000007ac (size=0x000002ab) count: 31
     Code start=0x000007b0 end=0x0001b258 (size=0x0001aaa8) count: 260
     Data start=0x0001b25c end=0x00020862 (size=0x00005606) count: 2
   Custom start=0x00020866 end=0x00034e9a (size=0x00014634) ".debug_info"
   Custom start=0x00034e9d end=0x00035f66 (size=0x000010c9) ".debug_pubtypes"
   Custom start=0x00035f6a end=0x000431a7 (size=0x0000d23d) ".debug_loc"
   Custom start=0x000431aa end=0x00044f60 (size=0x00001db6) ".debug_ranges"
   Custom start=0x00044f62 end=0x00044fa1 (size=0x0000003f) ".debug_aranges"
   Custom start=0x00044fa4 end=0x00046ef6 (size=0x00001f52) ".debug_abbrev"
   Custom start=0x00046efa end=0x00059503 (size=0x00012609) ".debug_line"
   Custom start=0x00059507 end=0x0006510b (size=0x0000bc04) ".debug_str"
   Custom start=0x0006510f end=0x0006bf6b (size=0x00006e5c) ".debug_pubnames"
   Custom start=0x0006bf6e end=0x0006e6e8 (size=0x0000277a) "name"
   Custom start=0x0006e6eb end=0x0006e778 (size=0x0000008d) "producers"
```

[...]

PR #70, opened 2021-12-10
jit: support unreachable instruciton, and stack trace.

mathetake:

part of #65


wasm/jit/engine.go
```diff
@@ -34,20 +35,43 @@ type engine struct {
...
+	defer func() {
+		if v := recover(); v != nil {
+			top := e.callFrameStack
+			var traces []string
+			var counter int
+			for top != prevFrame {
+				traces = append(traces, fmt.Sprintf("\t%d: %s", counter, top.getFunctionName()))
+				top = top.caller
+				counter++
+				// TODO: include DWARF symbols.
```
mathetake, line 53: Add reference to #58 

```suggestion
				// TODO: include DWARF symbols. See #58 
```

wasm/jit/engine.go
```diff
@@ -226,18 +254,34 @@ type callFrame struct {
...
+}
+
+type compiledHostFunction = struct {
+	f    func(ctx *wasm.HostFunctionCallContext)
+	name string
+}
+
 type compiledWasmFunction struct {
+	// FunctionInstance from which this is compiled.
+	originalFunctionInstance *wasm.FunctionInstance
```
codefromthecrypt, line 284: maybe `source` ?

wasm/jit/testdata/unreachable.wat
```diff
@@ -0,0 +1,14 @@
...
+		(call $one)
+	)
+	(func $one
+		(call $two)
+	)
+	(func $two
+		(call $three)
+	)
+	(func $three
+		(unreachable)
```
codefromthecrypt, line 12: is there a more intuitive use of unreachable? like something where it is may occur in real life (not a forged example)

wasm/jit/engine.go
```diff
@@ -34,20 +35,43 @@ type engine struct {
...
 }
 
-type hostFunction = func(ctx *wasm.HostFunctionCallContext)
-
 func (e *engine) Call(f *wasm.FunctionInstance, args ...uint64) (returns []uint64, err error) {
+	prevFrame := e.callFrameStack
+	defer func() {
+		if v := recover(); v != nil {
+			top := e.callFrameStack
+			var traces []string
```
codefromthecrypt, line 47:
```suggestion
			var frames []string
```

wasm/jit/engine.go
```diff
@@ -34,20 +35,43 @@ type engine struct {
...
-	hostFunctionIndex map[*wasm.FunctionInstance]int64
+	compiledHostFunctions     []*compiledHostFunction
+	compiledHostFunctionIndex map[*wasm.FunctionInstance]int64
 }
 
-type hostFunction = func(ctx *wasm.HostFunctionCallContext)
-
 func (e *engine) Call(f *wasm.FunctionInstance, args ...uint64) (returns []uint64, err error) {
+	prevFrame := e.callFrameStack
+	defer func() {
```
codefromthecrypt, line 44: comment why specifically here we are handling panic?

codefromthecrypt (approved):
as usual, mostly naming and semantics!

wasm/jit/testdata/unreachable.wat
mathetake: for example `unreachable` will be placed in a branch for the case when the slice boundary check failed (like TinyGO does emit unreachable on "slice out of bounds")!

mathetake: But it is essentially the same as this example

codefromthecrypt:
thanks for making it possible to uhh become unreachable!

PR #72, opened 2021-12-10
introduce wasi.Errno type and switch error codes to constants

### LINKED ISSUE #64: Add wasi.Errno type implementing the error interface
Hello,

I'm opening this issue to discuss changing the way error codes in https://github.com/tetratelabs/wazero/blob/main/wasi/errno.go are represented.

Instead of being simple `uint32`, how about making the a type implementing the `error` interface. This would be useful to return error codes from custom implementations of the `wasi.FS` and `wasi.File` interfaces, which are otherwise converted to the generic `EIO` code in functions like `fd_read` or `fd_write`, for example https://github.com/tetratelabs/wazero/blob/main/wasi/wasi.go#L258

The `Errno` type would be defined as such:

```go
type Errno uint32

func (e Errno) Error() string { return fmt.Sprintf("wasi.Errno(%d)", uint32(e)) }
```

The error codes declared in errno.go would be declared as constants of the `Errno` type:

```go
const (
  ESUCCESS Errno = 0
  ...
)
```

The signature of functions like `fd_read` can then be modified to return an `Errno` value instead of a plain `uint32`. I believe this should remain compatible with the current implementation since the underlying type remains unchanged, the use of reflection to discover the function signature would keep seeing return values of kind `reflect.Uint32`.

Let me know if you have any concerns about it or I have overlooked implementation details, I can send a pull request if we agree the change would be useful.

achille-roussel:
This PR implements the changes discussed in https://github.com/tetratelabs/wazero/issues/64

I modified the original proposal a bit so that the `Error` method will return the name of the error code rather than the numeric code, the intent being to be more meaningful when reading formatted errors in application logs.

Let me know if you have any comments!

resolves #64


mathetake (approved):
LGTM. Thank you @achille-roussel for your contribution!🏅

codefromthecrypt:
fyi we had to revert this to ErrnoName because all errnos are not errors (ex errnosuccess isn't an error), plus we have some thinking to do wrt how to allow host functions to raise errors #258. Finally, the type alias complicates things like reflection, and generation which are easier if only use go types

ps if you are using something besides tinygo and/or need a WASI API not defined here, please pipe up in the issue https://github.com/tetratelabs/wazero/issues/271

PR #75, opened 2021-12-16
Allows NULL per spec

codefromthecrypt:
It seems strange to me to intentionally allow NULL byte in the middle of comments and strings, but that's allowed by spec! This switches the malformed test to use 0x80 which is invalid unicode.

### DIFF
--- a/wasm/wat/lexer.go
+++ b/wasm/wat/lexer.go
@@ -14,10 +14,10 @@ import (
 // Returning an error will short-circuit any future invocations.
 type parseToken func(source []byte, tok tokenType, beginLine, beginCol, beginPos, endPos int) error
 
-// lex invokes the parser function for each token, the source is exhausted.
+// lex invokes the parser function for each token, the source is exhausted. The fileName is used in error messages.
 //
 // Errors from the parser or during tokenization exit early, such as dangling block comments or unexpected characters.
-func lex(source []byte, parser parseToken) error {
+func lex(fileName string, source []byte, parser parseToken) error {
 	length := len(source)
 	// p is the position in the source, and a parameter to the parser.
 	p := 0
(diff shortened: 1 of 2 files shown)


wasm/wat/lexer.go
```diff
@@ -14,10 +14,10 @@ import (
 // Returning an error will short-circuit any future invocations.
 type parseToken func(source []byte, tok tokenType, beginLine, beginCol, beginPos, endPos int) error
 
-// lex invokes the parser function for each token, the source is exhausted.
+// lex invokes the parser function for each token, the source is exhausted. The fileName is used in error messages.
 //
 // Errors from the parser or during tokenization exit early, such as dangling block comments or unexpected characters.
-func lex(source []byte, parser parseToken) error {
+func lex(fileName string, source []byte, parser parseToken) error {
```
codefromthecrypt, line 20: oh I also added the fileName even though we don't use it, yet. I noticed wabt does and it makes sense to.

wasm/wat/lexer.go
```diff
@@ -28,12 +28,12 @@ func lex(source []byte, parser parseToken) error {
 
 	// Web assembly expressions are grouped by parenthesis, even the minimal example "(module)". We track nesting level
 	// to help report problems instead of bubbling to the parser layer.
-	nestLevel := 0
+	parenDepth := 0
```
codefromthecrypt, line 31: wasm-tools uses the word "depth" which I think is better than "level" so borrowing that, too

wasm/wat/lexer.go
```diff
@@ -221,22 +221,24 @@ func lex(source []byte, parser parseToken) error {
...
-		return fmt.Errorf("%d:%d expected ')', but reached end of input", line, col+1)
+	if parenDepth > 0 {
+		return fmt.Errorf("%s:%d:%d: expected ')', but reached end of input", fileName, line, col+1)
 	}
 	return nil // EOF
 }
 
 // utf8Size returns the size of the UTF-8 rune based on its first byte, or zero.
 //
+// Note: The null byte (0x00) is here as it is valid in comments. See WebAssembly/spec#1372
```
codefromthecrypt, line 235:
```suggestion
// Note: The null byte (0x00) is here as it is valid in string tokens and comments. See WebAssembly/spec#1372
```

codefromthecrypt (commented):
notes

mathetake (approved):

PR #76, opened 2021-12-18
Renames OptCode to Opcode

codefromthecrypt:
In the spec, Opcode is a single word (not optcode). This corrects and
adds a verification which was at the context I noticed the naming
glitch. Advice on that verification welcome!


wasm/naivevm/vm.go
```diff
@@ -112,9 +112,17 @@ func (vm *naiveVirtualMachine) Compile(f *wasm.FunctionInstance) error {
...
 		}
-	} else {
-		// Compile native Wasm functions.
-		f.Body[len(f.Body)-1] = byte(wasm.OptCodeReturn)
+	} else { // Compile native Wasm functions.
+
+		lastIndex := len(f.Body) - 1
+		switch f.Body[lastIndex] {
+		case wasm.OpcodeReturn:
+		case wasm.OpcodeEnd: // coerce end to return: TODO why?
```
codefromthecrypt, line 120: This is the new check. all other change is mechanical.

wasm/naivevm/vm.go
```diff
@@ -176,7 +184,7 @@ func (vm *naiveVirtualMachine) exec(f *wasm.FunctionInstance) (errRet error) {
 	for vm.activeFrame != prevActive {
 		if buildoptions.IsDebugMode {
 			fmt.Printf("0x%x: op=%s (Label SP=%d, Operand SP=%d, Frame SP=%d) \n",
-				vm.activeFrame.pc, buildoptions.OptcodeStrs[vm.activeFrame.f.Body[vm.activeFrame.pc]],
+				vm.activeFrame.pc, buildoptions.OpcodeStrs[vm.activeFrame.f.Body[vm.activeFrame.pc]],
```
codefromthecrypt, line 187: ps do you think it is important to hide the string constants in debugmode? ack it will grow the code by an array of 256 (unless we trim it to fit)

mathetake: I think we don't need to hide the strings as it seems over optimization😄

wasm/naivevm/vm.go
mathetake: do we need to have this swtich? From the spec, I understand that every function end with "end" opcode so we shouldn't reach case `wasm.OpcodeReturn` above. Maybe I am missing something though

codefromthecrypt: ok I will check to see if anything hits "return" if not, then take out the case statement!

codefromthecrypt: at least our tests hit the return branch... if I panic, one test dies

```
--- FAIL: TestSpecification (0.57s)
    --- FAIL: TestSpecification/imports.wast (0.00s)
        --- FAIL: TestSpecification/imports.wast/naivevm (0.00s)
            --- FAIL: TestSpecification/imports.wast/naivevm/module/line:24 (0.00s)
panic: return at end [recovered]
	panic: return at end

goroutine 30428 [running]:
```

so, there are a couple questions I guess.
* is it a bug in translation that ^^ ends with a return?
* for the cases that end in .. "end", why do we change that to return? (I need to add a comment per TODO)

mathetake: > is it a bug in translation that ^^ ends with a return?
naivevm doesn't translate so I think my understanding of spec was wrong. Function can end with "return" instead of "end"

> for the cases that end in .. "end", why do we change that to return? (I need to add a comment per TODO)
 
that's just because I implemented the naivevm in the way that all the function must end with "return" :D

codefromthecrypt: will revert this per https://github.com/tetratelabs/wazero/issues/79 as we don't need to spend time on dead code

wasm/instruction.go
```diff
@@ -0,0 +1,398 @@
...
+	OpcodeF64PromoteF32:     "f64.promote_f32",
+	OpcodeI32ReinterpretF32: "i32.reinterpret_f32",
+	OpcodeI64ReinterpretF64: "i64.reinterpret_f64",
+	OpcodeF32ReinterpretI32: "f32.reinterpret_i32",
+	OpcodeF64ReinterpretI64: "f64.reinterpret_i64",
+}
+
+// InstructionName returns the instruction corresponding to this binary Opcode.
+// See https://www.w3.org/TR/wasm-core-1/#a7-index-of-instructions
+func InstructionName(oc Opcode) string {
```
codefromthecrypt, line 396: fyi I renamed the file to instruction.go as the binary opcode is just one column describing an instruction.

mathetake (approved):
img

mathetake (approved):

codefromthecrypt:
thanks for the look @mathetake

PR #77, opened 2021-12-20
Centralizes error handling of line and column numbers 

codefromthecrypt:
This changes the lex API to return the line and column number it stopped
at. This removes a lot of redundant formatting code and allow users to
control error formatting centrally.

### DIFF
--- a/wasm/wat/lexer_test.go
+++ b/wasm/wat/lexer_test.go
@@ -1,6 +1,7 @@
 package wat
 
 import (
+	"errors"
 	"fmt"
 	"testing"
 	"unicode/utf8"
@@ -36,8 +37,6 @@ const exampleWat = `(module
 
 // TestLex_Example is intentionally verbose to catch line/column/position bugs
 func TestLex_Example(t *testing.T) {
-	tokens, e := lexTokens(exampleWat)
-	require.NoError(t, e)
 	require.Equal(t, []*token{
 		{tokenLParen, 1, 1, "("},
 		{tokenKeyword, 1, 2, "module"},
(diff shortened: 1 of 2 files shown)


wasm/wat/lexer_test.go
```diff
@@ -442,101 +442,127 @@ func TestLex(t *testing.T) {
...
-			input:       []byte(")("),
-			expectedErr: "test.wasm:1:1: found ')' before '('",
+			name:         "open paren at end of input",
+			input:        []byte("("),
+			expectedLine: 1,
+			expectedCol:  1,
+			expectedErr:  "found '(' at end of input",
 		},
 		{
-			name:        "unbalanced nesting",
```
codefromthecrypt, line 455: apologies that removing the paren depth tests misaligned the diff.

codefromthecrypt:
marked as draft until I have representative usage example

codefromthecrypt:
one design I tried and dismissed was having the parser return a result of the next parser. While neat, this slowed benchmarks 5x!

Next step will be to show how to use this API to parse a module iteratively.

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/78 should go before this, and show that this mostly just helps with error handling. we may not actually need to iteratively parse afterall so I may repurpose this solely to centralize error formatting with line/col

wasm/wat/lexer_test.go
```diff
@@ -571,9 +647,9 @@ func BenchmarkLex(b *testing.B) {
 		b.Run(bm.name, func(b *testing.B) {
 			b.ReportAllocs()
 			for i := 0; i < b.N; i++ {
-				err := lex(noopParseToken, "test.wasm", bm.data)
+				line, col, err := lex(noopParseToken, bm.data)
 				if err != nil {
-					panic(err)
+					panic(fmt.Errorf("%d:%d: %w", line, col, err))
```
codefromthecrypt, line 652: a real caller who knows the filename could put that in the format, also!

mathetake (approved):

wasm/wat/lexer_test.go
mathetake: 💯

codefromthecrypt:
thanks for the patience with the api thrash @mathetake!

PR #86, opened 2021-12-23
jit: support memory.load instruction

mathetake:


wasm/jit/jit_amd64.go
```diff
@@ -98,7 +98,9 @@ func (e *engine) compileWasmFunction(f *wasm.FunctionInstance) (*compiledWasmFun
 				return nil, fmt.Errorf("error handling global.set operation: %w", err)
 			}
 		case *wazeroir.OperationLoad:
-			return nil, fmt.Errorf("unsupported operation in JIT compiler: %v", o)
+			if err := builder.handleLoad(o); err != nil {
+				return nil, fmt.Errorf("error handling global.set operation: %w", err)
```
codefromthecrypt, line 102: ps can we use a string form to generify the error handling across these branches?

```suggestion
				return nil, fmt.Errorf("error handling load operation: %w", err)
```

wasm/jit/jit_amd64.go
```diff
@@ -1096,6 +1098,90 @@ func (b *amd64Builder) handleLe(o *wazeroir.OperationLe) error {
 	return nil
 }
 
+func (b *amd64Builder) handleLoad(o *wazeroir.OperationLoad) error {
+	base := b.locationStack.pop()
+	if base.onStack() {
+		if err := b.moveStackToRegisterWithAllocation(generalPurposeRegisterTypeInt, base); err != nil {
```
codefromthecrypt, line 1104: how about this?

```suggestion
		if err := b.moveStackToRegisterWithAllocation(generalPurposeIntRegister, base); err != nil {
```

wasm/jit/jit_amd64.go
```diff
@@ -1096,6 +1098,90 @@ func (b *amd64Builder) handleLe(o *wazeroir.OperationLe) error {
...
+			return err
+		}
+	} else if base.onConditionalRegister() {
+		if err := b.moveConditionalToGPRegister(base); err != nil {
```
codefromthecrypt, line 1108:
```suggestion
		if err := b.moveConditionalToGeneralPurposeRegister(base); err != nil {
```

wasm/jit/jit_amd64.go
```diff
@@ -1096,6 +1098,90 @@ func (b *amd64Builder) handleLe(o *wazeroir.OperationLe) error {
...
+
+	// TODO: Emit instructions here to check memory out of bounds as
+	// potentially it would be an security risk.
+
+	var (
+		isIntType bool
+		movInst   obj.As
+	)
+	switch o.Type {
+	case wazeroir.SignLessTypeI32:
```
codefromthecrypt, line 1134: so assuming signless is not the same as unsigned, would it be better to use wasm conventions and say `TypeI32` (instead of `TypeU32`)? https://www.w3.org/TR/wasm-core-1/#integers%E2%91%A0

wasm/jit/jit_amd64.go
```diff
@@ -1096,6 +1098,90 @@ func (b *amd64Builder) handleLe(o *wazeroir.OperationLe) error {
...
+	case wazeroir.SignLessTypeF32:
+		isIntType = false
+		movInst = x86.AMOVL
+	case wazeroir.SignLessTypeF64:
+		isIntType = false
+		movInst = x86.AMOVQ
+	}
+
+	if isIntType {
+		// For integer types, we just read the corresponding bytes from the offset to the memory
```
codefromthecrypt, line 1149: "just" makes things seem simple and even if it is, let's avoid it. I saw a presentation once where the speaker called "just" a magician!

```suggestion
		// For integer types, read the corresponding bytes from the offset to the memory
```

wasm/jit/jit_amd64.go
```diff
@@ -1096,6 +1098,90 @@ func (b *amd64Builder) handleLe(o *wazeroir.OperationLe) error {
...
+		moveFromMemory.To.Type = obj.TYPE_REG
+		moveFromMemory.To.Reg = floatReg
+		moveFromMemory.From.Type = obj.TYPE_MEM
+		moveFromMemory.From.Reg = reservedRegisterForMemory
+		moveFromMemory.From.Index = reg
+		moveFromMemory.From.Scale = 1
+		b.addInstruction(moveFromMemory)
+		top := b.locationStack.pushValueOnRegister(floatReg)
+		top.setValueType(o.Type)
+		// We no longer need the int register so mark it unused.
```
codefromthecrypt, line 1179: good comment!

wasm/jit/jit_amd64_test.go
```diff
@@ -1402,6 +1402,97 @@ func TestAmd64Builder_handleCall(t *testing.T) {
...
+		wazeroir.SignLessTypeI64,
+		wazeroir.SignLessTypeF32,
+		wazeroir.SignLessTypeF64,
+	} {
+		tp := tp
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			eng := newEngine()
+			builder := requireNewBuilder(t)
+			builder.initializeReservedRegisters()
+			// Before load operations, we must push the base offset value.
```
codefromthecrypt, line 1417: is 100 a magic number? what determines the base? if arbitrary, finish with "... for this test, we use an arbitrary value"

wasm/jit/jit_amd64_test.go
```diff
@@ -1402,6 +1402,97 @@ func TestAmd64Builder_handleCall(t *testing.T) {
...
+			const baseOffset = 100
+			base := builder.locationStack.pushValueOnStack()
+			eng.stack[base.stackPointer] = baseOffset
+			// Emit the memory load instructions.
+			o := &wazeroir.OperationLoad{Type: tp, Arg: &wazeroir.MemoryImmediate{Offest: 361}}
+			err := builder.handleLoad(o)
+			require.NoError(t, err)
+			// At this point, the loaded value must be on top of the stack, and placed on a register.
```
codefromthecrypt, line 1425: when making comments try to line break before as it reads easier

wasm/jit/jit_amd64_test.go
```diff
@@ -1402,6 +1402,97 @@ func TestAmd64Builder_handleCall(t *testing.T) {
...
+			loadedValue := builder.locationStack.peek()
+			require.Equal(t, o.Type, loadedValue.valueType)
+			require.True(t, loadedValue.onRegister())
+			// Double the laoded value in order to verify the behavior.
```
codefromthecrypt, line 1429:
```suggestion
			// Double the loaded value in order to verify the behavior.
```

wasm/jit/jit_amd64_test.go
```diff
@@ -1402,6 +1402,97 @@ func TestAmd64Builder_handleCall(t *testing.T) {
...
+			doubleLoadedValue.As = addInst
+			doubleLoadedValue.To.Type = obj.TYPE_REG
+			doubleLoadedValue.To.Reg = loadedValue.register
+			doubleLoadedValue.From.Type = obj.TYPE_REG
+			doubleLoadedValue.From.Reg = loadedValue.register
+			builder.addInstruction(doubleLoadedValue)
+
+			// Compile.
+			builder.releaseAllRegistersToStack()
+			builder.returnFunction()
```
codefromthecrypt, line 1455: why we are doing releaseAllRegistersToStack and returnFunction is special knowledge unless you comment why :D

wasm/jit/jit_amd64_test.go
```diff
@@ -1402,6 +1402,97 @@ func TestAmd64Builder_handleCall(t *testing.T) {
...
+			code, err := builder.compile()
+			require.NoError(t, err)
+
+			// Place the load target value to the memory.
+			mem := newMemoryInst()
+			var expValue uint64
+			switch tp {
+			case wazeroir.SignLessTypeI32:
+				original := uint32(100)
+				binary.LittleEndian.PutUint32(mem.Buffer[baseOffset+o.Arg.Offest:], original)
```
codefromthecrypt, line 1465: extract `mem.Buffer[baseOffset+o.Arg.Offest:]` with a variable?

wasm/jit/jit_amd64_test.go
```diff
@@ -1402,6 +1402,97 @@ func TestAmd64Builder_handleCall(t *testing.T) {
...
+				expValue = uint64(original * 2)
+			case wazeroir.SignLessTypeI64:
+				original := uint64(math.MaxUint32 + 123)
```
codefromthecrypt, line 1468: when using magic numbers use a trailing comment or define a variable por favor 🇪🇸

wasm/jit/jit_amd64_test.go
```diff
@@ -1402,6 +1402,97 @@ func TestAmd64Builder_handleCall(t *testing.T) {
...
+			jitcall(
+				uintptr(unsafe.Pointer(&code[0])),
+				uintptr(unsafe.Pointer(eng)),
+				uintptr(unsafe.Pointer(&mem.Buffer[0])),
+			)
+
+			// Load instruction must push the loaded value to the top of the stack,
+			// so the stack pointer must be incremented.
+			require.Equal(t, uint64(1), eng.currentStackPointer)
+			require.Equal(t, expValue, eng.stack[eng.currentStackPointer-1])
```
codefromthecrypt, line 1491: comment above is good about what the current stack pointer is, but since we are reading the pointer-1 does it imply currentStackPointer is more like "next stack pointer?"

codefromthecrypt (approved):
another round of Q/A in pull request format!

wasm/jit/jit_amd64.go
mathetake: I prefer as-is because the prefix `generalPurposeRegisterType*` is actually the type name :D

wasm/jit/jit_amd64.go
mathetake: > so assuming signless is not the same as unsigned

well, signless is the same as unsigned, so I think it is fine. (Maybe we should rethink about the name "Signless")

codefromthecrypt: hmm unsigned is far more common language than signless, and it is used in the wasm spec.. if they are the same, how about preferring unsigned then?

wasm/jit/jit_amd64_test.go
mathetake: "stack pointer" in general means the next free address in the stack, so the word "stack pointer" itself means next in general!

wasm/jit/jit_amd64.go
mathetake: will raise a PR to rename shortly

wasm/jit/jit_amd64_test.go
codefromthecrypt: works for me... do we need the word "current" then?

codefromthecrypt:
great job!

wasm/jit/jit_amd64_test.go
mathetake: that would be less redundant! will raise a PR!

PR #88, opened 2021-12-23
jit: remove 'current' prefixes on engine

mathetake:
per https://github.com/tetratelabs/wazero/pull/86#discussion_r774401112

### DIFF
--- a/wasm/jit/engine.go
+++ b/wasm/jit/engine.go
@@ -17,9 +17,9 @@ type engine struct {
 	// The actual Go-allocated stack.
 	// Note that we NEVER edit len or cap in JITed code so we won't get screwed when GC comes in.
 	stack []uint64
-	// Wasm stack pointer on .stack field which is accessed by currentStackBasePointer+currentStackPointer
-	currentStackPointer     uint64
-	currentStackBasePointer uint64
+	// Wasm stack pointer on .stack field which is accessed by stackBasePointer+stackPointer
+	stackPointer     uint64
+	stackBasePointer uint64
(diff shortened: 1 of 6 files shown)


wasm/jit/engine.go
```diff
@@ -17,9 +17,9 @@ type engine struct {
 	// The actual Go-allocated stack.
 	// Note that we NEVER edit len or cap in JITed code so we won't get screwed when GC comes in.
 	stack []uint64
-	// Wasm stack pointer on .stack field which is accessed by currentStackBasePointer+currentStackPointer
-	currentStackPointer     uint64
-	currentStackBasePointer uint64
+	// Wasm stack pointer on .stack field which is accessed by stackBasePointer+stackPointer
+	stackPointer     uint64
+	stackBasePointer uint64
```
codefromthecrypt, line 22: add a comment on base pointer? since it is referenced in the above...

codefromthecrypt (approved):
I'm currently of the opinion, this is a good change

I'm of the opinion this is a good change!

wasm/jit/engine.go
codefromthecrypt: ex what is the base? is it the program running's initial address? (just sayin)

mathetake: https://github.com/tetratelabs/wazero/pull/88/commits/066b4b5f0ae8a53340327ce0f4871689147b94b7 commented (note that "base pointer" or stack base pointer is a general term in calling conventions)

PR #91, opened 2021-12-24
Uses authoritative name for token String()

codefromthecrypt:


wasm/wat/token.go
```diff
@@ -68,7 +68,7 @@ const (
 
 	// tokenID is a sequence of idChar characters prefixed by a '$':
 	//
-	// For example, in the below, '$y' is an id:
+	// For example, in the below, '$y' is an ID:
```
codefromthecrypt, line 71: in english, though I think it is still worth using ID for id 😎

mathetake (approved):
img

PR #92, opened 2021-12-24
Adds initial WebAssembly Text Format (%.wat) parser

codefromthecrypt:
This builds on the lexing layer to parse the WebAssembly Text Format.

The PR after this will convert the following text format to a `wasm.Module`

```wat
(module
	(start $hello)
)
```

### DIFF
--- a/wasm/wat/parser.go
+++ b/wasm/wat/parser.go
@@ -6,9 +6,8 @@ import (
 )
 
 type ModuleParser struct {
-	source                     []byte
-	tm                         *textModule
-	parenDepth, skipUntilDepth int
+	source []byte
+	tm     *textModule
 	// currentStringCount allows us to unquote the textImport.module and textImport.name fields, differentiating empty
 	// from never set, without making textImport.module and textImport.name pointers.
 	currentStringCount int
@@ -42,19 +41,9 @@ func ParseModule(source []byte) (*textModule, error) {
 // reaching a final tokenRParen. This implies nested paren handling.
 type fieldHandler func(fieldName []byte) (tokenParser, error)
 
+// parse calls the delegate ModuleParser.tokenParser
 func (p *ModuleParser) parse(tok tokenType, tokenBytes []byte, line, col int) error {
-	if p.skipUntilDepth == 0 {
-		return p.tokenParser(tok, tokenBytes, line, col)
-	}
-	if tok == tokenLParen {
-		p.parenDepth = p.parenDepth + 1
-	} else if tok == tokenRParen {
-		if p.parenDepth == p.skipUntilDepth {
-			p.skipUntilDepth = 0
-		}
-		p.parenDepth = p.parenDepth - 1
-	}
-	return nil
+	return p.tokenParser(tok, tokenBytes, line, col)
 }
 
 func (p *ModuleParser) startField(tok tokenType, tokenBytes []byte, _, _ int) error {
@@ -203,7 +192,6 @@ func (p *ModuleParser) startFile(tok tokenType, _ []byte, _, _ int) error {
 	if tok != tokenLParen {
 		return fmt.Errorf("expected '(', but found %s", tok)
 	}
-	p.parenDepth = p.parenDepth + 1
 	p.tokenParser = p.startField
 	p.fieldHandler = p.startModule
 	return nil
--- a/wasm/wat/parser.go
+++ b/wasm/wat/parser.go
@@ -160,7 +160,7 @@ func (p *ModuleParser) parseImport(tok tokenType, tokenBytes []byte, _, _ int) e
 	return nil
 }
 
-func (p *ModuleParser) startImportField(fieldName []byte) (tokenParser, error) {
+func (p *ModuleParser) importFieldHandler(fieldName []byte) (tokenParser, error) {
 	switch string(fieldName) {
 	case "func":
 		p.currentField = fieldModuleImportFunc
--- a/wasm/wat/parser.go
+++ b/wasm/wat/parser.go
@@ -1,0 +1,230 @@
+package wat
+
+import (
+	"fmt"
+)
+
+type ModuleParser struct {
+	source                     []byte
+	m                          *textModule
+	parenDepth, skipUntilDepth int
+	// currentStringCount allows us to unquote the Import.module and Import.name fields, differentiating empty from
+	// never set, without making Import.module and Import.name pointers.
+	currentStringCount int
+	pf                 fieldHandler // TODO: crappy name
+	pt                 parseToken   // TODO: crappy name
+	// afterInlining sets the scope to return to after parsing a function.
+	// This is needed because a function can be defined at module scope or an inner scope such as an import.
+	// TODO: https://www.w3.org/TR/wasm-core-1/#abbreviations%E2%91%A8
+	afterInlining parseToken
+	currentImport *textImport
+	currentFunc   *textFunc
+}
+
+// fieldHandler returns a parseToken index that resumes parsing after "($fieldName". This must handle all tokens until
+// reaching a final tokenRParen. This implies nested paren handling.
+type fieldHandler func(fieldName []byte) (parseToken, error)
+
+func (p *ModuleParser) parse(tok tokenType, tokenBytes []byte, line, col int) error {
+	if p.skipUntilDepth == 0 {
+		return p.pt(tok, tokenBytes, line, col)
+	}
+	if tok == tokenLParen {
+		p.parenDepth = p.parenDepth + 1
+	} else if tok == tokenRParen {
+		if p.parenDepth == p.skipUntilDepth {
+			p.skipUntilDepth = 0
+		}
+		p.parenDepth = p.parenDepth - 1
+	}
+	return nil
+}
+
+func (p *ModuleParser) startField(tok tokenType, tokenBytes []byte, _, _ int) error {
+	if tok != tokenKeyword {
+		return fmt.Errorf("%s has a %s where a field name was expected", p.errorContext(), tok)
+	}
+
+	np, err := p.pf(tokenBytes)
+	if err != nil {
+		return err
+	}
+	p.pt = np
+	return nil
+}
+
+func (p *ModuleParser) startModule(fieldName []byte) (parseToken, error) {
+	if string(fieldName) == "module" {
+		return p.parseModule, nil
+	} else {
+		return nil, fmt.Errorf("%s should not have the field: %s", p.errorContext(), string(fieldName))
+	}
+}
+
+func (p *ModuleParser) parseModule(tok tokenType, tokenBytes []byte, _, _ int) error {
+	switch tok {
+	case tokenID:
+		name := string(tokenBytes)
+		if p.m.name != "" {
+			return fmt.Errorf("%s has a redundant name: %s", p.errorContext(), name)
+		}
+		p.m.name = name
+	case tokenLParen:
+		p.pt = p.startField       // after this look for a field name
+		p.pf = p.startModuleField // this defines the field names accepted
+		return nil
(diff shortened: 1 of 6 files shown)


wasm/wat/module_test.go
```diff
@@ -0,0 +1,130 @@
...
+			name:     "with func",
+			input:    "(module (func $dummy))",
+			expected: &Module{Functions: []*Function{{ID: "$dummy"}}},
+		},
+		{
+			name:  "simple example",
+			input: string(simpleExample),
+			expected: &Module{
+				Imports:   []*Import{{Module: "\"\"", Name: "\"hello\"", Desc: &Function{ID: "$hello"}}},
+				Functions: []*Function{{ /* TODO (export "run") (call $hello) */ }},
```
codefromthecrypt, line 46: this is what I'm working on now..

wasm/wat/parser.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+type ModuleParser struct {
+	source                     []byte
+	m                          *textModule
+	parenDepth, skipUntilDepth int
+	// currentStringCount allows us to unquote the Import.module and Import.name fields, differentiating empty from
+	// never set, without making Import.module and Import.name pointers.
+	currentStringCount int
+	pf                 fieldHandler // TODO: crappy name
```
codefromthecrypt, line 14: I will address this and the below TODO prior to marking for review (other TODO will remain to keep line count down)

wasm/wat/parser_test.go
```diff
@@ -0,0 +1,217 @@
...
+			expectedLine: 1,
+			expectedCol:  19,
+			expectedErr:  "import[1] is missing its name",
+		},
+		{
+			name:         "import unquoted module",
+			input:        []byte("(module (import foo bar))"),
+			expectedLine: 1,
+			expectedCol:  17,
+			expectedErr:  "import[1] has an unexpected keyword: foo",
```
codefromthecrypt, line 130: I will also fix this as the index is wrong

codefromthecrypt:
ok really ready now. I removed all code except what's required at the moment

wasm/wat/parser.go
```diff
@@ -0,0 +1,209 @@
...
+	// currentStringCount allows us to unquote the _import.module and _import.name fields, differentiating empty
+	// from never set, without making _import.module and _import.name pointers.
+	currentStringCount int
+	fieldHandler       fieldHandler
+	tokenParser        tokenParser
+	// afterInlining sets the scope to return to after parsing a function.
+	// This is needed because a function can be defined at module scope or an inlined scope such as an import.
+	// TODO: https://www.w3.org/TR/wasm-core-1/#abbreviations%E2%91%A8
+	afterInlining tokenParser
+	currentImport *_import
```
mathetake, line 20: Need comments!

wasm/wat/parser.go
```diff
@@ -0,0 +1,209 @@
...
+	switch string(fieldName) {
+	case "func":
+		p.currentImport.importFunc = &importFunc{}
+		p.afterInlining = p.parseImport
+		return p.parseImportFunc, nil
+	}
+	return nil, fmt.Errorf("unexpected field: %s", string(fieldName))
+}
+
+func (p *ModuleParser) parseImportFunc(tok tokenType, tokenBytes []byte, _, _ int) error {
```
mathetake, line 145: needs TODO comment on handling function types!

wasm/wat/parser.go
```diff
@@ -0,0 +1,209 @@
...
+		p.currentStringCount = 0
+		p.tokenParser = p.parseModule
+	default:
+		return p.unexpectedToken(tok, tokenBytes)
+	}
+	return nil
+}
+
+func (p *ModuleParser) startImportField(fieldName []byte) (tokenParser, error) {
+	switch string(fieldName) {
```
mathetake, line 136: needs TODO comment for handling other imports!

wasm/wat/types.go
```diff
@@ -0,0 +1,83 @@
...
+	// See https://www.w3.org/TR/wasm-core-1/#binary-namesec
+	name string
+
+	// TODO: typeuse https://www.w3.org/TR/wasm-core-1/#text-typeuse
+}
+
+// _import corresponds to the text format of a WebAssembly import.
+//
+// See https://www.w3.org/TR/wasm-core-1/#imports%E2%91%A0
+type _import struct { // note: this is named _import because import is reserved in golang
```
mathetake, line 53: How about?

```suggestion
type importField struct { // note: this is named _import because import is reserved in golang
```

mathetake (approved):

wasm/wat/parser.go
```diff
@@ -0,0 +1,209 @@
+package wat
+
+import (
+	"errors"
+	"fmt"
+)
+
+type ModuleParser struct {
```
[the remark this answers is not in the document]
codefromthecrypt: sg

wasm/wat/parser.go
```diff
@@ -0,0 +1,209 @@
...
+	source []byte
+	module *module
+	// currentStringCount allows us to unquote the _import.module and _import.name fields, differentiating empty
+	// from never set, without making _import.module and _import.name pointers.
+	currentStringCount int
+	fieldHandler       fieldHandler
+	tokenParser        tokenParser
```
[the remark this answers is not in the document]
codefromthecrypt: good idea. I will comment this

wasm/wat/parser.go
```diff
@@ -0,0 +1,209 @@
...
+		p.currentImport = nil
+		p.currentStringCount = 0
+		p.tokenParser = p.parseModule
+	default:
+		return p.unexpectedToken(tok, tokenBytes)
+	}
+	return nil
+}
+
+func (p *ModuleParser) startImportField(fieldName []byte) (tokenParser, error) {
```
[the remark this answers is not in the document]
codefromthecrypt: sure thing!

wasm/wat/types.go
codefromthecrypt: only thing is we use the word import field to describe a field in the import ex func. ex importFieldHandler handles func

codefromthecrypt:
ok I did another pass.

wasm/wat/types.go
codefromthecrypt: once I had `moduleImport` but I reverted it. exporting was the other way to avoid `_`, but then also raises another question. it is a crappy thing that we can name a string variable "string", but we can't name a type "import" :p

codefromthecrypt: ok I found another way out, which is to group import types together. For example, `importFunc` doesn't clash with the keyword `import``

wasm/wat/types.go
```diff
@@ -0,0 +1,83 @@
...
+	//
+	// Note: When in raw numeric form, this is relative to imports. See wasm.Module StartSection for more.
+	//
+	// See https://www.w3.org/TR/wasm-core-1/#start-function%E2%91%A4
+	startFunction string
+}
+
+// importFunc corresponds to the text format of a WebAssembly function import.
+//
+// Note: nothing is required per specification. Ex `(import "" "" (func))` is valid!
```
mathetake, line 33: Didn't know that!

wasm/wat/types.go
```diff
@@ -0,0 +1,83 @@
...
+	//
+	// Note: This is not necessarily the module.name, so it does not need to begin with '$'!
+	module string
+
+	// name is the possibly empty entity name to import. Ex. "" or "PI"
+	//
+	// Note: This is not necessarily the funcName, so it does not need to begin with '$'!
+	name string
+
+	// name is optional and starts with '$'. For example, "$main".
```
mathetake, line 50: I see everything is optional
```suggestion
	// funcName is starts with '$'. For example, "$main".
```

wasm/wat/types.go
```diff
@@ -0,0 +1,83 @@
...
+	//
+	// Note: The name may also be stored in the wasm.Module CustomSection under the key "name" subsection 0.
+	// See https://www.w3.org/TR/wasm-core-1/#binary-namesec
+	name string
+
+	// importFuncs are imports describing functions "(import... (func...))" added in insertion order.
+	importFuncs []*importFunc
+
+	// startFunction is the function to call during wasm.Store Instantiate. The value is a importFunc.name, such as
+	// "$main", or its equivalent raw numeric index, such as "2".
```
mathetake, line 23: wanna confirm that this is the reason why we have string here right?

wasm/wat/parser.go
```diff
@@ -0,0 +1,248 @@
...
+
+func (p *ModuleParser) ensureLParen(tok tokenType, _ []byte, _, _ int) error {
+	if tok != tokenLParen {
+		return fmt.Errorf("expected '(', but found %s", tok)
+	}
+	p.tokenParser = p.startField
+	return nil
+}
+
+func (p *ModuleParser) startField(tok tokenType, tokenBytes []byte, _, _ int) (err error) {
```
mathetake, line 73: wanna know that `Field` is a term in spec (for text format) or specific in this code?

wasm/wat/parser.go
```diff
@@ -0,0 +1,248 @@
...
+		p.tokenParser, err = p.moduleFieldHandler(tokenBytes)
+	case fieldModuleImport:
+		p.tokenParser, err = p.importFieldHandler(tokenBytes)
+	default:
+		return fmt.Errorf("unexpected current field %d", p.currentField)
+	}
+	return
+}
+
+// initialFieldHandler parses the top-level fields in the WebAssembly source.
```
mathetake, line 94:
```suggestion
// initialFieldHandler returns a tokenParser for the top-level fields in the WebAssembly source.
```

wasm/wat/parser.go
codefromthecrypt: field is used a lot in the spec indeed. incidentally, I learned that after looking at Wabt, then verified also as I thought it kind strange term. it isn't consistent but "A module consists of a sequence of fields that can occur in any order." https://www.w3.org/TR/wasm-core-1/#modules%E2%91%A0%E2%91%A2

codefromthecrypt: of course they say "any order" then later say imports must be first 🤷

wasm/wat/types.go
codefromthecrypt: at the moment, yes. A future refactoring maybe we can eagerly map to an integer and drop the name, but we should do that after we start encoding the name custom section.

wasm/wat/types.go
codefromthecrypt: not sure it is intended to be so vague, but heh the grammar allows it

wasm/wat/parser.go
mathetake: 🤷 🧐

wasm/wat/parser.go
```diff
@@ -0,0 +1,248 @@
...
+		p.currentField = fieldModuleStart
+		return p.parseStart, nil
+	}
+	return nil, fmt.Errorf("unexpected field: %s", string(fieldName))
+}
+
+func (p *ModuleParser) parseImport(tok tokenType, tokenBytes []byte, _, _ int) error {
+	switch tok {
+	case tokenString: // Ex. "" or "foo" including quotes!
+		name := string(tokenBytes[1 : len(tokenBytes)-1]) // unquote
```
mathetake, line 138: maybe better add comment about we already ensure that len(tokenBytes) >= 2 in lexer?

mathetake (approved):

wasm/wat/parser.go
codefromthecrypt: sure

codefromthecrypt:
Thanks as always, but always thankful @mathetake!

PR #94, opened 2021-12-28
Adds a new example and adjusts some naming

codefromthecrypt:
This adds a simpler example, which helps understand the basic ins and
outs. This will be used later when the %.wat parser is implemented.

Along the way, I noticed naming inconsistency. Most of the time,
function inputs are named params and outputs are named results. Using
the same vocab helps when mapping the spec to our code.

See https://www.w3.org/TR/wasm-core-1/#function-types%E2%91%A6


mathetake:
> Along the way, I noticed naming inconsistency. Most of the time,
function inputs are named params and outputs are named results. Using
the same vocab helps when mapping the spec to our code.

excellent! 🚀

examples/simple_test.go
```diff
@@ -0,0 +1,42 @@
...
+	buf, err := os.ReadFile("wasm/simple.wasm")
+	require.NoError(t, err)
+	mod, err := wasm.DecodeModule(buf)
+	require.NoError(t, err)
+
+	// Create a new store and add the function "hello" which the module imports
+	store := wasm.NewStore(wazeroir.NewEngine())
+	hello := func(_ *wasm.HostFunctionCallContext) {
+		fmt.Println("Hello from Wazero!")
+	}
```
mathetake, line 31:
```suggestion
	var called bool
	hello := func(_ *wasm.HostFunctionCallContext) {
		called = false
	}
	
	// ...
	
	require.True(t, called)
```

mathetake (approved):

examples/simple_test.go
mathetake: I meant I want to ensure the host function is actually called! :D

wasm/jit/engine.go
```diff
@@ -108,11 +108,11 @@ func (e *engine) Call(f *wasm.FunctionInstance, args ...uint64) (returns []uint6
 		err = fmt.Errorf("function not compiled")
 		return
 	}
-	// Note the top value is the tail of the returns,
-	// so we assign the returns in reverse order.
-	returns = make([]uint64, len(f.Signature.ReturnTypes))
-	for i := range returns {
-		returns[len(returns)-1-i] = e.pop()
+	// Note the top value is the tail of the results,
+	// so we assign the results in reverse order.
```
codefromthecrypt, line 112:
```suggestion
	// so we assign them in reverse order.
```

wasm/store.go
```diff
@@ -486,6 +481,18 @@ func (s *Store) buildFunctionInstances(module *Module, target *ModuleInstance) (
 	return rollbackFuncs, nil
 }
 
+// getFunctionName gets the name of the function corresponding to the codeIndex.
+func getFunctionName(functionNames map[uint32]string, importedFunctionCount int, codeIndex int) string {
```
codefromthecrypt, line 485: I didn't verify this with the compiler flag, but I assume it is inlineable

examples/simple_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/tetratelabs/wazero/wasm"
+	"github.com/tetratelabs/wazero/wasm/wazeroir"
+)
+
+// Test_Simple implements a basic function in go ("hello"), called by a function defined Web Assembly ("run"):
+func Test_Simple(t *testing.T) {
+	// Decode simple.wasm which was pre-compiled from this text format:
+	//	(module
+	//		(import "" "hello" (func $hello))
```
mathetake, line 19: I didn't know that imports with empty module name is valid 😄

examples/simple_test.go
codefromthecrypt: ah good point! right now it is only a visual cue.

examples/simple_test.go
codefromthecrypt: nutty right? it is all over the wasm spec tests!

codefromthecrypt:
I switched the assertion to be more hello

mathetake (approved):

codefromthecrypt:
Thanks for the help @mathetake 🙇

PR #95, opened 2021-12-28
Simplifies simple example and clarifies the StartSection

codefromthecrypt:
A particularly confusing point is that function indices in Wasm are
preceded by imports. Hopefully, this helps reduce that confusion.


wasm/module.go
```diff
@@ -35,11 +35,16 @@ type Module struct {
...
 	GlobalSection   []*GlobalSegment
 	ExportSection   map[string]*ExportSegment
-	StartSection    *uint32
-	ElementSection  []*ElementSegment
-	CodeSection     []*CodeSegment
-	DataSection     []*DataSegment
-	CustomSections  map[string][]byte
+	// StartSection is a function index value when set.
+	//
+	// Note: The function index space is preceded by imported functions, but not all elements in the ImportSection are
```
mathetake, line 40: fyi, this is not only true for function index space, but also table, globals memory space. According to the spec: 

> The index space for functions, tables, memories and globals includes respective imports declared in the same module. The indices of these imports precede the indices of other definitions in the same index space.

mathetake (approved):

mathetake:
Coolio!

wasm/module.go
codefromthecrypt: good point! I will add the text for the others, too!

wasm/module.go
```diff
@@ -28,18 +28,60 @@ func (r *reader) Read(p []byte) (n int, err error) {
...
+	GlobalSection []*GlobalSegment
+	ExportSection map[string]*ExportSegment
+	// StartSection is the index of a function to call before returning from Store.Instantiate.
+	//
+	// Note: The function index space begins with any ImportKindFunction in ImportSection, then the FunctionSection.
+	// For example, if there are two imported functions and three defined in this module, the index space is five.
+	// Note: This is a pointer to avoid conflating no start section with the valid index zero.
+	StartSection   *uint32
+	ElementSection []*ElementSegment
+	// CodeSection is index-correlated with FunctionSection and contains each functions' locals and body.
```
codefromthecrypt, line 80:
```suggestion
	// CodeSection is index-correlated with FunctionSection and contains each function's locals and body.
```

wasm/module.go
```diff
@@ -28,20 +28,57 @@ func (r *reader) Read(p []byte) (n int, err error) {
...
+	// See https://www.w3.org/TR/wasm-core-1/#function-section%E2%91%A0
 	FunctionSection []uint32
-	TableSection    []*TableType
-	MemorySection   []*MemoryType
-	GlobalSection   []*GlobalSegment
-	ExportSection   map[string]*ExportSegment
-	// StartSection is a function index value when set.
+	// TableSection contains each table defined in this module.
+	//
+	// Table indexes are offset by any imported tables because the table index space begins with imports, followed by
```
mathetake, line 50: maybe worth noting in MVP, multiple tables are not supported!

wasm/module.go
```diff
@@ -28,20 +28,57 @@ func (r *reader) Read(p []byte) (n int, err error) {
...
+	// ones defined in this module. For example, if there are two imported tables and three defined in this module, the
+	// table at index 3 to be defined in this module at TableSection[0].
+	//
+	// See https://www.w3.org/TR/wasm-core-1/#table-section%E2%91%A0
+	TableSection []*TableType
+	// MemorySection contains each memory defined in this module.
+	//
+	// Memory indexes are offset by any imported memories because the memory index space begins with imports, followed
```
mathetake, line 58: maybe worth noting in MVP, multiple memories are not supported!

wasm/module.go
codefromthecrypt: ah right. and that MVP is spec 1.0 😎

mathetake (approved):

codefromthecrypt:
thanks for the guidance online and offline @mathetake

PR #96, opened 2021-12-30
jit: support for less than equal for floats.

mathetake:

### DIFF
--- a/wasm/jit/jit_amd64.go
+++ b/wasm/jit/jit_amd64.go
@@ -1038,64 +1038,56 @@ func (b *amd64Builder) handleSub(o *wazeroir.OperationSub) error {
 }
 
 func (b *amd64Builder) handleLe(o *wazeroir.OperationLe) error {
-	var resultConditionState conditionalRegisterState
-	var instruction obj.As
-	var tp generalPurposeRegisterType
-	switch o.Type {
-	case wazeroir.SignedTypeInt32:
-		resultConditionState = conditionalRegisterStateLE
-		instruction = x86.ACMPL
-		tp = generalPurposeRegisterTypeInt
-	case wazeroir.SignedTypeUint32:
-		resultConditionState = conditionalRegisterStateBE
-		instruction = x86.ACMPL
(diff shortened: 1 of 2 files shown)


wasm/jit/jit_amd64_test.go
```diff
@@ -1176,6 +1176,138 @@ func TestAmd64Builder_handleLe(t *testing.T) {
...
+			x1, x2 float64
+			exp    bool
+		}{
+			{x1: 100, x2: -1.1, exp: false},
+			{x1: -1, x2: 100, exp: true},
+			{x1: 100, x2: 200, exp: true},
+			{x1: 100.01234124, x2: 100.01234124, exp: true},
+			{x1: 100.01234124, x2: -100.01234124, exp: false},
+			{x1: 200.12315, x2: 100, exp: false},
+			{x1: math.Inf(1), x2: 100, exp: false},
```
codefromthecrypt, line 1256: should we add normal number that requires 64-bit precision also?

wasm/jit/jit_amd64.go
```diff
@@ -1038,65 +1038,57 @@ func (b *amd64Builder) handleSub(o *wazeroir.OperationSub) error {
...
-		tp = generalPurposeRegisterTypeInt
+		prog.As = x86.ACMPQ
+		prog.From.Reg = x1.register
+		prog.To.Reg = x2.register
 	case wazeroir.SignedTypeFloat32:
-		tp = generalPurposeRegisterTypeFloat
-		panic("add test!")
+		resultConditionState = conditionalRegisterStateBE
+		prog.As = x86.ACOMISS
+		prog.From.Reg = x2.register
```
codefromthecrypt, line 1080: maybe it is intuitive to people more familiar, but I found it odd that the from/to are permuted when a float..

codefromthecrypt (approved):
Looking good, can't find much to suggest as it is quite clean!

mathetake:
@codefromthecrypt thanks for the look!

PR #97, opened 2021-12-30
jit: add more test cases for add/sub operations.

mathetake:
This completes the support of add/sub operations for all types!


codefromthecrypt (approved):
what makes these easy to review is that the "check the stack" values make sense.

PR #101, opened 2021-12-31
Removes suffix Type from fields in FunctionType

codefromthecrypt:
Since there is no ambiguity in FunctionType about it being types, this
removes the suffix Type from its fields. I wouldn't do this, if for
example, there were mixed fields (ex type names). However, as it is, I
think this is less stuttering so overall better.


mathetake (approved):
img

mathetake (approved):
img

PR #102, opened 2021-12-31
Update Go cache in Pre-commit

mathetake:
Looks like a cache is so old that it doesn't contain any cache for goimports and golangci-lint

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -30,7 +30,7 @@ jobs:
           path: |
             ~/.cache/go-build
             ~/go/pkg/mod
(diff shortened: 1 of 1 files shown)


mathetake:
let me re-kick after the first check finished

mathetake:
### Before
Screen Shot 2021-12-31 at 10 18 07

### After
Screen Shot 2021-12-31 at 10 17 57

mathetake:
hmm though strange that we’ve updated go.sum

codefromthecrypt (approved):
yes this is totally strange, but at the end of the day it is a black box we are dealing with. tainting the expression may be the only work around for now. We can try removing it later

PR #103, opened 2021-12-31
jit: support for all comparison instructions

mathetake:


codefromthecrypt (approved):
fantastico

no nits worth the breath!

PR #105, opened 2021-12-31
TypeIndexPtr -> FuncTypeIndex

codefromthecrypt:
It is easier to reason with the function type index as a number instead
of a pointer. It is also easier to unit test. We don't need the second
pointer guard because the import definition already checks the type (ex
func vs table) first.

This also fixes a typo


mathetake (approved):
img

codefromthecrypt:
thanks for the look again!

PR #106, opened 2021-12-31
Rename Go source dirs for Wasms to testdata.

mathetake:
Previously we couldn't, for example `go get ./...` as these directories contain Go sources only valid for TinyGO: 

```
$ go get ./...                                
# github.com/tetratelabs/wazero/bench/case
bench/case/case.go:23:6: missing function body
bench/case/case.go:33:3: cannot use uintptr(bufSize) (type uintptr) as type int in field value
bench/case/case.go:34:3: cannot use uintptr(bufSize) (type uintptr) as type int in field value
# github.com/tetratelabs/wazero/examples/wasm
examples/wasm/file_system.go:8:6: main redeclared in this block
        /home/mathetake/gasm/examples/wasm/fibonacci.go:3:6: previous declaration
examples/wasm/host_func.go:10:6: main redeclared in this block
        /home/mathetake/gasm/examples/wasm/file_system.go:8:6: previous declaration
examples/wasm/host_func.go:22:6: missing function body
examples/wasm/host_func.go:32:3: cannot use uintptr(bufSize) (type uintptr) as type int in field value
examples/wasm/host_func.go:33:3: cannot use uintptr(bufSize) (type uintptr) as type int in field value
examples/wasm/stdio.go:10:6: main redeclared in this block
        /home/mathetake/gasm/examples/wasm/host_func.go:10:6: previous declaration
examples/wasm/trap.go:3:6: main redeclared in this block
        /home/mathetake/gasm/examples/wasm/stdio.go:10:6: previous declaration
```

This commit renames these directories as `testdata` which is recognized and ignored by official Go toolchain.


mathetake:
this will allow us to check that this library is compilable for other targets than amd64 and Linux via `GOOS=linux GOARCH=arm64 go build ./...`

codefromthecrypt (approved):
good idea!

PR #108, opened 2021-12-31
Completes basic implementation of wat.TextToBinary

codefromthecrypt:
The completes the minimal features required to run the following:

```
(module
	(start $hello)
)
```

This hasn't been tuned, yet, but is able to perform the conversion in
just over a microsecond on my laptop:
```
goos: darwin
goarch: amd64
pkg: github.com/tetratelabs/wazero/wasm/wat
BenchmarkTextToBinaryExample
BenchmarkTextToBinaryExample/vs_utf8.Valid
BenchmarkTextToBinaryExample/vs_utf8.Valid-16         	79973917	        14.30 ns/op	       0 B/op	       0 allocs/op
BenchmarkTextToBinaryExample/vs_wasmtime.Wat2Wasm
BenchmarkTextToBinaryExample/vs_wasmtime.Wat2Wasm-16  	  181326	      6061 ns/op	     128 B/op	       3 allocs/op
BenchmarkTextToBinaryExample/vs_wasm.DecodeModule
BenchmarkTextToBinaryExample/vs_wasm.DecodeModule-16  	  890084	      1310 ns/op	     949 B/op	      37 allocs/op
BenchmarkTextToBinaryExample/vs_wat.lex
BenchmarkTextToBinaryExample/vs_wat.lex-16            	10745912	       108.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkTextToBinaryExample/vs_wat.parseModule
BenchmarkTextToBinaryExample/vs_wat.parseModule-16    	 1322486	       851.6 ns/op	     512 B/op	      21 allocs/op
BenchmarkTextToBinaryExample/TextToBinary
BenchmarkTextToBinaryExample/TextToBinary-16          	  864080	      1165 ns/op	     952 B/op
```


examples/simple_test.go
```diff
@@ -3,25 +3,22 @@ package examples
...
 // Test_Simple implements a basic function in go: hello. This is imported as the Wasm name "$hello" and run on start.
-//	(module
-//		(import "" "hello" (func $hello))
-//		(start $hello)
-//	)
 func Test_Simple(t *testing.T) {
-	buf, err := os.ReadFile("wasm/simple.wasm")
-	require.NoError(t, err)
-	mod, err := wasm.DecodeModule(buf)
+	mod, err := wat.TextToBinary([]byte(`(module
```
codefromthecrypt, line 18: yay! though don't change this source to do anything else, yet :p

mathetake:
Super!!!!!!!!!!! (will review tomorrow :D)

examples/simple_test.go
mathetake: :heart:

wasm/wat/convert_test.go
```diff
@@ -0,0 +1,214 @@
...
+		})
+	}
+}
+
+func TestTextToBinary_Errors(t *testing.T) {
+	tests := []struct{ name, input, expectedErr string }{
+		{
+			name:        "start, but no funcs",
+			input:       "(module (start $main))",
+			expectedErr: "1:16: unknown function name: $main in module.start",
```
codefromthecrypt, line 117: First time using line/col outside the scope of the tokenParser function

wasm/wat/convert.go
```diff
@@ -0,0 +1,66 @@
...
+// an error occurs.
+//
+// Here's a description of the return values:
+// * result is the module parsed or nil on error
+// * err is a FormatError invoking the parser, dangling block comments or unexpected characters.
+func TextToBinary(source []byte) (result *wasm.Module, err error) {
+	var m *module
+	m, err = parseModule(source)
+	if err != nil {
+		return nil, err
```
jcchavezs, line 20: I think it is a good practice telling what were you trying to accomplish when the error happened e.g. `fmt.Errorf("failed to parse module: %v", err)` otherwise errors would look so random.

wasm/wat/convert.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+	}
+
+	// TODO: encode CustomSection["name"] with module function and local names
+	return
+}
+
+func parseStartSection(m *module) (*uint32, error) {
+	start := m.startFunction
+	if parsed, e := strconv.ParseUint(start.index, 0, 32); e == nil {
```
jcchavezs, line 47: worth to add what happens when `e==nil`, I guess index is numeric and hence we return it directly`. An small nit would be to use `parsingErr` instead of `e`, I had to double check the return types of `ParseUint` to be sure what `e` was.

wasm/wat/convert.go
codefromthecrypt: parseModule already says this, so I think we are good. ex `1:30: redundant funcidx: $main in module.start`

wasm/wat/convert.go
codefromthecrypt: sure I'll add a comment here agree!

wasm/wat/convert.go
codefromthecrypt: what I mean to say is that if we wrap we will lose the format and not gain anything in this case, but I totally agree we should think about error context and if it has it already or not. thanks for making me think

codefromthecrypt: ok! I added a comment. hopefully it works or make new line comments por favor!

codefromthecrypt:
thanks for the feedback @jcchavezs I was lazy in hindsight in comments, and it took a while to build the context. better to comment so that building context doesn't steal everyone's time also!

wasm/wat/convert_test.go
mathetake: maybe we have `${filename}:${line}:${col}` format as that is recognizable by some editors and we could just click it to jump to the position (e.g. VSCode)

mathetake: should

mathetake: ah - filename does not exist in this context!

mathetake: nvm

wasm/wat/convert.go
```diff
@@ -0,0 +1,83 @@
...
+// an error occurs.
+//
+// Here's a description of the return values:
+// * result is the module parsed or nil on error
+// * err is a FormatError invoking the parser, dangling block comments or unexpected characters.
+func TextToBinary(source []byte) (result *wasm.Module, err error) {
+	// First, attempt to parse the module into a basic structure representing the text format. If this errs, return
+	// immediately without wrapping parseModule returns FormatError, which is pre-formatted.
+	var m *module
+	if m, err = parseModule(source); err != nil {
```
mathetake, line 20: why not just ?
```suggestion
	m, err = parseModule(source)
	if err != nil {
```

wasm/wat/convert_test.go
codefromthecrypt: yeah that's why I exported `FormatError`, so that the formatting can be changed even if the default formatting could be ok for inlined text (which doesn't have a filename)

wasm/wat/convert.go
codefromthecrypt: because that adds an IDE warning about re-using the name `err` from the return type

mathetake: sg!

mathetake (approved):

codefromthecrypt:
Thanks for the feedback @jcchavezs and @mathetake! This is hardly done so will be improving further in the many following pull requests to implement more features including functions with instructions. If you have more to say on this, I'll weave it into one of those.

PR #109, opened 2022-01-01
Adds custom name section encoder

codefromthecrypt:
I'd like to test the output vs another tool, but it seems like perhaps the name section is encoded differently in wabt, ex maybe they strip the leading '$'. I'd like to get a binary dump of some name section from maybe wasm-tools and/or wabt to compare both against the spec and the values encoded for the same input.

```
(module $simple
	(start $hello)
)
```

### DIFF
--- a/wasm/leb128/leb128.go
+++ b/wasm/leb128/leb128.go
@@ -17,6 +17,30 @@ var (
 	errOverflow64 = errors.New("overflows a 64-bit integer")
 )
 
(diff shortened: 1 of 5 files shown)


mathetake (approved):

codefromthecrypt:
Next PR will move this into the binary encoding package so that we can do round-trip testing. When I do that I'll add some examples from wabt and/or wasm-tools, not so much about what's in the spec in the binary encoding, but what's actually encoded (ex "$main" vs "main")

PR #110, opened 2022-01-02
ci: include Job ID and Go version in the cache key.

mathetake:

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -30,7 +30,7 @@ jobs:
           path: |
             ~/.cache/go-build
             ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ hashFiles('**/go.sum') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/commit.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
           path: |
             ~/.cache/go-build
             ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-20211231
+          key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ hashFiles('**/go.sum') }}
```
mathetake, line 33: we should use different key than benches as that is the main cause of golangci-lint cache non-existence

mathetake:
strange `{{ env.GITHUB_JOB }}` seems empty..

codefromthecrypt:
maybe you can't reference by env.X in this scope?

mathetake:
yeah got it https://david.gardiner.net.au/2021/07/github-actions-default-environment-variables.html

codefromthecrypt (approved):

PR #111, opened 2022-01-02
jit: support for mul

mathetake:


wasm/jit/jit_amd64.go
```diff
@@ -766,7 +766,128 @@ func (c *amd64Compiler) compileSub(o *wazeroir.OperationSub) error {
...
+	c.addInstruction(prog)
+
+	// We no longer need x2 register after ADD operation here,
+	// so we release it.
+	c.locationStack.releaseRegister(x2)
+	return nil
+}
+
+func (c *amd64Compiler) compileMul(o *wazeroir.OperationMul) (err error) {
+	switch o.Type {
```
codefromthecrypt, line 785: It might be helpful to comment here that those unfamiliar with x86 instructions can look at https://www.felixcloutier.com/x86/index.html for explanations, noting why these are prefixed 'A' :p

wasm/jit/jit_amd64.go
```diff
@@ -766,7 +766,128 @@ func (c *amd64Compiler) compileSub(o *wazeroir.OperationSub) error {
...
+		err = c.compileMulForFloats(x86.AMULSS)
+	case wazeroir.UnsignedTypeF64:
+		err = c.compileMulForFloats(x86.AMULSD)
+	}
+	return
+}
+
+func (c *amd64Compiler) compileMulForInts(is32Bit bool, mulInstruction obj.As) error {
+	// See https://www.felixcloutier.com/x86/mul if unfamiliar with the convention for
+	// integer multiplication. In summary, the destination operand must be on RAX
```
codefromthecrypt, line 800: because RAX register is redundant, we should either remove the R or use the same form as the constant ex REG_AX

```suggestion
	// integer multiplication. In summary, the destination operand must be on the AX
```

wasm/jit/jit_amd64.go
```diff
@@ -766,7 +766,128 @@ func (c *amd64Compiler) compileSub(o *wazeroir.OperationSub) error {
...
+	// register and the overflow info is stored in RDX. So, we have to ensure that
```
codefromthecrypt, line 801:
```suggestion
	// register and the overflow info is stored in REG_DX. So, we have to ensure that
```

wasm/jit/jit_amd64.go
```diff
@@ -766,7 +766,128 @@ func (c *amd64Compiler) compileSub(o *wazeroir.OperationSub) error {
...
+	// 1) Previously located value on RDX must be saved to memory stack.
```
codefromthecrypt, line 802: I'll make comments like this to keep things consistent. We can later think if we want to make our own constants instead that map to these, in which case we can use RDX instead of REG_DX when abbreviating

```suggestion
	// 1) Previously located value on REG_DX must be saved to memory stack.
```

wasm/jit/jit_amd64.go
```diff
@@ -766,7 +766,128 @@ func (c *amd64Compiler) compileSub(o *wazeroir.OperationSub) error {
...
+	// 2) One of the operands (x1 or x2) must be on RAX register.
```
codefromthecrypt, line 803:
```suggestion
	// 2) One of the operands (x1 or x2) must be on REG_AX (the AX register).
```

wasm/jit/jit_amd64.go
```diff
@@ -766,7 +766,128 @@ func (c *amd64Compiler) compileSub(o *wazeroir.OperationSub) error {
...
+
+	// We have to make sure that at this point the operands must be on registers.
+	if err := c.ensureOnGeneralPurposeRegister(x2); err != nil {
+		return err
+	}
+	if err := c.ensureOnGeneralPurposeRegister(x1); err != nil {
+		return err
+	}
+
+	// We have to save the existing value on RDX.
```
codefromthecrypt, line 854:
```suggestion
	// We have to save the existing value on REG_DX.
```

wasm/jit/jit_amd64.go
```diff
@@ -1532,6 +1653,7 @@ func (c *amd64Compiler) moveStackToRegister(loc *valueLocation) {
 	c.addInstruction(prog)
 }
 
+// Move the valuie on a conditional register to a free general purpose register.
```
codefromthecrypt, line 1656: less typos is you don't convert the function name to English 😎 
```suggestion
```

wasm/jit/jit_amd64.go
```diff
@@ -1532,6 +1653,7 @@ func (c *amd64Compiler) moveStackToRegister(loc *valueLocation) {
...
 func (c *amd64Compiler) moveConditionalToGeneralPurposeRegister(loc *valueLocation) error {
```
codefromthecrypt, line 1657:
```suggestion
func (c *amd64Compiler) moveConditionalToFreeGeneralPurposeRegister(loc *valueLocation) error {
```

wasm/jit/jit_amd64.go
```diff
@@ -1542,10 +1664,17 @@ func (c *amd64Compiler) moveConditionalToGeneralPurposeRegister(loc *valueLocati
 		return fmt.Errorf("conditional register mov requires a free register")
 	}
 
+	c.moveConditionalRegister(loc, reg)
+	return nil
+}
+
+// Move the valuie on a conditional register to the given general purpose register.
```
codefromthecrypt, line 1671:
```suggestion
```

wasm/jit/jit_amd64.go
```diff
@@ -1542,10 +1664,17 @@ func (c *amd64Compiler) moveConditionalToGeneralPurposeRegister(loc *valueLocati
...
+func (c *amd64Compiler) moveConditionalRegister(loc *valueLocation, reg int16) {
```
codefromthecrypt, line 1672:
```suggestion
func (c *amd64Compiler) moveConditionalToGeneralPurposeRegister(loc *valueLocation, reg int16) {
```

wasm/jit/jit_amd64.go
```diff
@@ -1683,7 +1811,7 @@ func (c *amd64Compiler) callFunctionFromRegisterIndex(reg int16) {
 
 func (c *amd64Compiler) releaseAllRegistersToStack() {
 	used := len(c.locationStack.usedRegisters)
-	for i := len(c.locationStack.stack) - 1; i >= 0 && used > 0; i-- {
+	for i := uint64(0); i < c.locationStack.sp && used > 0; i++ {
```
codefromthecrypt, line 1814: We generally ok on  `i++`? I don't remember why but some time ago someone said in go we are supposed to never use that and always laboriously do `i = i+1`

wasm/jit/jit_amd64.go
```diff
@@ -1752,6 +1880,18 @@ func (c *amd64Compiler) setFunctionCallIndexFromConst(index int64) {
 	c.addInstruction(prog)
 }
 
+// ensureRegisterUnused ensures that a given register is not used by
+// any value. If the register is currently used, release the value
+// to the stack.
+func (c *amd64Compiler) ensureRegisterUnused(reg int16) {
```
codefromthecrypt, line 1886: maybe this and then you can drop the doc?

```suggestion
func (c *amd64Compiler) onValueReleaseRegisterToStack(reg int16) {
```

wasm/jit/jit_amd64_test.go
```diff
@@ -2301,6 +2301,326 @@ func TestAmd64Compiler_compileSub(t *testing.T) {
...
 }
 
+func TestAmd64Compiler_compileMul(t *testing.T) {
+	t.Run("int32", func(t *testing.T) {
+		for _, tc := range []struct {
+			name         string
+			x1Reg, x2Reg int16
+		}{
+			{
+				name:  "x1:rax,x2:random_reg",
```
codefromthecrypt, line 2311: if we like to use `rax` vs `reg_ax` let's make a constant file now then everything can have parity. We don't need to be beholdent to the imported names I think, but we should be consistent.

wasm/jit/jit_amd64_test.go
```diff
@@ -2301,6 +2301,326 @@ func TestAmd64Compiler_compileSub(t *testing.T) {
...
+			t.Run(tc.name, func(t *testing.T) {
+				const x1Value uint32 = 1 << 11
+				const x2Value uint32 = 51
+				const dxValue uint64 = 111111
+
+				eng := newEngine()
+				compiler := requireNewCompiler(t)
+				compiler.initializeReservedRegisters()
+
+				// Put the value on DX register -- the value must be saved to memory in compileMul.
```
codefromthecrypt, line 2361:
```suggestion
				// Pretend there was an existing value on the DX register. We expect compileMul to save this to the stack.
```

wasm/jit/jit_amd64_test.go
```diff
@@ -2301,6 +2301,326 @@ func TestAmd64Compiler_compileSub(t *testing.T) {
...
+					uintptr(unsafe.Pointer(&code[0])),
+					uintptr(unsafe.Pointer(eng)),
+					0,
+				)
+				fmt.Println(eng.stack[:3])
+
+				// Check the stack.
+				require.Equal(t, uint64(2), eng.stackPointer)
+				require.Equal(t, uint64(x1Value*x2Value), eng.stack[eng.stackPointer-1])
+				// The value on DX must be saved to stack.
```
codefromthecrypt, line 2406:
```suggestion
				// Verify the previous DX register value was saved to the stack.
```

wasm/jit/jit_amd64.go
```diff
@@ -766,7 +766,128 @@ func (c *amd64Compiler) compileSub(o *wazeroir.OperationSub) error {
...
+	prog.As = instruction
+	c.addInstruction(prog)
+
+	// We no longer need x2 register after ADD operation here,
+	// so we release it.
+	c.locationStack.releaseRegister(x2)
+	return nil
+}
+
+func (c *amd64Compiler) compileMul(o *wazeroir.OperationMul) (err error) {
```
codefromthecrypt, line 784:
```suggestion
// compileMul blah blah and saves any prior value on the DX register to the stack
func (c *amd64Compiler) compileMul(o *wazeroir.OperationMul) (err error) {
```

wasm/jit/jit_amd64_test.go
```diff
@@ -2301,6 +2301,326 @@ func TestAmd64Compiler_compileSub(t *testing.T) {
...
+				require.Equal(t, dxValue, eng.stack[eng.stackPointer-2])
+			})
+		}
+	})
+	t.Run("int64", func(t *testing.T) {
+		for _, tc := range []struct {
+			name string
+			// Interpret -1 as stack.
```
codefromthecrypt, line 2414: this is awkward, but fixing it is laborious until we refactor out more test helper methods. OK for now

wasm/jit/jit_amd64_test.go
```diff
@@ -2301,6 +2301,326 @@ func TestAmd64Compiler_compileSub(t *testing.T) {
...
+			t.Run(tc.name, func(t *testing.T) {
+				const x1Value uint64 = 1 << 35
+				const x2Value uint64 = 51
+				const dxValue uint64 = 111111
+
+				eng := newEngine()
+				compiler := requireNewCompiler(t)
+				compiler.initializeReservedRegisters()
+
+				// Put the value on DX register. The value must be saved to memory in compileMul.
```
codefromthecrypt, line 2468:
```suggestion
				// Pretend there was an existing value on the DX register. We expect compileMul to save this to the stack.
``````

File: wasm/jit/jit_amd64_test.go [hunk 19]
```diff
@@ -2301,6 +2301,326 @@ func TestAmd64Compiler_compileSub(t *testing.T) {
...
+					uintptr(unsafe.Pointer(&code[0])),
+					uintptr(unsafe.Pointer(eng)),
+					0,
+				)
+				fmt.Println(eng.stack[:3])
+
+				// Check the stack.
+				require.Equal(t, uint64(2), eng.stackPointer)
+				require.Equal(t, uint64(x1Value*x2Value), eng.stack[eng.stackPointer-1])
+				// The value on DX register must be saved to stack.
```
codefromthecrypt, line 2513:
```suggestion
				// Verify the previous DX register value was saved to the stack.
```

wasm/jit/jit_value_location_amd64.go
```diff
@@ -250,3 +250,15 @@ func (s *valueLocationStack) takeStealTargetFromUsedRegister(tp generalPurposeRe
 	}
 	return nil, false
 }
+
+// findValueForRegister finds the valueLocation which uses the given register.
```
codefromthecrypt, line 254:
```suggestion
// findValueForRegister returns the valueLocation of the given register or nil if not found.
```

codefromthecrypt (approved):
Looks good. I'm starting to notice an opportunity for consistency though it wasn't introduced here. We can think about it now or later.

wasm/jit/jit_amd64.go
codefromthecrypt: generally I think it would be neat to start doing docs on the compileMul which minimally say at least the stack post conditions. I know preconditions are very complicated, but where simple enough those also.

Ex.

```
// compileMul adds instructions to multiply two operands which may be on the stack or registers.
// After execution, the following results will be on the stack:
// * stackPointer-1: result of the multiplication
// * stackPointer-2: the prior value of the DX register
```

wasm/jit/jit_amd64.go
mathetake: https://github.com/tetratelabs/wazero/pull/111/commits/1a057d8f22077c92adff6857f5b7a3405224eea5

wasm/jit/jit_amd64.go
mathetake: > some time ago someone said in go we are supposed to never use that and always laboriously do i = i+1

I don't think so, we abuse `i++` or `i--` everywhere 😄

wasm/jit/jit_amd64_test.go
mathetake: agreed!

wasm/jit/jit_amd64_test.go
mathetake: `reg_ax` is a variable name only exiting in x86 assembler pkg, but the true name is `RAX`. So I will stack to RAX or whatever in doc or comments

wasm/jit/jit_amd64.go
mathetake: noted!

codefromthecrypt:
yay! I found the long form tedious

wasm/jit/jit_amd64.go
```diff
@@ -781,6 +783,129 @@ func (c *amd64Compiler) compileSub(o *wazeroir.OperationSub) error {
 	return nil
 }
 
+// compileMul adds instructions to multiply two operands which may be on the stack or registers.
+// After execution, the result of multiplication pushed onto the stack.
```
codefromthecrypt, line 787: that the prior RDX is pushed on the stack.. is that intuitive enough to assume as being a result of multiplication?

codefromthecrypt (approved):
approve!

wasm/jit/jit_amd64.go
mathetake: actually RDX is not pushed on the stack, RDX holds the overflow info according to the manual, but we don't need to use it in Wasm convention

mathetake:
@codefromthecrypt thanks for the thorough review!

PR #112, opened 2022-01-02
Delete simple.wasm

mathetake:
simple.wasm is now unused as it is replaced by inline wat in simple_test.go


codefromthecrypt:
yeah that's true!

codefromthecrypt (approved):

PR #115, opened 2022-01-04
Matches how wabt encodes the binary name section in wat2wasm

codefromthecrypt:
This PR is the result of hours of trying to figure out how people use
the custom name section in practice. The time loss was largely caused
by it being underspecified and also undocumented in implementations.


wasm/wat/names_test.go
```diff
@@ -10,38 +10,38 @@ func TestEncodeNameSection(t *testing.T) {
 	m, err := parseModule(simpleExample)
```
mathetake, line 10: Maybe we should add case where module name doesn’t exist but function names exist in thhe custom section?

mathetake: Actually that happens for TinyGo produce such section and it is of course valid!

mathetake (approved):
🏌️ 💯

wasm/wat/names_test.go
codefromthecrypt: oh cool. will do!

codefromthecrypt:
thanks for the pointers @mathetake!

PR #117, opened 2022-01-06
Text Format: reduces the amount of parsing state

codefromthecrypt:
This generifies some parser state so that it can be reused next for
parsing parameter names. This also cleans up some ++ stuff.


mathetake (approved):

codefromthecrypt:
thanks for the look @mathetake and PS I measured and it is same perf and allocation count. If we tried to track names with string pointers, that increases allocations, so taking advantage of the fact we already have []byte here, we end up a wash even with the less code.

PR #120, opened 2022-01-07
wat: support import functions with inlined types

codefromthecrypt:
This begins support of real functions as before we hard-coded all are
nullary (void in void out). As there's a lot of refactoring left to do,
this starts only by supporting inlined type (signatures) for imported
functions. This does not handle inlined parameter names which are a bit
complicated so would make the PR too hard to review.

### DIFF
--- a/wasm/wat/parser.go
+++ b/wasm/wat/parser.go
@@ -440,9 +440,12 @@ func (p *moduleParser) errorContext() string {
 		return "module"
 	case fieldModuleStart:
 		return "module.start"
-	case fieldModuleImport, fieldModuleImportFunc, fieldModuleImportFuncParam:
+	case fieldModuleImport, fieldModuleImportFunc, fieldModuleImportFuncParam, fieldModuleImportFuncResult:
 		if p.currentField == fieldModuleImportFuncParam {
 			return fmt.Sprintf("module.import[%d].func.param[%d]", p.currentImportIndex, p.currentParamIndex)
+		}
+		if p.currentField == fieldModuleImportFuncResult {
+			return fmt.Sprintf("module.import[%d].func.result", p.currentImportIndex)
 		}
 		if p.currentField == fieldModuleImportFunc {
 			return fmt.Sprintf("module.import[%d].func", p.currentImportIndex)
--- a/wasm/wat/parser.go
+++ b/wasm/wat/parser.go
@@ -345,7 +345,7 @@ func (p *moduleParser) parseImportFuncParam(tok tokenType, tokenBytes []byte, _,
 		}
 		p.currentType.params = append(p.currentType.params, p.currentParams...)
 
-		// since multiple param fields are valid, prepare for any next.
+		// since multiple param fields are valid, ex `(func (param i32) (param i64))`, prepare for any next.
 		p.currentParams = nil
 		p.currentParamIndex++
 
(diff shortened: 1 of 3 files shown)


wasm/wat/parser.go
```diff
@@ -80,39 +101,101 @@ func (p *moduleParser) ensureLParen(tok tokenType, _ []byte, _, _ uint32) error
...
+		p.tokenParser = p.parseImport
+	case fieldModuleStart, fieldModuleImport:
 		p.currentField = fieldModule
-		return p.parseModule, nil
-	} else {
-		return nil, fmt.Errorf("unexpected field: %s", string(fieldName))
+		p.tokenParser = p.parseModule
+	case fieldModule:
+		p.currentField = fieldInitial
+		p.tokenParser = p.parseUnexpectedToken // only one module is allowed and nothing else
```
codefromthecrypt, line 194: while there's more to do, at least grouping begin and end allows us to raise this error now

wasm/wat/parser.go
```diff
@@ -222,24 +283,126 @@ func (p *moduleParser) parseImportFunc(tok tokenType, tokenBytes []byte, _, _ ui
...
 		}
-		p.module.importFuncs[len(p.module.importFuncs)-1].typeIndex = 0
-		p.currentField = fieldModuleImport
-		p.tokenParser = p.parseImport
+
+		fn := p.module.importFuncs[len(p.module.importFuncs)-1]
+
+		// search for an existing signature that matches the current type
+		for i, t := range p.module.types {
+			if bytes.Equal(p.currentType.params, t.params) && bytes.Equal(p.currentType.results, t.results) {
```
codefromthecrypt, line 298: this is where it is helpful you used a type alias @mathetake!

codefromthecrypt (commented):
notes

wasm/wat/parser.go
```diff
@@ -222,24 +283,126 @@ func (p *moduleParser) parseImportFunc(tok tokenType, tokenBytes []byte, _, _ ui
...
+			return errors.New("expected a type")
+		}
+
+		// add the parameters to the current type
+		if p.currentType == nil {
+			p.currentType = &typeFunc{}
+		}
+		p.currentType.params = append(p.currentType.params, p.currentParams...)
+
+		// since multiple param fields are valid, prepare for any next.
```
mathetake, line 348: sorry what do you mean by "multiple param fields"? any example?

mathetake (approved):
Brilliant!

wasm/wat/parser.go
mathetake: 🕺

wasm/wat/parser.go
codefromthecrypt: good question!

```suggestion
		// since multiple param fields are valid, ex `(func (param i32) (param i64)`, prepare for any next.
```

wasm/wat/parser.go
```diff
@@ -345,7 +345,7 @@ func (p *moduleParser) parseImportFuncParam(tok tokenType, tokenBytes []byte, _,
 		}
 		p.currentType.params = append(p.currentType.params, p.currentParams...)
 
-		// since multiple param fields are valid, prepare for any next.
+		// since multiple param fields are valid, ex `(func (param i32) (param i64))`, prepare for any next.
```
mathetake, line 348: I think we should add test case for it (I see no case to cover this)

codefromthecrypt: I also noticed some more tests missing just now. will fix thanks!

PR #123, opened 2022-01-10
wat: enforces signature order

codefromthecrypt:
This change does two things:
 * enforces field names are first or not present
   * Ex. (func $main) $main is the name of the func field
 * enforces the order requirements of the type grammar. (param)* result?
   * To reduce complexity, this logic is extracted to type_parser.

### Unit test coverage is the same
There is still 100pct coverage except TODO and panics

### benchmark speed decrease and allocation increase

Extracting typeParser resulted in a performance decrease on the trivial benchmark because it initializes, but never needs a type parser. I think this is ok because in real life, %.wat files with anything non-trivial will end up with at least a few types.

decrease in speed
```
BenchmarkTextToBinaryExample/vs_wat.parseModule-16    	 1429618	       832.1 ns/op	     568 B/op	      23 allocs/op
BenchmarkTextToBinaryExample/vs_wat.parseModule-16    	 1293886	       977.1 ns/op	     664 B/op	      27 allocs/op
```

example input
```
(module $simple
	(start $hello)
)
```

Once we implement functions, we should re-do the benchmark anyway.

### Function pointer fake
The parser used by the lexer changes when fields change. It now also changes when we enter a common grammar (type use). To reduce stack depth and allocations, I'm updating moduleParser.tokenParser via a pointer to the moduleParser as it seems there's no function pointer in go.

### DIFF
--- a/wasm/wat/convert.go
+++ b/wasm/wat/convert.go
@@ -25,7 +25,11 @@ func TextToBinary(source []byte) (result *wasm.Module, err error) {
 	// difference is that the text format has type names and the binary format does not.
 	result = &wasm.Module{}
 	for _, t := range m.types {
-		result.TypeSection = append(result.TypeSection, &wasm.FunctionType{Params: t.params, Results: t.results})
+		var results []wasm.ValueType
+		if t.result != 0 {
+			results = []wasm.ValueType{t.result}
+		}
+		result.TypeSection = append(result.TypeSection, &wasm.FunctionType{Params: t.params, Results: results})
 	}
 
 	// Now, handle any imported functions. Notably, we retain the same insertion order as defined in the text format in
@@ -35,7 +39,7 @@ func TextToBinary(source []byte) (result *wasm.Module, err error) {
 			Module: f.module, Name: f.name,
 			Desc: &wasm.ImportDesc{
 				Kind:          wasm.ImportKindFunction,
-				FuncTypeIndex: uint32(f.typeIndex),
+				FuncTypeIndex: f.typeIndex,
 			},
 		})
 	}
--- a/wasm/wat/parser_test.go
+++ b/wasm/wat/parser_test.go
@@ -278,6 +278,11 @@ func TestParseModule_Errors(t *testing.T) {
 			expectedErr: "1:15: found ')' before '('",
 		},
 		{
+			name:        "module name after import",
+			input:       "(module (import \"\" \"\" (func) $Math)",
+			expectedErr: "1:30: unexpected id: $Math in module.import[0]",
+		},
+		{
 			name:        "import missing module",
 			input:       "(module (import))",
 			expectedErr: "1:16: missing module and name in module.import[0]",
--- a/wasm/wat/parser_test.go
+++ b/wasm/wat/parser_test.go
@@ -209,32 +209,6 @@ func TestParseModule(t *testing.T) {
 	}
 }
 
-func TestParseValueType(t *testing.T) {
-	tests := []struct {
-		input    string
-		expected wasm.ValueType
-	}{
-		{"i32", wasm.ValueTypeI32},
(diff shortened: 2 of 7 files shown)


wasm/wat/convert.go
```diff
@@ -25,7 +25,11 @@ func TextToBinary(source []byte) (result *wasm.Module, err error) {
 	// difference is that the text format has type names and the binary format does not.
 	result = &wasm.Module{}
 	for _, t := range m.types {
-		result.TypeSection = append(result.TypeSection, &wasm.FunctionType{Params: t.params, Results: t.results})
+		var results []wasm.ValueType
+		if t.result != 0 {
```
codefromthecrypt, line 29: Since we are enforcing only one result per signature in the parser anyway, I made the tracking simpler by using a single value instead.

wasm/wat/parser_test.go
```diff
@@ -348,7 +355,7 @@ func TestParseModule_Errors(t *testing.T) {
 		{
 			name:        "import func double result",
 			input:       "(module (import \"\" \"\" (func (param i32) (result i32) (result i32))))",
-			expectedErr: "1:55: redundant result field in module.import[0].func",
```
codefromthecrypt, line 351: saying redundant result field would be too much logic vs below. one important thing this PR does is prepares us to re-use the logic implementing the type use in imported functions for both module type and function parsing. This will be next.

codefromthecrypt:
To check my assumptions about performance, I ran benchmarks against a file that has two types in it. The perf is basically the same, while both more strict enforcing grammar and decoupled code vs before.


before
```
BenchmarkTextToBinaryExample/vs_wat.parseModule
BenchmarkTextToBinaryExample/vs_wat.parseModule-16    	  433776	      2401 ns/op	    1200 B/op	      66 allocs/op
BenchmarkTextToBinaryExample/TextToBinary
BenchmarkTextToBinaryExample/TextToBinary-16          	  387208	      2809 ns/op	    1744 B/op	      77 allocs/op
```

Now
```
BenchmarkTextToBinaryExample/vs_wat.parseModule
BenchmarkTextToBinaryExample/vs_wat.parseModule-16    	  480280	      2411 ns/op	    1184 B/op	      65 allocs/op
BenchmarkTextToBinaryExample/TextToBinary
BenchmarkTextToBinaryExample/TextToBinary-16          	  409278	      2960 ns/op	    1728 B/op	      78 allocs/op
```

```
(module
	(import "wasi_snapshot_preview1" "path_open" (func $runtime.path_open (param i32 i32 i32 i32 i32 i64 i64 i32 i32) (result i32)))
	(import "wasi_snapshot_preview1" "fd_write" (func $runtime.fd_write (param i32 i32 i32 i32) (result i32)))
)
```

wasm/wat/type_parser.go
```diff
@@ -0,0 +1,206 @@
...
+		vt = wasm.ValueTypeI64
+	case "f32":
+		vt = wasm.ValueTypeF32
+	case "f64":
+		vt = wasm.ValueTypeF64
+	default:
+		return 0, fmt.Errorf("unknown type: %s", t)
+	}
+	return vt, nil
+}
```
mathetake, line 206:
```suggestion
func parseValueType(tokenBytes []byte) (vt wasm.ValueType, err error) {
	t := string(tokenBytes)
	switch t {
	case "i32":
		vt = wasm.ValueTypeI32
	case "i64":
		vt = wasm.ValueTypeI64
	case "f32":
		vt = wasm.ValueTypeF32
	case "f64":
		vt = wasm.ValueTypeF64
	default:
		err = fmt.Errorf("unknown type: %s", t)
	}
	return
}
```

mathetake (approved):

codefromthecrypt:
Thanks for the look.. more to come!

PR #125, opened 2022-01-10
wat: refactors in preparation of module defined types

codefromthecrypt:
This refactors the code in preparation of parsing module types. There is
a lot of nuance around type signatures, so this restructures the code to
fit that idea and enhances the documentation in preparation.


codefromthecrypt:
PS this takes an early punch on state tracking, so perf dipped a little to track the indexes even if we aren't yet using them.

Ex full conversion on trivial example before:
```
1233 ns/op	    1096 B/op	      35 allocs/op
```

now
```
1370 ns/op	    1232 B/op	      38 allocs/op
```

Once we can parse the whole structure, I will revisit perf as a whole as I think there are enough common patterns to improve it later

mathetake (approved):

codefromthecrypt:
image

PR #128, opened 2022-01-12
wat: refactors symbolic index identifier handling

codefromthecrypt:
Before, we punted symbolic index binding until we converted into binary
format. This put logic in an awkward spot, which was ok if only the
start function, but not for generic index handling (such as type
binding).

This pulls a new file out for handling bindings, and also eagerly
validates numbers when we read an index. Most importantly, this binds
symbolic identifiers to numeric offsets on parse, which matches how
they work at runtime.

### DIFF
--- a/wasm/wat/bind.go
+++ b/wasm/wat/bind.go
@@ -1,0 +1,47 @@
+package wat
+
+import (
+	"fmt"
+)
+
+// bindIndices ensures any indices point are numeric or returns a FormatError if they cannot be bound.
+func bindIndices(m *module) error {
+	if m.startFunction != nil {
+		if err := bindStartFunction(m); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
looks like github is having a bad day

```
 Download action repository 'actions/setup-go@v2' (SHA:424fc82d43fa5a37540bae62709ddcc23d9520d4)
Download action repository 'actions/checkout@v2' (SHA:ec3a7ce113134d7a93b817d10a8272cb61118579)
Error: '/Users/runner/work/_actions/_temp_be395fa4-9b3a-468c-9e68-e23f8decbe7d/4d026012-194f-4f32-9d20-87c90bfee65d.tar.gz' contains '0' directories
```

mathetake (approved):

codefromthecrypt:
thanks for the look @mathetake

image

PR #129, opened 2022-01-12
wat: adds support for type parsing

codefromthecrypt:
This adds support for parsing module types including index resolution.
This also changes the benchmark example to use more features, so we can
notice the cost of each.


wasm/wat/bind.go
```diff
@@ -6,42 +6,103 @@ import (
 
 // bindIndices ensures any indices point are numeric or returns a FormatError if they cannot be bound.
 func bindIndices(m *module) error {
+	typeToIndex := map[*typeFunc]uint32{}
+	typeNameToIndex := map[string]uint32{}
+	for i, t := range m.typeFuncs {
+		if t.name != "" {
+			typeNameToIndex[t.name] = uint32(i)
+		}
+		typeToIndex[t] = uint32(i)
```
mathetake, line 15:
```suggestion
		index := uint32(i)
		if t.name != "" {
			typeNameToIndex[t.name] = index
		}
		typeToIndex[t] = index
```

mathetake (approved):

PR #130, opened 2022-01-13
wat: adds param name parsing

codefromthecrypt:
parameter name parsing is a little tricky due to a number of spec
features:
 * you can choose to only name some parameters and not others
 * abbreviated params cannot be named
 * you can mix named params with abbreviated names
 * override rules are needed when inlined params are named for a type
   index which is also named.

Note: this does not encode the name section because it is currently
unverifiable due to missing features in both wabt and wasm-tools. Next
PR will be a wasm encoder so we can proceed with accurrate benchmarks.


wasm/wat/parser_test.go
```diff
@@ -11,20 +11,25 @@ import (
 // example has a work-in-progress of supported functionality, used primarily for benchmarking. This includes:
 // * module and function names
 // * explicit, and inlined type definitions (including anonymous)
+// * inlined and overridden type parameter names
 // * start function
 //
-// NOTE: this is different from exampleWat because the parser doesn't yet support all features
+// Note: This is different from exampleWat because the parser doesn't yet support all features.
+// Note: When changing this, regenerate example.wasm like `wat2wasm --debug-names --debug-parser -v example.wat`
+// TODO: this has drifted because neither wabt, nor wasm-tools properly encodes the name section.
```
codefromthecrypt, line 19: so basically `names.go` hasn't been updated and neither has `testdata/example.wasm` I'll do that in the next PR

mathetake (approved):
:1000:

PR #133, opened 2022-01-14
Adds CustomNameSection codec

codefromthecrypt:
This adds `CustomNameSection` and codec operations so that we can
decouple the text format from the binary one.
    
Note: This does not completely implement name decoding. It stops at
converting the old logic as the change became very big. I will follow-up
later to finish that.


wasm/names_test.go
```diff
@@ -0,0 +1,129 @@
...
+	//		(start $hello)
+	//	)
+	require.Equal(t, []byte{
+		SectionIDCustom,
+		0x1d, // 24 bytes after this point
+		0x04, // the custom second name "name" is 4 bytes long
+		'n', 'a', 'm', 'e',
+		0x00, // module subsection ID zero
+		0x07, // 7 bytes to follow
+		0x06, // the module name simple is 6 bytes long
```
codefromthecrypt, line 39: due to these changes, to add the section prefix, and the changes to append an empty local subsection, the git diff considers this a new file

wasm/section.go
```diff
@@ -93,16 +93,16 @@ func (m *Module) readSectionCustom(r *reader, sectionSize int) error {
...
 	}
-	contentLen := int(sectionSize) - int(nameLenBytes) - int(nameLen)
-	if contentLen < 0 {
+	dataSize := sectionSize - int(nameLenSize) - int(nameLen)
+	if dataSize < 0 {
 		return fmt.Errorf("malformed custom section %s", string(nameBuf))
 	}
-	contents := make([]byte, contentLen)
-	_, err = io.ReadFull(r, contents)
+	data := make([]byte, dataSize)
```
codefromthecrypt, line 100: data is used in the doc, so figured we could

wasm/names.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"github.com/tetratelabs/wazero/wasm/leb128"
+)
+
+// nameSectionPrefix a length-prefixed 'name'. subsection data should follow.
+var nameSectionPrefix = encodeName("name")
+
+// encodeCustomNameSection encodes a possibly empty buffer representing the "name" wasm.Module CustomSection.
+// See https://www.w3.org/TR/wasm-core-1/#name-section%E2%91%A0
+func encodeCustomNameSection(m *Module) []byte {
```
mathetake, line 12: Given that `*Module` already contains all the custom sections in module.CustomSections and possibly original `"name"` section there, what's the usecase for this?

wasm/segment.go
```diff
@@ -23,6 +23,13 @@ type ImportDesc struct {
 	// FuncTypeIndex is the index in Module.TypeSection corresponding to this import's FunctionType.
 	// This is only valid when Kind equals ImportKindFunction
 	FuncTypeIndex uint32
+	// FuncName is the optional symbolic identifier read from the text format or SectionIDCustom under the key "name"
+	// subsection 1. Ex. main
+	//
+	// See https://www.w3.org/TR/wasm-core-1/#binary-namesec
+	//
+	// Note: When read from the text format, the leading '$' is stripped to maintain parity with other tools like wabt.
+	FuncName      string
```
mathetake, line 32: maybe I don't want to have this one -- I would like to keep wasm pkg is separated from text representation

mathetake:
LGTM except moving the concepts from Wat to wasm pkg

wasm/segment.go
mathetake: well i want to keep this aligned with the binary representation of ImportDesc in the spec

mathetake: See https://www.w3.org/TR/wasm-core-1/#binary-importdesc

codefromthecrypt:
I think being strict about names is a cure worse than the disease, because that means to access names means serialization each time, or redefining another type that is almost the same as the binary format, just to manifest names. That said, I also don't think it is important enough yet and it can be undone later. So, I'll move the code to be pure binary serialization meaning to compare names == hex editor. If later we find this difficult, I'll do the code again in a different way.

codefromthecrypt:
opened https://github.com/tetratelabs/wazero/issues/134 because I don't agree, this is just to voice an opinion. That I've done that, I'll proceed as requested! Thanks for the look

codefromthecrypt:
I will rewrite this PR as a serializer since there's no way to verify without binary comparison with the current direction. This will result in a larger PR, but that's also cool as we eventually need to serialize anyway.

wasm/names.go
codefromthecrypt: here are the use cases:
 * round tripping test not requiring hex editor and reviewers to know exact encoding rules
 * less performance lost wat to wasm
 * more intuitive debugging as the name serialized by the compiler ex even tinygo is visible

we can address these later, though, so resolving!

codefromthecrypt:
I think the best way to decouple is to make a `NamesEncoder` interface with the implementation where all the other binary serialization logic is (in the wasm package which is currently implicitly the same as the binary format). Then, I can make the wat tests implement that just catching names, which avoids the problem we have now where it has to know the binary format.

mathetake:
Also I think we also should have not only for names, but also for `Encoder` interface like

```
type WasmEncoder struct {
    // original is either WebAssembly text or WebAssembly binary.
    Encode(original []byte) *wasm.Module
}
```

then wasm package should be agnostic to text or binary encoding :D

mathetake:
oh maybe this should be called Decoder

mathetake:
Anyway thank you for putting a lot of thoughts into my comments!

codefromthecrypt:
heh since you are the reviewer I think my next PR update will give you enough.. neither of us are in a position to eagerly serialize back into text format, so let's punt that part until we are farther ahead.

codefromthecrypt:
ok this was completely rewritten PTAL!

wasm/section.go
```diff
@@ -25,24 +25,82 @@ const (
...
+// writeSections appends this module's sections into the buffer in the order required by the specification, custom
+// section first.
+// See https://www.w3.org/TR/wasm-core-1/#modules%E2%91%A0%E2%93%AA
+func (m *Module) encodeSections(buffer []byte) (bytes []byte) {
+	bytes = buffer
+	for name, data := range m.CustomSections {
+		bytes = append(bytes, encodeCustomSection(name, data)...)
+	}
+	if len(m.TypeSection) > 0 {
+		panic("TODO: TypeSection")
```
codefromthecrypt, line 37: note: I used panic here as we will never need to return an error encoding even if we always do on the decode side

wasm/leb128/leb128.go
```diff
@@ -94,26 +94,26 @@ func DecodeInt32(r io.Reader) (ret int32, num uint64, err error) {
...
+			} else if bytesRead == 5 && ret >= 0 && unused != 0x00 {
 				return 0, 0, errOverflow32
 			}
 			return
 		}
 	}
 }
 
-func DecodeInt33AsInt64(r io.Reader) (ret int64, num uint64, err error) {
+func DecodeInt33AsInt64(r io.Reader) (ret int64, bytesRead uint64, err error) {
```
codefromthecrypt, line 116: Is there an int33 type?

wasm/module.go
```diff
@@ -130,61 +136,3 @@ func DecodeModule(binary []byte) (*Module, error) {
 	}
 	return ret, nil
 }
-
-func (m *Module) DecodeCustomNameSection() (map[uint32]string, error) {
```
codefromthecrypt, line 134: ps I ported this logic, but it is left alone. I'll revisit it in another PR so that we can read all the standard subsections

wasm/names_test.go
```diff
@@ -0,0 +1,218 @@
...
+				},
+			},
+			expected: []byte{
+				0x01, // function subsection ID one
+				0x2b, // 43 bytes to follow
+				0x02, // two function names
+				0x00, // the function index is zero
+				0x16, // the function name runtime.args_sizes_get is 22 bytes long
+				'r', 'u', 'n', 't', 'i', 'm', 'e', '.', 'a', 'r', 'g', 's', '_', 's', 'i', 'z', 'e', 's', '_', 'g', 'e', 't',
+				0x01, // the function index is zero
```
codefromthecrypt, line 77:
```suggestion
				0x01, // the function index is one
```

wasm/names_test.go
```diff
@@ -0,0 +1,272 @@
...
+				FunctionNames: map[uint32]string{
+					0x00: "mul",
+					0x01: "add",
+				},
+				LocalNames: map[uint32]map[uint32]string{
+					0x00: {0x00: "x", 0x01: "y"},
+					0x01: {0x00: "l", 0x01: "r"},
+				},
+			},
+			expected: []byte{
```
codefromthecrypt, line 93: this is hard to compare against something else because wasm-tools doesn't encode locals properly and wabt doesn't display nice. However, if you take a best-of you can verify this manually against something besides the spec. Encode the example with `wat2wasm --debug-names foo.wat`, then print it with `wasm-tools dump foo.wasm`

```bash
$ wasm-tools dump foo.wasm
--snip--
0x0031 | 01 0b       | function names
0x0033 | 02          | 2 count
0x0034 | 00 03 6d 75 | Naming { index: 0, name: "mul" }
       | 6c         
0x0039 | 01 03 61 64 | Naming { index: 1, name: "add" }
       | 64         
0x003e | 02 11       | local names
0x0040 | 02          | 2 count
0x0041 | 00          | function 0 locals
0x0042 | 02          | 2 count
0x0043 | 00 01 78    | Naming { index: 0, name: "x" }
0x0046 | 01 01 79    | Naming { index: 1, name: "y" }
0x0049 | 01          | function 1 locals
0x004a | 02          | 2 count
0x004b | 00 01 6c    | Naming { index: 0, name: "l" }
0x004e | 01 01 72    | Naming { index: 1, name: "r" }

**Comment by codefromthecrypt** [comment]
pulled out an issue for codec type refactoring as it is cart before the horse. Highest priority is converting the text format into the binary format, not for the sake of writing it, but because currently all code depends on the binary structure. Once that's done, we can refactor it and probably do it cleaner than now anyway!

https://github.com/tetratelabs/wazero/issues/136

File: wasm/leb128/leb128.go (hunk 6 above)
Comment (mathetake) [reply wasm/leb128/leb128.go:116]: as in the latest spec: https://webassembly.github.io/spec/core/binary/instructions.html#control-instructions

> Unlike any other occurrence, the type index in a block type is encoded as a positive signed integer, so that its signed LEB128 bit pattern cannot collide with the encoding of value types or the special code 0x40, which correspond to the LEB128 encoding of negative integers. To avoid any loss in the range of allowed indices, it is treated as a 33 bit signed integer.

and this is used for decoding "block type" for control instructions which can be multiple post MVP, so I think that's why I implemented this and use in `ReadBlockType` function.  Even though this is overkilling for MVP spec, but I think this is harmless as it is

File: wasm/names.go [hunk 10]
```diff
@@ -0,0 +1,201 @@
...
+	return append(leb128.EncodeUint32(i), encodeSizePrefixed(data)...)
+}
+
+// encodeSizePrefixed encodes the data prefixed by their size.
+func encodeSizePrefixed(data []byte) []byte {
+	size := leb128.EncodeUint32(uint32(len(data)))
+	return append(size, data...)
+}
+
+// uint32Slice implements sort.Interface
```
mathetake, line 136: I think sort.Interface is outdated one, and instead we are recommended to do
```go
var keys []uint32
// appends....
sort.Slice(keys, func (i, j int) bool { return keys[i] < keys[j] } )
```

wasm/names_test.go
```diff
@@ -0,0 +1,272 @@
...
+	name := []byte("hello")
+	require.Equal(t, []byte{byte(index), 5, 'h', 'e', 'l', 'l', 'o'}, encodeNameMapEntry(index, name))
+}
+
+func TestEncodeSizePrefixed(t *testing.T) {
+	// We expect size in bytes (LEB128 encoded) then the bytes
+	require.Equal(t, []byte{5, 'h', 'e', 'l', 'l', 'o'}, encodeSizePrefixed([]byte("hello")))
+}
+
+func TestDecodeCustomNameSection(t *testing.T) {
```
mathetake, line 168: thank you for rewriting test!

mathetake (approved):

wasm/leb128/leb128.go
codefromthecrypt: thanks for giving me the contents for the comment 😎

wasm/names.go
codefromthecrypt: cool. I also found this awkward!

codefromthecrypt:
Thanks for the hand @mathetake! I think the code structure is better now

Issue #83, opened 2021-12-22
Flatten type hierarchy removing single-implementation types


codefromthecrypt:
#79 removed the naivevm interpreter implementation leaving what we called wazeroir. We had to name the other implementation because there were multiple possibilities. In reality the maintenance of multiple stores implies another will never live. Let's try to reduce the amount of types and packages knowing there is only wazeroir. Part of this could be "denaming" wazeroir as it becomes the default and only impl. Any other opportunities to flatten package and type hierarchies, and generally concepts or terms should be taken as well!

codefromthecrypt:
one thing I noticed starting this is that compiler/operations.go has a lot of conceptual overlap with instructions.go most of the symbols have an exact or near match with opcodes, for example. This is probably an area we can be more direct.

codefromthecrypt:
everything is internal now so we can do any of this without impact. in fact, we've already done a lot of this

PR #135, opened 2022-01-15
Initialize module instance with `Exports` map

summerwind:
Hi, Thanks for developing a great wasm runtime!

This PR fixes `panic: assignment to entry in nil map` that occurs when `AddGlobal()`, `AddTableInstance()` or `AddMemoryInstance()` is called before `AddHostFunction()`.

This panic does not occur when `AddHostFunction()` called, because `ModuleInstance` is initialized with `Exports` field in `AddHostFunction()` as follows. This PR makes sure that other functions are initialized with `Exports` field as well.
https://github.com/tetratelabs/wazero/blob/6a309797efa5361f45c6a5cb64d567ecf5ceb16d/wasm/store.go#L1637

### DIFF
--- a/wasm/store_test.go
+++ b/wasm/store_test.go
@@ -1,0 +1,22 @@
+package wasm
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetModuleInstance(t *testing.T) {
+	name := "test"
+
+	// 'jit' and 'wazeroir' package cannot be used because of circular import.
+	// Here we will use 'nil' instead. Should we have an Engine for testing?
+	s := NewStore(nil)
+
+	m1 := s.getModuleInstance(name)
+	assert.Equal(t, m1, s.ModuleInstances[name])
+	assert.NotNil(t, m1.Exports)
+
+	m2 := s.getModuleInstance(name)
+	assert.Equal(t, m1, m2)
+}
--- a/wasm/store_test.go
+++ b/wasm/store_test.go
@@ -3,7 +3,7 @@ package wasm
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestGetModuleInstance(t *testing.T) {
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
Thanks for coming by and fixing this @summerwind (song is in my head now!)

Sorry this file doesn't have unit tests, yet. One way to fix both could be to make a function like this and reuse it everywhere. then make a new file store_test.go which has `TestGetModuleInstance` which has cases for when it exists and doesn't. Wanna give it a try?

```go
// getModuleInstance returns an existing ModuleInstance if exists, or assigns a new one.
func (s *Store) getModuleInstance(name string) *ModuleInstance {
	m, ok := s.ModuleInstances[name]
	if !ok {
		m = &ModuleInstance{Exports: map[string]*ExportInstance{}}
		s.ModuleInstances[name] = m
	}
	return m
}
```

summerwind:
Thanks for the comment!
That's make sense for me. I'll try to add unit test.

summerwind:
@codefromthecrypt Added a unit test as you guided!
Please let me know if there is anything I need to fix.

wasm/store_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetModuleInstance(t *testing.T) {
+	name := "test"
+
+	// 'jit' and 'wazeroir' package cannot be used because of circular import.
+	// Here we will use 'nil' instead. Should we have an Engine for testing?
```
mathetake, line 13: thanks for this comment, and yeah actually this is problematic. We could solve this as part of #83. Meanwhile, I'm fine using nil for this test.

mathetake (approved):
Change makes sense, and LGTM! @codefromthecrypt wanna take final look?

codefromthecrypt (approved):
good! the tests cover the cases of interest (particularly we don't overwrite by accident on second call to getModuleInstance)

wasm/store_test.go
```diff
@@ -0,0 +1,22 @@
...
+	s := NewStore(nil)
+
+	m1 := s.getModuleInstance(name)
+	assert.Equal(t, m1, s.ModuleInstances[name])
```
codefromthecrypt, line 17: nit is we prefer require vs assert as getting errors early helps avoid test panics

summerwind: Thanks for the feedback!
Fixed in a66e636.

mathetake:
Thank you for your contribution! 👯  💯 @summerwind

summerwind:
Thank you!

PR #138, opened 2022-01-17
Centralizes name section encoding and usage

### LINKED ISSUE #134: Custom name section isn't that custom
The web assembly specification points out that "name" is a custom section, yet it is also defined, including the subsections. This is directly mapped to the text format and implementations rely on this knowledge for round-tripping and debug information.

Right now, we leave "name" as a custom section because wasm.Module is defined needing to be the binary representation. OTOH, binary is only one representation, there's also text and the abstract representation. Moreover, implementors are not required to name types exactly to the spec. In other words, we can choose to top level the type in a way that helps including in round-tripping from the text format as well any future debug information.

Aside: Being flexible where it helps implementation isn't uncommon, even in standard implementations. For example, the %.wat parser is actually the wast parser in Wasm-tools, for convenience. That said, this doesn't mean declaring a type for the representation of the name section ends up more useful than decoding it each time, either!

codefromthecrypt:
`Store.Instantiate` needs the name section when building function
instances. Before, we had to lazy decode this as it wasn't decoded along
with the other data defined by the binary format. This acknowledges that
the name section is a standard field and top-levels it in the
`wasm.Module` type. It clarifies that anything else will stay in the
`Module.CustomSections` until we support more of them (perhaps Dwarf).

This also renames `CustomNameSection` to `NameSection` because the
standard refers it that way. It is the "name section", just for whatever
reason, they decided to store it as a custom section.

Nice side effects include centralization logic of binary decoding, and
also no more double-encoding of names from the text format. Before this
change, when converting text to binary, we had to re-encode names back
into the binary format because later `Store.Instantiate` needed to read
them as raw bytes. Doing it this way saves dozens of allocations vs
before, so cool.

This also allows the name section and also the custom sections to be nil
as there's no reason to set them when we have no data.

Fixes #134

### DIFF
--- a/wasm/section.go
+++ b/wasm/section.go
@@ -11,7 +11,7 @@ import (
 type SectionID = byte
 
 const (
-	// SectionIDCustom includes the standard defined CustomNameSection and possibly others not defined in the standard.
+	// SectionIDCustom includes the standard defined NameSection and possibly others not defined in the standard.
 	SectionIDCustom   SectionID = 0
 	SectionIDType     SectionID = 1
 	SectionIDImport   SectionID = 2
@@ -31,6 +31,9 @@ const (
 // See https://www.w3.org/TR/wasm-core-1/#modules%E2%91%A0%E2%93%AA
 func (m *Module) encodeSections(buffer []byte) (bytes []byte) {
 	bytes = buffer
+	for name, data := range m.CustomSections {
+		bytes = append(bytes, encodeCustomSection(name, data)...)
+	}
 	if len(m.TypeSection) > 0 {
 		panic("TODO: TypeSection")
 	}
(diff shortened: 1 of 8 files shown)


wasm/section.go
```diff
@@ -156,16 +159,33 @@ func (m *Module) readSectionCustom(r *reader, sectionSize int) error {
...
+		return fmt.Errorf("malformed custom section %s", name)
 	}
 	data := make([]byte, dataSize)
 	_, err = io.ReadFull(r, data)
 	if err != nil {
 		return fmt.Errorf("cannot read custom section data: %v", err)
 	}
-	m.CustomSections[string(nameBuf)] = data
+
+	// Decode any known custom sections after checking for redundant keys
```
codefromthecrypt, line 174: oops forgot the redundant section tests. will add them

mathetake (approved):
brilliant!

codefromthecrypt:
thanks for the stolen attention from call indirect!

PR #139, opened 2022-01-17
Adds binary encoding of TypeSection

codefromthecrypt:


wasm/type.go
```diff
@@ -11,6 +11,30 @@ type FunctionType struct {
...
+var nullary = []byte{0x60, 0, 0}
+
+var encodedOneParam = buildEncodedFuncTypes(true)
+var encodedOneResult = buildEncodedFuncTypes(false)
+
+// buildEncodedFuncTypes is like buildEncodedValTypes except it is parameter or result specific and includes the func
+// type prefix 0x60.
+//
+// See https://www.w3.org/TR/wasm-core-1/#function-types%E2%91%A4
+func buildEncodedFuncTypes(param bool) (encodedTypes [ValueTypeI32 + 1][]byte) {
```
mathetake, line 23: `[ValueTypeI32 + 1][]byte` seems too big and redundant because `ValueTypeI32 = 0x7f`?

mathetake: maybe `ValueTypeF64 - ValueTypeI32`

wasm/value.go
```diff
@@ -35,6 +35,42 @@ func formatValueType(t ValueType) (ret string) {
...
 
+var noValType = []byte{0}
+
+// encodedValTypes is a cache of size prefixed binary encoding of known val types.
+var encodedValTypes = buildEncodedValTypes()
+
+// buildEncodedValTypes builds results for encodeValTypes for known value types.
+//
+// Note: this is length ValueTypeI32+1 because the largest known val type is that.
+func buildEncodedValTypes() (encodedTypes [ValueTypeI32 + 1][]byte) {
```
mathetake, line 46: same

wasm/type.go
codefromthecrypt: yeah we can make it smaller, but then to access the map entry, we have subtract first

wasm/value.go
codefromthecrypt: it will be a bit slower, but sure!

wasm/type.go
codefromthecrypt: at this point probably just as well to use a switch statement.

wasm/value.go
codefromthecrypt: done!

wasm/value.go
```diff
@@ -19,6 +19,13 @@ const (
 	ValueTypeI64 ValueType = 0x7e
 	ValueTypeF32 ValueType = 0x7d
 	ValueTypeF64 ValueType = 0x7c
+	// valTypeRange allows us to cache common valueType calculations. To determine the array offset, subtract
+	// valTypeFloor from the value of interest.
+	valTypeRange = ValueTypeI32 - valTypeFloor + 1
+	// valTypeFloor is the lowest numeric value of a value type.
+	//
+	// Note: the floor value changes after WebAssembly 1.0 (MVP)
```
codefromthecrypt, line 27: ps only reason I added this is I looked it up before punting to a larger array before, as it changes to https://webassembly.github.io/spec/core/binary/types.html#reference-types

I don't care about it, just made a comment to remind later.

mathetake (approved):

codefromthecrypt:
I redid this as a map, because it is the more natural type and by the time we start checking ranges, etc, the other approach started falling apart.

mathetake (approved):
I like it better!

codefromthecrypt:
Thanks for your patience and advice @mathetake!

PR #140, opened 2022-01-17
Flattens Export and Import and dodges Segment suffix

codefromthecrypt:
This flattens the type hierarchy of Export and Import a little as it
felt easier to reason with considering there are only 4 kinds.

This also dodges the Segment suffix because that only applies to
"element" and "data"

### DIFF
--- a/wasm/module.go
+++ b/wasm/module.go
@@ -28,15 +28,21 @@ func (r *reader) Read(p []byte) (n int, err error) {
 
 // Module is a WebAssembly binary representation.
 // See https://www.w3.org/TR/wasm-core-1/#modules%E2%91%A8
+//
+// Differences from the specification:
+// * The NameSection is decoded, so not present as a key "name" in CustomSections.
 type Module struct {
 	// TypeSection contains the unique FunctionType of functions imported or defined in this module.
 	//
 	// See https://www.w3.org/TR/wasm-core-1/#type-section%E2%91%A0
 	TypeSection []*FunctionType
-	// ImportSection contains any types, tables, memories or globals imported into this module.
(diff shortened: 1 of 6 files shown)


wasm/module.go
```diff
@@ -77,11 +83,11 @@ type Module struct {
...
 	//
 	// See https://www.w3.org/TR/wasm-core-1/#global-section%E2%91%A0
-	GlobalSection []*GlobalSegment
-	ExportSection map[string]*ExportSegment
+	GlobalSection []*Global
+	ExportSection map[string]*Export
 	// StartSection is the index of a function to call before returning from Store.Instantiate.
 	//
-	// Note: The function index space begins with any ImportKindFunction in ImportSection, then the FunctionSection.
+	// Note: The function index space begins with any ImportKindFunc in ImportSection, then the FunctionSection.
```
codefromthecrypt, line 90: went with func here because literally it calls it that

wasm/segment.go
```diff
@@ -8,104 +8,80 @@ import (
...
 	case ImportKindTable:
-		tt, err := readTableType(r)
-		if err != nil {
-			return nil, fmt.Errorf("read table type: %v", err)
+		if i.DescTable, err = readTableType(r); err != nil {
+			return nil, fmt.Errorf("error decoding import table desc: %w", err)
 		}
-		return &ImportDesc{
-			Kind:         ImportKindTable,
-			TableTypePtr: tt,
```
codefromthecrypt, line 54: ps one of the motivations for flattening was reducing the amount of things we have to name. ex now we don't need to say things like Desc.TableTypePtr

mathetake (approved):

codefromthecrypt:
image

PR #142, opened 2022-01-18
Adds wasm.EncodeModule and wasm.DecodeModule

### LINKED ISSUE #136: Refactor out a codec package
Follow-up from https://github.com/tetratelabs/wazero/pull/133#issuecomment-1013800357

Currently, the `wasm` package is conflated with binary format and codec. This has led to some confusion for example, where logic for writing the binary format of names should live. If there was a separate package, this would be a no-brainer as the part that decodes should also have the logic that encodes.

Since later we need to refactor out an internal package, we should probably consider refactoring a text and binary codec package also.

Note: This isn't the highest priority as we don't even support the full text format, yet, and doing this before we have a functional second format is questionable. However, it makes sense to do!

codefromthecrypt:
This moves encoding logic into the packages binary and text, adding a
new top-level function for encoding and decoding:

```go
// DecodeModule parses the configured source into a wasm.Module. This function returns when the source is exhausted or
// an error occurs. The result can be initialized for use via Store.Instantiate.
//
// Here's a description of the return values:
// * result is the module parsed or nil on error
// * err is a FormatError invoking the parser, dangling block comments or unexpected characters.
// See binary.DecodeModule and text.DecodeModule
type DecodeModule func(source []byte) (result *Module, err error)

// EncodeModule encodes the given module into a byte slice depending on the format of the implementation.
// See binary.EncodeModule
type EncodeModule func(m *Module) (bytes []byte)
```

Fixes #136


wasm/store.go
```diff
@@ -343,9 +344,9 @@ func (s *Store) applyFunctionImport(target *ModuleInstance, typeIndex uint32, ex
 		return fmt.Errorf("unknown type for function import")
 	}
 	expectedType := target.Types[typeIndex].Type
-	if !ValueTypesEqual(expectedType.Results, f.FunctionType.Type.Results) {
+	if !bytes.Equal(expectedType.Results, f.FunctionType.Type.Results) {
 		return fmt.Errorf("return signature mimatch: %#x != %#x", expectedType.Results, f.FunctionType.Type.Results)
-	} else if !ValueTypesEqual(expectedType.Params, f.FunctionType.Type.Params) {
+	} else if !bytes.Equal(expectedType.Params, f.FunctionType.Type.Params) {
```
codefromthecrypt, line 349: this is unrelated

wasm/store.go
```diff
@@ -427,13 +428,13 @@ func (s *Store) executeConstExpression(target *ModuleInstance, expr *ConstantExp
 		}
 		return v, ValueTypeI64, nil
 	case OpcodeF32Const:
-		v, err = readFloat32(r)
+		v, err = ieee754.DecodeFloat32(r)
```
codefromthecrypt, line 431: this was just for organization

wasm/binary/section.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func decodeDataSection(r *reader) ([]*wasm.DataSegment, error) {
+	vs, _, err := leb128.DecodeUint32(r)
+	if err != nil {
+		return nil, fmt.Errorf("get size of vector: %w", err)
+	}
+
+	result := make([]*wasm.DataSegment, vs)
+	for i := uint32(0); i < vs; i++ {
```
codefromthecrypt, line 236: beside changing all for loops to be generic, the code is the same as before

wasm/binary/decoder.go
```diff
@@ -0,0 +1,124 @@
...
+		sectionSize, _, err := leb128.DecodeUint32(r)
+		if err != nil {
+			return nil, fmt.Errorf("get size of section for id=%d: %v", sectionID[0], err)
+		}
+
+		sectionContentStart := r.read
+		switch sectionID[0] {
+		case SectionIDCustom:
+			// First, validate the section and determine if the section for this name has already been set
+			name, dataSize, decodeErr := decodeCustomSectionNameAndDataSize(r, sectionSize)
```
codefromthecrypt, line 59: I left this section here as it is the only one special (this avoids having to pass module to each of the type decoders, as it is only needed here)

wasm/binary/decoder.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// DecodeModule implements wasm.DecodeModule for the WebAssembly 1.0 (MVP) Binary Format
+// See https://www.w3.org/TR/wasm-core-1/#binary-format%E2%91%A0
+func DecodeModule(binary []byte) (*wasm.Module, error) {
+	r := &reader{binary: binary, buffer: bytes.NewBuffer(binary)}
+
+	// Magic number.
+	buf := make([]byte, 4)
+	if _, err := io.ReadFull(r, buf); err != nil || !bytes.Equal(buf, magic) {
```
codefromthecrypt, line 32: used bytes.Equal instead of doing it manually as before

codefromthecrypt (commented):
code is the same as before with noted exceptions. just that splitting out into files made diffs useless

mathetake:
I love this!!!

wasm/binary/code.go
```diff
@@ -0,0 +1,76 @@
...
+	"math"
+
+	"github.com/tetratelabs/wazero/wasm"
+	"github.com/tetratelabs/wazero/wasm/leb128"
+)
+
+func decodeCode(r io.Reader) (*wasm.Code, error) {
+	ss, _, err := leb128.DecodeUint32(r)
+	if err != nil {
+		return nil, fmt.Errorf("get the size of code segment: %w", err)
```
mathetake, line 15:
```suggestion
		return nil, fmt.Errorf("get the size of code: %w", err)
```

wasm/binary/section.go
```diff
@@ -0,0 +1,275 @@
...
+func decodeGlobalSection(r *reader) ([]*wasm.Global, error) {
+	vs, _, err := leb128.DecodeUint32(r)
+	if err != nil {
+		return nil, fmt.Errorf("get size of vector: %w", err)
+	}
+
+	result := make([]*wasm.Global, vs)
+	for i := uint32(0); i < vs; i++ {
+		if result[i], err = decodeGlobal(r); err != nil {
+			return nil, fmt.Errorf("read global segment: %v ", err)
```
mathetake, line 165:
```suggestion
			return nil, fmt.Errorf("read global: %v ", err)
```

wasm/module.go
```diff
@@ -115,43 +103,178 @@ type Module struct {
...
+type Export struct {
+	Kind ExportKind
+	// Name is what the host refers to this definition as.
+	Name string
+	// Index is the index of the definition to export, the index namespace is by Kind
+	// Ex. If ExportKindFunc, this is an index to ModuleInstance.Functions
+	Index uint32
+}
+
+type ElementSegment struct {
```
mathetake, line 231: shouldn't we delete `segment` suffix for consistency?

wasm/module.go
```diff
@@ -115,43 +103,178 @@ type Module struct {
...
+	Init       []uint32
+}
+
+type Code struct {
+	NumLocals  uint32
+	LocalTypes []ValueType
+	Body       []byte
+}
+
+type DataSegment struct {
```
mathetake, line 243: same

mathetake (approved):

wasm/module.go
codefromthecrypt: spec calls it this though, so let's leave it. I'll tidy this part up when get to the bottom of the file https://www.w3.org/TR/wasm-core-1/#element-segments%E2%91%A4

wasm/module.go
codefromthecrypt: similar only other place they use the term segment https://www.w3.org/TR/wasm-core-1/#data-segments%E2%91%A4

codefromthecrypt:
thanks for the look and the fixies @mathetake

PR #143, opened 2022-01-18
Consolidates reader code in binary decoder

codefromthecrypt:


wasm/binary/value.go
```diff
@@ -60,20 +61,22 @@ func decodeValueTypes(r io.Reader, num uint32) ([]wasm.ValueType, error) {
 	return ret, nil
 }
 
-func decodeNameValue(r io.Reader) (string, error) {
-	vs, _, err := leb128.DecodeUint32(r)
+// decodeName decodes a size prefixed string from the reader, returning it and the count of bytes read.
```
codefromthecrypt, line 64:
```suggestion
// decodeUTF8 decodes a size prefixed string from the reader, returning it and the count of bytes read.
```

codefromthecrypt:
put a random ordering change I forgot from the last one as it isn't worth its own PR

mathetake (approved):

PR #149, opened 2022-01-20
wat: adds func parsing

codefromthecrypt:
This adds function parsing, consolidating type use code in order to
reduce replication. This implements a few instructions to make a simple
example pass. Ex.

```
(module
    (func $addInt
        (param $value_1 i32) (param $value_2 i32)
        (result i32)
        local.get 0
        local.get 1
        i32.add
    )
    (export "AddInt" (func $addInt))
)
```


wasm/text/func_parser.go
```diff
@@ -0,0 +1,131 @@
...
+	currentCode []byte
+}
+
+var end = []byte{wasm.OpcodeEnd}
+
+func (p *funcParser) getCode() []byte {
+	if p.currentCode == nil {
+		return end
+	}
+	return append(p.currentCode, wasm.OpcodeEnd)
```
mathetake, line 39: better add comment about the reason why we always end with end optcode: https://www.w3.org/TR/wasm-core-1/#expressions%E2%91%A0

codefromthecrypt:
PS I will redo the first pass not parsing the body in S-expressions. I want to get @summerwind first demo working :) https://github.com/summerwind/the-art-of-webassembly-go/blob/main/chapter1/addint/addint.wat

codefromthecrypt:
I know this is huge, but it could be worse (ex if also implemented locals etc) and most of it is tests. Functions was going to be huge whenever it started, but yeah appreciate the patience with this mountain!

wasm/internal/wazeroir/compiler.go
```diff
@@ -133,7 +133,7 @@ func (c *compiler) resetUnreachable() {
 }
 
 type CompilationResult struct {
-	// Operations holds wazerois operations compiled from Wasm instructions in a Wasm function.
+	// Operations holds wazeror operations compiled from Wasm instructions in a Wasm function.
```
mathetake, line 136: 😄 

```suggestion
	// Operations holds wazeroir operations compiled from Wasm instructions in a Wasm function.
```

wasm/binary/import.go
```diff
@@ -69,3 +69,12 @@ func encodeImport(i *wasm.Import) []byte {
 	}
 	return data
 }
+
+// encodeCode returns the wasm.Code encoded in WebAssembly 1.0 (MVP) Binary Format.
+//
+// See https://www.w3.org/TR/wasm-core-1/#binary-code
+func encodeCode(c *wasm.Code) []byte {
+	code := append(leb128.EncodeUint32(c.NumLocals), c.LocalTypes...)
```
mathetake, line 77: I think this encoding maybe isn't correct as 

local types are encoded as `uint32 (number of locals for specific type), byte (value type),...` 

See https://github.com/tetratelabs/wazero/blob/main/wasm/binary/code.go#L31-L47

mathetake (approved):
awesome!

wasm/internal/leb128/leb128.go
```diff
@@ -17,10 +17,26 @@ var (
 	errOverflow64 = errors.New("overflows a 64-bit integer")
 )
 
+// encodeCache reduces allocations by returning a constant slice of size one for values that can encode in a single byte
+var encodeCache = [0x80][]byte{
```
codefromthecrypt, line 21: here's the new cache

wasm/binary/code.go
```diff
@@ -64,13 +64,47 @@ func decodeCode(r io.Reader) (*wasm.Code, error) {
...
+//
+// See https://www.w3.org/TR/wasm-core-1/#binary-code
+func encodeCode(c *wasm.Code) []byte {
+	// local blocks compress locals while preserving index order by grouping locals of the same type.
+	// https://www.w3.org/TR/wasm-core-1/#code-section%E2%91%A0
+	localBlockCount := uint32(0) // how many blocks of locals with the same type (types can repeat!)
+	var localBlocks []byte
+	localTypeLen := len(c.LocalTypes)
+	if localTypeLen > 0 {
+		i := localTypeLen - 1
```
codefromthecrypt, line 84: here's the correct encoding logic

codefromthecrypt (commented):
@mathetake PTAL if you don't mind

wasm/text/parser.go
```diff
@@ -486,6 +492,80 @@ func (p *moduleParser) parseImportFuncEnd(tok tokenType, tokenBytes []byte, _, _
...
+	if tok == tokenID { // Ex. $main
+		na := &wasm.NameAssoc{Index: p.currentFuncIndex, Name: string(stripDollar(tokenBytes))}
+		p.module.funcNames = append(p.module.funcNames, na)
+		p.tokenParser = p.parseFunc
+		return nil
+	}
+	return p.parseFunc(tok, tokenBytes, line, col)
+}
+
+// parseFunc is the second parser inside the ed function field. This passes control to the typeParser until
```
nullpo-head, line 514: Perhaps is `ed` a typo? > `the ed function`

codefromthecrypt: yep! it was partial deletion of "imported function"

mathetake (approved):

wasm/text/parser.go
codefromthecrypt, line 514:
```suggestion
// parseFunc is the second parser inside a module defined function field. This passes control to the typeParser until
```

codefromthecrypt:
Thanks for the review help @mathetake and @nullpo-head!

PR #152, opened 2022-01-21
Consolidate runtime errors

mathetake:
This commit centralizes which runtime errors are possibly returned by engines, 
and document all of them on under which condition they occur.

Notably, now our spectest for `assert_trap` not only checks if the runtime returns error,
but also checks the type of error. This way, we can ensure that 
interpreter and JIT engines return exactly the same error for the same condition.
In order to achieve that, I found some discrepancies in the semantics between 
JIT and interpreter and resolved them. Notably, I refactored the code for division and 
remainder calculation in JIT so that they are alined with interpreter and spectests.

### DIFF
--- a/wasm/errors.go
+++ b/wasm/errors.go
@@ -2,7 +2,7 @@ package wasm
 
 import "errors"
 
-// All the erros were returned by Engine during the excecution of Wasm functions,
+// All the erros are returned by Engine during the excecution of Wasm functions,
 // and they indicate that the Wasm virtual machine's state is unrecoverable.
 var (
 	// ErrRuntimeCallStackOverflow indicates that there are too many function calls,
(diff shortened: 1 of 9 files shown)


wasm/errors.go
```diff
@@ -2,4 +2,30 @@ package wasm
 
 import "errors"
 
-var ErrCallStackOverflow = errors.New("callstack overflow")
+// All the erros are returned by Engine during the excecution of Wasm functions,
```
codefromthecrypt, line 5: thanks for these!

wasm/spectests/spec_test.go
```diff
@@ -149,6 +150,35 @@ func (v commandActionVal) toUint64() uint64 {
 	}
 }
 
+// Returns the expected runtime error when the command type equals assert_trap
```
codefromthecrypt, line 153:
```suggestion
// expectedError returns the expected runtime error when the command type equals assert_trap
```

codefromthecrypt (approved):
Looks great. If you have any links to explain any weirdness in corner cases, please add them as comments in either the code or the unit test that hits those corner cases

mathetake:
thanks for the look!

Issue #154, opened 2022-01-21
JIT: tests with GOOS=windows


mathetake:

mathetake:
oh this is closed by #300

PR #155, opened 2022-01-21
jit: enable build on Windows 

mathetake:
part of #154. Currently we only implement the mmap for `linux` and `darwin`,
therefore JIT is only available for these GOOSes. This commit at least makes it 
possible for Windows to do `go build ./...`.


wasm/jit/mmap_darwin.go
```diff
@@ -1,5 +1,24 @@
+//go:build darwin
+// +build darwin
+
 package jit
 
 import "syscall"
 
 const mmapFlags = syscall.MAP_ANON
+
+// Copy the code into the executable region
```
codefromthecrypt, line 10:
```suggestion
// mmapCodeSegment copies the code into the executable region
```

wasm/jit/mmap_darwin.go
```diff
@@ -1,5 +1,24 @@
...
+// and returns the byte slice of the region.
+func mmapCodeSegment(code []byte) ([]byte, error) {
+	mmapFunc, err := syscall.Mmap(
+		-1,
+		0,
+		len(code),
+		syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC, syscall.MAP_PRIVATE|mmapFlags,
```
codefromthecrypt, line 17: comment why these flags using a link if necessary

wasm/jit/mmap_linux.go
```diff
@@ -1,5 +1,24 @@
+//go:build linux
+// +build linux
+
 package jit
 
 import "syscall"
 
 const mmapFlags = syscall.MAP_ANONYMOUS
+
+// Copy the code into the executable region
```
codefromthecrypt, line 10:
```suggestion
// mmapCodeSegment copies the code into the executable region
```

wasm/jit/mmap_other.go
```diff
@@ -0,0 +1,12 @@
+//go:build !darwin && !linux
+// +build !darwin,!linux
+
+package jit
+
+const mmapFlags = 0
+
+// Copy the code into the executable region
```
codefromthecrypt: change this to `mmapCodeSegment is unsupported on windows.`

codefromthecrypt (commented):
instead of copy/paste, maybe resurrect `mmap.go` and include `mmapCodeSegment` as `mmapCodeSegmentPosix` or some better name. then instead of 2 files for darwin and linux, use one with flags for both and in that do..

```
func mmapCodeSegment().. {
   return mmapCodeSegmentPosix()
}
```

that way mmap.go can have unit tests and also those tests don't need to be copy/pasted

wdyt?

wasm/jit/mmap_other.go
```diff
@@ -0,0 +1,8 @@
+//go:build !darwin && !linux
```
codefromthecrypt, line 1: ps I think it is fine to rename this to _windows and wait for someone to ask for something besides darwin and linux. we would likely want CI if that were the case anyway

codefromthecrypt (approved):

wasm/jit/mmap_darwin.go
codefromthecrypt: @mathetake ready except I don't know what to put here

mathetake: https://github.com/tetratelabs/wazero/pull/155/commits/2f3a70f403e9ea78dbb66a4ea239203d54871f7d Added comments and link!

mathetake:
Thanks for the look and polish!

codefromthecrypt:
thanks @mathetake your comments taught me some things!

PR #158, opened 2022-01-24
Introduce internal pkg, reorganize directories

mathetake:
This PR introduces `internal` directory and put all the following non-user-relevant packages there:
* wazeroir
* leb128
* ieee754 
* wasm/example 
* spectests

Also this creates `interpreter` package where we put the wazeroir interpreter taken out of wazeroir pkg
since the fact wazeroir.NewEngine returns interpreter is not intuitive to users and IR is IR so it shouldn’t
be the matter of users.

Before:

```
$ tree -d
.
├── bench
│   └── testdata
├── examples
│   └── testdata
├── wasi
└── wasm
    ├── binary
    ├── buildoptions
    ├── example
    │   └── testdata
    ├── ieee754
    ├── jit
    │   └── testdata
    ├── leb128
    ├── spectests
    │   └── cases
    ├── text
    └── wazeroir
```

After:

```
$ tree -d    
.
├── bench
│   └── testdata
├── examples
│   └── testdata
├── wasi
└── wasm
    ├── binary
    ├── buildoptions
    ├── internal
    │   ├── example
    │   │   └── testdata
    │   ├── ieee754
    │   ├── leb128
    │   ├── spectests
    │   │   └── cases
    │   └── wazeroir
    ├── interpreter
    ├── jit
    │   └── testdata
    └── text
```


codefromthecrypt (approved):
This looks great. since minor changes can come later, but end up coherent with the plan here, I'll refrain from org nits and just raise any follow-up PR as I notice things

PR #159, opened 2022-01-24
Ignore spectest cases from lang stat

mathetake:
missed in #158


codefromthecrypt:
thanks for catching this

PR #161, opened 2022-01-24
jit: implement function calls in fully assembly

### LINKED ISSUE #151: JIT: implement function calls in assembly to speed up "call" and "call_indirect"
The [current implementation](https://github.com/tetratelabs/wazero/blob/main/wasm/jit/RATIONALE.md) proxies all the function calls through the Go world. That is, we go back to the callsite of JITed code whenever makes function calls, and then we re-enter the caller's JITed code again from the Go world.

This comes with a huge cost and is the current bottleneck as you can see thorough the benchmarks for iterative factorial vs recursive factorial.

mathetake:
This commit implements (almost) all the code required to make function calls in assembly.
Notably, now that function calls are never proxied through Go world, but instead directly 
calling back and forth among functions in JIT compiled native code.

Now, the overhead of function calls is almost nothing (compared to the past calling convention),
and the benchmark results showed that this improves the performance of JIT up around 100 times
in call-heavy cases.

Due to the limitations in JIT compiler in Go, there's one exception in direct function calls 
from assembly. That is when the code tries to make host function calls. As we cannot use
direct jump into the Go function's address in order to avoid conflicts with Go runtime,
instead we achieve the call just like before by proxies through the JIT engine before 
calling the actual host functions.

The basic idea is that we create (push and pop) function call frames onto the Go-allocated 
slice and  never use `call` and `return` instructions to achieve the function calls. 
If doing so, that might result in manipulating Go-routine stacks from JITed native code and 
there might be conflicts with Goruntimes. Instead we just use the absolute address `jmp` 
to back and forth among functions.

resolves #151 

The follow-up PR would add more detailed design doc in RATIONALE.md and for now I would leave it as TODO 😄


mathetake:
ok done 80%

mathetake:
Ok I will mark this ready after polishing docs and observe the benchmark improvements

mathetake:
Here's the bench result for Recursive fibonacci function written in wat. 
This proves that the cost of function calls was the bottleneck and now we resolved the issue! 

#### before
```
BenchmarkFibonacci
BenchmarkFibonacci/jit
BenchmarkFibonacci/jit-32         	     477	   2480705 ns/op	 1051497 B/op	   21893 allocs/op
BenchmarkFibonacci/interpreter
BenchmarkFibonacci/interpreter-32 	     555	   2242362 ns/op	  350296 B/op	   21893 allocs/op
```

#### after
```
goos: linux
goarch: amd64
pkg: github.com/tetratelabs/wazero/wasm/jit
BenchmarkFibonacci
BenchmarkFibonacci/jit
BenchmarkFibonacci/jit-32         	    7714	    149843 ns/op	      61 B/op	       2 allocs/op
BenchmarkFibonacci/interpreter
BenchmarkFibonacci/interpreter-32 	     564	   2182406 ns/op	  350304 B/op	   21893 allocs/op
```

wasm/jit/engine_test.go
```diff
@@ -194,36 +248,3 @@ func TestEngine_RecursiveEntry(t *testing.T) {
 	_, _, err = store.CallFunction("test", "main", uint64(1))
 	require.NoError(t, err)
 }
-
-func TestEngine_maybeGrowStack(t *testing.T) {
```
mathetake, line 198: now this is implemented in assembly so deleted the test

wasm/jit/engine_tmp_test.go
```diff
@@ -1,75 +0,0 @@
-package jit
```
mathetake, line 1: well this was a file I used to debug but somehow checked in so I deleted here 😄

wasm/jit/testdata/tmp.wat
```diff
@@ -1,21 +0,0 @@
-
```
mathetake, line 1: this is also accidentally checked in

wasm/jit/RATIONALE.md
```diff
@@ -40,62 +40,4 @@ And `jitcall` is actually implemented in [jit_amd64.s](./jit_amd64.s) as a conve
...
-    callerFrame := currentFrame.caller
-    e.callFrameStack = callerFrame
-    e.stackBasePointer = callerFrame.stackBasePointer
-    e.stackPointer = callerFrame.continuationStackPointer
-```
-
-To summarize, every function call is achieved by returning back to Go code (`engine.exec`'s main loop) with some continuation infor, and enter the callee native code (or host functions) from there. That, of course, comes with a bit of overhead because each function call is implemented by two steps (returning back to `jitcall` callsite AND entering `jitcall` again) vs just `call` instruction (or `jmp`) in usual native codes.
-
-Note that this mechanism is a minimal PoC impl, so in the near future, we would achieve the function calls without returning back to `engine.exec`'s main loop and instead `jmp` directly to the callee native code.
+TODO:
```
mathetake, line 43: Maybe won't resolve this TODO in this PR 😄

mathetake:
sorry I must admit this PR is massive.. 😄
  img

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
-	// callFrameNum tracks the current number of call frames.
-	// Note: this is not len(callFrameStack) because the stack is implemented as a linked list.
-	callFrameNum uint64
-
-	// The following fields are only used during compilation.
-
-	// Store the compiled functions.
-	compiledFunctions map[wasm.FunctionAddress]*compiledFunction
-}
+type (
```
codefromthecrypt, line 18: nit: not sure the extra indents are worth it here

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+	engine struct {
+		// There contexts are read and written by JITed code.
+		// Note that we embed these structs so we can reduce the costs to access fields inside of them.
+		// Also, that eases the calculation of offsets to each field.
```
codefromthecrypt, line 22: to ensure this isn't conflated with the godoc on globalContext
```suggestion
		// Also, that eases the calculation of offsets to each field.

```

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
-	callFrameNum uint64
-
-	// The following fields are only used during compilation.
-
-	// Store the compiled functions.
-	compiledFunctions map[wasm.FunctionAddress]*compiledFunction
-}
+type (
+	engine struct {
+		// There contexts are read and written by JITed code.
```
codefromthecrypt, line 20:
```suggestion
		// These contexts are read and written by JITed code.
```

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+		// Note that we embed these structs so we can reduce the costs to access fields inside of them.
+		// Also, that eases the calculation of offsets to each field.
+		globalContext
+		moduleContext
+		valueStackContext
+		exitContext
+
+		// The following fields are NOT acceessed by JITed code directly.
+
+		// The actual Go-allocated stack for holding Wasm values.
```
codefromthecrypt, line 30:
```suggestion
		// valueStack is the go-allocated stack for holding Wasm values.
```

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+		// Note that we NEVER edit len or cap in JITed code so we won't get screwed when GC comes in.
```
codefromthecrypt, line 31:
```suggestion
		// Note: We NEVER edit len or cap in JITed code so we won't get screwed when GC comes in.
```

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+		valueStack []uint64
+
+		// Function call frames. The new one is set at callFrameStack[callFrameStackPoitner].
```
codefromthecrypt, line 34:
```suggestion
		// callFrameStack is initially callFrameStack[callFrameStackPointer].
```

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+		// The currently executed function call frame lives at callFrameStack[callFrameStackPoitner-1]
+		// and that is equivalent to  engine.callFrameTop().
+		callFrameStack []callFrame
+
+		// Store the compiled functions.
```
codefromthecrypt, line 39: here and everywhere

```suggestion
		// compiledFunctions are the currently compiled functions.
```

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+		// as the underlying memory region is accessed by assembly directly by
+		// using compiledFunctionsFirstItemAddress.
+		compiledFunctions []*compiledFunction
+	}
+
+	// globalContext holds the data which stays same across multiple function calls.
+	globalContext struct {
+		// &engine.valueStack[0] as uintptr.
+		// Note: this is updated when growing the stack in builtinFunctionGrowValueStack.
+		valueStackFirstItemAddress uintptr
```
codefromthecrypt, line 50: here and everywhere as element is more often used than item

```suggestion
		valueStackElement0Address uintptr
```

wasm/jit/engine_test.go
```diff
@@ -14,19 +15,73 @@ import (
...
-	require.Equal(t, int(unsafe.Offsetof((&engine{}).globalSliceAddress)), engineglobalSliceAddressOffset)
-	require.Equal(t, int(unsafe.Offsetof((&engine{}).memorySliceAddress)), engineMemorySliceAddressOffset)
-	require.Equal(t, int(unsafe.Offsetof((&engine{}).memorySliceLen)), engineMemorySliceLenOffset)
-	require.Equal(t, int(unsafe.Offsetof((&engine{}).tableSliceAddress)), engineTableSliceAddressOffset)
-	require.Equal(t, int(unsafe.Offsetof((&engine{}).tableSliceLen)), engineTableSliceLenOffset)
+// Ensures that the offset consts do not drift when we manipulate the target structs.
+func TestVeifyOffsetValue(t *testing.T) {
+	var eng engine
+	// Offsets for engine.globalContext.
+	// var gc globalContext
```
mathetake, line 22:
```suggestion
```

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+		callFrameStack []callFrame
+
+		// Store the compiled functions.
+		// The index means wasm.FunctionAddress, but we intentionally avoid using map
+		// as the underlying memory region is accessed by assembly directly by
+		// using compiledFunctionsFirstItemAddress.
+		compiledFunctions []*compiledFunction
+	}
+
+	// globalContext holds the data which stays same across multiple function calls.
```
codefromthecrypt, line 46:
```suggestion
	// globalContext holds the data which is constant across multiple function calls.
```

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+		memorySliceLen uint64
+		// The address of the first item in the globa slice, i.e. &ModuleInstance.Tables[0].Table[0] as uintptr.
+		tableFirstItemAddress uintptr
+		// The length of the memory buffer, i.e. len(ModuleInstance.Tables[0].Table).
+		tableSliceLen uint64
+	}
+
+	// valueStackContext stores the data to access engine.valueStack.
+	valueStackContext struct {
+		// Stack pointer on .value field which is accessed by [stackBasePointer] + [stackPointer].
```
codefromthecrypt, line 100: .value definitely won't resolve, so pin it.

wasm/jit/engine.go
```diff
@@ -15,70 +15,327 @@ import (
...
+		// This is only used in the Go world and set when jit exiting as native code know
```
codefromthecrypt, line 101: assume the reader is in the go world, and rewrite the other way :D

wasm/jit/engine.go
```diff
@@ -148,217 +401,52 @@ func (e *engine) Call(f *wasm.FunctionInstance, params ...uint64) (results []uin
...
-	}
-	return nil
-}
-
 func NewEngine() wasm.Engine {
 	return newEngine()
 }
 
-const initialStackSize = 1024
+// TODO: better make them configurable?
```
codefromthecrypt, line 413: should we? think about other JIT look at them and cite if they configure this! (when you get to the TODO)

wasm/jit/engine.go
```diff
@@ -148,217 +401,52 @@ func (e *engine) Call(f *wasm.FunctionInstance, params ...uint64) (results []uin
...
-		ret = "returned"
-	case jitCallStatusCodeCallFunction:
-		ret = "call_function"
-	case jitCallStatusCodeCallBuiltInFunction:
-		ret = "call_builtin_function"
-	case jitCallStatusCodeUnreachable:
-		ret = "unreachable"
-	}
+func (e *engine) popValue() (ret uint64) {
+	ret = e.valueStack[e.valueStackContext.stackBasePointer+e.valueStackContext.stackPointer-1]
```
codefromthecrypt, line 431: this repeats, extract a function (that will end up inlined) so you can document it

wasm/jit/engine.go
```diff
@@ -400,122 +488,223 @@ func (e *engine) execHostFunction(f *reflect.Value, ctx *wasm.HostFunctionCallCo
...
+	// initialize moduleContext's other fields in the function preamble inside of JITed code.
+	e.moduleContext.moduleInstanceAddress = f.moduleInstanceAddress
+
+	// Push the new frame to the top of stack.
+	e.callFrameStack[e.globalContext.callFrameStackPointer] = callFrame{returnAddress: f.codeInitialAddress, compiledFunction: f}
+	e.globalContext.callFrameStackPointer++
+
+	// For example, if we have the following state (where "_" means no value pushed),
+	//       base            sp
+	//        |              |
```
codefromthecrypt, line 580: nice

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+	// At this point, callFrameStackPointerRegister holds the offset in call frame slice in bytes,
+	// and tmpRegister holds the absolute address of the first item of call frame slice.
+	// To illustrate the situation:
+	//
+	//  tmpRegister (holding the absolute address of &callFrame[0])
+	//      |
+	//      [ra.0, rb.0, rc.0, _, ra.1, rb.1, rc.1, _, ra.next, rb.next, rc.next, ...]  <--- call frame stack's data region (somewhere in the memory)
+	//      |                                        |
+	//      <---------------------------------------->
+	//          callFrameStackPointerRegister (holding the offset from &callFrame[0] in bytes.)
```
codefromthecrypt, line 4711: the diagram is ambiguous as it shows an arrow at both ends, which makes it confusing what callFrameStackPointerRegister is actually pointing to. use another LF if needed and you want to show the range and also what the pointer is at

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+	calcCallFrameStackTopAddress.As = x86.ALEAQ
+	calcCallFrameStackTopAddress.To.Type = obj.TYPE_REG
+	calcCallFrameStackTopAddress.To.Reg = callFrameStackTopAddressRegister
+	calcCallFrameStackTopAddress.From.Type = obj.TYPE_MEM
+	calcCallFrameStackTopAddress.From.Reg = tmpRegister
+	calcCallFrameStackTopAddress.From.Index = callFrameStackPointerRegister
+	calcCallFrameStackTopAddress.From.Scale = 1
+	c.addInstruction(calcCallFrameStackTopAddress)
+
+	// Complete 1).
```
codefromthecrypt, line 4740: inline what 1 is or extract a function and don't use numbers

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+		saveCurrentStackBasePointerIntoCallFrameFromTmpRegister.To.Type = obj.TYPE_MEM
+		saveCurrentStackBasePointerIntoCallFrameFromTmpRegister.To.Reg = callFrameStackTopAddressRegister
+		// "rb.1" is BELOW the top address. See the above example for detail.
+		saveCurrentStackBasePointerIntoCallFrameFromTmpRegister.To.Offset = -(callFrameDataSize - callFrameReturnStackBasePointerOffset)
+		saveCurrentStackBasePointerIntoCallFrameFromTmpRegister.From.Type = obj.TYPE_REG
+		saveCurrentStackBasePointerIntoCallFrameFromTmpRegister.From.Reg = tmpRegister
+		c.addInstruction(saveCurrentStackBasePointerIntoCallFrameFromTmpRegister)
+	}
+
+	// Complete 2).
```
codefromthecrypt, line 4764: inline what 2 is or extract a function and don't use numbers

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+		putCompiledFunctionFunctionAddressOnNewCallFrame.As = x86.AMOVQ
+		putCompiledFunctionFunctionAddressOnNewCallFrame.To.Type = obj.TYPE_MEM
+		putCompiledFunctionFunctionAddressOnNewCallFrame.To.Reg = callFrameStackTopAddressRegister
+		putCompiledFunctionFunctionAddressOnNewCallFrame.To.Offset = callFrameCompiledFunctionOffset
+		putCompiledFunctionFunctionAddressOnNewCallFrame.From.Type = obj.TYPE_REG
+		putCompiledFunctionFunctionAddressOnNewCallFrame.From.Reg = compiledFunctionAddressRegister
+		c.addInstruction(putCompiledFunctionFunctionAddressOnNewCallFrame)
+	}
+
+	// Complete 4).
```
codefromthecrypt, line 4832: same

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+		setEngineModuleInstanceAddress.As = x86.AMOVQ
+		setEngineModuleInstanceAddress.From.Type = obj.TYPE_REG
+		setEngineModuleInstanceAddress.From.Reg = tmpRegister
+		setEngineModuleInstanceAddress.To.Type = obj.TYPE_MEM
+		setEngineModuleInstanceAddress.To.Reg = reservedRegisterForEngine
+		setEngineModuleInstanceAddress.To.Offset = engineModuleContextModuleInstanceAddressOffset
+		c.addInstruction(setEngineModuleInstanceAddress)
+	}
+
+	// Now ready to complete 5).
```
codefromthecrypt, line 4855: same

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+	// and tmpRegister holds the absolute address of the first item of call frame slice.
+	// To illustrate the situation:
+	//
+	//  tmpRegister (holding the absolute address of &callFrame[0])
+	//      |                              callFrameStackTopAddressRegister (absolute address of tmpRegister+decrementedCallFrameStackPointerRegister)
+	//      |                                           |
+	//      [......., ra.caller, rb.caller, rc.caller, _, ra.current, rb.current, rc.current, _, ...]  <--- call frame stack's data region (somewhere in the memory)
+	//      |                                           |
+	//      <------------------------------------------->
+	//           decrementedCallFrameStackPointerRegister (holding the offset from &callFrame[0] in bytes.)
```
codefromthecrypt, line 5030: same comment about diagram.. use a line if needed to point precisely

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+	//      rb.* = callFrame.returnStackBasePointer
+	//      rc.* = callFrame.compiledFunction
+	//      _  = callFrame's padding (see comment on callFrame._ field.)
+	//
+	// What we have to do in the following is that
+	//   1) Set engine.valueStackContext.stackBasePointer to the value on "rb.caller"
+	//   2) Set engine.moduleContext.moduleInstanceAddress to the one stored in "rc.caller".
+	//   3) Jump into the address of "ra.caller"
+
+	// Complete 1).
```
codefromthecrypt, line 5043: same on numbers, inline definition or extract a function

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+	c.locationStack.markRegisterUnused(regs...)
+	return nil
+}
 
-	// Release all the registers as our calling convention requires the callee-save.
+// callGoFunction adds instructions to call a Go function whose address equals the addr parameter.
+// jitStatus is set before making call, and it should be either jitCallStatusCodeCallBuiltInFunction or
+// jitCallStatusCodeCallHostFunction.
+func (c *amd64Compiler) callGoFunction(jitStatus jitCallStatusCode, addr wasm.FunctionAddress) error {
+	// Set the functionAddress to the engine.exitContext.functionCallAddress.
```
codefromthecrypt, line 5127: double nested usually doesn't resolve

```suggestion
	// Set the functionAddress to the engine.exitContext functionCallAddress.
```

wasm/jit/jit_amd64.go
```diff
@@ -4594,66 +4569,695 @@ func (c *amd64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
...
+	// Since the assembler cannot directly emit "LEA foo [rip + bar]", we use the some hack here:
+	// We intentionally use x86.REG_BP here so that the resulting instruction sequence becomes
+	// exactly the same as "LEA foo [rip + bar]" except the most significant bit of the third byte.
+	// We do the rewrite in onGenerateCallbacks which is invoked after the assembler emitted the code.
+	readInstructionAddressAfterReturn.From.Reg = x86.REG_BP
+	c.addInstruction(readInstructionAddressAfterReturn)
+
+	return func(instructionAfterReadInstructionAddress, addressAcquisitionTargetInstruction *obj.Prog) {
+		c.onGenerateCallbacks = append(c.onGenerateCallbacks, func(code []byte) error {
+			// See the comment at readInstructionAddressAfterReturn.From.Offset.
```
codefromthecrypt, line 5241: extract this to a function so that you can use godoc links (note only direct symbols link, not nested)

codefromthecrypt (approved):
epic! unleashing comments!

wasm/jit/jit_amd64.go
mathetake: I don't think we could extract this as a function because  this is the closure on readInstructionAddressAfterReturn..

mathetake: at least simplified! https://github.com/tetratelabs/wazero/pull/161/commits/6bff87ebff3d103eb6d68cb384a4d868703f123d

mathetake:
Thanks for the reviews! @codefromthecrypt

PR #162, opened 2022-01-24
wat: begins type consolidation with wasm.X

codefromthecrypt:
This reduces the amount of conversion work by using wasm.X types when
parsing the text format. A follow-up change is needed to complete this
as remaining sections imply index resolution.

Before:
```
BenchmarkCodecExample/text.DecodeModule
BenchmarkCodecExample/text.DecodeModule-16           	   79468	     15224 ns/op	    7456 B/op	     331 allocs/op
```

After:
```
BenchmarkCodecExample/text.DecodeModule
BenchmarkCodecExample/text.DecodeModule-16           	   76357	     14908 ns/op	    7328 B/op	     323 allocs/op
```


codefromthecrypt:
actually, I can add name section parsing into this PR

wasm/text/bind.go
```diff
@@ -140,3 +130,47 @@ func checkIndexInRange(idx *index, count uint32, context string, contextArg0 int
 	}
 	return nil
 }
+
+// mergeLocalNames produces wasm.NameSection LocalNames. This has to be done post-parse as types can be defined after
+// functions that use them.
+func mergeLocalNames(m *module, typeParamNames map[wasm.Index]wasm.NameMap) wasm.IndirectNameMap {
```
codefromthecrypt, line 136: this logic was moved from another file without altering it

mathetake (approved):

PR #163, opened 2022-01-25
Enforces uniqueness within symbolic indices

codefromthecrypt:
> An identifier context is well-formed if no index space contains duplicate identifiers.

We weren't checking this, and now we are!. This comes at a cost, though,
but I don't see a good alternative to initializing maps wrt uniqueness.

```
BenchmarkCodecExample/text.DecodeModule
BenchmarkCodecExample/text.DecodeModule-16           	   78700	     14460 ns/op	    7328 B/op	     323 allocs/op
```

After
```
BenchmarkCodecExample/text.DecodeModule
BenchmarkCodecExample/text.DecodeModule-16           	   74214	     16921 ns/op	    8288 B/op	     333 allocs/op
```

See https://www.w3.org/TR/wasm-core-1/#text-context

### DIFF
--- a/wasm/module.go
+++ b/wasm/module.go
@@ -1,7 +1,5 @@
 package wasm
 
-import "sort"
-
 // DecodeModule parses the configured source into a wasm.Module. This function returns when the source is exhausted or
 // an error occurs. The result can be initialized for use via Store.Instantiate.
 //
@@ -333,18 +331,6 @@ type NameMap []*NameAssoc
 type NameAssoc struct {
 	Index Index
 	Name  string
-}
-
-// NewNameMap constructs a new NameMap from the given name to index map and returns it ordered by index ascending.
-func NewNameMap(m map[string]Index) NameMap {
-	result := make(NameMap, 0, len(m))
-	for n, idx := range m {
-		result = append(result, &NameAssoc{Name: n, Index: idx})
-	}
-	sort.Slice(result, func(i, j int) bool {
-		return result[i].Index < result[j].Index
-	})
-	return result
 }
 
 // IndirectNameMap associates an index with an association of names.
--- a/wasm/module.go
+++ b/wasm/module.go
@@ -1,5 +1,7 @@
 package wasm
 
+import "sort"
+
 // DecodeModule parses the configured source into a wasm.Module. This function returns when the source is exhausted or
 // an error occurs. The result can be initialized for use via Store.Instantiate.
 //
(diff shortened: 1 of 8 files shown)


wasm/module.go
```diff
@@ -333,6 +335,20 @@ type NameAssoc struct {
 	Name  string
 }
 
+// NewNameMap constructs a new NameMap from the given name to index map and returns it ordered by index ascending.
+func NewNameMap(m map[string]Index) NameMap {
```
codefromthecrypt, line 339: I will delete this and build it iteratively as I looked at the call sites and we can

wasm/module.go
```diff
@@ -333,6 +335,20 @@ type NameAssoc struct {
...
+	result := make(NameMap, len(m))
+	i := 0
+	for n, idx := range m {
+		result[i] = &NameAssoc{Name: n, Index: idx}
+		i++
+	}
```
mathetake, line 345: less variable 
```suggestion
	result := make(NameMap, 0, len(m))
	for n, idx := range m {
		result = append(result, &NameAssoc{Name: n, Index: idx})
	}
```

codefromthecrypt: nice!

codefromthecrypt:
ok I reduced the change which also shaved off a couple allocations and backfilled a missing test

wasm/text/parser.go
```diff
@@ -590,15 +611,16 @@ func (p *moduleParser) parseFuncBodyField(tok tokenType, tokenBytes []byte, line
 // and/or Func.typeInlined and sets the next parser to parse.
 func (p *moduleParser) parseFuncEnd(tok tokenType, tokenBytes []byte, _, _ uint32) error {
 	if tok == tokenRParen {
-		tu, paramNames := p.typeParser.getTypeUse()
+		tu, _, paramNames := p.typeParser.getTypeUse()
```
codefromthecrypt, line 614: this `_` will be used when we implement locals to range check.. if that doesn't end up true, I'll change the sig and delete it :D

wasm/text/types.go
```diff
@@ -87,6 +87,17 @@ type index struct {
 	col uint32
 }
 
+type idContext map[string]wasm.Index
+
+func (ctx idContext) setID(idToken []byte, idx uint32) (string, error) {
```
mathetake, line 92: comments?

mathetake (approved):

nullpo-head (approved):
looks good to me!

wasm/text/types.go
codefromthecrypt: nice catch, let me add it!

codefromthecrypt:
thanks for the help!

PR #166, opened 2022-01-27
Implement args_get and args_sizes_get

nullpo-head:
Implement simple system calls, args_get and args_sizes_get first.  This PR also introduces APIs to safely access any guest address given by the user without the runtime crashing.


examples/testdata/wasi_args.go
```diff
@@ -0,0 +1,53 @@
+package main
```
nullpo-head, line 1: I guess I will make `wasi` subfolder later since a LOT of WASI related test files will be introduced in `examples`.

mathetake:
looks good overall.. I will review within today!

wasi/wasi.go
```diff
@@ -277,16 +323,28 @@ func (w *WASIEnvironment) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32)
 	return ESUCCESS
 }
 
-func args_sizes_get(ctx *wasm.HostFunctionCallContext, argcPtr uint32, argvPtr uint32) (err Errno) {
-	// not implemented yet
-	binary.LittleEndian.PutUint32(ctx.Memory.Buffer[argcPtr:], 0)
-	binary.LittleEndian.PutUint32(ctx.Memory.Buffer[argvPtr:], 0)
-	return 0
+func (w *WASIEnvironment) args_sizes_get(ctx *wasm.HostFunctionCallContext, argsCountPtr uint32, argsBufSizePtr uint32) Errno {
```
codefromthecrypt, line 326: we are missing wasi_test.go which can be started with unit tests for these two

examples/testdata/wasi_args.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+//go:wasm-module wasi_snapshot_preview1
+//export args_get
```
codefromthecrypt, line 10: I'm a little confused because I would have expected a test to import args_get, not export it. can you comment why?

examples/wasi_args_test.go
```diff
@@ -0,0 +1,38 @@
...
+	require.NoError(t, err)
+	wasiEnv := wasi.NewEnvironment(args, wasi.Stdout(os.Stderr))
+
+	err = wasiEnv.Register(store)
+	require.NoError(t, err)
+
+	err = store.Instantiate(mod, "test")
+	require.NoError(t, err)
+
+	// Let TinyGo runtime initialize the WASI environment by calling main.
```
codefromthecrypt, line 35: this feels like a test that should be in tinygo.. what makes me curious is why we have this and not for example a %.wat file importing the expected functions, then using testify here to validate they match wasi.args

codefromthecrypt (commented):
Thanks for starting this, I'm sure I'm missing a critical part of how WASI interacts, so appreciate any follow-up to the PR comments.

examples/wasi_args_test.go
codefromthecrypt: ex. to be concrete, you can export and imported function.. that seems like the easiest path to check wasi arg functions are returning the correct stuff. It feels like the tinygo runtime sanity tests might be better put into tinygo where they can maintain it.

examples/testdata/wasi_args.go
```diff
@@ -0,0 +1,53 @@
...
+func args_get(uint32, uint32) uint16
+
+//go:wasm-module wasi_snapshot_preview1
+//export args_sizes_get
+func args_sizes_get(uint32, uint32) uint16
+func main() {
```
codefromthecrypt, line 16: I think what most confused me are things that look like unit tests here. since this is an example, how about just printing out the arg count and values. and verifying that in the unit test?

examples/testdata/wasi_args.go
```diff
@@ -0,0 +1,53 @@
...
+func testGetWASIArgs() {
+	var expectedArgs = []string{
+		"foo",
+		"bar",
+		"foobar",
+		"",
+		"baz",
+	}
+
+	if len(os.Args) != len(expectedArgs) {
```
codefromthecrypt, line 30: basically verify this by checking console in the unit tests, which reduces code also as you can use testify.

examples/testdata/wasi_args.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	for i, expectedArg := range expectedArgs {
+		if os.Args[i] != expectedArg {
+			panic(fmt.Sprintf("os.Args are not what are expected. expected: %v, actual: %v", expectedArgs, os.Args))
+		}
+	}
+}
+
+func testInvalidArguments() {
```
codefromthecrypt, line 41: this sort of test I think we can handle using the text format and not in an example. The less we do here the less scary the example is. If we aren't confident about something the example relies on being tested in tinygo, we can raise a PR there to make it better.

codefromthecrypt (commented):
I think I understand what's making me concerned. examples should be very little code and the tests against them only ensure a user is likely to be able to run them.

We should avoid putting things that look like unit tests into wasm as that makes the example look scary, or is smell we are missing tests somewhere (in our repo or tinygo)

hope this feedback helps steer!

mathetake (changes requested):
In the example (or maybe should be moved to wasi/**_test.go), we should use`os.Args` rather than directly invoking them because at this point test doesn't ensure that "actual" use case works.

mathetake:
oh sorry I missed that you used there!

mathetake:
agreed we shouldn't put things like unittests under examples/**, let's move them into wasi/**_test.go as adrian suggested!

nullpo-head:
Thanks for the feedbacks! I will move the tests to under `wasi` directory (not under examples but under `/wasi`). 

This will be another topic and would be another PR, but I think it might be better to make the purpose of `example` more explicit. From its name, it reads like it's an example directory for users of wazero, but its README clearly states that it is a directory for e2e tests and it also has some files which are rather tests than examples for users, which makes the purpose of `example` ambiguous. I feel it's better to move some files like `file_system.go` to WASI's directory as well and elaborate the README. Will do it later.

examples/testdata/wasi_args.go
nullpo-head: `export` is TinyGo's syntax to specify its the link-time name, not only for exporting a function. This makes TinyGo import external functions from other wasm modules. https://github.com/tinygo-org/tinygo/issues/1120 https://github.com/tinygo-org/tinygo/blob/06df31944c6af706d6cb8b30a2e92d19060436d2/src/runtime/runtime_wasm_wasi.go#L115-L117

examples/testdata/wasi_args.go
nullpo-head: Yeah checking the output from the guest in the host is one way to check the expected conditions. However, it works well for simple system calls like this, but I guess it's easier to read if the comparison gets complicated such as comparing the expected struct with the returned value. Indeed we cannot use testify in the guest, though. What do you think?

examples/testdata/wasi_args.go
nullpo-head: Moved from the example directory.

examples/testdata/wasi_args.go
nullpo-head: sorry, could you elaborate what made you feel this test validates tinyGo, not the implementation / spec of the WASI system call? It's testing that WASI's argument AP's behavior

examples/testdata/wasi_args.go
codefromthecrypt: if we are interested in grey box testing, it would be not in an example, maybe we can put something in the memory section and verify it later, if the aim is also to use the start component. generally I think it is conflating as it isn't required to use the start component to reach the values (we can export the function and call it directly)

examples/testdata/wasi_args.go
codefromthecrypt: sorry it was the above function (misplaced comment) which was looking at `os.Args`. here really it is most direct to just export the function and test it vs trying to invoke test code in wasm

wasm/store.go
```diff
@@ -757,6 +758,23 @@ func (s *Store) buildExportInstances(module *Module, target *ModuleInstance) (ro
 	return
 }
 
+// ValidateAddrRange checks if the given address range is a valid address range.
+// It accepts rangeSize as uint64 so that callers can add or multiply two uint32 addresses
+// without overflow. eg.) ValidateAddrRange(uint32Offset, uint64(uint32Size) + 1)
+func (m *MemoryInstance) ValidateAddrRange(addr uint32, rangeSize uint64) bool {
+	return uint64(addr) < uint64(len(m.Buffer)) && rangeSize <= uint64(len(m.Buffer))-uint64(addr)
```
jcchavezs, line 765: Can `rangeSize` be zero here (`from the example it looks like not`)? If not, we just need `uint64(len(m.Buffer))-uint64(addr) >= rangeSize` condition.

If rangeSize can be 0 you can do 

```
if diffSize := uint64(len(m.Buffer))-uint64(addr); diffSize > 0 && diffSize >= rangeSize {
   return true
}

return false
```

nullpo-head: This will result in underflow when `addr` is bigger than `len(m.Buffer)`.
`uint64(len(m.Buffer))-uint64(addr)`

jcchavezs: Yeah, if that happens the difference `uint64(len(m.Buffer))-uint64(addr)` is negative but that will never be `>=` than rangeSize which is a `uint64`.

nullpo-head: sorry, maybe you've lost me. 

>  uint64(len(m.Buffer))-uint64(addr) is negative 

IIUC, this never be negative because it's unsigned. For example, `uint64(1) - `uint64(2)` is `18446744073709551615` and will be  bigger than almost any `rangeSize`

wasi/wasi.go
```diff
@@ -107,6 +110,48 @@ func Stderr(writer io.Writer) Option {
...
+	if args == nil {
+		return &wasiStringArray{strings: [][]byte{}}, nil
+	}
+	if len(args) > math.MaxUint32 {
+		return nil, fmt.Errorf("the length of the args exceeds the max of uint32: %v", len(args))
+	}
+	strings := make([][]byte, len(args))
+	totalBufSize := uint32(0)
+	for i, arg := range args {
+		argLen := uint64(len(arg)) + 1 // + 1 for '\x00'
```
codefromthecrypt, line 132: maybe add some unicode to the tests so that this breaks?

wasi/wasi.go
codefromthecrypt: looking at `wasm.HostFunctionCallContext`, it seems simple enough to instantiate so that we can unit test this and below. jit.Test_Simple has some of this kind of thing, and hopefully we'll have more of it soon enough!

mathetake: +1

nullpo-head: Updated the codes based on your suggestions. PTAL. Thanks!

wasi/wasi.go
nullpo-head: Added. Thanks.

wasi/wasi_test.go
```diff
@@ -0,0 +1,199 @@
...
+				[]byte("bar\x00"),
+				[]byte("foobar\x00"),
+				[]byte("\x00"),
+				[]byte("baz\x00"),
+			},
+			expectedBufSize: 20,
+		},
+		{
+			name: "utf-8 string",
+			// "😨", "🤣", and "️🏃‍♀️" have 4, 4, and 13 bytes respectively
```
codefromthecrypt, line 44: love it!

wasi/wasi_test.go
```diff
@@ -0,0 +1,199 @@
...
+	require.NoError(t, err)
+	wasiEnv := NewEnvironment(argsOpt)
+	hostFunctionCallContext := buildMockHostFunctionCallContext()
+
+	outOfBounds := uint32(len(hostFunctionCallContext.Memory.Buffer))
+
+	argsArray, err := newWASIStringArray(dummyArgs)
+	require.NoError(t, err)
+
+	errno := wasiEnv.args_sizes_get(hostFunctionCallContext, outOfBounds, 0)
```
codefromthecrypt, line 121: we tend to use matrixes for tests that aren't inherently stateful as that allows you to name them. also it breaks it up visually a bit better than here. mind doing that and naming why the inputs are interesting? you can still re-use the setup above the matrix if helpful.

If these are stateful (between assertions), then you can comment why or do inline tests to name them after a previous. ex. `t.Run("second call fails", ...`

wasm/store_test.go
```diff
@@ -52,3 +54,45 @@ func TestBuildFunctionInstances_FunctionNames(t *testing.T) {
...
+	mi := s.getModuleInstance(name)
+
+	m := &Module{
+		MemorySection: []*LimitsType{{Min: 100}},
+	}
+
+	_, err := s.buildMemoryInstances(m, mi)
+	require.NoError(t, err)
+
+	require.True(t, mi.Memory.ValidateAddrRange(uint32(0), uint64(0)))
```
codefromthecrypt, line 70: same comment here about matrix to document why the inputs or results are interesting

wasm/store_test.go
```diff
@@ -52,3 +54,45 @@ func TestBuildFunctionInstances_FunctionNames(t *testing.T) {
...
+	m := &Module{
+		MemorySection: []*LimitsType{{Min: 100}},
+	}
+
+	_, err := s.buildMemoryInstances(m, mi)
+	require.NoError(t, err)
+
+	maxUint32 := uint32(math.MaxUint32)
+	asymmetryBitsVal := uint32(0xfffffffe)
+	require.True(t, mi.Memory.PutUint32(uint32(0), asymmetryBitsVal))
```
codefromthecrypt, line 91: same comment here about matrix to document why the inputs or results are interesting

wasi/wasi_test.go
```diff
@@ -0,0 +1,199 @@
...
+	require.Equal(t, EINVAL, errno)
+}
+
+func buildMockHostFunctionCallContext() *wasm.HostFunctionCallContext {
+	return &wasm.HostFunctionCallContext{
+		Memory: &wasm.MemoryInstance{Buffer: make([]byte, wasm.PageSize), Min: 1},
+	}
+}
+
+// Test whether the specs implemented by the args API match those understood by a language runtime.
```
codefromthecrypt, line 150: This test isn't a unit test, it is closest to `interop` for lack of them. That this keeps coming up is important, but it is also important to not add more tinygo deps to the main tree just for lack of a better directory. The best time to solve this is now,

Can you please make a `interop/wasi` root directory and move this here. Inside make a test for args_get and args_sizes_get with the pre-compiled wasm in `interop/wasi/testdata`. Make can refresh that from tinygo files also in that directory.

Ex you can start a file `interop/wasi/wasi_interop_tests.go`:
```
// package wasi_interop tests wazero's WASI XX_spec_draft_XXX is compatible with at least one runtime.
//
// We use tinygo, to perform these tests as the maintainers are familiar with it.
//
// Note: Use `make XXX` to compile `testdata/*.wasm` from `testdata/*.go` 
// Note: This also substitutes for WASI spec tests until we have another option
package wasi_interop

// argsGetBinary is compiled from testdata/args_get.go
//go:embed testdata/args_get.wasm
var argsGetBinary []byte
func Test_ArgsGet(t *testing.T) {
```

wasi/wasi.go
codefromthecrypt, line 326: This file isn't maintained, yet, but needs to have godoc on each new method (until someone backfills the others) that summarizes what's going on and cites a link to the spec we are following (permalink to a specific pre-release of wasi vs a master link if available)

wasi/wasi.go
```diff
@@ -277,16 +323,28 @@ func (w *WASIEnvironment) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32)
...
+		return EINVAL
+	}
+
+	return ESUCCESS
 }
 
-func args_get(*wasm.HostFunctionCallContext, uint32, uint32) (err Errno) {
-	// not implemented yet
-	return
+func (w *WASIEnvironment) args_get(ctx *wasm.HostFunctionCallContext, argsPtr uint32, argsBufPtr uint32) (err Errno) {
```
codefromthecrypt, line 337: same comment about.. godoc (`// args_get does blah... // See https://link`) particularly where there are pointers involved, it is helpful to have a bullet per arg saying what they are supposed to represent and what we are doing with them as a result.

Ex. as there's no link I will guess :D

seems like argsPtr is where to write memory locations of each arg (as uint64). I'm not sure how this works without writing the count first (ex how does the other side know when to stop reading?), but maybe there's something in docs we can say as to how that works. 

seems like argsBufPtr is where we write the null terminated strings. 

Seems that the locations aren't assumed to be sequential in the buffer, just we do that for convenience.

wasi/wasi.go
```diff
@@ -277,16 +323,28 @@ func (w *WASIEnvironment) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32)
...
+	if !ctx.Memory.ValidateAddrRange(argsPtr, uint64(len(w.args.strings))*4) || !ctx.Memory.ValidateAddrRange(argsBufPtr, uint64(w.args.totalBufSize)) {
+		return EINVAL
+	}
+	for _, arg := range w.args.strings {
+		binary.LittleEndian.PutUint32(ctx.Memory.Buffer[argsPtr:], argsBufPtr)
+		argsPtr += 4
```
codefromthecrypt, line 343: trailing comment any magic numbers and they lose their magic!
```suggestion
		argsPtr += 4 // size of uint32
```

wasi/wasi.go
```diff
@@ -107,6 +110,48 @@ func Stderr(writer io.Writer) Option {
...
 }
 
+// wasiStringArray holds null-terminated strings. It ensures that
+// its length and total buffer size don't exceed the max of uint32.
+type wasiStringArray struct {
+	strings      [][]byte
+	totalBufSize uint32
+}
+
+// newWASIStringArray creates a cStringArray from the given string slice. It returns an error
```
codefromthecrypt, line 120: when using jargon like cstringarray add a `// See https://link` to describe what that helps with. The general assumption here will be that comparing to go things is something you can assume, so anything else, add a link!

wasi/wasi.go
```diff
@@ -107,6 +110,48 @@ func Stderr(writer io.Writer) Option {
 	}
 }
 
+// wasiStringArray holds null-terminated strings. It ensures that
+// its length and total buffer size don't exceed the max of uint32.
+type wasiStringArray struct {
+	strings      [][]byte
```
codefromthecrypt, line 116: rename to wasiArgs until it is reused somewhere besides args.

```suggestion
	// nullTerminatedValues are the UTF-8 encoded arg values with a NULL suffix for convenience writing to `argsBufPtr`
	nullTerminatedValues      [][]byte
```

wasi/testdata/args.go
```diff
@@ -0,0 +1,10 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+func main() {
```
codefromthecrypt, line 8: I know you insist on using `main` to test this, please comment why, and if we expect args to never be able to be called except from main

codefromthecrypt (changes requested):
While this is good, I insist the tinygo compat test is moved outside the normal unit test tree. I rarely hit "request changes" but once this is done, consider it approved!

It is also good to start better practice on wasi documentation (similar to what we've been doing in all other code). I added pointers to that end.

Thanks for the continued work here, and I bet you understand now why we suggested a simple change as a first one as there are project concerns lingering in the structure and rigor in WASI. Since all the comments needed to solve that are here, we really should solve them now since the code is small otherwise.

wasi/testdata/args.go
codefromthecrypt: Ex. if this were me, and I had read the spec and noticed something like XXX must be able to be called from _start. I would make that the comment and link it.

If there was also a spec saying XXX must never be able to be called except from _start, I would note that and also link it.

Anything that's personal interpretation, ex. something says should not.. are our choices and should also be documented. Ex if we choose not to test, expect from `main`/`_start`, and the spec doesn't indicate that (ex it says MAY SHOULD NOT or anything besides MUST or MUST NOT), we can say either we choose to not allow something (and test it isn't allowed) or comment that we aren't testing from another place as it is undefined.

wasi/wasi.go
```diff
@@ -107,6 +110,48 @@ func Stderr(writer io.Writer) Option {
...
+		return nil, fmt.Errorf("the length of the args exceeds the max of uint32: %v", len(args))
+	}
+	strings := make([][]byte, len(args))
+	totalBufSize := uint32(0)
+	for i, arg := range args {
+		argLen := uint64(len(arg)) + 1 // + 1 for '\x00'
+		if argLen > uint64(math.MaxUint32-totalBufSize) {
+			return nil, fmt.Errorf("the required buffer size for the args exceeds the max of uint32: %v", uint64(totalBufSize)+argLen)
+		}
+		totalBufSize += uint32(argLen)
```
mathetake, line 136: what if len(args) < math.MaxUint32 but totalBufSize exceeds math.MaxUint32? does WASI spec say anything about this?

mathetake: oh yeah that result in OOM in wasm in anyway :D

wasi/wasi.go
```diff
@@ -107,6 +110,48 @@ func Stderr(writer io.Writer) Option {
...
+// wasiStringArray holds null-terminated strings. It ensures that
+// its length and total buffer size don't exceed the max of uint32.
+type wasiStringArray struct {
+	strings      [][]byte
+	totalBufSize uint32
+}
+
+// newWASIStringArray creates a cStringArray from the given string slice. It returns an error
+// if the length or the total buffer size of the result WASIStringArray exceeds the max of uint32
+func newWASIStringArray(args []string) (*wasiStringArray, error) {
```
mathetake, line 122: maybe we want the small test on this

wasi/wasi.go
codefromthecrypt: easiest way is to adapt in the text from here and link it as it also explains why the count isn't returned. notably, we should avoid using words like "array" though as we can narrow more specifically or less confusingly (ex the var name is but, so in s/array/buffer/ at least)

```
  ;;; Read command-line argument data.
  ;;; The size of the array should match that returned by `args_sizes_get`
```

https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/witx/wasi_snapshot_preview1.witx#L15-L29

wasi/wasi_test.go
nullpo-head: Improved. Thanks!

wasm/store_test.go
nullpo-head: Improved. Thanks!

wasm/store_test.go
nullpo-head: Improved. Thanks!

wasi/wasi_test.go
nullpo-head: Moved them. Thanks for the feedback!

wasi/wasi.go
nullpo-head: Added docs. PTAL

wasi/wasi.go
nullpo-head: Added docs. PTAL

wasi/wasi.go
nullpo-head: I've introduced a file to define the sizes of types like `uint32`. This is in `. /wasi/type_size.go` for now, but we may want to move it to a more general directory later, since we had tended to use magic values like `4` to indicate the size of a type everywhere, and we may want to refactor those as well. JFYI @mathetake

wasi/wasi.go
nullpo-head: Nice catch! Actually, I just missed to fix `cStringArray` to `wasiStringArray`. `cStringArray` was an old name of this struct. Fixed it. Thanks!

wasi/wasi.go
nullpo-head: Changed the field's name as suggested. I'd like to keep `wasiStringArray` because I already know `environ_get` uses it. (https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-environ_getenviron-pointerpointeru8-environ_buf-pointeru8---result-errno) wdyt?

wasi/type_size.go
```diff
@@ -0,0 +1,6 @@
+package wasi
+
+// Constants of the size of frequent types in bytes
+const (
+	SIZE_UINT32 = 4
```
mathetake, line 5: we don't need to define the obvious consts like this

wasi/wasi.go
mathetake: I believe what @codefromthecrypt means is not to add a file and define consts for this, but rather just add the comment. It doesn't matter whether the existing code base has comments or not. Having consts and user everywhere doesn't improve anything because the callsites of these consts still needs a comment.

interop/wasi/wasi_interop_test.go
```diff
@@ -0,0 +1,93 @@
+// package wasi_interop tests wazero's WASI wasi_snapshot_preview1 is compatible with at least one runtime.
```
mathetake, line 1: runtime?

wasi/wasi_test.go
mathetake: Maybe I'd prefer the name `tests` directory rather than `interop` where I plan to move `wasm/internal/spectests` into `tests/spectests` and will add bunch of test modules under `tests/wasm` which I will use to find compiler bugs

wasi/wasi.go
```diff
@@ -107,6 +110,55 @@ func Stderr(writer io.Writer) Option {
...
+	totalBufSize := uint32(0)
+	for i, arg := range args {
+		argLen := uint64(len(arg)) + 1 // + 1 for "\x00"
+		if argLen > uint64(math.MaxUint32-totalBufSize) {
+			return nil, fmt.Errorf("the required buffer size for the args exceeds the max of uint32: %v", uint64(totalBufSize)+argLen)
+		}
+		totalBufSize += uint32(argLen)
+		strings[i] = make([]byte, argLen)
+		copy(strings[i], arg)
+		strings[i][argLen-1] = byte(0)
```
mathetake, line 146: `byte` defaults to zero

```suggestion
```

wasi/wasi.go
```diff
@@ -107,6 +110,55 @@ func Stderr(writer io.Writer) Option {
...
+	strings := make([][]byte, len(args))
+	totalBufSize := uint32(0)
+	for i, arg := range args {
+		argLen := uint64(len(arg)) + 1 // + 1 for "\x00"
+		if argLen > uint64(math.MaxUint32-totalBufSize) {
+			return nil, fmt.Errorf("the required buffer size for the args exceeds the max of uint32: %v", uint64(totalBufSize)+argLen)
+		}
+		totalBufSize += uint32(argLen)
+		strings[i] = make([]byte, argLen)
+		copy(strings[i], arg)
```
mathetake, line 145:
```suggestion
		strings[i] = []byte(arg)
```

wasi/wasi.go
```diff
@@ -107,6 +110,55 @@ func Stderr(writer io.Writer) Option {
...
+		strings[i][argLen-1] = byte(0)
+	}
+
+	return &wasiStringArray{nullTerminatedValues: strings, totalBufSize: totalBufSize}, nil
+}
+
+func Args(args []string) (Option, error) {
```
mathetake, line 152: Needs godoc as this is the public API

wasi/wasi_test.go
```diff
@@ -0,0 +1,261 @@
...
+import (
+	"encoding/binary"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/wazero/wasm"
+)
+
+func TestWasiStringArray(t *testing.T) {
```
mathetake, line 12:
```suggestion
func TestNewWasiStringArray(t *testing.T) {
```

wasm/store_test.go
```diff
@@ -52,3 +54,45 @@ func TestBuildFunctionInstances_FunctionNames(t *testing.T) {
...
+	require.NoError(t, err)
+
+	require.True(t, mi.Memory.ValidateAddrRange(uint32(0), uint64(0)))
+	require.True(t, mi.Memory.ValidateAddrRange(uint32(0), uint64(100*PageSize)))
+	require.False(t, mi.Memory.ValidateAddrRange(uint32(0), uint64(100*PageSize+1)))
+	require.False(t, mi.Memory.ValidateAddrRange(uint32(1), uint64(100*PageSize)))
+	require.False(t, mi.Memory.ValidateAddrRange(uint32(100*PageSize), uint64(0)))
+}
+
+func TestPutUint32(t *testing.T) {
```
mathetake, line 77: looks like this could be written in table driven

wasm/store_test.go
```diff
@@ -52,3 +54,45 @@ func TestBuildFunctionInstances_FunctionNames(t *testing.T) {
 	// We expect unknown for any functions missing data in the NameSection
 	require.Equal(t, []string{"unknown", "two", "unknown", "four", "five"}, names)
 }
+
+func TestValidateAddrRange(t *testing.T) {
```
mathetake, line 58: same

wasi/wasi_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+func TestArgsSizesGetReturnError(t *testing.T) {
+	dummyArgs := []string{"foo", "bar", "baz"}
+	argsOpt, err := Args(dummyArgs)
+	require.NoError(t, err)
+	wasiEnv := NewEnvironment(argsOpt)
+	hostFunctionCallContext := buildMockHostFunctionCallContext()
+
+	memorySize := uint32(len(hostFunctionCallContext.Memory.Buffer))
+	inMemory := uint32(0)
```
mathetake, line 169: document?

wasi/wasi_test.go
```diff
@@ -0,0 +1,261 @@
...
+		tc := tt
+
+		t.Run(tc.name, func(t *testing.T) {
+			errno := wasiEnv.args_get(hostFunctionCallContext, tc.argsPtr, tc.argsBufPtr)
+			require.Equal(t, EINVAL, errno)
+		})
+	}
+}
+
+func buildMockHostFunctionCallContext() *wasm.HostFunctionCallContext {
```
mathetake, line 257: nit: In what sense this is mock?

interop/wasi/testdata/args.go
```diff
@@ -0,0 +1,14 @@
...
+	"os"
+)
+
+func main() {
+}
+
+//export PrintArgs
+func PrintArgs() {
+	fmt.Printf("os.Args: %v", os.Args)
+}
```
mathetake, line 14: naive question (I haven't followed the previous discussion): what's value we could have over directly calling os.Args in main, considering that we end up having longer test codes?

interop/wasi/wasi_interop_test.go
```diff
@@ -0,0 +1,96 @@
...
+// Links:
+//    - Spec: https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md
+//    - Witx: https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/witx/wasi_snapshot_preview1.witx
+//
+// We use tinygo, to perform these tests as the maintainers are familiar with it.
+//
+// Note: Use `make build.interop` to compile `testdata/*.wasm` from `testdata/*.go`
+// Note: This also substitutes for WASI spec tests until we have another option
+package wasi_interop
```
mathetake, line 10: In general, we avoid using multi word (or hyphened words) in package name (see https://go.dev/blog/package-names)
```suggestion
package wasi
```

wasm/store_test.go
```diff
@@ -52,3 +54,45 @@ func TestBuildFunctionInstances_FunctionNames(t *testing.T) {
...
+	require.False(t, mi.Memory.ValidateAddrRange(uint32(100*PageSize), uint64(0)))
+}
+
+func TestPutUint32(t *testing.T) {
+	name := "test"
+	s := NewStore(nil)
+	mi := s.getModuleInstance(name)
+
+	m := &Module{
+		MemorySection: []*LimitsType{{Min: 100}},
```
mathetake, line 83: where's this limit used?

wasi/wasi_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/wazero/wasm"
+)
+
+func TestWasiStringArray(t *testing.T) {
+	tests := []struct {
+		name            string
+		args            []string
```
mathetake, line 15: We should add a case where args == nil as I see the block `if arga == nil` is not covered

wasi/wasi_test.go
```diff
@@ -0,0 +1,261 @@
...
+			expectedBufSize: 12,
+		},
+	}
+
+	for _, tt := range tests {
+		tc := tt
+
+		t.Run(tc.name, func(t *testing.T) {
+			opts := []Option{}
+			if tc.args != nil {
```
mathetake, line 118: args == nil is correctly handled in `newWASIStringArray` so do we need this branch?

wasi/wasi.go
```diff
@@ -18,6 +20,7 @@ const (
 )
 
 type WASIEnvironment struct {
+	args  wasiStringArray
```
mathetake, line 23: Use pointer then we don't need to copy `wasiStringArray` in option 
```suggestion
	args *wasiStringArray
```

wasm/store_test.go
```diff
@@ -52,3 +54,45 @@ func TestBuildFunctionInstances_FunctionNames(t *testing.T) {
...
+	require.True(t, mi.Memory.ValidateAddrRange(uint32(0), uint64(100*PageSize)))
+	require.False(t, mi.Memory.ValidateAddrRange(uint32(0), uint64(100*PageSize+1)))
+	require.False(t, mi.Memory.ValidateAddrRange(uint32(1), uint64(100*PageSize)))
+	require.False(t, mi.Memory.ValidateAddrRange(uint32(100*PageSize), uint64(0)))
+}
+
+func TestPutUint32(t *testing.T) {
+	name := "test"
+	s := NewStore(nil)
+	mi := s.getModuleInstance(name)
```
mathetake, line 80: same comment for overkiling

wasm/store_test.go
mathetake: plus the test name is too broad.. 
```suggestion
func TestMemoryInstance_PutUint32(t *testing.T) {
```

wasi/testdata/args.go
nullpo-head: Thanks for the feedback. In fact, I don't insist on using `main`, to be precise calling all functions from `main`, at all. Actually I once wrote the test so that the test go code calls exported functions outside `_start`. The comment `// Let TinyGo runtime initialize the WASI environment by calling _start` you commented before was an old comment I wrote at that time, and it's now back.
So, I've updated so that the test code calls exported functions after calling `_start` only once. I've also commented why it calls `_start` and also stated that the WASI spec's restriction doesn't affect wazero side.

wasi/wasi_test.go
codefromthecrypt: if by "tests" you mean "tests/interop" works for me. I would like to isolate where tinygo is required as we can centralize the README and also make can invalidate easier when things are in the same place like that.

wasi/wasi.go
codefromthecrypt: yep that's what I meant. I can try to be more clear later, but I also use trailing comments a lot to remind what it is as ex not everything is fixed encoding either

wasi/wasi.go
codefromthecrypt: I'm not sure internal types need to have the same name as a spec especially when this isn't an array, but I won't fuss over it. Usually specs should try not to micro-manage internal details and leave it to naming the interface and butting out otherwise

codefromthecrypt:
Thanks for the progress, @nullpo-head I will take another review pass after you have addressed @mathetake's comments, to help avoid feedback storm!

wasi/wasi_test.go
mathetake: Ah I see! sg!

mathetake: so I've created the top `tests` directory, feel free to put `interop` under there

wasm/store_test.go
nullpo-head: I surely fixed it this time. Thanks.

wasm/store_test.go
nullpo-head: I surely fixed it this time. Thanks.

wasi/wasi_test.go
nullpo-head: I've moved to them `tests/wasi` as we discussed offline. Thanks all.

wasi/wasi.go
nullpo-head: Added it. Thanks!

interop/wasi/wasi_interop_test.go
nullpo-head: clarified as `language runtime`.

wasi/wasi.go
nullpo-head: Added. PTAL. Thanks!

wasi/wasi_test.go
nullpo-head: Fixed. Thanks!

wasm/store_test.go
nullpo-head: Ouch. Thanks for the nice catch. Fixed it.

wasm/store_test.go
nullpo-head: Fixed it. Thanks!

wasi/wasi_test.go
nullpo-head: Improved the comment. I though you mentioned about comments because it seems the even recent test codes don't seem to have godocs for each test functions, but let me know if your intention by `document` is not the comments.

wasi/wasi_test.go
nullpo-head: Yeah, this is a real instance of `HostFunctionCallContext`, and `Mock` was not a right name. Fixed it. Thanks!

Incidentally, the reason why I named it that in the first place is because this function directly builds the `HostFunctionCallContext` and does not guarantee invariant that might be guaranteed if it was built with the original method. For example, in the current implementation, the length of the memory is not a multiple of the page size, which is not a problem in this test though.

interop/wasi/wasi_interop_test.go
nullpo-head: Fixed. Thanks.

wasm/store_test.go
nullpo-head: Now these codes are removed due to simplification. JFYI, it was used to build a memory instance.

wasi/wasi_test.go
nullpo-head: Added it. Thanks!

wasi/wasi_test.go
nullpo-head: This branch is necessary to test the behavior of args API when a `wasiEnvironment` is instantiated without `Args` option in the first place, while the test case of `args == nil` of `newWasiStringArray` tests the behavior of `newWasiStringArray` when nil is explicitly passed to `newWasiStringArray`.

wasi/wasi.go
nullpo-head: Improved it. Thanks!

wasm/store_test.go
```diff
@@ -52,3 +54,45 @@ func TestBuildFunctionInstances_FunctionNames(t *testing.T) {
...
+
+	m := &Module{
+		MemorySection: []*LimitsType{{Min: 100}},
+	}
+
+	_, err := s.buildMemoryInstances(m, mi)
```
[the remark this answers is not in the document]
nullpo-head: Fixed it so that it builds an ad-hoc `MemoryInstance` from scratch instead of calling `buildMemoryInstances`.

wasm/store_test.go
nullpo-head: Fixed it so that it builds an ad-hoc MemoryInstance from scratch instead of calling buildMemoryInstances.

wasi/type_size.go
nullpo-head: Removed.

interop/wasi/testdata/args.go
nullpo-head: In my understanding, it is to make the expected behavior of the wasm module more explicit by calling named exported function rather than calling `_start` and everything is done inside `_start`, which was adrian's concern. It may have only a little difference this time because it has only one output, though. Correct me if I'm wrong @codefromthecrypt 

The discussion is here (https://github.com/tetratelabs/wazero/pull/166#discussion_r796160590) and I remember some are on Slack.

nullpo-head: ps: I'm also open to having it revert to calling print from `_start`. I think the both style makes sense this time.

codefromthecrypt: my point was missed slightly, but almost! I'll try to be clear

1. I think all wasi functions should be tested via export
2. I agree that there's value in testing wasi via tinygo in the adhoc-area

what was missed is that I don't think we do the "re-export for test" approach with tinygo. That's easier and more direct in the main source tree with %.wat even if we temporarily have to check in the result of wat2wasm. Put another way, tests that just call a wasi method re-exported shouldn't rely on tinygo.

codefromthecrypt: basically I would expect the interop style test to simply call the stuff via main func (which is invoked via wasi _start).

I would expect proxying the methods to be something easy to do in %.wat or even manually building the opcodes, and that should be with the normal unit tests.

wasi/wasi.go
```diff
@@ -277,16 +331,47 @@ func (w *WASIEnvironment) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32)
...
+//     A pointer to a buffer. args_get writes multiple *C.char pointers in sequence there. In other words, this is an array of *char in C.
+//     Each *C.char of them points to a command-line argument that is a null-terminated string.
+//     The number of this *C.char matches the value that args_sizes_get returns in argsCountPtr.
+// * argsBufPtr:
+//     A pointer to a buffer. args_get writes the command line arguments as null-terminated strings in the given buffer.
+//     The total number of bytes written there is the value that args_sizes_get returns in argsBufSizePtr. The caller must ensure that
+//     the buffer has the enough size.
+//     Each *C.char pointer that can be obtained from argsPtr points to the beginning of each of these null-terminated strings.
+// Link to the actual spec: https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-args_getargv-pointerpointeru8-argv_buf-pointeru8---errno
+// Reference: https://en.wikipedia.org/wiki/Null-terminated_string
```
codefromthecrypt, line 362:
```suggestion
// See: https://en.wikipedia.org/wiki/Null-terminated_string
```

wasi/wasi.go
```diff
@@ -277,16 +331,47 @@ func (w *WASIEnvironment) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32)
...
+// * argsPtr:
+//     A pointer to a buffer. args_get writes multiple *C.char pointers in sequence there. In other words, this is an array of *char in C.
+//     Each *C.char of them points to a command-line argument that is a null-terminated string.
+//     The number of this *C.char matches the value that args_sizes_get returns in argsCountPtr.
+// * argsBufPtr:
+//     A pointer to a buffer. args_get writes the command line arguments as null-terminated strings in the given buffer.
+//     The total number of bytes written there is the value that args_sizes_get returns in argsBufSizePtr. The caller must ensure that
+//     the buffer has the enough size.
+//     Each *C.char pointer that can be obtained from argsPtr points to the beginning of each of these null-terminated strings.
+// Link to the actual spec: https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-args_getargv-pointerpointeru8-argv_buf-pointeru8---errno
```
codefromthecrypt, line 361:
```suggestion
// See: https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-args_getargv-pointerpointeru8-argv_buf-pointeru8---errno
```

Makefile
```diff
@@ -21,11 +21,19 @@ bench_testdata_dir := tests/bench/testdata
 build.bench:
 	tinygo build -o $(bench_testdata_dir)/case.wasm -scheduler=none -target=wasi $(bench_testdata_dir)/case.go
 
-examples_testdata_dir := examples/testdata
+wasi_testdata_dir := ./examples/testdata ./tests/wasi/testdata
```
codefromthecrypt, line 24: is this a bad merge?

Makefile
```diff
@@ -21,11 +21,19 @@ bench_testdata_dir := tests/bench/testdata
...
 
 .PHONY: build.examples
 build.examples:
-	@find $(examples_testdata_dir) -type f -name "*.go" | xargs -Ip /bin/sh -c 'tinygo build -o $$(echo p | sed -e 's/\.go/\.wasm/') -scheduler=none -target=wasi p'
+	@$(MAKE) wasi_testdata_dir=./examples/testdata build.wasi
```
codefromthecrypt, line 28: same.. did you intend to change this?

tests/wasi/testdata/args.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"fmt"
+	"os"
+)
+
+func main() {
+}
+
+//export PrintArgs
+func PrintArgs() {
```
codefromthecrypt, line 12: please remove this and just call it from main as the comments clarifying this are completely distracting. MAY is not an issue except if we touch something. avoid that and just inline the printf in main.

as mentioned before, to call the function directly is perfectly in our ability to control and allow, and calling that directly should be in a unit test as "tests" won't necessarily ever be run by someone on change

tests/wasi/wasi_test.go
```diff
@@ -0,0 +1,96 @@
...
+			require.NoError(t, err)
+			wasiEnv := wasi.NewEnvironment(args, wasi.Stdout(stdoutBuf))
+
+			err = wasiEnv.Register(store)
+			require.NoError(t, err)
+
+			err = store.Instantiate(mod, "test")
+			require.NoError(t, err)
+
+			// XXX Strictly speaking, this test code violates the WASI specification.
```
codefromthecrypt, line 71: All this stuff is interesting, but is more about tinygo than our project, please remove it. In fact this is precisely why I've mentioned multiple times to not rely on tinygo to test anything important. If you are using it, don't abuse it if it has some very tight things it currently is able to do.


Remember, it is not for the sake of wazero, just that you wrote this test that we are doing anything. Please choose the other path by calling main and simplify the comment to "main is the only stable way to call a function on tinygo"

the paragraphs will drift with the tinygo impl and if anywhere is needed should be in tinygo docs.

interop/wasi/testdata/args.go
nullpo-head: Let me summarize your points, firstly points related to this `tests/wasi` file. So, your opinion on this time's change is 
- calling `fmt.Printf("...", os.Args)` in `main` is OK because this is an "interop style test".
- The current style which calls exported `PrintArgs` is the "re-export for test" approach.

Is it right? Then introduction of `PrintArgs` was my misunderstanding, and will revert it. Confirming just in case, because I once read the part below as it discouraged to do that. (sorry for not being able to read the nuances right!

>  I know you insist on using `main` to test this, please comment why, and if we expect args to never be able to be called except from main

codefromthecrypt (commented):
I think we are not finding the right balance of code docs yet, but also this change isn't worth the amount of effort by so many people put towards it. please address the comments and I'll clean it up post merge and we can discuss on that PR why

codefromthecrypt (approved):
approve on timeout! 

Let's handle the %.wat test post merge (I'll do it) because if we don't finish this now, it will force us into confusing discussions with other people.

Issue #170, opened 2022-01-29
Create InstantiatedModule type


codefromthecrypt:
There's a lifecycle requirement that is tricky to prove here:

> The start function is intended for initializing the state of a module. The module and its exports are not accessible before this initialization has completed.

https://www.w3.org/TR/wasm-core-1/#start-function%E2%91%A0

Right now, `Store` includes the instantiated module state internally. 
```go
	err = store.Instantiate(mod, "test")
        store.CallFunction("test", "fac-iter", in)
```

Due to this, to satisfy the above things, runtime checks like this have to happen internally.
```go
	m, ok := s.ModuleInstances[moduleName]
	if !ok {
		return nil, nil, fmt.Errorf("module '%s' not instantiated", moduleName)
	}
```

This are subject to racing which makes them hard to test also. Finally, it burdens the user a little as they have a way to invoke something when it can be in the wrong state. Ex.
```go
	// forgot to call store.Instantiate, or some other goroutine isn't done that.
        store.CallFunction("test", "fac-iter", in) // bombs
```

I think a cleaner design would be to make an InstantiatedModule type and move features that require start being invoked first, to be on that type. That features are only on the correct type, people wouldn't be tempted to do something at the wrong time or when something isn't available

Ex.
```go
	store.CallFunction("test", "fac-iter", in) // compile error!
	imod, err = store.Instantiate(mod, "test")
        imod.CallFunction("fac-iter", in) // compile error!
        factIter, ok = imod.GetFunction("fac-iter") // if ok, definitely that function exists
        factIter.invoke(ctx, in) // naming this "call" is also fine. whichever you can use it as often as you like
```

codefromthecrypt:
If we do this I also think we should think hard about whether adding and replacing functions on an instantiated module is something to support or not. Safety-wise it feels we shouldn't keep doing this, rather allow someone to change the imports/exports etc only up to `.Instantiate()`, then after that they are fixed.

codefromthecrypt:
from @mathetake 

> let's add store.GetModuleInstance as well or return *ModuleInstance on store.Instantiate to line with moduleInstance.GetExport API 😄

codefromthecrypt:
this is done via `api.Module`

PR #171, opened 2022-01-29
Handles export name collision and cleans up module instantiation

codefromthecrypt:
It is odd, but an export name can be empty, but cannot repeat. This adds
tests for that.

This also cleans up API usage where the caller had to remember which
name the module they instantiated had, by storing the name filed in the
module instance (similar to export in fact!).

Finally, this moves section names to the wasm package, where they can be
re-used.


wasm/store.go
```diff
@@ -716,43 +716,42 @@ func (s *Store) buildTableInstances(module *Module, target *ModuleInstance) (rol
 
 func (s *Store) buildExportInstances(module *Module, target *ModuleInstance) (rollbackFuncs []func(), err error) {
 	target.Exports = make(map[string]*ExportInstance, len(module.ExportSection))
-	for name, exp := range module.ExportSection {
```
codefromthecrypt, line 719: notice we don't really need the export section to be a map, and if it isn't, we retain insertion order, making late errors have more context

codefromthecrypt:
this allows parsing, but it is still unusable. I'll fix the other code

codefromthecrypt:
ok I refactored to include tests that access exported modules. I also added module name because it is error prone to have people remember the name they looked up with in the first place

wasm/jit/engine.go
```diff
@@ -223,9 +223,9 @@ const (
 	tableElementFunctionTypeIDOffest  = 8
 
 	// Offsets for wasm.ModuleInstance
-	moduleInstanceGlobalsOffset = 32
-	moduleInstanceMemoryOffset  = 56
-	moduleInstanceTablesOffset  = 64
+	moduleInstanceGlobalsOffset = 80
```
codefromthecrypt, line 226: @mathetake your tests saved my ass!

codefromthecrypt:
ok done polishing 😎

codefromthecrypt:
maps can have empty keys, I made a mistake somewhere and will fix

wasm/store.go
```diff
@@ -177,6 +178,27 @@ type (
 	FunctionTypeID uint64
 )
 
+// addExport adds and indexes the given export or errs if the name is already exported.
+func (m *ModuleInstance) addExport(name string, e *ExportInstance) error {
```
codefromthecrypt, line 182: since module knows its own name, now, we can centralize the errors in the module itself (including the name part of the error message), rather than always passing it the name we instantiated it with.

In the future we can also allow someone to instantiate a module with the name that was serialized into the namesection or in the text format, but I didn't do that now.

wasm/store.go
```diff
@@ -1765,57 +1767,39 @@ func (s *Store) AddHostFunction(moduleName, funcName string, fn reflect.Value) e
 	if err := s.engine.Compile(f); err != nil {
 		return fmt.Errorf("failed to compile %s: %v", f.Name, err)
 	}
-	m.Exports[funcName] = &ExportInstance{Kind: ExportKindFunc, Function: f}
 	s.addFunctionInstance(f)
-	return nil
+
+	return m.addExport(funcName, &ExportInstance{Kind: ExportKindFunc, Function: f})
```
codefromthecrypt, line 1772: This is done at the bottom to save code duplication. Ack this instantiates a conflicting function for no reason, but I think saving code and duplication is worth it!

If there is something we need to undo on err I don't mind adding that.

codefromthecrypt: actually I'll do that now!

wasm/internal/spectests/spec_test.go
```diff
@@ -294,10 +294,8 @@ func runTest(t *testing.T, newEngine func() wasm.Engine) {
 							}
 							inst, ok := store.ModuleInstances[moduleName]
 							require.True(t, ok, msg)
```
mathetake, line 296: let's add store.GetModuleInstance as well or return *ModuleInstance on store.Instantiate to line with moduleInstance.GetExport API 😄

mathetake (approved):
LGTM (the last comment maybe worth another PR)

wasm/internal/spectests/spec_test.go
codefromthecrypt: I would like to clean this up further in #170 once done the parser.. I've added it to the list and thanks for noticing!

codefromthecrypt:
Thanks for the look and the notes for later @mathetake !

PR #172, opened 2022-01-30
text: wasm.Module native text decoder

codefromthecrypt:
This reorganizes the code and strips all types except those needed to handle index resolution. The net result is we can parse directly to wasm.Module, now.

Here are some ancillary changes that helped with consolidation:
* replaces decoder.go with what was in parser.go since types are consolidated
* adds a tokenParser result to the lexer to consolidate dispatch logic formerly happening via resetting moduleParser.tokenParser
* introduces indexNamespace type to track ID -> index mappings
* extracts type_parser to handle the entire field and backfilled tests
* consolidates constants used for source position
* reduces wordiness of docs
* fixes broken logic on type use verification. Ex. `(func (type 12) (param i32))` must match type 12 or crash (it didn't!)


codefromthecrypt:
@mathetake I'm pretty done on the main code, there are some unit tests to backfill and that will take a day. Meanwhile, if you can help with review, I'd appreciate it!

codefromthecrypt:
ps benchmark results are effectively the same as presence of intermediate types didn't slow down anything vs now. in fact, splitting up the code by responsibility added one allocation. This change wasn't to increase perf anyway, it was to remove indirection, centralize responsibility and cut down some rather large types into slightly smaller ones.

wasm/text/decoder.go
```diff
@@ -1,64 +1,741 @@
 package text
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/tetratelabs/wazero/wasm"
 )
 
+// parserPosition holds the positional state of a parser. Values are also useful as they allow you to do a reference
```
codefromthecrypt, line 10: this file used to be parser.go, but merging it with decoder.go caused enough changes it is basically a new file

wasm/text/index_namespace.go
```diff
@@ -0,0 +1,177 @@
...
+	"fmt"
+
+	"github.com/tetratelabs/wazero/wasm"
+)
+
+func newIndexNamespace() *indexNamespace {
+	return &indexNamespace{idToIdx: map[string]wasm.Index{}}
+}
+
+// indexNamespace contains the count in an index namespace and any association of symbolic IDs to numeric indices.
```
codefromthecrypt, line 14: this type I hope to be able to mostly re-use in order to handle function locals, which are the next complex thing to attack and why I spent several days preparing for it here.

wasm/text/lexer.go
```diff
@@ -17,7 +17,7 @@ import (
 //
 // Note: Do not include the line and column number in a parsing error as that will be attached automatically. Line and
 // column are here for storing the source location, such as for use in runtime stack traces.
-type tokenParser func(tok tokenType, tokenBytes []byte, line, col uint32) error
+type tokenParser func(tok tokenType, tokenBytes []byte, line, col uint32) (tokenParser, error)
```
codefromthecrypt, line 20: this change reduced some complexity as we no longer have to use a module field to swap it.

mathetake (approved):

wasm/text/numbers.go
```diff
@@ -1,29 +1,25 @@
...
-// decodeUint32 decodes an uint32 from a tokenUN.
+// decodeUint32 decodes an uint32 from a tokenUN or returns false on overflow
 //
 // Note: Bit length interpretation is not defined at the lexing layer, so this may fail on overflow due to invalid
 // length known at the parsing layer.
 //
 // Note: This is similar to, but cannot use strconv.Atoi because WebAssembly allows underscore characters in numeric
 // representation. See https://www.w3.org/TR/wasm-core-1/#integers%E2%91%A6
-func decodeUint32(tokenBytes []byte) (uint32, error) { // TODO: hex
+func decodeUint32(tokenBytes []byte) (uint32, bool) { // TODO: hex
```
codefromthecrypt, line 10: only case was ever overflow

wasm/text/type_use_parser.go
```diff
@@ -0,0 +1,465 @@
...
+// typeUseParser parses an inlined type from a field such "func" and dispatches to onTypeUseEnd or onUnknownField.
+//
+// Ex. `(import "Math" "PI" (func $math.pi (result f32))`
+//                           starts here --^           ^
+//                         onTypeUseEnd resumes here --+
+//
+// Note: Unlike normal parsers, this is not used for an entire field (enclosed by parens). Rather, this only handles
+// "type", "param" and "result" inner fields in the correct order.
+// Note: typeUseParser is reusable. The caller resets via begin.
+type typeUseParser struct {
```
codefromthecrypt, line 24: this no longer has code inside to parse module-defined types, which allows it to be smaller than it was before. It was getting out-of-hand.

codefromthecrypt (commented):
notes!

codefromthecrypt:
still have work to go, but on the downward slope of the mountain. the type use parser is now unit testable and on its way to 95% coverage goal

codefromthecrypt:
once type use parser is unit tested, I will remove all the redundant tests in decoder_test, then clean up import and func tests if needed before merge

codefromthecrypt:
I'll stop here at 97.7% coverage, at a lower priority I'll move it closer to 100 and dedupe other tests. Meanwhile, I need to focus on the next step which is parsing locals. Thanks for the look @mathetake!

PR #174, opened 2022-01-31
jit: remove hostcall trampoline

mathetake:


mathetake:
sad but this has resulted in increasing the size of code in Go and ended up slowing down.. will come back to this PR later after we actually see this is the bottleneck

codefromthecrypt:
thanks for the update!

codefromthecrypt:
should we close this and reopen when actionable?

PR #180, opened 2022-02-01
Move wasm/internal/{spectets,example} under the top tests directory.

mathetake:
This commit creates the top level `tests` directory and moves:
* `wasm/internal/spectests`  -> `tests/spectest`
* `wasm/internal/example` -> `tests/codec`

The latter is to clarify the meaning of test which didn't seem obvious,
but not sure there's better! 

The rationale for this is that `internal` is not for e2e testing (actually it is the 
reserved keyword in Go), but for the actual packages for use.

In the future, I plan to move the current e2e-ish examples here


codefromthecrypt (approved):
great. This makes sense!

PR #181, opened 2022-02-02
Move e2e tests in JIT under tests dir

mathetake:
This PR generalizes e2e tests in `jit/engine_test.go`and moves them
`tests/adhoc` directory to run these tests against interpreter as well.

Notably this removes all the `testdata` directory from `wasm/*`
and helps isolates the usage of raw binary or texts into `examples`,
`bench` and `tests` directory.

During the course of generalization, I found the inconsistency on the 
panic handling between JIT and Interpreter which seems a bug of latter
so I fixed it.


examples/trap_test.go
```diff
@@ -1,41 +0,0 @@
-package examples
```
mathetake, line 1: this was definitely not an `example` but an e2e test, and is a dup of the e2e test used in jit previously. So I deleted it

examples/README.md
```diff
@@ -1,4 +0,0 @@
-
-This is where we put e2e tests of virtual machine. Please place examples so that `examples/foo_test.go` is testing
```
mathetake, line 2: this is a remnants from my hobby project day and confusing so deleted for now

wasm/interpreter/interpreter.go
```diff
@@ -427,34 +427,46 @@ func (it *interpreter) lowerIROps(f *wasm.FunctionInstance,
...
+
+	// shouldRecover is true when a panic at the origin of callstack should be recovered
+	//
+	// If this is the recursive call into Wasm (prevFrameLen != 0), we do not recover, and delegate the
+	// recovery to the first interpreter.Call.
+	//
+	// For example, given the call stack:
+	//	 "original host function" --(interpreter.Call)--> Wasm func A --> Host func --(interpreter.Call)--> Wasm function B,
+	// if the top Wasm function panics, we go back to the "original host function".
+	shouldRecover := prevFrameLen == 0
```
mathetake, line 439: that was the inconsistency I found

tests/README.md
```diff
@@ -1,3 +1,4 @@
 This directory contains tests which use multiple packages. For example:
+- `adhoc` contains variety of e2e tests, mainly to ensure the consistency in the behavior between engines.
```
codefromthecrypt, line 2: maybe call the directory engine instead, then say the tests inside are ad-hoc?

codefromthecrypt (approved):
fine either way, but I think the directory may be better "engine"

PR #182, opened 2022-02-02
Move benchmarks under tests directory.

mathetake:
This completes the reorganization and centralizes which directory 
contains raw Wasm text or binary files. Notably, now only `examples`
and `tests` top directory contain such files.


codefromthecrypt (approved):
progress!

PR #185, opened 2022-02-02
Instantiate wat instead of HostFunctionCallContext in unit tests

nullpo-head:
This PR changes the design of WASI unit tests so that it instantiates a wasm module from `wat` instead of instantiating `HostFunctionCallContext`. I hope this aligns with @codefromthecrypt's intention, but let me know if it's not.


wasi/testdata/args.wat
```diff
@@ -0,0 +1,30 @@
...
+;; this file is put here for now, because this is a temporary file until the parser supports
+;; the enough syntax, and this file will be embedded in unit test codes after that.
+(module
+  (import "wasi_snapshot_preview1" "args_sizes_get" (func $wasi_args_sizes_get (param i32 i32) (result i32)))
+  (import "wasi_snapshot_preview1" "args_get"       (func $wasi_args_get (param i32 i32) (result i32)))
+  (memory 0x200)  ;; just an arbitrary size big enough for tests
+  (export "memory" (memory 0))
+  ;; Define wrapper functions instead of just exporting the imported WASI APIS for now
+  ;; because wazero's interpreter has a bug that it crashes when an exported host function is called
+  ;; from the host environment, which will be fixed soon.
```
nullpo-head, line 16: I'll open an issue of it soon later. JFYA @mathetake

wasi/wasi_test.go
```diff
@@ -126,7 +129,8 @@ func TestArgsAPISucceed(t *testing.T) {
 				opts = append(opts, argsOpt)
 			}
 			wasiEnv := NewEnvironment(opts...)
-			hostFunctionCallContext := buildHostFunctionCallContext()
+			store, err := instantiateWasmStore("testdata/args.wasm", wasiEnv)
```
nullpo-head, line 132: I ended up changing all WASI API calls with `buildHostFunctionCallContext` to invocation through `wat` rather than adding new tests with the wat module invocation which tests only a part of test cases of the existing tests or duplicated test cases. 
However, I also remember that @codefromthecrypt recommended to add tests instead of removing tests (I guess it includes modifying as well) as follows, let me know if this doesn't align with your original suggestion.
```
rather adding a test that does this. you don’t need to remove the tests you made
```

wasi/wasi_test.go
```diff
@@ -254,14 +263,30 @@ func TestArgsGetAPIReturnError(t *testing.T) {
...
+			require.Equal(t, uint64(EINVAL), ret[0]) // ret[0] is returned errno
 		})
 	}
 }
 
-func buildHostFunctionCallContext() *wasm.HostFunctionCallContext {
-	return &wasm.HostFunctionCallContext{
-		Memory: &wasm.MemoryInstance{Buffer: make([]byte, wasm.MemoryPageSize), Min: 1},
+func instantiateWasmStore(wasmFile string, wasiEnv *WASIEnvironment) (*wasm.Store, error) {
+	buf, err := os.ReadFile(wasmFile)
```
codefromthecrypt, line 274: please use go:embed instead of this as less is best and also IDE can catch mistakes on path names. ex. https://github.com/tetratelabs/wazero/blob/main/tests/codec/codec_test.go#L27-L29

wasi/wasi_test.go
```diff
@@ -254,14 +263,30 @@ func TestArgsGetAPIReturnError(t *testing.T) {
...
+	if err != nil {
+		return nil, err
+	}
+	mod, err := wbinary.DecodeModule(buf)
+	if err != nil {
```
codefromthecrypt, line 279: since this is a test, pass `t *testing.T` then you can use require.NoErr which is less crufty as so many less if statements

wasi/wasi_test.go
```diff
@@ -254,14 +263,30 @@ func TestArgsGetAPIReturnError(t *testing.T) {
...
+		return nil, err
+	}
+	store := wasm.NewStore(interpreter.NewEngine())
+	err = wasiEnv.Register(store)
+	if err != nil {
+		return nil, err
+	}
+	err = store.Instantiate(mod, "test")
```
codefromthecrypt, line 287: pass the name of the module as a var then in your tests, you can use `moduleName` which will be obviously same as here

wasi/wasi_test.go
codefromthecrypt: I like this better, just I wanted to give you flexibility to keep the old ones if you wanted

codefromthecrypt (approved):
looks good! I will follow-up with a PR that uses the text compiler.

I think it is the best place and exactly what I wanted to see (here vs in /tests) because it can and should be run on change

wasi/testdata/args.wat
mathetake: oh nice! Thank you for finding this!

mathetake (approved):

wasi/wasi_test.go
nullpo-head: ah nice catch! thanks, I've updated it.

wasi/wasi_test.go
nullpo-head: Updated as suggested. Thanks.

wasi/wasi_test.go
nullpo-head: Updated as suggested. Thanks!

PR #186, opened 2022-02-03
Add wazero specific limitations

mathetake:
This PR add more documentations on wazero specific limits,
and impose the limitation in the module instantiation phase.

Also, this backfills tests for executeConstantExpression and buildGlobalInstances
and fixed a bug in the former.

The implementation on the function value stack height limit will be in a followup PR
as it would be rather large change.

PR #189, opened 2022-02-03
Return error on stack height limit exceed

mathetake:
This commit implements static checks on the limit of maximum 
values on the stack a function can potentially reaches.
This is the left TODO from #186.

This also moves the `validateFunction` from store.go into func_validation.go


wasm/func_validation.go
```diff
@@ -0,0 +1,948 @@
+package wasm
```
mathetake, line 1: this is extracted from store.go as this can be decoupled from store and a bit complex
compared to others in store.go!

wasm/func_validation.go
```diff
@@ -0,0 +1,948 @@
...
+	}
+
+	if len(controlBloclStack) > 0 {
+		return fmt.Errorf("ill-nested block exists")
+	}
+
+	if valueTypeStack.maximumStackPointer > maxStackValues {
+		return fmt.Errorf("function too large: potentially could have %d values on the stack with the limit %d",
+			valueTypeStack.maximumStackPointer, maxStackValues)
+	}
```
mathetake, line 827: this (and the addition of `maximumStackPointer` on `valueTypeStack`) is the only change in this commit

wasm/func_validation_test.go
```diff
@@ -0,0 +1,38 @@
...
+	f.Body = append(f.Body, OpcodeEnd)
+
+	t.Run("not exceed", func(t *testing.T) {
+		err := validateFunctionInstance(f, nil, nil, nil, nil, nil, max+1)
+		require.NoError(t, err)
+	})
+	t.Run("exceed", func(t *testing.T) {
+		err := validateFunctionInstance(f, nil, nil, nil, nil, nil, max)
+		require.Error(t, err)
+		expMsg := fmt.Sprintf("function too large: potentially could have %d values on the stack with the limit %d", valuesNum, max)
```
codefromthecrypt, line 35: is there a similar message somewhere from another tool that crashes when there are too many stack values?

I would suggest adding the function name or index to the message.

Also, "potentially" and "could have" are both passive. maybe "function[%d] may have %d stack values, which  exceeds limit %d"

wasm/module.go
```diff
@@ -433,3 +433,27 @@ type NameMapAssoc struct {
 	Index   Index
 	NameMap NameMap
 }
+
+// allDeclarations returns all declarations for functions, globals, memories and tables in a module including imported ones.
+func (m *Module) allDeclarations() (functions []Index, globals []*GlobalType, memories []*MemoryType, tables []*TableType) {
```
codefromthecrypt, line 438: TestModule_AllDeclarations?

codefromthecrypt (approved):
message nit and a mechanical unit test, then good to go!

mathetake:
Thanks for the help!

PR #192, opened 2022-02-03
text: adds memory and memory export

codefromthecrypt:
This allows the WASI test to not require a checked in binary


wasi/testdata/args.wat
```diff
@@ -9,7 +9,7 @@
 (module
   (import "wasi_snapshot_preview1" "args_sizes_get" (func $wasi_args_sizes_get (param i32 i32) (result i32)))
   (import "wasi_snapshot_preview1" "args_get"       (func $wasi_args_get (param i32 i32) (result i32)))
-  (memory 0x200)  ;; just an arbitrary size big enough for tests
```
codefromthecrypt, line 12: punted hex for now

wasi/testdata/args.wat
```diff
@@ -9,7 +9,7 @@
...
+  (memory 512)  ;; just an arbitrary size big enough for tests
```
mathetake, line 12: this is the page size so maybe this is too big.. 😄  maybe single digits or at most tens is enough I think 😄

codefromthecrypt: oh interesting. I was wondering and thought maybe it was bytes!

mathetake (approved):
This is so exciting!!!

nullpo-head (approved):

codefromthecrypt:
thanks for the looks @mathetake and @nullpo-head!

PR #194, opened 2022-02-03
Make address validations and WASI handle no-memory modules correctly

nullpo-head:
The current address validation functions and WASI implementations didn't take into account the case of no memory. This PR fixes it.

### DIFF
--- a/wasm/store.go
+++ b/wasm/store.go
@@ -789,13 +789,19 @@ func (s *Store) buildExportInstances(module *Module, target *ModuleInstance) (ro
 
 // ValidateAddrRange checks if the given address range is a valid address range.
 // It accepts rangeSize as uint64 so that callers can add or multiply two uint32 addresses
-// without overflow. eg.) ValidateAddrRange(uint32Offset, uint64(uint32Size) + 1)
+// without overflow. For example, `ValidateAddrRange(uint32Offset, uint64(uint32Size) + 1)`.
+// ValidateAddrRange works even if `m` is nil so that memory range validation
+// against a module with no memory exported can be done in a consistent way.
 func (m *MemoryInstance) ValidateAddrRange(addr uint32, rangeSize uint64) bool {
+	if m == nil {
+		// Address validation is done for a module with no memory exported.
+		return false
+	}
 	return uint64(addr) < uint64(len(m.Buffer)) && rangeSize <= uint64(len(m.Buffer))-uint64(addr)
 }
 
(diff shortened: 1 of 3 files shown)


wasm/store.go
```diff
@@ -789,13 +789,19 @@ func (s *Store) buildExportInstances(module *Module, target *ModuleInstance) (ro
 
 // ValidateAddrRange checks if the given address range is a valid address range.
 // It accepts rangeSize as uint64 so that callers can add or multiply two uint32 addresses
-// without overflow. eg.) ValidateAddrRange(uint32Offset, uint64(uint32Size) + 1)
+// without overflow. For example, `ValidateAddrRange(uint32Offset, uint64(uint32Size) + 1)`.
+// ValidateAddrRange works even if `m` is nil so that memory range validation
+// against a module with no memory exported can be done in a consistent way.
 func (m *MemoryInstance) ValidateAddrRange(addr uint32, rangeSize uint64) bool {
+	if m == nil {
```
codefromthecrypt, line 796: would this be less awkward if we instead made sure this is never nil and use a constant `MemoryInstance(0,0,[]byte{})` instead?

mathetake: Starting with the memory length zero but later growing is valid so nil (memory is not declared in module) and memory with length zero is different

mathetake: But yeah max zero sounds ok for non memory case

codefromthecrypt: yeah because I noticed earlier it is valid to decode max zero, so if we have a bug where are writing we'll need to fix it anyway

nullpo-head: yeah, the idea of always having at least  0-max memory sounds good. I'll change the code.

nullpo-head: considering that @codefromthecrypt is doing refactoring on `MemoryInstance`, how about @codefromthecrypt doing the introduction of `NoMemoryInstance` constant at the same time and close this PR? It will conflict in a lot of places.

codefromthecrypt: you mean #204? works for me

codefromthecrypt: only thing is that my brain is on another change, so probably won't do 204 until tomorrow or weds

nullpo-head: yes, I mean #204. It sounds ok to me. Thanks!

nullpo-head:
Closing this PR in favor of the coming refactoring on the interface of `MemoryInstance` for #204

Issue #195, opened 2022-02-03
Host functions called from another host function cannot know the right MemoryInstance


nullpo-head:
In the current implementation, host functions try to get the `MemoryInstance` of `HostFunctionCallContext` from `<function>.ModuleInstance.Memory` of the parent stack's (in interpreter) or of the function itself (in jit).
However, a host function doesn't have associated `MemoryInstance` nor wasm modules in the first place. So, a host function only finds `nil` as `MemoryInstance` when it's called from another host function. 

FYI, it looks wasmtime also doesn't provide the access to memory instances when a host function is called from another host function because there's no caller context in that case, but let me check.

https://github.com/tetratelabs/wazero/blob/a6cd1549ba093e35d09fe07cd1a049b70e798185/wasm/interpreter/interpreter.go#L516
https://github.com/tetratelabs/wazero/blob/a6cd1549ba093e35d09fe07cd1a049b70e798185/wasm/jit/engine.go#L390

This also prevents the WASI implementation from testing APIs by "import-and-exported-back" style, where tests call each WASI API by having a wasm module import a WASI API and re-export it to the host.

codefromthecrypt:
I was thinking in the PR maybe we can default to a zero length constant Memory or initialize an empty one if we are afraid some code might mutate it.

```
zero := uint32(0)
noMemory := &MemoryInstance{Min: 0, Max: &zero, Buffer: []byte{}}
```

Meanwhile these safety issues keep coming up which casts doubt on the idea that exposing the memory buffer directly is a good idea. At some point we may want to use an interface to centralize all the bug handling and make it less easy to make them

mathetake:
I think this is something unspecified so I just passes the memory instance from the module on this host function is initialized currently if it is directly called by other host funcs, but year we should check how other runtimes do..

codefromthecrypt:
for this one, please use the available memory (if someone forgot to add a memory, we can treat nil as zero max). I don't think there's a reasonable alternative as WebAssembly 1.0 (MVP) can have at most 1 memory. If we later find some new information, we can revise, but meanwhile not crash

PR #196, opened 2022-02-03
Fix interpreter cannot call "imported-and-exported-back" function

nullpo-head:
This PR fixes a bug that interpreter crashes when it calls such a guest function that was imported from the host into a wasm module, but was exported to the host again. This is an interpreter-only bug and JIT has been correctly handling this without problem.

Please note that this PR alone will not allow us to test the WASI APIs by calling such "imported and exported back" WASI API functions. It turned out that there is another issue. See #195

### DIFF
--- a/tests/engine/adhoc_test.go
+++ b/tests/engine/adhoc_test.go
@@ -32,6 +32,7 @@ func runTests(t *testing.T, newEngine func() wasm.Engine) {
 	unreachable(t, newEngine)
 	memory(t, newEngine)
 	recursiveEntry(t, newEngine)
+	importedAndExportedFunc(t, newEngine)
 }
 
 func fibonacci(t *testing.T, newEngine func() wasm.Engine) {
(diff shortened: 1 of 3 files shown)


tests/engine/adhoc_test.go
```diff
@@ -32,6 +32,7 @@ func runTests(t *testing.T, newEngine func() wasm.Engine) {
 	unreachable(t, newEngine)
 	memory(t, newEngine)
 	recursiveEntry(t, newEngine)
+	importedAndExportedFunc(t, newEngine)
```
nullpo-head, line 35: Can we rewrite here like the following? Do we intentionally avoid it or is it ok to do that?
```go

	t.Run("fibonacci", func(t *testing.T) {
		fibonacci(t, newEngine)
	})
	t.Run("fac", func(t *testing.T) {
		fac(t, newEngine)
	})
	t.Run("unreachable", func(t *testing.T) {
		unreachable(t, newEngine)
	})
	t.Run("memory", func(t *testing.T) {
		memory(t, newEngine)
	})
	t.Run("recursive entry", func(t *testing.T) {
		recursiveEntry(t, newEngine)
	})
	t.Run("imported-and-exported func", func(t *testing.T) {
		importedAndExportedFunc(t, newEngine)
	})
```

codefromthecrypt: sounds good

tests/engine/adhoc_test.go
```diff
@@ -170,3 +172,28 @@ func recursiveEntry(t *testing.T, newEngine func() wasm.Engine) {
 	_, _, err = store.CallFunction("test", "main", uint64(1))
 	require.NoError(t, err)
 }
+
+func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
+	buf, err := os.ReadFile("testdata/imported_and_exported_func.wat")
```
codefromthecrypt, line 177: maybe just backtrack this as for small text inlining > go:embed > os.ReadFile

tests/engine/adhoc_test.go
```diff
@@ -170,3 +172,28 @@ func recursiveEntry(t *testing.T, newEngine func() wasm.Engine) {
...
+	require.NoError(t, err)
+	mod, err := text.DecodeModule(buf)
+	require.NoError(t, err)
+
+	store := wasm.NewStore(newEngine())
+
+	// addInt is imported by the wasm module, and the wasm module exports addInt back
```
codefromthecrypt, line 184: ex if you inline you will already have the comments about what's going on in the source in the same test

codefromthecrypt (approved):
some style suggestions, thanks for fixing the first bug in the import-and-export-back!

mathetake (approved):
Awesome!

wasm/interpreter/interpreter.go
```diff
@@ -477,12 +477,12 @@ func (it *interpreter) Call(f *wasm.FunctionInstance, params ...uint64) (results
 		return
 	}
 
+	for _, param := range params {
+		it.push(param)
+	}
 	if g.hostFn != nil {
 		it.callHostFunc(g, params...)
```
mathetake, line 484: Could you delete `_ ...uint64` from `callHostFunc`'s signature? https://github.com/tetratelabs/wazero/blob/07248537ca015313b4dcc16e2507f4905c8eac91/wasm/interpreter/interpreter.go#L495

nullpo-head: sur

tests/engine/adhoc_test.go
nullpo-head: inlined!

PR #197, opened 2022-02-03
Use goembed for all tests

mathetake:

### DIFF
--- a/tests/spectest/spec_test.go
+++ b/tests/spectest/spec_test.go
@@ -20,7 +20,8 @@ import (
 	"github.com/tetratelabs/wazero/wasm/jit"
 )
 
-//go:embed testdata/*
+//go:embed testdata/*.wasm
+//go:embed testdata/*.json
 var testcases embed.FS
 
 type (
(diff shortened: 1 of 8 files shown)


mathetake:
Oh didn't know that goembed can be used for directory so I can use it for spectests as well! lemme

mathetake:
ok ready!

tests/spectest/spec_test.go
```diff
@@ -20,6 +20,9 @@ import (
 	"github.com/tetratelabs/wazero/wasm/jit"
 )
 
+//go:embed testdata/*
```
codefromthecrypt, line 23: nit does this accept *.wasm? it will help me in inventory of which tests still use binaries

codefromthecrypt (approved):
this is great. at least in goland this is tighter as the IDE can tell when you reference a file that doesn't exist. That's as good imho as the reduction of error handling

tests/spectest/spec_test.go
mathetake: not really, we have to read *.json first and inside of it wasm binary names are specified.

mathetake: https://github.com/tetratelabs/wazero/pull/197/commits/58cdd090272be91978c2fef72ceaf40d10c42172 specified at least!

PR #198, opened 2022-02-03
ci: add arm64 test job

mathetake:
This commit adds a CI job for running test on arm64 arch with qemu. The idea 
is just building test binaries with `go test -c` on amd64, not arm64 and 
running  these built test binaries inside the emulation.  This would reduce
the cost of build time in emulation.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -108,10 +108,6 @@ jobs:
           image: tonistiigi/binfmt:latest
           platforms: arm64
 
-      # Reference: https://www.stereolabs.com/docs/docker/building-arm-container-on-x86/
-      - name: Setup the multiarch docker settings
-        run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes # This step will execute the registering scripts
-
       - name: Run built test binaries inside the qemu
         run: docker run -w /tmp/wazero -v $(pwd):/tmp/wazero --rm -t arm64v8/ubuntu /bin/bash -c 'find . -name "*.test" | xargs -Ibin bash -c "bin test.v"'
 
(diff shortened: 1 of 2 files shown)


mathetake:
virtually no overhead!

.github/workflows/commit.yaml
```diff
@@ -70,6 +70,46 @@ jobs:
...
+        env:
+          GOARCH: arm64
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: arm64
+
+      - name: Run built test binaries inside the qemu
+        run: docker run -w /tmp/wazero -v $(pwd):/tmp/wazero --rm -t arm64v8/ubuntu /bin/bash -c 'find . -name "*.test" | xargs -Ibin bash -c "bin test.v"'
```
codefromthecrypt, line 111: maybe comment why we can't use a make target here instead?

codefromthecrypt (approved):
I have reservations about using docker+qemu for tests vs real metal, but more reservations about not testing :D Let's start with this and as long as the emu is correct and tests not super slow, we can keep it!

.github/workflows/commit.yaml
```diff
@@ -71,7 +71,13 @@ jobs:
       - run: make test
 
   test_arm64:
-    name: Test (ubuntu, Go-${{ matrix.go-version }}) / arm64
+    # Here we cannot use "make test" to test on arm64. We have to run tests inside the emulation, but running all tests in the emulation
+    # comes with costs as we have to install all the build dependency there and build time itself can be a bottleneck.
+    # Therefore, we choose to build arm64 test binaries on usual amd64 environment so we can leverage Go's cross compilation
+    # functionality and reduce the maintenance cost of build dependencies outside of amd64 platform, and just we run the 
+    # built test binaries inside of the emulation. This way, the overhead of emulation becomes ignorable.
+    # Note: this can be generalized to any GOARCH.
```
codefromthecrypt, line 80: fantastic!

Issue #187, opened 2022-02-03
Tracking issue for arm64 JIT compiler 


mathetake:

mathetake:
For those interested; after #220 and #217, 40% done, but the all the ground work has done so everything left should be mechanical and easy because we already have amd64 impl as the reference!

mathetake:
now the calling functions is supported so everything should be mechanical except br_table which must be tricky

mathetake:
22 instructions left!

mathetake:
after #259, 2 instructions left

mathetake:
ok the left is to pass spectests!

mathetake:
Enjoy!

PR #199, opened 2022-02-04
jit: scaffold arm64 JIT engine

mathetake:
This commit builds the scaffold for arm64 JIT engine.
Notably, we can run go test ./wasm/jit/... and pass some
GOARCH independent tests on arm64.

part of #187

### DIFF
--- a/wasm/jit/jit_arm64.s
+++ b/wasm/jit/jit_arm64.s
@@ -1,0 +1,11 @@
+//go:build arm64
+// +build arm64
+
+#include "funcdata.h"
+#include "textflag.h"
+
+// jitcall(codeSegment, engine)
+TEXT ·jitcall(SB),NOSPLIT|NOFRAME,$0-16
+        MOVD codeSegment+0(FP),R1  // Load the address of native code.
+        MOVD engine+8(FP),R0       // Load the address of engine.
+        JMP (R1)                   // Jump to native code.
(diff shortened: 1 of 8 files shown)


wasm/jit/jit_value_location.go
```diff
@@ -0,0 +1,287 @@
+package jit
```
mathetake, line 1: This is just extracted the GOARCH independent parts of code from `jit_value_location_amd64.go` without any change!

wasm/jit/jit_value_location_test.go
```diff
@@ -1,14 +1,10 @@
-//go:build amd64
```
mathetake, line 1: not this is arch-independent tests passing both on arm64 and amd64.

mathetake: now

wasm/jit/jit_arm64.s
```diff
@@ -0,0 +1,11 @@
+//go:build arm64
+// +build arm64
+
+#include "funcdata.h"
+#include "textflag.h"
+
+// jitcall(codeSegment, engine)
+TEXT ·jitcall(SB),NOSPLIT|NOFRAME,$0-16
+        MOVD codeSegment+0(FP),R1  // Load the address of native code.
+        MOVD engine+8(FP),R0       // Load the address of engine.
```
mathetake, line 10: only difference from amd64.s is that MOVD is for 64bit regs with arm64 assembler, not MOVQ as in amd64

nullpo-head (approved):

codefromthecrypt (commented):
cooooool

PR #200, opened 2022-02-04
ci: generalize arm64 job to RISC-V 64.

mathetake:


mathetake:
whew all tests pass on RISCV!

nullpo-head (approved):
LGTM

.github/workflows/commit.yaml
```diff
@@ -70,25 +68,25 @@ jobs:
 
       - run: make test
 
-  test_arm64:
-    # Here we cannot use "make test" to test on arm64. We have to run tests inside the emulation, but running all tests 
+  test_non_amd64:
+    # Here we don't use "make test" to test on non-amd64. We have to run tests inside the emulation, but running all tests
```
codefromthecrypt, line 72: Instead of double-negative, reword a bit.

"Only amd64 is viable with "make test" because other architectures require emulation: the process of installing depedencies and building the project through an emulator is unacceptably slow.

Instead, we build cross-platform test binaries on our amd64 runner, which limits what is emulated to only executing the tests. This is why below uses `test.v` (the test binary) instead of `make`

codefromthecrypt (approved):
added a suggestion for rewording the rationale in a simpler way

PR #202, opened 2022-02-04
text: fixes code section ID parsing a function

codefromthecrypt:
This moves logic to func_parser.go, which allows us to keep a coherent
function index when parsing code.


wasm/text/decoder.go
```diff
@@ -625,8 +586,6 @@ func (p *moduleParser) resolveTypeIndices(module *wasm.Module) error {
 
 // resolveFunctionIndices ensures any indices point are numeric or returns a FormatError if they cannot be bound.
 func (p *moduleParser) resolveFunctionIndices(module *wasm.Module) error {
-	// TODO: remove when funcParser points to the correct index
```
codefromthecrypt, line 628: this is the main thing

mathetake (approved):

codefromthecrypt:
gracias para verlo

PR #208, opened 2022-02-07
ci: remove build.lib

mathetake:


codefromthecrypt (approved):
sounds good because our CI will catch problems and as we centralize source more, things left uncompiled per platform become smaller, and so does drift risk

PR #211, opened 2022-02-07
Fix float32-type parameter not working for host function

nullpo-head:
This PR fixes that `float32` parameters were not correctly passed to the host function.

### DIFF
--- a/tests/engine/adhoc_test.go
+++ b/tests/engine/adhoc_test.go
@@ -230,34 +230,35 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
 func hostFuncWithFloat32Param(t *testing.T, newEngine func() wasm.Engine) {
 	ctx := context.Background()
 	mod, err := text.DecodeModule([]byte(`(module
-		;; 'identity_f32' is a host function which accepts a float32 param.
-		;; It should just return the given value without breaking it.
+		;; 'identity_f32' just returns the given value as is.
 		(import "test" "identity_f32" (func $identity_f32 (param f32) (result f32)))
 
-		;; 'call_identity_f32' proxies test.identity_f32 via call in order to test floats aren't corrupted
-		(func $call_identity_f32 (param f32) (result f32)
+		;; 'call->identity_f32' proxies test.identity_f32 via call in order to test floats aren't corrupted
+		;; when the float values are routing through OpCodeCall
+		(func $call->test.identity_f32 (param f32) (result f32)
 			local.get 0
 			call $identity_f32
 		)
-		(export "call_identity_f32" (func $call_identity_f32))
+		(export "call->test.identity_f32" (func $call->test.identity_f32))
 		)`))
 	require.NoError(t, err)
 
 	store := wasm.NewStore(newEngine())
 
-	// identityF32 is a host function that just returns the given float32 parameter to the guest.
+	// identityF32 is a host function that just returns the given float32 parameter as is to the module.
 	identityF32 := func(ctx *wasm.HostFunctionCallContext, value float32) float32 {
 		return value
 	}
-	err = store.AddHostFunction("env", "identity_f32", reflect.ValueOf(identityF32))
+	err = store.AddHostFunction("test", "identity_f32", reflect.ValueOf(identityF32))
 	require.NoError(t, err)
 
-	err = store.Instantiate(mod, "test")
+	err = store.Instantiate(mod, "mod")
 	require.NoError(t, err)
 
 	expectedF32Val := float32(math.MaxFloat32) // arbitrary f32 value
-	// This call should return the given `expectedF32Val` without corrupting the value.
-	results, resultTypes, err := store.CallFunction(ctx, "test", "call_identity_f32", uint64(math.Float32bits(expectedF32Val)))
+	// This call should return the given `expectedF32Val` as is.
+	// That ensures that the float values are not corrupted when they are routed through OpCodeCall.
+	results, resultTypes, err := store.CallFunction(ctx, "mod", "call->test.identity_f32", uint64(math.Float32bits(expectedF32Val)))
 	require.NoError(t, err)
 	require.Len(t, results, len(resultTypes))
 	require.Equal(t, wasm.ValueTypeF32, resultTypes[0])
--- a/tests/engine/adhoc_test.go
+++ b/tests/engine/adhoc_test.go
@@ -249,31 +249,28 @@ func hostFuncWithFloatParam(t *testing.T, newEngine func() wasm.Engine) {
 
 	tests := []struct {
 		testName          string
-		typeName          string        // "f32" or "f64"
+		floatType         byte          // wasm.ValueTypeF32 or wasm.ValueTypeF64
 		identityFloatFunc reflect.Value // imported as 'identity_float' to the wasm module
 		floatParam        uint64        // passed to identityFloatFunc as a parameter
-		expectedFloatType byte          // wasm.ValueTypeF32 or wasm.ValueTypeF64
 		expectedFloatVal  float64
 	}{
 		{
-			testName: "host function with f32 param",
-			typeName: "f32",
+			testName:  "host function with f32 param",
+			floatType: wasm.ValueTypeF32,
 			identityFloatFunc: reflect.ValueOf(func(ctx *wasm.HostFunctionCallContext, value float32) float32 {
 				return value
 			}),
-			floatParam:        uint64(math.Float32bits(math.MaxFloat32)), // float bits as a uint32 value, but casted to uint64 to be passed to CallFunction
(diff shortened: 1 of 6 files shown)


tests/engine/adhoc_test.go
```diff
@@ -194,9 +198,9 @@ func recursiveEntry(t *testing.T, newEngine func() wasm.Engine) {
 	require.NoError(t, err)
 }
 
+// Test that the engine can call "imported-and-then-exported-back" function correctly
```
codefromthecrypt, line 201:
```suggestion
// importedAndExportedFunc fails if the engine cannot call an "imported-and-then-exported-back" function
```

mathetake (approved):
Nice..

tests/engine/adhoc_test.go
```diff
@@ -221,3 +225,41 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
 	require.NoError(t, err)
 	require.Equal(t, uint64(42), results[0])
 }
+
+// Test that host functions can handle float parameter without corrupting the value
```
codefromthecrypt, line 229:
```suggestion
//  hostFuncWithFloat32Param fails if a float parameter corrupts a host function value
```

tests/engine/adhoc_test.go
```diff
@@ -221,3 +225,41 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
...
+func hostFuncWithFloat32Param(t *testing.T, newEngine func() wasm.Engine) {
+	ctx := context.Background()
+	mod, err := text.DecodeModule([]byte(`(module
+		;; 'test_f32param' is a host function which accepts a f32 param
+		;; and tests if it is an expected value.
+		(import "env" "test_f32param" (func $test_f32param (param f32)))
+
+		;; 'call_test_f32' is an exported guest function to call
```
codefromthecrypt, line 237: "guest" is jargon not defined in wasm iirc, so let's avoid it. or if it is the right jargon let me know :)

tests/engine/adhoc_test.go
```diff
@@ -221,3 +225,41 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
 	require.NoError(t, err)
 	require.Equal(t, uint64(42), results[0])
 }
+
+// Test that host functions can handle float parameter without corrupting the value
+func hostFuncWithFloat32Param(t *testing.T, newEngine func() wasm.Engine) {
+	ctx := context.Background()
+	mod, err := text.DecodeModule([]byte(`(module
+		;; 'test_f32param' is a host function which accepts a f32 param
```
codefromthecrypt, line 233: since import implies it is a host function (in a single module setup anyway) and the signature says itself anyway we can skip clarifying these things. focused comments will end up winning. Not sure if it is normal or not to quote the function name, so feel free to adjust:

`;; test_f32param crashes the runtime if the parameter is unexpected.

tests/engine/adhoc_test.go
```diff
@@ -221,3 +225,41 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
...
+		;; and tests if it is an expected value.
+		(import "env" "test_f32param" (func $test_f32param (param f32)))
```
codefromthecrypt, line 235:
```suggestion
		(import "test" "f32param" (func $test.f32param (param f32)))
```

mathetake:
I was kind of under the impression that this is tested in spectests, but I realized that all the host funcs used there is just a mocks :D https://github.com/tetratelabs/wazero/blob/main/tests/spectest/spec_test.go#L189-L197

tests/engine/adhoc_test.go
```diff
@@ -221,3 +225,41 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
...
+		)
+		(export "call_test_f32param" (func $call_test_f32param))
+		)`))
+	require.NoError(t, err)
+
+	store := wasm.NewStore(newEngine())
+
+	expectedFloat32Val := float32(math.MaxFloat32) // arbitrary float32 value.
+	// testF32Param is called from the guest function, call_test_f32param.
+	// This function expects the guest to pass expectedF32Val.
```
codefromthecrypt, line 251: // testF32Param is a host function (accessed via import), that crashes the runtime (panic) if the float value it receives is unexpected.

tests/engine/adhoc_test.go
```diff
@@ -221,3 +225,41 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
...
+	testF32Param := func(ctx *wasm.HostFunctionCallContext, value float32) {
+		require.Equal(t, expectedFloat32Val, value)
+	}
+	err = store.AddHostFunction("env", "test_f32param", reflect.ValueOf(testF32Param))
+	require.NoError(t, err)
+
+	err = store.Instantiate(mod, "test")
+	require.NoError(t, err)
+
+	// `call_test_f32param` calls the host function `testF32Param` with `expectedF32Val` as its parameter,
```
codefromthecrypt, line 261: In general, we don't need to summarize the signature if it is one line below. The impact is the important part. Ex. 

`This invokes call->test.f32param to ensure float parameters aren't corrupted when routing through OpCodeCall`

tests/engine/adhoc_test.go
```diff
@@ -221,3 +225,41 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
...
+func hostFuncWithFloat32Param(t *testing.T, newEngine func() wasm.Engine) {
+	ctx := context.Background()
+	mod, err := text.DecodeModule([]byte(`(module
+		;; 'test_f32param' is a host function which accepts a f32 param
+		;; and tests if it is an expected value.
+		(import "env" "test_f32param" (func $test_f32param (param f32)))
+
+		;; 'call_test_f32' is an exported guest function to call
+		;; the host function, 'test_f32param', from the guest.
+		(func $call_test_f32param (param f32)
```
codefromthecrypt, line 239: the ID namespace is amazingly open so you can be very literal.

```suggestion
		;; call->test.f32param proxies test.f32param via call in order to test floats aren't corrupted.
		(func $call->test.f32param (param f32)
```

codefromthecrypt (approved):
neat stuff. If we are interested in float corruption, should we widen the signature and test both 32 and 64 at the same time (add one more param)?

tests/engine/adhoc_test.go
mathetake: +1

nullpo-head: Fixed it. Thanks!

tests/engine/adhoc_test.go
nullpo-head: `test_f32param` is replaced with `identity_float` now, and I skipped commenting about its signature as suggested.

tests/engine/adhoc_test.go
nullpo-head: Updated as suggested. Thanks.

nullpo-head:
I've updated so that 1. it does `require.Equals` on the result of the callFunction, 2. test `flot64` as well. PTAL @codefromthecrypt . 
The latter change introduced a bit duplicated code, let me know if there's more sophisticated way to do this.

codefromthecrypt (approved):
Good as is, though I wonder if you can use a test matrix instead of repeating the t.Run body.

For example, wazero currently requires parameters and results to be captured as uint64. Seems like almost everything except defining the host functions themselves can be test parameters (ex. fn reflect.Value as a test struct field, and the two functions used to populate those defined before the matrix)

wdyt?

nullpo-head:
You're right, it can be a table-driven one. I had wrongly assumed that it couldn't because I had lost track of the fact that `reflect.value` could be used and that the parameters and return value were always `uint64` after all. I've improved it. Thank you!

tests/engine/adhoc_test.go
```diff
@@ -221,3 +227,83 @@ func importedAndExportedFunc(t *testing.T, newEngine func() wasm.Engine) {
...
+			local.get 0
+			call $test.identity_float
+		)
+		(export "call->test.identity_float" (func $call->test.identity_float))
+		)`)
+	require.NoError(t, err)
+
+	tests := []struct {
+		testName          string
+		typeName          string        // "f32" or "f64"
```
codefromthecrypt, line 252: maybe export `valueTypeName` then you can use `ValueType` here instead of a string (you can use `valueTypeName(vt)` at the call site I mean

codefromthecrypt (approved):
Looks great. merge when ready

tests/engine/adhoc_test.go
nullpo-head: improved. Thanks.

PR #212, opened 2022-02-07
jit(arm64): support for const operations

mathetake:
part of #187

### DIFF
--- a/wasm/jit/jit_arm64.go
+++ b/wasm/jit/jit_arm64.go
@@ -136,7 +136,7 @@ func (c *arm64Compiler) returnFunction() {
 	c.exit(jitCallStatusCodeReturned)
 }
 
-// exit adds instruction to give the control back to engine.exec with the given status code.
+// exit adds instructions to give the control back to engine.exec with the given status code.
 func (c *arm64Compiler) exit(status jitCallStatusCode) {
 	tmp, _ := c.locationStack.takeFreeRegister(generalPurposeRegisterTypeInt)
 
--- a/wasm/jit/jit_arm64.go
+++ b/wasm/jit/jit_arm64.go
@@ -486,7 +486,7 @@ func (c *arm64Compiler) compileConstI64(o *wazeroir.OperationConstI64) error {
 	return c.emitIntConstant(false, o.Value)
 }
 
-// emitIntConstant adds instruction to load an integer constant.
+// emitIntConstant adds instructions to load an integer constant.
 // is32bit is true if the target value is originally 32-bit const, false otherwise.
 // value holds the (zero-extended for 32-bit case) load target constant.
 func (c *arm64Compiler) emitIntConstant(is32bit bool, value uint64) error {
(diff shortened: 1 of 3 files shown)


wasm/jit/jit_value_location_arm64.go
```diff
@@ -15,14 +15,18 @@ const (
...
-		arm64.REG_V18, arm64.REG_V19, arm64.REG_V20, arm64.REG_V21, arm64.REG_V22, arm64.REG_V23,
-		arm64.REG_V24, arm64.REG_V25, arm64.REG_V26, arm64.REG_V27, arm64.REG_V28, arm64.REG_V29,
-		arm64.REG_V30, arm64.REG_V31,
+		arm64.REG_F0, arm64.REG_F1, arm64.REG_F2, arm64.REG_F3,
+		arm64.REG_F4, arm64.REG_F5, arm64.REG_F6, arm64.REG_F7, arm64.REG_F8,
+		arm64.REG_F9, arm64.REG_F10, arm64.REG_F11, arm64.REG_F12, arm64.REG_F13,
+		arm64.REG_F14, arm64.REG_F15, arm64.REG_F16, arm64.REG_F17, arm64.REG_F18,
+		arm64.REG_F19, arm64.REG_F20, arm64.REG_F21, arm64.REG_F22, arm64.REG_F23,
+		arm64.REG_F24, arm64.REG_F25, arm64.REG_F26, arm64.REG_F27, arm64.REG_F28,
+		arm64.REG_F29, arm64.REG_F30, arm64.REG_F31,
```
mathetake, line 29: V* registers are for SIMD operations, and F* are for scaler floats :D

wasm/jit/jit_arm64.go
```diff
@@ -667,7 +667,20 @@ func (c *arm64Compiler) initializeReservedStackBasePointerRegister() {
...
+	setShiftedRegister(calcStackBasePointerAddress, true, tmpReg, 3)
 	c.addInstruction(calcStackBasePointerAddress)
 }
+
+// setShiftedRegister modifies the given *obj.Prog so that .From or .To
+// becomes the "shifted register". For example, this is used to emit instruction like
+// "add  x1, x2, x3, lsl #3" which means "x1 = x2 + (x3 << 3)".
+// See https://github.com/twitchyliquid64/golang-asm/blob/v0.15.1/obj/link.go#L120-L131.
+func setShiftedRegister(inst *obj.Prog, onFrom bool, register int16, shiftNum int64) {
+	if onFrom {
```
codefromthecrypt: maybe var addr Addr... if onFrom add = inst.From else addr = inst.To then addr.Type=... that way no copy/paste

mathetake: removed onFrom in the first place as I realized that setting this to .To is illegal and we never encounter the usecase 😄

mathetake:
ok ready!

wasm/jit/jit_arm64.go
```diff
@@ -451,18 +476,214 @@ func (c *arm64Compiler) compileMemorySize() error {
...
+	if value == 0 {
+		c.pushZeroValue()
+	} else {
+		// Take a register to load the value.
+		reg, err := c.allocateRegister(generalPurposeRegisterTypeInt)
+		if err != nil {
+			return err
+		}
+
+		// Then, load the const.
```
codefromthecrypt, line 502: this section down looks close to or the same for the float case. can this be extracted and reused if for no other reason re-using the comments around From.Offset?

codefromthecrypt (approved):
Nice to see this coming along

wasm/jit/jit_arm64.go
mathetake: cool I found a lot of duplication which I was too enough to consolidate (in amd64).. arm64 code would be much less code in this way!

mathetake: https://github.com/tetratelabs/wazero/pull/212/commits/c3268d5957349f848ba654b83bf7c5a34a046b82

mathetake:
Thanks for the help!!

PR #214, opened 2022-02-08
jit(arm64): support for Pick, Drop, Le 

mathetake:
pretty much in wip

### DIFF
--- a/wasm/jit/jit_arm64.go
+++ b/wasm/jit/jit_arm64.go
@@ -336,7 +336,38 @@ func (c *arm64Compiler) compileSelect() error {
 }
 
 func (c *arm64Compiler) compilePick(o *wazeroir.OperationPick) error {
-	return fmt.Errorf("TODO: unsupported on arm64")
+	pickTarget := c.locationStack.stack[c.locationStack.sp-1-uint64(o.Depth)]
+
+	pickedRegister, err := c.allocateRegister(pickTarget.registerType())
+	if err != nil {
+		return err
+	}
+
+	if pickTarget.onRegister() {
+		// If the pick target is on a register, we simply copy the value
+		// to the pickedRegister.
+		var inst obj.As
+		switch pickTarget.registerType() {
+		case generalPurposeRegisterTypeInt:
+			inst = arm64.AMOVD
+		case generalPurposeRegisterTypeFloat:
+			inst = arm64.AFMOVD
+		}
+		c.applyRegisterToRegisterInstruction(inst, pickTarget.register, pickedRegister)
+	} else if pickTarget.onStack() {
+		// Temporarily assign a register to the pick target, and then load the value.
+		pickTarget.setRegister(pickedRegister)
+		c.loadValueOnStackToRegister(pickTarget)
+		// After the load, we revert the register assignment to the pick target.
+		pickTarget.setRegister(nilRegister)
+	} else if pickTarget.onConditionalRegister() {
+		return fmt.Errorf("TODO: support conditional register target in pick")
+	}
+
+	// Now we have the value of the target on the pickedRegister,
+	// so push the location.
+	c.locationStack.pushValueOnRegister(pickedRegister)
+	return nil
 }
 
 // compileAdd implements compiler.compileAdd for the arm64 architecture.
--- a/wasm/jit/jit_arm64_test.go
+++ b/wasm/jit/jit_arm64_test.go
@@ -519,3 +519,106 @@ func TestArm64Compiler_compile_Add_Sub_Mul(t *testing.T) {
 		})
 	}
 }
+
+func TestArm64Compiler_compielePick(t *testing.T) {
+	const pickTargetValue uint64 = 12345
+	op := &wazeroir.OperationPick{Depth: 1}
+
+	for _, tc := range []struct {
+		name                                      string
+		pickTargetSetupFunc                       func(compiler *arm64Compiler, eng *engine) error
+		isPickTargetFloat, isPickTargetOnRegister bool
+	}{
+		{
+			name: "float on register",
+			pickTargetSetupFunc: func(compiler *arm64Compiler, eng *engine) error {
+				return compiler.compileConstF64(&wazeroir.OperationConstF64{Value: math.Float64frombits(pickTargetValue)})
+			},
+			isPickTargetFloat:      true,
+			isPickTargetOnRegister: true,
+		},
+		{
+			name: "int on register",
+			pickTargetSetupFunc: func(compiler *arm64Compiler, eng *engine) error {
+				return compiler.compileConstI64(&wazeroir.OperationConstI64{Value: pickTargetValue})
+			},
+			isPickTargetFloat:      false,
+			isPickTargetOnRegister: true,
+		},
+		{
+			name: "float on stack",
+			pickTargetSetupFunc: func(compiler *arm64Compiler, eng *engine) error {
+				pickTargetLocation := compiler.locationStack.pushValueOnStack()
+				pickTargetLocation.setRegisterType(generalPurposeRegisterTypeFloat)
+				eng.valueStack[pickTargetLocation.stackPointer] = pickTargetValue
+				return nil
+			},
+			isPickTargetFloat:      true,
+			isPickTargetOnRegister: false,
+		},
+		{
+			name: "int on stack",
+			pickTargetSetupFunc: func(compiler *arm64Compiler, eng *engine) error {
+				pickTargetLocation := compiler.locationStack.pushValueOnStack()
+				pickTargetLocation.setRegisterType(generalPurposeRegisterTypeInt)
+				eng.valueStack[pickTargetLocation.stackPointer] = pickTargetValue
+				return nil
+			},
+			isPickTargetFloat:      false,
+			isPickTargetOnRegister: false,
+		},
+	} {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			env := newJITEnvironment()
+			compiler := env.requireNewCompiler(t)
+			err := compiler.emitPreamble()
+			require.NoError(t, err)
+
+			// Set up the stack before picking.
+			err = tc.pickTargetSetupFunc(compiler, env.engine())
+			require.NoError(t, err)
+			pickTargetLocation := compiler.locationStack.peek()
+
+			// Push the unused median value.
+			_ = compiler.locationStack.pushValueOnStack()
+			require.Equal(t, uint64(2), compiler.locationStack.sp)
+
+			// Now ready to compile Pick operation.
+			err = compiler.compilePick(op)
+			require.NoError(t, err)
+			require.Equal(t, uint64(3), compiler.locationStack.sp)
+
+			pickedLocation := compiler.locationStack.peek()
+			require.True(t, pickedLocation.onRegister())
+			require.Equal(t, pickTargetLocation.registerType(), pickedLocation.registerType())
+
+			// Release the value to the memory stack again to verify the operation, and then return.
+			compiler.releaseRegisterToStack(pickedLocation)
+			if tc.isPickTargetOnRegister {
+				compiler.releaseRegisterToStack(pickTargetLocation)
+			}
+			compiler.returnFunction()
+
+			// Generate the code under test.
+			code, _, _, err := compiler.compile()
+			require.NoError(t, err)
+
+			// Run code.
+			env.exec(code)
+
+			// Check the returned status and stack pointer.
+			require.Equal(t, jitCallStatusCodeReturned, env.jitStatus())
+			require.Equal(t, uint64(3), env.stackPointer())
+
+			// Verify the top value is the picked one and the pick target's value stays the same.
+			if tc.isPickTargetFloat {
+				require.Equal(t, math.Float64frombits(pickTargetValue), env.stackTopAsFloat64())
+				require.Equal(t, math.Float64frombits(pickTargetValue), math.Float64frombits(env.stack()[pickTargetLocation.stackPointer]))
+			} else {
+				require.Equal(t, pickTargetValue, env.stackTopAsUint64())
+				require.Equal(t, pickTargetValue, env.stack()[pickTargetLocation.stackPointer])
+			}
+		})
+	}
+}


mathetake:
lol arm64 is simple enough that I could finish all of lt, le, gt, ge operations all at once 😄

mathetake:
This has been straightforward after amd64, but I re-wrote the tests 👯  PTAL!

wasm/jit/jit_arm64.go
```diff
@@ -200,16 +204,41 @@ func (c *arm64Compiler) applyTwoRegistersToRegisterInstruction(instruction obj.A
 	c.addInstruction(inst)
 }
 
+// applyTwoRegistersToNoneInstruction adds an instruction which takes two source operands on registers.
+func (c *arm64Compiler) applyTwoRegistersToNoneInstruction(instruction obj.As, src1, src2 int16) {
+	inst := c.newProg()
+	inst.As = instruction
+	inst.To.Type = obj.TYPE_NONE
```
codefromthecrypt, line 211: what's the impact of To.Type == None?

wasm/jit/jit_arm64.go
```diff
@@ -200,16 +204,41 @@ func (c *arm64Compiler) applyTwoRegistersToRegisterInstruction(instruction obj.A
...
+	inst.From.Type = obj.TYPE_REG
+	inst.From.Reg = src1
+	inst.Reg = src2
+	c.addInstruction(inst)
+}
+
 func (c *arm64Compiler) String() (ret string) { return }
 
+func (c *arm64Compiler) pushFunctionParams() {
+	if c.f != nil && c.f.FunctionType != nil {
```
codefromthecrypt, line 221: nit I prefer early return on things like this as reduces the indent level

```suggestion
	if c.f == nil || c.f.FunctionType == nil {
		return
	}
```

wasm/jit/jit_arm64.go
```diff
@@ -200,16 +204,41 @@ func (c *arm64Compiler) applyTwoRegistersToRegisterInstruction(instruction obj.A
...
+	inst.To.Type = obj.TYPE_NONE
+	inst.From.Type = obj.TYPE_REG
+	inst.From.Reg = src1
+	inst.Reg = src2
+	c.addInstruction(inst)
+}
+
 func (c *arm64Compiler) String() (ret string) { return }
 
+func (c *arm64Compiler) pushFunctionParams() {
```
codefromthecrypt, line 220:
```suggestion
// pushFunctionParams pushes any function parameters onto the stack, setting appropriate register types.
func (c *arm64Compiler) pushFunctionParams() {
```

wasm/jit/jit_arm64.go
```diff
@@ -327,16 +356,87 @@ func (c *arm64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
...
 }
 
+// compileDrop implements compiler.compileDrop for the arm64 architecture.
 func (c *arm64Compiler) compileDrop(o *wazeroir.OperationDrop) error {
-	return fmt.Errorf("TODO: unsupported on arm64")
+	r := o.Range
+	if r == nil {
+		return nil
+	} else if r.Start == 0 {
+		// If the drop starts from the top, we just mark all the regiters
```
codefromthecrypt, line 365: nit: "just " is almost always deletable

codefromthecrypt: reword, and insert the why for the what
"When the drop starts from the top the stack, mark all registers unused because XXX"

wasm/jit/jit_arm64.go
```diff
@@ -327,16 +356,87 @@ func (c *arm64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
...
+		// used by drop targets unused.
+		for i := 0; i <= r.End; i++ {
+			if loc := c.locationStack.pop(); loc.onRegister() {
+				c.markRegisterUnused(loc.register)
+			}
+		}
+		return nil
+	}
+
+	// Save the non-drop target values.
```
codefromthecrypt, line 375: add the why. ex. "so that XXX". also if "non-drop" == "live" then choose one or the other term, not both

wasm/jit/jit_arm64.go
```diff
@@ -327,16 +356,87 @@ func (c *arm64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
...
+	liveValues := make([]*valueLocation, r.Start)
+	for i := 0; i < r.Start; i++ {
+		live := c.locationStack.pop()
+		liveValues[i] = live
+	}
+
+	// Then mark all registers used by drop tragets unused.
```
codefromthecrypt, line 382: s/tragets/targets/g and mention why `r.End-r.Start+1` or extract that into a local which name says why

wasm/jit/jit_arm64.go
```diff
@@ -327,16 +356,87 @@ func (c *arm64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
...
+	for i := 0; i < r.End-r.Start+1; i++ {
+		if loc := c.locationStack.pop(); loc.onRegister() {
+			c.markRegisterUnused(loc.register)
+		}
+	}
+
+	for i := range liveValues {
+		live := liveValues[len(liveValues)-1-i]
+		// If the value is on a memory, we have to move it to a register,
```
codefromthecrypt, line 391: good comment! what and why

wasm/jit/jit_arm64.go
```diff
@@ -327,16 +356,87 @@ func (c *arm64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
...
+
+	// Then mark all registers used by drop tragets unused.
+	for i := 0; i < r.End-r.Start+1; i++ {
+		if loc := c.locationStack.pop(); loc.onRegister() {
+			c.markRegisterUnused(loc.register)
+		}
+	}
+
+	for i := range liveValues {
+		live := liveValues[len(liveValues)-1-i]
```
mathetake, line 390: I think this needs comment

wasm/jit/jit_arm64.go
```diff
@@ -327,16 +356,87 @@ func (c *arm64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
...
+		// If the value is on a memory, we have to move it to a register,
+		// otherwise the memory location is overriden by other values
+		// after this drop instructin.
+		if err := c.ensureOnGeneralPurposeRegister(live); err != nil {
+			return err
+		}
+		// Modify the stack location by pushing onto the location stack.
```
codefromthecrypt, line 397: It might be confusing that we are implying above "ensure it is on a GP register" and implying "it is on the stack" here :D This may be less confusing to people working on this all the time, though.

nit: modify is correct, but I think update is more telling. 
```suggestion
		// Update the stack location by pushing onto the location stack.
```

wasm/jit/jit_arm64.go
```diff
@@ -327,16 +356,87 @@ func (c *arm64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
...
+	c.maybeMoveTopConditionalToFreeGeneralPurposeRegister()
+
+	pickTarget := c.locationStack.stack[c.locationStack.sp-1-uint64(o.Depth)]
+	pickedRegister, err := c.allocateRegister(pickTarget.registerType())
+	if err != nil {
+		return err
+	}
+
+	if pickTarget.onRegister() {
+		// If the pick target is on a register, we simply copy the value
```
codefromthecrypt, line 418: since "if " already says pickTarget.onRegister. sometimes I use a trialing comment to save lines and doc drift.

ex. 

```
if pickTarget.onRegister() { // copy the value to the picked register
```

wasm/jit/jit_arm64.go
```diff
@@ -327,16 +356,87 @@ func (c *arm64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
...
+		c.applyRegisterToRegisterInstruction(inst, pickTarget.register, pickedRegister)
+	} else if pickTarget.onStack() {
+		// Temporarily assign a register to the pick target, and then load the value.
+		pickTarget.setRegister(pickedRegister)
+		c.loadValueOnStackToRegister(pickTarget)
+		// After the load, we revert the register assignment to the pick target.
+		pickTarget.setRegister(nilRegister)
+	}
+
+	// Now we have the value of the target on the pickedRegister,
```
codefromthecrypt, line 436: this comment doesn't match the code. maybe the function name isn't matching.

ex. `locationStack.pushValueOnRegister(pickedRegister)` reads "push the value on the location stack to the picked register"

perhaps I misunderstood, but think if the function name could be more accurate if not.

wasm/jit/jit_arm64.go
```diff
@@ -581,20 +681,152 @@ func (c *arm64Compiler) compileEqz(o *wazeroir.OperationEqz) error {
...
+		inst = arm64.AFCMPS
+		conditionalRegister = arm64.COND_MI
+	case wazeroir.SignedTypeFloat64:
+		inst = arm64.AFCMPD
+		conditionalRegister = arm64.COND_MI
+	}
+
+	c.applyTwoRegistersToNoneInstruction(inst, x2.register, x1.register)
+
+	// Push the comparison result as a conditional register placed value.
```
codefromthecrypt, line 716: "conditional register placed value" is a mouthful and makes me wonder if the word "placed" is meaningful. If it isn't, delete it, or perhaps move the significance of "placed" terminology to doc on `pushValueOnConditionalRegister` and leave it out here. similar comment for all "placed" later in the file

wasm/jit/jit_arm64.go
```diff
@@ -745,11 +981,42 @@ func (c *arm64Compiler) ensureOnGeneralPurposeRegister(loc *valueLocation) (err
...
 		err = c.loadValueOnStackToRegister(loc)
 	} else if loc.onConditionalRegister() {
-		err = fmt.Errorf("TODO: support moving conditional to general purpose register")
+		c.loadConditionalRegisterToGeneralPurposeRegister(loc)
 	}
 	return
 }
 
+// maybeMoveTopConditionalToFreeGeneralPurposeRegister moves the top value on the stack
+// if the value is located on a conditional register. This is usually called at the beginning of
```
codefromthecrypt, line 990: add 2 line breaks before "This"

```
...conditional register.
//
// This

File: wasm/jit/jit_arm64.go [hunk 14]
```diff
@@ -745,11 +981,42 @@ func (c *arm64Compiler) ensureOnGeneralPurposeRegister(loc *valueLocation) (err
...
+			c.loadConditionalRegisterToGeneralPurposeRegister(loc)
+		}
+	}
+}
+
+func (c *arm64Compiler) loadConditionalRegisterToGeneralPurposeRegister(loc *valueLocation) {
+	reg, _ := c.locationStack.takeFreeRegister(generalPurposeRegisterTypeInt)
+	c.markRegisterUsed(reg)
+
+	// Use CSET instruction to set 1 on the register if the condition satisfies.
```
codefromthecrypt, line 1006: move this to a Note: on the function doc

wasm/jit/jit_arm64.go
```diff
@@ -808,6 +1075,24 @@ func (c *arm64Compiler) allocateRegister(t generalPurposeRegisterType) (reg int1
 	return
 }
 
+// releaseAllRegistersToStack adds instructions to store all the values located on either general purpuse or conditional
```
codefromthecrypt, line 1072: add  // See releaseRegisterToStack

wasm/jit/jit_arm64_test.go
```diff
@@ -238,6 +238,8 @@ func TestArm64Compiler_releaseRegisterToStack(t *testing.T) {
 			// Setup the location stack so that we push the const on the specified height.
 			compiler.locationStack.sp = tc.stackPointer
 			compiler.locationStack.stack = make([]*valueLocation, tc.stackPointer)
+			// Peek must be non-nil. Otherwise, compileConst* would fail.
```
codefromthecrypt, line 241: good comment

wasm/jit/jit_arm64_test.go
```diff
@@ -351,6 +353,193 @@ func TestArm64Compiler_loadValueOnStackToRegister(t *testing.T) {
...
+
+							// Release the value to the memory stack again to verify the operation, and then return.
+							compiler.releaseRegisterToStack(resultLocation)
+							compiler.returnFunction()
+
+							// Generate the code under test.
+							code, _, _, err := compiler.compile()
+							require.NoError(t, err)
+
+							// Run code.
```
codefromthecrypt, line 459: nit but this is probably not that telling. maybe make the former comment

```
// Compile and execute the code under test.
```

and squash the newlines and this comment between this and compile()

same here and everywhere

wasm/jit/jit_arm64_test.go
```diff
@@ -351,6 +353,193 @@ func TestArm64Compiler_loadValueOnStackToRegister(t *testing.T) {
...
+							env.exec(code)
+
+							// Check the stack.
```
codefromthecrypt, line 462: be more specific, as the assertion on stackPointer says we are checking it somehow

```suggestion
							// There should only be one value on the stack
```

wasm/jit/jit_arm64_test.go
```diff
@@ -519,3 +708,241 @@ func TestArm64Compiler_compile_Add_Sub_Mul(t *testing.T) {
...
+			// Now ready to compile Pick operation.
+			err = compiler.compilePick(op)
+			require.NoError(t, err)
+			require.Equal(t, uint64(3), compiler.locationStack.sp)
+
+			pickedLocation := compiler.locationStack.peek()
+			require.True(t, pickedLocation.onRegister())
+			require.Equal(t, pickTargetLocation.registerType(), pickedLocation.registerType())
+
+			// Release the value to the memory stack again to verify the operation, and then return.
```
codefromthecrypt, line 785: what happens if we don't return?

maybe delete the "and then return" here and put a comment above compiler.returnFunction of why this is needed (ex does this end the program and it would abend if we didn't?

codefromthecrypt (approved):
Looks good. I got lazy lately, so went for a more thorough review on comments this time. Feel free to choose which to address now, later or never

wasm/jit/jit_arm64.go
mathetake: So the reason why this is confusing is that we have two stacks -- one is valueLocationStack which we use at compilation time to track all the wasm values' location (memory stack, registers, or conditional register). Another is the stack in a memory at runtime. I must admit this is really confusing..

mathetake: https://github.com/tetratelabs/wazero/pull/214/commits/435623b84683600b96bdd9a13a6dde545bc6f9f9 I did my best only on this line, but I have to revisit all the comments later

wasm/jit/jit_arm64.go
mathetake: location stack is for compilation, and this means that "pushes a value location with a register (pickedRegister)".

mathetake: let me rethink and at least document the value location stack methods

mathetake:
Thanks for the look!!!!!!

codefromthecrypt:
> So the reason why this is confusing is that we have two stacks -- one is valueLocationStack which we use at compilation time to track all the wasm values' location (memory stack, registers, or conditional register). Another is the stack in a memory at runtime. I must admit this is really confusing..

ack. this is tricky

PR #215, opened 2022-02-08
text: adds expansion of abbreviated import and exports

codefromthecrypt:
needs to be integrated with the main decoder still

### DIFF
--- a/wasm/text/abbreviation_parser.go
+++ b/wasm/text/abbreviation_parser.go
@@ -1,0 +1,263 @@
+package text
+
+import (
+	"errors"
+	"fmt"
+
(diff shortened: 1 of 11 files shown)


codefromthecrypt:
still a lot of work left to do on this, mostly looking very very carefully, backfilling tests, organizing code and finishing memory parsing.

codefromthecrypt:
to buy back some room to land this, I'll firstly change how error positions are handled. Switching from logical position (ex. module.func[3].param[2]) to source position (ex display the line than erred and a ^^ to the token that failed). This will allow us to reuse complexity currently only needed for errors to be instead used for parser state.

PR #216, opened 2022-02-09
wasi: Implement environ_get and environ_sizes_get API

nullpo-head:
This PR implements `environ_get` and `environ_sizes_get`.

### DIFF
--- a/wasi/strings.go
+++ b/wasi/strings.go
@@ -19,20 +19,20 @@ type nullTerminatedStrings struct {
 
 // newNullTerminatedStrings creates a nullTerminatedStrings from the given string slice. It returns an error
 // if the length or the total buffer size of the result nullTerminatedStrings exceeds the maxBufSize
-func newNullTerminatedStrings(maxBufSize uint32, args ...string) (*nullTerminatedStrings, error) {
-	if len(args) == 0 {
+func newNullTerminatedStrings(maxBufSize uint32, valName string, vals ...string) (*nullTerminatedStrings, error) {
+	if len(vals) == 0 {
 		return &nullTerminatedStrings{nullTerminatedValues: [][]byte{}}, nil
 	}
 	var strings [][]byte // don't pre-allocate as this function is size bound
 	totalBufSize := uint32(0)
-	for i, arg := range args {
+	for i, arg := range vals {
 		if !utf8.ValidString(arg) {
-			return nil, fmt.Errorf("value at %d is not a valid UTF-8 string", i)
+			return nil, fmt.Errorf("%s[%d] is not a valid UTF-8 string", valName, i)
 		}
 		argLen := uint64(len(arg)) + 1 // + 1 for "\x00"; uint64 in case this one arg is huge
 		nextSize := uint64(totalBufSize) + argLen
 		if nextSize > uint64(maxBufSize) {
-			return nil, fmt.Errorf("value at %d will exceed max buffer size %d", i, maxBufSize)
+			return nil, fmt.Errorf("%s[%d] will exceed max buffer size %d", valName, i, maxBufSize)
 		}
 		totalBufSize = uint32(nextSize)
 		strings = append(strings, append([]byte(arg), 0))
--- a/wasi/strings.go
+++ b/wasi/strings.go
@@ -18,7 +18,7 @@ type nullTerminatedStrings struct {
 }
 
 // newNullTerminatedStrings creates a nullTerminatedStrings from the given string slice. It returns an error
-// if the length or the total buffer size of the result WASIStringArray exceeds the maxBufSize
+// if the length or the total buffer size of the result nullTerminatedStrings exceeds the maxBufSize
 func newNullTerminatedStrings(maxBufSize uint32, args ...string) (*nullTerminatedStrings, error) {
 	if len(args) == 0 {
 		return &nullTerminatedStrings{nullTerminatedValues: [][]byte{}}, nil
@@ -27,12 +27,12 @@ func newNullTerminatedStrings(maxBufSize uint32, args ...string) (*nullTerminate
 	totalBufSize := uint32(0)
 	for i, arg := range args {
 		if !utf8.ValidString(arg) {
-			return nil, fmt.Errorf("arg[%d] is not a valid UTF-8 string", i)
+			return nil, fmt.Errorf("value at %d is not a valid UTF-8 string", i)
 		}
 		argLen := uint64(len(arg)) + 1 // + 1 for "\x00"; uint64 in case this one arg is huge
 		nextSize := uint64(totalBufSize) + argLen
 		if nextSize > uint64(maxBufSize) {
-			return nil, fmt.Errorf("arg[%d] will exceed max buffer size %d", i, maxBufSize)
+			return nil, fmt.Errorf("value at %d will exceed max buffer size %d", i, maxBufSize)
 		}
 		totalBufSize = uint32(nextSize)
 		strings = append(strings, append([]byte(arg), 0))
--- a/wasi/strings_test.go
+++ b/wasi/strings_test.go
@@ -65,7 +65,7 @@ func TestNewNullTerminatedStrings(t *testing.T) {
 		tc := tt
 
 		t.Run(tc.name, func(t *testing.T) {
-			s, err := newNullTerminatedStrings(100, tc.input...)
+			s, err := newNullTerminatedStrings(100, "", tc.input...)
 			require.NoError(t, err)
 			require.Equal(t, tc.expected, s)
 		})
@@ -74,15 +74,15 @@ func TestNewNullTerminatedStrings(t *testing.T) {
 
 func TestNewNullTerminatedStrings_Errors(t *testing.T) {
 	t.Run("invalid utf-8", func(t *testing.T) {
-		_, err := newNullTerminatedStrings(100, "\xff\xfe\xfd", "foo", "bar")
-		require.EqualError(t, err, "value at 0 is not a valid UTF-8 string")
+		_, err := newNullTerminatedStrings(100, "arg", "\xff\xfe\xfd", "foo", "bar")
+		require.EqualError(t, err, "arg[0] is not a valid UTF-8 string")
 	})
-	t.Run("value at 0 too large", func(t *testing.T) {
-		_, err := newNullTerminatedStrings(1, "a", "bc")
-		require.EqualError(t, err, "value at 0 will exceed max buffer size 1")
+	t.Run("arg[0] too large", func(t *testing.T) {
+		_, err := newNullTerminatedStrings(1, "arg", "a", "bc")
+		require.EqualError(t, err, "arg[0] will exceed max buffer size 1")
 	})
 	t.Run("empty arg too large due to null terminator", func(t *testing.T) {
-		_, err := newNullTerminatedStrings(2, "a", "", "bc")
-		require.EqualError(t, err, "value at 1 will exceed max buffer size 2")
+		_, err := newNullTerminatedStrings(2, "arg", "a", "", "bc")
+		require.EqualError(t, err, "arg[1] will exceed max buffer size 2")
 	})
 }
--- a/wasi/strings_test.go
+++ b/wasi/strings_test.go
@@ -75,14 +75,14 @@ func TestNewNullTerminatedStrings(t *testing.T) {
 func TestNewNullTerminatedStrings_Errors(t *testing.T) {
 	t.Run("invalid utf-8", func(t *testing.T) {
 		_, err := newNullTerminatedStrings(100, "\xff\xfe\xfd", "foo", "bar")
-		require.EqualError(t, err, "arg[0] is not a valid UTF-8 string")
+		require.EqualError(t, err, "value at 0 is not a valid UTF-8 string")
 	})
-	t.Run("arg[0] too large", func(t *testing.T) {
+	t.Run("value at 0 too large", func(t *testing.T) {
 		_, err := newNullTerminatedStrings(1, "a", "bc")
-		require.EqualError(t, err, "arg[0] will exceed max buffer size 1")
+		require.EqualError(t, err, "value at 0 will exceed max buffer size 1")
 	})
 	t.Run("empty arg too large due to null terminator", func(t *testing.T) {
 		_, err := newNullTerminatedStrings(2, "a", "", "bc")
-		require.EqualError(t, err, "arg[1] will exceed max buffer size 2")
+		require.EqualError(t, err, "value at 1 will exceed max buffer size 2")
 	})
 }
--- a/wasi/wasi.go
+++ b/wasi/wasi.go
@@ -86,7 +86,7 @@ type API interface {
 
 	// EnvironGet is the WASI function that reads environment variables.
 	//
-	// There are two parameters. Both are offsets in api.HostFunctionCallContext Memory. If either are invalid due to
+	// There are two parameters. Both are offsets in wasm.HostFunctionCallContext Memory. If either are invalid due to
 	// memory constraints, this returns ErrnoFault.
 	//
 	// * environ - is the offset to begin writing environment variables offsets in uint32 little-endian encoding.
@@ -105,15 +105,15 @@ type API interface {
 	//                              environ offset for "a=b" --+           |
 	//                                         environ offset for "b=cd" --+
 	//
-	// Note: FunctionEnvironGet documentation shows this signature in the WebAssembly 1.0 (MVP) Text Format.
+	// Note: ImportEnvironGet shows this signature in the WebAssembly 1.0 (MVP) Text Format.
 	// See EnvironSizesGet
 	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#environ_get
 	// See https://en.wikipedia.org/wiki/Null-terminated_string
-	EnvironGet(ctx wasm.HostFunctionCallContext, resultEnviron, resultEnvironBuf uint32) Errno
+	EnvironGet(ctx wasm.HostFunctionCallContext, environ, environBuf uint32) Errno
 
 	// EnvironSizesGet is a WASI function that reads environment variable (Environ) sizes.
 	//
-	// There are two result parameters: these are offsets in the api.HostFunctionCallContext Memory to write
+	// There are two result parameters: these are offsets in the wasi.HostFunctionCallContext Memory to write
 	// corresponding sizes in uint32 little-endian encoding. If either are invalid due to memory constraints, this
 	// returns ErrnoFault.
 	//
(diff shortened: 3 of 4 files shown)


tests/wasi/wasi_test.go
```diff
@@ -78,3 +78,61 @@ func Test_ArgsSizesGet_ArgsGet(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			store := wasm.NewStore(interpreter.NewEngine())
+			require.NoError(t, err)
+
+			stdoutBuf := bytes.NewBuffer(nil)
+			envOpt, err := wasi.Environ(tc.keys, tc.values)
+			require.NoError(t, err)
+			wasiEnv := wasi.NewEnvironment(envOpt, wasi.Stdout(stdoutBuf))
+
+			err = wasiEnv.Register(store)
```
codefromthecrypt, line 125: my guess is some of this init logic is reused across wasi tests and can be a separate function (still in test source)?

tests/wasi/wasi_test.go
```diff
@@ -78,3 +78,61 @@ func Test_ArgsSizesGet_ArgsGet(t *testing.T) {
...
 	}
 }
+
+//go:embed testdata/environ.wasm
+var environWasm []byte
+
+// Test environ_sizes_get and environ_get. environ_sizes_get must be used to know the length and the
+// size of the result of environ_get, so TinyGo calls the both of them together to retrieve the
+// WASI's environment variables. Any other language runtime would do the same things.
+func Test_environ_sizes_get_environ_get(t *testing.T) {
```
codefromthecrypt, line 88: too bad we can't have a separate func per wasi here as this is really hard to read vs `Test_environ_sizes_get` and `Test_environ_get`. Any way we can do that, as if we can it is also easier to count which things are done by virtue of the test results

wasi/wasi.go
```diff
@@ -170,6 +173,30 @@ func Args(args []string) (Option, error) {
 	}, nil
 }
 
+// Environ returns an option to set environmental variables to the WASIEnvironment.
+// * keys: the names of environment variables to set
+// * values: the corresponding values of environment variables to set
+// Environ returns an error if the length of the keys and values doesn't match,
+// or the length or the total size of those strings exceed the max of uint32.
+func Environ(keys []string, values []string) (Option, error) {
```
codefromthecrypt, line 181: `os.Environ` is []string, should we match that up for convenience (considering this is a go project). It also seems to match the representation. Moreover a map is a more idiomatic way for key/value if we were going for that. I think we should re-use how `os.Environ` does it.

Then, we can say `Note: while this signature matches os.Environ conventions, env propagation into WASI is intentionally done manually here`

wdyt?

wasi/wasi.go
```diff
@@ -396,13 +424,48 @@ func proc_exit(*wasm.HostFunctionCallContext, uint32) {
...
+
+	return ESUCCESS
 }
 
-func environ_get(*wasm.HostFunctionCallContext, uint32, uint32) (err Errno) {
-	return
+// environ_get is a WASI API to read the environment variable data.
+// * environPtr:
+//     A pointer to a buffer. environ_get writes multiple *C.char pointers in sequence there. In other words, this is an array of *char in C.
+//     The pointers point to environment variables. Each environment variable is a key/value pair joined with `=`, and it is a null-terminated string.
```
codefromthecrypt, line 447: if we match signatures with how we set this, we should be able to say "See Environ for notes on syntax`

wasi/wasi.go
```diff
@@ -396,13 +424,48 @@ func proc_exit(*wasm.HostFunctionCallContext, uint32) {
...
+//     The number of the environment variables matches the value that environ_sizes_get returns.
+// * environBufPtr:
+//     A pointer to a buffer. envrion_get writes the environment variables as null-terminated strings in the given buffer.
+//     The total number of bytes written there is the value that args_sizes_get returns in argsBufSizePtr. The caller must ensure that
+//     the buffer has the enough size.
+//     Each *C.char pointer that can be obtained from environPtr points to the beginning of each of these null-terminated strings.
```
codefromthecrypt, line 453: do we think `*C.char pointer` will be understandable and relevant in go code, given choices of how to describe this?

codefromthecrypt (commented):
I think this is good work. Main things will be to try to make things less cryptic. Doing that would be shedding some "C brain" a bit and deferring to go idioms since this is a go project. Another way is to link to other code when in doubt as that helps docs be concise. Ex. instead of re-explaining key/value encoding link to something in go that already does this.

When you feel ready, I'll probably come back with more suggestions of how to reduce word count on some of these docs by half while keeping the signal high and relevant. By doing this sooner, we can make sure we have a good "concision baseline" to apply to other functions missing docs.

wasi/wasi.go
mathetake, line 181: is it valid to have duplicated keys?

mathetake: This make me wonder whether it would be better to have `map[key]value` as inputs as I imagine users usually hold environments as maps rather than index aligned two slices like this

mathetake (commented):
One quick question

wasi/wasi.go
mathetake: oh maybe this is dup from @codefromthecrypt's review :D

codefromthecrypt: I also thought about dupes, but I doubt the spec will actually enforce this. Also, pros and cons on this are easier if we defer to os.Environ (ex this was go's fault not ours ;P)

codefromthecrypt: also if we do map insertion order is lost which may be fine, but also complicates. dupe could be checked internally if we wanted to

nullpo-head: Yeah, I agree with the format of `os.Environ` is the best choice for ergonomics, considering that the common use cases are calling `os.Environ` to get the all environment variables, and pass it to this `wasi.Environ` (possibly with some filtering). I'd like to merge the design discussion on vs map interface here. wdyt @mathetake?

nullpo-head: BTW, why my first design was the same as that of wasmtime, but now I think the format of `os.Environ` is better than that for the reason above.

nullpo-head: I replied about the part of the discussion on `map[key]value` vs `os.Environ` in another thread.

nullpo-head: As for the question whether a duplicated environment variable is allowed or not, I doubt WASI has any rule on duplication right now. So, there may not be any guarantee from the perspective of WASI spec that the result of args_get is unique, though it will be also ok for wazero to accept only unique environment variables, mentioning that in the rationale note. Anyway, we don't have to think about this behavior by accepting the same format as `os.Environ`.

side note: incidentally, operating systems such as Linux don't guarantee the uniqueness of the environment variables, though WASI may not follow that weird spec.

codefromthecrypt: PS if anyone ever finds any value validation of anything in wasi, please point it out. So far, you are lucky if a name is defined for a parameter.

nullpo-head:
Hi @mathetake @codefromthecrypt, I've updated this PR based on the latest code base. PTAL. Thanks!

wasi/wasi.go
```diff
@@ -82,9 +84,60 @@ type API interface {
...
 
-	// TODO: EnvironGet(ctx api.hostFunctionCallContext, environ, environBuf uint32) Errno
+	// EnvironGet is the WASI function that reads environment variables.
+	//
+	// There are two parameters. Both are offsets in api.HostFunctionCallContext Memory. If either are invalid due to
+	// memory constraints, this returns ErrnoFault.
+	//
+	// * environ - is the offset to begin writing environment variables offsets in uint32 little-endian encoding.
+	//   * EnvironSizesGet result environc * 4 bytes are written to this offset
+	// * environBuf - is the offset to write the null terminated environment variables in the form of "key=val", the same format as os.Environ(), to wasm.MemoryInstance.
```
mathetake, line 94: I think we should avoid mentions to `wasm.MemoryInstance` as it's hidden behind the interface

wasi/wasi.go
```diff
@@ -82,9 +84,60 @@ type API interface {
 	// See https://en.wikipedia.org/wiki/Null-terminated_string
 	ArgsSizesGet(ctx wasm.HostFunctionCallContext, resultArgc, resultArgvBufSize uint32) Errno
 
-	// TODO: EnvironGet(ctx api.hostFunctionCallContext, environ, environBuf uint32) Errno
+	// EnvironGet is the WASI function that reads environment variables.
+	//
+	// There are two parameters. Both are offsets in api.HostFunctionCallContext Memory. If either are invalid due to
```
mathetake, line 89: I think this is a drift from @codefromthecrypt as `api` doesn't exist :D

wasi/strings.go
```diff
@@ -27,12 +27,12 @@ func newNullTerminatedStrings(maxBufSize uint32, args ...string) (*nullTerminate
 	totalBufSize := uint32(0)
 	for i, arg := range args {
 		if !utf8.ValidString(arg) {
-			return nil, fmt.Errorf("arg[%d] is not a valid UTF-8 string", i)
+			return nil, fmt.Errorf("value at %d is not a valid UTF-8 string", i)
```
codefromthecrypt, line 30: I see what you are doing, but I think it is better to change the message instead.

ex. `(maxBufSize uint32, type string, vals ...string)`

then `%s[%d]..., type, i`

when you call say "arg" or "env"

remember this is internal code!

wasi/strings_test.go
```diff
@@ -75,14 +75,14 @@ func TestNewNullTerminatedStrings(t *testing.T) {
 func TestNewNullTerminatedStrings_Errors(t *testing.T) {
 	t.Run("invalid utf-8", func(t *testing.T) {
 		_, err := newNullTerminatedStrings(100, "\xff\xfe\xfd", "foo", "bar")
-		require.EqualError(t, err, "arg[0] is not a valid UTF-8 string")
```
codefromthecrypt, line 78: same here, revert and set "type" to "arg" instead

wasi/wasi.go
```diff
@@ -38,7 +40,7 @@ type API interface {
 	//   * ArgsSizesGet result argv_buf_size bytes are written to this offset
 	//
 	// For example, if ArgsSizesGet wrote argc=2 and argvBufSize=5 for arguments: "a" and "bc"
-	//   and ArgsGet results argv=7 and argvBuf=1, we expect `ctx.Memory.Buffer` to contain:
+	//   and ArgsGet parameters are argv=7 and argvBuf=1, we expect `ctx.Memory.Buffer` to contain:
```
codefromthecrypt, line 43: spec is so crap that sometimes out params (offsets to write data) are classified as params and other times results. this is ok how you've interpreted it!

wasi/wasi.go
```diff
@@ -65,7 +67,7 @@ type API interface {
 	// * resultArgvBufSize - is the offset to write the null-terminated argument length to wasm.MemoryInstance Buffer
 	//
 	// For example, if Args are []string{"a","bc"} and
-	//   ArgsSizesGet parameters resultArgc=1 and resultArgvBufSize=6, we expect `ctx.Memory.Buffer` to contain:
+	//   ArgsSizesGet parameters are resultArgc=1 and resultArgvBufSize=6, we expect `ctx.Memory.Buffer` to contain:
```
codefromthecrypt, line 70: unnecessary if you read from the top, but ok

wasi/wasi.go
```diff
@@ -82,9 +84,60 @@ type API interface {
...
+	//
+	//                           environBufSize                 uint32le    uint32le
+	//              +------------------------------------+     +--------+  +--------+
+	//              |                                    |     |        |  |        |
+	//   []byte{?, 'a', '=', 'b', 0, 'b', '=', 'c', 'd', 0, ?, 1, 0, 0, 0, 5, 0, 0, 0, ?}
+	// environBuf --^                                          ^           ^
+	//                              environ offset for "a=b" --+           |
+	//                                         environ offset for "b=cd" --+
+	//
+	// Note: FunctionEnvironGet documentation shows this signature in the WebAssembly 1.0 (MVP) Text Format.
```
codefromthecrypt, line 108:
```suggestion
	// Note: ImportEnvironGet shows this signature in the WebAssembly 1.0 (MVP) Text Format.
```

wasi/wasi.go
codefromthecrypt, line 89: I have the other drift in another PR.

```suggestion
	// There are two parameters. Both are offsets in wasm.HostFunctionCallContext Memory. If either are invalid due to
```

wasi/wasi.go
```diff
@@ -82,9 +84,60 @@ type API interface {
...
+	// See EnvironSizesGet
+	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#environ_get
+	// See https://en.wikipedia.org/wiki/Null-terminated_string
+	EnvironGet(ctx wasm.HostFunctionCallContext, resultEnviron, resultEnvironBuf uint32) Errno
```
codefromthecrypt, line 112: ack this is confusing particularly as you noted above about `ArgsGet`. in the docs, https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#environ_get doesn't classify these as results, rather params, so we should stick with that I think. that or switch `ArgsGet` params to be prefixed result. etc


However, since you have other things saying 'environ' and that's what doc says, might as well converge to that
```suggestion
	EnvironGet(ctx wasm.HostFunctionCallContext, environ, environBuf uint32) Errno
```

wasi/wasi.go
nullpo-head: That makes sense. `ctx.Memory.Buffer` will need the update as well. So, let's change them to `wasm.Memory`, the interface name.

wasi/wasi.go
```diff
@@ -82,9 +84,60 @@ type API interface {
...
 
-	// TODO: EnvironSizesGet(ctx api.hostFunctionCallContext, resulEnvironc, resultEnvironBufSize uint32) Errno
+	// EnvironSizesGet is a WASI function that reads environment variable (Environ) sizes.
+	//
+	// There are two result parameters: these are offsets in the api.HostFunctionCallContext Memory to write
```
codefromthecrypt, line 116:
```suggestion
	// There are two result parameters: these are offsets in the wasi.HostFunctionCallContext Memory to write
```

wasi/wasi.go
```diff
@@ -82,9 +84,60 @@ type API interface {
...
+	// * resultEnvironBufSize - is the offset to write the null-terminated environment variable length to wasm.MemoryInstance Buffer
+	//
+	// For example, if Environ is []string{"a=b","b=cd"} and
+	//   EnvironSizesGet parameters are resultEnvironc=1 and resultEncironBufSize=6, we expect `ctx.Memory.Buffer` to contain:
+	//
+	//                   uint32le       uint32le
+	//                  +--------+     +--------+
+	//                  |        |     |        |
+	//        []byte{?, 2, 0, 0, 0, ?, 9, 0, 0, 0, ?}
+	// resultEnvironc --^              ^
```
codefromthecrypt, line 130: cool as the docs say these are results

wasi/wasi.go
```diff
@@ -174,8 +227,11 @@ const (
 )
 
 type wasiAPI struct {
-	args  *nullTerminatedStrings
-	stdin io.Reader
+	args *nullTerminatedStrings
+	// environ stores each environment variable in the form of "key=value",
+	// which is convenient for the implementation of environ_get.
```
codefromthecrypt, line 232:
```suggestion
	// which is both convenient for the implementation of environ_get and matches os.Environ
```

wasi/wasi.go
```diff
@@ -278,9 +334,34 @@ func (a *wasiAPI) ArgsSizesGet(ctx wasm.HostFunctionCallContext, resultArgc, res
 	return ErrnoSuccess
 }
 
-// TODO: func (a *wasiAPI) EnvironGet
+// EnvironGet implements API.EnvironGet
+func (w *wasiAPI) EnvironGet(ctx wasm.HostFunctionCallContext, resultEnviron uint32, resultEnvironBuf uint32) (err Errno) {
```
codefromthecrypt, line 338: same note on result vs not consistency

wasi/wasi.go
```diff
@@ -333,6 +414,29 @@ func Args(args ...string) (Option, error) {
 	}, nil
 }
 
+// Environ returns an option to set environment variables to the API.
+// Environ returns an error if the input contains a string not joined with `=`, or if the inputs are too large.
+//  * environ: environment variables in the same format as that of `os.Environ`, where key/value pairs are joined with `=`.
+// See https://pkg.go.dev/os#Environ
```
codefromthecrypt, line 420: this is fine

```suggestion
// See os.Environ
```

wasi/wasi_test.go
```diff
@@ -200,8 +200,210 @@ func TestAPI_ArgsSizesGet_Errors(t *testing.T) {
...
+	})
+
+	errorTests := []struct {
+		name        string
+		environ     string
+		errorMesage string
+	}{
+		{name: "error invalid utf-8",
+			environ:     "non_utf8=\xff\xfe\xfd",
+			errorMesage: "value at 0 is not a valid UTF-8 string"},
```
codefromthecrypt, line 224: I'm expecting environ[0]

codefromthecrypt (commented):
cool I noticed something confusing about how getX and getXsizes is defined in WASI, but let's be consistent with their confusion vs inconsistent in it. (about result vs not result) basically, use the docs when there's a conflict

wasi/wasi.go
codefromthecrypt: yeah and I will fix any other related drift in next PR. sorry

wasi/wasi.go
codefromthecrypt: indeed

wasi/wasi.go
```diff
@@ -82,9 +84,60 @@ type API interface {
 	// See https://en.wikipedia.org/wiki/Null-terminated_string
 	ArgsSizesGet(ctx wasm.HostFunctionCallContext, resultArgc, resultArgvBufSize uint32) Errno
 
-	// TODO: EnvironGet(ctx api.hostFunctionCallContext, environ, environBuf uint32) Errno
+	// EnvironGet is the WASI function that reads environment variables.
```
codefromthecrypt, line 87: link to the option that controls it, which should be named `Environ` for consistency both in wasi and also in go

```suggestion
	// EnvironGet is the WASI function that reads environment variables (Environ).
```

wasi/wasi_test.go
```diff
@@ -200,8 +200,210 @@ func TestAPI_ArgsSizesGet_Errors(t *testing.T) {
...
+		'?',                // resultEnvironBufSize is after this
+		0x9, 0x0, 0x0, 0x0, // little endian-encoded size of null terminated strings
+		'?', // stopped after encoding
+	}
+	store, wasiAPI := instantiateWasmStore(t, FunctionEnvironSizesGet, ImportEnvironSizesGet, "test", envOpt)
+
+	t.Run("API.EnvironSizesGet", func(t *testing.T) {
+		maskMemory(store, maskLength)
+
+		// provide a host context we call directly
```
mathetake, line 342: super nit but usually independent comment lines should start with capital letters and end with period here and everywhere 

```suggestion
		// Provide a host context we call directly.
```

mathetake (approved):

wasi/wasi_test.go
codefromthecrypt: I'll start to do that also, I think I haven't

PR #217, opened 2022-02-09
jit(arm64): support for br, br_if and labels

mathetake:
part of #187 

there's some of overlaps with amd64, but I intentionally not share the code,
as they are slightly different, and I'm not sure if the new JIT target (e.g. RISCV)
can reuse the code.

### DIFF
--- a/wasm/jit/jit_arm64.go
+++ b/wasm/jit/jit_arm64.go
@@ -428,7 +428,7 @@ func (c *arm64Compiler) compileBr(o *wazeroir.OperationBr) error {
 		labelKey := o.Target.String()
 		targetLabel := c.label(labelKey)
 		if targetLabel.callers > 1 {
-			// If the number of callers to the target label is larget than one,
+			// If the number of callers to the target label is larger than one,
 			// we have multiple origins to the target branch. In that case,
 			// we must have unique register state.
 			if err := c.releaseAllRegistersToStack(); err != nil {
--- a/wasm/jit/jit_arm64.go
+++ b/wasm/jit/jit_arm64.go
@@ -59,6 +59,12 @@ func newCompiler(f *wasm.FunctionInstance, ir *wazeroir.CompilationResult) (comp
 		locationStack: newValueLocationStack(),
 	}
 
+	if ir != nil {
+		compiler.labels = make(map[string]*labelInfo, len(ir.LabelCallers))
+		for key, callers := range ir.LabelCallers {
+			compiler.labels[key] = &labelInfo{callers: callers}
+		}
+	}
 	return compiler, nil
 }
 
--- a/wasm/jit/jit_arm64.go
+++ b/wasm/jit/jit_arm64.go
@@ -68,15 +68,47 @@ type arm64Compiler struct {
 	// locationStack holds the state of wazeroir virtual stack.
 	// and each item is either placed in register or the actual memory stack.
 	locationStack *valueLocationStack
+	// labels holds per wazeroir label specific informationsin this function.
+	labels map[string]*labelInfo
+	// maxStackPointer tracks the maximum value of stack pointer (from valueLocationStack).
+	maxStackPointer uint64
 }
 
 // compile implements compiler.compile for the arm64 architecture.
 func (c *arm64Compiler) compile() (code []byte, staticData compiledFunctionStaticData, maxStackPointer uint64, err error) {
+	// c.maxStackPointer tracks the maximum stack pointer across all valueLocationStack(s)
+	// used for all labels (via replaceLocationStack), excluding the current one.
+	// Hence, we check here if the final block's max one exceeds the current c.maxStackPointer.
+	maxStackPointer = c.maxStackPointer
+	if maxStackPointer < c.locationStack.maxStackPointer {
+		maxStackPointer = c.locationStack.maxStackPointer
+	}
+
 	code, err = mmapCodeSegment(c.builder.Assemble())
 	if err != nil {
 		return
 	}
 	return
+}
+
+type labelInfo struct {
+	// callers is the number of call sites which may jump into this label.
+	callers int
+	// initialInstruction is the initial instruction for this label so other block can jump into it.
+	initialInstruction *obj.Prog
+	// initialStack is the initial value location stack from which we start compiling this label.
+	initialStack *valueLocationStack
+	// labelBeginningCallbacks holds callbacks should to be called with initialInstruction
+	labelBeginningCallbacks []func(*obj.Prog)
+}
+
+func (c *arm64Compiler) label(labelKey string) *labelInfo {
+	ret, ok := c.labels[labelKey]
+	if ok {
+		return ret
+	}
+	c.labels[labelKey] = &labelInfo{}
+	return c.labels[labelKey]
 }
 
 func (c *arm64Compiler) newProg() (inst *obj.Prog) {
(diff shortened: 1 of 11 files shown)


mathetake:
ok done with label and br including tests.. left is notorious br_if

mathetake:
cool done with all

mathetake:
After this, the only thing is to implement `call` to support fibonacci!

wasm/jit/jit_arm64.go
```diff
@@ -59,35 +59,84 @@ func newCompiler(f *wasm.FunctionInstance, ir *wazeroir.CompilationResult) (comp
...
+
 	code, err = mmapCodeSegment(c.builder.Assemble())
 	if err != nil {
 		return
 	}
 	return
 }
 
+type labelInfo struct {
+	// callers is the number of call sites which may jump into this label.
```
codefromthecrypt, line 104: "may" or is will this crash? wondering again if this is max vs ceil (ex call sites we are aware of due to compilation). This also clarifies impact, ex what is this for? if a max we crash on breach. if a ceil, we optimize according to it. Is it potentially zero? if so why would we care?

wasm/jit/jit_arm64.go
```diff
@@ -336,15 +431,122 @@ func (c *arm64Compiler) compileGlobalSet(o *wazeroir.OperationGlobalSet) error {
...
 
-func (c *arm64Compiler) compileBrIf(o *wazeroir.OperationBrIf) error {
-	return fmt.Errorf("TODO: unsupported on arm64")
+// assignJumpTarget assigns the given label's initial instruction to the destination of jmp.
+func (c *arm64Compiler) assignJumpTarget(labelKey string, jmp *obj.Prog) {
+	target := c.label(labelKey)
+	if target.initialInstruction != nil {
+		jmp.To.SetTarget(target.initialInstruction)
+	} else {
+		// This case, the target label hasn't been compilde yet, so we append the callback and assign
```
codefromthecrypt, line 544: s/compilde/compiled/g

wasm/jit/jit_arm64_test.go
```diff
@@ -938,3 +961,322 @@ func TestArm64Compiler_compieleDrop(t *testing.T) {
...
+		compiler := env.requireNewCompiler(t)
+		err := compiler.emitPreamble()
+		require.NoError(t, err)
+
+		// Branch into nil label is interpreted as return. See BranchTarget.IsReturnTarget
+		err = compiler.compileBr(&wazeroir.OperationBr{Target: &wazeroir.BranchTarget{Label: nil}})
+		require.NoError(t, err)
+
+		// Compile and execute the code under test.
+		// Note that we don't invoke "compiler.return()" as the code emitted by compilerBr is enough to exit.
```
codefromthecrypt, line 1010: nice:
```suggestion
		// Note: we don't invoke "compiler.return()" as the code emitted by compilerBr is enough to exit.
```

wasm/jit/jit_arm64_test.go
```diff
@@ -938,3 +961,322 @@ func TestArm64Compiler_compieleDrop(t *testing.T) {
...
+					x2, x1 = x1, x2
+				}
+				requirePushTwoInt32Consts(t, x1, x2, compiler)
+				// Lt on unsigned integer produces the value on COND_LO register.
+				err := compiler.compileLt(&wazeroir.OperationLt{Type: wazeroir.SignedTypeUint32})
+				require.NoError(t, err)
+			},
+		},
+		{
+			name: "LT",
```
codefromthecrypt, line 1206: this makes me feel we can do compileLT instead of compileLt

codefromthecrypt (approved):
Looks good! note also to @nullpo-head I'm being increasingly "asky" on PRs as we proceed. You'll notice more comments like this to establish better and better baselines in case they are copy/pasted as copy/pasted things amplify tech debt later (think later compilers)

wasm/jit/jit_arm64.go
mathetake: no, may means we have conditional jumps (br_if) where we may or may not jump.

wasm/jit/jit_arm64.go
```diff
@@ -59,35 +59,84 @@ func newCompiler(f *wasm.FunctionInstance, ir *wazeroir.CompilationResult) (comp
...
 	code, err = mmapCodeSegment(c.builder.Assemble())
 	if err != nil {
 		return
 	}
 	return
 }
 
+type labelInfo struct {
+	// callers is the number of call sites which may jump into this label.
+	callers int
```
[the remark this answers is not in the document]
mathetake: Ceil doesn't make sense here as I said the number of callsites is static and doesn't  change for label.

wasm/jit/jit_arm64.go
mathetake: Will clarify more in the comment.

codefromthecrypt: ok maybe if that's the case, is it ok to be literal and use the symbol name? 

Ex ... `brIfCount is the count of "br_if" (OpcodeBrIf) sites that may or may not jump to this label`

Where possible we can use symbol names in wasm to differentiate wasm lingo from arm64 ones, where using most precise wins

mathetake: I would just drop "may"! !https://github.com/tetratelabs/wazero/commit/ffcee6cd4aaa550862333dfd2b13a1f8e6f6a6fd

wasm/jit/jit_arm64.go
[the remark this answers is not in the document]
mathetake: uint32 is enough as that means at least 4GB instructions for a function :D

codefromthecrypt:
>
>
> uint32 is enough as that means at least 4GB instructions for a function :D
>
I was actually wondering smaller, but we can do that by limit vs via type
when/if needed

wasm/jit/jit_arm64.go
```diff
@@ -336,15 +431,122 @@ func (c *arm64Compiler) compileGlobalSet(o *wazeroir.OperationGlobalSet) error {
...
+}
+
+func (c *arm64Compiler) branchInto(target *wazeroir.BranchTarget) error {
+	if target.IsReturnTarget() {
 		return c.returnFunction()
 	} else {
-		return fmt.Errorf("TODO: only return target is available on arm64")
+		labelKey := target.String()
+		targetLabel := c.label(labelKey)
+		if targetLabel.callers > 1 {
```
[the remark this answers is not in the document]
mathetake: zero possible and allowed in wasm:

```
(block
  (br 0)
  (block i32.const 1111)
)
```

This example the label corresponding to `(block i32.const 1111)` is never be reached _at runtime_ because `br 0` exits the function before we reach there

wasm/jit/jit_arm64_test.go
mathetake: so I will keep as-is but renaming to compilerLT works for me in anyway

wasm/jit/jit_arm64.go
```diff
@@ -314,7 +371,45 @@ func (c *arm64Compiler) compileHostFunction(address wasm.FunctionAddress) error
...
+	labelInfo := c.label(labelKey)
+
+	// If initialStack is not set, that means this label has never been reached.
+	if labelInfo.initialStack == nil {
+		skipThisLabel = true
+		return
+	}
+
+	// We use NOP as a beginning of instructions in a label.
+	// This should be eventually optimized out by assembler.
```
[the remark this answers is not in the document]
mathetake: because it is NOP, so that's why I omitted

codefromthecrypt:
can you add the clarification on zero to the docs on the field name? If you tab indent after '//" at least goland formats properly. Edge case notes like this are worth keeping in the doc vs PR comments.

```
// Note: zero possible and allowed in wasm. Ex.
//
//	(block
//	  (br 0)
//	  (block i32.const 1111)
//	)
//
// This example the label corresponding to `(block i32.const 1111)` is never be reached at runtime because `br 0` exits the function before we reach there
```

mathetake:
added on the wazeroir side!

mathetake:
I think I should reuse the compilationResult to reduce the drifts.. on it

PR #218, opened 2022-02-09
wasi: removes manually run integration tests of tinygo+wasi

codefromthecrypt:
We recently added test files to ensure TinyGo is compatible with our
implementation of WASI functions. The presumed intent is that due to
unavailability or us not using an available WASI TCK, we had to
subsitute that manually ad-hoc with TinyGo.

Doing this ends up checking in even more binaries into our source tree
for very little gain. For example, it is nearly 400K to test only that
args are printed out. Meanwhile, we already test in multiple ways our
args functionality, both unit tests and also by accessing them in wasm.

The additional value of also adding a tree in the tests directory is
minimal at best, and doing this would only help if it were run on
change, and it wasn't. No automation built this directory, and I don't
suggest we "fix" that, because it is digging a hole.

If somehow we are able to express re-used text format signatures, test
our impl and also still get it wrong, we should figure out why and
correct it. Meanwhile, checking in binaries in an antipractice for
reasons including bloat of the source repo. The more extra work we do,
the longer change review takes and the more tech debt is accumulated in
the process. Stopping this now is the very best time to do it.


codefromthecrypt:
cc also @pkedy as you've been in this code lately, too!

tests/codec/codec_test.go
```diff
@@ -93,11 +91,6 @@ func newExample() *wasm.Module {
 
 func TestExampleUpToDate(t *testing.T) {
 	ctx := context.Background()
-	encoded := binary.EncodeModule(example)
-	// This means we changed something. Overwrite the example wasm file rather than force maintainers to use hex editor!
```
codefromthecrypt, line 97: this isn't necessary as a round-trip test happens below and serializing the same bytes is redundant

codefromthecrypt:
PS one might argue that we shouldn't check in binaries for other things like the engine tests. agreed. this is mainly waiting on feature parity on the text compiler which is horsepower limited. Meanwhile the scale of that problem isn't much an issue tactically:

```
$ du -sk tests/engine/testdata/
44	tests/engine/testdata/
```

mathetake (approved):
Ack the rationale

nullpo-head (approved):
sounds reasonable

codefromthecrypt:
Thanks for the consideration, and looking forward to one less thing to have to ask contributors to do when implementing WASI also, something very easy to forget to ask

Issue #204, opened 2022-02-05
Make wasm.HostFunctionCallContext an interface


codefromthecrypt:
Right now, someone can accidentally mess up the memory of the host function, I think both the context and its members should be accessed via interfaces, which can provide functions to manipulate memory instead of direct buffer access. This will ensure bounds etc aren't violated.

Particularly this means not exporting `Memory.Buffer` as Memory is becomes an interface and it has interface functions to perform common tasks.

mathetake:
+1

mathetake:
One thing we have to care about is that the cost of accessing via interface vs just unexporting memory instance field

mathetake:
As each host call is equivalent to making system calls so people want them as fast as possible!

codefromthecrypt:
I guess in any case there will be an unexported field, so what you are saying is maybe we leave it as a struct with exported functions to manipulate mem vs define those same functions in an interface?

ex.

```
	MemoryInstance struct {
		buffer []byte // inexpert only this
		Min    uint32 // PS especially below would make sense to also not export as export makes it mutable
		Max    *uint32
	}

func (m *MemoryInstance) PutUint32(addr uint32, val uint32) bool {
--snip--
```

vs defining and matching an interface
```
type MemoryInstance interface {
  PutUint32(addr uint32, val uint32) bool
  Min() uint32
  Max() *uint32
}

	memoryInstance struct {
--
	}

func (m *memoryInstance) PutUint32(addr uint32, val uint32) bool { // matches 
--snip--
``

That we expect the type dispatch to drop performance in a way we should care about. Possible! and luckily benchmark able. Meanwhile our own code can avoid that type dispatch as we can also choose to export it, but put it in an internal package. So the main thing is if user-code using an interface is slow enough to never do that, not here or anywhere else.

Do I get it right?

codefromthecrypt:
I guess one point I mean to say is that this is about the user perspective. Anything we do for speed internally is possible thankfully for go's internal special package. Since we won't allow alternate implementations to be returned, we can guarantee what type is used and make sure that internally it is always specific, and so our jit can use that type and not use the interface

codefromthecrypt:
Looking at #205 sounds helpful to prefer structs over interfaces internally when we are doing a lot of dispatch, as it can save 1.5ns due to the extra pointer indirection.

OTOH for what we refer to as the public API, I think interfaces are better as it prevents field mutability and generally reduces the surface area of bugs such as we've gotten lately. 

[...]

PR #219, opened 2022-02-09
wasi: adds wasi.RegisterAPI and rewrites documentation to prefer go and text format

codefromthecrypt:
Before, we wrote WASI functions with jargon that is not typically used
in go. As this project is in go and implements a text compiler, we can
relay points more directly and specifically. This is the first example
to move to this.

This centralizes documentation through a new interface `wasi.API`
currently instantiated in `wasi.RegisterAPI(store, opts)`. Later, we
can figure out how to export the context needed to use it properly.


wasi/wasi.go
```diff
@@ -15,6 +15,75 @@ import (
...
 )
 
+// Api is a documentation interface for WASI exported functions in version "wasi_snapshot_preview1"
+//
+// Note: In WebAssembly 1.0 (MVP), there may be up to one Memory per store, which means the precise memory is always
+// wasm.Store Memories index zero: `store.Memories[0].Buffer`
+//
+// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md
+// See https://www.w3.org/TR/wasm-core-1/#memory-instances%E2%91%A0.
+type Api interface {
```
mathetake, line 25: another pov: this is especially great as this provides a simple way for users to provide there own WAIS impls. I love this.

wasi/wasi.go
```diff
@@ -15,6 +15,75 @@ import (
...
+	//   * FunctionArgsSizesGet argv_buf_size are written to this offset
+	//
+	// In WebAssembly 1.0 (MVP) Text format, this signature is:
+	//	(import "wasi_snapshot_preview1" "args_get"
+	//		(func ($wasi_args_get (param $argv i32) (param $argv_buf i32) (result i32))
+	//
+	// For example, if FunctionArgsSizesGet wrote argc=2 and argvBufSize=5 for arguments: "a" and "bc"
+	//   and FunctionArgsGet parameters argv=6 and argvBuf=1, we expect `store.Memories[0].Buffer` to contain:
+	//
+	//               argvBufSize                argc * 4
```
codefromthecrypt, line 72:
```suggestion
	//               argvBufSize                argc * 2
```

mathetake:
Then maybe we should unexport `WASIEnvironment` and modify `NewEnvironment` as 

```
func NewApi(opt ..) *Api {
    ...
    return &wasiEnvironment{} // Unexported
}
```

wasi/wasi.go
```diff
@@ -15,6 +15,75 @@ import (
...
+	//            +----------------+     +--------------------+
+	//            |                |     |                    |
+	// []byte{?, 'a', 0, 'b', 'c', 0, ?, 0, 0, 0, 1, 0, 0, 0, 3, ?}
```
codefromthecrypt, line 75: ps, later I'll rewrite the unit tests (another PR) to look exactly like this. When there are less than a hundred bytes, it is sometimes best to have tests be literal in expectations vs doing codec internal to test functions

mathetake:
nit: maybe `API` not `Api` (as Go prefers all cap for abbreviation), but either works for me, up to you

wasi/wasi.go
codefromthecrypt: I will rewrite them now as this actually shows a problem.. basically this is little-endian and it becomes obvious when tests are literal

wasi/wasi.go
```diff
@@ -15,6 +15,75 @@ import (
...
+	//	(import "wasi_snapshot_preview1" "args_sizes_get"
+	//		(func ($wasi_args_sizes_get (param $argc i32) (param $argv_buf_size i32) (result i32))
+	//
+	// For example, if Args are []string{"a","bc"} and
+	//   FunctionArgsSizesGet parameters argc=1 and argvBufSize=6, we expect `store.Memories[0].Buffer` to contain:
+	//
+	//             uint32         uint32
+	//           +--------+     +--------+
+	//           |        |     |        |
+	// []byte{?, 0, 0, 0, 2, ?, 0, 0, 0, 5, ?}
```
nullpo-head, line 44: Since, this is little-endian, rather it would be 
```suggestion
	// []byte{?, 2, 0, 0, 0, ?, 5, 0, 0, 0, ?}
```

codefromthecrypt: good point and yeah that's what I'm redoing the tests about now. binary encoding is not forget when asserted literally.

codefromthecrypt:
>> Then maybe we should unexport `WASIEnvironment` and modify `NewEnvironment` as 

interesting... what this would promote is someone using `WASIEnvironment` to directly invoke functions like GetArgs (as opposed to via wasm). I haven't thought through this.. should we do that?

wasi/wasi.go
codefromthecrypt: sorry about my late comment, I think you just here mentioned the intent is to allow folks direct access. works for me! only thing we need to add to the interface is `Register(store *wasm.Store) (err error)`

mathetake:
Then at least we should have something like `var _ Api = &WASIEnvironmet{}` (the one you hate:D) in order to stop the drifts like WASIEnvironmet somehow doesn't implement Api anymore, etc

mathetake:
either seems fine to me but the point is I want to enforce somehow WASIEnvironment actually implements Api rather than only comments

codefromthecrypt:
hehe I only hate the compile check when there's no comment on it 😁

wasi/wasi.go
codefromthecrypt: We have to think more about returning this to end users, as each function needs a `HostFunctionCallContext`. Probably we should turn that into an interface before exposing it as a return val or something. See #204

codefromthecrypt:
ok this reworked the tests in a way that makes memory writes obvious (IMHO) PTAL!

wasi/wasi_test.go
```diff
@@ -15,290 +15,298 @@ import (
...
+	store, wasiAPI := instantiateWasmStore(t, argsWat, "test", args)
+
+	t.Run("API.ArgsSizesGet", func(t *testing.T) {
+		maskMemory(store, maskLength)
+
+		// provide a host context we call directly
+		hContext := wasm.NewHostFunctionCallContext(ctx, store.Memories[0])
+
+		// invoke ArgsSizesGet directly and check the memory side-effects!
+		errno := wasiAPI.ArgsSizesGet(hContext, argc, argvBufSize)
```
codefromthecrypt, line 141: this is pretty cool imho as we know nothing in the re-exporting stuff can interfere with it. this will pass even if the other doesn't due to some odd reason

wasi/wasi_test.go
```diff
@@ -15,290 +15,298 @@ import (
...
-	}
+	args, err := Args("a", "bc")
+	require.NoError(t, err)
 
-	for _, tt := range tests {
-		tc := tt
+	argc := uint32(1)        // arbitrary offset
+	argvBufSize := uint32(6) // arbitrary offset
+	maskLength := 11         // number of bytes to write '?' to tell what we've written
+	expectedMemory := []byte{
```
codefromthecrypt, line 125: if our docs don't match this, there's a bug in impl or the docs!

wasi/wasi.go
```diff
@@ -100,88 +170,102 @@ type fileEntry struct {
...
 }
 
-// newWASIStringArray creates a wasiStringArray from the given string slice. It returns an error
-// if the length or the total buffer size of the result WASIStringArray exceeds the max of uint32
-func newWASIStringArray(args []string) (*wasiStringArray, error) {
-	if args == nil {
-		return &wasiStringArray{nullTerminatedValues: [][]byte{}}, nil
+// newNullTerminatedStrings creates a nullTerminatedStrings from the given string slice. It returns an error
+// if the length or the total buffer size of the result WASIStringArray exceeds the maxBufSize
+func newNullTerminatedStrings(maxBufSize uint32, args ...string) (*nullTerminatedStrings, error) {
```
codefromthecrypt, line 209: I rewrote this for a little less logic and also control over errors. maxUint32 is probably never a correct value and that policy should be decoupled. Once it is decoupled, tests are easy

wasi/wasi.go
```diff
@@ -100,88 +170,102 @@ type fileEntry struct {
...
-func NewEnvironment(opts ...Option) *WASIEnvironment {
-	ret := &WASIEnvironment{
-		args:   &wasiStringArray{},
+// RegisterAPI adds each function API to the wasm.Store via AddHostFunction.
+func RegisterAPI(store *wasm.Store, opts ...Option) error {
+	_, err := registerAPI(store, opts...)
+	return err
+}
+
+// TODO: we can't export a return with API until we figure out how to give users a wasm.HostFunctionCallContext
```
codefromthecrypt, line 259: hmm maybe it is ok, just we need an example. Ex if you define your own host function, invocations to that will have an instance of wasm.HostFunctionCallContext, which you could pass to API. In other words, `wasi.API` would only be used inside something that already has a context (apart from these tests). Make sense?

wasi/wasi.go
```diff
@@ -100,88 +170,102 @@ type fileEntry struct {
...
 	for i, arg := range args {
-		argLen := uint64(len(arg)) + 1 // + 1 for "\x00"
-		if argLen > uint64(math.MaxUint32-totalBufSize) {
-			return nil, fmt.Errorf("the required buffer size for the args exceeds the max of uint32: %v", uint64(totalBufSize)+argLen)
+		if !utf8.ValidString(arg) {
+			return nil, fmt.Errorf("arg[%d] is not a valid UTF-8 string", i)
+		}
+		argLen := uint64(len(arg)) + 1 // + 1 for "\x00"; uint64 in case this one arg is huge
+		nextSize := uint64(totalBufSize) + argLen
+		if nextSize > uint64(maxBufSize) { //
```
mathetake, line 221: looks like you're dying while writing :D

codefromthecrypt: lol

codefromthecrypt: I moved part of the comment up and left this..
```suggestion
		if nextSize > uint64(maxBufSize) {
```

wasi/wasi.go
mathetake: oh maybe I was wrong this doesn't work  for us

mathetake: nvm

mathetake (approved):
Brilliant!

wasi/wasi.go
mathetake: well I was right:

```

type API interface {
	Foo()
}

type wazeroImpl struct{}

func (a *wazeroImpl) Foo() {
	fmt.Println("wazeroImpl")
}

func newAPI() API {
	return &wazeroImpl{}
}

type userImpl struct {
	defaultImpl API // Embed the interface
}

func (i *userImpl) Foo() {
	i.defaultImpl.Foo()
	fmt.Println("userImpl")
}

func callFoo(impl API) {
	v := reflect.ValueOf(impl.Foo)
	v.Call(nil)
}

func Test_a(t *testing.T) {
	ownImpl := userImpl{newAPI()}
	callFoo(&ownImpl)
}
```

mathetake: as long as we have an API something like `RegisterOverriddenAPI(userImpl, store)` we could allow users to override the WASI func as much as they like __in theory__

mathetake: anyway LGTM for now

wasi/wasi.go
```diff
@@ -100,88 +170,102 @@ type fileEntry struct {
...
-	return func(w *WASIEnvironment) {
+	return func(w *api) {
 		w.stderr = writer
 	}
 }
 
-// wasiStringArray holds null-terminated strings. It ensures that
+// nullTerminatedStrings holds null-terminated strings. It ensures that
 // its length and total buffer size don't exceed the max of uint32.
 // Each string can have arbitrary byte values, not only utf-8 encoded text.
```
nullpo-head, line 195: We can delete this line since now we're restring it to be a utf-8 string,
```suggestion
```

wasi/wasi.go
```diff
@@ -100,88 +170,102 @@ type fileEntry struct {
...
-// wasiStringArray are convenience struct for args_get and environ_get. (environ_get is not implemented yet)
+// nullTerminatedStrings are convenience struct for args_get and environ_get. (environ_get is not implemented yet)
 //
 // A Null-terminated string is a byte string with a NULL suffix ("\x00").
 // See https://en.wikipedia.org/wiki/Null-terminated_string
-type wasiStringArray struct {
+type nullTerminatedStrings struct {
 	// nullTerminatedValues are null-terminated values with a NULL suffix.
 	// Each string can have arbitrary byte values, not only utf-8 encoded text.
```
nullpo-head, line 202: We can delete this line since now we're restring it to be a utf-8 string,
```suggestion
```

nullpo-head (approved):
Thanks a lot!

wasi/wasi.go
codefromthecrypt: cool I think we are talking about mildly different topics, but agree this change is probably a good unit to start with and also help unblock @nullpo-head while we think more

codefromthecrypt:
Thanks for the feedback and help. Hope this sets a good bedrock moving forward 👍

PR #221, opened 2022-02-10
jit(arm64): support call instruction

mathetake:
part of #187


mathetake:
ok cool passed tests

mathetake:
this is kind of a first milestone as this enables arm64 JIT to run fibonacci function!

wasm/jit/jit_amd64.go
```diff
@@ -4870,7 +4870,7 @@ func (c *amd64Compiler) returnFunction() error {
 	// Obtain the temporary registers to be used in the followings.
 	regs, found := c.locationStack.takeFreeRegisters(generalPurposeRegisterTypeInt, 3)
 	if !found {
-		// This in theory never happen as all the registers must be free except addrReg.
+		// This in theory never happen as all the registers must be free.
```
codefromthecrypt, line 4873:
```suggestion
		// This unexpected as all the registers must be free.
```

wasm/jit/jit_arm64.go
```diff
@@ -158,74 +171,63 @@ func (c *arm64Compiler) applyConstToRegisterInstruction(instruction obj.As, cons
...
-func (c *arm64Compiler) applyRegisterToMemoryInstruction(instruction obj.As, baseRegister int16, offset int64, source int16) (err error) {
-	if offset > math.MaxInt16 {
-		// This is a bug in JIT copmiler: caller must check the offset at compilation time, and avoid such a large offset
-		// by loading the const to the register beforehand and then using applyRegisterToRegisterOffsetMemoryInstruction instead.
-		err = fmt.Errorf("memory offset must be smaller than or equal %d, but got %d", math.MaxInt16, offset)
-		return
+func (c *arm64Compiler) applyRegisterToMemoryInstruction(instruction obj.As, sourceRegister int16, destinationBaseRegister int16, destinationOffsetConst int64) {
+	if destinationOffsetConst > math.MaxInt16 {
+		// The assembler can take care of offsets larger than 2^15-1 by emitting additional instructions to load such large offset,
+		// but it uses "its" temporary register which we cannot track. Therefore, we avoid directly emitting memory load with large offsets,
```
codefromthecrypt, line 206:
```suggestion
		// but we cannot track its temporary register. Therefore, we avoid directly emitting memory load with large offsets:
```

wasm/jit/jit_arm64.go
```diff
@@ -158,74 +171,63 @@ func (c *arm64Compiler) applyConstToRegisterInstruction(instruction obj.As, cons
...
+		// but instead load the constant manually to "our" temporary register, then emit the load with it.
```
codefromthecrypt, line 207:
```suggestion
		// load the constant manually to "our" temporary register, then emit the load with it.
```

wasm/jit/jit_arm64.go
```diff
@@ -303,56 +313,116 @@ func (c *arm64Compiler) emitPreamble() error {
...
+	}
 
 	// Since we return from the function, we need to decrement the callframe stack pointer, and write it back.
-	callFramePointerReg, _ := c.locationStack.takeFreeRegister(generalPurposeRegisterTypeInt)
-	if err := c.applyMemoryToRegisterInstruction(arm64.AMOVD, reservedRegisterForEngine,
-		engineGlobalContextCallFrameStackPointerOffset, callFramePointerReg); err != nil {
-		return err
+	tmpRegs, found := c.locationStack.takeFreeRegisters(generalPurposeRegisterTypeInt, 3)
+	if !found {
+		// This should never happen as all the registers must be free at this point.
```
codefromthecrypt, line 326: this comment says no more than the panic, so DRY!
```suggestion
```

wasm/jit/jit_arm64.go
```diff
@@ -303,56 +313,116 @@ func (c *arm64Compiler) emitPreamble() error {
...
+		return fmt.Errorf("BUG: could not find enough free registers")
```
codefromthecrypt, line 327:
```suggestion
		return fmt.Errorf("BUG: all the registers should be free at this point")
```

wasm/jit/jit_amd64.go
```diff
@@ -4870,7 +4870,7 @@ func (c *amd64Compiler) returnFunction() error {
...
 		return fmt.Errorf("could not find enough free registers")
```
codefromthecrypt, line 4874:
```suggestion
		return fmt.Errorf("BUG: all the registers should be free at this point")
```

wasm/jit/jit_arm64.go
```diff
@@ -303,56 +313,116 @@ func (c *arm64Compiler) emitPreamble() error {
...
+	c.emitAddInstructionWithLeftShiftedRegister(
+		callFramePointerReg, callFrameDataSizeMostSignificantSetBit,
+		tmpReg,
+		callFrameStackTopAddressRegister,
+	)
+
+	// At this point, we have
+	//
+	//      [......., ra.caller, rb.caller, rc.caller, _, ra.current, rb.current, rc.current, _, ...]  <- call frame stack's data region (somewhere in the memory)
+	//                                                  |
```
codefromthecrypt, line 372: nit: point to the position not the comma 😛 
```suggestion
	//                                                 ^
```

wasm/jit/jit_arm64.go
```diff
@@ -303,56 +313,116 @@ func (c *arm64Compiler) emitPreamble() error {
...
+		reservedRegisterForEngine, engineValueStackContextStackBasePointerOffset)
+
+	// 2) Branch into the address of "ra.caller".
+	c.applyMemoryToRegisterInstruction(arm64.AMOVD,
+		// "rb.caller" is below the top address.
+		callFrameStackTopAddressRegister, -(callFrameDataSize - callFrameReturnAddressOffset),
+		tmpReg)
+	c.emitUnconditionalBranchToAddressOnRegister(tmpReg)
+
+	// They were temporarily used, so we mark them unused.
```
codefromthecrypt, line 402: when comments say the same thing as the english of the function calls, you an omit them.
```suggestion
```

wasm/jit/jit_arm64.go
```diff
@@ -554,8 +624,229 @@ func (c *arm64Compiler) compileBrTable(o *wazeroir.OperationBrTable) error {
...
+	// Release all the registers as our calling convention requires the caller-save.
+	if err := c.releaseAllRegistersToStack(); err != nil {
+		return err
+	}
+
+	// Obtain the temporary registers to be used in the followings.
+	tmpRegisters, found := c.locationStack.takeFreeRegisters(generalPurposeRegisterTypeInt, 3)
+	if !found {
+		// This in theory never happen as all the registers must be free except addrReg.
+		return fmt.Errorf("could not find enough free registers")
```
codefromthecrypt, line 650:
```suggestion
		return fmt.Errorf("BUG: all registers except addrReg should be free at this point")
```

wasm/jit/jit_arm64.go
```diff
@@ -554,8 +624,229 @@ func (c *arm64Compiler) compileBrTable(o *wazeroir.OperationBrTable) error {
...
+
+	// Release all the registers as our calling convention requires the caller-save.
+	if err := c.releaseAllRegistersToStack(); err != nil {
+		return err
+	}
+
+	// Obtain the temporary registers to be used in the followings.
+	tmpRegisters, found := c.locationStack.takeFreeRegisters(generalPurposeRegisterTypeInt, 3)
+	if !found {
+		// This in theory never happen as all the registers must be free except addrReg.
```
codefromthecrypt, line 649:
```suggestion
```

wasm/jit/jit_arm64.go
```diff
@@ -554,8 +624,229 @@ func (c *arm64Compiler) compileBrTable(o *wazeroir.OperationBrTable) error {
...
+// compileCall implements compiler.compileCall and compiler.compileCallIndirect (TODO) for the arm64 architecture.
+func (c *arm64Compiler) callFunction(addr wasm.FunctionAddress, functype *wasm.FunctionType) error {
+	// TODO: the following code can be generalized for CallIndirect.
+
+	// Release all the registers as our calling convention requires the caller-save.
+	if err := c.releaseAllRegistersToStack(); err != nil {
+		return err
+	}
+
+	// Obtain the temporary registers to be used in the followings.
```
codefromthecrypt, line 646: somewhere above here, we claimed 3 registers and their usage was visually close (a couple lines below all were declared. That's not true here, so we may need to describe them.

In fact, I prefer special-casing so you can declare inline. Ex define

`takeFreeRegister` `takeTwoFreeRegisters` (if ever there's 2) and `takeThreeFreeRegisters`, and that way you can use the results immediately and prevent the indirection of what they are used for

ex.

```
purposeX, purposeY, purposeZ, found := c.locationStack.takeThreeFreeRegisters(generalPurposeRegisterTypeInt)
```

wasm/jit/jit_arm64.go
```diff
@@ -554,8 +624,229 @@ func (c *arm64Compiler) compileBrTable(o *wazeroir.OperationBrTable) error {
...
+	c.emitAddInstructionWithLeftShiftedRegister(
+		tmpRegisters[1], callFrameDataSizeMostSignificantSetBit,
+		tmpRegisters[2],
+		callFrameStackTopAddressRegister,
+	)
+
+	// At this point, we have:
+	//
+	//    [..., ra.current, rb.current, rc.current, _, ra.next, rb.next, rc.next, ...]  <- call frame stack's data region (somewhere in the memory)
+	//                                               |
```
codefromthecrypt, line 674:
```suggestion
	//                                              ^
```

wasm/jit/jit_arm64.go
```diff
@@ -554,8 +624,229 @@ func (c *arm64Compiler) compileBrTable(o *wazeroir.OperationBrTable) error {
...
+		reservedRegisterForEngine, engineValueStackContextStackBasePointerOffset,
+		tmpRegisters[1])
+	c.applyRegisterToMemoryInstruction(arm64.AMOVD,
+		tmpRegisters[1],
+		// "rb.current" is BELOW the top address. See the above example for detail.
+		callFrameStackTopAddressRegister, -(callFrameDataSize - callFrameReturnStackBasePointerOffset))
+
+	// 2) Set engine.valueStackContext.stackBasePointer for the next function.
+	//
+	// At this point, tmpRegisters[1] holds the old stack base pointer. We could get the new frame's
```
codefromthecrypt, line 703: after you refactor special arity takeNFreeRegisters, you can stop `tmpRegisters[1] holds the old stack base pointer` and just say oldBasePointer instead!

wasm/jit/jit_arm64.go
```diff
@@ -554,8 +624,229 @@ func (c *arm64Compiler) compileBrTable(o *wazeroir.OperationBrTable) error {
...
+}
+
+// readInstructionAddress adds an ADR instruction to set the absolute address of "target instruction"
+// into destinationRegister. "target instruction" is specified by beforeTargetInst argument and
+// the target is determined by "the instruction right after beforeTargetInst type".
+//
+// For example, if beforeTargetInst == RET and we have the instruction sequence like
+// ADR -> X -> Y -> ... -> RET -> MOV, then the ADR instruction emitted by this function set the absolute
+// address of MOV instruction into the destination register.
+func (c *arm64Compiler) readInstructionAddress(beforeTargetInst obj.As, destinationRegister int16) {
```
codefromthecrypt, line 795: should this be compileXXX?

wasm/jit/jit_arm64.go
```diff
@@ -554,8 +624,229 @@ func (c *arm64Compiler) compileBrTable(o *wazeroir.OperationBrTable) error {
...
+	}
+
+	// On the function return, we initialize the state for this function.
+	c.initializeReservedStackBasePointerRegister()
+
+	// TODO: initialize module context, and memory pointer.
+	return nil
+}
+
+// readInstructionAddress adds an ADR instruction to set the absolute address of "target instruction"
```
codefromthecrypt, line 788: if you mean a parameter name, use it directly. You can save a sentence by clarifying it on first use

```suggestion
// readInstructionAddress adds an ADR instruction to set the absolute address of beforeTargetInst (the target instruction)
```

wasm/jit/jit_arm64_test.go
```diff
@@ -50,84 +49,84 @@ func (j *jitEnv) requireNewCompiler(t *testing.T) *arm64Compiler {
...
-	env := newJITEnvironment()
+	t.Run("exit", func(t *testing.T) {
+		env := newJITEnvironment()
 
-	// Build code.
-	compiler := env.requireNewCompiler(t)
-	err := compiler.emitPreamble()
-	require.NoError(t, err)
-	compiler.returnFunction()
+		// Build code.
```
codefromthecrypt, line 62: It is my guess eventually we'll have a test util that self-describes when when/where we do preamble, returnFunction, etc. so I'll hold off repetitive comment nits till then, as the test util would centralize such explanations

wasm/jit/jit_arm64_test.go
```diff
@@ -1566,3 +1587,160 @@ func TestArm64Compiler_compileBrIf(t *testing.T) {
...
+		env.exec(code)
+
+		require.Equal(t, jitCallStatusCodeReturned, env.jitStatus())
+	})
+}
+
+func TestArm64Compiler_compieleCall(t *testing.T) {
+	t.Run("need to grow call frame stack", func(t *testing.T) {
+		// TODO
+		t.Skip()
```
codefromthecrypt, line 1675:
```suggestion
		t.Skip("TODO")
```

wasm/jit/jit_arm64_test.go
```diff
@@ -1566,3 +1587,160 @@ func TestArm64Compiler_compileBrIf(t *testing.T) {
...
+
+		env.exec(code)
+
+		require.Equal(t, jitCallStatusCodeReturned, env.jitStatus())
+	})
+}
+
+func TestArm64Compiler_compieleCall(t *testing.T) {
+	t.Run("need to grow call frame stack", func(t *testing.T) {
+		// TODO
```
codefromthecrypt, line 1674:
```suggestion
```

codefromthecrypt (approved):
Looks good. cc @summerwind so you can become more acclimated to the type of feedback on compiler stuff

wasm/jit/jit_arm64.go
mathetake: no, `beforeTargetInst` is not the target and that's why I don't use the param name here.

wasm/jit/jit_arm64.go
mathetake: nope, this is arm64 specific and don't need to be exposed as compiler interface

codefromthecrypt: well we add the XX implements XXX when it is part of the interface, we can still use a common prefix when there are functions that do the same sort of things. That was actually part of why I suggested we enforce the interface in the first place, so that we can tell the difference between naming conventions vs what are common across

wasm/jit/jit_arm64.go
mathetake: in anyway, we end up literally having `temporaryRegisterX` `temporaryRegisterY` and they are no different from using `tmpRegs[0]` and `tmpReg[1]` and that's why I intentionally used `tmpRegs[0, 1, ...]` and  only alias the ones with purpose

mathetake: but I understood the concern, let me refine

wasm/jit/jit_arm64.go
```diff
@@ -554,8 +622,228 @@ func (c *arm64Compiler) compileBrTable(o *wazeroir.OperationBrTable) error {
...
+	// See the comments in engine.pushCallFrame which does exactly the same calculation in Go.
+	c.applyConstToRegisterInstruction(arm64.AADD,
+		int64(c.locationStack.sp)-int64(len(functype.Params)),
+		tmpRegisters[1])
+	c.applyRegisterToMemoryInstruction(arm64.AMOVD,
+		tmpRegisters[1],
+		reservedRegisterForEngine, engineValueStackContextStackBasePointerOffset)
+
+	// Alias for readability.
+	compiledFunctionAddressRegister := tmpRegisters[1]
```
codefromthecrypt, line 711: it is this sort of thing I meant, but ack [1] could be reused for many things, then only now forward used for this purpose. However, if they are always ever only used for one thing, best to alias early. If always temp well, it is what it is, but at least you'd know explicitly!

codefromthecrypt: pps I get that returning a slice is handy on cleanup vs tracking all of them individually. It may be nice to think if a defer function could ever work here. then you don't need to wander around looking if they are deallocated at the end.

wasm/jit/jit_arm64.go
```diff
@@ -368,7 +368,7 @@ func (c *arm64Compiler) returnFunction() error {
 	// At this point, we have
 	//
 	//      [......., ra.caller, rb.caller, rc.caller, _, ra.current, rb.current, rc.current, _, ...]  <- call frame stack's data region (somewhere in the memory)
-	//                                                 ^
+	//                                                  ^
```
mathetake, line 371: actually this should be on comma as this points to the address of a struct, not any specific field.

wasm/jit/jit_arm64.go
mathetake: no I don't think so, this could be re-written and removed in anyway after #221 and I don't think we need this for other archs (e.g. RISC V).  Having methods on interface which are not used anywhere seems weird to me, so let me leave it as-is.

codefromthecrypt: you misunderstood. but the outcome is fine. What I meant is that naming convention != interface. we can re-use a naming convention for things that add instructions without adding them to the interface. put another way no reason for us to require adding an interface just to prefix "compileX".

mathetake: ah ok got it.

mathetake: so then in what way `compile*` (or whatever prefix) common prefix convention helps here?

mathetake: example would help

codefromthecrypt: compileXXX is used currently for operations that end up adding instructions, effecting state via `c.addInstruction(inst)`. I don't see a point using different prefixes to begin describing things that end up doing that. I have been looking at things and if I see compileXXX I think first this is going to add instructions, then I look at what it is doing if it matches the intent either via inherited doc from the interface or what's on the non-shared thing. I look to see if there's an "addInstruction" as formerly "compile" was a keyword for me to look for state changes vs ancillary functions that do not generate instructions. Does this make sense?

mathetake: yeah make sense esp this

> effecting state via c.addInstruction(inst).

PR #223, opened 2022-02-10
wasi: fix typo in WASI `ArgsGet` and `ArgsSizesGet` methods comments

r8d8:
Fix small typo in methods description for mentioned WASI methods

### DIFF
--- a/wasi/wasi.go
+++ b/wasi/wasi.go
@@ -50,7 +50,7 @@ type API interface {
 	//                     offset that begins "bc" --+
 	//
 	// Note: FunctionArgsGet documentation has an example of this signature in the WebAssembly 1.0 (MVP) Text Format.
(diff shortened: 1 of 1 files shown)


wasi/wasi.go
```diff
@@ -50,7 +50,7 @@ type API interface {
 	//                     offset that begins "bc" --+
 	//
 	// Note: FunctionArgsGet documentation has an example of this signature in the WebAssembly 1.0 (MVP) Text Format.
-	// See ArgsSizesGet
```
codefromthecrypt, line 53: oh this was intended as it is telling you to look at the other function which is relevant to this one. If we make the change here, it would be a loop!

r8d8: Ah, ok, I see.
 Will cancel PR

PR #224, opened 2022-02-10
wasi: refactors existing logic and tests of clock_time_get

codefromthecrypt:
This doesn't add new functionality, just ports it to current style


wasi/wasi.go
```diff
@@ -211,6 +231,48 @@ func (a *api) register(store *wasm.Store) (err error) {
 	return nil
 }
 
+// ArgsGet implements API.ArgsGet
```
codefromthecrypt, line 234: moved these to maintain order, no logic changed

wasi/wasi.go
```diff
@@ -87,7 +86,27 @@ type API interface {
...
 
-	// TODO: ClockTimeGet(ctx *wasm.HostFunctionCallContext, id uint32, precision uint64, resultTimestamp uint32) Errno
+	// ClockTimeGet is a WASI function that returns the time value of a clock (time.Now).
+	//
+	// * id - The clock id for which to return the time.
+	// * precision - The maximum lag (exclusive) that the returned time value may have, compared to its actual value.
+	// * resultTimestamp - the offset to write the timestamp to wasm.MemoryInstance Buffer
+	//   * the timestamp is epoch nanoseconds encoded as a uint64 little-endian encoding.
+	//
+	// For example, if time.Now returned exactly midnight UTC 2022-01-01 (1640995200000000000), and
```
nullpo-head, line 96: Expressing it in hex might help readers to compare the value and also help to understand the endianness a little easier.
```suggestion
	// For example, if time.Now returned exactly midnight UTC 2022-01-01 (0x16c5fc70a61f0000), and
```

nullpo-head (approved):

wasi/wasi.go
codefromthecrypt: what I did was make sure each api (not just this one) says little endian two times. once in the description and once in describing the encoding.

codefromthecrypt:
thanks for the feedback and lightning fast review @nullpo-head

PR #225, opened 2022-02-11
jit(arm64): support for and, or, and xor

summerwind:
Hello, The JIT support for Arm64 is really nice!
This PR adds support of `and`, `or`, and `xor` instruction for Arm64 JIT.
This might be a part of #187.

I have implemented this with reference to the x86 JIT code.
If you have any concerns about this, I would appreciate your review comments!

### DIFF
--- a/wasm/jit/jit_arm64.go
+++ b/wasm/jit/jit_arm64.go
@@ -771,16 +771,94 @@ func (c *arm64Compiler) compileRem(o *wazeroir.OperationRem) error {
 	return fmt.Errorf("TODO: unsupported on arm64")
 }
 
+// compileAnd implements compiler.compileAnd for the arm64 architecture.
 func (c *arm64Compiler) compileAnd(o *wazeroir.OperationAnd) error {
-	return fmt.Errorf("TODO: unsupported on arm64")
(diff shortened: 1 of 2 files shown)


wasm/jit/jit_arm64.go
```diff
@@ -771,16 +771,86 @@ func (c *arm64Compiler) compileRem(o *wazeroir.OperationRem) error {
...
+		return err
+	}
+
+	// At this point, at least one of x1 or x2 registers is non zero.
+	// Choose the non-zero register as destination.
+	var destinationReg int16 = x1.register
+	if isZeroRegister(x1.register) {
+		destinationReg = x2.register
+	}
+
```
mathetake, line 850: could you choose either `AEOR` or `AEORW` according to o.Type (size of integer)?

mathetake (commented):
Awesome! Thank you for this, one small fix and then good to go!

mathetake (approved):
LGTM!

mathetake:
Thank you for your contribution (again)! 👯

summerwind:
Thank you!

codefromthecrypt:
it is exciting and motivating to see you again @summerwind 🙇

PR #229, opened 2022-02-12
Fix `fibonacci` example in Readme

r8d8:


codefromthecrypt (approved):
Thanks for the help! indeed we drifted this.. unless we switch to asciidoc I think drift is likely for a while and we'll have to be more careful. Thanks for the help!

Usually I don't tend to put full imports an example, but I can see why this is helpful as it clarifies more than I would have guessed. 👍 

PS Later, we'll revise the text a little as there's no need to talk about TinyGo for fibonacci as that's implementable easy enough in the text format. We'll delay that until our compiler is complete, so no action needed now!

codefromthecrypt:
Thanks @r8d8!

PR #231, opened 2022-02-13
Adds Module.SectionSize and constrains text parsing rules

codefromthecrypt:
This consolidates more code around SectionID, notably consistently
getting and using section ID length. This also enforces rules in text
parsing such as up to one memory, and imported functions can't be
defined after regular ones.

### DIFF
--- a/wasm/module.go
+++ b/wasm/module.go
@@ -1,6 +1,8 @@
 package wasm
 
-// DecodeModule parses the configured source into a wasm.Module. This function returns when the source is exhausted or
+import "fmt"
+
+// DecodeModule parses the configured source into a Module. This function returns when the source is exhausted or
 // an error occurs. The result can be initialized for use via Store.Instantiate.
 //
 // Here's a description of the return values:
@@ -115,7 +117,8 @@ type Module struct {
 	// this module at TableSection[0].
 	//
 	// Note: Version 1.0 (MVP) of the WebAssembly spec allows at most one table definition per module, so the length of
-	// the TableSection can be zero or one.
+	// the TableSection can be zero or one, and can only be one if there is no ImportKindTable.
+	//
 	// Note: In the Binary Format, this is SectionIDTable.
 	//
 	// See https://www.w3.org/TR/wasm-core-1/#table-section%E2%91%A0
@@ -128,7 +131,8 @@ type Module struct {
 	// this module at TableSection[0].
 	//
 	// Note: Version 1.0 (MVP) of the WebAssembly spec allows at most one memory definition per module, so the length of
-	// the MemorySection can be zero or one.
+	// the MemorySection can be zero or one, and can only be one if there is no ImportKindMemory.
+	//
 	// Note: In the Binary Format, this is SectionIDMemory.
 	//
 	// See https://www.w3.org/TR/wasm-core-1/#memory-section%E2%91%A0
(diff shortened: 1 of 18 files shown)


wasm/text/decoder.go
```diff
@@ -18,17 +18,32 @@ const (
...
-	positionModuleType
-	positionModuleImport
-	positionModuleImportFunc
-	positionModuleFunc
-	positionModuleMemory
-	positionModuleExport
-	positionModuleExportFunc
-	positionModuleExportMemory
-	positionModuleStart
+	positionImport
```
codefromthecrypt, line 21: a couple more refactors left (not now) and I'll be able to substitute sectionID for the top-level parsing position, and take these out.

wasm/text/decoder.go
```diff
@@ -331,7 +361,7 @@ func (p *moduleParser) addFunctionName(name string) {
 // Ex. If there is no signature `(import "" "main" (func))`
 //                     calls onImportFunc here ---^
 func (p *moduleParser) parseImportFunc(tok tokenType, tokenBytes []byte, line, col uint32) (tokenParser, error) {
-	idx := wasm.Index(len(p.module.ImportSection))
+	idx := p.module.SectionSize(wasm.SectionIDImport)
```
codefromthecrypt, line 364: I have in-flight work to make an import parser. Due to this `module.SectionSize` function, I can decouple the parser from needing a reference to the module via a function type like... `func(SectionID) uint32`

wasm/module.go
```diff
@@ -461,3 +466,42 @@ func (m *Module) allDeclarations() (functions []Index, globals []*GlobalType, me
 	tables = append(tables, m.TableSection...)
 	return
 }
+
+// SectionSize returns the count of items in a given section
+func (m *Module) SectionSize(sectionID SectionID) uint32 {
+	switch sectionID {
+	case SectionIDCustom:
+		count := uint32(len(m.CustomSections))
+		if m.NameSection != nil {
```
mathetake, line 475: this seems inconsistent to others, as we have multiple custom sections vs unique one for others, and the comment `// SectionSize returns the count of items in a given section` is not true for custom section, right?

mathetake (commented):
looks fine except the inconsistency on custom sections

wasm/module.go
codefromthecrypt: you are right, that is indeed a special case!

we can do one of two things I think.

* Change the wording of that so that it is true even in the special case: `// SectionSize returns the count of items for a given section ID`
* Exclude CustomSection and then the docs around it

I feel the former is less to explain wdyt?

mathetake: what confuses me is that there's no "items" for custom section ID, but there's multiple "sections" for custom section ID. 
`// SectionSize returns the count of items for a given section ID except that if the id is custom section id, returns the number of custom sections` would wok for me

codefromthecrypt:
ps there's another need not addressed here, which is count by second index. Ex.
  * custom sections by subsection ID
  * imports/exports by kind

I didn't address this, yet but I think at some point we can refactor things so that you can for example, tell the function index namespace count w/o a for loop.

mathetake (approved):

wasm/module.go
codefromthecrypt: The spec itself is the confusing thing. I addressed this part and gave examples of "items" instead. How about it?

```
// SectionSize returns the count of items for a given section ID
//
// For example, given...
// * SectionIDType this returns the count of FunctionType
// * SectionIDCustom this returns the count of unique section names
// * SectionIDExport this returns the count of unique export names
```

mathetake: sg

mathetake:
`SectionItemCounts` would be exact? :D

mathetake (approved):

codefromthecrypt:
Thanks for the notes on `CustomSection` @mathetake  it is messy and seems per-spec it is allowed to repeat by name also. I'm inclined to drop support for it and just skip anything not the name section (in another PR)

Issue #232, opened 2022-02-14
Move contents of wasm.Module to an internal package and replace it with an interface 


codefromthecrypt:
Currently, we are doing some re-validation because `wasm.Module` is not required to be created by a `ModuleDecoder`. For example, we double-check certain things like the start function pointing to a valid function type. These sorts of checks can and should be done when parsing/decoding as otherwise you don't know where the failure line/col was. Moreover, there are so many rules that it is a fools errand to repeat each one in instantiate: we'd end up with a lot of code that isn't necessary if decoded properly. Finally, exposing fields like the `CodeSection` to mutation allow some pretty severe bugs to be possible. For example, someone could by accident mess up a signature order, something that decoders already enforce.

I don't think 3rd party manual instantiation of `wasm.Module` is viable. The only sustainable way is via validating decoders which we already maintain. I believe that we should instead move the existing logic of `wasm.Module` internal, guaranteeing it is only created by our trusted decoders. We'd backfill `wasm.Module` as a struct or interface that has no exported mutable fields, but could be implemented by or contain an internal `wasm.Module` created by the decoder.

Doing this allows us to move all validation possible to do during decoding to the edge and guarantee we can count on that having happened prior to instantiate. This also removes a large bug source area of random or misunderstood contents in function code or constant declarations.

codefromthecrypt:
this is done via `api.Module`

PR #234, opened 2022-02-14
Removes CustomSection except NameSection

codefromthecrypt:
Per #232, we will be unexporting fields only used during instantiate. Any
additional custom sections are unusable otherwise, so this skips them.
In doing so, this eliminates a length check bug, some TODO and some
logic that enforced strict name uniqueness even though that only rule
only applied to the name section.

Note: this doesn't restrict future use of custom sections, ex for DWARF,
just we don't buffer sections we don't use anymore.

See https://www.w3.org/TR/wasm-core-1/#custom-section%E2%91%A0

### DIFF
--- a/wasm/module.go
+++ b/wasm/module.go
@@ -92,6 +92,7 @@ type Module struct {
 	// (Store.Instantiate).
 	//
 	// Note: there are no unique constraints relating to the two-level namespace of Import.Module and Import.Name.
+	//
 	// Note: In the Binary Format, this is SectionIDImport.
 	//
 	// See https://www.w3.org/TR/wasm-core-1/#import-section%E2%91%A0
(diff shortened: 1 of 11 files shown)


wasm/module.go
```diff
@@ -467,20 +463,19 @@ func (m *Module) allDeclarations() (functions []Index, globals []*GlobalType, me
...
-// For example, given...
-// * SectionIDType this returns the count of FunctionType
-// * SectionIDCustom this returns the count of unique section names
-// * SectionIDExport this returns the count of unique export names
-func (m *Module) SectionSize(sectionID SectionID) uint32 {
+// For example...
+// * SectionIDType returns the count of FunctionType
+// * SectionIDCustom returns one if the NameSection is present
+// * SectionIDExport returns the count of unique export names
+func (m *Module) SectionElementCount(sectionID SectionID) uint32 { // element as in vector elements!
```
mathetake, line 472: I like this better!

mathetake (approved):

nullpo-head (approved):
looks nice!

codefromthecrypt:
Thanks for the peek fellas

PR #235, opened 2022-02-14
jit(arm64): Add common 'compile'/'maybeCompile' prefix

mathetake:
exception are `popValueOnRegister`, `popTwoValuesOnRegisters` and `allocateRegister` and they are left with TODO


mathetake:
amd64 is messy to enforce this as-is, so I will do that in another PR

codefromthecrypt (approved):
thanks a lot!

PR #236, opened 2022-02-14
jit(arm64): support for select, swap, unreachable, globals

mathetake:
part of #187 

 - [x] unreachable
 - [x] select
 - [x] swap
 - [x] global.get
 - [x] global.set


mathetake:
Done. So I will polish comments tomorrow then ready

wasm/jit/jit_arm64.go
```diff
@@ -171,33 +176,37 @@ func (c *arm64Compiler) compileConstToRegisterInstruction(instruction obj.As, co
 
 // compileMemoryToRegisterInstruction adds an instruction where source operand is a memory location and destination is a register.
 // baseRegister is the base absolute address in the memory, and offset is the offset from the absolute address in baseRegister.
```
codefromthecrypt, line 178: I think some of these docs drifted from the param names

wasm/jit/jit_arm64.go
```diff
@@ -300,15 +329,24 @@ func (c *arm64Compiler) pushFunctionParams() {
...
-	nop := c.newProg()
-	nop.As = obj.ANOP
-	c.addInstruction(nop)
+	c.compileNOP()
 
 	c.pushFunctionParams()
 
+	// TODO: Add maybeGrowValueStack() .
+
 	// Before excuting function body, we must initialize the stack base pointer register
```
codefromthecrypt, line 338:
```suggestion
	// Before executing function body, we must initialize the stack base pointer register
```

wasm/jit/jit_arm64.go
```diff
@@ -476,20 +512,141 @@ func (c *arm64Compiler) compileLabel(o *wazeroir.OperationLabel) (skipThisLabel
...
+	return c.exit(jitCallStatusCodeUnreachable)
 }
 
+// compileSwap implements compiler.compileSwap for the arm64 architecture.
 func (c *arm64Compiler) compileSwap(o *wazeroir.OperationSwap) error {
-	return fmt.Errorf("TODO: unsupported on arm64")
+	x := c.locationStack.peek()
+	y := c.locationStack.stack[int(c.locationStack.sp)-1-o.Depth]
+
+	if err := c.maybeCompileEnsureOnGeneralPurposeRegister(x); err != nil {
```
codefromthecrypt, line 525: are maybe and ensure redundant? I think ensure was used to say "unless already" which is same as maybe unless I am missing something

wasm/jit/jit_arm64.go
```diff
@@ -476,20 +512,141 @@ func (c *arm64Compiler) compileLabel(o *wazeroir.OperationLabel) (skipThisLabel
...
 func (c *arm64Compiler) compileUnreachable() error {
-	return fmt.Errorf("TODO: unsupported on arm64")
+	return c.exit(jitCallStatusCodeUnreachable)
 }
 
+// compileSwap implements compiler.compileSwap for the arm64 architecture.
 func (c *arm64Compiler) compileSwap(o *wazeroir.OperationSwap) error {
-	return fmt.Errorf("TODO: unsupported on arm64")
+	x := c.locationStack.peek()
+	y := c.locationStack.stack[int(c.locationStack.sp)-1-o.Depth]
```
codefromthecrypt, line 523: when there's multiple arithmetic maybe a trailing comment like `// Depth is relative to the last stack value` this makes it easier to think. just remember to update that comment if it ever changes :D

wasm/jit/jit_arm64.go
```diff
@@ -476,20 +512,141 @@ func (c *arm64Compiler) compileLabel(o *wazeroir.OperationLabel) (skipThisLabel
...
+	case wasm.ValueTypeI64:
+		mov = arm64.AMOVD
+	case wasm.ValueTypeF32:
+		mov = arm64.AFMOVS
+	case wasm.ValueTypeF64:
+		mov = arm64.AFMOVD
+	}
+
+	// At this point "globalInstanceAddressRegister = globals[index]".
+	// Therefore, this means "globals[index].Val = val.register"
```
codefromthecrypt, line 607: nice

wasm/jit/jit_arm64.go
```diff
@@ -476,20 +512,141 @@ func (c *arm64Compiler) compileLabel(o *wazeroir.OperationLabel) (skipThisLabel
...
+		mov,
+		val.register,
+		globalInstanceAddressRegister, globalInstanceValueOffset,
+	)
+
+	c.markRegisterUnused(val.register)
+	return nil
+}
+
+// compileReadGlobalAddress adds instruction to read the absolute address of the global instance whose index equals globalIndex,
```
codefromthecrypt, line 618: try to tighten if it still means the same
"adds instructions to store the absolute address of the global instance at globalIndex into a register"

wasm/jit/jit_arm64_test.go
```diff
@@ -1743,3 +1746,277 @@ func TestArm64Compiler_compieleCall(t *testing.T) {
...
+			moduleInstance: &wasm.ModuleInstance{},
+		},
+	} {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			env := newJITEnvironment()
+			compiler := env.requireNewCompiler(t)
+			compiler.f.ModuleInstance = tc.moduleInstance
+
+			// The assembler skips the first instruction so we intentionally add NOP here.
```
codefromthecrypt, line 1881: any note to add about why it does that?

codefromthecrypt (approved):
particularly module instantiation is some interesting code (that generates code)

wasm/jit/jit_arm64_test.go
mathetake: I have no idea for the reason but I just confirmed from the code (in golang-asm).. this will become unnecessary after implementing our assembler :D

mathetake:
thanks for the look!

PR #237, opened 2022-02-14
wasi: Implement `random_get` function

r8d8:


wasm/store.go
```diff
@@ -868,15 +870,25 @@ type HostFunctionCallContext struct {
 	ctx context.Context
 	// Memory is the currently used memory instance at the time when the host function call is made.
 	Memory *MemoryInstance
+	// Used in `RandomGet` method
+	Randomizer *rand.Rand
```
mathetake, line 874: we should _not_ define this here because this results in introducing WASI thingy into Wasm space. 
Let's have this in wasm.api struct and have an option to control this

mathetake (changes requested):
Thanks for this! In general we must not have WASI concepts into wasm package

wasi/wasi.go
```diff
@@ -503,6 +520,22 @@ func (a *api) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32) (err Errno)
 	return ErrnoSuccess
 }
 
+func (a *api) RandomGet(ctx *wasm.HostFunctionCallContext, buf uint32, buf_len uint32) (errno Errno) {
+	if !ctx.Memory.ValidateAddrRange(buf, uint64(buf_len)) {
+		return ErrnoInval
+	}
+
+	random_bytes := make([]byte, buf_len)
+	_, err := ctx.Randomizer.Read(random_bytes)
```
mathetake, line 529:
```suggestion
	_, err := a.randSource.Read(random_bytes)
```

mathetake: once you have `randSource` in wasi.api struct, then the code look like this ^^ and you don't need to expose the WASI concepts into wasm space

wasi/wasi.go
```diff
@@ -143,7 +143,24 @@ type API interface {
...
+	// with random seed `rand.NewSource(42)`, we expect `ctx.Memory.Buffer` to contain:
+	//
+	//                             buf_len (5)
+	//                    +------------------------+
+	//                    |                        |
+	//          []byte{?, 0x53, 0x8c, 0x7f, 0x96, 0xb1, ?}
+	//              buf --^
+	//
+	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-random_getbuf-pointeru8-buf_len-size---errno
+	RandomGet(ctx *wasm.HostFunctionCallContext, buf, buf_len uint32) Errno
```
codefromthecrypt, line 162:
```suggestion
	RandomGet(ctx *wasm.HostFunctionCallContext, buf, bufLen uint32) Errno
```

wasi/wasi.go
```diff
@@ -143,7 +143,24 @@ type API interface {
 	// TODO: ProcExit
 	// TODO: ProcRaise
 	// TODO: SchedYield
-	// TODO: RandomGet
+
+	// RandomGet is a WASI function that write random data in buffer (rand.Read()).
+	//
+	// * buf - is a offset to write random values
+	// * buf_len - size of random data in bytes
```
codefromthecrypt, line 150:
```suggestion
	// * bufLen - size of random data in bytes
```

wasi/wasi.go
```diff
@@ -143,7 +143,24 @@ type API interface {
...
+	//
+	// For example, if `HostFunctionCallContext.Randomizer` initialized
+	// with random seed `rand.NewSource(42)`, we expect `ctx.Memory.Buffer` to contain:
+	//
+	//                             buf_len (5)
```
codefromthecrypt, line 155:
```suggestion
	//                             bufLen (5)
```

wasi/wasi.go
```diff
@@ -143,7 +143,24 @@ type API interface {
...
+	//                    +------------------------+
```
codefromthecrypt, line 156:
```suggestion
	//                    +--------------------------+
```

wasi/wasi.go
```diff
@@ -143,7 +143,24 @@ type API interface {
...
+	//                    |                        |
```
codefromthecrypt, line 157:
```suggestion
	//                    |                          |
```

wasi/testdata/random.wat
```diff
@@ -0,0 +1,24 @@
...
+  ;; because wazero's interpreter has a bug that it crashes when an imported-and-exported host function
+  ;; is called from the host environment, which will be fixed soon.
+  ;; After it's fixed, these wrapper functions are no longer necessary.
+  (func $random_get (param i32 i32) (result i32)
+        local.get 0
+        local.get 1
+        call $wasi.random_get
+        )
+  (export "random_get" (func $random_get))
+  )
```
codefromthecrypt, line 24: supernit: EOL before EOF please!

wasi/wasi.go
```diff
@@ -503,6 +520,22 @@ func (a *api) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32) (err Errno)
 	return ErrnoSuccess
 }
 
+func (a *api) RandomGet(ctx *wasm.HostFunctionCallContext, buf uint32, buf_len uint32) (errno Errno) {
```
codefromthecrypt, line 523:
```suggestion
// RandomGet implements API.RandomGet
func (a *api) RandomGet(ctx *wasm.HostFunctionCallContext, buf uint32, bufLen uint32) (errno Errno) {
```

wasi/wasi_test.go
```diff
@@ -312,7 +312,69 @@ func TestAPI_ClockTimeGet_Errors(t *testing.T) {
...
+	var buf_len = uint32(5) // arbitrary buffer size,
+	var buf = uint32(1)     // offset,
+	var seed = int64(42)    // and seed value
+
+	t.Run("API.RandomGet", func(t *testing.T) {
+		maskMemory(store, maskLength)
+		// provide a host context with a seed value for random generator
+		hContext := wasm.NewHostFunctionCallContextWithSeed(context.Background(), store.Memories[0], seed)
+
+		errno := wasiAPI.RandomGet(hContext, buf, buf_len)
```
codefromthecrypt, line 338: maybe just change this all over from buf_len -> bufLen except in the %.wat

codefromthecrypt (commented):
You've done a great job. Once the randomizer is re-organized (and probably only exposed for testing), it will be a quick final review!

wasi/wasi_test.go
r8d8: Thanks @codefromthecrypt 
Got it, will use bufLen overall (except %.wat)

codefromthecrypt:
ps we added the DCO today, something we forgot to add before, so that's why your PR is red

wasi/wasi.go
```diff
@@ -162,6 +179,7 @@ type api struct {
 	opened map[uint32]fileEntry
 	// timeNowUnixNano is mutable for testing
 	timeNowUnixNano func() uint64
+	randSource      *rand.Rand
```
mathetake, line 182: I imagine one would expect this to be "cryptographically secure random number generator" rather than the pseudo one (See getrandom(2) on linux https://man7.org/linux/man-pages/man2/getrandom.2.html, and wasi's get_random is the counter part for that).

How about using crypto/rand instead of math.Rand?

wasi/wasi.go
```diff
@@ -354,6 +374,10 @@ func newAPI(opts ...Option) *api {
 	return ret
 }
 
+func (a *api) seedRandSource(seed int64) {
```
mathetake, line 377: I think we don't need this here as this is only for tests

wasi/wasi.go
r8d8: we can, for sure
WASI spec says "high-quality" random values, and it's somewhat confusing what they mean under this.
Will move now to crypto/rand

wasi/wasi.go
```diff
@@ -143,7 +143,24 @@ type API interface {
...
 	// TODO: ProcRaise
 	// TODO: SchedYield
-	// TODO: RandomGet
+
+	// RandomGet is a WASI function that write random data in buffer (rand.Read()).
+	//
+	// * buf - is a offset to write random values
+	// * bufLen - size of random data in bytes
+	//
+	// For example, if `HostFunctionCallContext.Randomizer` initialized
```
codefromthecrypt, line 152: this drifted and also we are probably being too specific on the mechanics of random.

maybe `For example, if bufLen = 5, we expect `ctx.Memory.Buffer` to contain 5 random bytes:`

wasi/wasi.go
codefromthecrypt: yeah I agree crypto is better as many things in WASI promise things of higher security, it fits

wasi/wasi.go
r8d8: @codefromthecrypt 
You mean not to use in comments concrete random bytes from specified seed?
And also not to check concrete random bytes form seed, cause crypto/rand non-deterministic and we can't seed it, right? 
Just check in test that memory is filled

codefromthecrypt: no I mean in this godoc we don't need to get into how secure random works. I still like that you are using tests that are resulting in a known value

r8d8: or we can seed math/rand from crypto/rand, like this
```
var seed int64
binary.Read(crypto_rand.Reader, binary.BigEndian, &seed)
math_rand.NewSource(seed)
```
and still use simple seed and check concrete bytes in tests

r8d8: > no I mean in this godoc we don't need to get into how secure random works. I still like that you are using tests that are resulting in a known value

got it, thanks for your feedback

r8d8:
Will cancel this PR and open new one 
Rebase to fix signoff broked branch
Sorry for inconvinience

PR #238, opened 2022-02-14
Refactors out public API from internals

codefromthecrypt:
This moves to a new end-user API under the root package `wazero`. This
simplifies call sites while hardening function calls to their known
return value. Most importantly, this moves most logic internal, as
noted in the RATIONALE.md.

Ex.

```go
	// Read WebAssembly binary containing an exported "fac" function.
	source, _ := os.ReadFile("./tests/engine/testdata/fac.wasm")

	// Decode the binary as WebAssembly module.
	mod, _ := wazero.DecodeModuleBinary(source)

	// Initialize the execution environment called "store" with Interpreter-based engine.
	store := wazero.NewStore()

	// Instantiate the module, which returns its exported functions
	functions, _ := store.Instantiate(mod)

	// Get the factorial function
	fac, _ := functions.GetFunctionI64Return("fac")

	// Discover 7! is 5040
	fmt.Println(fac(context.Background(), 7))

```

PS I changed the README to factorial because the wat version of
fibonacci is not consistent with the TinyGo one!


mathetake:
> +3,029 −2,271 

img

codefromthecrypt:
sorry refactoring resulted in wasm2 import.. just got rid of that, but yeah
it affects all code due to imports!

codefromthecrypt:
ps I moved the minimal things to public /wasi and /wasm packages, this is mentioned in the RATIONALE why, but these files have no mutable types including var (Error). we can be less strict on that over time and future PRs can move more things into the public area. just I didn't do anything more than needed as it is already a massive change

mathetake:
pushed a commit to fix arm64

RATIONALE.md
```diff
@@ -1,5 +1,67 @@
...
+in a high quality experience. Even if someone had these skills, supporting external code would introduce variables which
+are constants in the central one. Supporting an external codebase is harder on the project team, and could starve time
+from the already large burden on the central codebase.
+
+The tradeoffs of internal packages are a larger codebase and responsibility to implement all standard features. It also
+implies thinking about extension more as forking is not viable for reasons above also. The primary mitigation of these
+realities are friendly OSS licensing, high rigor and a collaborative spirit which aim to make contribution in the shared
+codebase productive.
+
+### Avoiding cyclic dependencies
```
mathetake, line 30: maybe it's time to use the graph? https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/ :D

wasm/types.go
```diff
@@ -0,0 +1,234 @@
+package wasm
```
mathetake, line 1: I doubt that if we need this public `wasm`... at least looks almost everything is only used in internal pakages except interfaces `ModuleFunctions`, `HostFunctionCallContext` and `Memory`..

The reason for that is that maybe it seems annoying to have `internalwasm.Foo` all the time..

wasm/types.go
```diff
@@ -0,0 +1,234 @@
...
+	"context"
+)
+
+// SectionID identifies the sections of a Module in the WebAssembly 1.0 (MVP) Binary Format.
+//
+// Note: these are defined in the wasm package, instead of the binary package, as a key per section is needed regardless
+// of format, and deferring to the binary type avoids confusion.
+//
+// See https://www.w3.org/TR/wasm-core-1/#sections%E2%91%A0
+type SectionID = byte
```
mathetake, line 13: this can be internalized at least

wasm/types.go
mathetake: it is difficult to enforce the naming `internalwasm` when there's not collision possibility, and I think there's no lint for such a thing

mathetake: actually I see `wasm` is used for internal/wasm in engine.go but `internalwasm` is used in jit_amd64.go..

tests/codec/codec_test.go
```diff
@@ -164,7 +162,7 @@ func BenchmarkCodecExample(b *testing.B) {
 		}
 	})
 	// Note: We don't know if wasmer.Wat2Wasm encodes the custom name section or not.
-	// Note: wasmer.Wat2Wasm calls wasmer via CGO which is eventually implemented by wasm-tools
+	// Note: wasmer.Wat2Wasm calls wasmer via CGO which is eventually implemented by internalwasm-tools
```
nullpo-head, line 165: maybe due to replace
```suggestion
	// Note: wasmer.Wat2Wasm calls wasmer via CGO which is eventually implemented by wasm-tools
```

tests/codec/codec_test.go
```diff
@@ -175,7 +173,7 @@ func BenchmarkCodecExample(b *testing.B) {
 		}
 	})
 	// Note: We don't know if wasmtime.Wat2Wasm encodes the custom name section or not.
-	// Note: wasmtime.Wat2Wasm calls wasmtime via CGO which is eventually implemented by wasm-tools
+	// Note: wasmtime.Wat2Wasm calls wasmtime via CGO which is eventually implemented by internalwasm-tools
```
nullpo-head, line 176: maybe due to replace
```suggestion
	// Note: wasmer.Wat2Wasm calls wasmer via CGO which is eventually implemented by wasm-tools
```

wasi/functions.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// FunctionArgsGet reads command-line argument data.
+	//
+	// See ImportArgsGet
+	// See FunctionArgsSizesGet
+	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-args_getargv-pointerpointeru8-argv_buf-pointeru8---errno
+	FunctionArgsGet = "args_get"
+
+	// ImportArgsGet is the WebAssembly 1.0 (MVP) Text format import of FunctionArgsGet
+	ImportArgsGet = `(import "wasi_snapshot_preview1" "args_get"
```
mathetake, line 28: how this text format will be used by end user?

codefromthecrypt: I can move it internal, but at some point we need to describe what we support, and without this, there's just a link and a function name. I had imagined this describes literally what we test vs the N interpretations of imports out there.

codefromthecrypt:
I will move SectionID and ValueType internal. I'm sure users will need to know error codes for WASI as there was maintenance on those before.

internal/wasm/binary/export.go
```diff
@@ -5,8 +5,9 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/tetratelabs/wazero/wasm"
-	"github.com/tetratelabs/wazero/wasm/internal/leb128"
+	"github.com/tetratelabs/wazero/internal/leb128"
+	wasm "github.com/tetratelabs/wazero/internal/wasm"
+	wasm2 "github.com/tetratelabs/wazero/wasm"
```
mathetake, line 10: lol wasm2

internal/wasm/host.go
```diff
@@ -0,0 +1,390 @@
...
+	GoFunc       *reflect.Value
+}
+
+// GetFunctionType returns the function type corresponding to the function signature or errs if invalid.
+func GetFunctionType(name string, fn *reflect.Value) (*FunctionType, error) {
+	if fn.Kind() != reflect.Func {
+		return nil, fmt.Errorf("%s value is not a reflect.Func: %s", name, fn.String())
+	}
+	p := fn.Type()
+	if p.NumIn() == 0 { // TODO: actually check the type
```
mathetake, line 25: yeah actually this is my TODO :D

internal/wasm/host.go
```diff
@@ -0,0 +1,390 @@
...
+	"math"
+	"reflect"
+
+	"github.com/tetratelabs/wazero/wasm"
+)
+
+type HostFunction struct {
+	Name         string
+	FunctionType *FunctionType
+	GoFunc       *reflect.Value
```
mathetake, line 16: I guess these can be unexported?

internal/wasm/jit/engine.go
```diff
@@ -277,7 +276,7 @@ func (s jitCallStatusCode) causePanic() {
 	case jitCallStatusCodeMemoryOutOfBounds:
 		err = wasm.ErrRuntimeOutOfBoundsMemoryAccess
 	case jitCallStatusCodeInvalidTableAccess:
-		err = wasm.ErrRuntimeInvalidTableAcces
+		err = wasm.ErrRuntimeInvalidTableAccess
```
mathetake, line 279: thanks!

internal/wasm/host.go
codefromthecrypt: I can take a look. In general post this change I think a lot of code can be refactored, but export matters less since internal.

README.md
```diff
@@ -56,9 +45,9 @@ There are two types of engines are available for wazero, and you have to choose
 
 Both of engines passes 100% of [WebAssembly spec test suites]((https://github.com/WebAssembly/spec/tree/wg-1.0/test/core)) (on supported platforms).
 
-| Engine     | Usage|GOARCH=amd64 | GOARCH=others | 
+| Engine     | Usage|GOARCH=amd64 | GOARCH=others |
 |:----------:|:---:|:-------------:|:------:|
-| Interpreter|`interpreter.NewEngine()`| ✅    | ✅ | 
+| Interpreter|`interpreter.NewEngine()`| ✅    | ✅ |
 | JIT engine |`jit.NewEngine()`|   ✅   | ❌  |
```
mathetake, line 51: I think these lines on Usage have drifted

codefromthecrypt: cool will fix

mathetake:
OK agreed this direction and I will approve once the planed change has pushed

wasi.go
```diff
@@ -0,0 +1,110 @@
...
+// or function call error.
+//
+// Prerequisites of the "Current Unstable ABI" from wasi.ModuleSnapshotPreview1:
+// * "_start" is an exported nullary function and does not export "_initialize"
+// * "memory" is an exported memory.
+//
+// Note: Exporting "__indirect_function_table" is mentioned as required, but not enforced.
+// Note: All TinyGo Wasm are WASI commands. They initialize memory on "_start" and import "fd_write" to implement panic.
+// See https://github.com/WebAssembly/WASI/blob/snapshot-01/design/application-abi.md#current-unstable-abi
+func StartWASICommand(store *Store, module *Module) (wasm.ModuleFunctions, error) {
```
nullpo-head, line 94: It may be worth mentioning in the comment for users that a wasi command may assume that none of their exports are accessed after `_start` returned according to [the WASI spec](https://github.com/WebAssembly/WASI/blob/snapshot-01/design/application-abi.md#current-unstable-abi).

nullpo-head (approved):
Thanks a lot for your hard and awesome work!

wasi.go
codefromthecrypt: I can add a comment like the other that we aren't choosing to restrict exports

nullpo-head: sounds nice.

codefromthecrypt: // Note: The wasm.ModuleFunctions return value does not restrict exports after "_start" as allowed in the specification.

mathetake (approved):

mathetake:
maybe we could also make every names in wasi/functions.go ?

internal/wasm/host.go
codefromthecrypt: indeed! unexpected

codefromthecrypt: unexported I meant

codefromthecrypt:
> maybe we could also make every names in wasi/functions.go internal ?

sure

codefromthecrypt:
thanks for all the feedback especially giving 100pct attention @mathetake @nullpo-head as that avoided mutual drift brutality

PR #239, opened 2022-02-14
Adds CONTRIBUTING template

codefromthecrypt:
we forgot this some time ago. This is exactly the same as func-e

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,70 @@
+# Contributing
+
+We welcome contributions from the community. Please read the following guidelines carefully to maximize the chances of your PR being merged.
+
+## Coding Style
+
+- To ensure your change passes format checks, use run `make check`. To format your files, you can run `make format`.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
+# Contributing
+
+We welcome contributions from the community. Please read the following guidelines carefully to maximize the chances of your PR being merged.
+
+## Coding Style
+
+- The code is linted using a relatively stringent [golang-ci config](./.golangci.yml). To run this linter (and a few others) use run `make check`. To format your files, you can run `make format`.
```
mathetake, line 7: we  don't have `.golangci.yml` :D

mathetake (approved):

CONTRIBUTING.md
codefromthecrypt: I keep forgetting hitting "resolve" doesn't add the pending comments typed! sorry, I meant to ack that this is now done.

PR #241, opened 2022-02-14
wasi: Implement `random_get` function

r8d8:
Fixed sign-off from original PR #237


wasi/wasi.go
```diff
@@ -154,6 +172,17 @@ const (
 	wasiSnapshotPreview1Name = "wasi_snapshot_preview1"
 )
 
+type RandomSource interface {
```
mathetake, line 175: No need to export

wasi/wasi.go
```diff
@@ -154,6 +172,17 @@ const (
...
+	Read([]byte) (int, error)
+}
+
+// Non-deterministic random source using crypto/rand
+type CryptoRandomSource struct{}
```
mathetake, line 180: Same

wasi/wasi.go
```diff
@@ -503,6 +535,23 @@ func (a *api) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32) (err Errno)
 	return ErrnoSuccess
 }
 
+// RandomGet implements API.RandomGet
+func (a *api) RandomGet(ctx *wasm.HostFunctionCallContext, buf uint32, bufLen uint32) (errno Errno) {
+	if !ctx.Memory.ValidateAddrRange(buf, uint64(bufLen)) {
+		return ErrnoInval
+	}
+
+	random_bytes := make([]byte, bufLen)
```
mathetake, line 544: We never use snake case for variables in Go. Could you take a look at https://go.dev/doc/effective_go throughly?

wasi/wasi.go
mathetake: We do the same for clock so could you reference that and please make the aligned

wasi/wasi.go
mathetake: And plus returning int can be removed

wasi/wasi.go
```diff
@@ -503,6 +535,23 @@ func (a *api) fd_close(ctx *wasm.HostFunctionCallContext, fd uint32) (err Errno)
...
+	_, err := a.randSource.Read(random_bytes)
+	if err != nil {
+		return ErrnoInval
```
mathetake, line 547: ErronoInval is for invalid arguments right? Is this the correct error for here?

mathetake: I would recommend take a look at spec and also WASI-libc and see how other impl handle errors

wasi/wasi.go
```diff
@@ -355,7 +386,8 @@ func newAPI(opts ...Option) *api {
 }
 
 func (a *api) randUnusedFD() uint32 {
-	fd := uint32(rand.Int31())
```
mathetake, line 358: Could you revert the unnecessary change?

wasi/wasi.go
r8d8: * Wasmer return ErrnoIO for this case
https://github.com/wasmerio/wasmer/blob/53c2c29bed1cf1fee8c5616632ee8c2bccc27880/lib/wasi/src/syscalls/mod.rs#L2541

* wasi-libc just cast error fron syscal, and it can return bunch of them
https://man7.org/linux/man-pages/man2/getrandom.2.html

* wasmtime also maps syscal error into Errno
https://github.com/bytecodealliance/wasmtime/blob/853a025613e012c6c29002ddcccfced67073a8d0/crates/wasi-common/src/snapshots/preview_1.rs#L111

codefromthecrypt: FYI for bad memory offsets, we discussed offline to use ErrnoFault. Main thing is to document what is used and be consistent even if no one else is :D

Later, people can argue validly with these choices and probably we need to raise an issue on the wasi spec as it is underspecified. Main thing now is to document what you do.

r8d8: @codefromthecrypt thanks for info.
So, for invalid `buf`, `bufLen`  return error is `ErrnoFault`. Am I getting right?
And how to deal with errors from random source ?

codefromthecrypt: I think ErrnoIO for the source is fine for now. we can change it easily later if we find a most standard code

wasi/testdata/random.wat
```diff
@@ -0,0 +1,24 @@
+;; This is a wat file to just export clock WASI API to the host environment for testing the APIs.
```
codefromthecrypt, line 1: PS you can delete this file now

wasi/wasi_test.go
```diff
@@ -305,7 +306,73 @@ func TestAPI_ClockTimeGet_Errors(t *testing.T) {
...
+
+	for _, tt := range tests {
+		tc := tt
+
+		t.Run(tc.name, func(t *testing.T) {
+			ret, _, err := store.CallFunction(context.Background(), "test", FunctionRandomGet, uint64(tc.buf), uint64(tc.bufLen))
+			require.NoError(t, err)
+			require.Equal(t, uint64(ErrnoInval), ret[0]) // ret[0] is returned errno
+		})
+	}
```
codefromthecrypt, line 373: after this you can do a one-off for the IOError, something like

```
t.Run("returns ErrnoXXX on random source err", func...
wasiAPI.(*api).randSource = func(p []byte) error {
  return errors.New("random error")
}
require...
``

codefromthecrypt (commented):
nearly there!

codefromthecrypt:
drift alert and apologies for it (again *blush*) https://github.com/tetratelabs/wazero/pull/246

if you need help resolving conflicts ask and I can do that for you.

wasi/wasi_test.go
r8d8: Got it. Done

codefromthecrypt (approved):
Thanks for the great work.

r8d8:
Thanks for patience  @codefromthecrypt @mathetake

mathetake (approved):

mathetake:
Thank you for you contribution! 💯🙌🏄‍♂️ @r8d8

wasi/wasi.go
codefromthecrypt: I found the issue that has been punted quite a while so far https://github.com/WebAssembly/WASI/issues/215

r8d8: Error codes stuff gets you confused even more in functions that a more compilated than `random_get`. Spec need to address it.

codefromthecrypt: totally agree. appreciate if you can make your case on the spec issue because it was 2 years already so somehow they aren't prioritizing. maybe if more scream they will 🤷

PR #242, opened 2022-02-15
Uses importing module memory for host function calls

### LINKED ISSUE #195: Host functions called from another host function cannot know the right MemoryInstance

codefromthecrypt:
Fixes #195


codefromthecrypt:
I raised for review while I backfill unit tests

mathetake:
cool!

tests/engine/adhoc_test.go
```diff
@@ -201,31 +202,40 @@ func recursiveEntry(t *testing.T, newEngine func() wasm.Engine) {
...
 
-	// We should be able to call the exported add_int and it should add two ints.
-	results, _, err := store.CallFunction(ctx, "test", "add_int", uint64(12), uint64(30))
+	// We should be able to call the exported store_int and it should store two ints.
+	results, _, err := store.CallFunction(ctx, "test", "store_int", 1, math.MaxUint64)
 	require.NoError(t, err)
-	require.Equal(t, uint64(42), results[0])
+	require.Equal(t, wasi.ErrnoSuccess, wasi.Errno(results[0]))
+
+	require.Equal(t, []byte{0x0, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x0}, store.Memories[0].Buffer[0:10])
```
mathetake, line 238: maybe comment on from what we have 0xff, 0xff

mathetake (approved):
Thank you for fixing (my) bug!

tests/engine/adhoc_test.go
codefromthecrypt: sure

codefromthecrypt:
Changed to make WASI tests test the constants mentioned in docs (notably the import is now a constant instead of embedded godoc)

tests/engine/adhoc_test.go
```diff
@@ -201,31 +202,41 @@ func recursiveEntry(t *testing.T, newEngine func() wasm.Engine) {
...
+		;; store_int is imported from the environment, but it's also exported back to the environment
+		(export "store_int" (func $store_int))
 		)`))
 	require.NoError(t, err)
 
 	store := wasm.NewStore(newEngine())
 
-	addInt := func(ctx *wasm.HostFunctionCallContext, x int32, y int32) int32 {
-		return x + y
+	storeInt := func(ctx *wasm.HostFunctionCallContext, offset uint32, val uint64) wasi.Errno {
```
mathetake, line 220: I don't see any reason to use wasi.Errono.. simply use uint32?

codefromthecrypt: ok sure

mathetake (approved):

codefromthecrypt:
FYI @nullpo-head @r8d8 this PR drifts your PRs in a good way as testing the wasi functions is easier (since you can use a constant for the imports instead of copy/pasta of a %.wat file)

codefromthecrypt:
Thanks for the review @mathetake and the hints on the issue and offline @nullpo-head

PR #243, opened 2022-02-15
jit(amd64): less instruction memory bounds check

mathetake:


codefromthecrypt (approved):
nice

PR #244, opened 2022-02-15
jit(amd64): less instructions for memory bounds check: part 2

mathetake:

### DIFF
--- a/wasm/jit/jit_amd64.go
+++ b/wasm/jit/jit_amd64.go
@@ -3930,7 +3930,7 @@ func (c *amd64Compiler) compileLoad(o *wazeroir.OperationLoad) error {
 		targetSizeInBytes = 64 / 8
 	}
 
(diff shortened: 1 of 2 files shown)


wasm/jit/jit_amd64.go
```diff
@@ -3996,6 +3999,7 @@ func (c *amd64Compiler) compileLoad8(o *wazeroir.OperationLoad8) error {
 	moveFromMemory.To.Reg = reg
 	moveFromMemory.From.Type = obj.TYPE_MEM
 	moveFromMemory.From.Reg = reservedRegisterForMemory
+	moveFromMemory.From.Offset = -targetSizeInBytes
```
codefromthecrypt, line 4002: negatives are usually odd, so you can trailing comment why this is negative? ex. `because this is accessed as memory.Buffer[ceil-targetSizeInBytes: ceil]`

same everywhere with what you decide

codefromthecrypt (approved):
just one note as I found it surprising until I read other source docs later

PR #246, opened 2022-02-15
Adds host function interfaces

### LINKED ISSUE #204: Make wasm.HostFunctionCallContext an interface

codefromthecrypt:
This adds two interfaces with only one implementation each, used to
decouple end-user code from internals.

* wasm.HostFunctionCallContext - allows access to go Context and Memory
* wasm.Memory - read/write standard types and arbitrary data

This currently ports existing code to use the interfaces. It is true
that over time we can use structs instead of interfaces internally to
save a pointer reference in WASI call sites. However, I'm choosing not
to do this right now, as it is more important to see the interfaces
work. We should defer optimizing towards structs until after we
repackage implementations under the "internal" directory.

Fixes #204

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -1,0 +1,69 @@
+// Package api only includes constants and interfaces. This ensures no package cycles in implementation.
+package api
+
+import "context"
+
+// HostFunctionCallContext is the first argument of all host functions.
+type HostFunctionCallContext interface {
+	// Context returns the host call's context.
+	//
+	// The returned context is always non-nil; it defaults to the background context.
+	Context() context.Context
+
+	// Memory allows restricted access to an importing module's memory during a host function call.
+	Memory() Memory
+}
+
+// Memory allows restricted access to a module's memory. Notably, this does not allow growing.
+//
+// Note: This includes all value types available in WebAssembly 1.0 (MVP) and all are encoded little-endian.
+// See https://www.w3.org/TR/wasm-core-1/#storage%E2%91%A0
+type Memory interface {
+	// Len returns the size in bytes available. Ex. If the underlying memory has 1 page: 65536
+	//
+	// Note: this will not grow during a host function call, even if the underlying memory can.  Ex. If the underlying
+	// memory has min 0 and max 2 pages, this returns zero.
+	Len() uint32
+
+	// ReadUint32Le reads a uint32 in little-endian encoding from the underlying buffer at the offset in or returns
+	// false if out of range.
+	ReadUint32Le(offset uint32) (uint32, bool)
+
+	// ReadFloat32Le reads a float32 from 32 IEEE 754 little-endian encoded bits in the underlying buffer at the offset
+	// or returns false if out of range.
+	// See math.Float32bits
+	ReadFloat32Le(offset uint32) (float32, bool)
+
+	// ReadUint64Le reads a uint64 in little-endian encoding from the underlying buffer at the offset or returns false
+	// if out of range.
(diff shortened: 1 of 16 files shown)


wasi/wasi.go
```diff
@@ -52,12 +53,13 @@ type API interface {
...
 	// See https://en.wikipedia.org/wiki/Null-terminated_string
-	ArgsGet(ctx *wasm.HostFunctionCallContext, argv, argvBuf uint32) Errno
+	ArgsGet(ctx api.HostFunctionCallContext, argv, argvBuf uint32) Errno
 
 	// ArgsSizesGet is a WASI function that reads command-line argument data (Args) sizes.
 	//
-	// There are two result parameters: these are offsets in the wasm.HostFunctionCallContext Memory Buffer to write
-	// corresponding sizes in uint32 little-endian encoding.
+	// There are two result parameters: these are offsets in the api.HostFunctionCallContext Memory to write
+	// corresponding sizes in uint32 little-endian encoding. If either are invalid due to memory constraints, this
```
codefromthecrypt, line 61: fyi @r8d8, this is the sort of comment I'm adding about errno choices. you can make one similarly, and if we need to change later we can update 

> If either are invalid due to memory constraints, this returns ErrnoFault.

examples/simple_test.go
```diff
@@ -25,7 +26,7 @@ func Test_Simple(t *testing.T) {
 	store := wasm.NewStore(interpreter.NewEngine())
 
 	stdout := new(bytes.Buffer)
-	hostFunction := func(_ *wasm.HostFunctionCallContext) {
+	hostFunction := func(_ api.HostFunctionCallContext) {
```
mathetake, line 29:
```suggestion
	hostFunction := func(api.HostFunctionCallContext) {
```

wasm/host.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Len implements api.HostFunctionCallContext Len
+func (m *MemoryInstance) Len() uint32 {
+	return uint32(len(m.Buffer))
+}
+
+// hasLen returns true if Len is sufficient for sizeInBytes at the given offset.
+func (m *MemoryInstance) hasLen(offset uint32, sizeInBytes uint32) bool {
+	return uint64(offset+sizeInBytes) <= uint64(m.Len()) // uint64 prevents overflow on add
```
mathetake, line 41: nice for overflow prevention

mathetake (approved):
One thing I would add is naming -- maybe putting top level of repo and wazero.HostFunctionCallContext, etc.. together with other future user APIa

api/host.go
```diff
@@ -0,0 +1,69 @@
+// Package api only includes constants and interfaces. This ensures no package cycles in implementation.
```
mathetake, line 1: oh I see nvm the comment ^^

codefromthecrypt: incidentally I think we can temporarily put this in the "wasm" package

codefromthecrypt: changed to wasm package for now as I think even with top-level entry points in the wazero package we'll still be able to keep packages for things like constants if we choose to. I'll show what I mean post-merge in #238

PR #254, opened 2022-02-16
jit(arm64): support for clz, ctz, popcnt, div, rem

mathetake:
part of #187

### DIFF
--- a/wasm/jit/jit_arm64.go
+++ b/wasm/jit/jit_arm64.go
@@ -40,17 +40,21 @@ type archContext struct {
 	// executing "ret" instruction with this value. See arm64Compiler.exit.
 	jitCallReturnAddress uint64
 
-	// Used for overflow checks on division operations.
+	// Loading large constants in arm64 is a bit costly so we place the following
+	// consts on engine struct so that we can quickly access them during various operations.
+
+	// minimum32BitSignedInt is used for overflow check for 32-bit signed division.
 	minimum32BitSignedInt int32
+	// minimum64BitSignedInt is used for overflow check for 32-bit signed division.
 	minimum64BitSignedInt int64
 }
 
 const (
 	// engineArchContextJITCallReturnAddressOffset is the offset of archContext.jitCallReturnAddress in engine.
 	engineArchContextJITCallReturnAddressOffset = 136
-	// engineArchContextMinimum32BitSignedIntAddressOffset is the offset of archContext.minimum32BitSignedIntAddress in engine.
+	// engineArchContextMinimum32BitSignedIntOffset is the offset of archContext.minimum32BitSignedIntAddress in engine.
 	engineArchContextMinimum32BitSignedIntOffset = 144
-	// engineArchContextMinimum64BitSignedIntAddressOffset is the offset of archContext.minimum64BitSignedIntAddress in engine.
+	// engineArchContextMinimum64BitSignedIntOffset is the offset of archContext.minimum64BitSignedIntAddress in engine.
 	engineArchContextMinimum64BitSignedIntOffset = 152
 )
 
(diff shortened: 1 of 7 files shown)


wasm/jit/jit_arm64.go
```diff
@@ -1357,11 +1357,68 @@ func (c *arm64Compiler) compileMul(o *wazeroir.OperationMul) error {
...
+		c.compileRegisterToRegisterInstruction(arm64.ACLZW, reg, reg)
+	} else {
+		c.compileRegisterToRegisterInstruction(arm64.ARBIT, reg, reg)
+		c.compileRegisterToRegisterInstruction(arm64.ACLZ, reg, reg)
+	}
+	c.locationStack.pushValueLocationOnRegister(reg)
+	return nil
 }
 
 func (c *arm64Compiler) compilePopcnt(o *wazeroir.OperationPopcnt) error {
```
mathetake, line 1424: note to myself: https://godbolt.org/z/hf3aM5Td1

mathetake:
ok done with div, the left is rem

mathetake:
ok done... completed it 10x faster than amd64 😄

wasm/jit/jit_arm64.go
```diff
@@ -25,17 +25,38 @@ import (
...
+
 // archContext is embedded in Engine in order to store architecture-specific data.
 type archContext struct {
 	// jitCallReturnAddress holds the absolute return address for jitcall.
 	// The value is set whenever jitcall is executed and done in jit_arm64.s
 	// Native code can return back to the engine.exec's main loop back by
 	// executing "ret" instruction with this value. See arm64Compiler.exit.
 	jitCallReturnAddress uint64
+
+	// Loading large constants in arm64 is a bit costly so we place the following
```
nullpo-head, line 43: I guess "costly" means the fact arm64 doesn't have a straight-forward way to move a 32-bit/64-bit immediate to a register in general. On the other hand, because the constants are minimum int32 (1 << 31) and minimum int64 (1 << 63), isn't it possible to move `1` to a register and shift it instead of reading memory?

nullpo-head (approved):

wasm/jit/jit_arm64.go
mathetake: oh  yeah lemme do some benchmarks on it

mathetake: So a bit surprisingly but the benchmark shows that

```
MOV tmRegister, $1
CMP dividend, (tmpRegister LSL #63)
```

is slower than the const load by using this constant location.

```
BenchmarkAAA
BenchmarkAAA/const_load_32-bit
BenchmarkAAA/const_load_32-bit-10               11085193               101.9 ns/op           240 B/op          4 allocs/op
BenchmarkAAA/const_load_64-bit
BenchmarkAAA/const_load_64-bit-10               11745105               102.8 ns/op           240 B/op          4 allocs/op
BenchmarkAAA/load_shift_32-bit
BenchmarkAAA/load_shift_32-bit-10               11373872               103.4 ns/op           240 B/op          4 allocs/op
BenchmarkAAA/load_shift_64-bit
BenchmarkAAA/load_shift_64-bit-10               11382826               104.3 ns/op           240 B/op          4 allocs/op
PASS
```

so I will leave it as-is. Anyway thanks!

codefromthecrypt: maybe add a footnote what you mean by costly wrt "is slower than the const load by using this constant location."

wasm/jit/jit_arm64.go
```diff
@@ -272,6 +293,19 @@ func (c *arm64Compiler) compileTwoRegistersToRegisterInstruction(instruction obj
 	c.addInstruction(inst)
 }
 
+// compileTwoRegistersToRegisterInstruction adds an instruction which takes two source and destination register operands.
+func (c *arm64Compiler) compileTwoRegistersToTwoRegisterInstruction(instruction obj.As, src1, src2, dst1, dst2 int16) {
```
codefromthecrypt, line 297: maybe simpler as compileTwoRegisterInstruction

codefromthecrypt (approved):
now we can divide by zero even on arm64!

wasm/jit/jit_arm64.go
mathetake: added!

mathetake:
Thanks for the look!

PR #256, opened 2022-02-17
doc: fix drifts

mathetake:


mathetake:
merging as this is harmless

PR #260, opened 2022-02-18
Removes requirement to pass a HostFunctionCallContext

codefromthecrypt:
This allows users to decouple from wazero code when authoring host
functions. Notably, this allows them to opt out of using a context, or
only using a Go context instead of HostFunctionCallContext.

This backfills docs on how to write host functions (in simple terms).

Finally, this does not optimize engines to avoid propagating context or
looking up memory if it would never be used. That could be done later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,8 @@ func main() {
 	// Decode the binary as WebAssembly module.
 	mod, _ := wazero.DecodeModuleBinary(source)
 
-	// Initialize the execution environment called "store" with Interpreter-based engine.
-	store := wazero.NewStore()
-
-	// Instantiate the module, which returns its exported functions
-	functions, _ := store.Instantiate(mod)
+	// Instantiate the module with a Wasm Interpreter, to return its exported functions
+	functions, _ := wazero.NewStore().Instantiate(mod)
 
 	// Get the factorial function
 	fac, _ := functions.GetFunctionI64Return("fac")
--- a/internal/wasi/wasi.go
+++ b/internal/wasi/wasi.go
@@ -113,7 +113,7 @@ const (
 	FunctionPathUnlinkFile       = "path_unlink_file"
 	FunctionPollOneoff           = "poll_oneoff"
 
-	// ProcExit terminates the execution of the module with an exit code.
+	// FunctionProcExit terminates the execution of the module with an exit code.
 	// See https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#proc_exit
 	FunctionProcExit = "proc_exit"
 
@@ -339,7 +339,7 @@ type SnapshotPreview1 interface {
 	//
 	// Note: ImportProcExit shows this signature in the WebAssembly 1.0 (MVP) Text Format.
 	// See https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#proc_exit
-	ProcExit(ctx wasm.HostFunctionCallContext, rval uint32)
+	ProcExit(rval uint32)
 
 	// TODO: ProcRaise
 	// TODO: SchedYield
@@ -505,8 +505,8 @@ func (a *wasiAPI) ClockTimeGet(ctx wasm.HostFunctionCallContext, id uint32, prec
 	return wasi.ErrnoSuccess
 }
 
-// ProcExit implements API.ProcExit
-func (a *wasiAPI) ProcExit(ctx wasm.HostFunctionCallContext, exitCode uint32) {
+// ProcExit implements SnapshotPreview1.ProcExit
+func (a *wasiAPI) ProcExit(exitCode uint32) {
 	// Panic in a host function is caught by the engines, and the value of the panic is returned as the error of the CallFunction.
 	// See the document of API.ProcExit.
 	panic(wasi.ExitCode(exitCode))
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -19,11 +19,8 @@ func main() {
 	// Decode the binary as WebAssembly module.
 	mod, _ := wazero.DecodeModuleBinary(source)
 
-	// Initialize the execution environment called "store" with Interpreter-based engine.
-	store := wazero.NewStore()
-
-	// Instantiate the module, which returns its exported functions
-	functions, _ := store.Instantiate(mod)
+	// Instantiate the module with a Wasm Interpreter, to return its exported functions
+	functions, _ := wazero.NewStore().Instantiate(mod)
```
codefromthecrypt, line 23: might as well reduce step count

internal/wasi/wasi.go
```diff
@@ -339,7 +339,7 @@ type SnapshotPreview1 interface {
 	//
 	// Note: ImportProcExit shows this signature in the WebAssembly 1.0 (MVP) Text Format.
 	// See https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#proc_exit
-	ProcExit(ctx wasm.HostFunctionCallContext, rval uint32)
```
codefromthecrypt, line 342: not needed!

store.go
```diff
@@ -20,10 +20,44 @@ func NewEngineJIT() *Engine { // TODO: compiler?
 }
 
 // HostFunctions are functions written in Go, which a WebAssembly Module can import.
+//
```
codefromthecrypt, line 23: Important docs here!

codefromthecrypt (commented):
Notes

internal/wasm/host.go
```diff
@@ -10,42 +10,131 @@ import (
...
-		if t, ok := getTypeOf(kind); !ok {
-			return nil, fmt.Errorf("%s param[%d] is unsupported: %s", name, i, kind.String())
-		} else {
-			paramTypes[i] = t
+	pOffset := 0
+	pCount := p.NumIn()
+	fk = FunctionKindHost
+	if pCount > 0 && p.In(0).Kind() == reflect.Interface {
+		p0 := p.In(0)
+		if hc := p0.Implements(hostFunctionCallContextType); hc {
```
nullpo-head, line 80: Is there any reason to define `hc` (and `gc` below), instead of just doing `if p0.Implements`?

internal/wasm/host.go
```diff
@@ -10,42 +10,131 @@ import (
...
+	for i := 0; i < len(ft.Params); i++ {
+		pI := p.In(i + pOffset)
+		if t, ok := getTypeOf(pI.Kind()); ok {
+			ft.Params[i] = t
+			continue
+		}
+
+		// Now, we will definitely err, decide which message is best
+		var arg0Type reflect.Type
+		if hc := pI.Implements(hostFunctionCallContextType); hc {
```
nullpo-head, line 109: Is there any reason to define `hc` (and `gc` below), instead of just doing `if p0.Implements`?

internal/wasm/jit/engine.go
```diff
@@ -450,22 +450,25 @@ const (
...
 	in := make([]reflect.Value, tp.NumIn())
 
 	// We pop the value and pass them as arguments in a reverse order according to the
 	// stack machine convention.
-	for i := len(in) - 1; i >= 1; i-- {
+	wasmParamOffset := 0
+	if fk != wasm.FunctionKindHost {
+		wasmParamOffset = 1
+	}
+	for i := len(in) - 1; i >= wasmParamOffset; i-- {
```
mathetake, line 471: so this and the `GetHostFunctionCallContextValue` below are the additional codes on the host call's runtime pass right? TODO comment would be appreciated for future performance optimizations

nullpo-head (approved):
Nice to see now the signature of a host function looks more natural as a Golang function

internal/wasm/jit/engine.go
codefromthecrypt: so the fk is compared twice.  here to determine if there is an offset or not, and inside GetHostFunctionCallContextValue to determine what to set. so this is read twice vs the code that existed before and if reading a number twice is a perf hit we are concerned with we can optimize!

internal/wasm/host.go
codefromthecrypt: nope what you mention is better!

internal/wasm/jit/engine.go
codefromthecrypt: I made a general note instead as I think this was too specific, and can be handled in a broader scope inside that same function and ideally get better perf regardless of context. we'll see..

internal/wasm/host.go
```diff
@@ -10,42 +10,131 @@ import (
...
 )
 
+// FunctionKind identifies the type of function that can be called.
+type FunctionKind byte
+
+const (
+	// FunctionKindWasm is not a host function: it is implemented in Wasm.
+	FunctionKindWasm FunctionKind = iota
+	// FunctionKindHost is a function implemented in Go, with a signature matching FunctionType.
+	FunctionKindHost
```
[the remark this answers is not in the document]
codefromthecrypt: ok sure

codefromthecrypt:
thanks for the help. It is better now!

PR #264, opened 2022-02-19
Adds wazero.MakeWasmFunc and MakeHostFunc

codefromthecrypt:
This adds `wazero.MakeWasmFunc` and `MakeHostFunc` which allow
users to avoid knowing how to encode parameters or results.

```
var addFloat func(float64, float64) float64
_ = wazero.MakeWasmFunc(exports, name, &addFloat)
res := addFloat(1.5, 1.5) // 3.0!
```

This should help avoid problems and make the API easier to deal with.
Someone can also use the export API to make a codegen alternative to
MakeWasmFunc!

Note: the name `MakeWasmFunc` is similar to `reflect.MakeFunc` and
that's how it is currently implemented.

`MakeHostFunc` is also available for the same reason in host functions.

### DIFF
--- a/internal/makefunc/makefunc.go
+++ b/internal/makefunc/makefunc.go
@@ -1,0 +1,148 @@
+package makefunc
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	wasm "github.com/tetratelabs/wazero/internal/wasm"
+	publicwasm "github.com/tetratelabs/wazero/wasm"
+)
+
+var uint64Zero = reflect.Zero(reflect.TypeOf(uint64(0)))
+var errorType = reflect.TypeOf((*error)(nil)).Elem()
+
+// MakeWasmFunc implements the goFuncPtr to call the provided wasmFunction.
+//
+// See reflect.MakeFunc
+// TODO: Maybe we can optimize our own variant of reflect.MakeFunc in JIT and add this to the Engine interface.
+func MakeWasmFunc(ctx *wasm.ModuleContext, name string, impl *wasm.FunctionInstance, goFuncPtr interface{}) error {
+	// goFuncPtr should point to nil if used properly, but we don't care as we clobber it.
+	fn := reflect.ValueOf(goFuncPtr).Elem()
+
+	// Ensure the signature is correct and if it has an error result or not.
+	fk, ft, hasErrorResult, err := wasm.GetFunctionType(name, &fn, true)
+	if err != nil {
+		return err
+	}
+
+	cf := &wasmFunc{
+		ctx:                  ctx,
+		impl:                 impl,
+		goFuncKind:           fk,
+		goFuncResultCount:    uint32(len(ft.Results)),
+		goFuncHasErrorResult: hasErrorResult,
+	}
+
+	if cf.goFuncResultCount == 1 {
+		cf.goFuncResultKind = fn.Type().Out(0).Kind()
+	}
+	if cf.goFuncHasErrorResult {
+		cf.goFuncResultCount++
+	}
+
+	// Now that we know the type is compatible with Wasm, make a reflective invoker.
+	v := reflect.MakeFunc(fn.Type(), cf.makeFunc)
+
+	// Set the pointer value to the implementation that calls the Wasm function.
+	fn.Set(v)
+	return nil
+}
+
+type wasmFunc struct {
+	ctx                  *wasm.ModuleContext
+	impl                 *wasm.FunctionInstance
+	goFuncKind           wasm.FunctionKind
+	goFuncResultKind     reflect.Kind
+	goFuncResultCount    uint32
+	goFuncHasErrorResult bool
+}
+
+func (f *wasmFunc) makeFunc(args []reflect.Value) (results []reflect.Value) {
+	wasmParamOffset := 1
+	ctx := f.ctx
+	switch f.goFuncKind {
+	case wasm.FunctionKindGoNoContext: // no special param zero
+		wasmParamOffset = 0
+	case wasm.FunctionKindGoContext:
+		ctx = ctx.WithContext(args[0].Interface().(context.Context))
+	case wasm.FunctionKindGoModuleContext:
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
Added benchmark which go as expected (most logic is moved to init, so that takes the biggest hit vs runtime invocation of function calls).

The TL;DR; is there is an overhead to using reflection. Since we avail both APIs, users can make a judgement call on the tradeoff of convenience vs perf!

```
goos: darwin
goarch: amd64
pkg: github.com/tetratelabs/wazero/tests/bench
BenchmarkFacIter_Init
BenchmarkFacIter_Init/Interpreter_GetFunction
BenchmarkFacIter_Init/Interpreter_GetFunction-16         	   59826	     18773 ns/op
BenchmarkFacIter_Init/Interpreter_MakeFunc
BenchmarkFacIter_Init/Interpreter_MakeFunc-16            	   62368	     19479 ns/op
BenchmarkFacIter_Init/JIT_GetFunction
BenchmarkFacIter_Init/JIT_GetFunction-16                 	    5080	    250171 ns/op
BenchmarkFacIter_Init/JIT_MakeFunc
BenchmarkFacIter_Init/JIT_MakeFunc-16                    	    4263	    265755 ns/op
BenchmarkFacIter_Init/wasmer-go
BenchmarkFacIter_Init/wasmer-go-16                       	    4070	    295139 ns/op
BenchmarkFacIter_Init/wasmtime-go
BenchmarkFacIter_Init/wasmtime-go-16                     	    3556	    284662 ns/op
BenchmarkFacIter_Invoke
BenchmarkFacIter_Invoke/Interpreter_GetFunction
BenchmarkFacIter_Invoke/Interpreter_GetFunction-16       	  729180	      1577 ns/op
BenchmarkFacIter_Invoke/Interpreter_MakeFunc
BenchmarkFacIter_Invoke/Interpreter_MakeFunc-16          	  627548	      1890 ns/op
BenchmarkFacIter_Invoke/JIT_GetFunction
BenchmarkFacIter_Invoke/JIT_GetFunction-16               	 7274833	       146.0 ns/op
BenchmarkFacIter_Invoke/JIT_MakeFunc
BenchmarkFacIter_Invoke/JIT_MakeFunc-16                  	 2998050	       432.2 ns/op
BenchmarkFacIter_Invoke/wasmer-go
BenchmarkFacIter_Invoke/wasmer-go-16                     	  782532	      1348 ns/op
BenchmarkFacIter_Invoke/wasmtime-go
BenchmarkFacIter_Invoke/wasmtime-go-16                   	  477218	      2489 ns/op
```

codefromthecrypt:
renamed to `MakeWasmFunc` as unlike `reflect.MakeFunc`, there's only one implementation: to call an exported Wasm function

codefromthecrypt:
switched to a utility to narrow the implementation after realizing we can still special-case based on engine provided `MakeWasmFunc` stays in the same repo

codefromthecrypt:
I want to rejig things slightly so that host functions can be used with this (easily).. will report back

codefromthecrypt:
converting back to draft until this is a plugin example!

wasm/wasm.go
```diff
@@ -109,3 +122,27 @@ type Memory interface {
...
+// EncodeF32 converts the input so that it can be used as a Function F32 parameter or result.
+// See DecodeF32
+func EncodeF32(input float32) uint64 {
+	return uint64(math.Float32bits(input))
+}
+
+// DecodeF32 converts the Function F32 parameter or result to a float32.
+// See DecodeF32
+func DecodeF32(input uint64) float32 {
+	return float32(math.Float64frombits(input))
```
mathetake, line 135: not sure, but I think this should be?

```suggestion
	return math.Float32frombits(uint32(input))
```

tests/spectest/spec_test.go
```diff
@@ -187,13 +187,13 @@ func (c command) expectedError() (err error) {
...
-	var printF64 = func(publicwasm.HostFunctionCallContext, float64) {}
-	var printI32F32 = func(publicwasm.HostFunctionCallContext, uint32, float32) {}
-	var printF64F64 = func(publicwasm.HostFunctionCallContext, float64, float64) {}
+	var printV = func(publicwasm.ModuleContext) {}
+	var printI32 = func(publicwasm.ModuleContext, uint32) {}
+	var printF32 = func(publicwasm.ModuleContext, float32) {}
+	var printI64 = func(publicwasm.ModuleContext, uint64) {}
+	var printF64 = func(publicwasm.ModuleContext, float64) {}
+	var printI32F32 = func(publicwasm.ModuleContext, uint32, float32) {}
+	var printF64F64 = func(publicwasm.ModuleContext, float64, float64) {}
```
mathetake, line 196: Let's simply remove 
```suggestion
	var printV = func() {}
	var printI32 = func(uint32) {}
	var printF32 = func(float32) {}
	var printI64 = func(uint64) {}
	var printF64 = func(float64) {}
	var printI32F32 = func(uint32, float32) {}
	var printF64F64 = func(float64, float64) {}
```

wasm/wasm.go
codefromthecrypt: didn't work in signed case! I owe backfilling of this test :D

codefromthecrypt:
pulled everything except reflection into this PR https://github.com/tetratelabs/wazero/pull/266

codefromthecrypt:
pared down the change and contemplating how to deal with the fact that Go can't currently implement interfaces. Probably, we need an example where struct functions are overridden as codegen is too much for one PR.

See https://medium.com/capital-one-tech/learning-to-use-go-reflection-822a0aed74b7

codefromthecrypt:
parking this for now as other word needs attention. To proceed this will be figuring out if there's an elegant pattern we can use for implementing plugins with wasm. function pointers are "ok" it would be nicer if we can overwrite struct functions, or at least use those as signatures. If we can't find anything better to group functions in a module together (in go) we can stick with `func` at a time as is implemented here

codefromthecrypt:
closing this as I'm not really sure this is worthwhile. We can revisit if something changes.

PR #266, opened 2022-02-21
Adds wasm.Store API to get exported module and host functions

codefromthecrypt:
This adds this interface `wasm.Store` which gives access to functions in
a store without leaking an API to change the store. This is primarily to
support configuration use cases where post-initialization, there's no
need or desire to mutate the store. This also backfills codecs needed to
handle float results.

```go
// Store allows access to instantiated modules and host functions
type Store interface {
	// ModuleExports returns exports from an instantiated module or false if it wasn't.
	ModuleExports(moduleName string) (ModuleExports, bool)

	// HostExports returns exported host functions for the moduleName or false there are none.
	HostExports(moduleName string) (HostExports, bool)
}
```

While `ModuleExports` are returned on `wazero.InstantiateModule`, they can also be accessed by module name later. Here's an example of decoupling where central configuration configures the store and plugins use it:
```go
// Config wiring can instantiate the store and locate the name of the function desired
exports, _ := store.ModuleExports("math")
addFloat, _ := exports.Function("AddFloat")

// That function could be used directly or implement a friendlier interface.
results, _ := addFloat(context.Background(), wasm.EncodeF64(1.5),
wasm.EncodeF64(1.5))
res := wasm.DecodeF64(result[0])) // 3.0!
```

This design also allows unwinding of function dependencies, ex where a host function needs to use another host function:
```go
	// built-in WASI function to write a random value to memory
	randomGet := func(ctx wasm.ModuleContext, buf, bufLen uint32) wasi.Errno {
		panic("unimplemented")
	}

	goFunc := func(ctx wasm.ModuleContext) {
		// Write 8 random bytes to memory using WASI.
		errno := randomGet(ctx, 0, 8)
	--snip--

	// Configure WASI and implement the function to use it
	we, err := wazero.ExportHostFunctions(store, wasi.ModuleSnapshotPreview1, wazero.WASISnapshotPreview1())
	require.NoError(t, err)
	randomGetFn, ok := we.Function("random_get")
	require.True(t, ok)

	// Implement the function pointer. This mainly shows how you can decouple a host function dependency.
	randomGet = func(ctx wasm.ModuleContext, buf, bufLen uint32) wasi.Errno {
		res, err := randomGetFn(ctx, uint64(buf), uint64(bufLen))
		require.NoError(t, err)
		return wasi.Errno(res[0])
	}
```

Ancillary changes:
* `NewStoreWithConfig()` no longer returns an error or takes host function config
* `HostFunctionCallContext` has been renamed to `ModuleContext`
* `wasi.Errno` was converted back to a uint32 alias to match mapped types

### DIFF
--- a/store.go
+++ b/store.go
@@ -1,6 +1,8 @@
 package wazero
 
 import (
+	"fmt"
+
 	internalwasm "github.com/tetratelabs/wazero/internal/wasm"
 	"github.com/tetratelabs/wazero/internal/wasm/interpreter"
 	"github.com/tetratelabs/wazero/internal/wasm/jit"
@@ -19,7 +21,39 @@ func NewEngineJIT() *Engine { // TODO: compiler?
 	return &Engine{e: jit.NewEngine()}
 }
 
-// HostFunctions are functions written in Go, which a WebAssembly Module can import.
+// StoreConfig allows customization of a Store via NewStoreWithConfig
+type StoreConfig struct {
+	// Engine defaults to NewEngineInterpreter
+	Engine *Engine
+}
+
+func NewStore() wasm.Store {
+	return internalwasm.NewStore(interpreter.NewEngine())
+}
+
+// NewStoreWithConfig returns a store with the given configuration.
+func NewStoreWithConfig(config *StoreConfig) wasm.Store {
+	engine := config.Engine
+	if engine == nil {
+		engine = NewEngineInterpreter()
+	}
+	return internalwasm.NewStore(engine.e)
+}
+
+// InstantiateModule instantiates the module namespace or errs if the configuration was invalid.
+func InstantiateModule(store wasm.Store, module *Module) (wasm.ModuleExports, error) {
+	internal, ok := store.(*internalwasm.Store)
+	if !ok {
+		return nil, fmt.Errorf("unsupported Store implementation: %s", store)
+	}
+	return internal.Instantiate(module.wasm, module.name)
+}
+
+// ExportHostFunctions adds functions written in Go, which a WebAssembly Module can import.
+//
+// Here is a description of parameters:
+// * moduleName is the module name that these functions are imported with. Ex. wasi.ModuleSnapshotPreview1
+// * nameToGoFunc map key is the name to export and the value is the func. Ex. WASISnapshotPreview1
 //
 // Noting a context exception described later, all parameters or result types must match WebAssembly 1.0 (MVP) value
 // types. This means uint32, uint64, float32 or float64. Up to one result can be returned.
@@ -30,7 +64,7 @@ func NewEngineJIT() *Engine { // TODO: compiler?
 //		return x + y
 //	}
 //
-// Host functions may also have an initial parameter (param[0]) of type context.Context or wasm.HostFunctionCallContext.
+// Host functions may also have an initial parameter (param[0]) of type context.Context or wasm.ModuleContext.
 //
 // Ex. This uses a Go Context:
 //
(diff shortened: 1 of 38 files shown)


wasm/wasm.go
```diff
@@ -109,3 +134,27 @@ type Memory interface {
...
+// EncodeF32 converts the input so that it can be used as a Function F32 parameter or result.
+// See DecodeF32
+func EncodeF32(input float32) uint64 {
+	return uint64(math.Float32bits(input))
+}
+
+// DecodeF32 converts the Function F32 parameter or result to a float32.
+// See DecodeF32
+func DecodeF32(input uint64) float32 {
+	return float32(math.Float64frombits(input))
```
mathetake, line 147: yeah this seems wrong to me again, and should be `math.Float32frombits(uint32(input))` otherwise something must be wrong..

store.go
```diff
@@ -19,7 +21,39 @@ func NewEngineJIT() *Engine { // TODO: compiler?
 	return &Engine{e: jit.NewEngine()}
 }
 
-// HostFunctions are functions written in Go, which a WebAssembly Module can import.
+// StoreConfig allows customization of a Store via NewStoreWithConfig
+type StoreConfig struct {
+	// Engine defaults to NewEngineInterpreter
+	Engine *Engine
```
mathetake, line 27: do we need to export?

mathetake: oh this is config! nvm

wasm/wasm.go
mathetake: https://github.com/tetratelabs/wazero/blob/cleaner-api/tests/spectest/spec_test.go#L79

mathetake: https://github.com/tetratelabs/wazero/blob/cleaner-api/tests/spectest/spec_test.go#L441-L442

wasi/wasi.go
```diff
@@ -41,17 +41,18 @@ type FS interface {
...
 // See https://linux.die.net/man/3/errno
-type Errno uint32
+type Errno = uint32 // alias for parity with internalwasm.ValueType
 
-// Error returns the POSIX error code name, except ErrnoSuccess, which isn't defined. Ex. Errno2big -> "E2BIG"
-func (err Errno) Error() string {
-	if int(err) < len(errnoToString) {
-		return errnoToString[err]
+// ErrnoName returns the POSIX error code name, except ErrnoSuccess, which is not an error. Ex. Errno2big -> "E2BIG"
+func ErrnoName(errno Errno) string {
```
nullpo-head, line 51: This change removes golang's `Error` interface from `Errno`. Is it intended? 
By the way, I think it's ok to remove `Error` interface from `Errno` anyway. Considering that in the future we may give host functions the ability to return an `Error` to trigger immediate termination, we will not want `Errno` to become `Error`, because it means even `ErrnoSuccess` triggers immediate termination.

nullpo-head (approved):
LGTM except the potential float encoding bug Takeshi pointed out.

wasi/wasi.go
codefromthecrypt: yeah this is intended as it isn't an error, as you mentioned. the other benefit is it keeps the function signature easy to implement reflectively (as you don't need to instantiate non-go type)

wasm/wasm.go
codefromthecrypt: changing this breaks tests as the sign bit is lost, so I'll have to change this and also the encode side..

store.go
codefromthecrypt: nope!

wasm/wasm.go
mathetake: this (`math.Float32frombits(uint32(input))`) itself shouldn't lose the signbit so something internal going wrong?

mathetake: ok found the flaw

mathetake: lemme

mathetake: https://github.com/tetratelabs/wazero/pull/268

mathetake: there was a flaw in engine :D

tests/engine/adhoc_test.go
```diff
@@ -241,41 +239,41 @@ func importedAndExportedFunc(t *testing.T, newEngine func() *wazero.Engine) {
...
+	require.Equal(t, uint64(0), results[0])
 
 	// Since offset=1 and val=math.MaxUint64, we expect to have written exactly 8 bytes, with all bits set, at index 1.
 	require.Equal(t, []byte{0x0, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x0}, memory.Buffer[0:10])
 }
 
-//  hostFunctions ensures arg0 is optionally a context, and fails if a float parameter corrupts a host function value
-func hostFunctions(t *testing.T, newEngine func() *wazero.Engine) {
+func TestHostFunctions(t *testing.T) {
+	testHostFunctions(t, wazero.NewEngineInterpreter)
```
mathetake, line 272: lets's run this for JIT as well :D

codefromthecrypt: accident.. the way we run tests makes it hard to run one. Later I will refactor but for now deleted this as it is already called

store.go
codefromthecrypt: well otherwise an end-user can't choose an engine, so yeah I think we need to keep this exported.

codefromthecrypt:
I added a comment to all the places we do manual NaN comparison which are ultimately caused by this "won't fix" https://github.com/stretchr/testify/issues/624

codefromthecrypt:
Thanks for the feedback and the help with the float bugs @mathetake and @nullpo-head!

PR #267, opened 2022-02-21
jit(amd64): less instructions on call/call_indirect

mathetake:
previous

```
if type == uninitialized {
    exit invalid_call
}
if type != expected {
    exit type_mismatch
}

call func
```

now:

```
if type != expected {
    if type == uninitialied {
        exit invalid_call
    } else {
        exit type_mismatch
    }
}

call func
```

therefore less cycles consumed on the regular case (which should be easy to predict the branch)

### DIFF
--- a/internal/wasm/jit/jit_amd64.go
+++ b/internal/wasm/jit/jit_amd64.go
@@ -1001,9 +1001,23 @@ func (c *amd64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
 	movTableSliceAddress.From.Offset = engineModuleContextTableElement0AddressOffset
 	c.addInstruction(movTableSliceAddress)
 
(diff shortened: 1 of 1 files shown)


internal/wasm/jit/jit_amd64.go
```diff
@@ -4702,7 +4700,7 @@ func (c *amd64Compiler) callFunction(addr wasm.FunctionAddress, addrReg int16, f
 	}
 
 	// 2) Set engine.valueStackContext.stackBasePointer for the next function.
-	{
+	if offset := (int64(c.locationStack.sp) - int64(len(functype.Params))); offset > 0 {
```
mathetake, line 4703: if zero, the emitted code is virtually noop so I simply omit at the compilation phase

nullpo-head (approved):

PR #272, opened 2022-02-21
wasi: backfill docs and tests of fd_close fd_prestat_get fd_prestat_dir_name fd_read

nullpo-head:
This PR backfills the docs and the tests of part of existing WASI APIs.

### DIFF
--- a/internal/wasi/wasi.go
+++ b/internal/wasi/wasi.go
@@ -394,7 +394,67 @@ type SnapshotPreview1 interface {
 	FdPrestatDirName(ctx wasm.ModuleContext, fd uint32, resultPrestat uint32) wasi.Errno
 
 	// TODO: wasi.FdPwrite
+
 	// TODO: wasi.FdRead
+	// FdRead is the WASI function to read from a file descriptor.
+	//
+	// * fd - the file decriptor to read from
+	// * iovs - the offset in `ctx.Memory` that contains a series of `IOVec` struct, which indicates where to write the bytes read. See the example below.
+	// * iovsLen - the number of `IOVec`s in `iovs`
+	// * resultSize - the offset in `ctx.Memory` to write the number of bytes read
+	//
+	// FdRead returns the following errors.
+	// * ErrnoBadf - if `fd` is invalid
+	// * ErrnoFault - if `iovs` or `resultSize` contain an invalid offset due to the memory constraint
+	// * ErrnoIo - if an IO related error happens during the operation
+	//
+	// For example, suppose fd 3 is a file with the contents "test",
+	// and FdRead parameters fd = 3, iovs = 1, and iovsLen = 2, resultSize = 24
+	// and two `IOVec`s are located at the offset 1.
+	// Those two `IOVec`s are as follows.
+	//   []IOVec {
+	//     {
+	//       buf: 18    /* type: uint32 */,
+	//       bufLen: 2  /* type: uint32 */,
+	//     },
+	//     {
+	//       buf: 21    /* type: uint32 */,
+	//       bufLen: 2  /* type: uint32 */,
+	//     },
+	//   }
+	// Note: Wazero does not export nor define IOVec struct explicitly. These are examples.
+	// Now, `ctx.Memory` contains
+	//
+	//                       IOVec                    IOVec
+	//              +--------------------+   +--------------------+
+	//              |uint32le    uint32le|   |uint32le    uint32le|
+	//              +--------+  +--------+   +--------+  +--------+
+	//              |        |  |        |   |        |  |        |
+	//   []byte{?, 18, 0, 0, 0, 2, 0, 0, 0, 21, 0, 0, 0, 2, 0, 0, 0, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ? }
+	//       iovs --^           ^            ^           ^                   resultSize --^
+	//              | iovs[0].  |  iovs[1].  | iovs[1].  |                     (= 24th)
+	//    iovs[0].  |  bufLen --+      buf --+  bufLen --+
+	//        buf --+
+	//
+	// After FdRead completes, we expect `ctx.Memory` to contain:
+	//
+	//                       IOVec                    IOVec
+	//              +--------------------+   +--------------------+      iovs[0]      iovs[1]
+	//              |uint32le    uint32le|   |uint32le    uint32le|      bufLen       bufLen       uint32le
+	//              +--------+  +--------+   +--------+  +--------+      +----+       +----+      +--------+
+	//              |        |  |        |   |        |  |        |      |    |       |    |      |        |
+	//   []byte{?, 18, 0, 0, 0, 2, 0, 0, 0, 21, 0, 0, 0, 2, 0, 0, 0, ?, 't', 'e', ?, 's', 't', ?, 4, 0, 0, 0 }
+	//       iovs --^                                      iovs[0].buf --^            ^           ^
+	//                                                        (= 18th)  iovs[1].buf --+           |
+	//                                                                     (= 21st)  resultSize --+
+	//                                                                                 (= 24th)
+	//
+	// Note: ImportFdRead shows this signature in the WebAssembly 1.0 (MVP) Text Format.
+	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#fd_read
+	// Note: This is similar to `readv` in POSIX.
+	// See https://linux.die.net/man/3/readv
+	// FdRead(ctx wasm.ModuleContext, fd uint32, iovs uint32, iovsLen uint32, resultSize uint32) wasi.Errno
+
 	// TODO: wasi.FdReaddir
 	// TODO: wasi.FdRenumber
 	// TODO: wasi.FdSeek
--- a/internal/wasi/wasi.go
+++ b/internal/wasi/wasi.go
@@ -80,9 +80,16 @@ const (
 	ImportClockTimeGet = `(import "wasi_snapshot_preview1" "clock_time_get"
     (func $wasi.clock_time_get (param $id i32) (param $precision i64) (param $result.timestamp i32) (result (;errno;) i32)))`
 
-	FunctionFdAdvise             = "fd_advise"
-	FunctionFdAllocate           = "fd_allocate"
-	FunctionFdClose              = "fd_close"
+	FunctionFdAdvise   = "fd_advise"
+	FunctionFdAllocate = "fd_allocate"
+
+	// FunctionFdClose closes a file descriptor.
+	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#fd_close
+	FunctionFdClose = "fd_close"
+	// ImportFdClose is the WebAssembly 1.0 (MVP) Text format import of FunctionFdClose
+	ImportFdClose = `(import "wasi_snapshot_preview1" "fd_close"
+    (func $wasi.fd_close (param $fd i32) (result (;errno;) i32)))`
+
 	FunctionFdDataSync           = "fd_datasync"
 	FunctionFdFdstatGet          = "fd_fdstat_get"
 	FunctionFdFdstatSetFlags     = "fd_fdstat_set_flags"
@@ -297,7 +304,17 @@ type SnapshotPreview1 interface {
 
 	// TODO: wasi.FdAdvise
 	// TODO: wasi.FdAllocate
-	// TODO: wasi.FdClose
+
+	// FdClose is the WASI function to close a file descriptor. This returns ErrnoBadf if the fd is invalid.
+	//
+	// * fd - the file decriptor to close
+	//
+	// Note: ImportFdClose shows this signature in the WebAssembly 1.0 (MVP) Text Format.
+	// Note: This is similar to `close` in POSIX.
+	// See https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#fd_close
+	// See https://linux.die.net/man/3/close
+	FdClose(ctx wasm.ModuleContext, fd uint32) wasi.Errno
+
 	// TODO: wasi.FdDataSync
 	// TODO: wasi.FdFdstatGet
 	// TODO: wasi.FdFdstatSetFlags
@@ -396,7 +413,7 @@ func SnapshotPreview1Functions(opts ...Option) (nameToGoFunc map[string]interfac
 		FunctionClockTimeGet: a.ClockTimeGet,
 		// TODO: FunctionFdAdvise
 		// TODO: FunctionFdAllocate
-		FunctionFdClose: a.fd_close,
+		FunctionFdClose: a.FdClose,
 		// TODO: FunctionFdDataSync
 		FunctionFdFdstatGet: a.fd_fdstat_get,
 		// TODO: FunctionFdFdstatSetFlags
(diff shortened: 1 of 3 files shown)


internal/wasi/wasi.go
```diff
@@ -504,6 +649,53 @@ func (a *wasiAPI) ClockTimeGet(ctx wasm.ModuleContext, id uint32, precision uint
...
+		return wasi.ErrnoBadf
+	}
+	return wasi.ErrnoSuccess
+}
+
+// FdPrestatDirName implements SnahpshotPreview1.FdPrestatDirName
+func (a *wasiAPI) FdPrestatDirName(ctx wasm.ModuleContext, fd uint32, pathPtr uint32, pathLen uint32) wasi.Errno {
+	f, ok := a.opened[fd]
+	if !ok {
+		return wasi.ErrnoBadf
```
nullpo-head, line 681: I've updated this line so that it returns `ErrnoBadf` instead of `ErrnoInval`

internal/wasi/wasi.go
```diff
@@ -504,6 +649,53 @@ func (a *wasiAPI) ClockTimeGet(ctx wasm.ModuleContext, id uint32, precision uint
...
+	}
+
+	// The semantics of pathLen in wazero follows that of wasmtime.
+	// See https://github.com/bytecodealliance/wasmtime/blob/2ca01ae9478f199337cf743a6ab543e8c3f3b238/crates/wasi-common/src/snapshots/preview_1.rs#L578-L582
+	// Some runtimes such as wasmer may have another semantics, where they consider pathLen as the maximum length to write, instead of the exact length to write.
+	// See https://github.com/wasmerio/wasmer/blob/3463c51268ed551933392a4063bd4f8e7498b0f6/lib/wasi/src/syscalls/mod.rs#L764
```
nullpo-head, line 687: Added the comments

internal/wasi/wasi.go
```diff
@@ -306,10 +336,125 @@ type SnapshotPreview1 interface {
...
+	// For example, if fd 3 is a file with a prestat
+	//   /* Note: Prestat is not defined in wazero yet. This is an example for explanation. */
+	//   Prestat {
+	//     Tag: PrestatDir /* 0 (type: uint8, offset: 0) */
+	//     PrNameLen: 3    /* 3 (type: uint32, offset: 4) */
+	//   }
+	// and FdPrestatGet parameters fd = 3, resultPrestat = 1, we expect `ctx.Memory` to contain:
+	//
+	//                     padding   uint32le
+	//          uint8 --+  +-----+  +--------+
```
codefromthecrypt, line 356: good example. Instead of defining a datatype, you can have other docs refer here to the bit layout of the tag and prNameLen. Reason is that we cannot pass structs and don't want to repeat ourselves, so I think this is the better of evils for now.

internal/wasi/wasi.go
```diff
@@ -306,10 +336,125 @@ type SnapshotPreview1 interface {
...
+	//     {
+	//       buf: 18    /* type: uint32 */,
+	//       bufLen: 2  /* type: uint32 */,
+	//     },
+	//     {
+	//       buf: 21    /* type: uint32 */,
+	//       bufLen: 2  /* type: uint32 */,
+	//     },
+	//   }
+	// Note: Wazero does not export nor define IOVec struct explicitly. These are examples.
```
codefromthecrypt, line 425: let's not define things as a struct. stick with the bit layout. Because I think many times we'll have offset, length pairs in this spec even if the name varies. 

Ex. in text
each element of iovs is a memory offset/length pair encoded sequentially as uint32le

I can see the diagram is getting crowded, and applaud the try. Let's try a different layout for this more complicated one..

In the diagram switch from pseudo-struct, s/IOVec/iovs[N/ and re-optimize the space to use what was used for uint32le to buf, bufLen as it was getting crowded. We can still put a reminder to substitute.

```
        // Note: each buf and bufLen are encoded as uint32le
       //
	//                       iovs[0]                    iovs[1]    
	//              +--------------------+   +--------------------+
	//              |      buf     |   |bufLen  |    |   buf      |   |  bufLen  |
	//              +--------+  +--------+   +--------+  +--------+
	//              |        |  |        |   |        |  |        |
	
```

internal/wasi/wasi.go
```diff
@@ -306,10 +336,125 @@ type SnapshotPreview1 interface {
...
+	// Now, `ctx.Memory` contains
+	//
+	//                       IOVec                    IOVec
+	//              +--------------------+   +--------------------+
+	//              |uint32le    uint32le|   |uint32le    uint32le|
+	//              +--------+  +--------+   +--------+  +--------+
+	//              |        |  |        |   |        |  |        |
+	//   []byte{?, 18, 0, 0, 0, 2, 0, 0, 0, 21, 0, 0, 0, 2, 0, 0, 0, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ? }
+	//       iovs --^           ^            ^           ^                   resultSize --^
```
codefromthecrypt, line 434: skip adding resultSize here I think as we are describing input data

internal/wasi/wasi.go
```diff
@@ -306,10 +336,125 @@ type SnapshotPreview1 interface {
...
+	//        buf --+
+	//
+	// After FdRead completes, we expect `ctx.Memory` to contain:
+	//
+	//                       IOVec                    IOVec
+	//              +--------------------+   +--------------------+      iovs[0]      iovs[1]
+	//              |uint32le    uint32le|   |uint32le    uint32le|      bufLen       bufLen       uint32le
+	//              +--------+  +--------+   +--------+  +--------+      +----+       +----+      +--------+
+	//              |        |  |        |   |        |  |        |      |    |       |    |      |        |
+	//   []byte{?, 18, 0, 0, 0, 2, 0, 0, 0, 21, 0, 0, 0, 2, 0, 0, 0, ?, 't', 'e', ?, 's', 't', ?, 4, 0, 0, 0 }
```
codefromthecrypt, line 446: skip redefining the part we didn't change.... you can still use the same offset by padding some text. it is 

```
[]byte{?, -- input bytes from above remain unchanged --  

```

This also allows you to move this whole diagram up several lines which makes it easier to understand visually against the input one

internal/wasi/wasi.go
codefromthecrypt: put in RATIONALE.md as we are accumulating a lot of these and will accumulate more due to underspecification. Then, you can line comment or trailing comment the anchor text if you like.

internal/wasi/wasi.go
```diff
@@ -775,21 +944,6 @@ func (a *wasiAPI) fd_read(ctx wasm.ModuleContext, fd uint32, iovsPtr uint32, iov
 	return wasi.ErrnoSuccess
 }
 
-func (a *wasiAPI) fd_close(ctx wasm.ModuleContext, fd uint32) wasi.Errno {
```
codefromthecrypt, line 778: keep definition in the same order as IIRC I had re-ordered these to match spec order. not a huge deal, but better to be consistent

internal/wasi/wasi_test.go
```diff
@@ -474,7 +474,35 @@ func TestAPI_ClockTimeGet_Errors(t *testing.T) {
 
 // TODO: TestAPI_FdAdvise TestAPI_FdAdvise_Errors
 // TODO: TestAPI_FdAllocate TestAPI_FdAllocate_Errors
-// TODO: TestAPI_FdClose TestAPI_FdClose_Errors
+
+func TestAPI_FdClose(t *testing.T) {
+	opt := Preopen("test", &MemFS{})
+	fd := uint32(3) // fd 3 will be opened for the "test" directory after 0, 1, and 2, that are stdin/out/err
```
codefromthecrypt, line 480: great comment

internal/wasi/wasi_test.go
```diff
@@ -474,7 +474,35 @@ func TestAPI_ClockTimeGet_Errors(t *testing.T) {
...
+	store, ctx, fn := instantiateWasmStore(t, FunctionFdClose, ImportFdClose, "test", opt)
+
+	t.Run("SnapshotPreview1.FdClose", func(t *testing.T) {
+		api := newAPI(opt)
+		errno := api.FdClose(ctx, fd)
+		require.Equal(t, wasi.ErrnoSuccess, errno)
+		require.NotContains(t, api.opened, fd) // Fd is closed and removed from the opned FDs.
```
codefromthecrypt, line 487:
```suggestion
		require.NotContains(t, api.opened, fd) // Fd is closed and removed from the opened FDs.
```

internal/wasi/wasi_test.go
```diff
@@ -474,7 +474,35 @@ func TestAPI_ClockTimeGet_Errors(t *testing.T) {
 
 // TODO: TestAPI_FdAdvise TestAPI_FdAdvise_Errors
 // TODO: TestAPI_FdAllocate TestAPI_FdAllocate_Errors
-// TODO: TestAPI_FdClose TestAPI_FdClose_Errors
+
+func TestAPI_FdClose(t *testing.T) {
+	opt := Preopen("test", &MemFS{})
```
codefromthecrypt, line 479: Let's be more direct as it explains more what's going on. The only complicated API (ex that uses the filesystem) is fd_open. For the others we can be really direct as it make later assertions clear on side-effects.

```
fd := uint32(3) // example FD past stdin/out/err
otherFD := uint32(5)

var wasiAPI *wasiAPI
store, ctx, fn := instantiateWasmStore(t, FunctionFdClose, ImportFdClose, "test", func(a *wasiAPI) {
		a.opened = map[uint32]fileEntry{
                    fd: fileEntry{path: "to be closed"},
                    otherFD: fileEntry{path: "shouldn't be closed"},
                 }
wasiAPI = a // for later assertions
})
```

internal/wasi/wasi.go
```diff
@@ -80,19 +80,39 @@ const (
...
+	FunctionFdPread            = "fd_pread"
+
+	// FunctionFdPrestatGet returns the prestat data of a file descriptor.
+	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#fd_prestat_get
+	FunctionFdPrestatGet = "fd_prestat_get"
+	// ImportFdPrestatGet is the WebAssembly 1.0 (MVP) Text format import of FunctionFdPrestatGet
+	ImportFdPrestatGet = `(import "wasi_snapshot_preview1" "fd_prestat_get"
+    (func $wasi.fd_prestat_get (param $fd i32) (param $result.prestat i32) (result (;errno;) i32)))`
+
+	// FunctionFdPrestatDirName returns the path of the prestat directory of a file descriptor.
```
mathetake, line 109: what is prestat? (I don't know why it is named in that way in the spec though :D)

```suggestion
	// FunctionFdPrestatDirName returns the path of the preopended directory of a file descriptor.
```

internal/wasi/wasi_test.go
```diff
@@ -474,7 +474,35 @@ func TestAPI_ClockTimeGet_Errors(t *testing.T) {
...
+	})
+	t.Run(FunctionFdClose, func(t *testing.T) {
+		ret, err := store.Engine.Call(ctx, fn, uint64(fd), uint64(fd))
+		require.NoError(t, err)
+		require.Equal(t, wasi.ErrnoSuccess, wasi.Errno(ret[0])) // cast because results are always uint64
+
+		// Ensure that it's closed by checking that closing it again returns ErrnoBadf.
+		ret, err = store.Engine.Call(ctx, fn, uint64(fd), uint64(fd))
+		require.NoError(t, err)
+		require.Equal(t, wasi.ErrnoBadf, wasi.Errno(ret[0])) // cast because results are always uint64
```
codefromthecrypt, line 497: require wasiAPI.opened[fd] is nil and otherFD is not

internal/wasi/wasi.go
```diff
@@ -504,6 +649,53 @@ func (a *wasiAPI) ClockTimeGet(ctx wasm.ModuleContext, id uint32, precision uint
...
+// FdPrestatDirName implements SnahpshotPreview1.FdPrestatDirName
+func (a *wasiAPI) FdPrestatDirName(ctx wasm.ModuleContext, fd uint32, pathPtr uint32, pathLen uint32) wasi.Errno {
+	f, ok := a.opened[fd]
+	if !ok {
+		return wasi.ErrnoBadf
+	}
+
+	// The semantics of pathLen in wazero follows that of wasmtime.
+	// See https://github.com/bytecodealliance/wasmtime/blob/2ca01ae9478f199337cf743a6ab543e8c3f3b238/crates/wasi-common/src/snapshots/preview_1.rs#L578-L582
+	// Some runtimes such as wasmer may have another semantics, where they consider pathLen as the maximum length to write, instead of the exact length to write.
```
mathetake, line 686: oh interesting and I see the wasmtime is correct is another is 🤷

internal/wasi/wasi_test.go
```diff
@@ -484,7 +512,96 @@ func TestAPI_ClockTimeGet_Errors(t *testing.T) {
 // TODO: TestAPI_FdFilestatSetTimes TestAPI_FdFilestatSetTimes_Errors
 // TODO: TestAPI_FdPread TestAPI_FdPread_Errors
 // TODO: TestAPI_FdPrestatGet TestAPI_FdPrestatGet_Errors
-// TODO: TestAPI_FdPrestatDirName TestAPI_FdPrestatDirName_Errors
+
+func TestAPI_FdPrestatDirName(t *testing.T) {
+	opt := Preopen("test", &MemFS{})
```
codefromthecrypt, line 517: ex similar to the above comment, make a literal map change with the expected dir name, then this test gets a lot more direct.

```
fd := uint32(3) // example FD past stdin/out/err
expectedDirName := "/tmp"

store, ctx, fn := instantiateWasmStore(t, FunctionFdClose, ImportFdClose, "test", func(a *wasiAPI) {
		a.opened = map[uint32]fileEntry{fd: fileEntry{path: "/etc/wazero"}}
})
```

internal/wasi/wasi_test.go
```diff
@@ -484,7 +512,96 @@ func TestAPI_ClockTimeGet_Errors(t *testing.T) {
...
+	store, ctx, fn := instantiateWasmStore(t, FunctionFdPrestatDirName, ImportFdPrestatDirName, "test", opt)
+	fd := uint32(3) // fd 3 will be opened for the "test" directory after 0, 1, and 2, that are stdin/out/err
+
+	path := uint32(1)    // arbitrary offset
+	pathLen := uint32(3) // arbitrary length
```
codefromthecrypt, line 522: when you make the literal map, please use len(expectedDirName) here instead of arbitrary

codefromthecrypt (approved):
Thanks a lot for biting off some tricky APIs. I made a lot of feedback, but I think once addressed this is good to go.

mathetake (approved):

codefromthecrypt:
ps I changed the title. let's keep this PR at max 4 functions (existing here are fd_close fd_prestat_get fd_prestat_dir_name fd_read) and of course you can axe one if you like.

internal/wasi/wasi.go
nullpo-head: nice catch! They were simply my typos...

internal/wasi/wasi.go
nullpo-head: after separating the diagram into the input and the output, the diagram becomes as  simple as Environ and Args's one, so I keep using the byte by byte memory diagram now. Other points are fixes as discussed

internal/wasi/wasi.go
nullpo-head: Actually it wat not, but I've fixed as discussed. Thanks!

codefromthecrypt (commented):
looks great. thanks

PR #274, opened 2022-02-22
Makes API easier by returning nil when a function isn't available

codefromthecrypt:
Before, we emulated the map api by returning `fn, ok` to tell if a
function was there or not. This made a lot of cruft vs returning `nil`
instead, especially as many times configuration of functions is near
where they are called in the source (so they aren't likely to be nil).

### DIFF
--- a/wasm/wasm.go
+++ b/wasm/wasm.go
@@ -19,8 +19,8 @@ type Store interface {
 //
 // Note: This is an interface for decoupling, not third-party implementations. All implementations are in wazero.
 type ModuleExports interface {
-	// Function returns a function exported from this module or false if it wasn't.
-	Function(name string) (Function, bool)
+	// Function returns a Wasm function exported from this module or nil if it wasn't.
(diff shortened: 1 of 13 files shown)


wasm/wasm.go
```diff
@@ -19,8 +19,8 @@ type Store interface {
 //
 // Note: This is an interface for decoupling, not third-party implementations. All implementations are in wazero.
 type ModuleExports interface {
-	// Function returns a function exported from this module or false if it wasn't.
-	Function(name string) (Function, bool)
+	// Function returns a Wasm function exported from this module or nil if it wasn't.
```
codefromthecrypt, line 22:
```suggestion
	// Function returns a function exported from this module or nil if it wasn't.
```

nullpo-head (approved):

mathetake (approved):

codefromthecrypt:
thanks again for the fast review

Issue #233, opened 2022-02-14
JIT: replace golang-asm with a thread-safe one (from scratch)


mathetake:
In aarch64, golang-asm is not thread-safe, and [the multi-goroutine engine test case](https://github.com/tetratelabs/wazero/blob/0186e41b27e572964e9f018cd9cd8a655eb6d3f8/tests/engine/adhoc_test.go#L67-L87) actually fails if gorutine num > 1.

We could introduce a mutex in engine.Compile and lock the golang-asm usage, then the test passes, but it's not ideal.

Instead, we should replace golang-asm with a thread-safe, but there's not alternative out there. 

That means we have to implement an assembler from scratch.

mathetake:
In anyway, removing golang-asm dependency is good and well aligned with our main goal "zero dependency".

codefromthecrypt:
FWIW I totally agree with this change

codefromthecrypt:
please make it internal though :) I don't expect we need to support this api for third party use

mathetake:
done with amd64.. now is the time to enjoy arm64!

PR #276, opened 2022-02-22
Complete JIT compilation engine for arm64 target.

### LINKED ISSUE #187: Tracking issue for arm64 JIT compiler

mathetake:
This commit completes the baseline single pass JIT engine for arm64 target.
The implementation passes 100% of specification tests and all the e2e tests
that have been used for amd64. Notably, the engine is stable under high
concurrency where multiple gorutines are holding stores and each of them
has Wasm execution environment.

One thing to note is that  the assembler (golang-asm) is not goroutine-safe,
so we have to take a lock on the assembler usage, therefore the compilation 
cannot scale to multiple CPU cores. This will be resolved once we build our
homemade assembler in #233.

resolves #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,15 +35,15 @@ wazero is an early project, so APIs are subject to change until version 1.0.
 There's the concept called "engine" in wazero (which is a word commonly used in Wasm runtimes). Engines are responsible for compiling and executing WebAssembly modules.
 There are two types of engines are available for wazero:
 
-1. _Interpreter_: a naive interpreter-based implementation of Wasm virtual machine. Its implementation doesn't have any platform (GOARCH, GOOS) specific code, therefore _interpreter_ engine can be used for any compilation target available for Go (such as `arm64`).
-2. _JIT engine_: compiles WebAssembly modules, generates the machine code, and executing it all at runtime. Currently wazero only implements the JIT compiler for `amd64` target. Generally speaking, _JIT engine_ is faster than _Interpreter_ by order of magnitude. However, the implementation is immature and has a bunch of aspects that could be improved (for example, it just does a singlepass compilation and doesn't do any optimizations, etc.). Please refer to [internal/wasm/jit/RATIONALE.md](internal/wasm/jit/RATIONALE.md) for the design choices and considerations in our JIT engine.
+1. _Interpreter_: a naive interpreter-based implementation of Wasm virtual machine. Its implementation doesn't have any platform (GOARCH, GOOS) specific code, therefore _interpreter_ engine can be used for any compilation target available for Go (such as `riscv64`).
+2. _JIT engine_: compiles WebAssembly modules, generates the machine code, and executing it all at runtime. Currently wazero implements the JIT compiler for `amd64` and `arm64` target. Generally speaking, _JIT engine_ is faster than _Interpreter_ by order of magnitude. However, the implementation is immature and has a bunch of aspects that could be improved (for example, it just does a singlepass compilation and doesn't do any optimizations, etc.). Please refer to [internal/wasm/jit/RATIONALE.md](internal/wasm/jit/RATIONALE.md) for the design choices and considerations in our JIT engine.
 
 Both of engines passes 100% of [WebAssembly spec test suites]((https://github.com/WebAssembly/spec/tree/wg-1.0/test/core)) (on supported platforms).
 
-| Engine     | Usage|GOARCH=amd64 | GOARCH=others |
-|:----------:|:---:|:-------------:|:------:|
-| Interpreter|`wazero.NewEngineInterpreter()`| ✅    | ✅ |
-| JIT engine |`wazero.NewEngineJIT()`|   ✅   | ❌  |
+| Engine     | Usage|GOARCH=amd64 | GOARCH=arm64 |GOARCH=others |
+|:---|:---:|:---:|:---:|:---:|
+| Interpreter|`wazero.NewEngineInterpreter()`| ✅    |✅    | ✅ |
+| JIT engine |`wazero.NewEngineJIT()`|   ✅   |✅    |  ❌  |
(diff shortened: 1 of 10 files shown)


internal/wasm/jit/jit_arm64.go
```diff
@@ -69,14 +70,26 @@ const (
...
+	assemblerMutex.Unlock()
+}
+
 // newCompiler returns a new compiler interface which can be used to compile the given function instance.
 // Note: ir param can be nil for host functions.
-func newCompiler(f *wasm.FunctionInstance, ir *wazeroir.CompilationResult) (compiler, error) {
+func newCompiler(f *wasm.FunctionInstance, ir *wazeroir.CompilationResult) (c compiler, done func(), err error) {
+	// golang-asm is not goroutine-safe so we take lock until we complete the compilation.
+	// TODO: delete after https://github.com/tetratelabs/wazero/issues/233
+	assemblerMutex.Lock()
```
nullpo-head, line 86: I may be missing some contexts, but what is the reason `newCompiler` is returning the unlock function and asking the caller to call it, instead of `newCompiler` itself calling `defer assemblerMutex.Unlock()` or something like that?

internal/wasm/jit/engine.go
```diff
@@ -669,11 +661,15 @@ func (e *engine) addCompiledFunction(addr wasm.FunctionAddress, compiled *compil
 }
 
 func compileHostFunction(f *wasm.FunctionInstance) (*compiledFunction, error) {
-	compiler, err := newCompiler(f, nil)
+	compiler, done, err := newCompiler(f, nil)
 	if err != nil {
 		return nil, err
 	}
```
nullpo-head, line 667: This error case will cause the dead lock of the mutex. Maybe better to call `done` before this block, as long as it's not nil

internal/wasm/jit/jit_amd64.go
```diff
@@ -90,12 +90,12 @@ func newArchContext() (ret archContext) { return }
 
 // newCompiler returns a new compiler interface which can be used to compile the given function instance.
 // Note: ir param can be nil for host functions.
```
nullpo-head, line 92: I'd appreciate it if the doc explains what the caller must do against the returned func value

internal/wasm/jit/engine.go
mathetake: nice!

internal/wasm/jit/jit_arm64.go
nullpo-head: oops never mind. I've realized that the caller is not calling `done` but deferring it, and this `done` must be called after all compile related functions complete

README.md
```diff
@@ -35,15 +35,15 @@ wazero is an early project, so APIs are subject to change until version 1.0.
...
 Both of engines passes 100% of [WebAssembly spec test suites]((https://github.com/WebAssembly/spec/tree/wg-1.0/test/core)) (on supported platforms).
 
-| Engine     | Usage|GOARCH=amd64 | GOARCH=others |
-|:----------:|:---:|:-------------:|:------:|
-| Interpreter|`wazero.NewEngineInterpreter()`| ✅    | ✅ |
-| JIT engine |`wazero.NewEngineJIT()`|   ✅   | ❌  |
+| Engine     | Usage| amd64 | arm64 | others |
+|:---:|:---:|:---:|:---:|:---:|
+| Interpreter|`wazero.NewEngineInterpreter()`|✅ |✅|✅|
+| JIT engine |`wazero.NewEngineJIT()`|✅|✅ |❌|
```
codefromthecrypt, line 46:
```suggestion
| JIT engine |`wazero.NewEngineJIT()`|✅|✅ |❌|

*Note:* JIT does not yet work on Windows. Please use the interpreter and track [this issue](https://github.com/tetratelabs/wazero/issues/270) if interested.
```

internal/wasm/jit/engine.go
```diff
@@ -688,6 +682,10 @@ func compileHostFunction(f *wasm.FunctionInstance) (*compiledFunction, error) {
 		stackPointerCeil = res
 	}
 
+	if buildoptions.IsDebugMode {
+		fmt.Printf("compiled code in hex: %s\n", hex.EncodeToString(code))
```
codefromthecrypt, line 686: want this? I guess it could be huge even if not in our tests.

internal/wasm/jit/jit_amd64.go
```diff
@@ -90,12 +90,12 @@ func newArchContext() (ret archContext) { return }
 
 // newCompiler returns a new compiler interface which can be used to compile the given function instance.
```
codefromthecrypt, line 91:
```suggestion
// newCompiler returns a new compiler interface which can be used to compile the given function instance.
// The function returned must be invoked when finished compiling, so use `defer` to ensure this.
```

internal/wasm/jit/jit_arm64.go
```diff
@@ -69,14 +70,26 @@ const (
...
 
+// golang-asm is not goroutine-safe so we take lock until we complete the compilation.
+// TODO: delete after https://github.com/tetratelabs/wazero/issues/233
+var assemblerMutex = &sync.Mutex{}
+
+func unlockAssembler() {
+	assemblerMutex.Unlock()
+}
+
 // newCompiler returns a new compiler interface which can be used to compile the given function instance.
```
codefromthecrypt, line 81:
```suggestion
// newCompiler returns a new compiler interface which can be used to compile the given function instance.
// The function returned must be invoked when finished compiling, so use `defer` to ensure this.
```

codefromthecrypt (approved):
Looks good and nice tactical workaround with the mutex!

nullpo-head (approved):

mathetake:
Thanks for the look!

Issue #279, opened 2022-02-23
Support host exports besides function


codefromthecrypt:
I started porting https://github.com/summerwind/the-art-of-webassembly-go by @summerwind to the latest API and realized we need to finish APIs exportable from the host, including memory, table and globals. Currently, we only support host functions.

codefromthecrypt:
implementing this would be best done via a `HostModule` type because we currently enforce host modules once setup are immutable. This means we can't later add other things to them (ex globals)

codefromthecrypt:
step 1: https://github.com/tetratelabs/wazero/pull/280

mathetake:
let me take this as a part of #305

lthibault:
Would this allow me to make arbitrary Go types (like an interface) available to guest programs?  I've been trying to figure out how to do so for the better part of a week 😄 

Relatedly, would any of you be able to provide a bit more context on how this might work, and what would be involved?  I'd really appreciate it, as I'm trying to familiarize myself with WASM internals as I venture into Wazero.

As I understand it, the mechanism through which a guest program written in Go interacts with host objects is through the `syscall/js` package, and as such, implementing this feature would involve mapping native Go structs to `js.Value`s.  Is that right?  Any noteworthy details here?

Thanks!

codefromthecrypt:
@lthibault so there are a couple ways.

* one way is to use a data binding approach such as https://github.com/knqyf263/go-plugin or https://karmem.org.
* another is to define host functions that access certain parts of the type in question.
  * implicit instance ex on the host it is in the go context
    * guest sig - (func $get_field (param $name) (param $name_len))
  * explicit instance ex the host has a lookup table `map[uint32]*thing{}`
    * guest sig - (func $get_field (param $this i32) (param $name) (param $name_len))

The implicit case is how we deal with http request/response (implicit as middleware is sync) https://github.com/http-wasm/http-wasm-guest-tinygo

In the explicit case, the guest could get the `this` (some call it context ID) param from propagating it everywhere, or from memory or another way. Note: Commonly, people use uint32 if mapping, though sometimes people use as uint64. If you use uint64, resist passing pointers directly as the guest is untrusted and could do arithmetic on the number.

There's yet another way, to use "reference types", which allows you to pass a host ref (like uintptr) on the stack, but the problem is at the moment no compiler supports it #532. We can support it when a compiler does.

does this help?

codefromthecrypt:
[...]

PR #280, opened 2022-02-23
Consolidates API to config input types

codefromthecrypt:
This makes instantiating host or wasm-defined modules similar, by using
XXXConfig types. Doing so also allows configuration engines to seed
properties.

Since decoded modules are only usable during instantiation, this pushes
decoding inside those functions. By doing so, the API is easier to use
as it has less choices and less errors to catch. Detection is done
internally by peeking at the magic number.

See https://github.com/tetratelabs/wazero/issues/279


examples/wasi_test.go
```diff
@@ -30,22 +30,14 @@ func Test_WASI(t *testing.T) {
...
 	store := wazero.NewStore()
 
 	// Host functions can be exported as any module name, including the empty string.
-	_, err = wazero.ExportHostFunctions(store, "", map[string]interface{}{"random": goFunc})
-	require.NoError(t, err)
+	env := &wazero.HostModuleConfig{Name: "", Functions: map[string]interface{}{"random": goFunc}}
+	_, err := wazero.InstantiateHostModule(store, env)
 
 	// Configure WASI and implement the function to use it
-	we, err := wazero.ExportHostFunctions(store, wasi.ModuleSnapshotPreview1, wazero.WASISnapshotPreview1())
```
codefromthecrypt, line 48: this is one of the more obvious simplifications. At the moment, WASI only exports functions, so we are allowed a simpler config for it!

internal/wasm/binary/errors.go
```diff
@@ -4,7 +4,7 @@ import "errors"
 
 var (
 	ErrInvalidByte           = errors.New("invalid byte")
-	ErrInvalidMagicNumber    = errors.New("invalid magic number")
+	ErrInvalidMagicNumber    = errors.New("invalid Magic number")
```
codefromthecrypt, line 7:
```suggestion
	ErrInvalidMagicNumber    = errors.New("invalid magic number")
```

README.md
```diff
@@ -17,11 +17,8 @@ func main() {
 	// * Ex. (func (export "fac") (param i64) (result i64) ...
 	source, _ := os.ReadFile("./tests/engine/testdata/fac.wasm")
 
-	// Decode the binary as WebAssembly module.
-	mod, _ := wazero.DecodeModuleBinary(source)
-
 	// Instantiate the module with a Wasm Interpreter, to return its exported functions
-	exports, _ := wazero.InstantiateModule(wazero.NewStore(), mod)
+	exports, _ := wazero.InstantiateModule(wazero.NewStore(), &wazero.ModuleConfig{Source: source})
```
codefromthecrypt, line 21: now, we are down to minimally 3 steps to invoke a wasm function, including reading the source.

README.md
```diff
@@ -17,11 +17,8 @@ func main() {
 	// * Ex. (func (export "fac") (param i64) (result i64) ...
 	source, _ := os.ReadFile("./tests/engine/testdata/fac.wasm")
 
-	// Decode the binary as WebAssembly module.
```
codefromthecrypt, line 20: there was no use of this type except passing it to instantiate, so merging it with config is a net win for simplicity. If we ever want to make a wat2wasm thing we can do that as a separate utility and without messing up this api

wasm.go
```diff
@@ -1,66 +1,171 @@
...
 }
 
-// EncodeModule encodes the given module into a byte slice depending on the format of the implementation.
-// See binary.EncodeModule
-type EncodeModule func(m *Module) []byte
-
-var EncodeModuleBinary EncodeModule = func(m *Module) []byte {
-	return encodeModule(binary.EncodeModule, m)
+// ModuleConfig defines the WebAssembly 1.0 (MVP) module to instantiate.
+type ModuleConfig struct {
```
codefromthecrypt, line 62: PS there's no chance of misinterpreting text for binary format as the magic is not valid text. However, if later we find a user who really wants to not detect, we can add a flag to force the encoding of the source. I'm intentionally not doing that as each knob should pull its weight and I really don't think this would be worth describing.

wasm.go
```diff
@@ -1,66 +1,171 @@
...
+	}
+	return
 }
 
-// Name defaults to what's decoded from the custom name section and can be overridden WithName.
-// See https://www.w3.org/TR/wasm-core-1/#name-section%E2%91%A0
-func (m *Module) Name() string {
-	return m.name
+// HostModuleConfig are WebAssembly 1.0 (MVP) exports from the host bound to a module name used by InstantiateHostModule.
+type HostModuleConfig struct {
```
codefromthecrypt, line 117: This type will later have globals, memory and tables added to it. These are able to be added later without breaking API, hence this change being raised first.

mathetake (approved):

codefromthecrypt:
thanks for the look!

PR #283, opened 2022-02-23
Backfills integration tests of WithName to ensure bug is fixed

codefromthecrypt:
Sorry about the bug

### DIFF
--- a/wasm.go
+++ b/wasm.go
@@ -74,7 +74,7 @@ type ModuleConfig struct {
 //
 // This is used to pre-flight check and cache the module for later instantiation.
 func (m *ModuleConfig) Validate() (err error) {
-	_, _, err = decodeModule(m)
+	_, err = decodeModule(m)
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


wasm.go
```diff
@@ -136,11 +144,6 @@ func decodeModule(module *ModuleConfig) (m *internalwasm.Module, name string, er
 	// Cache as tools like wapc-go re-instantiate the same module many times.
 	module.validatedSource = module.Source
 	module.decodedModule = m
-
-	name = module.Name
```
codefromthecrypt, line 140: this was the bug. this mutated something that was cached

wasm_test.go
```diff
@@ -120,6 +117,22 @@ func TestDecodeModule_Errors(t *testing.T) {
 	}
 }
 
+// TestInstantiateModule_Renamed tests that we can pre-validate (cache) a module and instantiate it under different
+// names. This pattern is used in wapc-go.
+func TestInstantiateModule_Renamed(t *testing.T) {
```
codefromthecrypt, line 122: this is the exact pattern

mathetake (approved):

PR #284, opened 2022-02-23
Exports wasm.ValueType and makes Functions interfaces for signature verification

codefromthecrypt:
This moves code for `ValueType` constants public, so that we can
verify function signatures. To achieve this without complicating
function calls this makes `Function` and `HostFunction` `interface`
instead of `func` types.

Note: We'll also need `ValueType` when re-introducing a Globals API.

To keep complexity down in consideration that there are only 4 types, I
copied the constants into the internalwasm package. This reduces the
import complexity otherwise would have caused.

### DIFF
--- a/internal/wasm/host.go
+++ b/internal/wasm/host.go
@@ -67,22 +67,28 @@ func (c *ModuleContext) Function(name string) publicwasm.Function {
 	if err != nil {
 		return nil
 	}
-	return (&function{c: c, f: exp.Function}).Call
+	return &function{c: c, f: exp.Function}
 }
 
+// function wraps FunctionInstance to implement wasm.Function as otherwise there's a signature clash on Call.
 type function struct {
 	c *ModuleContext
 	f *FunctionInstance
 }
 
+// ParamTypes implements wasm.Function ParamTypes
+func (f *function) ParamTypes() []publicwasm.ValueType {
+	return f.f.ParamTypes()
+}
+
+// ResultTypes implements wasm.Function ResultTypes
+func (f *function) ResultTypes() []publicwasm.ValueType {
+	return f.f.ResultTypes()
+}
+
+// Call implements wasm.Function Call
 func (f *function) Call(ctx context.Context, params ...uint64) ([]uint64, error) {
-	paramSignature := f.f.FunctionType.Type.Params
-	paramCount := len(params)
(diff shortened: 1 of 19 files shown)


codefromthecrypt:
Revised to make `Function` and `HostFunction` interfaces as otherwise, there's no clean way to validate the signature corresponding to parameter inputs without shared magic knowledge.

The non-clean way would be to add a separate thing to Store to query for signatures. Since all other "exports" are represented by interfaces, doing the same thing for Function while also adding methods to return the signature is cleaner.

codefromthecrypt:
While writing this, I noticed we inconsistently validated parameter length depending on whether a function was a Wasm or host function. I'll rejig tests to prove this is done consistently now.

wasm/wasm.go
```diff
@@ -30,37 +78,34 @@ type ModuleExports interface {
...
+	ParamTypes() []ValueType
+
+	// ResultTypes are the possibly empty sequence of value types returned by a function with this signature.
+	//
+	// Note: In WebAssembly 1.0 (MVP), there can be at most one result.
+	// See https://www.w3.org/TR/wasm-core-1/#result-types%E2%91%A0
+	// See ValueType documentation for decoding rules.
+	ResultTypes() []ValueType
+
+	// Call is invokes the function with parameters encoded according to ParamTypes. Up to one result is returned,
```
nullpo-head, line 95: typo
```suggestion
	// Call invokes the function with parameters encoded according to ParamTypes. Up to one result is returned,
```

nullpo-head (approved):
thanks!

internal/wasm/host.go
```diff
@@ -67,22 +67,28 @@ func (c *ModuleContext) Function(name string) publicwasm.Function {
...
+}
+
+// ResultTypes implements wasm.Function ResultTypes
+func (f *function) ResultTypes() []publicwasm.ValueType {
+	return f.f.ResultTypes()
+}
+
+// Call implements wasm.Function Call
 func (f *function) Call(ctx context.Context, params ...uint64) ([]uint64, error) {
-	paramSignature := f.f.FunctionType.Type.Params
```
codefromthecrypt, line 79: #284 handles testing of parameter length in a more centralized way

codefromthecrypt:
thanks for the help @nullpo-head

PR #286, opened 2022-02-24
Pushes param length validation to engine

codefromthecrypt:
This pushes parameter length validation to inside the engine to avoid bugs. Formerly, we were inconsistent checking this depending on whether the function was Wasm or not.

This also replicates the tests for interpreter to JIT and renames `interpreter.interpreter` to `interpreter.engine` for consistency

### DIFF
--- a/internal/wasm/host.go
+++ b/internal/wasm/host.go
@@ -67,28 +67,29 @@ func (c *ModuleContext) Function(name string) publicwasm.Function {
 	if err != nil {
 		return nil
 	}
-	return &function{c: c, f: exp.Function}
+	return &exportedFunction{module: c, function: exp.Function}
 }
 
-// function wraps FunctionInstance to implement wasm.Function as otherwise there's a signature clash on Call.
-type function struct {
-	c *ModuleContext
-	f *FunctionInstance
+// exportedFunction wraps FunctionInstance so that it is called in context of the exporting module.
+type exportedFunction struct {
+	module   *ModuleContext
+	function *FunctionInstance
 }
 
 // ParamTypes implements wasm.Function ParamTypes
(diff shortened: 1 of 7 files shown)


internal/wasm/interpreter/interpreter.go
```diff
@@ -425,16 +424,22 @@ func (it *interpreter) lowerIROps(f *wasm.FunctionInstance,
 }
 
 // Call implements an interpreted wasm.Engine.
-func (it *interpreter) Call(ctx *wasm.ModuleContext, f *wasm.FunctionInstance, params ...uint64) (results []uint64, err error) {
-	prevFrameLen := len(it.frames)
+func (e *engine) Call(ctx *wasm.ModuleContext, f *wasm.FunctionInstance, params ...uint64) (results []uint64, err error) {
+	paramSignature := f.FunctionType.Type.Params
```
codefromthecrypt, line 428: this is the only new code. everything else is due to renaming interpreter to engine

codefromthecrypt:
rebased

internal/wasm/host.go
```diff
@@ -67,28 +67,29 @@ func (c *ModuleContext) Function(name string) publicwasm.Function {
...
 	}
-	return &function{c: c, f: exp.Function}
+	return &exportedFunction{module: c, function: exp.Function}
 }
 
-// function wraps FunctionInstance to implement wasm.Function as otherwise there's a signature clash on Call.
-type function struct {
-	c *ModuleContext
-	f *FunctionInstance
+// exportedFunction wraps FunctionInstance so that it is called in context of the exporting module.
```
codefromthecrypt, line 73: clarified the naming here as the more important point is that this assigns the correct module context

mathetake (approved):

PR #287, opened 2022-02-24
engine: add per-invocation context for thread-safe functions

mathetake:
This commit introduces `virtualMachine` type in each engine, which holds 
the context for each `engine.Call` invocation. Previously, Engines were holding
value stacks and callframe stacks, and therefore it was impossible to execute 
multiple functions in a single store concurrently as each of them were trying to
manipulate these stacks in a single engine (tied with each store).

As a side effect, we are no longer able to track "recursive call" and engines cannot
judge if the current invocation is the recursive call or not (previously it was possible
by tracking the current callframe stack's state before invoking any function).
Therefore, if the recursive call traps, now only the most recent `store.Call` returns error,
and the panic doesn't propagate and unwind all the way back to the original `store.Call`.
But this is in line with other runtimes behavior, so I hope this is fine. If the necessity comes up,
we will revisit this "how to propagate panic on recursive function call" problem.


mathetake:
oh I am totally surprised all the tests passed 😄

mathetake:
ok arm64 and interpreter done (passes all tests)

mathetake:
yay all passes tests

internal/wasm/jit/engine.go
```diff
@@ -17,17 +18,33 @@ import (
...
+		// as the underlying memory region is accessed by assembly directly by
+		// using compiledFunctionsElement0Address.
+		compiledFunctions []*compiledFunction
+
+		// mux is used for read/write access to compiledFunctions slice.
+		// This is necessary as each compiled function will access the slice from the native code
+		// when they make function calls while engine might be modifying the underlying slice when
+		// adding a new compiled function. We take read lock when creating new virtualMachine
+		// for each function invocation while take write lock in engine.addCompiledFunction.
+		mux sync.RWMutex
```
mathetake, line 32: this can be removed by assuming that `engine.Compile` would never be called from a different goroutine, but I think sooner or later we would add. That's why I added, but I'm fine removing this for now

tests/engine/adhoc_test.go
```diff
@@ -71,16 +71,22 @@ func testFibonacci(t *testing.T, newEngine func() *wazero.Engine) {
 	// We execute 1000 times in order to ensure the JIT engine is stable under high concurrency
 	// and we have no conflict with Go's runtime.
 	const goroutines = 1000
+
```
mathetake, line 74: this change in testFibonacci verifies that the concurrent execution of module functions is thread-safe!

internal/wasm/interpreter/interpreter.go
```diff
@@ -19,43 +20,63 @@ var callStackCeiling = buildoptions.CallStackCeiling
...
+
+const initialCompiledFunctionsSliceSize = 128
+
+func NewEngine() wasm.Engine {
+	return &engine{
+		functions: make([]*interpreterFunction, initialCompiledFunctionsSliceSize),
+	}
+}
+
+// virtualMachine holds context per engine.Call, and shared across all the
```
codefromthecrypt, line 42: maybe call this rootCall?

mathetake: oh you mean rootCall holds `stack []uint64` and `callFrame []*callFrame` ?

mathetake: I don't think it would be appropriate

mathetake: other runtimes call this kind of struct `vmCtx` so I followed it but I'm open to rename to similar one

codefromthecrypt (approved):
Excellent! my only beef is the term virtualMachine/vm which should be banished for something else :D

nullpo-head (approved):
kudos!

internal/wasm/interpreter/interpreter.go
```diff
@@ -431,73 +443,59 @@ func (e *engine) Call(ctx *wasm.ModuleContext, f *wasm.FunctionInstance, params
...
-
-	// shouldRecover is true when a panic at the origin of callstack should be recovered
-	//
-	// If this is the recursive call into Wasm (prevFrameLen != 0), we do not recover, and delegate the
-	// recovery to the first engine.Call.
-	//
-	// For example, given the call stack:
-	//	 "original host function" --(engine.Call)--> Wasm func A --> Host func --(engine.Call)--> Wasm function B,
-	// if the top Wasm function panics, we go back to the "original host function".
-	shouldRecover := prevFrameLen == 0
```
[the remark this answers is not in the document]
mathetake: I think the previous behavior to unwind back to the origin is odder (compared to other runtimes) than the current one, so I will leave no TODO for now!

mathetake:
thanks for the look!

Issue #271, opened 2022-02-21
Complete minimum viable WASI wasi_snapshot_preview1


codefromthecrypt:
Here is the status of our implementation of wasi_snapshot_preview1
- [x] args_get
- [x] args_sizes_get
- [x] environ_get
- [x] environ_sizes_get
- [ ] clock_time_get (partially implemented)
- [ ] poll_oneoff (TinyGo needs this)
- [x] fd_close
- [ ] fd_fdstat_get (partially implemented)
- [ ] fd_prestat_get
- [x] fd_prestat_dir_name
- [x] fd_read
- [ ] fd_seek @nullpo-head working on
- [x] fd_write
- [ ] path_open (partially implemented)
- [x] fd_write
- [x] random_get
- [x] proc_exit

The above should only be checked when all parts of the API are implemented or we know what parts we won't implement. For example, don't tick a box if an API is skipping parameters unless we actively rule out a parameter. Definitely no box can be ticked until it is properly documented and tested in wasi.go and wasi_test.go

Notes: Per #263 there are some problems in `wasi_snapshot_preview1`, so it may not make sense to burn time on implementing all features and constants of it. That said, there are code using it, notably TinyGo, and we should complete a minimum viable subset of the API and its features based on what users want.

pims:
`proc_raise` would be nice to be able to use Grain Lang compiled webassembly.

nullpo-head:
> `proc_raise` would be nice to be able to use Grain Lang compiled webassembly.

Hi @pims, thanks for the feedback! Could you share some references on how Gain Lang uses `proc_raise`, please? I personally tried Grain but I couldn't see the compiled webassembly import `proc_raise`, at least for hello-world.

pims:
@nullpo-head 

Given the following grain program:
```
import Process from "sys/process"
import Array from "array"

match (Process.env()) {
 Ok(arr) => Array.forEach(print, arr),
 Err(err) => throw err
}
```

The following go code:
```go
store, err := wazero.NewStoreWithConfig(&wazero.StoreConfig{
		ModuleToHostFunctions: map[string]*wazero.HostFunctions{
			wasi.ModuleSnapshotPreview1: wazero.WASISnapshotPreview1WithConfig(wasiConfig),
		},
	})
	if err != nil {
		log.Fatal(err)
	}
	// StartWASICommand runs the "_start" function which is what TinyGo compiles "main" to
	_, err = wazero.StartWASICommand(store, s.mod)
[...]

PR #290, opened 2022-02-24
Implement proc_raise

nullpo-head:

### DIFF
--- a/internal/wasi/wasi.go
+++ b/internal/wasi/wasi.go
@@ -86,6 +86,7 @@ const (
 	// FunctionFdClose closes a file descriptor.
 	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#fd_close
 	FunctionFdClose = "fd_close"
+
 	// ImportFdClose is the WebAssembly 1.0 (MVP) Text format import of FunctionFdClose
 	ImportFdClose = `(import "wasi_snapshot_preview1" "fd_close"
     (func $wasi.fd_close (param $fd i32) (result (;errno;) i32)))`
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
thanks for the straw man. I think this is best decoupled from OS handling especially what is supported by what OS is out of scope for us especially in documenting the WASI part.

Here's the suggestion:

add this type to the public wasi package, documenting that it is for proc_raise
```
type SignalHandler func(sig uint32) wasi.Errno
```

Add a default impl which only handles the exit signal and calls exit 1

ex.
```
var defaultSignalHandler SignalHandler = func(sig uint32) wasi.Errno {
   if exit or kill ... panic(wasi.ExitCode(exitCode))
}
```

add an option to override `SignalHandler`

Implement  a built-in one for the current process, which I guess is only viable for Wasm-only processes. implement and cast the signals as syscall.Signal() and map go error to an errno
ex.
```

type currentPidSignalHandler { process}
func CurrentPidSignalHandler() (SignalHandler, err) {
// lookup process or fail
   return &currentPidSignalHandler{process}
}

// implementation is mainly just calling process.Signal
```
 
This allows unit tests to be easy as you can override `SignalHandler`. Also you can stick with cursory tests for the CurrentPidSignalHandler() for example, only testing when the OS isn't windows for now. What Go supports is out of scope and if someone wants to kill a process on windows we can ask and re-use code from func-e at that point.

codefromthecrypt:
ps only add the `SignalHandler` type to the public wasi package.. the others leave internal. Of course similar to other things holes will need to be punched in wazero.WASIConfig

codefromthecrypt:
Looks like we need to implement https://github.com/tetratelabs/wazero/issues/293 and handle at least sigabrt (for AssemblyScript) and whatever grain ends up calling. Process exit is deprioritized as current use cases don't affect the host process. In fact, we can probably drop that entirely as no-one asked for it and stick with module or maybe also store scope. This focuses the pull review on code needed and leaves the gnarly os.Signal stuff out of it

nullpo-head:
As discussed in https://github.com/tetratelabs/wazero/issues/271#issuecomment-1056440634, `proc_raise` is no longer necessary. I close this draft PR.

Issue #293, opened 2022-02-24
Support deinstantiation (mod.Close)


codefromthecrypt:
We need to eventually handle long lived processes, and graceful shutdown. If we had a way to deinstantiate a module, I think we can also handle a WASI quit signal scoped to module instead of the host process (#290)

I think we should make a new type that implements ModuleExports called `wazero.InstantiatedModule` (we maybe need this for host also, but let's start with normal modules). The only additional function there would be .Close and this would have the following effects.

* set a goroutine visible status on the internal moduleInstance  `closeCalled`
  * this may need to be an array of length one to ensure visibility across goroutines, and 1 means closeCalled
* make shared implementations err if they see this status which will cause existing functions to exit
  * make engine.call implementations check this status prior to invoking new functions and return an error (I think this handles table also as table is only call_indirect and not exposed otherwise)
  * make memory, global implementations check this status before read/write and return an error
* ensure calls begin with a `CancelContext` added to the callEngine, and invoke that hook so that any host functions doing I/O finish.
  * this could possibly be limited to wrapping the context in `wazero.StartWASICommand` as maybe non-wasi shouldn't be doing I/O anyway
* remove all the references for this module in Store added by instantiate
  * ensure nice errors when calling an imported function, memory, table or global that was removed

codefromthecrypt:
This would also be needed for wapc-go (cc @pkedy) because it is otherwise a memory leak. For example, modules are instantiated with instance counters, which I assume would eventually run out of memory depending on the scope and frequency of creation (remember memory is required by WASI and minimum memory is 64KB).

```
// Instantiate creates a single instance of the module with its own memory.
func (m *Module) Instantiate() (wapc.Instance, error) {
	moduleName := fmt.Sprintf("%d", atomic.AddUint64(&m.instanceCounter, 1))
```

codefromthecrypt:
There's a not really documented thing in AssemblyScript that would also needs this for abort handling. If someone can find docs on it please do!

```go
// envAbort is the AssemblyScript abort handler.
// Ex in WebAssembly 1.0 (MVP) Text Format:
//	(import "env" "abort" (func $~lib/builtins/abort (param i32 i32 i32 i32)))
func (t *thing) envAbort(ctx wasm.ModuleContext, messageOffset, fileOffset, line, col uint32) {
	t.module.Close()
}
```

Note: AssemblyScript maybe maps this function to WASI sigabrt ...

mathetake:
[...]

PR #294, opened 2022-02-25
Refactors store and adds ReleaseModuleInstance.

mathetake:
This PR refactors store.go and adds store.ReleaseModuleInstance.
Notably, this removes the "rollback" functions in InstantiateModule
which we had used to rollback the mutated state when we encounter
the initialization failure. Instead, we separate out the validation from
initialization and migrate most of the validation logics into module.go

As for ReleaseModuleInstance, that is necessary to complete #293,
will be leveraged to make it possible for store to be used for long-running
processes and environment.

### DIFF
--- a/internal/wasm/func_validation.go
+++ b/internal/wasm/func_validation.go
@@ -3,12 +3,13 @@ package internalwasm
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/tetratelabs/wazero/internal/leb128"
 )
 
-// validateFunctionInstance validates the instruction sequence of a function instance.
+// validateFunction validates the instruction sequence of a function.
 // following the specification https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#instructions%E2%91%A2.
 //
 // f is the validation target function instance.
@@ -21,8 +22,10 @@ import (
 //
 // Returns an error if the instruction sequence is not valid,
 // or potentially it can exceed the maximum number of values on the stack.
-func validateFunctionInstance(
-	f *FunctionInstance,
+func validateFunction(
+	functionType *FunctionType,
+	body []byte,
+	localTypes []ValueType,
 	functions []Index,
 	globals []*GlobalType,
 	memories []*MemoryType,
@@ -36,20 +39,20 @@ func validateFunctionInstance(
 	hasTable := len(tables) > 0
 
 	// We start with the outermost control block which is for function return if the code branches into it.
-	controlBloclStack := []*controlBlock{{blockType: f.FunctionType.Type}}
+	controlBloclStack := []*controlBlock{{blockType: functionType}}
 	// Create the valueTypeStack to track the state of Wasm value stacks at anypoint of execution.
 	valueTypeStack := &valueTypeStack{}
 
 	// Now start walking through all the instructions in the body while tracking
 	// control blocks and value types to check the validity of all instructions.
-	for pc := uint64(0); pc < uint64(len(f.Body)); pc++ {
-		op := f.Body[pc]
+	for pc := uint64(0); pc < uint64(len(body)); pc++ {
+		op := body[pc]
 		if OpcodeI32Load <= op && op <= OpcodeI64Store32 {
 			if !hasMemory {
 				return fmt.Errorf("unknown memory access")
 			}
 			pc++
-			align, num, err := leb128.DecodeUint32(bytes.NewBuffer(f.Body[pc:]))
+			align, num, err := leb128.DecodeUint32(bytes.NewBuffer(body[pc:]))
 			if err != nil {
 				return fmt.Errorf("read memory align: %v", err)
 			}
@@ -227,7 +230,7 @@ func validateFunctionInstance(
 			}
 			pc += num
 			// offset
-			_, num, err = leb128.DecodeUint32(bytes.NewBuffer(f.Body[pc:]))
+			_, num, err = leb128.DecodeUint32(bytes.NewBuffer(body[pc:]))
 			if err != nil {
 				return fmt.Errorf("read memory offset: %v", err)
 			}
@@ -237,7 +240,7 @@ func validateFunctionInstance(
 				return fmt.Errorf("unknown memory access")
 			}
 			pc++
-			val, num, err := leb128.DecodeUint32(bytes.NewBuffer(f.Body[pc:]))
+			val, num, err := leb128.DecodeUint32(bytes.NewBuffer(body[pc:]))
 			if err != nil {
 				return fmt.Errorf("read immediate: %v", err)
 			}
@@ -258,14 +261,14 @@ func validateFunctionInstance(
 			pc++
 			switch Opcode(op) {
 			case OpcodeI32Const:
-				_, num, err := leb128.DecodeInt32(bytes.NewBuffer(f.Body[pc:]))
+				_, num, err := leb128.DecodeInt32(bytes.NewBuffer(body[pc:]))
 				if err != nil {
 					return fmt.Errorf("read i32 immediate: %s", err)
 				}
 				pc += num - 1
 				valueTypeStack.push(ValueTypeI32)
 			case OpcodeI64Const:
-				_, num, err := leb128.DecodeInt64(bytes.NewBuffer(f.Body[pc:]))
+				_, num, err := leb128.DecodeInt64(bytes.NewBuffer(body[pc:]))
 				if err != nil {
 					return fmt.Errorf("read i64 immediate: %v", err)
 				}
@@ -280,46 +283,46 @@ func validateFunctionInstance(
 			}
 		} else if OpcodeLocalGet <= op && op <= OpcodeGlobalSet {
 			pc++
-			index, num, err := leb128.DecodeUint32(bytes.NewBuffer(f.Body[pc:]))
+			index, num, err := leb128.DecodeUint32(bytes.NewBuffer(body[pc:]))
 			if err != nil {
 				return fmt.Errorf("read immediate: %v", err)
 			}
 			pc += num - 1
 			switch op {
 			case OpcodeLocalGet:
-				inputLen := uint32(len(f.FunctionType.Type.Params))
-				if l := uint32(len(f.LocalTypes)) + inputLen; index >= l {
+				inputLen := uint32(len(functionType.Params))
+				if l := uint32(len(localTypes)) + inputLen; index >= l {
 					return fmt.Errorf("invalid local index for local.get %d >= %d(=len(locals)+len(parameters))", index, l)
 				}
 				if index < inputLen {
-					valueTypeStack.push(f.FunctionType.Type.Params[index])
+					valueTypeStack.push(functionType.Params[index])
 				} else {
-					valueTypeStack.push(f.LocalTypes[index-inputLen])
+					valueTypeStack.push(localTypes[index-inputLen])
 				}
 			case OpcodeLocalSet:
-				inputLen := uint32(len(f.FunctionType.Type.Params))
-				if l := uint32(len(f.LocalTypes)) + inputLen; index >= l {
+				inputLen := uint32(len(functionType.Params))
+				if l := uint32(len(localTypes)) + inputLen; index >= l {
 					return fmt.Errorf("invalid local index for local.set %d >= %d(=len(locals)+len(parameters))", index, l)
 				}
 				var expType ValueType
 				if index < inputLen {
-					expType = f.FunctionType.Type.Params[index]
+					expType = functionType.Params[index]
 				} else {
-					expType = f.LocalTypes[index-inputLen]
+					expType = localTypes[index-inputLen]
 				}
 				if err := valueTypeStack.popAndVerifyType(expType); err != nil {
 					return err
 				}
 			case OpcodeLocalTee:
-				inputLen := uint32(len(f.FunctionType.Type.Params))
-				if l := uint32(len(f.LocalTypes)) + inputLen; index >= l {
+				inputLen := uint32(len(functionType.Params))
+				if l := uint32(len(localTypes)) + inputLen; index >= l {
 					return fmt.Errorf("invalid local index for local.tee %d >= %d(=len(locals)+len(parameters))", index, l)
 				}
 				var expType ValueType
 				if index < inputLen {
-					expType = f.FunctionType.Type.Params[index]
+					expType = functionType.Params[index]
 				} else {
-					expType = f.LocalTypes[index-inputLen]
+					expType = localTypes[index-inputLen]
 				}
 				if err := valueTypeStack.popAndVerifyType(expType); err != nil {
 					return err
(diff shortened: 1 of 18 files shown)


mathetake:
note to myself: move refcounters to store

mathetake:
I will rework a bit so that we have refcount per moduleInstance rather than per each instance (memory, function, etc) as the latter needlessly makes the code complex while we benefit a little

mathetake:
so I need to do backfilling tests a lot..

mathetake:
some assert_invalid failing but almost there

mathetake:
ok passed spectests.. will polish

internal/wasm/module.go
```diff
@@ -170,6 +173,270 @@ func (m *Module) TypeOfFunction(funcIdx Index) *FunctionType {
...
+	if actualType != expectedType {
+		return fmt.Errorf("const expression type mismatch")
+	}
+	return nil
+}
+
+func (m *Module) buildGlobalInstances(importedGlobals []*GlobalInstance) (globals []*GlobalInstance) {
+	for _, gs := range m.GlobalSection {
+		var gv uint64
+		switch v := executeConstExpression(append(importedGlobals, globals...), gs.Init).(type) {
```
mathetake, line 370: note to myself tomorrow: write test case which uses imported global

mathetake:
so I will stop caring about concurrency in this PR as it bloats further more

mathetake:
so I will stop caring about concurrency in this PR as it bloats further more

internal/wasm/module.go
```diff
@@ -170,6 +173,274 @@ func (m *Module) TypeOfFunction(funcIdx Index) *FunctionType {
...
+			idx := m.SectionElementCount(SectionIDFunction) - 1
+			return fmt.Errorf("invalid function (%d/%d): %v", codeIndex, idx, err)
+		}
+	}
+	return nil
+}
+
+func (m *Module) validateTables(tables []*TableType, globals []*GlobalType) error {
+	if len(tables) > 1 {
+		return fmt.Errorf("multiple tables are not supported")
```
mathetake, line 265: @codefromthecrypt fyi now this is the place we enforce at most one table (before instantiation phase)

mathetake:
ok needs to fix amd64 then..

internal/wasm/func_validation.go
```diff
@@ -944,3 +947,47 @@ type controlBlock struct {
...
+		}
+		return types[index], nil
+	}, r)
+}
+
+// DecodeBlockTypeWithTypeInstances is exported for use in the compiler
+func DecodeBlockTypeWithTypeInstances(types []*TypeInstance, r io.Reader) (*FunctionType, uint64, error) {
+	return decodeBlockTypeImpl(func(index int64) (*FunctionType, error) {
+		if index < 0 || (index >= int64(len(types))) {
+			return nil, fmt.Errorf("invalid block type: %d", index)
```
codefromthecrypt, line 965:
```suggestion
			return nil, fmt.Errorf("type index out of range: %d", index)
```

internal/wasm/func_validation.go
```diff
@@ -944,3 +947,47 @@ type controlBlock struct {
 	isLoop                 bool
 	isIf                   bool
 }
+
+// decodeBlockType is exported for use in the compiler
+func decodeBlockType(types []*FunctionType, r io.Reader) (*FunctionType, uint64, error) {
+	return decodeBlockTypeImpl(func(index int64) (*FunctionType, error) {
+		if index < 0 || (index >= int64(len(types))) {
+			return nil, fmt.Errorf("invalid block type: %d", index)
```
codefromthecrypt, line 955:
```suggestion
			return nil, fmt.Errorf("type index out of range: %d", index)
```

internal/wasm/func_validation.go
```diff
@@ -944,3 +947,47 @@ type controlBlock struct {
 	isLoop                 bool
 	isIf                   bool
 }
+
+// decodeBlockType is exported for use in the compiler
```
codefromthecrypt, line 951:
```suggestion
```

internal/wasm/interpreter/interpreter.go
```diff
@@ -132,6 +132,13 @@ type interpreterOp struct {
 	rs     []*wazeroir.InclusiveRange
 }
 
+func (e *engine) Release(f *wasm.FunctionInstance) error {
```
codefromthecrypt, line 135:
```suggestion
// Release implements wasm.Engine Release
func (e *engine) Release(f *wasm.FunctionInstance) error {
```

internal/wasm/jit/engine.go
```diff
@@ -316,6 +316,15 @@ func (c *callFrame) String() string {
 	)
 }
 
+func (e *engine) Release(f *wasm.FunctionInstance) error {
```
codefromthecrypt, line 319:
```suggestion
// Release implements wasm.Engine Release
func (e *engine) Release(f *wasm.FunctionInstance) error {
```

internal/wasm/jit/jit_amd64.go
```diff
@@ -5414,27 +5414,19 @@ func (c *amd64Compiler) initializeModuleContext() error {
...
-		resolveTableInstanceAddressFromPointer.As = x86.AMOVQ
-		resolveTableInstanceAddressFromPointer.To.Type = obj.TYPE_REG
-		resolveTableInstanceAddressFromPointer.To.Reg = tmpRegister
-		resolveTableInstanceAddressFromPointer.From.Type = obj.TYPE_MEM
-		resolveTableInstanceAddressFromPointer.From.Reg = tmpRegister
-		c.addInstruction(resolveTableInstanceAddressFromPointer)
-
-		// At this point, tmpRegister holds the address of ModuleIntance.Tables[0].
-		// So we are ready to read and put the first item's address stored in Tables[0].Table.
+		// At this point, tmpRegister holds the address of ModuleIntance.TableInstance.
```
codefromthecrypt, line 5428: s/Intance/Instance/g

internal/wasm/store.go
```diff
@@ -53,9 +54,11 @@ type (
 		//  maximumFunctionTypes represents the limit on the number of function types in a store.
 		// Note: this is fixed to 2^27 but have this a field for testability.
 		maximumFunctionTypes int
-		// maximumGlobals is the maximum number of globals that can be declared in a module.
-		// Note: this is fixed to 2^27 but have this a field for testability.
-		maximumGlobals int
+
+		releasedFunctionAddress []FunctionAddress
```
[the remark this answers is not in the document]
codefromthecrypt: add docs or a general TODO for what these are going to be reshaped as

mathetake:
some of the funcs lacks godoc but eagerly mark this ready for review

internal/wasm/store.go
```diff
@@ -153,6 +158,10 @@ type (
 		Type *GlobalType
 		// Val holds a 64-bit representation of the actual value.
 		Val uint64
+
+		// address is the globaladdr(https://www.w3.org/TR/wasm-core-1/#syntax-globaladdr) of this global instance.
+		// In other words, this equals the index of this global instance in store.Globals.
```
codefromthecrypt, line 163: probably better to call this `index` and use the existing type `wasm.Index` vs the confusing uint64 which is at best an implementation detail of additions.

```suggestion
		// address is the index of this global instance in Store.Globals.
		// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-globaladdr
```

internal/wasm/store.go
```diff
@@ -153,6 +158,10 @@ type (
 		Type *GlobalType
 		// Val holds a 64-bit representation of the actual value.
 		Val uint64
+
+		// address is the globaladdr(https://www.w3.org/TR/wasm-core-1/#syntax-globaladdr) of this global instance.
```
codefromthecrypt, line 162:
```suggestion
```

internal/wasm/store.go
```diff
@@ -172,6 +181,10 @@ type (
 		Max   *uint32
 		// Currently fixed to 0x70 (funcref type).
 		ElemType byte
+
+		// address is the tableaddr(https://www.w3.org/TR/wasm-core-1/#syntax-tableaddr) of this table instance.
+		// In other words, this equals the index of this table instance in store.Tables.
```
codefromthecrypt, line 186: probably better to call this `index` and use the existing type `wasm.Index` vs the confusing uint64 which is at best an implementation detail of additions.

```suggestion
		// address is the index of this table instance in Store.Tables.
		// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-tableaddr
```

internal/wasm/store.go
```diff
@@ -172,6 +181,10 @@ type (
 		Max   *uint32
 		// Currently fixed to 0x70 (funcref type).
 		ElemType byte
+
+		// address is the tableaddr(https://www.w3.org/TR/wasm-core-1/#syntax-tableaddr) of this table instance.
```
codefromthecrypt, line 185:
```suggestion
```

internal/wasm/store.go
```diff
@@ -196,11 +209,24 @@ type (
 		Buffer []byte
 		Min    uint32
 		Max    *uint32
+
+		// address is the memoryaddr(https://www.w3.org/TR/wasm-core-1/#syntax-tableaddr) of this memory instance.
+		// In other words, this equals the index of this memory instance in store.Memories.
```
codefromthecrypt, line 214: probably better to call this `index` and use the existing type `wasm.Index` vs the confusing uint64 which is at best an implementation detail of additions.

```suggestion
		// address is the index of this memory instance in Store.Memories.
		// See See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-tableaddr
```

internal/wasm/store.go
```diff
@@ -196,11 +209,24 @@ type (
...
+		address memoryAddress
 	}
 
 	// FunctionAddress is funcaddr (https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-funcaddr),
```
codefromthecrypt, line 218: Same comment vs above.. probably you can simplify docs by linking to the other.

Ex.

```
// XXIndex holds a copy of the last value of xx.index.
```

codefromthecrypt (commented):
adding some comments because the uint64 thing shouldn't survive another PR

internal/wasm/store.go
codefromthecrypt: actually these look like types, so just delete them

wasm.go
```diff
@@ -74,8 +74,12 @@ type ModuleConfig struct {
 //
 // This is used to pre-flight check and cache the module for later instantiation.
 func (m *ModuleConfig) Validate() (err error) {
-	_, err = decodeModule(m)
-	return err
+	mod, err := decodeModule(m)
+	if err != nil {
+		return
+	}
+	err = mod.Validate()
```
codefromthecrypt, line 81:
```suggestion
	// TODO: decoders should validate before returning, as that allows
	// them to err with the correct source position.
	err = mod.Validate()
```

tests/spectest/spec_test.go
```diff
@@ -435,6 +418,17 @@ func runTest(t *testing.T, newEngine func() wasm.Engine) {
 	}
 }
 
+func requireInstantiationFailure(t *testing.T, store *wasm.Store, buf []byte, msg string) {
```
codefromthecrypt, line 421:
```suggestion
func requireInstantiationError(t *testing.T, store *wasm.Store, buf []byte, msg string) {
```

internal/wazeroir/compiler.go
```diff
@@ -246,7 +246,7 @@ operatorSwitch:
 		c.controlFrames.push(frame)
 
 	case wasm.OpcodeLoop:
-		bt, num, err := wasm.DecodeBlockType(c.f.ModuleInstance.Types,
+		bt, num, err := wasm.DecodeBlockTypeWithTypeInstances(c.f.ModuleInstance.Types,
```
codefromthecrypt, line 249: nit since DecodeBlockType isn't exported, you could sneaky use the same name

internal/wasm/store_test.go
```diff
@@ -327,186 +286,641 @@ func TestStore_getTypeInstance(t *testing.T) {
...
+			expectedAddress := prevMaxAddr + 1 + i
+			f := &FunctionInstance{}
+			s.addFunctionInstances(f)
+
+			// After adding function intance to store, an funcaddr must be assigned.
+			require.Equal(t, expectedAddress, f.Address)
+		}
 	})
-	t.Run("invalid constant expression", func(t *testing.T) {
+	t.Run("use released address", func(t *testing.T) {
```
codefromthecrypt, line 424:
```suggestion
	t.Run("reuse released address", func(t *testing.T) {
```

internal/wasm/store.go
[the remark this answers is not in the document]
codefromthecrypt: in this TODO make sure it notes this is currently unbounded. Either mention we have to bound it, or document how it is implicitly bounded

internal/wasm/module.go
```diff
@@ -170,6 +173,274 @@ func (m *Module) TypeOfFunction(funcIdx Index) *FunctionType {
...
+	if m.NameSection != nil {
+		functionNames = m.NameSection.FunctionNames
+	}
+
+	importCount := m.ImportFuncCount()
+	n, nLen := 0, len(functionNames)
+	for codeIndex := range m.FunctionSection {
+		funcIdx := Index(importCount + uint32(len(functions)))
+		// Seek to see if there's a better name than "unknown"
+		name := "unknown"
```
codefromthecrypt, line 404: probably better to default to $funcIdx and set that after the loop so we don't allocate unless we found no name.

codefromthecrypt (approved):
Good work!

I would probably move validate* functions to validate.go and validate_test.go even if they have a module target. At some point we will export more of these to ensure source context is correct on failure.

internal/wasm/store.go
```diff
@@ -212,10 +238,120 @@ type (
 const (
 	maximumFunctionAddress = 1 << 27
 	maximumFunctionTypes   = 1 << 27
-	maximumGlobals         = 1 << 27
-	maximumValuesOnStack   = 1 << 27
 )
 
+// newModuleInstance bundles all the instances for a module and creates a new module isntance.
```
nullpo-head, line 251:
```suggestion
// newModuleInstance bundles all the instances for a module and creates a new module instance.
```

internal/wasm/store.go
```diff
@@ -246,101 +382,238 @@ func NewStore(ctx context.Context, engine Engine) *Store {
 		Engine:                 engine,
 		maximumFunctionAddress: maximumFunctionAddress,
 		maximumFunctionTypes:   maximumFunctionTypes,
-		maximumGlobals:         maximumGlobals,
 	}
 }
 
+// checkFuncAddrOverflow checks if there would be too many function instances in a store.
+func (s *Store) checkFuncAddrOverflow(newInstanceNum int) error {
+	if len(s.Functions)+newInstanceNum > int(s.maximumFunctionAddress) {
```
nullpo-head: Is there guarantee that `len(s.Functions)+newInstanceNum` doesn't overflow and doesn't negative value? e.g. it is validated that `newInsanceNum` is small enough beforehand. If not, this add may overflow and the validation can be bypassed, so may be better to replace it with `len(s.Functions) > int(s.maximumFunctionAddress) - newInstaneNum`

nullpo-head (approved):

internal/wasm/store.go
mathetake: good point lemme

mathetake:
Thanks for the look!

PR #295, opened 2022-02-25
Stops proliferation of MVP jargon

codefromthecrypt:
MVP was a term used by WebAssembly insiders when ramping up to the 1.0
spec. While these folks still use that term it is confusing and
unnecessary way to qualify a W3C version. Here are some of the problems:

* MVP does not match a W3C published URL
* MVP does not match a git tag on the spec repo
* MVP was a work in progress, so there are text that say "not in MVP"
  which ended up in 1.0 (as MVP became more than it was).
* MVP is jargon to people who don't know that stands for Minimum Viable Product.

This stops this practice and instead uses the W3C 1.0 Draft version
instead: 20191205

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -77,20 +77,20 @@ https://github.com/bytecodealliance/wasmtime/blob/v0.29.0/crates/lightbeam/src/m
 
 ## Implementation limitations
 
-WebAssembly 1.0 (MVP) specification allows runtimes to [limit certain aspects of Wasm module or execution](https://www.w3.org/TR/wasm-core-1/#a2-implementation-limitations).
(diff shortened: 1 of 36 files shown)


mathetake (approved):
Like it

codefromthecrypt:
ps I added this to hopefully align the spec repo with the actual published URL, but since the webassembly repositories are spread out and a bit confusing anyway, I think it is best to go with W3C permalinks instead of git tags where possible https://github.com/WebAssembly/spec/issues/1425

nullpo-head (approved):

codefromthecrypt:
thanks for the support!

PR #297, opened 2022-02-28
Folds ImportKind and ExportKind into ExternType

codefromthecrypt:
This centralizes our management of external types by choosing the word
"external" also used in the spec instead of defining the same constants
for imports vs exports. One advantage is more coherent reference
searching as constants are no longer split. Another is future work can
use a single "kind" type to manage namespace length checks.

### DIFF
--- a/internal/wasm/module.go
+++ b/internal/wasm/module.go
@@ -148,7 +148,7 @@ func (m *Module) TypeOfFunction(funcIdx Index) *FunctionType {
 	}
 	funcImportCount := Index(0)
 	for i, im := range m.ImportSection {
-		if im.Kind == ExternalKindFunc {
+		if im.Type == ExternTypeFunc {
 			if funcIdx == Index(i) {
 				if im.DescFunc >= typeSectionLength {
 					return nil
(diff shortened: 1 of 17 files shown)


internal/wasm/module.go
```diff
@@ -257,12 +257,12 @@ type ConstantExpression struct {
...
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#binary-export
 type Export struct {
-	Kind ExportKind
+	Kind ExternalKind
 	// Name is what the host refers to this definition as.
 	Name string
 	// Index is the index of the definition to export, the index namespace is by Kind
-	// Ex. If ExportKindFunc, this is an index to ModuleInstance.Functions
-	Index Index // TODO: can you export an import? If so, rewrite the doc ^^ to say "index namespace"
+	// Ex. If ExternalKindFunc, this is an index in the function index namespace.
```
codefromthecrypt, line 264:
```suggestion
	// Ex. If ExternalKindFunc, this is an position in the function index namespace.
```

mathetake (approved):
Shortend "ExternKind"  would also work but yeah either works for me

nullpo-head (approved):
looks good to me

codefromthecrypt:
Went with ExternType as that's the name in the spec!

codefromthecrypt:
thanks for the look, compadres

PR #298, opened 2022-02-28
Enforces at most one table by type

codefromthecrypt:
This tries to match the logic of memory, but fails due to some skew in JIT.

PR #299, opened 2022-02-28
Adds import count functions to unlock shared text decoding

codefromthecrypt:
Right now, it is a lot of copy/paste to decode inlined import
abbreviation with the same parser as module-defined ones. The primary
reason for this is that the import section is only guaranteed before
module-defined *after* expanding abbreviations. For example,

`(func (import "foo" "bar") ...` follows imports and may be followed by
any number of abbreviations. This means you can't know the end of
imports until you see the first `(func` which doesn't abbreviate one.

While certain external types, namely memory and table, can only be
imported once, this also applies to globals.

This change makes counting the current imports a `func() uint32` so that
it can be decoupled and used for any external type. Other PRs can
simplify memory and table which can only be imported once, but the
general signatures here will still work.

### DIFF
--- a/internal/wasm/counts.go
+++ b/internal/wasm/counts.go
@@ -1,0 +1,81 @@
+package internalwasm
+
+import "fmt"
+
+// ImportFuncCount returns the possibly empty count of imported functions. This plus SectionElementCount of
+// SectionIDFunction is the size of the function index namespace.
+func (m *Module) ImportFuncCount() uint32 {
+	return m.importCount(ExternTypeFunc)
+}
+
+// ImportTableCount returns the possibly empty count of imported tables. This plus SectionElementCount of SectionIDTable
+// is the size of the table index namespace.
+func (m *Module) ImportTableCount() uint32 {
+	return m.importCount(ExternTypeTable) // TODO: once validation happens on decode, this is zero or one.
+}
+
+// ImportMemoryCount returns the possibly empty count of imported memories. This plus SectionElementCount of
+// SectionIDMemory is the size of the memory index namespace.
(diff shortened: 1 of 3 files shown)


internal/wasm/counts.go
```diff
@@ -0,0 +1,81 @@
...
+	return
+}
+
+// SectionElementCount returns the count of elements in a given section ID
+//
+// For example...
+// * SectionIDType returns the count of FunctionType
+// * SectionIDCustom returns one if the NameSection is present
+// * SectionIDExport returns the count of unique export names
+func (m *Module) SectionElementCount(sectionID SectionID) uint32 { // element as in vector elements!
```
codefromthecrypt, line 46: this and tests are the same as before

codefromthecrypt (commented):
I'll rebase #298 after this, but this is mostly for #215.

Note: I pulled this into a separate file as module.go was getting big, and also pulling this out helps drift control

internal/wasm/counts_test.go
```diff
@@ -0,0 +1,298 @@
...
+	for _, tt := range tests {
+		tc := tt
+
+		t.Run(tc.name, func(t *testing.T) {
+			require.Equal(t, tc.expected, tc.input.ImportFuncCount())
+		})
+	}
+}
+
+// TODO: once we fix up-front validation, this only needs to check zero or one
```
codefromthecrypt, line 54: ps almost certainly I will delete this and the function for counting memory once there can be only one and it is validated during decode. I'm only putting this in now as there are no such guarantees yet. deleting this later will help show how useful centralized decode logic is!

mathetake (approved):
thanks I will reuse this for #294 as well!

PR #300, opened 2022-02-28
jit: support Windows system

### LINKED ISSUE #269: Support MMap on Windows for JIT engine
An issue in using the JIT engine on Windows, is the lack of support of MMap, after looking at MSDN and GitHub, we can support MMap in the same fashion as this portable library - https://github.com/edsrzf/mmap-go/blob/main/mmap_windows.go

We can either introduce the portable mmap library as a dependency, or reimplement the features.

wdvxdr1123:
Fixes #269


wdvxdr1123:
@mathetake @codefromthecrypt Hello, I implemented the jit by using Win32 memory api VirtualAlloc. Currently this code imports a new package(golang/x/sys/windows), I can replace it with std syscall instead if you like.

codefromthecrypt (approved):
Thanks for the help. I'm going to approve without making small comments as @mathetake  later will have some slight adjustments to all os/arch for MMap. If we run into any problems, maybe will ask your help, but this is fantastic progress especially first change.

Welcome to wazero @wdvxdr1123!

PR #302, opened 2022-03-01
text: changes string value of externtype "mem" to "memory"

codefromthecrypt:
This changes the string value of "mem" to "memory" so that the same
constants currently used for error context can also be used for field
parsing dispatch.

### DIFF
--- a/internal/wasm/module_test.go
+++ b/internal/wasm/module_test.go
@@ -68,7 +68,7 @@ func TestExternTypeName(t *testing.T) {
 	}{
 		{"func", ExternTypeFunc, "func"},
 		{"table", ExternTypeTable, "table"},
-		{"mem", ExternTypeMemory, "mem"},
+		{"mem", ExternTypeMemory, "memory"},
 		{"global", ExternTypeGlobal, "global"},
 		{"unknown", 100, "0x64"},
 	}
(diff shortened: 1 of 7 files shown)


internal/wasm/module_test.go
```diff
@@ -68,7 +68,7 @@ func TestExternTypeName(t *testing.T) {
 	}{
 		{"func", ExternTypeFunc, "func"},
 		{"table", ExternTypeTable, "table"},
-		{"mem", ExternTypeMemory, "mem"},
+		{"mem", ExternTypeMemory, "memory"},
```
codefromthecrypt, line 71: this is the only impact and except in text parsing, this is only used for late errors on invalid memory

internal/wasm/text/decoder.go
```diff
@@ -91,6 +91,10 @@ type moduleParser struct {
 
 	// unresolvedExports holds any exports whose type index wasn't resolvable when parsed.
 	unresolvedExports map[wasm.Index]*wasm.Export
+
+	// field counts can be different from the count in a section when abbreviated imports exist. To give an accurate
+	// errorContext, we count explicitly.
+	fieldCountFunc, fieldCountMemory uint32
```
codefromthecrypt, line 97: these will be moved in a later PR to `fieldCount [4]uint32`, so I didn't doc them individually now.

mathetake (approved):

codefromthecrypt:
thanks for the look @mathetake !

PR #305, opened 2022-03-01
Limit visibility on fields of store to clarify where to lock

mathetake:
This commit unexports all the fields of store so that we could
clarify where to take lock for making store goroutine-safe.


mathetake:
well, I will narrow the scope to unexporting store's fields and renaming to InstantiatedModule

mathetake:
ok tomorrow I will backfill the tests and docs, but yeah now spectests and wasi tests only uses the public APIs

internal/wasm/module_context.go
```diff
@@ -61,13 +61,22 @@ func (c *ModuleContext) Memory() publicwasm.Memory {
...
-	exp, err := c.Module.GetExport(name, ExternTypeFunc)
+	exp, err := c.module.getExport(name, ExternTypeFunc)
 	if err != nil {
 		return nil
 	}
 	return &exportedFunction{module: c, function: exp.Function}
 }
 
+// Function implements wasm.ModuleContext Global
+func (c *ModuleContext) Global(name string) publicwasm.Global {
```
mathetake, line 72: this necessary for spectests

internal/wasm/module_instance_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestModuleInstance_Memory(t *testing.T) {
+	t.Skip() // TODO fix
```
mathetake, line 11: TODO tomorrow

internal/wasm/store_host_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/wazero/wasm"
+)
+
+func TestStore_AddHostFunction(t *testing.T) {
+	t.Skip() // TODO: fix
```
mathetake, line 14: TODO tomorrow

internal/wasm/store.go
```diff
@@ -614,46 +330,35 @@ func (s *Store) addMemoryInstance(m *MemoryInstance) {
...
+	if m, ok := s.moduleContexts[moduleName]; !ok {
 		return nil
 	} else {
-		return &ModuleExports{s, m}
+		return m
 	}
 }
 
-// ModuleExports implements wasm.ModuleExports
-type ModuleExports struct {
```
mathetake, line 635: TODO tomorrow: revert this

wasm.go
```diff
@@ -157,59 +157,16 @@ func decodeModule(module *ModuleConfig) (m *internalwasm.Module, err error) {
 	return
 }
 
-// HostModuleConfig are WebAssembly 1.0 (20191205) exports from the host bound to a module name used by InstantiateHostModule.
-type HostModuleConfig struct {
```
codefromthecrypt, line 161: ps I had intentionally kept "config" types out of the /wasm and /wasi packages because they are mutable and probably not very safe to share to internal code.

wasm/wasm.go
```diff
@@ -66,16 +66,17 @@ type Store interface {
...
+//
+// TODO: rename this to InstantiatedModule per https://github.com/tetratelabs/wazero/issues/293.
 type ModuleExports interface {
 	// Memory returns a memory exported from this module or nil if it wasn't.
-	//
-	// Note: WASI modules require exporting a Memory named "memory". This means that a module successfully initialized
-	// as a WASI Command or Reactor will never return nil for this name.
-	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/design/application-abi.md#current-unstable-abi
-	Memory(name string) Memory
+	Memory() Memory
```
codefromthecrypt, line 73: still need the memory name I think in order to validate WASI rules where it is named "export" remember these are exported symbols (external types)

wasm/wasm.go
```diff
@@ -107,6 +108,11 @@ type Function interface {
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
 }
 
+type Global interface {
+	Value() uint64
```
codefromthecrypt, line 112: What I was thinking was `Get() uint64` and `Set(uint64) bool` where the latter can return bool false which means it is unsupported for any reason including configuration or the global isn't mutable.

```suggestion
	Get() uint64
```

internal/wasm/store_host.go
```diff
@@ -0,0 +1,147 @@
...
+	t := newTableInstance(config.Min, config.Max)
+
+	// TODO: check if the module already has memory, and if so, returns error.
+	m.TableInstance = t
+	s.addTableInstance(t)
+
+	return m.addExport(config.Name, &ExportInstance{Type: ExternTypeTable, Table: t})
+}
+
+func (s *Store) exportHostMemoryInstance(m *ModuleInstance, config *publicwasm.HostModuleConfigMemory) error {
```
codefromthecrypt, line 131: yeah this tight coupling makes me nervous as it is easy to accidentally hold a reference which allows the user to muck with something at runtime. Particularly memory is dead simple to configure, so is a no brainer (min/max).

wasm/wasm.go
```diff
@@ -233,3 +239,67 @@ func EncodeF64(input float64) uint64 {
...
+	//		return x + y
+	//	}
+	//
+	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#host-functions%E2%91%A2
+	Functions map[string]interface{}
+
+	// TODO
+	Globals map[string]*HostModuleConfigGlobal
+	// TODO
+	Table *HostModuleConfigTable
```
codefromthecrypt, line 288: still need the name both here and below. Even if there is a limit of only one, the structure is more coherent if it is a map. especially consider this is likely to be encoded as json or yaml.

wasm/wasm.go
```diff
@@ -233,3 +239,67 @@ func EncodeF64(input float64) uint64 {
...
+	//		x, _ := ctx.Memory().ReadUint32Le(offset)
+	//		y, _ := ctx.Memory().ReadUint32Le(offset + 4) // 32 bits == 4 bytes!
+	//		return x + y
+	//	}
+	//
+	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#host-functions%E2%91%A2
+	Functions map[string]interface{}
+
+	// TODO
+	Globals map[string]*HostModuleConfigGlobal
```
codefromthecrypt, line 286: I had hoped to hide how the current global configuration was implemented as it is messy. That's why I didn't implement #279, yet. I really had hoped we could defer this, and I hope you can reconsider conflating this with lock hunting. You can see you are already tempted to conflating internal details with the public config api just to do both at the same time. It is far better to separate the topics. public api can and probably should have nothing to do with the current internal impl. rushing a public api is a bad idea.

The main idea I had but didn't implement yet was that the user shouldn't have any of this mess in their code. Simply they can configure either a global or a mutable global. so that means two maps. `map[string]interface{}`: `Globals` and `MutableGlobals` this makes it very hard to get wrong. the value is only one of the existing supported value types which we already have code to convert from go types. the value of the ex. uint64 value is the initial value. There's no difference between unset and zero we don't support that. This makes the only weird case besides someone passing the wrong type (which they already can do in `Functions` so that's nothing new) is declaring the same export name in both Globals and MutableGlobals. that's easy enough to document so that's the best I could do with no less than 4 days thinking about this field.

codefromthecrypt (commented):
really not a fan of adding a public api at the same time as clarifying where to lock. the api deserves more than being rushed like this

wasm/wasm.go
codefromthecrypt: ps the `Get` `Set` API of globals won't look like ^^ for the same reason the API to access exported functions don't look like how the host defines them. Each public type especially the even more weird "table" should be done in a different PR. Internals can change 4times a day and we shouldn't care, but we must care about the public API, the what, how, why and for who on each and every symbol

mathetake:
Yeah you got me and let me rework in the way that there’s no public API change

mathetake:
This is literally getting out of control

mathetake:
ok reworked and removed all the public API change..

mathetake:
mostly code move and mechanical renaming stuff plus spectest and wasi fixes

mathetake:
mark as unready until the public API refactoring done as it might affect the internal code change here

mathetake:
so I will redo this 😄

PR #309, opened 2022-03-02
Isolates CGO comparisons to their own go.mod file

### LINKED ISSUE #306: Consider placing separate go.mod under `tests/*`
To prevent users from worrying that wazero would use CGO as currently `go get wazero` adds `wasmtime` and `wasmer` dependencies into their go.sum.

In order to achieve this, we have to organize the codes in `tests/*` so that they don't use internal.

codefromthecrypt:
Fixes #306


codefromthecrypt:
thanks for the look and the idea @mathetake

PR #310, opened 2022-03-02
Adds Global API and implements for module-defined globals

codefromthecrypt:
Global is a WebAssembly 1.0 (20191205) global exported from an
instantiated module (wazero.Runtime NewModule).

Ex. If the value is not mutable, you can read it once:
```go
offset := module.Global("memory.offset").Get()
```

Globals are allowed by specification to be mutable. However, this can
be disabled by configuration. When in doubt, safe cast to find out if
the value can change.

Ex.
```go
offset := module.Global("memory.offset")
if _, ok := offset.(wasm.MutableGlobal); ok {
	value can change
} else {
	value is constant
}
```

See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#globals%E2%91%A0

### DIFF
--- a/internal/wasm/global.go
+++ b/internal/wasm/global.go
@@ -1,0 +1,147 @@
+package internalwasm
+
+import (
+	"fmt"
+
+	publicwasm "github.com/tetratelabs/wazero/wasm"
+)
+
+type mutableGlobal struct {
+	g *GlobalInstance
+}
+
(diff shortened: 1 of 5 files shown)


nullpo-head (approved):

codefromthecrypt:
thanks for the look and sorry was so late landing this. I should have done this Friday

PR #311, opened 2022-03-02
jit(arm64): restore callee-saved registers on exits from native code 

### LINKED ISSUE #308: Default JIT when arm64 or amd64
When we sort out CI, let's set JIT as default.

mathetake:
fixes #308


config_supported.go
```diff
@@ -6,5 +6,5 @@ package wazero
 // NewRuntimeConfig returns NewRuntimeConfigInterpreter
 // TODO: switch back to NewRuntimeConfigJIT https://github.com/tetratelabs/wazero/issues/308
```
codefromthecrypt, line 7:
```suggestion
```

codefromthecrypt (commented):
When ready... :D

codefromthecrypt (approved):
Wow keen eye fixing this one!

examples/testdata/host_func.go
```diff
@@ -42,6 +42,6 @@ func base64OnString(num uint32) {
 	for i := uint32(0); i < num; i++ {
 		msg := getRandomString()
 		encoded := base64.StdEncoding.EncodeToString([]byte(msg))
-		fmt.Printf("base64 encoded string '%s'\n", encoded[:10])
```
codefromthecrypt, line 45: anything interesting here? except maybe the massive reduction in wasm :D

mathetake: oh nothing special but funny that after rebuild amd64 has started failing on this due to "out of range" so I just deleted :D

mathetake: it seems like a TinyGo's bug as I see this succeeds after changing this to 
```
		v := encoded[:10]
		fmt.Printf("base64 encoded string '%s'\n", v)
```

mathetake: something buggy on inline slicing :D

PR #312, opened 2022-03-02
Stubs WASI functions for GrainLang

codefromthecrypt:
This stubs all of the known WASI functions so that runtimes like Grain can compile even if they don't use them.

See:

https://github.com/meshx-org/meshx-old/blob/197aa5f7f84f4decc9587cf3477ba378ee68778f/crates/meshx_wasm_runner/src/syscalls/wasi.rs
https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md
https://github.com/pgavlin/warp/blob/2bac1014889b11205d410a6c0930b036067c0e61/wasi/module_definition.go
https://github.com/ziglang/zig/blob/master/lib/std/os/wasi.zig


codefromthecrypt:
picking this back up a while

codefromthecrypt:
backfilled stub tests and now will add more stub

mathetake (approved):

PR #313, opened 2022-03-02
Wasi backfill docs/tests of path_open, fd_seek, and fd_fdstat_get

nullpo-head:
This PR backfills the documents of `path_open`, `fd_seek`, and `fd_fdstat_get`, and backfills the test of `path_open`. I didn't write tests of `fd_seek` and `fd_fdstat_get` because they are almost just a stub.


internal/wasi/wasi_test.go
```diff
@@ -958,7 +959,168 @@ func createFile(t *testing.T, path string, contents []byte) (*memFile, *MemFS) {
...
+				path:    ".",
+				fileSys: memFS,
+			},
+		}
+	})
+
+	tests := []struct {
+		name                                                                                        string
+		fd, lookupFlags, path, pathLen, oFlags, fsRightsBase, fsRightsInheriting, fdFlags, resultFD uint64
+		memory                                                                                      []byte
```
codefromthecrypt, line 1064: I don't think we need to test partial writes to memory unless you are using this to know which position in the source you got to (due to multiple possibility of the same code) if this is why, maybe comment it.

internal/wasi/wasi.go
```diff
@@ -1230,14 +1294,14 @@ func (a *wasiAPI) PathLink(ctx wasm.ModuleContext, oldFd, oldFlags, oldPath, old
...
+func (a *wasiAPI) PathOpen(ctx wasm.ModuleContext, fd, dirflags, pathString, pathLen, oflags uint32, fsRightsBase,
+	fsRightsInheriting uint64, fdflags, resultOpenedFD uint32) (errno wasi.Errno) {
 	dir, ok := a.opened[fd]
 	if !ok || dir.fileSys == nil {
-		return wasi.ErrnoInval
+		return wasi.ErrnoBadf
 	}
 
-	b, ok := ctx.Memory().Read(path, pathLen)
+	b, ok := ctx.Memory().Read(pathString, pathLen)
```
codefromthecrypt, line 1304: nit: not sure the suffix is necessary as we use xx xxLen elsewhere

internal/wasi/wasi_test.go
```diff
@@ -1128,9 +1129,9 @@ func TestSnapshotPreview1_FdWrite(t *testing.T) {
 			errno := tc.fdWrite()(ctx(mem), fd, iovs, iovsCount, resultSize)
 			require.Equal(t, wasi.ErrnoSuccess, errno)
 
-			actual, ok := mem.Read(0, uint32(len(expectedMemory)))
+			actualMemory, ok := mem.Read(0, uint32(len(expectedMemory)))
```
codefromthecrypt, line 1132: nit unless there are multiple "actual" we don't need to suffix it. you'll notice 'actual' is a common name in assertions regardless of the type it points to

same for expected 

ex.
```
func Equal(t TestingT, expected interface{}, actual interface{}
```

codefromthecrypt: my guess is you did this because above we say `expectedMemory` and `initialMemory` etc. if we choose to change actual to actualMemory, do it everywhere. doing something in only a couple places out of the dozens in the same file leads to mental glitches in future maintenance.

internal/wasi/wasi_test.go
nullpo-head: wow, thanks for the feedback early! Could you tell me more about the meaning of "test partial writes"? Sorry, maybe I'm not following you.

internal/wasi/wasi_test.go
nullpo-head: Thanks, that totally makes sense and your guess is right. Let me fix it.

internal/wasi/wasi_test.go
```diff
@@ -1300,7 +1301,165 @@ func TestSnapshotPreview1_PathLink(t *testing.T) {
...
+		'?',                          // `path` is after this
+		'w', 'a', 'z', 'e', 'r', 'o', // path
+		'?', // `resultOpenedFD` is after this
+	}
+	expectedMemory := append(
+		initialMemory,
+		4, 0, 0, 0, // resultOpenedFD
+		'?',
+	)
+	expectedFd := uint32(4) // arbitrary expected FD
```
nullpo-head, line 1324: I'm thinking whether we should consistently use `FD` or `Fd`. In general we should use `FD`, like `URL`  or `HTTP` as Golang community often follows https://github.com/golang/go/wiki/CodeReviewComments#initialisms. On the other hand, Golang's std library uses the word `Fd` https://pkg.go.dev/os#File.Fd. Plus considering that `FdClose` or other `FdSomething` looks more readable than `FDClose`, I'd like to say `Fd` is better. wdyt?

codefromthecrypt: for translated things, I think better to follow the delimiter rules as ex fd_close -> FdClose is more obvious and mechanical than FDClose. Also bear in mind these functions aren't public.

For things we define that aren't mapped to something external, probably we would tend to use Go rules which as you note aren't followed always. I often am in the same boat, where I'm looking to see how often someone does HTTPThing vs HttpThing and if it is a clear majority or not 😅

codefromthecrypt: one reason I've been using normal case format conversion for things from the spec as there is so much to translate and sometimes it is subtle, like something ending in _s. We end up being near code generators and to manage the scale of that is easier vs having to think if something is really an abbreviation or not. ex iovs? To not have to think carefully *1000 things is its own benefit and if someone gets really upset it would only be internal code so we can change that any time.

Basically we can save the higher scrutiny for code outside /internal as that's the stuff if you change breaks api (or is it API?) 😎

nullpo-head: Adopted "Fd" for semi-automatic conversions, and "FD" for normal Go names. Thanks!

codefromthecrypt (approved):
looks great!

PR #314, opened 2022-03-03
Simplify host_call example

mathetake:
no longer output the base64 encoded string into stdout instead, we write it into the buffer and verify.


mathetake:
No longer output the garbase "base64 encoded string: ..." in non amd64 tests

examples/host_func_test.go
```diff
@@ -19,22 +22,22 @@ type testKey struct{}
 var hostFuncWasm []byte
 
 func Test_hostFunc(t *testing.T) {
-	allocateBuffer := func(context.Context, uint32) uint32 {
```
codefromthecrypt, line 22: actually I intended this to show how you can stub a function without wazero types.. replacing with `wasm.Function` simplifies for us, but doesn't show how you can do this..

examples/host_func_test.go
```diff
@@ -47,34 +50,34 @@ func Test_hostFunc(t *testing.T) {
...
+	stdout := bytes.NewBuffer(nil)
+	_, err = r.NewHostModule(wazero.WASISnapshotPreview1WithConfig(&wazero.WASIConfig{Stdout: stdout}))
 	require.NoError(t, err)
 
 	module, err := wazero.StartWASICommandFromSource(r, hostFuncWasm)
 	require.NoError(t, err)
 
-	allocateBufferFn := module.Function("allocate_buffer")
-
-	// Implement the function pointer. This mainly shows how you can decouple a module function dependency.
```
codefromthecrypt, line 67: yeah this was the point, not just decoupling the dependency from a chicken-egg, but also from its signature requiring wazero types. Maybe I should have clarified that..

codefromthecrypt (approved):
I'm ok with the change, but think about it before merging!

examples/host_func_test.go
```diff
@@ -19,22 +22,22 @@ type testKey struct{}
...
 		require.Equal(t, int64(12345), contextValue)
 
-		const bufferSize = 10000 // force memory space grow to ensure eager failures on missing setup
-		// Allocate the in-Wasm memory region so we can store the generated string.
-		// Note that this is recursive call. That means that this is the VM function call during the VM function call.
-		// More precisely, we call test.base64 (in Wasm), and the function in turn calls this get_random_string function,
-		// and we call test.allocate_buffer (in Wasm) here: host->vm->host->vm.
-		offset := allocateBuffer(ctx.Context(), bufferSize)
+		const bufferSize = 10
+		res, err := allocateInWasmBufferFn.Call(ctx.Context(), uint64(bufferSize))
```
nullpo-head, line 38: Aside from the Adiran's discussion on showing how to stub, if we don't show such stub and just show calling a wasm function here, then getting the function by `ctx.Function("allocate_buffer")` inside `getRandomBytes` would be current simpler API.

codefromthecrypt: yep agreed. there's no point to play "the shell game" unless we are trying to show how to unwind deps.

mathetake:
thanks, I reverted the unwinding stuff!

PR #316, opened 2022-03-03
jit: reduce the cost of each function call

mathetake:
Previously, we allocate stacks per engine, therefore there were no
allocation overhead for each function call. Since #287, we create
stack per store.Call and resulted in increasing the cost of each
store.Call. This commit reduces the pre-allocation stack size
so that we could reduce the each store.Call.


codefromthecrypt (approved):
more reasonable!

PR #317, opened 2022-03-03
Sets floor go version to 1.17

codefromthecrypt:
We temporarily supported go 1.16 to allow migration. By setting the
floor to 1.17, we can improve performance and focus on only two versions
(1.17 and soon 1.18).

See https://go.googlesource.com/proposal/+/refs/changes/78/248178/1/design/40724-register-calling.md
See #37

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -56,7 +56,6 @@ jobs:
         os: [ubuntu-20.04, macos-11, windows-2022]
         go-version:
         - "1.17" # == ${{ env.GO_VERSION }} because matrix cannot expand env variables
-        - "1.16" # temporarily support go 1.16 per #37
 
     steps:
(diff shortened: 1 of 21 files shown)


.github/workflows/commit.yaml
```diff
@@ -86,7 +85,6 @@ jobs:
       matrix:
         go-version:
         - "1.17"  # == ${{ env.GO_VERSION }} because matrix cannot expand env variables
-        - "1.16"  # temporarily support go 1.16 per #37
```
codefromthecrypt, line 89: We can add 1.18 once it is out of beta. Typically github adds it a week or so after GA. Then, you can look for it like so: https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#go

mathetake (approved):

PR #318, opened 2022-03-03
ci: add bench.check target to ensure JIT is the fastest 

mathetake:
This commit adds `bench.check` make target to ensure that JIT is the 
fastest engine among others.  There was a performance degration 
in #287 which has been resolved in #316, but only found manually.

The target sets a linker flag in vs/bench_test.go which results in running 
`TestFacIter_JIT_Fastest`  instead of `BenchmarkFacIter_Invoke`. 
The `bench` make target doesn't set the flag so 
`BenchmarkFacIter_Invoke` is run and produces rich benchmark output 
into stdout and doesn't run `TestFacIter_JIT_Fastest.

### DIFF
--- a/internal/wasm/jit/engine.go
+++ b/internal/wasm/jit/engine.go
@@ -432,8 +432,8 @@ func newEngine() *engine {
 
 // TODO: better make them configurable?
 const (
-	initialValueStackSize             = 1024
-	initialCallFrameStackSize         = 1024
+	initialValueStackSize             = 64
+	initialCallFrameStackSize         = 16
 	initialCompiledFunctionsSliceSize = 128
 )
 
--- a/internal/wasm/jit/engine.go
+++ b/internal/wasm/jit/engine.go
@@ -432,8 +432,8 @@ func newEngine() *engine {
 
 // TODO: better make them configurable?
 const (
-	initialValueStackSize             = 64
-	initialCallFrameStackSize         = 16
+	initialValueStackSize             = 1024
+	initialCallFrameStackSize         = 1024
 	initialCompiledFunctionsSliceSize = 128
 )
 
--- a/vs/bench_fac_iter_test.go
+++ b/vs/bench_fac_iter_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	_ "embed"
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/bytecodealliance/wasmtime-go"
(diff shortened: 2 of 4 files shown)


internal/wasm/jit/engine.go
```diff
@@ -432,8 +432,8 @@ func newEngine() *engine {
 
 // TODO: better make them configurable?
 const (
-	initialValueStackSize             = 64
-	initialCallFrameStackSize         = 16
+	initialValueStackSize             = 1024
+	initialCallFrameStackSize         = 1024
```
mathetake, line 436: revert after verifying this causes CI failure

mathetake:
OK cool this works

```
JIT   137134	      8126 ns/op
interpreter   523346	      2434 ns/op
wasmer-go   897148	      1441 ns/op
wasmtime-go   406878	      3125 ns/op
--- FAIL: TestFacIter_JIT_Fastest (5.13s)
    --- FAIL: TestFacIter_JIT_Fastest/interpreter (0.00s)
        assertion_compare.go:332: 
            	Error Trace:	bench_fac_iter_test.go:169
            	Error:      	"8125.853340528242" is not less than "2434.2552403190243"
            	Test:       	TestFacIter_JIT_Fastest/interpreter
            	Messages:   	[JIT engine must be faster than interpreter. Run BenchmarkFacIter_Invoke with ensureJITFastest=false instead to see the detailed result]
    --- FAIL: TestFacIter_JIT_Fastest/wasmer-go (0.00s)
        assertion_compare.go:332: 
            	Error Trace:	bench_fac_iter_test.go:169
            	Error:      	"8125.853340528242" is not less than "1440.9133754965735"
            	Test:       	TestFacIter_JIT_Fastest/wasmer-go
            	Messages:   	[JIT engine must be faster than wasmer-go. Run BenchmarkFacIter_Invoke with ensureJITFastest=false instead to see the detailed result]
    --- FAIL: TestFacIter_JIT_Fastest/wasmtime-go (0.00s)
        assertion_compare.go:332: 
            	Error Trace:	bench_fac_iter_test.go:169
            	Error:      	"8125.853340528242" is not less than "3124.905979187865"
            	Test:       	TestFacIter_JIT_Fastest/wasmtime-go
            	Messages:   	[JIT engine must be faster than wasmtime-go. Run BenchmarkFacIter_Invoke with ensureJITFastest=false instead to see the detailed result]
FAIL

```

vs/bench_fac_iter_test.go
```diff
@@ -18,6 +19,8 @@ import (
 	"github.com/tetratelabs/wazero/wasm"
 )
 
+var ensureJITFastest string = "false"
```
codefromthecrypt, line 22:
```suggestion
// ensureJITFastest is overridable via ldflags. Ex.
//	-ldflags '-X github.com/tetratelabs/wazero/vs.ensureJITFastest=true'
var ensureJITFastest string = "false"
```

codefromthecrypt (approved):
Thanks for adding insurance!

PR #327, opened 2022-03-04
Reduces exports and prepares ReleaseModuleInstance for testing

codefromthecrypt:
This removes more exports and adds a toehold test for
`ReleaseModuleInstance` so that it can later be completed for both
wasm and host modules.

### DIFF
--- a/internal/wasm/host.go
+++ b/internal/wasm/host.go
@@ -91,23 +91,24 @@ func (f *exportedFunction) Call(ctx context.Context, params ...uint64) ([]uint64
 }
 
 // NewHostModule is defined internally for use in WASI tests and to keep the code size in the root directory small.
-func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interface{}) (publicwasm.HostModule, error) {
+func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interface{}) (*HostModule, error) {
 	if err := s.requireModuleUnused(moduleName); err != nil {
 		return nil, err
 	}
 
 	exportCount := len(nameToGoFunc)
 
-	ret := &HostModule{NameToFunctionInstance: make(map[string]*FunctionInstance, exportCount)}
+	ret := &HostModule{name: moduleName, NameToFunctionInstance: make(map[string]*FunctionInstance, exportCount)}
 	hostModule := &ModuleInstance{
-		Name: moduleName, Exports: make(map[string]*ExportInstance, exportCount),
+		Name:       moduleName,
+		Exports:    make(map[string]*ExportInstance, exportCount),
 		hostModule: ret,
 	}
 	s.moduleInstances[moduleName] = hostModule
 	for name, goFunc := range nameToGoFunc {
 		if hf, err := NewGoFunc(name, goFunc); err != nil {
 			return nil, err
-		} else if function, err := s.AddHostFunction(hostModule, hf); err != nil {
+		} else if function, err := s.addHostFunction(hostModule, hf); err != nil {
 			return nil, err
 		} else {
 			ret.NameToFunctionInstance[name] = function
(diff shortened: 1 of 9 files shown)


internal/wasm/host.go
```diff
@@ -120,7 +121,7 @@ func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interfa
 // exists for this module and name it is ignored rather than overwritten.
 //
 // Note: The wasm.Memory of the fn will be from the importing module.
-func (s *Store) AddHostFunction(m *ModuleInstance, hf *GoFunc) (*FunctionInstance, error) {
+func (s *Store) addHostFunction(m *ModuleInstance, hf *GoFunc) (*FunctionInstance, error) {
```
codefromthecrypt, line 124: this is not used anywhere except AddHostModule, so there's no chance of export conflicts (the keys are a map)

internal/wasm/store.go
```diff
@@ -455,7 +456,7 @@ func (s *Store) Instantiate(module *Module, name string) (*PublicModule, error)
 	instance.applyElements(module.ElementSection)
 	instance.applyData(module.DataSection)
 
-	// Persist the instances other than functions (which we already persisted before compilation).
+	// Persist the instances other tha functions (which we already persisted before compilation).
```
codefromthecrypt, line 459:
```suggestion
	// Persist the instances other than functions (which we already persisted before compilation).
```

internal/wasm/store.go
```diff
@@ -477,30 +478,44 @@ func (s *Store) Instantiate(module *Module, name string) (*PublicModule, error)
 			return nil, fmt.Errorf("module[%s] start function failed: %w", name, err)
 		}
 	}
-	return &PublicModule{s, instance}, nil
+	return &PublicModule{s: s, instance: instance}, nil
 }
 
-func (s *Store) ReleaseModuleInstance(instance *ModuleInstance) error {
+// ReleaseModuleInstance deallocates resources if a module with the given name exists.
+func (s *Store) ReleaseModuleInstance(moduleName string) error {
```
codefromthecrypt, line 485: using a string name here is fine because module name uniqueness is guaranteed regardless of host or wasm module. later, this will allow us to call easier from an outside package (perhaps needing to export the name field in PublicXX)

internal/wasm/store_test.go
```diff
@@ -86,48 +85,84 @@ func TestModuleInstance_Memory(t *testing.T) {
 	}
 }
 
-func TestStore_AddHostFunction(t *testing.T) {
+func TestStore_NewHostModule(t *testing.T) {
```
mathetake, line 88: should we move this to host_test.go?

internal/wasm/store_test.go
```diff
@@ -86,48 +85,84 @@ func TestModuleInstance_Memory(t *testing.T) {
...
+
+	// Ensure paths that can create the host module can see the name.
+	m, err := s.Instantiate(&Module{}, "module")
+	require.NoError(t, err)
+	require.Equal(t, "Module[module]", m.String())
+	require.Equal(t, "Module[module]", s.Module(m.instance.Name).String())
 }
 
-func TestStore_ExportImportedHostFunction(t *testing.T) {
+func TestHostModule_String(t *testing.T) {
```
mathetake, line 124: same goes for this

internal/wasm/store_test.go
codefromthecrypt: hmm but these are code implemented in store

internal/wasm/store_test.go
codefromthecrypt: same here, this is making sure store which is instantiating it, is setting the name field..

internal/wasm/store_test.go
codefromthecrypt: I am wrong this code was moved (probably by me) to host.go I'll fix this!

internal/wasm/store_test.go
codefromthecrypt: same .. I forgot even though these are functions with store pointers that do the stuff, it is in host.go

mathetake (approved):

PR #329, opened 2022-03-04
Simplify features const declarations

mathetake:
iota abbreviation works well with bit shifts as well


codefromthecrypt (approved):
oops :D

Issue #328, opened 2022-03-04
Flake on jit_test.go: TestAmd64Compiler_compileBrTable


codefromthecrypt:
at the time, the code that failed was this line, which was nil.

```
	jitcall(
		uintptr(unsafe.Pointer(&code[0])),
		uintptr(unsafe.Pointer(j.vm)),
	)
```

It should not be nil due to jit_amd64.s
```
// jitcall(codeSegment, callEngine)
TEXT ·jitcall(SB),NOSPLIT|NOFRAME,$0-16
--snip--
```

```
--- FAIL: TestAmd64Compiler_compileBrTable (0.39s)
[18](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:18)
    --- FAIL: TestAmd64Compiler_compileBrTable/index_register=2074,tmpRegister=2064 (0.02s)
[19](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:19)
        --- FAIL: TestAmd64Compiler_compileBrTable/index_register=[20](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:20)74,tmpRegister=2064/select_last_with_five_targets (0.01s)
20
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[21](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:21)
	panic: runtime error: invalid memory address or nil pointer dereference
[22](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:22)
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x24ae066]
[23](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:23)

[24](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:24)
goroutine 862 [running]:
[25](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:25)
testing.tRunner.func1.2({0x12ae360, 0x1513370})
[26](https://github.com/tetratelabs/wazero/runs/5415751283?check_suite_focus=true#step:5:26)
	/Users/runner/hostedtoolcache/go/1.17.7/x64/src/testing/testing.go:1209 +0x24e
[...]

PR #330, opened 2022-03-04
Adds test case where use up all unreserved registers.

mathetake:


mathetake:
intends to resolve #328, but in anyway this helps us to find bugs like #311

tests/engine/testdata/hugestack.wat
```diff
@@ -0,0 +1,24 @@
+(module
```
mathetake, line 1: I think we need formatter for wat :D

mathetake:
so found that no need to save callee saved registers in arm64 ABI, but we are on Go's ABI so it's only necessary not to use X18 and X28

mathetake:
for arm64

tests/engine/adhoc_test.go
```diff
@@ -67,28 +72,50 @@ func runTests(t *testing.T, newRuntimeConfig func() *wazero.RuntimeConfig) {
 	})
 }
 
-func testFibonacci(t *testing.T, newRuntimeConfig func() *wazero.RuntimeConfig) {
+func testHugeStack(t *testing.T, newRuntimeConfig func() *wazero.RuntimeConfig) {
 	// We execute 1000 times in order to ensure the JIT engine is stable under high concurrency
```
mathetake, line 76: I think in another PR I will separate out the counccurent testing in conccurent_test.go

codefromthecrypt (approved):
good find!

PR #332, opened 2022-03-04
uses less code testing by not repeating the read-side in error tests

codefromthecrypt:


internal/wasi/wasi_test.go
```diff
@@ -1147,11 +1147,19 @@ func TestSnapshotPreview1_FdWrite_Errors(t *testing.T) {
 		}
 	})
 
+	// Setup valid test memory
+	iovs, iovsCount := uint64(0), uint64(1)
```
codefromthecrypt, line 1151: so basically here, we aren't truncating manually anymore, rather using the same memory for the "read side" of a 2 step operation like fdwrite. Ack it is easy to goof the offsets as the errno is less specific than normal go code where you'd use an `error` which would have some description of what context failed (ex reading offset vs len)

codefromthecrypt:
@nullpo-head feel free to merge or obviate this as you see fit in another PR!

nullpo-head (approved):
thanks!

PR #333, opened 2022-03-04
untangles host module instantiation

codefromthecrypt:
This unravels host module instantiation a little. Particularly it is more obvious that some preparation can be done prior to mutating the store. There may be future possibilities from here.

### DIFF
--- a/internal/wasm/host.go
+++ b/internal/wasm/host.go
@@ -129,7 +129,7 @@ func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interfa
 	return ret, nil
 }
 
-func (s *Store) compileFunction(f *FunctionInstance, hf *GoFunc) (err error) {
+func (s *Store) compileHostFunction(f *FunctionInstance, hf *GoFunc) (err error) {
 	f.FunctionType, err = s.getTypeInstance(hf.functionType)
 	if err != nil {
 		return err
--- a/internal/wasm/host.go
+++ b/internal/wasm/host.go
@@ -97,57 +97,53 @@ func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interfa
 	}
 
 	exportCount := len(nameToGoFunc)
-
 	ret := &HostModule{name: moduleName, NameToFunctionInstance: make(map[string]*FunctionInstance, exportCount)}
 	hostModule := &ModuleInstance{
 		Name:       moduleName,
 		Exports:    make(map[string]*ExportInstance, exportCount),
 		hostModule: ret,
 	}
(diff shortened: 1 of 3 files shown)


internal/wasm/host.go
```diff
@@ -97,57 +97,53 @@ func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interfa
...
 	return ret, nil
 }
 
-// AddHostFunction exports a function so that it can be imported under the given module and name. If a function already
-// exists for this module and name it is ignored rather than overwritten.
-//
-// Note: The wasm.Memory of the fn will be from the importing module.
-func (s *Store) addHostFunction(m *ModuleInstance, hf *GoFunc) (*FunctionInstance, error) {
-	typeInstance, err := s.getTypeInstance(hf.functionType)
+func (s *Store) compileFunction(f *FunctionInstance, hf *GoFunc) (err error) {
```
mathetake, line 132: I would prefer adding "host" element in the name as this is only for host
```suggestion
func (s *Store) compileHostFunction(f *FunctionInstance, hf *GoFunc) (err error) {
```

mathetake (approved):

internal/wasm/host.go
```diff
@@ -97,57 +97,53 @@ func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interfa
...
+			Name:           fmt.Sprintf("%s.%s", hostModule.Name, hf.wasmFunctionName),
+			HostFunction:   hf.goFunc,
+			FunctionKind:   hf.functionKind,
+			ModuleInstance: hostModule,
+		}
+		hostModule.Exports[hf.wasmFunctionName] = &ExportInstance{Type: ExternTypeFunc, Function: f}
+		hostModule.Functions = append(hostModule.Functions, f)
+		ret.NameToFunctionInstance[name] = f
+
+		if err = s.compileFunction(f, hf); err != nil {
```
codefromthecrypt, line 123: what I noticed after this is the only things that need the store (besides the initial check for "all clear" on the module name) is this which compiles the function and at the bottom, adding it to the module.

I'm thinking later we can have a similar relationship for host modules as wasm defined ones in terms of preparation errors vs instantiation ones.

Ex. Wasm-defined modules can fail early on decode for any reason (including compilation if we push it up that early)

There seems to be an option for failing early on host modules, for example if the signatures are invalid. Some similar "compile" prep may be useful in the future vs only being able to tell if something will fail when you use it.

codefromthecrypt:
merging as the build fail is unrelated and hopefully fixed in the next PR. Thanks for the look!

PR #335, opened 2022-03-05
Flattens FunctionInstance FunctionType field

codefromthecrypt:
This flattens `FunctionInstance.FunctionType` into `Type` and `TypeID`
fields, where the former is known prior to instantiation. This helps
pave a way for integration between Wasm declared and host defined
modules.

This also clarifies that `FunctionInstance.String` was used as a lookup
key, by renaming and caching its impl. While at it, I renamed "null" to
"v" in its output as I had been using v for void noticing others were
doing that also. Moreover, null is easy to misunderstand as a bug.

### DIFF
--- a/internal/wasm/host.go
+++ b/internal/wasm/host.go
@@ -105,22 +105,23 @@ func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interfa
 	}
 
 	for name, goFunc := range nameToGoFunc {
-		hf, err := NewGoFunc(name, goFunc)
+		hf, err := newGoFunc(goFunc)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("func[%s] %w", name, err)
 		}
 
 		f := &FunctionInstance{
-			Name:           fmt.Sprintf("%s.%s", hostModule.Name, hf.wasmFunctionName),
-			HostFunction:   hf.goFunc,
-			FunctionKind:   hf.functionKind,
-			ModuleInstance: hostModule,
+			Module: hostModule,
+			Name:   fmt.Sprintf("%s.%s", hostModule.Name, name),
+			Kind:   hf.functionKind,
+			Type:   hf.functionType,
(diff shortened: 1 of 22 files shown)


internal/wasm/host.go
```diff
@@ -111,16 +111,17 @@ func (s *Store) NewHostModule(moduleName string, nameToGoFunc map[string]interfa
...
 		f := &FunctionInstance{
-			Name:           fmt.Sprintf("%s.%s", hostModule.Name, hf.wasmFunctionName),
-			HostFunction:   hf.goFunc,
-			FunctionKind:   hf.functionKind,
-			ModuleInstance: hostModule,
+			Module: hostModule,
+			Name:   fmt.Sprintf("%s.%s", hostModule.Name, hf.name),
+			Kind:   hf.functionKind,
+			Type:   hf.functionType,
+			GoFunc: hf.goFunc,
```
codefromthecrypt, line 118: renamed this to GoFunc to match the kind name. ex `FunctionKindGoContext`

internal/wasm/interpreter/interpreter_test.go
```diff
@@ -89,15 +89,13 @@ func TestEngine_Call_HostFn(t *testing.T) {
...
-				Params:  []wasm.ValueType{wasm.ValueTypeI64},
-				Results: []wasm.ValueType{wasm.ValueTypeI64},
-			},
+		GoFunc: &hostFn,
+		Kind:   wasm.FunctionKindGoModuleContext,
+		Type: &wasm.FunctionType{
+			Params:  []wasm.ValueType{wasm.ValueTypeI64},
+			Results: []wasm.ValueType{wasm.ValueTypeI64},
 		},
-		ModuleInstance: module,
```
codefromthecrypt, line 100: the main reason we use the type name for the field name is when there's a method of the same name (so can't), or the name is the same as a reserved word.

I reverted the longer names in Function to improve readability. There are other places we can shorted also.

mathetake (approved):

PR #339, opened 2022-03-07
Opt-in support for sign-extend

mathetake:
Allows users to do the following to enable sign-extension-ops

```
r := wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfig().WithFeatureSignExtensionOps(true))
```

Resolves #66

PR #340, opened 2022-03-07
Polishes in preparation of first post 1.0 feature

codefromthecrypt:
will rebase #339 after this


mathetake (approved):

codefromthecrypt:
had too much in here, should be good now.

PR #342, opened 2022-03-07
Enable concurrent instantiation, and release

mathetake:
This makes sure that concurrent use of Instantiate and
ReleaseModuleInstance is goroutine-safe. 

Note that it is still not safe to expose Release in the public API
as we haven't taken into account the outstanding calls. That
would be addressed in the follow-up commit.

part of #293.

### DIFF
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"sync"
 
 	"github.com/tetratelabs/wazero/internal/ieee754"
 	"github.com/tetratelabs/wazero/internal/leb128"
@@ -89,6 +90,8 @@ type (
 		// in this store.
 		// The slice index is to be interpreted as tableaddr (https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-tableaddr).
 		tables []*TableInstance
+
+		mux sync.RWMutex
 	}
 
 	// ModuleInstance represents instantiated wasm module.
@@ -114,7 +117,8 @@ type (
 		hostModule *HostModule
 
 		// TODO: https://github.com/tetratelabs/wazero/issues/293
-		refCount      int
+		mux           sync.Mutex
+		importedCount int
 		moduleImports map[*ModuleInstance]struct{}
 	}
 
@@ -400,6 +404,8 @@ func NewStore(ctx context.Context, engine Engine, enabledFeatures Features) *Sto
 
 // checkFunctionIndexOverflow checks if there would be too many function instances in a store.
 func (s *Store) checkFunctionIndexOverflow(newInstanceNum int) error {
+	s.mux.RLock()
+	defer s.mux.RUnlock()
 	if len(s.functions) > int(s.maximumFunctionIndex)-newInstanceNum {
 		return fmt.Errorf("too many functions in a store")
 	}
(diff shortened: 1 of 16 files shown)


mathetake:
ok cool no race detected 

```
$ go test -race -run ^TestStore_concurrent$ github.com/tetratelabs/wazero/internal/wasm -count=100 
ok      github.com/tetratelabs/wazero/internal/wasm     19.425s
```

internal/wasm/store.go
```diff
@@ -484,28 +505,29 @@ func (s *Store) Instantiate(module *Module, name string) (*PublicModule, error)
...
+	instance.mux.Lock()
+	defer instance.mux.Unlock()
+
+	if instance.importedCount > 0 {
 		// This case other modules are importing this module instance and still alive.
-		return fmt.Errorf("%d modules import this and need to be closed first", instance.refCount)
+		return fmt.Errorf("%d modules import this and need to be closed first", instance.importedCount)
 	}
 
 	// TODO: check outstanding calls and wait until they exit.
```
mathetake, line 524: this will be resolved in the follow-up and then we could expose Close (or whatever we call) in the public API

codefromthecrypt:
I'm going to add some commits instead of making review comments (or at least to reduce them)

internal/wasm/store.go
```diff
@@ -101,21 +102,27 @@ type (
...
 		// hostModule holds HostModule if this is a "host module" which is created in store.NewHostModule.
 		hostModule *HostModule
 
-		// TODO: https://github.com/tetratelabs/wazero/issues/293
-		refCount      int
-		moduleImports map[*ModuleInstance]struct{}
+		// mux is used to guard the fields from concurrent access.
+		mux sync.Mutex
+
+		// dependentCount is the current number of modules which import this module. On Store.ReleaseModule, this number
```
codefromthecrypt, line 119: notably renamed these two fields

internal/wasm/store.go
```diff
@@ -482,56 +501,69 @@ func (s *Store) Instantiate(module *Module, name string) (*PublicModule, error)
...
+	s.releaseGlobal(m.Globals...)
 
-	// Explicitly assign nil so that we ensure this moduleInstance no longer holds reference to instances.
-	instance.Exports = nil
-	instance.Globals = nil
-	instance.Functions = nil
-	instance.TableInstance = nil
-	instance.MemoryInstance = nil
-	instance.Types = nil
+	// Explicitly assign nil so that we ensure this module no longer holds reference to instances.
```
codefromthecrypt, line 539: mention why it is important especially as we clear the reference below.. is it because we want things that have this reference to fail (panic)?

codefromthecrypt (approved):
one main question on the impl, otherwise works for me!


PS I think in the future I'd also like to understand why the store can't partition data by module as it seems the locks would be less blinding.  I guess there are store-wide type indexes that must be coherent across modules? If this is already explained in RATIONALE never mind, but one day if it isn't let's add something on the data partitioning strategy as it impacts locking.

mathetake:
> why the store can't partition data by module as it seems the locks would be less blinding.

Theoretically can be partitioned, but a live instance must be assigned a unique addr (funcaddr, memaddr, etc in the spec https://www.w3.org/TR/wasm-core-1/#addresses%E2%91%A0), and notably the function calls were made via funcaddr (which engine maps to the a *compiledFunction).  In the spec rather aggressively, a module instance only holds addresses (https://www.w3.org/TR/wasm-core-1/#module-instances%E2%91%A0).

But yeah I believe there's some way to work around this (especially once we are able to do precompilation namely after we could decouple compiledFunction from runtime representations).

internal/wasm/store.go
mathetake: deleted as this is unnecessary

mathetake:
Raise a follow-up for the ps comment! https://github.com/tetratelabs/wazero/pull/343

codefromthecrypt:
thanks for the work on this, getting close to safe Close!

PR #343, opened 2022-03-07
Decouples global, table and memory instance from store.

mathetake:
This decouples global, table and memory instances from
store, and now only ModuleInstances are holding references
to these types.

This is a follow-up on [the comment by @codefromthecrypt ](https://github.com/tetratelabs/wazero/pull/342#pullrequestreview-901580527) in #342


codefromthecrypt (approved):
love it

PR #345, opened 2022-03-08
Introduce ModuleEngine interface

mathetake:
This commit adds ModuleEngine interface which is used to 
make calls and created per-module instance. Notably, this 
enables us to remove the necessity for store to holds 
FunctionInstances and FunctionIndex. 

This is a follow-up from #342


mathetake:
ok the left is call_indirect ..

mathetake:
ok the left is linking.wast and imports.wast

mathetake:
ok started passing all tests on amd64 and riscv.. the left is to make changes to amd64 jits

mathetake:
cool passed

mathetake:
🤞

mathetake:
hmm..

mathetake:
windows might become flaky with this but in any case I will take care of the subsequent prs and in the worse case I will add t.Skip in ad-hoc for windows and raise issue

mathetake:
this PR itself is too big to debug the potential flake which might have existed in  the first place

codefromthecrypt:
I'll rebase and add any minor commits merging unless they become not minor :D

internal/wasm/store.go
```diff
@@ -49,27 +48,14 @@ type (
 
 		// maximumFunctionIndex represents the limit on the number of function addresses (= function instances) in a store.
 		// Note: this is fixed to 2^27 but have this a field for testability.
-		maximumFunctionIndex FunctionIndex
+		//
+		// TODO: Enforce this at module-level.
+		// maximumFunctionIndex FunctionIndex
```
mathetake, line 53:
```suggestion
```

internal/wasm/store.go
```diff
@@ -269,9 +230,12 @@ func (m *ModuleInstance) buildExports(exports map[string]*Export) {
 			ei = &ExportInstance{Type: exp.Type, Function: m.Functions[index]}
 			// The module instance of the host function is a fake that only includes the function and its types.
 			// We need to assign the ModuleInstance when re-exporting so that any memory defined in the target is
-			// available to the wasm.Module Memory.
+			// available to the wasm.ModuleContext Memory.
+			//
+			// TODO: shouldn't this be race if multiple modules re-exporting? maybe we should limit this to only tests.
```
mathetake, line 235: @codefromthecrypt maybe this is of your interests

codefromthecrypt: yeah though I need to think about this one more clearly tomorrow :D It looks bad though. I'll change the TODO to TODO(adrian):

internal/wasm/interpreter/interpreter.go
```diff
@@ -20,44 +21,57 @@ var callStackCeiling = buildoptions.CallStackCeiling
...
 	return &engine{
-		compiledFunctions: make([]*compiledFunction, initialCompiledFunctionsSliceSize),
+		compiledFunctions: make(map[*wasm.FunctionInstance]*compiledFunction),
 	}
 }
 
-// callEngine holds context per engine.Call, and shared across all the
-// function calls originating from the same engine.Call execution.
+func (e *engine) deleteCompiledFunction(f *wasm.FunctionInstance) {
+	e.mux.RLock()
```
codefromthecrypt, line 35: switching this to Lock as it is not read-only

codefromthecrypt (approved):
I like how this is going! Works for me when you are ready

internal/wasm/interpreter/interpreter.go
```diff
@@ -448,15 +464,26 @@ func (e *engine) lowerIROps(f *wasm.FunctionInstance,
 	return ret, nil
 }
 
-// Call implements an interpreted wasm.Engine.
-func (e *engine) Call(ctx *wasm.ModuleContext, f *wasm.FunctionInstance, params ...uint64) (results []uint64, err error) {
+// FunctionAddress implements internalwasm.ModuleEngine FunctionAddress.
+func (me *moduleEngine) FunctionAddress(index wasm.Index) uintptr {
+	return uintptr(unsafe.Pointer(me.compiledFunctions[index]))
```
nullpo-head, line 469: Does it work well after GC moved the address of compiledFunctions?

internal/wasm/interpreter/interpreter.go
```diff
@@ -619,20 +641,18 @@ func (ce *callEngine) callNativeFunc(ctx *wasm.ModuleContext, f *compiledFunctio
 				if offset >= uint64(len(table.Table)) {
 					panic(wasm.ErrRuntimeInvalidTableAccess)
 				}
-				tableElement := table.Table[offset]
-				// Type check.
-				if uint64(tableElement.FunctionTypeID) != op.us[1] {
-					if tableElement.FunctionTypeID == wasm.UninitializedTableElementTypeID {
-						panic(wasm.ErrRuntimeInvalidTableAccess)
-					}
+				targetCompiledFunction := (*compiledFunction)(unsafe.Pointer(table.Table[offset]))
```
nullpo-head, line 644: Does it work well after GC moved the address of compiledFunctions?

mathetake: go's GC doesn't do compaction or relocation, so as long as the *compilerFunction is alive, this is safe

nullpo-head: oh, I see. is that behavior just true for the current implementation, or guaranteed by spec?  But anyway no problem if you're sure. I asked just because I see the warning for this line on VSCode and found this document that warns it https://pkg.go.dev/unsafe#:~:text=A%20uintptr%20is%20an,object%20from%20being%20reclaimed.

mathetake:
haha they are still flaky clearly not ready 😄

mathetake:
`TestAmd64Compiler_compileCall` is the culprit for sure.. let me kill it tomorrow

internal/wasm/interpreter/interpreter.go
mathetake: GC's behavior is not something written in language's spec, but you can see https://github.com/golang/go/blob/master/src/runtime/mgc.go#L9

mathetake: but any way on second though, I think I will use `interface` instead of uintptr for safety in the subsequent PR

PR #347, opened 2022-03-08
Removes host-specific types

codefromthecrypt:
This converges host-defined modules with Wasm defined modules by
introducing a custom section for host-defined functions. The net result
are far less types and consistent initialization.

* HostModule is removed for Module
* HostFunction is removed for Function
* ModuleContext is removed for Module

Note: One impact of this is that the low-level API no longer accepts a
go context (context.Context), rather a `wasm.Module` which the function
is called in context of. This meant exposing `wasm.Module.WithContext`
to override the default.

Signed-off-by: Adrian Cole <adrian@tetrate.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ func main() {
 	module, _ := wazero.NewRuntime().NewModuleFromSource(source)
 
 	// Discover 7! is 5040
-	fmt.Println(module.Function("fac").Call(context.Background(), 7))
+	fmt.Println(module.ExportedFunction("fac").Call(nil, 7))
 }
 ```
 
--- a/internal/wasm/binary/types.go
+++ b/internal/wasm/binary/types.go
@@ -76,7 +76,7 @@ var encodedOneResult = map[wasm.ValueType][]byte{
 
 // encodeFunctionType returns the internalwasm.FunctionType encoded in WebAssembly 1.0 (20191205) Binary Format.
 //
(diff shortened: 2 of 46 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ func main() {
 	module, _ := wazero.NewRuntime().NewModuleFromSource(source)
 
 	// Discover 7! is 5040
-	fmt.Println(module.Function("fac").Call(context.Background(), 7))
+	fmt.Println(module.ExportedFunction("fac").Call(nil, 7))
```
codefromthecrypt, line 24: Note: this prefix is to retain access to `Memory()` without necessarily exporting it. Later changes may decide to do similar with functions and globals, but that would be too much here.

internal/wasm/binary/types.go
```diff
@@ -76,7 +76,7 @@ var encodedOneResult = map[wasm.ValueType][]byte{
 
 // encodeFunctionType returns the internalwasm.FunctionType encoded in WebAssembly 1.0 (20191205) Binary Format.
 //
-// Note: Function types are encoded by the byte 0x60 followed by the respective vectors of parameter and result types.
+// Note: ExportedFunction types are encoded by the byte 0x60 followed by the respective vectors of parameter and result types.
```
codefromthecrypt, line 79:
```suggestion
// Note: Function types are encoded by the byte 0x60 followed by the respective vectors of parameter and result types.
```

mathetake (approved):

PR #348, opened 2022-03-08
WASI: Implement fd_seek

nullpo-head:
This PR implements `fd_seek` API.

### DIFF
--- a/internal/wasi/fs.go
+++ b/internal/wasi/fs.go
@@ -77,15 +77,16 @@ func (m *MemFS) OpenWASI(dirFlags uint32, path string, oFlags uint32, fsRights, 
 	return ret, nil
 }
 
+// memFile implements wasi.File
 type memFile struct {
 	buf    []byte
 	offset int64
 	flush  func(bts []byte)
 }
 
-// Read implements wasi.File.Read
+// Read implements io.Reader
 func (f *memFile) Read(p []byte) (int, error) {
-	// Read must return 0, io.EOF when it reaches the end-of-file
+	// In memFile, the end of the buffer is the end of the file.
 	if f.offset == int64(len(f.buf)) {
 		return 0, io.EOF
 	}
@@ -94,7 +95,7 @@ func (f *memFile) Read(p []byte) (int, error) {
 	return nread, nil
 }
 
-// Write implements wasi.File.Write
+// Write implements io.Writer
 func (f *memFile) Write(p []byte) (int, error) {
 	nwritten := copy(f.buf[f.offset:], p)
 	f.buf = append(f.buf, p[nwritten:]...)
@@ -102,7 +103,7 @@ func (f *memFile) Write(p []byte) (int, error) {
 	return len(p), nil
 }
 
-// Seek implements wasi.File.Seek
+// Seek implements io.Seeker
 func (f *memFile) Seek(offset int64, whence int) (int64, error) {
 	switch whence {
 	case io.SeekStart:
@@ -117,7 +118,7 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
 	return f.offset, nil
 }
 
-// Close implements wasi.File.Close
+// Close implements io.Closer
 func (f *memFile) Close() error {
 	if f.flush != nil {
 		f.flush(f.buf)
--- a/internal/wasi/fs.go
+++ b/internal/wasi/fs.go
@@ -1,7 +1,8 @@
 package internalwasi
 
 import (
-	"bytes"
+	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"runtime"
@@ -65,11 +66,11 @@ func (m *MemFS) OpenWASI(dirFlags uint32, path string, oFlags uint32, fsRights, 
 		}
 	}
 
-	ret := &memFile{buf: bytes.NewBuffer(buf)}
+	ret := &memFile{buf: buf, offset: 0}
 
 	if fsRights&wasi.R_FD_WRITE != 0 {
 		ret.flush = func(bts []byte) {
-			m.Files[path] = bts
+			m.Files[path] = append([]byte{}, bts...)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


internal/wasi/wasi.go
```diff
@@ -1218,7 +1244,25 @@ func (a *wasiAPI) FdRenumber(ctx wasm.Module, fd, to uint32) wasi.Errno {
 
 // FdSeek implements SnapshotPreview1.FdSeek
 func (a *wasiAPI) FdSeek(ctx wasm.Module, fd uint32, offset uint64, whence uint32, resultNewoffset uint32) wasi.Errno {
-	return wasi.ErrnoNosys // stubbed for GrainLang per #271
+	f, ok := a.opened[fd]
+	if !ok || f.file == nil {
+		return wasi.ErrnoBadf
+	}
+
+	// Note: Golang's whence and WASI's whence are compatible.
```
mathetake, line 1252: what is "Golang's whence" here?

nullpo-head: I've clarified the comment so that it's more understandable that "Golang's whence" meant io package's Seek* constants.

wasi/wasi.go
```diff
@@ -28,9 +28,12 @@ const (
 	R_FD_WRITE
 )
 
+// File is an interface for an opened file similar to fs.File, but it supports addtional operations such as Write.
```
codefromthecrypt, line 31: Change this entirely as the it is a side-note that this happens to match another thing

```
File combines file I/O interfaces supported by ModuleSnapshotPreview1
```

os.File isn't important and we shouldn't mention it. The more important type is fs.File (because we want to end up with an impl of `fs.FS`)

IMHO, we shouldn't mention fs.File until we finish the things it supports. Then we embed the fs.File interface along with anything extra we support.

Ex. you'll notice we don't have stat here, yet.
```
// A File provides access to a single file.
// The File interface is the minimum implementation required of the file.
// Directory files should also implement ReadDirFile.
// A file may implement io.ReaderAt or io.Seeker as optimizations.
type File interface {
	Stat() (FileInfo, error)
	Read([]byte) (int, error)
	Close() error
}
```

internal/wasi/fs.go
```diff
@@ -65,33 +66,61 @@ func (m *MemFS) OpenWASI(dirFlags uint32, path string, oFlags uint32, fsRights,
 		}
 	}
 
-	ret := &memFile{buf: bytes.NewBuffer(buf)}
+	ret := &memFile{buf: buf}
 
 	if fsRights&wasi.R_FD_WRITE != 0 {
 		ret.flush = func(bts []byte) {
-			m.Files[path] = bts
+			m.Files[path] = append([]byte{}, bts...)
```
mathetake, line 73: pr comment on what's this change about?

codefromthecrypt (commented):
thanks for giving this a start! This needs tests or all three seek whences, and the only good time to do this is while it is already in the brain (now).

Most importantly, we need to ensure we know what `offset` is, and I suspect it is a signed number even if encoded as a uint64. Tests on SeekEnd should flush this out. If it is signed, the docs need to change on `offset` to say it can be negative.

codefromthecrypt:
PS hopefully the other interfaces needed for fs.FS will complete soon, then you can use go's `fstest.TestFS` to test it in addition to unit tests meanwhile.

However, new functions until that's the case should have unit tests. You'll notice a lot of similarity between go's MapFS and our memFile, so looking at code and tests around that may help (ack it mainly punts to `fstest.TestFS`)

internal/wasi/fs.go
nullpo-head: thanks, never mind it's just a garbage and I've deleted it. thanks!

internal/wasi/fs_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+	"testing/iotest"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestMemFile_Read_Seek tests the behavior of Read and Seek by iotest.TestReader.
+// See iotest.TestReader
+func TestMemFile_Read_Seek(t *testing.T) {
```
nullpo-head, line 12: let me know if you know of the better naming

internal/wasi/wasi.go
```diff
@@ -741,7 +741,33 @@ type SnapshotPreview1 interface {
...
+	//
+	//                             uint64le
+	//                    +-----------------------+
+	//                    |                       |
+	//          []byte{?, 4, 0, 0, 0, 0, 0, 0, 0, ? }
+	//  resultNewoffset --^
+	//
+	// Note: ImportFdSeek shows this signature in the WebAssembly 1.0 (20191205) Text Format.
+	// Note: This is similar to `lseek` in POSIX.
+	// See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#fd_seek
```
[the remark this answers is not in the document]
nullpo-head: @codefromthecrypt , I've realized that I cannot find `FdSeekName` symbol in any packages. I thought you were mentioning about `io.Seek*` in general, but they don't prefixes of `Fd` either. Could you tell me what you pointed at by `FdSeekName`, please?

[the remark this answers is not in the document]
codefromthecrypt: sorry I meant FunctionFdSeek

internal/wasi/fs_test.go
codefromthecrypt: fine with me!

internal/wasi/wasi.go
```diff
@@ -741,7 +741,37 @@ type SnapshotPreview1 interface {
 	// FdRenumber is the WASI function named FunctionFdRenumber
 	FdRenumber(ctx wasm.Module, fd, to uint32) wasi.Errno
 
-	// FdSeek is the WASI function named FunctionFdSeek
```
codefromthecrypt, line 744: I'll go back later and re-add the names to descriptions as I know they are missing mostly. no action needed here

codefromthecrypt (approved):
looking nice!🪞 💇

PR #352, opened 2022-03-09
Adds global binary encoder and fixes scope of mutable-global flag

codefromthecrypt:
Before, we denied mutable globals when mutable-global=false, when really
that flag is about import/export. This fixes the flag so that it
disables externalizing a mutable global. Non-exported mutable globals
are common as implenented by compilers such as Emscripten.

See https://github.com/emscripten-core/emscripten/blob/main/system%2Flib%2Fcompiler-rt%2Fstack_ops.S

### DIFF
--- a/internal/wasm/module_test.go
+++ b/internal/wasm/module_test.go
@@ -293,7 +293,7 @@ func TestModule_Validate_Errors(t *testing.T) {
 		expectedErr string
 	}{
 		{
-			name: "StartSection points to an invalid func",
+			name: "StartSection points to an infunc",
 			input: &Module{
 				TypeSection:     nil,
 				FunctionSection: []uint32{0},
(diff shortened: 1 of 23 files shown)


internal/wasm/module_test.go
```diff
@@ -293,7 +293,7 @@ func TestModule_Validate_Errors(t *testing.T) {
 		expectedErr string
 	}{
 		{
-			name: "StartSection points to an invalid func",
+			name: "StartSection points to an infunc",
```
nullpo-head, line 296: possibly due to some mechanical replacement
```suggestion
			name: "StartSection points to an invalid func",
```

nullpo-head (approved):

codefromthecrypt:
thanks for the look folks

PR #353, opened 2022-03-10
Use interface as table element to pass tests with `-race`

mathetake:
This is a followup on the comment https://github.com/tetratelabs/wazero/pull/345#discussion_r822587159
Notably, using `uintptr` and casting to objects is considered as invalid in Go
as `fatal error: checkptr: pointer arithmetic result points to invalid allocation`
is raised when running with `-race`

### DIFF
--- a/internal/wasm/engine.go
+++ b/internal/wasm/engine.go
@@ -9,11 +9,9 @@ type Engine interface {
 
 // ModuleEngine implements function calls for a given module.
 type ModuleEngine interface {
-	// FunctionAddress returns the absolute address of the compiled function for index.
-	// The returned address is stored and used as a TableInstance.Table element.
-	//
-	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-funcaddr
-	FunctionAddress(index Index) uintptr
+	// FunctionPointer returns the pointer the compiled function for index as interface{}.
+	// The returned interface is stored and used as a TableInstance.Table element.
+	FunctionPointer(index Index) interface{}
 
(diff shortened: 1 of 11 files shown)


internal/wasm/engine.go
```diff
@@ -9,11 +9,9 @@ type Engine interface {
...
 // ModuleEngine implements function calls for a given module.
 type ModuleEngine interface {
-	// FunctionAddress returns the absolute address of the compiled function for index.
-	// The returned address is stored and used as a TableInstance.Table element.
-	//
-	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-funcaddr
-	FunctionAddress(index Index) uintptr
+	// FunctionPointer returns the pointer the compiled function for index as interface{}.
+	// The returned interface is stored and used as a TableInstance.Table element.
+	FunctionPointer(index Index) interface{}
```
codefromthecrypt, line 14:
```suggestion
	// Note: the result is a `uintptr`, but returned as `interface{}` to prevent errors like below running with `-race`:
	// 	fatal error: checkptr: pointer arithmetic result points to invalid allocation
	FunctionPointer(index Index) interface{}
```

mathetake: oh actually the result is *compiledFunction, not uintptr now

codefromthecrypt: whoops this is not correct either :D

mathetake:
Had a discussion offline with @codefromthecrypt, and he will take over the task!

PR #358, opened 2022-03-11
Removes unrepresentative cases from adhoc

mathetake:
All of them are remnants from my initial amd64 JIT compiler and duplicates of spectests


codefromthecrypt (approved):
spec tests are very hard to troubleshoot as they aren't defined in go and very poorly documented if ever. Most times I have one failing, I lose a lot of time trying to understand what they intend to do. Let's not generally remove things only because spec tests exist!

That said, I assume the spec test overlap is a secondary reason and I'd prefer less drift to maintain, so go for it!

PR #360, opened 2022-03-11
Adds Module.Close API based on runtime.Finalizer

mathetake:
This commit adds Module.Close API, which can be used to indicate that
the module will no longer be used and setting finalizers for the allocated
resources. Notably, this will set finalizers on compiledFunction to deallocate
mapped memory regions as well as removing them from the store-wide
storage. As a result, users can re-instantiate a module for the same name
while having the safety -- Calling Module.Close is safe even when there are
outstanding function calls as we never explicitly deallocate but indirectly
do that via Goruntine's finalizers.

partially resolves: #293 

### DIFF
--- a/internal/wasm/jit/engine.go
+++ b/internal/wasm/jit/engine.go
@@ -347,7 +347,7 @@ func (e *engine) Compile(importedFunctions, moduleFunctions []*wasm.FunctionInst
 			compiled, err = compileHostFunction(f)
 		}
 		if err != nil {
-			_ = modEngine.Release()
+			modEngine.Close()
 			return nil, fmt.Errorf("function[%d/%d] %w", i, len(moduleFunctions)-1, err)
 		}
 		modEngine.compiledFunctions = append(modEngine.compiledFunctions, compiled)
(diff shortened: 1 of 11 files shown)


mathetake:
OK almost ready, but I would like to at least representative test case in conccurent_test.go.. other than that I think this is ready

mathetake:
will do the next week!

tests/engine/concurrency_test.go
```diff
@@ -54,3 +56,103 @@ func runAdhocTestUnderHighConcurrency(t *testing.T, newRuntimeConfig func() *waz
 	}
 	wg.Wait()
 }
+
+func singleModuleHighConcurrency(t *testing.T, newRuntimeConfig func() *wazero.RuntimeConfig) {
```
mathetake, line 60: so this e2e test shows what we can do now

internal/wasm/jit/engine.go
```diff
@@ -33,6 +34,11 @@ type (
 		// parentEngine holds *engine from which this module engine is created from.
 		parentEngine           *engine
 		importedFunctionCounts int
+
+		// Closed indicates whether Close was called for this moduleEngine, and used to prevent
+		// double-free of compiledFunctions.
+		// Note: intentionally use uint32 instead of bool for atomic operation.
```
codefromthecrypt, line 40: ❤️

codefromthecrypt (approved):
Thank you so much for help on this! Close is the most important API recently, as it allows people to instantiate modules with a different name many many times.

PR #362, opened 2022-03-11
WASI: `fs.FS` as Preopen parameter with implementation of `fd_prestat_get`

nullpo-head:
This PR changes the input type of preopen from `wasi.FS` to `fs.FS`, which is more natural for users.
This PR also includes implementation of `FdPrestatGet` and `PathOpen` flags, which are necessary for multiple pre-open directories to work correctly.


README.md
```diff
@@ -26,6 +28,69 @@ func main() {
...
+System Interface ([WASI](https://github.com/WebAssembly/WASI)). WASI defines
+how WebAssembly programs interact with the host embedding them. For example,
+WASI defines functions for reading the time, or a random number. Let's dig
+deeper into this with file I/O.
+
+Since we are Go developers, let's use Go's `fs.FS` interface to connect the
+dots. This [example](examples/testdata/file_system.wasm) allows a WebAssembly Module access read an `embed.FS` you
+define:
+
+```golang
```
codefromthecrypt, line 52: I like this as an examples directory but it has too much going on for the top-level README. I'll think about an alternative while moving along!

internal/wasi/fs.go
```diff
@@ -5,100 +5,137 @@ import (
...
-		pFlags |= os.O_RDWR
-	}
-	if oFlags&wasi.O_CREATE != 0 {
-		pFlags |= os.O_CREATE
-	}
-	if oFlags&wasi.O_EXCL != 0 {
-		pFlags |= os.O_EXCL
+// OpenFile implements wasi.OpenFileFS.OpenFile.
+func (dir DirFS) OpenFile(name string, flag int, perm os.FileMode) (fs.File, error) {
+	if !fs.ValidPath(name) || runtime.GOOS == "windows" && strings.ContainsAny(name, `\:`) {
```
mathetake, line 22: I would avoid doing `runtime.GOOS == "windows"` everytime `OpenFile` is called.. one idea is to have `const isWindows = runtime.GOOS == "windows"` and use it here (and eventually it will be optimized by compiler as it's const.

internal/wasi/fs.go
```diff
@@ -111,17 +148,159 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
...
+	entries []fs.DirEntry
+	// The memFSEntry this memFileInfo is about.
+	fsEntry *memFSEntry
+}
+
+func newMemDir(dir *memFSEntry, baseName string) *memDir {
+	// Cache the file list of the directory for ReadDir to return the result in the consistent order.
+	// Note that it's ok to return stale file list if the directory is modified after Open.
+	// The result of ReadDir is undefined in POSIX in that situation, so WASI will be the same.
+	entries := make([]fs.DirEntry, 0)
```
mathetake, line 175:
```suggestion
	entries := make([]fs.DirEntry, 0, len(dir.Entries))
```

internal/wasi/fs.go
```diff
@@ -111,17 +148,159 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
...
+	for name, entry := range dir.Entries {
+		entries = append(entries, &memFileInfo{name: path.Base(name), fsEntry: entry})
+	}
+	sort.Slice(entries, func(i, j int) bool { return entries[i].Name() < entries[j].Name() })
```
mathetake, line 179: q: is sorting dirs necessary?

internal/wasi/fs.go
```diff
@@ -111,17 +148,159 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
...
+		return 0, fmt.Errorf("invalid whence: %d", whence)
+	}
+	return d.offset, nil
+}
+
+// ReadDir implements fs.ReadDirFile.
+func (d *memDir) ReadDir(n int) ([]fs.DirEntry, error) {
+	remaining := int64(len(d.entries)) - d.offset
+	if remaining == 0 &&
+		n > 0 /* only when n > 0, since ReadDir should return empty slice instead of io.EOF when n <= 0 */ {
```
mathetake, line 222: large `/**/` is a bit less readable so maybe split the line?

internal/wasi/fs.go
```diff
@@ -5,100 +5,137 @@ import (
...
+	"path"
 	"runtime"
+	"sort"
 	"strings"
-
-	"github.com/tetratelabs/wazero/wasi"
+	"time"
 )
 
+// DirFS is a file system that implements fs.FS and wasi.OpenFileFS.
```
codefromthecrypt, line 15: do you know any go api for adding a file? maybe there's an open, closed or locked issue on go's repo if not.

internal/wasi/fs.go
codefromthecrypt, line 22: comment what all this is about, specifically the windows if

internal/wasi/wasi.go
```diff
@@ -1296,19 +1313,13 @@ func (a *wasiAPI) FdTell(ctx wasm.Module, fd, resultOffset uint32) wasi.Errno {
...
-	case 2:
-		writer = cfg.stderr
-	default:
-		f, ok := cfg.opened[fd]
-		if !ok || f.file == nil {
-			return wasi.ErrnoBadf
-		}
-		writer = f.file
+	f, ok := cfg.opened[fd]
+	if !ok || f.file == nil {
```
mathetake, line 1317: comment on what `f.file == nil` means?

internal/wasi/fs.go
```diff
@@ -5,100 +5,137 @@ import (
...
+// It is similar to os.DirFS, but supports file creation by OpenFile.
+// See os.DirFS.
 type DirFS string
```
codefromthecrypt, line 18: a type of string deserves a comment why it is string

internal/wasi/wasi.go
```diff
@@ -1296,19 +1313,13 @@ func (a *wasiAPI) FdTell(ctx wasm.Module, fd, resultOffset uint32) wasi.Errno {
...
+		return wasi.ErrnoBadf
+	}
+	writer, ok := f.file.(io.Writer)
+	if !ok {
```
mathetake, line 1321: q: in what case `!ok` in real world?

internal/wasi/wasi.go
```diff
@@ -1494,25 +1567,22 @@ type Config struct {
...
-	opened map[uint32]fileEntry
+	opened  map[uint32]fileEntry
 	// timeNowUnixNano is mutable for testing
 	timeNowUnixNano func() uint64
 	randSource      func([]byte) error
 }
 
 func (c *Config) Stdin(reader io.Reader) {
-	c.stdin = reader
+	c.opened[0] = fileEntry{file: &readerWriterFile{Reader: reader}}
```
mathetake, line 1577: maybe better have `0, 1, 2` as consts?

internal/wasi/fs.go
```diff
@@ -5,100 +5,137 @@ import (
...
+// Open implements fs.FS.Open.
+func (dir DirFS) Open(path string) (fs.File, error) {
+	return dir.OpenFile(path, os.O_RDONLY, 0)
 }
 
-type MemFS struct {
-	Files map[string][]byte
-}
+// MemFS is an in-memory file system that implements fs.FS and wasi.OpenFileFS.
+type MemFS map[string]*memFSEntry
```
codefromthecrypt, line 34: I'm not sure there's no read/write FS yet, but in any case, until we find that out we can say why in the comment.

```suggestion
// This is the only implementation that allows the Wasm to create or change files.
type MemFS map[string]*memFSEntry
```

internal/wasi/fs.go
```diff
@@ -5,100 +5,137 @@ import (
...
+	// Path "." indicates m itself as a directory.
+	if name == "." {
+		return newMemDir(dir, "."), nil
+	}
+
+	// Traverse directories.
+	files := strings.Split(name, "/")
+	for _, file := range files[:len(files)-1] {
+		var ok bool
+		dir, ok = dir.Entries[file]
```
codefromthecrypt, line 53: do you mean to reassign dir on each iteration of this function? since the `dir` is defined before this loop and also used after this loop, it feels a bit off. if this is intentional, maybe add a comment that we want `dir` to be effected on each path

internal/wasi/fs.go
```diff
@@ -5,100 +5,137 @@ import (
...
 	}
-	nread := copy(p, f.buf[f.offset:])
+	nread := copy(p, f.fsEntry.Contents[f.offset:])
 	f.offset += int64(nread)
 	return nread, nil
 }
 
+// Close implements fs.File.Close
+func (f *memFile) Close() error {
+	return nil
```
codefromthecrypt, line 132: trailing comment on no-op, but maybe we want to nil its body to allow faster GC on close?

internal/wasi/fs.go
```diff
@@ -111,17 +148,159 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
...
-// Close implements io.Closer
-func (f *memFile) Close() error {
-	if f.flush != nil {
-		f.flush(f.buf)
+// memDir represents an opened memFS directory.
+// memDir implements fs.File, fs.ReadDirFile, and io.Seeker.
+type memDir struct {
+	// name of the opened file, not a full path but a base name.
+	name string
+	// Seek offset of this opened directory.
```
codefromthecrypt, line 163: is this ever negative?

internal/wasi/fs.go
```diff
@@ -111,17 +148,159 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
...
+	offset int64
+	// Sorted file list of the dirEntries for ReadDir and Seek.
+	entries []fs.DirEntry
+	// The memFSEntry this memFileInfo is about.
+	fsEntry *memFSEntry
+}
+
+func newMemDir(dir *memFSEntry, baseName string) *memDir {
+	// Cache the file list of the directory for ReadDir to return the result in the consistent order.
```
codefromthecrypt, line 172: caching a read could be a performance problem. If only caching for tests, maybe sort in the test instead?

internal/wasi/fs.go
```diff
@@ -111,17 +148,159 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
...
+}
+
+// Close implements fs.File.Close
+func (d *memDir) Close() error {
+	return nil
+}
+
+// Seek implements io.Seeker
+func (d *memDir) Seek(offset int64, whence int) (int64, error) {
+	switch whence {
```
codefromthecrypt, line 205: should there be bounds checking to ensure d.offset doesn't end up negative?

internal/wasi/fs.go
```diff
@@ -111,17 +148,159 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
...
+// IsDir implements fs.FileInfo.IsDir
+func (f *memFileInfo) IsDir() bool { return f.Mode().IsDir() }
+
+// Sys implements fs.FileInfo.Sys
+func (f *memFileInfo) Sys() interface{} { return nil }
+
+// Info implements fs.DirEntry.Info
+func (f *memFileInfo) Info() (fs.FileInfo, error) { return f, nil }
+
+// readerWriterFile wraps io.Reader and io.Writer as fs.File.
```
codefromthecrypt, line 274: I don't really understand the relevance to "as fs.File" because fs.File has a different signature. It is also confusing that reader and writer are exported.

If you want to make a utility that is readerWriterFile, maybe that comment would be better

```
// compile-time check that readerWriterFile wraps io.Reader and io.Writer as fs.File.
_ io.File = &readerWriterFile{}
```

internal/wasi/fs.go
```diff
@@ -111,17 +148,159 @@ func (f *memFile) Seek(offset int64, whence int) (int64, error) {
...
+func (f *readerWriterFile) Write(p []byte) (int, error) {
+	if f.Writer == nil {
+		return len(p), nil
+	}
+	return f.Writer.Write(p)
+}
+
+// Close implements fs.File.Close
+func (f *readerWriterFile) Close() error {
 	return nil
```
codefromthecrypt, line 305: same comment.. if noop comment why or make it op :D

internal/wasi/fs_test.go
```diff
@@ -12,10 +17,242 @@ import (
...
+// It includes testing the behavior of fs.FS.Open, and fs.File.(Read, Seek, Stat, ReadDir)
+// See io/fs.FS
+// See io/fs.File
+// See fstest.TestFS
+func TestMemFS_FSInterface(t *testing.T) {
+	memFS := &MemFS{
+		"simple file": {Contents: []byte("simple cont")},
+		"directory": {
+			Entries: map[string]*memFSEntry{
+				"file1 inside directory": {Contents: []byte("simple cont file1 inside directory")},
```
codefromthecrypt, line 38: maybe use more realistic paths here? it would make it easier to read and know the intent.

ex would it be "directory/file1" because this includes the name of the key?

also the contents if dummy can be shorter like single character or single word

internal/wasi/fs_test.go
```diff
@@ -12,10 +17,242 @@ import (
 func TestMemFile_Read_Seek(t *testing.T) {
 	expectedFileContent := []byte("wazero") // arbitrary contents
 	memFile := &memFile{
-		buf: expectedFileContent,
+		fsEntry: &memFSEntry{Contents: expectedFileContent},
 	}
 	// TestReader tests that io.Reader correctly reads the expected contents.
 	// It also tests io.Seeker when it's implemented, which memFile does.
 	err := iotest.TestReader(memFile, expectedFileContent)
```
codefromthecrypt, line 24: maybe mention why we aren't using more of the iotest, yet, at the top of this file.. ex an overview comment about use of iotest and fstest (including status) then the functions

codefromthecrypt (commented):
Thanks for getting this started. I think there will be a fair amount of work to finish this up, and we don't want to hold the supporting changes hostage through that. Can you extract the wasi api changes and tests into a separate PR from the fs.FS stuff?

codefromthecrypt:
I'm going to take over this PR for some several minutes, rebase it after extracting the WASI functions to another PR then force push a more isolated variant to proceed with. bear with me a bit

internal/wasi/fs.go
nullpo-head: > If only caching for tests

cache is not for tests, just to avoid sorting the file list every time ReadDir is called. (btw, [sort is required for ReadDir](https://github.com/tetratelabs/wazero/pull/362/files#r828715370)) If your intention is caching it is weird or too much, yes we can sort the list inside ReadDir instead of here. wdyt? thanks!

internal/wasi/fs.go
nullpo-head: thanks for nice catch! Yes, there must be bounds checking. Let me add..

internal/wasi/fs.go
nullpo-head: I also investigated whether there is any Go's writable FS interface. There is an issue discussing creating/writing file on FS, but it lacks the attention from the Go team. https://github.com/golang/go/issues/45757

nullpo-head: ah, or if your intention is literally API to add a new file, it's exactly `os.OpenFile`, which also can do file creation. `os.Create` is simpler API that creates a file with default flags. It does `OpenFile(name, O_RDWR|O_CREATE|O_TRUNC, 0666)` internally.

internal/wasi/fs.go
nullpo-head: Will do. Thanks! ps: DirFS also allows file creation.

codefromthecrypt:
once https://github.com/tetratelabs/wazero/pull/388 is merged I'll rebase this and hand it back over to @nullpo-head. hopefully next time we can keep fixing WASI APIs separate from new API work that uses them.

codefromthecrypt:
okie dokie over to you!

internal/wasi/fs.go
codefromthecrypt: yeah I think it is better to keep the abstraction coherent. fs.FS shouldn't bleed too much into how we implement WASI, that way in case it isn't used, there's no perf hit. In the call site that requires caching, you can add a comment that it is required by the definition of fs.FS

codefromthecrypt: accidentally added a comment and clicked resolve. What I meant was add like you said sorting to ReadDir. Don't cache on first impl as it is possible no-one calls readDir and each thing we add to this PR == more code to review and edge case unit tests

internal/wasi/wasi.go
nullpo-head: It was when `f` was a pre-opened directory, but I've changed the logic so that `f.file` never becomes `nil`.

internal/wasi/fs.go
nullpo-head: During changing it, I've realized that `runtime.GOOS` is compile-time constant string in Golang, not a value obtained at runtime. And `runtime.GOOS == "windows"` is completely optimized out and disappears on non-Windows binaries, so it never runs at runtime. So, I think it would be better to leave it as is this time, rather than introducing build flags only for this. wdyt?

e.g.
on Windows, it calls `strings.IndexAny` (= `ContainsAny`) after `ValidPath`. 
```asm
0000000000591ec0 <github.com/tetratelabs/wazero/internal/wasi.DirFS.OpenWASI>:
  --snip--
  591f02:   e8 39 ff ee ff          callq  481e40 <io/fs.ValidPath>
  591f07:   84 c0                   test   %al,%al
  591f09:   74 24                   je     591f2f <github.com/tetratelabs/wazero/internal/wasi.DirFS.OpenWASI+0x6f>
  591f0b:   48 8b 44 24 70          mov    0x70(%rsp),%rax
  591f10:   48 8b 5c 24 40          mov    0x40(%rsp),%rbx
  591f15:   48 8d 0d 5e b8 05 00    lea    0x5b85e(%rip),%rcx        # 5ed77a <go.string.*+0x222>
  591f1c:   bf 02 00 00 00          mov    $0x2,%edi
  591f21:   e8 da b8 f1 ff          callq  4ad800 <strings.IndexAny>
--snip--
```

on Linux, that path disappears.
```asm
0000000000583440 <github.com/tetratelabs/wazero/internal/wasi.DirFS.OpenWASI>:
--snip--
  583482:   e8 99 28 ef ff          callq  475d20 <io/fs.ValidPath>
  583487:   84 c0                   test   %al,%al
  583489:   0f 84 e2 00 00 00       je     583571 <github.com/tetratelabs/wazero/internal/wasi.DirFS.OpenWASI+0x131>
  58348f:   48 8b 94 24 88 00 00    mov    0x88(%rsp),%rdx
  583496:   00
  583497:   0f ba e2 09             bt     $0x9,%edx
  58349b:   73 17                   jae    5834b4 <github.com/tetratelabs/wazero/internal/wasi.DirFS.OpenWASI+0x74>
  58349d:   48 f7 c2 08 00 00 00    test   $0x8,%rdx
  5834a4:   be 01 00 00 00          mov    $0x1,%esi
  5834a9:   ba 02 00 00 00          mov    $0x2,%edx
  5834ae:   48 0f 45 f2             cmovne %rdx,%rsi
```

btw, it looks the this line originally comes from Golang's standard library. https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/os/file.go;l=644

internal/wasi/fs_test.go
nullpo-head: I've changed the names to more realistic paths, but let me know if those are not what you intended. THanks!

nullpo-head:
ptal @codefromthecrypt @mathetake thanks!

codefromthecrypt (changes requested):
memFS looks suspiciously like `fstest.MapFS` Please delete it and use `fstest.MapFS` if you need a dummy filesystem besides embedFS. The main point of this change is to stop bespoke APIs and use fs.FS instead.

That does introduce discomfort, for example if someone doing something in a non
I would be very very surprised if there isn't an existing fs.FS library out there that supports writes.

If we couple `OpenFile(path string, flag int, perm os.FileMode) (fs.File, error)` with `fs.FS` we almost guarantee people can't bring their own impls as an existing impl wouldn't be using our interfaces. Instead, until we find an alternative to `OpenFile` (this is a TODO to find one!), make a separate Config option for a write path that accepts this interface.

ex. `config.WithOpenFile(path, OpenFile)` and put a TODO to look more into options for writes that don't imply completely custom implementations of fs.FS

wasi/wasi.go
```diff
@@ -12,38 +13,22 @@ const (
...
 
-// FS is an interface for a preopened directory.
-type FS interface {
-	// OpenWASI is a general method to open a file, similar to
-	// os.OpenFile, but with WASI flags and rights instead of POSIX.
-	OpenWASI(dirFlags uint32, path string, oFlags uint32, fsRights, fsRightsInheriting uint64, fdFlags uint32) (File, error)
+	// OpenFile opens the named file with specified flag in the same way as os.OpenFile.
+	// For example, if the path does not exist and flag includes os.O_CREATE, it will create a new file with perm.
+	//
+	// Unlike os.OpenFile, OpenFile must reject paths that do not satisfy `fs.ValidPath(path)`,
```
codefromthecrypt, line 25: I have a hard time believing there are any strict requirements that need to be made here based on the WASI spec. It doesn't have spec tests and at best very vaguely documented. Any constraints that make something behave different than something else in Go should be not just mentioned but also answered why. (Ex in RATIONALE is a good place as there's not likely any spec or spectest to say why)

Can you please document where these are coming from?

codefromthecrypt:
I'm going to fork this and try to salvage what I can from it, but generally this is far from ideal and time is better spent on another path. Hence closing it.

internal/wasi/wasi.go
```diff
@@ -1558,25 +1569,29 @@ type Config struct {
...
 
+// FD number constants of standard input / output / error.
+const (
+	stdinFD = iota
+	stdoutFD
+	stderrFD
+)
+
 func (c *Config) Stdin(reader io.Reader) {
-	c.stdin = reader
```
codefromthecrypt, line 1571: not currently planning to do anything to try to make stdin/stdout/stderr act like normal files as they aren't and it is more distracting forcing them though a struct that only has one side useful than the special casing in the first place.

codefromthecrypt:
There's nothing to salvage from this PR except experience. Once things are stable we can do another impl, but research and citation here are a prerequisite so that we don't fall into the same hole again https://github.com/tetratelabs/wazero/issues/390

Thanks for giving this a try

internal/wasi/fs.go
mathetake: oh that's good to know! thank you for confirming that

internal/wasi/fs.go
```diff
@@ -5,107 +5,317 @@ import (
...
-	}
-	mode := fs.FileMode(0644)
-	if oFlags&wasi.O_DIR != 0 {
-		mode |= fs.ModeDir
+// OpenFile implements wasi.OpenFileFS.OpenFile.
+func (dir DirFS) OpenFile(name string, flag int, perm os.FileMode) (fs.File, error) {
+	// fs.FS requires a FS to reject a path that doesn't satisfy fs.ValidPath.
+	if !fs.ValidPath(name) ||
+		// '\' works as alternative path separater and ':' allows to express a root drive directory.
+		// fs.FS implementation must reject those path on Windows. See Note in the doc of fs.ValidPath.
```
codefromthecrypt, line 26: this is a good example of a leaky abstraction. WASI isn't windows, and there's nothing about fs.FS that should ever allow windows paths to leak into it. Ex. if someone exposes their "c:" drive, that's not on us to enforce. The only sensible file mappings would be chrooted if even an actual filesystem.

codefromthecrypt: In other words windows-based file handing in a barely specified posix system layer is a non-goal. We can use this to simplify discussions in future work. By adding code to address things like OS-specific things inside WASI, we are actually making the problem more confused, especially to people new who may not know Wasm is effectively a virtual machine, and if shares anything in common with OS, it would be unix, not windows, that we'd optimize for.

PR #366, opened 2022-03-14
Refactor amd64 JIT compiler

mathetake:
This removes all the copy-and-pasted assembler code and
centralizes the usage of golang-asm in jit_amd64.go

In follow-up PRs, I will share majority of unit tests in amd64_test.go
with arm64 and introduce assembler interface.

### DIFF
--- a/internal/wasm/jit/jit_amd64.go
+++ b/internal/wasm/jit/jit_amd64.go
@@ -185,11 +185,11 @@ func (c *amd64Compiler) compileHostFunction() error {
 	// First we must update the location stack to reflect the number of host function inputs.
 	c.pushFunctionParams()
 
-	if err := c.callGoFunction(jitCallStatusCodeCallHostFunction, c.f.Index); err != nil {
+	if err := c.compileCallHostFunction(); err != nil {
 		return err
 	}
 
-	return c.returnFunction()
+	return c.compileReturnFunction()
 }
 
 // compile implements compiler.compile for the amd64 architecture.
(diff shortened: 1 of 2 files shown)


internal/wasm/jit/jit_amd64.go
```diff
@@ -4999,211 +3958,94 @@ func (c *amd64Compiler) returnFunction() error {
...
-		setFunctionCallAddress := c.newProg()
-		setFunctionCallAddress.As = x86.AMOVL
-		setFunctionCallAddress.From.Type = obj.TYPE_CONST
-		setFunctionCallAddress.From.Offset = int64(index)
-		setFunctionCallAddress.To.Type = obj.TYPE_MEM
-		setFunctionCallAddress.To.Reg = reservedRegisterForCallEngine
-		setFunctionCallAddress.To.Offset = callEngineExitContextBuiltinFunctionCallAddressOffset
-		c.addInstruction(setFunctionCallAddress)
-	}
+// TODO: this is deletable.
```
nullpo-head, line 3983: nit: the comment on why this is deletable might help. Maybe it means `compileCallHostFunction` can be just replaced with `compileCallHostFunction`?

nullpo-head (approved):
Thanks!

internal/wasm/jit/jit_amd64.go
mathetake: I intended to solve this shortly, so I choose to leave as-is!

PR #367, opened 2022-03-14
Adds StartWASICommandWithConfig to defer configuration

### LINKED ISSUE #364: wasi: recommended usage for module re-use
Hi 👋 ,

The wazero public interface has been going through major improvements recently, and before it gets finalized, I was wondering if you could shed some light on how how to re-use a "wasi module ".

Note: if this use-case is not something you'd want to support, I fully understand, I'm just trying to make sure I'm not missing anything obvious.

```go
r := wazero.NewRuntimeWithConfig(cfg)
decoded,  := r.CompileModule(source) // source parsed only once 


// this configuration is unique to each invocation based on message content
wasiConfig := &wazero.WASIConfig{
		Stdin:  stdinBuf,
		Stdout: stdoutBuf,
		Stderr: stderrBuf,
		Args:   []string{"rotate", "angle=90", "dir=cw"}, // <-- this comes from the body of the message received.
	}

if _, err := r.InstantiateModule(wazero.WASISnapshotPreview1WithConfig(wasiConfig)); err != nil {
		log.Fatal(err) // 👈🏼 this fails for the second r. r.InstantiateModule  call with `module wasi_snapshot_preview1 has already been instantiated`
	}
if _, err := wazero.StartWASICommand(r, decoded); err != nil {
		log.Fatal(err)
}
```

I can't really change the module name since it's required for wasi support.
```
(import "wasi_snapshot_preview1" "args_get" (func $fimport$1.linked.1052 (param i32 i32) (result i32)))
```

I understand how to achieve this with a regular (non wasi) wasm module, but I was hoping to not have to marshal strings via wasm memory.

One viable avenue is to wrap `WASIConfig` into a message/context aware struct and replacing the static args/env values.
Is that what you'd recommend?

Thanks!

codefromthecrypt:
StartWASICommandWithConfig is like StartWASICommand, except you can override configuration based on the importing
module. For example, you can use this to define different args depending on the importing module.

```go
// Initialize base configuration:
r := wazero.NewRuntime()
config := wazero.NewWASIConfig().WithStdout(buf)
wasi, _ := r.NewHostModule(wazero.WASISnapshotPreview1WithConfig(config))
decoded, _ := r.CompileModule(source)

// Assign configuration only when ready to instantiate.
module, _ := StartWASICommandWithConfig(r, decoded, config.WithArgs("rotate", "angle=90", "dir=cw"))
```

This also changes the config to be immutable since it may be reused many
times. Instead of assigning fields directly, use `WithXXX`.

Finally, this changes the examples to encourage using `.Close()`

Fixes #364


examples/testdata/wasi_arg.wat
```diff
@@ -0,0 +1,61 @@
+;; $wasi_arg is a WASI command which copies null-terminated args to stdout.
+(module $wasi_arg
```
codefromthecrypt, line 2: added a %.wat program (it works in other runtimes too). This is intentionally a blackbox test (not sneaking looks at internals to assume things work.

examples/testdata/wasi_arg.wat
```diff
@@ -0,0 +1,61 @@
...
+    ;; See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#memories%E2%91%A7
+    (memory (export "memory") 1)
+
+    ;; _start is a special function defined by a WASI Command that runs like a main function would.
+    ;;
+    ;; See https://github.com/WebAssembly/WASI/blob/snapshot-01/design/application-abi.md#current-unstable-abi
+    (func $main (export "_start") (local $iovs i32) (local $ignored i32)
+        ;; $iovs are offset/length pairs in memory fd_write copies to the file descriptor.
+        ;; This function will only write one offset/length pair, corresponding to null-terminated args.
+        (local.set $iovs (i32.const 1024)) ;; 1024 is an arbitrary offset larger than the args.
```
codefromthecrypt, line 33: probably less clutter to define these as globals since you can do a constant initializer..

mathetake (approved):
Love how this is solved

codefromthecrypt:
will have to ignore pre-commit check until GH feels better 😷 

Screen Shot 2022-03-14 at 2 38 13 PM

codefromthecrypt:
thanks for the look and the kind words @mathetake!

pims:
A million thanks @codefromthecrypt for a very convenient API change 🙏🏼

PR #368, opened 2022-03-14
Adds stricter enforcement of Table element validation.

codefromthecrypt:
This enforces all ElementSegment rules prior to creating a new module
engine. This allows flexibility and less errors inside that function.

This also uses interface as table element to pass tests with -race

### DIFF
--- a/internal/leb128/leb128.go
+++ b/internal/leb128/leb128.go
@@ -32,7 +32,14 @@ var encodeCache = [0x80][]byte{
 // EncodeUint32 encodes the value into a buffer in LEB128 format
 //
 // See https://en.wikipedia.org/wiki/LEB128#Encode_unsigned_integer
-func EncodeUint32(value uint32) (buf []byte) {
+func EncodeUint32(value uint32) []byte {
+	return EncodeUint64(uint64(value))
+}
+
+// EncodeUint64 encodes the value into a buffer in LEB128 format
+//
+// See https://en.wikipedia.org/wiki/LEB128#Encode_unsigned_integer
+func EncodeUint64(value uint64) (buf []byte) {
 	if value < 0x80 {
 		return encodeCache[value]
 	}
--- a/internal/wasm/jit/engine.go
+++ b/internal/wasm/jit/engine.go
@@ -332,7 +332,7 @@ func (c *callFrame) String() string {
 }
 
 // NewModuleEngine implements internalwasm.Engine NewModuleEngine
-func (e *engine) NewModuleEngine(name string, importedFunctions, moduleFunctions []*wasm.FunctionInstance) (wasm.ModuleEngine, error) {
+func (e *engine) NewModuleEngine(name string, importedFunctions, moduleFunctions []*wasm.FunctionInstance, table *wasm.TableInstance, tableInit map[wasm.Index]wasm.Index) (wasm.ModuleEngine, error) {
 	imported := len(importedFunctions)
 	me := &moduleEngine{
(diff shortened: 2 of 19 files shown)


codefromthecrypt:
@mathetake I picked in your https://github.com/tetratelabs/wazero/pull/353 as I couldn't find an easier way besides a polymorphic type to hold the tables at the moment. The problems with refactor it are about sharing and using the right table context in a function call. Right now, it seems like punting to interface is the best way forward.

Before I work more on tests, can you give a quick look that the approach is sound?

mathetake:
looks fine to me

codefromthecrypt:
ok next steps are basically backfilling tests of which there are many!

mathetake:
https://github.com/tetratelabs/wazero/runs/5536658163?check_suite_focus=true

oh something weird  happens.. maybe double free of compiledFunctions..

codefromthecrypt:
ok rebased... now will go back and backfill the unit tests.

internal/leb128/leb128.go
```diff
@@ -32,7 +32,14 @@ var encodeCache = [0x80][]byte{
...
 // See https://en.wikipedia.org/wiki/LEB128#Encode_unsigned_integer
-func EncodeUint32(value uint32) (buf []byte) {
+func EncodeUint32(value uint32) []byte {
+	return EncodeUint64(uint64(value))
+}
+
+// EncodeUint64 encodes the value into a buffer in LEB128 format
+//
+// See https://en.wikipedia.org/wiki/LEB128#Encode_unsigned_integer
+func EncodeUint64(value uint64) (buf []byte) {
```
mathetake, line 42: maybe we could just rename this to EncodeVarUint as the simple doesn't rely on the bit length!

mathetake, line 42: maybe we could just rename this to EncodeVarUint as the simple doesn't rely on the bit length!

mathetake: ah nvm!

internal/wasm/jit/engine.go
```diff
@@ -368,14 +368,12 @@ func (e *engine) NewModuleEngine(name string, importedFunctions, moduleFunctions
 		e.addCompiledFunction(f, compiled)
 	}
 
+	for elemIdx, funcidx := range tableInit { // Initialize any elements with compiled functions
```
codefromthecrypt, line 371: oops I forgot tests for this, will update the PR

internal/wasm/table.go
```diff
@@ -18,68 +24,197 @@ type ElementSegment struct {
...
+//
+// Note: The global imported at globalIdx may have an offset value that is out-of-bounds for the corresponding table.
+type validatedElementSegment struct {
+	// opcode is OpcodeGlobalGet or OpcodeI32Const
+	opcode Opcode
+
+	// arg0 is the arg to opcode, which results in the offset to add to init indices.
+	//  * OpcodeGlobalGet: position in the global index namespace of an imported Global ValueTypeI32 holding the offset.
+	//  * OpcodeI32Const: a constant ValueTypeI32 offset.
+	arg0 uint32
```
mathetake, line 51: I would drop `0` as in anyway there would be at most one arg

mathetake (approved):

internal/wasm/table.go
codefromthecrypt: sgtm!

codefromthecrypt:
Thanks for the help @mathetake!

PR #373, opened 2022-03-15
jit: introduce arch-independent and assembler-free unit tests

mathetake:
This commit consolidates `jit_{amd64, arm64}_test.go` into `jit_impl_test.go`,
where we have architecture-independent unit tests over majority of compiler 
semantics on compiler interface, meaning that `golang-asm` is not used in
`jit_impl_test.go`.

This is towards #233, and especially useful as this allows us to narrow down 
the golang-asm usage, combined with #366.

In the follow-up PR, I will further refactor and introduce interface for architecture,
where we have documentation interface over each architecture dependent functions
and structures.


internal/wasm/jit/jit_amd64.go
```diff
@@ -3980,7 +3977,6 @@ func (c *amd64Compiler) compileReturnFunction() error {
 	return nil
 }
 
-// TODO: this is deletable.
```
mathetake, line 3983: turns out it's not that simple to delete this so removed the comment in this PR 😄

internal/wasm/jit/jit_test.go
```diff
@@ -1,147 +1,13 @@
+//go:build !amd64 && !arm64
```
nullpo-head, line 1: If this file builds only when !amd64 && !arm64, then maybe the 12th line isn't necessary anymore?
```go
if runtime.GOARCH != "amd64" && runtime.GOARCH != "arm64" {
```

nullpo-head (approved):

internal/wasm/jit/jit_test.go
codefromthecrypt: good point though if on a wrong arch I think that TestMain is better (as it covers the whole package)? maybe revert the compile flags and leave it instead? Either way though, remove one or the other.

internal/wasm/jit/jit_impl_test.go
```diff
@@ -0,0 +1,4448 @@
...
+					if shouldGoToElse {
+						require.Equal(t, elseLabelExitStatus, env.jitStatus())
+					} else {
+						require.Equal(t, thenLabelExitStatus, env.jitStatus())
+					}
+				})
+			}
+		})
+	}
+}
```
codefromthecrypt, line 4448: this this file is too large. can you split it into files so that each are (ideally)_ not more than 500 lines? I suspect that at least >1.0 feature can be separate file, and then the 1.0 split by category to achieve this.

codefromthecrypt (approved):
I like this, mainly think in the tests TestMain is easier vs adding build flags to do the same to each test. Also, let's be careful about massive files as usually there's some natural categorization we can employ to segment them. Otherwise anything we ever do taints the same file.

PR #374, opened 2022-03-15
Propagates module name to module engine panic nicer

codefromthecrypt:
Before, if there was an unexpected panic on munmap of a code section,
we'd only see "invalid argument" and have to guess why. This propagates
context and stubs out the finalizer so we can test a better error, like:

"jit: failed to munmap code segment for test.function[0]: invalid argument"

This also attempts to improve troubleshooting time by making all module
instantiation that used empty module name, use the test name instead.

### DIFF
--- a/internal/wasm/engine.go
+++ b/internal/wasm/engine.go
@@ -1,20 +1,27 @@
 package internalwasm
 
+import "io"
+
 // Engine is a Store-scoped mechanism to compile functions declared or imported by a module.
 // This is a top-level type implemented by an interpreter or JIT compiler.
 type Engine interface {
 	// NewModuleEngine compiles down the function instances in a module, and returns ModuleEngine for the module.
 	//
+	// * name is the name the module was instantiated with used for error handling.
 	// * importedFunctions: functions this module imports, already compiled in this engine.
 	// * moduleFunctions: functions declared in this module that must be compiled.
 	//
 	// Note: Input parameters must be pre-validated with internalwasm.Module Validate, to ensure no fields are invalid
 	// due to reasons such as out-of-bounds.
-	NewModuleEngine(importedFunctions, moduleFunctions []*FunctionInstance) (ModuleEngine, error)
+	NewModuleEngine(name string, importedFunctions, moduleFunctions []*FunctionInstance) (ModuleEngine, error)
 }
 
 // ModuleEngine implements function calls for a given module.
 type ModuleEngine interface {
+	// Closer releases the resources allocated by functions in this ModuleEngine.
+	io.Closer
+	// ^^ io.Closer not due to I/O, but to allow future static analysis to catch leaks (unclosed Closers).
+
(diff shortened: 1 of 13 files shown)


internal/wasm/engine.go
```diff
@@ -1,20 +1,26 @@
...
 	// Note: Input parameters must be pre-validated with internalwasm.Module Validate, to ensure no fields are invalid
 	// due to reasons such as out-of-bounds.
-	NewModuleEngine(importedFunctions, moduleFunctions []*FunctionInstance) (ModuleEngine, error)
+	NewModuleEngine(name string, importedFunctions, moduleFunctions []*FunctionInstance) (ModuleEngine, error)
 }
 
 // ModuleEngine implements function calls for a given module.
 type ModuleEngine interface {
+	// Closer releases the resources allocated by functions in this ModuleEngine.
+	io.Closer
```
codefromthecrypt, line 22: ps this gives us space to use errors later as needed even if we currently implement close async via finalizer.

internal/wasm/jit/engine_test.go
```diff
@@ -298,11 +336,6 @@ func TestRelease(t *testing.T) {
 				require.NotContains(t, e.compiledFunctions, f, i)
 			}
 
-			// Release the reference to ModuleEngine -> *compiledFunctions for non-imported compiledFunctions
-			// must be munmapped.
-			modEngine = nil
-			runtime.GC()
```
mathetake, line 304: maybe this could be the cause of flake on win

codefromthecrypt: ok I'll put it back! I was assuming GC was because we were testing finalization side-effects. guess not heh.

codefromthecrypt: actually there's another close later that was still using finalizer..

mathetake: 😄

internal/wasm/engine.go
nullpo-head: just curious: is it a common practice in Golang for a non-IO interface to adopt `io.Closer` because it has `Close() error`?

nullpo-head (approved):

codefromthecrypt:
> just curious: is it a common practice in Golang for a non-IO interface to adopt io.Closer because it has Close() error?

good question. What I'm hoping for really is static analysis that helps you detect something you forgot to close. That said, haven't found a linter yet (ex not in golint-ci) https://github.com/dominikh/go-tools/issues/139

I'll add a comment why

codefromthecrypt:
I added code to panic on zero length code segments because that's a bug on our end (even in tests) as there's no valid code length zero (opcodeend is required). I also backfilled tests on freeing memory never mapped.

codefromthecrypt:
Thanks for the help @mathetake and @nullpo-head: the code is better with your feedback!

PR #375, opened 2022-03-15
Propagates error return Close to public API

codefromthecrypt:
Before, we only handled errors internally and swallowed them when re-exporting to the public API.


mathetake (approved):

codefromthecrypt:
ps we probably should have a convenience Runtime.Close to close all the modules in the right order.

PR #386, opened 2022-03-17
Adds description of WebAssembly feature status

codefromthecrypt:
This is as much to reduce confusion for us as it is to reduce confusion for end users, due to the last standard being published over two years ago, yet most compilers targeting >1.0 features by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ default. However, the following status covers what's currently possible with
 Note: While the above are specified in a WebAssembly GitHub repository, they
 are not W3C recommendations (standards). This means they can change further
 between now and any future update to the W3C WebAssembly core specification.
-Due to this we cannot guarantee future compatability. Please encourage the
+Due to this, we cannot guarantee future compatability. Please encourage the
 [WebAssembly community](https://www.w3.org/community/webassembly/) to formalize
 features you rely on, specifically to read the W3C recommendation (REC) phase.
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ The last W3C REC was at the end of 2019. There were other features that didn't
 make the cut or were developed in the years since. The community unofficially
 refers to these as [Finished Proposals](https://github.com/WebAssembly/proposals/blob/main/finished-proposals.md).
 
-To ensure compatability, wazero does not opt-into any non-standard feature by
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,6 +49,51 @@ r := wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfigInterpreter())
...
+default. However, the following status covers what's currently possible with
+`wazero.RuntimeConfig`.
+
+| Feature                               | Status |
+|:-------------------------------------:|:------:|
+| mutable-global                        |   ✅   |
+| nontrapping-float-to-int-conversions  |   ❌   |
+| sign-extension-ops                    |   ✅   |
+| multi-value                           |   ❌   |
+| JS-BigInt-integration                 |  N/A   |
```
codefromthecrypt, line 82: I didn't look closely, but I don't know why this would effect the core interpretation of Wasm. If wrong, I'll switch to X

codefromthecrypt:
fyi @nullpo-head I will add another section for WASI in a different PR. This has even more complicated caveats including that those authoring the spec don't recommend implementing it entirely as the subsequent version is incompatible due to complete redo of how rights work.

mathetake (approved):

README.md
```diff
@@ -49,6 +49,48 @@ r := wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfigInterpreter())
...
+meanwhile users can work around this using tools such as `wat2wasm` to compile
+the text format into the binary format.
+
+#### Post 1.0 Features
+
+The last W3C REC was at the end of 2019. There were other features that didn't
+make the cut or were developed in the years since. The community unofficially
+refers to these as [Finished Proposals](https://github.com/WebAssembly/proposals/blob/main/finished-proposals.md).
+
+To ensure compatability, wazero does not opt-into any non-standard feature by
```
nullpo-head, line 72: `opt-into` is not in a dictionary, so it may be a little bit difficult to guess the meaning for non-native English readers (yeah, like me). Maybe the same meaning as "opt-in"?

nullpo-head (approved):

README.md
codefromthecrypt: missing space :D

codefromthecrypt: thanks for looking close enough to notice, and thinking to say something. fixed!

PR #391, opened 2022-03-18
jit: make compiler implementations build-tag free

mathetake:
This commit makes compiler implementations build tag free.
Notably, now the arm64 and amd64 compilers are compilable 
on any architecture. Anything arch-specific is isolated in files
named with `arch_*` prefixes.

This reduces the pain to make some broad change across
multiple architectures and allow developers to notice errors
on whichever architecture they are working on.


codefromthecrypt (approved):
easy peasy

PR #392, opened 2022-03-18
CI: change non-amd64 job into scratch platform tests

### LINKED ISSUE #385: Add CI for the "FROM scratch" platform
Per https://github.com/tetratelabs/wazero/pull/384 wazero supports scratch. scratch is different than "distroless" as the latter still includes an operating system (albeit slim, often debian). We should be able to add a `scratch` platform to our CI dimensions that run spec tests, and verify that on the same archs as ubuntu.

one way is to build a docker image `FROM scratch` during CI, which copies the spec binary and sets the entrypoint to run it. Another way could be to have a standing `FROM scratch` image with a volume mount which would house the test binary. Other ways are possible too. Regardless, it would be best to run this like a normal OS dimension even if it relies on Docker, so that the CI output looks the same as an OS.

mathetake:
fixes #385

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -113,7 +113,7 @@ jobs:
           platforms: ${{ matrix.arch }}
 
       - name: Build scratch container
-        run:
+        run: |
           echo 'FROM scratch' >> Dockerfile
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
love it

Issue #390, opened 2022-03-18
wasi: allow users to configure a filesystem that can accept new files


codefromthecrypt:
WASI defines portable system calls to manipulate files. We currently use fs.FS to map to many of these. read-only operations are implicit, but the status quo goes beyond this in the standard library.

For example, fs.FS can write existing files as implementations are free to go beyond the functionality defined on fs.File. Some of this is encouraged. Concretely, [os.File](https://github.com/golang/go/blob/master/src/os/file.go) goes beyond with mutations such as chmod and write. [os.FS](https://github.com/golang/go/blob/master/src/os/file.go)

Also, there are standard extensions to `fs.FS` including `fs.ReadDirFS`. The point is that in many cases, there is an existing mapping of a go interface to WASI functions, even if it isn't required by `fs.FS`. By safe casts, you can get most functionality needed.

However, there are gaps, notably creating new files. This issue is solved when the core WASI APIs requested by end users have a corresponding implementation. This is ideally done without introducing any wazero-specific interfaces. As such research is needed and validation is needed prior to introducing an interface not in Go's standard library.

Moreover, as the README suggests, completion of all WASI functions is a non-goal. In other words, if there is a tricky API, and hasn't been asked directly by end-users yet.. wait until that's the case, then solve it. WASI is completely changing its filesystem API so we can conserve energy by only focusing on what users need today.

Solving this issue is a high priority and 1.0 will be held back until it is solved. However, it is likely this will be solved way before 1.0, like in a week or two.

References:
* Go has an issue tracking approaches to weave in write support to fs.FS https://github.com/golang/go/issues/45757
* Last year,  @Gaboose added a toe-hold implementation of many WASI features including a MemFS https://github.com/tetratelabs/wazero/pull/22 This was maintained over time, but we completely redid our API. Reads are now via fs.FS despite limitations of that.
* Recently, @nullpo-head attempted to make an extension of `fs.FS` in https://github.com/tetratelabs/wazero/pull/362 with some custom APIs to add write support to it. This was closed out because it coupled read-side with writes and the former is easier to implement. This issue was raised to carry that forward.
[...]

PR #394, opened 2022-03-18
wasi: replaces existing filesystem apis with fs.FS

codefromthecrypt:
This changes `SysConfig.WithFS` and `SysConfig.WithWorkDirFS` to accept an `fs.FS` for the root ("/") and working directory (".") file systems. This allows end users to leverage standard libraries that adhere to this, including OS, HTTP and embedded filesystems.

For example, here's how you can allow WebAssembly modules to read
"/work/home/a.txt" as "/a.txt" or "./a.txt":
```go
wasi, err := r.InstantiateModule(wazero.WASISnapshotPreview1())
defer wasi.Close()

sysConfig := wazero.NewSysConfig().WithFS(os.DirFS("/work/home"))
module, err := wazero.StartWASICommandWithConfig(r, compiled, sysConfig)
defer module.Close()
...
```

Note: Further development will happen after this, for example to allow creation of new directories, though priority of that is demand-based. If you need further FS features, you should ask on https://github.com/tetratelabs/wazero/issues/390 with your use-case!


examples/testdata/cat.go
```diff
@@ -0,0 +1,20 @@
...
+	"fmt"
+	"os"
+)
+
+// main is the same as cat.
+func main() {
+	for _, file := range os.Args {
+		bytes, err := os.ReadFile(file)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "couldn't read %s: %v", file, err)
```
codefromthecrypt, line 13: IMHO: generally we should do something like this instead of panic because panics from tinygo are infinitely difficult to troubleshoot.

codefromthecrypt:
@nullpo-head would you mind pushing a commit with `make build.examples` from linux? darwin output is massive

codefromthecrypt:
@nullpo-head would you mind pushing a commit with `make build.examples` from linux? darwin output is massive

internal/wasi/wasi.go
```diff
@@ -1383,87 +1377,69 @@ func (a *wasiAPI) PathLink(ctx wasm.Module, oldFd, oldFlags, oldPath, oldPathLen
...
-	// TODO: oflagDir
-	oflagExclusive
-	oflagTrunc
-
-	// WASI FS rights
-	rightFDRead  = 1 << iota
-	rightFDWrite = 0x200
-)
-
-func posixOpenFlags(oFlags uint32, fsRights uint64) (pFlags int) {
```
codefromthecrypt, line 1398: #390 re-introduce some aspects of this later. it would ignore fsRights as they aren't helpful or likely to be implemented in any consistent way. Rather, pay attention to mount config (static) oFlags and dirFlags See

wasi.go
```diff
@@ -4,33 +4,21 @@ import (
...
-	}
-}
-
 type WASIConfig struct {
-	stdin    io.Reader
-	stdout   io.Writer
-	stderr   io.Writer
-	args     []string
-	environ  map[string]string
-	preopens map[string]wasi.FS
```
codefromthecrypt, line 33: no reason to use a map because fs.FS can be composed by other libraries. plus we use things like path maps we end up needing to consider overlays. let fs.FS handle that

nullpo-head:
> @nullpo-head would you mind pushing a commit with `make build.examples` from linux? darwin output is massive

Do you mean files such as "cat.wasm", right? I tried but the size of the output was the same. I guess the target platform is wasm, so the output result of TinyGo doesn't change much on different platform?

codefromthecrypt:
> Do you mean files such as "cat.wasm", right? I tried but the size of the output was the same. I guess the target platform is wasm, so the output result of TinyGo doesn't change much on different platform?

OK yeah I think once it made a lot of difference. I rolled back the println on error as that increased the size by 140KB

codefromthecrypt:
lol there's no /tmp on scratch @mathetake  :) I'll try mounting a volume..

```
 --- FAIL: TestSnapshotPreview1_FdWrite (0.00s)
    --- FAIL: TestSnapshotPreview1_FdWrite/SnapshotPreview1.FdWrite (0.00s)
        wasi_test.go:2180: TempDir: stat /tmp: no such file or directory
    --- FAIL: TestSnapshotPreview1_FdWrite/fd_write (0.00s)
        wasi_test.go:2180: TempDir: stat /tmp: no such file or directory
--- FAIL: TestSnapshotPreview1_FdWrite_Errors (0.00s)
    wasi_test.go:2180: TempDir: stat /tmp: no such file or directory
```

examples/testdata/cat.go
codefromthecrypt: switched to just error code which is just as well as panic when there is only one function (_start). Removed the printf because it added 140KB to the wasm

codefromthecrypt:
I'll look into the windows stuff tomorrow. I remember some of this ordering issue in func-e tests, just not exactly how we unraveled it.

```
--- FAIL: TestSnapshotPreview1_FdWrite (0.00s)
    --- FAIL: TestSnapshotPreview1_FdWrite/SnapshotPreview1.FdWrite (0.00s)
        testing.go:967: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestSnapshotPreview1_FdWriteSnapshotPreview1.FdWrite603158801\001\arbitrary: The process cannot access the file because it is being used by another process.
    --- FAIL: TestSnapshotPreview1_FdWrite/fd_write (0.00s)
        testing.go:967: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestSnapshotPreview1_FdWritefd_write2424343077\001\arbitrary: The process cannot access the file because it is being used by another process.
--- FAIL: TestSnapshotPreview1_FdWrite_Errors (0.00s)
    testing.go:967: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestSnapshotPreview1_FdWrite_Errors4036609832\001\arbitrary: The process cannot access the file because it is being used by another process.
```

codefromthecrypt:
probably need a t.Cleanup() when opening test file descriptors as not all tests close them. This also hints maybe we should make mod.Close() actually close file descriptors when WASI

codefromthecrypt:
actually I will redo this in the next couple days because how opened file descriptors are kept is wrong. they should always be owned by the importing module.

codefromthecrypt:
So there are a couple next steps before merge. I'm going to do these in the same PR so that the most important features are retained, and when there's a new API it is safe. That windows tests are failing are a good incentive to fix this now, and we couldn't fix the safety issues before as we were relying on non-standard types (memfs)

1. Add a feature for PWD (ex allowing the directory named "." to resolve).
  * The two most common "preopen" on GitHub are working directory "." and "/"
  * Other subpaths can be achieved by composing a rootFS or working dir
2. Now that we are using standard types, move file descriptors to internalwasm.Module.
  * Right now, there are scope problems as FDs are shared across modules. this doesn't make sense in sandboxing.
  * Trying to manage them with a scoped object isn't correct as WASI can be imported twice
  * Moveover we may have a different WASI impl which means a different memory.
  * By moving FDs to internalwasm.Module, they are like memory or anything else and can be cleaned up coherently.
3. Add a toehold test for sanity checking file system works
  * Even though we have unit tests on what memory is written

codefromthecrypt:
FYI I tried to retrofit memFS to fs.FS timeboxed to a couple hours, but it is more messy than removing it. I'm not spending any more time on memFS

codefromthecrypt:
ok this is rebased. next steps before ready for review:

1. fix whatever's going on in the write tests in windows
2. Test_Cat is the only thing actually using this. Add tests to ensure working directory and root directory paths work
3. scrub for TODOs and fix here, leave them or open a new issue.

codefromthecrypt:
opened https://github.com/tetratelabs/wazero/issues/408 to follow-up at some point and add another kind of integration test, now that it is easy to compare against an `fs.FS`

config.go
```diff
@@ -216,21 +216,42 @@ func (c *SysConfig) WithEnv(key, value string) *SysConfig {
...
+//
+// Ex. This sets a read-only, embedded file-system as the root ("/"), and a mutable one as the working directory ("."):
+//
+//	//go:embed appA
+//	var rootFS embed.FS
+//
+//	// Files relative to this source under appA is available under "/" and files relative to "/work/appA" under ".".
+//	sysConfig := wazero.NewSysConfig().WithFS(rootFS).WithWorkDirFS(os.DirFS("/work/appA"))
+//
+// Note: os.DirFS documentation includes important notes about isolation, which also applies to fs.Sub. As of Go 1.18,
```
codefromthecrypt, line 246: took some reading to discover there are no plans for an actual chrooted FS. Until then, folks with high isolation needs will have to write their own fs.FS impl (or find some library that isn't abandoned that tries to do this!)

codefromthecrypt:
cc @pims this is an API break, but a good one as we've gotten rid of an incompatible FS and formalized the API. Hope this doesn't drift you too much!

mathetake (approved):

pims:
Thanks for the heads-up @codefromthecrypt 

This looks like a very sensible change and until 1.0 is released, no stress about API breaking changes.

PR #395, opened 2022-03-19
Makes filesystem test simpler and removes use of panic in examples

codefromthecrypt:
This reduces size of #394 so that when finished, it is easier to review.


examples/testdata/stdio.go
```diff
@@ -12,11 +12,11 @@ func main() {
 	for s.Scan() {
 		line := s.Text()
 		if _, err := fmt.Printf("Hello, %s!\n", strings.TrimSpace(line)); err != nil {
-			panic(err)
+			os.Exit(1)
```
codefromthecrypt, line 15: IIRC when we wrote these examples, we didn't implement exit, yet

codefromthecrypt:
I will raise a PR to document the WASIConfig because I think it is easy to forget arg[0] is the program name. In this case, I think it probably makes sense to default it to the WASM module name, but we can discuss on that PR.

mathetake (approved):

PR #396, opened 2022-03-19
wasi: renames WASIConfig to SysConfig and makes stdio defaults safer

codefromthecrypt:
This introduces `SysConfig` to replace `WASIConfig` and formalize documentation around system calls.

The only incompatible change planned after this is to switch from wasi.FS to fs.FS in #396

Implementation Notes:

Defaulting to os.Stdin os.Stdout and os.Stderr doesn't make sense for
the same reasons as why we don't propagate ENV or ARGV: it violates
sand-boxing. Moreover, these are worse as they prevent concurrency and
can also lead to console overload if accidentally not overridden.

This also changes default stdin to read EOF as that is safer than reading
from os.DevNull, which can run the host out of file descriptors.

Finally, this removes "WithPreopens" for "WithFS" and "WithWorkDirFS",
to focus on the intended result. Similar Docker, if the WorkDir isn't set, it
defaults to the same as root.


internal/wasi/wasi.go
```diff
@@ -1231,7 +1231,11 @@ func (a *wasiAPI) FdRead(ctx wasm.Module, fd, iovs, iovsCount, resultSize uint32
 
 	switch fd {
 	case 0:
-		reader = cfg.stdin
+		if cfg.stdin != nil {
+			reader = cfg.stdin
+		} else {
+			return wasi.ErrnoBadf
```
mathetake, line 1237: I would prefer having Nil Reader rather than returning Badf as maybe all programs would expect `0` is valid FD in anycase

mathetake (approved):

internal/wasi/wasi.go
codefromthecrypt: ok just we have to close the reader, so I think I'll make a TODO as the code is not organized well enough now to close resources.

codefromthecrypt: unless by Nil reader you mean something besides os.DevNull...

mathetake: maybe passing `bytes.NewReader(nil)` is one idea :D

codefromthecrypt: remember this is for multiple modules (when many import one wasi), hence the close resources problem... and also if the program really needs to read stdin shouldn't it know somehow to override?

one way out is to TODO for a few days until there's proper resource management of file descriptors.

mathetake: Oh we need closer..

mathetake: yep up to you!

codefromthecrypt: the general problem in the current code is that file descriptors are wildly scoped, which will be solved at or after #394 as file descriptors should never be runtime scoped (ex they should be like memory, owned by the importing module)

codefromthecrypt: I'll make a code comment TODO as I've tried and it is too hard to get resources to close until we use standard types. If someone complains in any case they should be setting stdin if they actually need it, and badf error is not a horrible way to inform that for a few days.

thx for the nag it is a real issue I agree.

mathetake: sg!

codefromthecrypt: I have a better idea.. just took a 🚿 to think of it. iostreams aren't tightly coupled with memfs in the current code, so I can move them to module ahead of the preopens stuff. That has the advantage of lowering future scoped work, too. So, I'll update this in a bit!

codefromthecrypt:
getting closer to this, just have to re-align the planet a bit :D

internal/wasi/wasi.go
codefromthecrypt: I found out that reading from os.DevNull was a very bad idea as it can run the host out of file descriptors easily if you forget to close. Reading EOF is a better default as if some code fails due to this, they should override stdin anyway

codefromthecrypt:
I owe some cleanup and tests, but this should be ready tomorrow morning asia time.

codefromthecrypt:
ok finished backfilling tests and also I made as many changes possible to support #396 here, notably removing WithPreopens for WithFS and WithWorkDirFS. This will hopefully make the next PR more mechanical to implement.

wasm.go
```diff
@@ -146,5 +144,9 @@ func (r *runtime) InstantiateModuleFromSource(source []byte) (wasm.Module, error
 
 // InstantiateModule implements Runtime.InstantiateModule
 func (r *runtime) InstantiateModule(module *Module) (wasm.Module, error) {
-	return r.store.Instantiate(r.ctx, module.module, module.name)
+	if sys, err := (&SysConfig{}).toSysContext(); err != nil {
```
mathetake, line 147: what's this?

internal/wasm/sys.go
```diff
@@ -20,102 +20,206 @@ type FileEntry struct {
...
+}
+
+// Stderr is like exec.Cmd Stderr and defaults to io.Discard.
+// See wazero.SysConfig WithStderr
+func (c *SysContext) Stderr() io.Writer {
+	return c.stderr
+}
+
+// EOFReader is safer than reading from os.DevNull as it can never overrun operating system file descriptors.
+type EOFReader struct{}
```
mathetake, line 101: does this need to be exported?

mathetake (approved):

internal/wasm/sys.go
codefromthecrypt: nope!

wasm.go
codefromthecrypt: goofy is what it is. I cleaned it up.

PR #397, opened 2022-03-19
Hide WASI system call lists by default in README

mathetake:


codefromthecrypt (approved):
I clicked rich view and was able to see that the collapse thingy works!

PR #399, opened 2022-03-20
wasi: moves test-only overrides out of config, simplifies error tests

codefromthecrypt:
This moves test-only overrides out of config in preparation of
normalized configuration. This also simplifies error tests by using the
API directly for all cases except exit.


mathetake (approved):

codefromthecrypt:
thanks for the 👁️ @mathetake

PR #401, opened 2022-03-21
wasi: moves system resource management to the importing module

codefromthecrypt:
The prior design had a problem, where multiple imports of WASI would end
up having different file descriptors for the same file. Moreover, there
was no means to close any of these when a module importing WASI was
closed.

This moves all existing functionality to a new type SystemContext, which
is owned by the importing module, similar to how it owns its memory.
While this PR doesn't fix some problems such as unclosed files, the code
is now organized in a way it can be, and these issues will be resolved
by #394.

In order to fix scope, `WASISnapshotPreview1WithConfig` had to be
removed.

Signed-off-by: Adrian Cole <adrian@tetrate.io>


wasi.go
```diff
@@ -69,33 +68,31 @@ func (c *WASIConfig) WithPreopens(preopens map[string]wasi.FS) *WASIConfig {
...
 	m, err := internalwasm.NewHostModule(wasi.ModuleSnapshotPreview1, fns)
 	if err != nil {
 		panic(fmt.Errorf("BUG: %w", err))
 	}
 	return &Module{name: wasi.ModuleSnapshotPreview1, module: m}
 }
 
-func newConfig(c *WASIConfig) *internalwasi.Config {
-	cfg := internalwasi.NewConfig()
+func newSystemContext(c *WASIConfig) (*internalwasm.SystemContext, error) {
```
codefromthecrypt, line 79: Specifically #394 will rename the WASIConfig type to ModuleConfig or SystemConfig and pass that to instantiate. This will prevent having to swap references later, and also ensure error handling happens before any instantiation might occur.

codefromthecrypt: I'm waiting to do that as it is a bigger break than this change, and we might as well fix wasi.FS -> fs.FS while in the same change.

mathetake (approved):

codefromthecrypt:
thanks for the look!

PR #403, opened 2022-03-22
asm: add skeleton amd64 assembler 

mathetake:
This adds the skeleton implementation of amd64 assembler.
The assembler is not able to produce binary at all, and the actual
encoding implementations follow in subsequent commits.

Part of #233


internal/wasm/jit/asm/amd64/assembler.go
```diff
@@ -10,35 +10,35 @@ func NewAssembler() (Assembler, error) {
 
 type Assembler interface {
 	asm.AssemblerBase
-	// CompileConstModeRegisterToRegister adds an instruction where source and destination
+	// CompileRegisterToRegisterWithMode adds an instruction where source and destination
 	// are `from` and `to` registers and the instruction's "mode" is specified by `mode`.
 	// For example, ROUND** instructions can be modified "mode" constant.
 	// See https://www.felixcloutier.com/x86/roundss for ROUNDSS as an example.
-	CompileConstModeRegisterToRegister(inst asm.Instruction, from, to asm.Register, mode int64)
+	CompileRegisterToRegisterWithMode(instruction asm.Instruction, from, to asm.Register, mode int64)
```
codefromthecrypt, line 17: should mode have a type alias to define modes?

internal/wasm/jit/asm/amd64/consts.go
```diff
@@ -2,7 +2,7 @@ package amd64
 
 import "github.com/tetratelabs/wazero/internal/wasm/jit/asm"
 
-// Arm64-specific register states.
+// AMD64-specific register states.
```
codefromthecrypt, line 5: add link to where someone can verify these (as well the instructions?)

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,382 @@
...
+
+	"github.com/tetratelabs/wazero/internal/wasm/jit/asm"
+)
+
+// nodeImpl implements asm.Node for amd64.
+type nodeImpl struct {
+	instruction asm.Instruction
+
+	// offsetInBinary represents the offset of this node in the final binary.
+	offsetInBinary int64
```
codefromthecrypt, line 17: add comment why the offset is signed or make it unsigned? also are we expecting a binary this size? if not, say why 64bit

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,382 @@
...
+	// jumpTarget holds the target node in the linked for the jump-kind instruction.
+	jumpTarget *nodeImpl
+	// prev holds the previous node to this node in the assembled linked list.
+	prev *nodeImpl
+	// next holds the next node from this node in the assembled linked list.
+	next *nodeImpl
+
+	types                    operandTypes
+	srcReg, dstReg           asm.Register
+	srcConst, dstConst       int64
```
codefromthecrypt, line 27: similar comment needed saying why 64 and why signed. actually this is everywhere so you can make a type alias to describe and use that to reduce comments. here and where they are used in parameters.

ex type sourcePosition = int64 (or something, then comment once)

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,382 @@
...
+func (n *nodeImpl) String() (ret string) {
+	instName := instructionName(n.instruction)
+	switch n.types {
+	case operandTypesNoneToNone:
+		ret = instName
+	case operandTypesNoneToRegister:
+		ret = fmt.Sprintf("%s , %s", instName, registerName(n.dstReg))
+	case operandTypesNoneToMemory:
+		if n.dstMemIndex != asm.NilRegister {
+			ret = fmt.Sprintf("%s , [%s + 0x%x + %s*0x%x]", instName,
```
codefromthecrypt, line 64: is space around comma intended? if so a-ok

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,382 @@
...
+func (n *nodeImpl) AssignSourceConstant(value int64) {
+	n.srcConst = value
+}
+
+// OffsetInBinary implements asm.Node.OffsetInBinary.
+func (n *nodeImpl) OffsetInBinary() int64 {
+	return n.offsetInBinary
+}
+
+// String implements fmt.Stringer.
```
codefromthecrypt, line 54: add a comment below this to say what format we are expected, maybe a simple example and a link

codefromthecrypt (commented):
unleashing some comments!

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,401 @@
...
+		} else {
+			ret = fmt.Sprintf("%s 0x%x, [%s + 0x%x]", instName, n.srcConst, registerName(n.dstReg), n.dstConst)
+		}
+	case operandTypesConstToRegister:
+		ret = fmt.Sprintf("%s 0x%x, %s", instName, n.srcConst, registerName(n.dstReg))
+	}
+	return
+}
+
+type operandType byte
```
codefromthecrypt, line 110: comment where this is from, if from somewhere (ex if we defined this ourselves for convenience there's no See link needed!)

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,401 @@
...
+		ret = "memory"
+	case operandTypeConst:
+		ret = "const"
+	case operandTypeBranch:
+		ret = "branch"
+	}
+	return
+}
+
+type operandTypes struct{ src, dst operandType }
```
codefromthecrypt, line 136: something like "operandTypes include the only combinations used by wazero" (unless they are the only combinations valid for the arch, then say that instead)

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,401 @@
...
+	operandTypesRegisterToRegister = operandTypes{operandTypeRegister, operandTypeRegister}
+	operandTypesRegisterToMemory   = operandTypes{operandTypeRegister, operandTypeMemory}
+	operandTypesRegisterToConst    = operandTypes{operandTypeRegister, operandTypeConst}
+	operandTypesMemoryToRegister   = operandTypes{operandTypeMemory, operandTypeRegister}
+	operandTypesMemoryToConst      = operandTypes{operandTypeMemory, operandTypeConst}
+	operandTypesConstToRegister    = operandTypes{operandTypeConst, operandTypeRegister}
+	operandTypesConstToMemory      = operandTypes{operandTypeConst, operandTypeMemory}
+)
+
+func (o operandTypes) String() string {
```
codefromthecrypt, line 153:
```suggestion
// String implements io.Stringer
func (o operandTypes) String() string {
```

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,401 @@
...
+		a.root = node
+		a.current = node
+	} else {
+		parent := a.current
+		parent.next = node
+		a.current = node
+	}
+}
+
+func (a *assemblerImpl) encodeNode(w io.Writer, n *nodeImpl) (err error) {
```
codefromthecrypt, line 190: as node functions sometimes traverse, maybe mention that this only encodes the current node, so that you don't have to click in to figure that out. unless it will traverse :D

internal/wasm/jit/asm/amd64/impl.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// CompileNoneToMemory implements assembler.CompileNoneToMemory
+func (a *assemblerImpl) CompileNoneToMemory(instruction asm.Instruction, baseReg asm.Register, offset int64) {
+	n := a.newNode(instruction, operandTypesNoneToMemory)
+	n.dstReg = baseReg
+	n.dstConst = offset
+}
+
+func (a *assemblerImpl) CompileConstToMemory(instruction asm.Instruction, value int64, dstbaseReg asm.Register, dstOffset int64) asm.Node {
```
codefromthecrypt: maybe this and below are missing the docs that say it implements the interface, or if it doesn't document them here 🙏

internal/wasm/jit/asm/impl.go
```diff
@@ -0,0 +1,47 @@
+package asm
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math"
+)
+
+type BaseAssemblerImpl struct {
```
codefromthecrypt, line 9:
```suggestion
// BaseAssemblerImpl includes code common to all architectures.
//
// Note: When possible, add code here instead of in architecture-specific files to reduce drift:
// As this is internal, exporting symbols only to reduce duplication is ok.
type BaseAssemblerImpl struct {
```

internal/wasm/jit/asm/impl.go
```diff
@@ -0,0 +1,47 @@
...
+	a.OnGenerateCallbacks = append(a.OnGenerateCallbacks, cb)
+}
+
+// BuildJumpTable implements AssemblerBase.BuildJumpTable
+func (a *BaseAssemblerImpl) BuildJumpTable(table []byte, labelInitialInstructions []Node) {
+	a.AddOnGenerateCallBack(func(code []byte) error {
+		// Build the offset table for each target.
+		base := labelInitialInstructions[0].OffsetInBinary()
+		for i, nop := range labelInitialInstructions {
+			if uint64(nop.OffsetInBinary())-uint64(base) >= math.MaxUint32 {
```
codefromthecrypt, line 33: can we define a constant in like internal/limits.go and move this comment there?

codefromthecrypt (approved):
Looking nice!

PR #404, opened 2022-03-22
ci: fix test binary name duplicates

mathetake:
Previously, the prebuilt binaries have been named as `${last part of import path}.test`
and placed on the project root. Therefore, the test binary for `./wasm/` has the same 
name for `internal/wasm/` and resulted in running only one of these two tests.
This fixes that by producing the binary on each pkg's parent directory.


nullpo-head (approved):
lgtm, thanks!

PR #405, opened 2022-03-23
Make ProcExit call Module.Close()

nullpo-head:
This PR makes ProcExit call `Module.Close()` when it's called. [`proc_exit` is expected to make the caller no longer usable]( https://github.com/WebAssembly/WASI/issues/26#issuecomment-1051176870), which means wazero can call `Module.Close()` and can release its resources. There may be further discussion that we can close the entire store, but closing module would be the good starting point anyway.

### DIFF
--- a/internal/wasi/wasi_test.go
+++ b/internal/wasi/wasi_test.go
@@ -38,7 +38,7 @@ func TestSnapshotPreview1_ArgsGet(t *testing.T) {
 		'?', // stopped after encoding
 	}
 
-	a, mod, fn := instantiateModule(t, ctx, FunctionArgsGet, ImportArgsGet, moduleName, sys)
+	a, _, mod, fn := instantiateModule(t, ctx, FunctionArgsGet, ImportArgsGet, moduleName, sys)
 	defer mod.Close()
 
 	t.Run("SnapshotPreview1.ArgsGet", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/wasi/wasi_test.go
```diff
@@ -1874,8 +1874,76 @@ func TestSnapshotPreview1_PollOneoff(t *testing.T) {
...
+	procExitFns := []struct {
+		name     string
+		procExit func() (procExitFn, *wasm.Store)
+	}{
+		// call proc_exit in a usual way, that is, call the imported-and-exported-back ProcExit.
+		{FunctionProcExit, func() (procExitFn, *wasm.Store) {
+			sys, err := newSysContext(nil, nil, nil)
+			require.NoError(t, err)
+
+			_, store, mod, fn := instantiateModule(t, context.Background(), FunctionProcExit, ImportProcExit, moduleName, sys)
```
codefromthecrypt, line 1891: AFAICT, the signature change to return a 4th result is only used here. can you revert it and instantiate this test case different way? That prevents drifting almost all tests in this file.

codefromthecrypt (commented):
So, I would roll back changes to wasi_test.go.

There are two sorts of tests going on. wasi_test is unit tests and shouldn't do more than that. Anything more heavy should be tested in ad_hoc test so that multiple engines are used. `testAdhocCloseWhileExecution` does almost exactly what you were looking for and so heavier tests should be slight modifications to that.

For unit tests, there is little to do here, so should be a lot less code change. There's only one thing new going on. `module.Close()` is being invoked. You only need to test this was called, and that you are ignoring its errors (since the code added ignores errors).

So, for a test like that you need two instance of `wasm.Module`: one that succeeds on close and one that fails on close. You also need to be able to detect if it has closed.

You can instantiate the Module any way you like, including literally instantiating it. This means you can avoid heavy changes to the file to test these two cases.

Remember, WASI is only used when importing, so the importing module is the context used, you control which module instance is used when you use `fn.Call`

Hope these points help!

nullpo-head:
Your points totally make sense and I'm working on fixing. This PR is not closed but will be completely rebased after it.

nullpo-head:
Closing in favor of the different changes

PR #406, opened 2022-03-23
asm: complete amd64 instruction encodings

mathetake:
This completes the implementation of our homemade assembler for amd64.
The implementation is _almost_ 100% compatible with the Go's official assembler
except that NOP padding strategy for [mitigation for Intel jump erratum](https://www.intel.com/content/dam/support/us/en/documents/processors/mitigations-jump-conditional-code-erratum.pdf) differs, 
which doesn't affect the semantics of produced machine code.

Note that our assembler here is not complete enough for public usage; The impl
has support only for instructions used by wazero, and the API is heavily coupled
with our JIT compiler internal implementation detail. Therefore, I don't have any
intention to make this public as a replacement for golang-asm.

This is the partial fix for #233 on amd64 target, and arm64 follows later.


mathetake:
ok 60 % 😄

mathetake:
97% done :D

mathetake:
ok the left is only to implement NOP padding for macro fused jump instructions..https://en.wikichip.org/wiki/macro-operation_fusion#Intel

mathetake:
ps the current test coverage is around 94.5 % :D

mathetake:
LOL I totally forgot the macos only instructions :D

mathetake:
anyway passed all tests!

mathetake:
ok finally ready!

internal/wasm/jit/asm/amd64/golang_asm.go
```diff
@@ -16,7 +16,7 @@ type assemblerGoAsmImpl struct {
 }
 
 func newGolangAsmAssembler() (*assemblerGoAsmImpl, error) {
```
codefromthecrypt, line 18: can we move internal/wasm/jit/asm -> internal/jit. I don't think it relies on anything in wasm or jit

Also, can we make the golang/twitchy stuff dependent on debug flag or arm64 to try to reduce the deps?
ex.

```
//go:build debug_mode || arm64
```

codefromthecrypt: I will try this

codefromthecrypt: it will be too much work to unravel this. I'll put a TODO until arm64 is finished. Right now, we are sharing logic and having it compile on both architectures, which is really helpful when completing something as it controls drift. We can later schedule to decouple better, as the decoupling will make drift hard to manage

codefromthecrypt:
going to try one more time to get twitchy out of amd64. I think it is because the code is referenced in the same package:
```
# github.com/twitchyliquid64/golang-asm
test
github.com/tetratelabs/wazero
github.com/tetratelabs/wazero/internal/wasm/jit
github.com/tetratelabs/wazero/internal/asm/amd64
github.com/tetratelabs/wazero/internal/asm/golang_asm
github.com/twitchyliquid64/golang-asm
```

codefromthecrypt:
good progress! only thing left is to exclude arm when building amd64

```
○ → go mod why  github.com/twitchyliquid64/golang-asm
# github.com/twitchyliquid64/golang-asm
test
github.com/tetratelabs/wazero
github.com/tetratelabs/wazero/internal/wasm/jit
github.com/tetratelabs/wazero/internal/asm/arm64
github.com/tetratelabs/wazero/internal/asm/golang_asm
github.com/twitchyliquid64/golang-asm
```

codefromthecrypt:
OK! So right now, arm64 is bringing in the twitchy dep because that's not been redone yet. However, I've checked and that dep is only due to the package scope of the dependency tree. I've made an end-to-end example and provably things are smaller now, and by a LOT!

Before
```
IMAGE          CREATED         CREATED BY                                      SIZE      COMMENT
38ae923f0378   7 seconds ago   /bin/sh -c #(nop)  ENTRYPOINT ["/wazero"]       0B        
945bf18902b6   8 seconds ago   /bin/sh -c #(nop) COPY file:7f9737dbb254898f…   5.18MB    
```

Now!!
```
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
41b4632dd4dc   10 seconds ago   /bin/sh -c #(nop)  ENTRYPOINT ["/wazero"]       0B        
3b386efaad77   10 seconds ago   /bin/sh -c #(nop) COPY file:869939117b01cf7d…   3.23MB    
```

codefromthecrypt:
Using a reference of `fmt.Println(1+2)`, this means wazero's image overhead inclusive of a wasm compiler and runtime is 1.47MB

```
○ → docker history wazero:latest
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
badbad20823d   12 seconds ago   /bin/sh -c #(nop)  ENTRYPOINT ["/wazero"]       0B        
b7998d3307da   12 seconds ago   /bin/sh -c #(nop) COPY file:ceeeafa23c4af7cc…   1.76MB    
```

PR #407, opened 2022-03-23
Adds tests to cover ModuleContext.Close

codefromthecrypt:
This backfills tests that ModuleContext.Close ends up calling
`Store.CloseModule` and `SysContext.Close`.

### DIFF
--- a/internal/wasm/module_context_test.go
+++ b/internal/wasm/module_context_test.go
@@ -1,0 +1,199 @@
+package internalwasm
+
+import (
+	"context"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestModuleContext_WithContext(t *testing.T) {
+	type key string
+	tests := []struct {
+		name       string
+		mod        *ModuleContext
+		ctx        context.Context
+		expectSame bool
+	}{
+		{
+			name:       "nil->nil: same",
+			mod:        &ModuleContext{},
+			ctx:        nil,
+			expectSame: true,
+		},
+		{
+			name:       "nil->ctx: not same",
+			mod:        &ModuleContext{},
+			ctx:        context.WithValue(context.Background(), key("a"), "b"),
+			expectSame: false,
+		},
+		{
+			name:       "ctx->nil: same",
+			mod:        &ModuleContext{ctx: context.Background()},
+			ctx:        nil,
+			expectSame: true,
+		},
+		{
+			name:       "ctx1->ctx2: not same",
+			mod:        &ModuleContext{ctx: context.Background()},
+			ctx:        context.WithValue(context.Background(), key("a"), "b"),
+			expectSame: false,
+		},
+	}
+
+	for _, tt := range tests {
+		tc := tt
+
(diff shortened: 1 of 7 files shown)


internal/wasm/module_context_test.go
```diff
@@ -0,0 +1,199 @@
...
+	s := newStore()
+
+	t.Run("calls store.CloseModule(module.name)", func(t *testing.T) {
+		moduleName := t.Name()
+		m, err := s.Instantiate(context.Background(), &Module{}, moduleName, nil)
+		require.NoError(t, err)
+
+		// We use side effects to determine if Close in fact called store.CloseModule (without repeating store_test.go).
+		// One side effect of store.CloseModule is that the moduleName can no longer be looked up. Verify our base case.
+		require.Equal(t, s.Module(moduleName), m)
```
codefromthecrypt, line 151: @nullpo-head here's an example of using other tests to build higher level signals that something happened.

For example, through this and other tests, we know that `Module.Close()` ends up resulting in `runtime.Module(moduleName) == nil` (provided you check the base case). This can lead to simpler tests when we are trying to see if a function that calls close executed. e.g. try to lookup the module it was called from.

internal/wasm/module_test.go
```diff
@@ -76,7 +76,7 @@ func TestExternTypeName(t *testing.T) {
 	}{
 		{"func", ExternTypeFunc, "func"},
 		{"table", ExternTypeTable, "table"},
-		{"mem", ExternTypeMemory, "memory"},
+		{"ctx", ExternTypeMemory, "memory"},
```
codefromthecrypt, line 79:
```suggestion
		{"mem", ExternTypeMemory, "memory"},
```

internal/wasm/sys_test.go
```diff
@@ -147,3 +150,57 @@ func TestNewSysContext_Environ(t *testing.T) {
...
+			},
+		)
+		require.NoError(t, err)
+
+		// Closing should delete the file descriptors after closing the files.
+		require.NoError(t, sys.Close())
+		require.Empty(t, sys.openedFiles)
+
+		// Verify it was actually closed, by trying to close it again.
+		err = file.Close()
```
codefromthecrypt, line 184: another side-effect you can use. Ex. if you can't easily reach the store or the runtime.. if you can open a file, you can check that it closed, by trying to close it again!

internal/wasm/sys_test.go
```diff
@@ -147,3 +150,57 @@ func TestNewSysContext_Environ(t *testing.T) {
...
+		require.Contains(t, err.Error(), "file already closed")
+
+		// No problem closing config again because the descriptors were removed, so they won't be called again.
+		require.NoError(t, sys.Close())
+	})
+
+	// TODO: fs but never used (ex file == nil)
```
codefromthecrypt, line 191: I won't be addressing these TODOs in this PR. This is just to unlock the other one. I'll do it in #394

mathetake (approved):

codefromthecrypt:
thanks for the look!

PR #409, opened 2022-03-24
wasi: stops enforcing _start function export

codefromthecrypt:
Currently, we have custom code in wapc-go because our library forces a
failure when a module that uses WASI doesn't define a "_start" function.
Using the same pragmatism that resulted in us not enforcing the WASI
table, this makes the "_start" function optional. This doesn't add a
flag as the spec is not a proper version anyway (snapshot-01), so
there's no need to further complicate configuration.

If a "_start" function exists, we enforce it is of the proper signature
and succeeds. Otherwise, we allow it to be absent.

### DIFF
--- a/wasi.go
+++ b/wasi.go
@@ -40,9 +40,9 @@ func StartWASICommandFromSource(r Runtime, source []byte) (wasm.Module, error) {
 	}
 }
 
-// StartWASICommand instantiates the module and starts its WASI Command function ("_start"). The return value are all
-// exported functions in the module. This errs if the module doesn't comply with prerequisites, or any instantiation
-// or function call error.
+// StartWASICommand instantiates the module and starts its WASI Command function ("_start") if present. The return value
+// are all exported functions in the module. This errs if the module doesn't export a memory named "memory", or there
+// are any instantiation or function call errors. On success, other modules can import wasi.ModuleSnapshotPreview1.
 //
 // Ex.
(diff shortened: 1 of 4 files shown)


wasi.go
```diff
@@ -53,13 +53,17 @@ func StartWASICommandFromSource(r Runtime, source []byte) (wasm.Module, error) {
...
 //
-// Prerequisites of the "Current Unstable ABI" from wasi.ModuleSnapshotPreview1:
-// * "_start" is an exported nullary function and does not export "_initialize"
-// * "memory" is an exported memory.
+// ## "memory" export
+// WASI snapshot-01 requires exporting a memory named "memory", and wazero enforces this as nearly all functions use
+// memory to implement multiple returns. StartWASICommand errs if there is no memory exported as "memory".
+//
+// ## "_start" function export
+// WASI snapshot-01 requires exporting a function named "_start", but wazero does not enforce this. If it is defined,
```
mathetake, line 61: nit but important as some embedders will use `_initialize` of WASI reactors e.g. Envoy (though we don't have `InitializeWASIReactor` yet :D)

```suggestion
// WASI snapshot-01 requires exporting a function named "_start" for WASI command, but wazero does not enforce this. If it is defined,
```

mathetake (approved):

wasi.go
codefromthecrypt: great point!

PR #410, opened 2022-03-25
wasi: exposes CloseWithExitCode to correct implementation of proc_exit

codefromthecrypt:
Before, when the WASI function "proc_exit" was invoked, not only did the
module stay alive, but also only the calling goroutine would receive an
error with the exit code.

```go
panic(wasi.ExitCode(exitCode))
```

Now, this corrects the implementation to actually shutdown the module,
and ensure any callers can see the exit code, not just the caller who
invoked exit.

```go
_ = m.CloseWithExitCode(exitCode)
```

The changes needed for this were numerous, but allow other
implementations, such as AssemblyScript abort handlers, to close the
same way:
* Adds `Module.CloseWithExitCode` which allows propagating an exit code
  to any callers of exported functions.
* Dispatches `Module.Close` to `Module.CloseWithExitCode(0)`
* Replace `wasi.ExitCode` with `sys.ExitError` which is returned to any
  in-flight callers. This formalizes the concept regardless of WASI.
* Ensures all "close-once" guards are tested on engines.
* Reuses the numeric field used for the closed flag to store the exit_code atomically.
* Weave in exit errors into existing test cases to reduce code heft.

This also folds the partially maintained WASI RATIONALE into the
top-level doc.


codefromthecrypt:
I can't tell if the CI fail is a race condition or a field visibility problem. I'll switch the impl to use atomic just in case

codefromthecrypt:
ok switched impl to pack exit code into high-order bits and read them back atomically. Anything left I think is a legit race I think and I'll update tests as needed if they fail again.

internal/wasm/interpreter/interpreter.go
```diff
@@ -507,6 +515,12 @@ func (me *moduleEngine) Call(ctx *wasm.ModuleContext, f *wasm.FunctionInstance,
 
 	ce := me.newCallEngine()
 	defer func() {
+		// Default to an exit error if another function closed the module, but allow a better error to override it.
+		if err == nil {
+			if closed := atomic.LoadUint64(&me.closed); closed > 0 {
```
codefromthecrypt, line 520: @mathetake pretty sure this is correct, just double-check me if you don't mind. I was hoping i could just read me.closed directly, but did like this just in case there's some visibility problem otherwise.

regardless, I think the tests still race and there's a chance things pass this barrier in concurrent scenarios. usually in other tests, I put some sort of ordering condition to make sure something happens after (usually a blocking queue in the tests). I'll sort out the race in tests later, mainly I'm curious your opinion of the impl itself!

codefromthecrypt: ps I will be refactoring the lock to be external as there's too much similar between here and jit and easier to test externally

codefromthecrypt:
ok externalized the lock. One interesting thing is since we are using loaduint64 in go, we probably could in JIT if we wanted JIT to figure out if it should close or not.

Meanwhile, I need to backfill some concurrency tests before marking out of draft: tomorrow or sunday.

internal/wasm/interpreter/interpreter.go
```diff
@@ -534,6 +527,7 @@ func (me *moduleEngine) Call(ctx *wasm.ModuleContext, f *wasm.FunctionInstance,
 				err = fmt.Errorf("%w\nwasm backtrace:\n%s", err, strings.Join(traces, "\n"))
 			}
 		}
+
```
codefromthecrypt, line 530:
```suggestion
```

codefromthecrypt:
looking into the flakiness now. It is contained by singleModuleHighConcurrency test and fails because sometimes no error ends up returned. I'll look to see if I enforce this properly.

codefromthecrypt:
at first glance there's more race going on than needed to prove the concept. "delay" while launching 1000 goroutines is very racey. I don't think we are necessarily trying to test race, rather contention when a many goroutines are outstanding while a module is closed. By changing this to a synchronous barrier, we can be far more precise in tests because we can count exactly how many should see an error and fail if that didn't happen have to make guesses due to inherent race.

I'll do that now.

internal/wasm/interpreter/interpreter.go
codefromthecrypt: this is correct. I did research and added to RATIONALE.md in https://github.com/tetratelabs/wazero/pull/411 as it was hard to find a concrete answer (TL;DR; it isn't concretely answered, but fairly understood)

codefromthecrypt:
I'll bump this again after https://github.com/tetratelabs/wazero/pull/411 as the new tests will be easier to troubleshoot if failed.

codefromthecrypt:
OK I owe new concurrency tests on internalwasm.Store that prove regardless of engine we both close only once, but also new calls to a closed module fail early. Otherwise, I think this is ready for review because the test flakes are gone.

internal/wasm/jit/engine.go
```diff
@@ -38,10 +37,8 @@ type (
 		parentEngine           *engine
 		importedFunctionCounts int
 
-		// Closed indicates whether Close was called for this moduleEngine, and used to prevent
-		// double-free of compiledFunctions.
-		// Note: intentionally use uint32 instead of bool for atomic operation.
-		closed uint32
+		// closed is the field updated atomically that guards Close
+		closed uint64
```
codefromthecrypt, line 41: I'm secretly hoping we can load64 this in assembly to quit wasm early when non-zero.

internal/wasm/jit/engine_test.go
```diff
@@ -454,20 +445,8 @@ func TestModuleEngine_Close_Concurrent(t *testing.T) {
 				mappedRegions = append(mappedRegions, e.compiledFunctions[f].codeSegment)
 			}
 
-			const goroutines = 100
```
codefromthecrypt, line 457: I will recreate something similar to this on store_test.go and that would make sure both interpreter and JIT only call once even if it is less important internally for interpreter. These tests will do things like ensure the first exit code is the one returned, even if called twice with different values.

internal/wasm/module_context.go
```diff
@@ -105,13 +109,31 @@ func (f *FunctionInstance) ResultTypes() []publicwasm.ValueType {
 }
 
 // Call implements wasm.Function Call
-func (f *FunctionInstance) Call(ctx publicwasm.Module, params ...uint64) ([]uint64, error) {
+func (f *FunctionInstance) Call(m publicwasm.Module, params ...uint64) (ret []uint64, err error) {
+	if err = f.failIfClosed(); err != nil {
```
codefromthecrypt, line 113: this guard is new. before we still allowed new calls to modules that were already closed (ex if you kept a ref to the module).

codefromthecrypt:
I think that even though the logic will be duplicated, it is better to keep the guard internal to the engine implementations vs extract it just to avoid copying a test twice. I'll make that change

codefromthecrypt:
rebased over last few PRs. now interpreter and jit have the same internal guards and there is base infrastructure so we can prove it.

next step: backfill enginetests to cover exit code logic. eta tomorrow

codefromthecrypt:
rebased and updated hammer tests. next step is to backfill normal unit tests.

codefromthecrypt:
ok I added new tests into the existing ad-hoc file rather than move them in the same change as this is big enough already. I can do more code organization PRs later.

internal/wasm/interpreter/interpreter.go
```diff
@@ -547,6 +565,14 @@ func (me *moduleEngine) Call(ctx *wasm.ModuleContext, f *wasm.FunctionInstance,
 	return
 }
 
+// failIfClosed returns a sys.ExitError if moduleEngine.CloseWithExitCode was called.
+func (me *moduleEngine) failIfClosed() error {
+	if closed := atomic.LoadUint64(&me.closed); closed != 0 {
```
mathetake, line 570: curious how much does this cost as this is invoked anytime we execute `Call`. Anything notable in bench result?

codefromthecrypt: I'll add a bench, sure. though I don't think there's a cheaper way to achieve safety and eventually we need this to allow code sections to eventually release (unless we can do it cheaper in assembly). First, I'll add the bench, though!

codefromthecrypt:
CI isn't all that reliable, but here's the run before:

```
JIT  2392832	       540.5 ns/op
interpreter   517441	      2416 ns/op
wasmer-go   977470	      1304 ns/op
wasmtime-go   467770	      3083 ns/op
go-wasm3  1280509	       910.2 ns/op
```

and here's the run after
```
JIT  1713376	       663.9 ns/op
interpreter   372816	      3179 ns/op
wasmer-go   714549	      1702 ns/op
wasmtime-go   390840	      3837 ns/op
go-wasm3  1000000	      1197 ns/op
```

you'll notice everything was slower last time, but we're still a lot faster. my laptop is equally unreliable on a end-to-end, and in any case it isn't effecting the end result of being significantly faster than others.

On my host best in 3 before is:
```
JIT  2722590	       440.4 ns/op
interpreter   614865	      1915 ns/op
wasmer-go  1000000	      1208 ns/op
wasmtime-go   654991	      2259 ns/op
go-wasm3  1411844	       823.7 ns/op
```

and best in 3 after is
```
JIT  2680701	       437.6 ns/op
interpreter   656510	      1877 ns/op
wasmer-go  1000000	      1146 ns/op
wasmtime-go   644338	      2256 ns/op
go-wasm3  1670329	       725.3 ns/op
```

mathetake (approved):

internal/wasm/interpreter/interpreter.go
codefromthecrypt: I changed the code to ensure this is only called once per call, wiith notes why.

codefromthecrypt:
have to redo-some of this code as once I added the module name into the error message, it became clear something was awry.

codefromthecrypt:
ok I fixed error imprecision limited to one module deep. I think going deeper than that is nastier than we need to do at the moment. Note: while code looks duplicated it is because the internal structure of the interpreter and jit look a little similar. They are in fact different, and that's why tests are used to enforce behavior vs types.

codefromthecrypt:
hmm.. flakey I'll look into it.

codefromthecrypt:
I'll rebase this again after #419 as I think part of that found the flake cause (engine was shared between parallel tests by accident)

codefromthecrypt:
figured out what was going on. basically we have no implementation to propagate imported module's closed status. I left a TODO as addressing that is too much.

codefromthecrypt:
Thanks for the review and the advice along the way @mathetake!

PR #411, opened 2022-03-26
Refactors concurrent close tests

codefromthecrypt:
This refactors the concurrent close tests to use barriers, as opposed to time or offsets to implement phases. In doing so, it is easier to reason with and debuggable without changing the sleep value.


mathetake (approved):

codefromthecrypt:
thanks for letting me interrupt your weekend @mathetake!

PR #412, opened 2022-03-27
Deduplicates internalwasm.Engine tests

codefromthecrypt:
This deduplicates Engine tests so that there's less chance of copy/paste
errors and less requirement to use ad-hoc tests which are outside the
relevant source tree.

### DIFF
--- a/internal/testing/enginetest/enginetest.go
+++ b/internal/testing/enginetest/enginetest.go
@@ -1,0 +1,197 @@
+// Package enginetest contains tests common to any internalwasm.Engine implementation. Defining these as top-level
+// functions is less burden than copy/pasting the implementations, while still allowing test caching to operate.
+//
+// Ex. In simplest case, dispatch:
+//	func TestEngine_Call(t *testing.T) {
+//		enginetest.RunTestEngine_Call(t, NewEngine)
+//	}
+//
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
If looks good so far, I'll proceed. I want to extract the existing "close" tests here before doing anything else on the proper exit handling.

internal/testing/enginetest/enginetest.go
```diff
@@ -0,0 +1,229 @@
...
+// addFunction assigns and adds a function to the module.
+func addFunction(module *wasm.ModuleInstance, fn *wasm.FunctionInstance) {
+	module.Functions = append(module.Functions, fn)
+	// This link is essential for all engines. For example, functions call other functions defined in the same module.
+	fn.Module = module
+}
+
+// newModuleContext creates an internalwasm.ModuleContext for unit tests.
+//
+// Note: This sets fields that are not needed in the interpreter, but are required by code compiled by JIT. If a new
```
codefromthecrypt, line 220: @mathetake took me a while to re-remember this, so hope it isn't too much to leave this comment in

mathetake (approved):

codefromthecrypt:
Thanks for the look. I think I have enough baseline to complete #410 now!

PR #415, opened 2022-03-29
Refactors concurrency tests as "hammer" tests and adjusts style

codefromthecrypt:
This refactors tests that hammer shared state in ways that use locks or
atomics, so that they are consistent and also follow practice internally
used by Go itself. Notably, this supports the `-test.short` flag also.


codefromthecrypt:
@mathetake mind taking a look? If the style works for you I'll apply it to the others and mark for review.

codefromthecrypt:
ps using the "hammer" name like https://github.com/golang/go/search?q=hammer and will update RATIONALE.md with notes before merge

internal/wasm/store_test.go
```diff
@@ -202,29 +199,48 @@ func TestStore_concurrent(t *testing.T) {
...
-
-	// Concurrent release.
-	wg.Add(goroutines)
-	for i := 0; i < goroutines; i++ {
-		go func(i int) {
-			defer wg.Done()
-			require.NoError(t, s.CloseModule(strconv.Itoa(i)))
-			require.NoError(t, err)
-		}(i)
+	defer runtime.GOMAXPROCS(runtime.GOMAXPROCS(P))
```
mathetake, line 210: maybe half to ensure that we actually need goroutine switch? 

```suggestion
	defer runtime.GOMAXPROCS(runtime.GOMAXPROCS(P/2))
```

mathetake:
sg

internal/wasm/store_test.go
codefromthecrypt: cool. I'll comment that also :D

codefromthecrypt:
PTAL if good I'll repeat this for other tests!

mathetake:
LG

RATIONALE.md
```diff
@@ -199,6 +199,62 @@ See [wasm/jit/RATIONALE.md](internal/wasm/jit/RATIONALE.md).
...
 ## Golang patterns
 
+### Hammer tests
+Code that uses concurrency primitives, such as locks or atomics, should include "hammer tests", which run large loops
+inside a bounded amount of goroutines, run by half that many `GOMAXPROCS`. These are named consistently "hammer", so
+they are easy to find. The name inherits from some existing tests in [golang/go](https://github.com/golang/go/search?q=hammer&type=code).
+
+Here is an annotated description of the key pieces of a hammer test:
+1. `P` declares the count of goroutines to use, defaulting to 8 or 4 if `testing.Short`.
+   * 8 is typically less than a modern laptop's available cores, which allow multiple "hammer" tests to run in parallel.
```
codefromthecrypt, line 209:
```suggestion
   * Half this amount are the cores used, and 4 is less than a modern laptop's CPU. This allows multiple "hammer" tests to run in parallel.
```

tests/engine/hammer_test.go
```diff
@@ -12,61 +12,21 @@ import (
...
-func TestInterpreterConcurrency(t *testing.T) {
-	runAdhocTestsUnderHighConcurrency(t, wazero.NewRuntimeConfigInterpreter)
-	singleModuleHighConcurrency(t, wazero.NewRuntimeConfigInterpreter)
+	runAllTests(t, hammers, wazero.NewRuntimeConfigJIT())
 }
 
-func runAdhocTestsUnderHighConcurrency(t *testing.T, newRuntimeConfig func() *wazero.RuntimeConfig) {
-	t.Run("huge stack", func(t *testing.T) {
-		t.Parallel()
-		runAdhocTestUnderHighConcurrency(t, newRuntimeConfig, testHugeStack)
```
codefromthecrypt, line 31: ps these didn't really do what we thought they would as each uses a separate runtime :D It would be a lot of refactoring to make each "ad-hoc" test work with a shared runtime, particularly around dodging module names (ex most tests use the same module name "host". Instead, I removed these from the concurrency aka hammer tests as they aren't valid and distract from what's actually testing concurrency vs just running the same test many times (like a benchmark)

mathetake (approved):

codefromthecrypt:
I will look at the flakes and fix before merge

codefromthecrypt:
flakes gone!

PR #416, opened 2022-03-30
Extracts hammer.Hammer and adds atomic release of goroutines

codefromthecrypt:
This extracts `hammer.Hammer` as a utility for re-use, notably adding a
feature that ensures all tests run concurrently. Before, tests start in
a loop that could be delayed due to goroutine sheduling.


mathetake (approved):
Like it

codefromthecrypt:
Thanks for the look!

PR #418, opened 2022-03-30
Adds Module.Name and ports all hammer tests to the same impl

codefromthecrypt:
This exposes `wasm.Module.Name()` so that it is more coherent to
redefine a module with the same name, or know the name to import.

This uses that to port existing tests to use the new hammer
infrastructure (without module name collision).

### DIFF
--- a/internal/wasm/module_context.go
+++ b/internal/wasm/module_context.go
@@ -28,6 +28,11 @@ type ModuleContext struct {
 	sys *SysContext
 }
 
+// Name implements the same method as documented on wasm.Module
+func (m *ModuleContext) Name() string {
+	return m.module.Name
+}
+
 // WithMemory allows overriding memory without re-allocation when the result would be the same.
 func (m *ModuleContext) WithMemory(memory *MemoryInstance) *ModuleContext {
 	if memory != nil && memory != m.memory { // only re-allocate if it will change the effective memory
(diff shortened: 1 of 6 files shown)


internal/wasm/module_context.go
```diff
@@ -51,18 +56,18 @@ func (m *ModuleContext) Sys() *SysContext {
 	return m.sys
 }
 
-// WithContext implements wasm.Module WithContext
```
codefromthecrypt, line 54: I did this because the last `WithContext` here is self-referencing

tests/engine/adhoc_test.go
```diff
@@ -15,14 +15,13 @@ import (
 )
 
 var tests = map[string]func(t *testing.T, r wazero.Runtime){
-	"huge stack":                                 testHugeStack,
-	"unreachable":                                testUnreachable,
-	"recursive entry":                            testRecursiveEntry,
-	"imported-and-exported func":                 testImportedAndExportedFunc,
-	"host function with context parameter":       testHostFunctionContextParameter,
-	"host function with numeric parameter":       testHostFunctionNumericParameter,
-	"close module with in-flight calls":          testCloseInFlight,
```
codefromthecrypt, line 24: these two are now in the same test which has a matrix to cover both cases

mathetake (approved):

PR #423, opened 2022-03-31
asm(amd64): ROL and ROR on Register To Memory type.

mathetake:

### DIFF
--- a/internal/wasm/jit/jit_numeric_test.go
+++ b/internal/wasm/jit/jit_numeric_test.go
@@ -290,23 +290,24 @@ func TestCompiler_compile_And_Or_Xor_Shl_Rotl_Rotr(t *testing.T) {
 								require.NoError(t, err)
 
 								// Emit consts operands.
+								var x1Location *valueLocation
 								switch unsignedInt {
 								case wazeroir.UnsignedInt32:
 									err = compiler.compileConstI32(&wazeroir.OperationConstI32{Value: uint32(x1)})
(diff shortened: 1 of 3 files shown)


internal/wasm/jit/jit_numeric_test.go
```diff
@@ -279,107 +279,119 @@ func TestCompiler_compile_And_Or_Xor_Shl_Rotl_Rotr(t *testing.T) {
 						{1 << 63, 1}, {1, 1 << 63}, {1 << 63, 1 << 63},
 					} {
 						x1, x2 := values[0], values[1]
-						t.Run(fmt.Sprintf("x1=0x%x,x2=0x%x", x1, x2), func(t *testing.T) {
-							env := newJITEnvironment()
-							compiler := env.requireNewCompiler(t, newCompiler, nil)
-							err := compiler.compilePreamble()
-							require.NoError(t, err)
+						for _, x1OnRegister := range []bool{
```
mathetake, line 282: this unit test change is for increasing the test coverage -- previously https://github.com/tetratelabs/wazero/blob/main/internal/wasm/jit/jit_impl_amd64.go#L1686-L1688 this block hasn't been covered and that was the cause of missing implementation in assembler.

mathetake: acutally this tests fails on the current main but passes in this branch!

codefromthecrypt (approved):
great. I've tested locally this solves the problem also!

Issue #425, opened 2022-03-31
API shape


neilalexander:
I've been trying to pick up all of the recent changes since changing from `gasm` to `wazero` and I am really struggling with the new API shape. There's obviously a high rate of change, which is nice, but many of the examples in the comments are now obsolete and even code samples from GitHub issues from a couple weeks ago are no longer relevant.

Not to mention that suddenly the `wasm` and `wasi` packages — which I was happily using before — are now suddenly hidden away in `internal` and therefore private, so basically the entire API I was used to using in `gasm` is now seemingly gone. *Ouch*.

Let's say I want to load a WASI application. I create a runtime called `runtime`. I then need `wazero.WASISnapshotPreview1()`, which is a `*wazero.Module`, but `runtime.InstantiateModule` expects a `wasm.Module`. So I use `runtime.InstantiateModule`, which gives me a `wasm.Module`. At this point it isn't clear what the difference is between `wasm.Module` and `*wazero.Module`.

I then want to load my own functions into `env`, so I create a `runtime.NewModuleBuilder`, export some functions and then `Instantiate`, which should instantiate the module back into `runtime` and then returns a `wasm.Module` too. 

Finally, I want to load my own application: `runtime.CompileModule` which returns a `*wazero.Module`, and then I pass that into `runtime.InstantiateModule` which returns a `wasm.Module`.

So I've now got a `wazero.Runtime` and three `wasm.Module`s: the WASI bit, `env` and my application module. 

Now I have a problem. `wazero.StartWASICommandWithConfig` doesn't want a `wasm.Module`, it wants a `*wazero.Module`, and `wazero.StartWASICommandWithConfig` is the only way to pass in a `*wazero.SysConfig` that lets me define stdout, stdin etc. The alternative seemingly is that I can use `module.ExportedFunction("_start").Call`, but that doesn't take a config that lets me define stdout, stdin etc.

If I want to get an exported function or variable from my application module later, well, that only works if it's a `wasm.Module` and not a `*wazero.Module` — the former has `ExportedFunction` whereas the latter doesn't. So it's once again not clear what I'm supposed to do here and why there are two types for seemingly the same concept?

[...]

PR #427, opened 2022-04-01
Simplifies API per feedback

codefromthecrypt:
During #425, @neilalexander gave constructive feedback that the API is
both moving fast, and not good enough yet. This attempts to reduce the
incidental complexity at the cost of a little conflation.

### odd presence of `wasm` and `wasi` packages -> `api` package

We had public API packages in wasm and wasi, which helped us avoid
leaking too many internals as public. That these had names that look
like there should be implementations in them cause unnecessary
confusion. This squashes both into one package "api" which has no
package collission with anything.

We've long struggled with the poorly specified and non-uniformly
implemented WASI specification. Trying to bring visibility to its
constraints knowing they are routinely invalid taints our API for no
good reason. This removes all `WASI` commands for a default to invoke
the function `_start` if it exists. In doing so, there's only one path
to start a module.

Moreover, this puts all wasi code in a top-level package "wasi" as it
isn't re-imported by any internal types.

### Reuse of Module for pre and post instantiation to `Binary` -> `Module`

Module is defined by WebAssembly in many phases, from decoded to
instantiated. However, using the same noun in multiple packages is very
confusing. We at one point tried a name "DecodedModule" or
"InstantiatedModule", but this is a fools errand. By deviating slightly
from the spec we can make it unambiguous what a module is.

This make a result of compilation a `Binary`, retaining `Module` for an
instantiated one. In doing so, there's no longer any name conflicts
whatsoever.

### Confusion about config -> `ModuleConfig`

Also caused by splitting wasm into wasm+wasi is configuration. This
conflates both into the same type `ModuleConfig` as it is simpler than
trying to explain a "will never be finished" api of wasi snapshot-01 in
routine use of WebAssembly. In other words, this further moves WASI out
of the foreground as it has been nothing but burden.

```diff
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ For example, here's how you can allow WebAssembly modules to read
-wm, err := r.InstantiateModule(wazero.WASISnapshotPreview1())
-defer wm.Close()
+wm, err := wasi.InstantiateSnapshotPreview1(r)
+defer wm.Close()

-sysConfig := wazero.NewSysConfig().WithFS(os.DirFS("/work/home"))
-module, err := wazero.StartWASICommandWithConfig(r, compiled, sysConfig)
+config := wazero.ModuleConfig().WithFS(os.DirFS("/work/home"))
+module, err := r.InstantiateModule(binary, config)
 defer module.Close()
 ...
```


codefromthecrypt:
cc @pims @pkedy @inkeliz @tonyhb @MutatePat @wdvxdr1123 @leighmcculloch @birros @wuhuizuo I'm usually in favor of less is best, so hope these changes are ok with you.

codefromthecrypt:
forked an issue about the spectest drift as that happens on main also https://github.com/tetratelabs/wazero/issues/428

config.go
```diff
@@ -109,38 +110,28 @@ func (c *RuntimeConfig) WithFeatureSignExtensionOps(enabled bool) *RuntimeConfig
...
-// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#name-section%E2%91%A0
-// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#custom-section%E2%91%A0
-// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#modules%E2%91%A0%E2%91%A2
-func (m *Module) WithName(name string) *Module {
-	m.name = name
-	return m
+// Note: In WebAssembly language, this is a decoded, validated, and possibly also compiled module. wazero avoids using
+// the name "Module" for both before and after instantiation as the name conflation has caused confusion.
+// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#semantic-phases%E2%91%A0
+type Binary struct {
```
mathetake, line 118: maybe I would prefer `CompiledCode` to line with `FromCode` ?

codefromthecrypt: fine by me

codefromthecrypt:
ps after a shower I realized we no longer need an internal package for wasi anymore. This allows us to put the whole thing in the the top-level wasi package which is more intuitive, especially as we'll likely need one for assemblyscript if for no other reason to stop copy/pasting the same abort handler :D

codefromthecrypt:
OK I've isolated wasi into the top-level package "wasi" and unexported everything except this.

```diff
-wasi, err := r.InstantiateModule(wazero.WASISnapshotPreview1())
-defer wasi.Close()
+wm, err := wasi.InstantiateSnapshotPreview1(r)
+defer wm.Close()
```

So the good news is now there's only one place wasi is used and there's no coupling at all between that and wazero. It could even be in another repo (though here is far more convenient).

codefromthecrypt:
Thanks for the 👍 folks, and thanks especially to @neilalexander for speaking up. We took our eyes off the ball and glad someone blew a whistle.

I'll leave #425 open a bit though as there are some issues to address including better docs (ex the code snippets aren't tested).

codefromthecrypt:
ps our soft target for API stability (at least end of thrash even if little things have to change) is [gluecon](https://www.gluecon.com/) which is the first conference we'll talk about wazero.

We don't want to to advertise things generally too much when the API isn't stable, as it is not nice to lead too many into something that will cause them pain. This is why we are working so hard to figure out what is the best even if perfection will never happen.

We appreciate that folks are here using wazero knowing it is unstable as there's a chicken egg problem on api stability.. you want things to not impact too many on change, but still need different perspectives or else you can cement something not great!

I will continue raising PRs on drift to anything I know uses wazero unless someone says don't (because they don't need it etc). I feel it is important to have responsibility and it also teaches me a lot each time, and as a project maintainer, I have to understand what people are doing or things wander. I really appreciate that folks have been helping me update their code.

Anyway, we still have 6 weeks even under this schedule to knock out any other kinks, so please speak up!

codefromthecrypt:
PS here's a comprehensive example that uses all our recent API changes https://github.com/wapc/wapc-go/pull/21

PR #430, opened 2022-04-01
Updates to latest wabt and pares back to 1.0

codefromthecrypt:


mathetake:
```
Run make build.examples
# command-line-arguments
examples/testdata/cat.go:11:20: ReadFile not declared by package os
make: *** [Makefile:22: build.examples] Error 123
Error: Process completed with exit code 2.
```

oh maybe we should ping Go version (which is implicitly used by TinyGo) in example workflow

Makefile
```diff
@@ -35,7 +35,14 @@ build.spectest: # Note: wabt by default uses >1.0 features, so wast2json flags m
 		perl -pi -e 's/\((assert_return_canonical_nan|assert_return_arithmetic_nan)\s(\(invoke\s"[a-z._0-9]+"\s\((f[0-9]{2})\.const\s[a-z0-9.+:-]+\)\))\)/\(assert_return $$2 \($$3.const nan\)\)/g' $$f; \
 		perl -pi -e 's/\((assert_return_canonical_nan|assert_return_arithmetic_nan)\s(\(invoke\s"[a-z._0-9]+"\s\((f[0-9]{2})\.const\s[a-z0-9.+:-]+\)\s\([a-z0-9.\s+-:]+\)\))\)/\(assert_return $$2 \($$3.const nan\)\)/g' $$f; \
 		perl -pi -e 's/\((assert_return_canonical_nan|assert_return_arithmetic_nan)\s(\(invoke\s"[a-z._0-9]+"\s\((f[0-9]{2})\.const\s[a-z0-9.+:-]+\)\))\)/\(assert_return $$2 \($$3.const nan\)\)/g' $$f; \
-		wast2json --debug-names --disable-bulk-memory $$f; \
+		wast2json \
```
mathetake, line 38: nope, the failure is not for this workflow actually

mathetake:
do we even need examples.yaml in the first place though? :D

Makefile
codefromthecrypt: cool. I'll leave this in anyway as we have no control over the default choices version to version.

mathetake: sg

codefromthecrypt:
if we stop compiling examples they won't build when we drift changes, so I think we have to

mathetake (approved):

codefromthecrypt:
Thanks for the help.

to clarify, for some reason our examples and tests weren't re-compiling.. not sure why. Two things drifted:
1. since whenever last, wabt started using bulk-memory by default. I turned off all >1.0 flags defensively.
2. tinygo doesn't have one os.ReadFile, so I switched to ioutil

Should be ok again and thanks for the help!

PR #431, opened 2022-04-01
asm: complete arm64 instruction encodings for homemade assembler

### LINKED ISSUE #233: JIT: replace golang-asm with a thread-safe one (from scratch)

mathetake:
This commit implements all the arm64 instruction encodings necessary 
for our JIT compiler and replaces the golang-asm assembler with our
handmade assembler on arm64 platform. Notably, this allows us to do 
concurrent compilations. 

This closes #233 combined with #406.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,9 @@ module github.com/tetratelabs/wazero
 go 1.17
 
 require (
+	// Test-only dependency.
 	github.com/stretchr/testify v1.7.0
-	// Once we reach some maturity, remove this dep and implement our own assembler.
+	// Test-only dependency.
 	github.com/twitchyliquid64/golang-asm v0.15.1
 )
 
--- a/internal/asm/arm64/impl.go
+++ b/internal/asm/arm64/impl.go
@@ -1859,7 +1859,7 @@ func (a *AssemblerImpl) EncodeConstToRegister(n *NodeImpl) (err error) {
 			// https://github.com/golang/go/blob/release-branch.go1.15/src/cmd/internal/obj/arm64/asm7.go#L6570-L6583
 			tmpRegBits := registerBits(a.temporaryRegister)
 			// OOR $c, tmpReg
-			a.loadConstViaBitMaskImmediate(uc, tmpRegBits, true, false)
+			a.loadConstViaBitMaskImmediate(uc, tmpRegBits, true)
 
 			// ADD/SUB tmpReg, dstReg
 			a.addOrSub64BitRegisters(sfops, dstRegBits, tmpRegBits)
@@ -1907,7 +1907,7 @@ func (a *AssemblerImpl) EncodeConstToRegister(n *NodeImpl) (err error) {
 		ic := int64(c32)
 		if ic >= 0 && (ic <= 0xfff || (ic&0xfff) == 0 && (uint64(ic>>12) <= 0xfff)) {
 			if isBitMaskImmediate(uint64(c)) {
-				a.loadConstViaBitMaskImmediate(uint64(c), dstRegBits, false, true)
+				a.loadConstViaBitMaskImmediate(uint64(c), dstRegBits, false)
 				return
 			}
 		}
@@ -1920,7 +1920,7 @@ func (a *AssemblerImpl) EncodeConstToRegister(n *NodeImpl) (err error) {
 			// Also if the reverse of the const can fit within 16-bit range, do the same ^^.
 			a.load16bitAlignedConst((int64(^c32) >> (16 * t)), byte(t), dstRegBits, true, false)
 		} else if isBitMaskImmediate(uint64(c)) {
-			a.loadConstViaBitMaskImmediate(uint64(c), dstRegBits, false, true)
+			a.loadConstViaBitMaskImmediate(uint64(c), dstRegBits, false)
 		} else {
 			// Othewise we use MOVZ and MOVK to load it.
 			// https://github.com/golang/go/blob/release-branch.go1.15/src/cmd/internal/obj/arm64/asm7.go#L6623-L6630
@@ -1948,7 +1948,7 @@ func (a *AssemblerImpl) EncodeConstToRegister(n *NodeImpl) (err error) {
 		// https://github.com/golang/go/blob/release-branch.go1.15/src/cmd/internal/obj/arm64/asm7.go#L1798-L1852
 		if c >= 0 && (c <= 0xfff || (c&0xfff) == 0 && (uint64(c>>12) <= 0xfff)) {
 			if isBitMaskImmediate(uint64(c)) {
-				a.loadConstViaBitMaskImmediate(uint64(c), dstRegBits, true, false)
+				a.loadConstViaBitMaskImmediate(uint64(c), dstRegBits, true)
 				return
 			}
 		}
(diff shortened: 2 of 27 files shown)


mathetake:
ok progress: 30%

mathetake:
OK the left is Memory <-> Register data transfers, meaning that I have to implement addressing, the hardest part :D

mathetake:
95% of encoding done.. the left is relative jumps, and then we could run e2e tests

mathetake:
ok finished all encodings.. debugging time to match the binary output between homemade and golang-asm

mathetake:
ok  turns out that the constant pool strategy differs from Go's assembler.. will do next Monday..

mathetake:
almost there :D

mathetake:
ok passed all tests and the output binary 100% matches the ones from golang-asm  👯  will polish in the next few hours and then good to go!
!img[https://media.giphy.com/media/26BnaOFdD5Xbo1RMQ/giphy.gif]

mathetake:
ok ready for review!

internal/wasm/jit/engine.go
```diff
@@ -786,14 +786,7 @@ func (ce *callEngine) builtinFunctionMemoryGrow(mem *wasm.MemoryInstance) {
 	ce.moduleContext.memoryElement0Address = bufSliceHeader.Data
 }
 
-// golang-asm is not goroutine-safe so we take lock until we complete the compilation.
-// TODO: delete after https://github.com/tetratelabs/wazero/issues/233
-var assemblerMutex = &sync.Mutex{}
```
mathetake, line 791: 🕺

go.mod
```diff
@@ -4,8 +4,9 @@ module github.com/tetratelabs/wazero
 go 1.17
 
 require (
+	// Test-only dependency.
 	github.com/stretchr/testify v1.7.0
-	// Once we reach some maturity, remove this dep and implement our own assembler.
+	// Test-only dependency.
```
mathetake, line 9: 💃

internal/asm/arm64/impl.go
```diff
@@ -0,0 +1,2340 @@
...
+	code := a.Bytes()
+	for _, cb := range a.OnGenerateCallbacks {
+		if err := cb(code); err != nil {
+			return nil, err
+		}
+	}
+	return code, nil
+}
+
+// maybeFlushConstPool maybe flushes the current constants in the constPool.
```
codefromthecrypt, line 268: maybeFlushConstPool flushes the constant pool if endOfBinary or a boundary condition was met.

internal/asm/arm64/impl.go
```diff
@@ -0,0 +1,2340 @@
...
+		err = a.EncodeSIMDByteToRegister(n)
+	case OperandTypesTwoSIMDBytesToSIMDByteRegister:
+		err = a.EncodeTwoSIMDBytesToSIMDByteRegister(n)
+	default:
+		err = fmt.Errorf("encoder undefined for [%s] operand type", n.Types)
+	}
+	return
+}
+
+// CompileStandAlone implements asm.AssemblerBase.CompileStandAlone
```
codefromthecrypt, line 379: nested linking doesn't seem to navigate, which is why I have been switching to the following when implementing something out of package.

```
Xxx implements the same method as documented on yyy.Zzzz.
```

internal/asm/arm64/impl.go
```diff
@@ -0,0 +1,2340 @@
...
+	n := a.newNode(CSET, OperandTypesRegisterToRegister)
+	n.SrcReg = conditionalRegisterStateToRegister(cond)
+	n.DstReg = dstReg
+}
+
+func errorEncodingUnsupported(n *NodeImpl) error {
+	return fmt.Errorf("%s is unsupported for %s type", InstructionName(n.Instruction), n.Types)
+}
+
+func (a *AssemblerImpl) EncodeNoneToNone(n *NodeImpl) (err error) {
```
codefromthecrypt, line 528: need to be exported? if so // EncodeNoneToNone is exported because XXX

internal/asm/arm64/impl.go
```diff
@@ -0,0 +1,2340 @@
...
+	if n.Instruction != NOP {
+		err = errorEncodingUnsupported(n)
+	}
+	return
+}
+
+func (a *AssemblerImpl) EncodeJumpToRegister(n *NodeImpl) (err error) {
```
codefromthecrypt, line 535: same on export

codefromthecrypt:
I'm going to make a pass to do the little doc nits and push a commit vs have you do them :)

codefromthecrypt:
a couple things maybe can be changed now (or a later PR)?

```go
// Note: this interface is coupled and heavily influenced by golang-asm's API (i.e. Go's official assembler).
// Therefore, we will do the refactoring after golang-asm removal.
type AssemblerBase interface {
--snip

	// BuildJumpTable calculates the offsets between the first instruction `initialInstructions[0]`
	// and others (e.g. initialInstructions[3]), and wrote the calculated offsets into pre-allocated
	// `table` slice in little endian.
	//
	// TODO: This can be hidden into assembler implementation after golang-asm removal.
	BuildJumpTable(table []byte, initialInstructions []Node)
```

codefromthecrypt:
another TODO?
```
	// CompileJumpToMemory adds jump-type instruction whose destination is stored in the memory address specified by `baseReg`,
	// and returns the corresponding Node in the assembled linked list.
	//
	// Note: this has exactly the same implementation as the same method in asm.AssemblerBase in the homemade assembler.
	// Therefore, this will be removed after golang-asm removal.
	CompileJumpToMemory(jmpInstruction asm.Instruction, baseReg asm.Register)
```

codefromthecrypt:
```
// Bytes returns the encoded binary.
func (a *AssemblerImpl) Bytes() []byte {
	// 16 bytes alignment to match our impl with golang-asm.
	// https://github.com/golang/go/blob/release-branch.go1.15/src/cmd/internal/obj/arm64/asm7.go#L62
	//
	// TODO: Delete after golang-asm removal.
	if pad := 16 - a.Buf.Len()%16; pad > 0 && pad != 16 {
		a.Buf.Write(make([]byte, pad))
	}
	return a.Buf.Bytes()
}
```

mathetake:
I will resolve these todos later after I’m feeling comfortable to completely delete golang-ask code since resolving them unable me to keep golang-ask working

codefromthecrypt (approved):
OK I looked through all the changes and some similar things from the amd64 impl as well. 

Besides adding the RATIONALE including why we tend towards being like go assembly when in doubt (for familiarity and help troubleshooting), one thing to keep in mind is right now, we are depending heavily on round-tripping through the "debug" assemblers in order to get decent test coverage.

We shouldn't hunt coverage now, as I noticed a lot of code is just there to interop with go's assembler, and that would be pointless busywork. However, sometime between now and when that's removed, let's backfill some coverage in the natural unit tests for the parts that will stay, especially anything tricky. One way to prioritize could be to make comments saying "TODO: especially unit test this part" for places you lost time on in the past.

Besides that advice for post-merge, I'm really happy with the progress here and looking forward to the merge when you feel ready!

mathetake:
Thanks for the help!!

PR #437, opened 2022-04-04
Attempts to not mutate FunctionInstance on export

codefromthecrypt:
OK this fails on imported calls on both interpreter and JIT (in the enginetest which is missing something). Also JIT fails on another test due to subtracting from a zero PC

### DIFF
--- a/internal/testing/enginetest/enginetest.go
+++ b/internal/testing/enginetest/enginetest.go
@@ -219,7 +219,8 @@ func RunTestModuleEngine_Call_HostFn(t *testing.T, et EngineTester) {
 
 	e := et.NewEngine()
 
-	imported, importedMe := setupCallTests(t, e)
+	imported, importedMe, importing, importingMe := setupCallTests(t, e)
+	defer importingMe.Close()
 	defer importedMe.Close()
 
 	// Ensure the base case doesn't fail: A single parameter should work as that matches the function signature.
@@ -243,6 +244,11 @@ func RunTestModuleEngine_Call_HostFn(t *testing.T, et EngineTester) {
 			module: imported.Ctx,
 			fn:     imported.Exports[callHostFnName].Function,
 		},
+		{
+			name:   callImportCallHostFnName,
+			module: importing.Ctx,
+			fn:     importing.Exports[callImportCallHostFnName].Function,
+		},
 	}
 	for _, tt := range tests {
 		tc := tt
@@ -260,7 +266,8 @@ func RunTestModuleEngine_Call_HostFn(t *testing.T, et EngineTester) {
 func RunTestModuleEngine_Call_Errors(t *testing.T, et EngineTester) {
 	e := et.NewEngine()
 
-	imported, importedMe := setupCallTests(t, e)
+	imported, importedMe, importing, importingMe := setupCallTests(t, e)
+	defer importingMe.Close()
 	defer importedMe.Close()
 
 	tests := []struct {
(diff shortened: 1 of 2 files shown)


internal/testing/enginetest/enginetest.go
```diff
@@ -260,112 +266,125 @@ func RunTestModuleEngine_Call_HostFn(t *testing.T, et EngineTester) {
...
 }
 
 const (
-	wasmFnName     = "wasm_div_by"
-	hostFnName     = "host_div_by"
-	callHostFnName = "call->" + hostFnName
+	wasmFnName               = "wasm_div_by"
+	hostFnName               = "host_div_by"
+	callHostFnName           = "call->" + hostFnName
+	callImportCallHostFnName = "call_import->" + callHostFnName
```
codefromthecrypt, line 387: all tests fail on this. maybe there's a setup issue

codefromthecrypt:
@mathetake my brain is mush so if you can take a look and maybe push a commit to get expected behavior on a module-defined function calling an imported one, much obliged. I can't see what's wrong and really want to be able to test moduleEngine independently inclusive of errors, without relying on Store side-effects, to preserve time in the future.

codefromthecrypt:
almost all code is here https://github.com/tetratelabs/wazero/pull/438

codefromthecrypt:
ok I rebased this over #438 to show what's different.

mathetake:
oh what's failing now?

mathetake:
ah CI isn't running ok 😄

codefromthecrypt:
clicked ready for review in case that kicks CI

codefromthecrypt:
oh it is because I put this against the other branch

codefromthecrypt:
https://github.com/tetratelabs/wazero/actions/runs/2088507390

internal/testing/enginetest/enginetest.go
```diff
@@ -402,7 +431,30 @@ func setupCallTests(t *testing.T, e wasm.Engine) (*wasm.ModuleInstance, wasm.Mod
...
+		Type:  ft,
+		Body:  []byte{wasm.OpcodeLocalGet, 0, wasm.OpcodeCall, 0 /* only one imported function */, wasm.OpcodeEnd},
+		Index: 1, // after import
+	}
+	addFunction(importing, callImportCallHostFnName, callImportedHostFn)
+
+	// Compile the importing module
+	importingMe, err := e.NewModuleEngine(importing.Name, importedFunctions, importing.Functions, nil, nil)
+	require.NoError(t, err)
+	linkModuleToEngine(importing, importedMe)
```
mathetake, line 452: This must be `importingMe`, not `importedMe`.
```suggestion
	linkModuleToEngine(importing, importingMe)
```

mathetake: still failure remains, but this makes the error constsitent between Iterpreter and JIT

mathetake:
oh yeah this solves the outstanding issues except ExpecteError=`TODO:` cases 😄

mathetake:
fix pushed!

mathetake:
ok cool passed on all platforms https://github.com/tetratelabs/wazero/runs/5813115061?check_suite_focus=true

internal/testing/enginetest/enginetest.go
```diff
@@ -449,7 +461,7 @@ func setupCallTests(t *testing.T, e wasm.Engine) (*wasm.ModuleInstance, wasm.Mod
 	// Compile the importing module
 	importingMe, err := e.NewModuleEngine(importing.Name, importedFunctions, importing.Functions, nil, nil)
 	require.NoError(t, err)
-	linkModuleToEngine(importing, importedMe)
```
codefromthecrypt: hah definitely appreciate the second eyes!

PR #438, opened 2022-04-04
Moves close responsibility to the module context

codefromthecrypt:
This moves the responsibility to close a module from the engine to the
context. The reason for this is that the engine is what defines the
function. When a module is closed, it is often from an imported host
function. If it is on the module engine, it is easy to accidentally
close an imported module.

This refactors the WASI tests also, to ensure they aren't cheating too
much. This allows us to know for example "proc_exit" works without too
much trouble.

### DIFF
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -183,16 +183,6 @@ func (m *ModuleInstance) buildExports(exports map[string]*Export) {
 		switch exp.Type {
 		case ExternTypeFunc:
 			ei = &ExportInstance{Type: exp.Type, Function: m.Functions[index]}
-			// The module instance of the host function is a fake that only includes the function and its types.
-			// We need to assign the ModuleInstance when re-exporting so that any memory defined in the target is
-			// available to the wasm.ModuleContext Memory.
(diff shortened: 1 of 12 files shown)


internal/wasm/store.go
```diff
@@ -183,16 +183,6 @@ func (m *ModuleInstance) buildExports(exports map[string]*Export) {
 		switch exp.Type {
 		case ExternTypeFunc:
 			ei = &ExportInstance{Type: exp.Type, Function: m.Functions[index]}
-			// The module instance of the host function is a fake that only includes the function and its types.
-			// We need to assign the ModuleInstance when re-exporting so that any memory defined in the target is
-			// available to the wasm.ModuleContext Memory.
-			//
-			// TODO(adrian): shouldn't this be race if multiple modules re-exporting?
```
codefromthecrypt, line 190: Note: this also stops mutating an imported function when exporting it. Notice all the other tests are still passing.

mathetake (approved):
Love it

codefromthecrypt:
thanks so much for the help!

PR #439, opened 2022-04-05
wat: syncs spectest with latest text decoder

codefromthecrypt:
This helps us prioritize text format features

### DIFF
--- a/tests/spectest/spec_test.go
+++ b/tests/spectest/spec_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/tetratelabs/wazero/internal/wasm/binary"
 	"github.com/tetratelabs/wazero/internal/wasm/interpreter"
 	"github.com/tetratelabs/wazero/internal/wasm/jit"
+	"github.com/tetratelabs/wazero/internal/wasm/text"
 	"github.com/tetratelabs/wazero/internal/wasmruntime"
 )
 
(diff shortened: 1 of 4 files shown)


tests/spectest/spec_test.go
```diff
@@ -196,79 +197,80 @@ func (c command) expectedError() (err error) {
...
+	mod.GlobalSection = append(mod.GlobalSection, &wasm.Global{
+		Type: &wasm.GlobalType{ValType: wasm.ValueTypeI64},
+		Init: &wasm.ConstantExpression{Opcode: wasm.OpcodeI64Const, Data: []byte{0x9a, 0x5} /* = 666 */},
+	})
+	mod.ExportSection["global_i64"] = &wasm.Export{Name: "global_i64", Index: 1, Type: wasm.ExternTypeGlobal}
+
+	// (global (export "global_f32") f32)
+	mod.GlobalSection = append(mod.GlobalSection, &wasm.Global{
+		Type: &wasm.GlobalType{ValType: wasm.ValueTypeF32},
+		Init: &wasm.ConstantExpression{Opcode: wasm.OpcodeF32Const, Data: []byte{0x44, 0x26, 0x80, 0x00}},
```
codefromthecrypt, line 257: ps I'm not sure in the docs where the magic constants are defined.. if you know, please let me and I'll comment them.

codefromthecrypt: or if we define them, I can just comment "arbitrary"

mathetake: I thought the const is here https://github.com/WebAssembly/spec/blob/ffb5e3b40db7a45237d4e2fc972c1552ac4abef5/interpreter/README.md#spectest-host-module but I couldn't find :D but they need to be these values otherwise fail so not arbitrary :D

mathetake (approved):
like it

tests/spectest/spec_test.go
mathetake: oh found it https://github.com/WebAssembly/spec/blob/f5a260a2025ba4d7d398654581c7c532e3a2c319/interpreter/script/js.ml#L44-L49

codefromthecrypt: I found them https://github.com/WebAssembly/spec/blob/wg-1.0/test/harness/sync_index.js#L79-L91

always 🕵️

codefromthecrypt:
thanks for the link. I noticed we exported an unused global so removed it.

PR #440, opened 2022-04-05
Adds ModuleBuilder.ExportMemory and ExportMemoryWithMax

codefromthecrypt:
This adds functions to configure memory with ModuleBuilder. This uses
two functions, ExportMemory and ExportMemoryWithMax, as working with
uint32 pointers is awkward.

See https://github.com/summerwind/the-art-of-webassembly-go/issues/1

### DIFF
--- a/internal/wasm/host.go
+++ b/internal/wasm/host.go
@@ -4,40 +4,62 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/tetratelabs/wazero/internal/wasmdebug"
 )
 
 // NewHostModule is defined internally for use in WASI tests and to keep the code size in the root directory small.
-func NewHostModule(moduleName string, nameToGoFunc map[string]interface{}) (*Module, error) {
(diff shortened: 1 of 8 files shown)


internal/wasm/host.go
```diff
@@ -4,90 +4,112 @@ import (
...
 		}
+		return fmt.Errorf("only one memory is allowed, but configured [%s]", strings.Join(memoryNames, ","))
 	}
 
-	result := m.SectionElementCount(SectionIDType)
-	m.TypeSection = append(m.TypeSection, ft)
-	return result
-}
-
-func (m *Module) validateHostFunctions() error {
```
codefromthecrypt, line 63: validateHostFunctions() was unneeded as they are only instantiated here and validated on insertion

mathetake (approved):

codefromthecrypt:
thanks for the look. next we're goin global

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/812 will revert this as the project mentioned stalled out and it is also incorrect assumption that memory defined on the host would end up used.

PR #441, opened 2022-04-05
Adds ModuleBuilder.ExportGlobalXXX to configure constants

codefromthecrypt:
I think this is the last to port @summerwind's lovely codebase

See https://github.com/summerwind/the-art-of-webassembly-go/issues/1

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/tetratelabs/wazero/api"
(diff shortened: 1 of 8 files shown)


mathetake (approved):

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/812 will revert this as the project mentioned stalled out and it is also incorrect assumption that globlals defined on the host would end up used.

PR #442, opened 2022-04-05
Fixes global numeric types to have max of signed encoding

codefromthecrypt:
This adjusts towards the exiting code which used int32/64 instead of
uint32/64. The reason is that the spec indicates intepretation as signed
numbers, which affects the maximum value.

See https://www.w3.org/TR/wasm-core-1/#value-types%E2%91%A2


codefromthecrypt:
Note: I left function signatures allowing uints because they often use numbers as raw bytes or use unsigned types to hint something can never be negative.

api/wasm.go
```diff
@@ -27,13 +27,19 @@ import (
 //	result := api.DecodeF64(result[0])
 //
 // Note: This is a type alias as it is easier to encode and decode in the binary format.
+// Note: Integer types are encoded as signed numbers, which affects their max value.
+// See https://www.w3.org/TR/wasm-core-1/#value-types%E2%91%A2
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#binary-valtype
 type ValueType = byte
 
 const (
+	// ValueTypeI32 is a 32-bit integer with max value math.MaxInt32.
```
mathetake, line 36: Oh I think this comes with confusion.. as what the spec is saying is that its *binary encoding* is signed leb128 (therefore the binary representation of i32 const or whatever has limit on math.MaxInt32), but it doesn't mean the value itself has limit on math.MaxInt32. For example, host function `func (i32, it32)` needs to handle (2**32).

mathetake: For globals, yes, the value must be encoded in text or binary format which means it has the limit on math.MaxtInt32 .

codefromthecrypt: gotcha.. I will move this to the global instead!

mathetake: For reference on what `i32` means:
```
The types i32 and i64 classify 32 and 64 bit integers, respectively. Integers are not inherently signed or unsigned, their interpretation is determined by individual operations.
```

codefromthecrypt: PTAL!

internal/u64/u64.go
```diff
@@ -0,0 +1,15 @@
+package u64
+
+// LeBytes returns a byte array corresponding to the 8 bytes in the uint64 in little-endian byte order.
+func LeBytes(v uint64) []byte {
```
mathetake, line 4: is this exactly the same as `binary.LittleEndian.PutUint64`?

mathetake (approved):

internal/u64/u64.go
codefromthecrypt: yep, except with that you have to pre-allocate an array to put it into which makes the call sites suffer more.

That's why before we extracted a similar named utility, yet in other places did things like this:

```
				// Allocate bytes with enough size for all types.
 				expr := &ConstantExpression{Data: make([]byte, 8)}
```

codefromthecrypt: oh if you are referring to the impl.. I wanted to add another utility for only the lower bits, but found some things rely on reading 8 even if only 4 are needed. I didn't want to dig into why during the same change, but in any case that's why I did inline declaration vs initializing a byte slice and passing it to PutUint64 which does the same thing

codefromthecrypt: pretty sure there will be more maintenance, so if we want to change to two step I'll do it in the next hunt!

codefromthecrypt:
thanks for the help with this @mathetake it is sometimes hard to keep track what is a binary representation limitation vs a runtime one!

PR #443, opened 2022-04-06
Consistently uses LEB128 signed encoding for global constants

codefromthecrypt:
Global constants can be defined in wasm or in ModuleBuilder. In either
case, they end up being decoded and interpreted during instantiation.
This chooses signed encoding to avoid surprises. A more comprehensive
explanation was added to RATIONALE.md, but the motivation was a global
100 coming out negative.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -107,7 +107,7 @@ type ModuleBuilder interface {
 	//
 	// Note: If a global is already exported with the same name, this overwrites it.
 	// Note: The maximum value of v is math.MaxInt32 to match constraints of initialization in binary format.
-	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#value-types%E2%91%A2
+	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#value-types%E2%91%A0
 	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-globaltype
 	ExportGlobalI32(name string, v int32) ModuleBuilder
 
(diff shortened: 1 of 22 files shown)


builder.go
```diff
@@ -202,8 +202,8 @@ func (b *moduleBuilder) ExportMemoryWithMax(name string, minPages, maxPages uint
 func (b *moduleBuilder) ExportGlobalI32(name string, v int32) ModuleBuilder {
 	b.nameToGlobal[name] = &wasm.Global{
 		Type: &wasm.GlobalType{ValType: wasm.ValueTypeI32},
-		// Signed per https://www.w3.org/TR/wasm-core-1/#value-types%E2%91%A2
-		Init: &wasm.ConstantExpression{Opcode: wasm.OpcodeI32Const, Data: leb128.EncodeUint32(uint32(v))},
```
codefromthecrypt, line 206: this was the bug as we decoded as int32

mathetake (approved):

wasm_test.go
```diff
@@ -164,6 +165,8 @@ func TestModule_Memory(t *testing.T) {
 
 // TestModule_Global only covers a couple cases to avoid duplication of internal/wasm/global_test.go
 func TestModule_Global(t *testing.T) {
+	globalVal := int64(100) // intentionally a value that differs in signed vs unsigned encoding
```
codefromthecrypt, line 168: this was the killer value, used in the art of webassembly helloworld example

codefromthecrypt:
Thanks for the help behind the scenes, @mathetake!

PR #445, opened 2022-04-06
Fix example typo

Taction:
The vm has been handled before, here should be the module.


mathetake (approved):

mathetake:
@Taction thank you for finding this! 👯

PR #446, opened 2022-04-07
Supports functions with multiple results (multi-value)

### LINKED ISSUE #433: Add example for how to do multiple returns with "result parameters"
WebAssembly 1.0 restricts results to a single value. The secret way people get around this is by passing a "result parameter" which is a memory offset to write the Nth return value. As this is easier to do than say, we need to add an example so we can use it to help instruct people.

codefromthecrypt:
This adds support for functions (and blocks) with multiple values. Notably, this allows functions to return more than one result without using memory offset parameters (such as done in WASI).

Ex. before:
```wasm
  (func $get_age (param $result_offset.age i32) (result (;errno;) i32)
    local.get 0   ;; stack = [$result_offset.age]
    i64.const 37  ;; stack = [$result_offset.age, 37]
    i64.store     ;; stack = []
    i32.const 0   ;; stack = [0]
  )
```

Now
```wasm
  (func $get_age (result (;age;) i64 (;errno;) i32)
    i64.const 37  ;; stack = [37]
    i32.const 0   ;; stack = [37, 0]
  )
```

This uses the ["multi-value" feature](https://github.com/WebAssembly/spec/blob/main/proposals/multi-value/Overview.md) and includes all spectests merged when that was.

Note: "multi-value" is finished status, but it is not yet a W3C recommendation. wazero only enables
W3C recommended features by default, so to enable "multi-value", you need configuration:
```go
r := wazero.NewRuntimeWithConfig(
	wazero.NewRuntimeConfig().WithFeatureMultiValue(true),
	// ^^ Note: You can enable all features via WithFinishedFeatures.
)
```

This adds an example to illustrate both wasm and go function usage.

Fixes #433


tests/post1_0/multi-value/multi_value_test.go
```diff
@@ -0,0 +1,60 @@
...
+		t.Skip()
+	}
+	testMultiValue(t, wazero.NewRuntimeConfigJIT)
+}
+
+func TestMultiValue_Interpreter(t *testing.T) {
+	testMultiValue(t, wazero.NewRuntimeConfigInterpreter)
+}
+
+func testMultiValue(t *testing.T, newRuntimeConfig func() *wazero.RuntimeConfig) {
```
codefromthecrypt, line 27: this is the test that currently fails, due to function validation

mathetake:
fixed a func validation.. next is to fix wazeroir's signature for blocks https://github.com/tetratelabs/wazero/blob/main/internal/wazeroir/signature.go#L154-L157

mathetake:
OK fixed

internal/wazeroir/compiler.go
```diff
@@ -263,11 +262,9 @@ operatorSwitch:
 		// Create a new frame -- entering loop.
 		frame := &controlFrame{
 			frameID:          c.nextID(),
-			originalStackLen: len(c.stack),
+			originalStackLen: len(c.stack) - len(bt.Params),
```
codefromthecrypt, line 265: @mathetake thanks tons for fixing all this and the added docs. This is the only one I'd like to doc better. Have a good answer for why we subtract the block type params from the original stack length?

mathetake: memory is really fuzzy lemme think about it a bit..

mathetake:
oh I sense that this still fails with params with if block..

internal/wazeroir/compiler.go
mathetake: renamed for clarity!

mathetake:
so pushed a commit for it https://github.com/tetratelabs/wazero/pull/446/commits/bc3bd89ba7d99c4c12ce0eaf74e74e41e57cfc14 

maybe we should add more spectests

codefromthecrypt:
Thanks a lot @mathetake great job!

TODO(Adrian):
* add func type parser (text)
* add tests for the new code
* add more tests from the spec test diff
* add example code for multiple returns before (1.0) and after (1.0+multi-value)

https://github.com/WebAssembly/spec/commit/484180ba3d9d7638ba1cb400b699ffede796927c

codefromthecrypt:
text format is done. now, need to backfill wazeroir (intermediate representation) tests, as these are the parts tricky to figure out if left untested until spectests.

codefromthecrypt:
I want to complete 100pct of the spec test changes before merge, so will be another couple days to finish (because it is a weekend)

tests/post1_0/multi-value/testdata/func.wat
```diff
@@ -0,0 +1,104 @@
...
+
+  (func (export "break-br_table-num-num") (param i32) (result i32 i64)
+    (br_table 0 0 (i32.const 50) (i64.const 51) (local.get 0))
+    (i32.const 51) (i64.const 52)
+  )
+
+  (func (export "break-br_table-nested-num-num") (param i32) (result i32 i32)
+    (i32.add
+      (block (result i32 i32)
+        (br_table 0 1 0 (i32.const 50) (i32.const 51) (local.get 0))
```
codefromthecrypt, line 77: this fails with input -3 on the interpreter 

runtime error: index out of range [-99]
--snip--
		/Users/adrian/oss/wazero/internal/wasm/interpreter/interpreter.go:635 +0x9ba7
        	            	github.com/tetratelabs/wazero/internal/wasm/interpreter.(*moduleEngine).Call(0x1181a30, 0xc00034c000, 0xc00033eaf0, {0xc0003413d8, 0x1, 0x1069ce0})
        	            		/Users/adrian/oss/wazero/internal/wasm/interpreter/interpreter.go:536 +0x43a

mathetake: https://github.com/tetratelabs/wazero/pull/446/commits/75ba48894f36c0674fa22049bd19547bb62e5045 pushed a fix

codefromthecrypt: thanks tons! back on track!

codefromthecrypt:
ps reason I'm adding all the spectests in this PR is because we keep finding glitches. If we didn't find any I would add them post merge. that this PR ends up bulky is what it is.. basically this is one of the bigger code effecting PRs due to its impact on block structure

internal/wasm/func_validation.go
```diff
@@ -904,28 +928,64 @@ func (s *valueTypeStack) popStackLimit() {
 	}
 }
 
-func (s *valueTypeStack) pushStackLimit() {
-	s.stackLimits = append(s.stackLimits, len(s.stack))
+// pushStackLimit pushes the control frame's bottom of the stack.
+func (s *valueTypeStack) pushStackLimit(params int) {
```
mathetake, line 932: `limit` is ambiguous so you can rename it as you like e.g.
```suggestion
func (s *valueTypeStack) pushStackFloor(params int) {
```

codefromthecrypt:
done all spectests

Next:
* add example code for multiple returns before (1.0) and after (1.0+multi-value)
* polish PR desc

examples/results_test.go
```diff
@@ -0,0 +1,165 @@
+package examples
```
codefromthecrypt, line 1: this is the example

internal/wasm/text/errors.go
```diff
@@ -70,6 +70,7 @@ func moreThanOneInvalidInSection(section wasm.SectionID) error {
 	return moreThanOneInvalid(wasm.SectionIDName(section))
 }
 
+// TODO: multi-value
```
mathetake, line 73: is this TODO for another PR?

mathetake (approved):

internal/wasm/text/errors.go
codefromthecrypt: oh that was a leftover. I'll scrub it!

codefromthecrypt:
Thanks for the week of help here and behind the scenes @mathetake!

PR #448, opened 2022-04-08
leb128: use fixed length slice for encoding cache

mathetake:


mathetake:
LOL not 127, but 128

codefromthecrypt:
thanks, good eye!

PR #454, opened 2022-04-12
jit: decouple compiled code from moduleInstance's pointer

mathetake:
This commit decouples the *wasm.ModuleInstance from the compiled code.

This is necessary to reuse same machine code for multiple module instances
originating from the same binary (or text), which is a necessary mechanism for
compilation caches I will do in the next PRs.

PR #457, opened 2022-04-13
Add compilation cache functionality and Close on CompiledCode.

mathetake:
Thanks to #454, now the compiled binary (code segment) can be reused for
multiple module instances originating from the same source (wasm.Module).

This commit introduces the caching mechanism on engine where it caches
compiled functions keyed on `wasm.Module`. As a result, this allows us to
do the fast module instantiation from the same *CompiledCode.

In order to release the cache properly, this also adds `Close` method 
on CompiledCode.

Here's some bench result for instantiating multiple modules from the same CompiledCode:

```
name                           old time/op    new time/op    delta
Initialization/interpreter-32    2.84ms ± 3%    0.06ms ± 1%  -97.73%  (p=0.008 n=5+5)
Initialization/jit-32            10.7ms ±18%     0.1ms ± 1%  -99.52%  (p=0.008 n=5+5)

name                           old alloc/op   new alloc/op   delta
Initialization/interpreter-32    1.25MB ± 0%    0.15MB ± 0%  -88.41%  (p=0.008 n=5+5)
Initialization/jit-32            4.46MB ± 0%    0.15MB ± 0%  -96.69%  (p=0.008 n=5+5)

name                           old allocs/op  new allocs/op  delta
Initialization/interpreter-32     35.2k ± 0%      0.3k ± 0%  -99.29%  (p=0.008 n=5+5)
Initialization/jit-32             94.1k ± 0%      0.2k ± 0%  -99.74%  (p=0.008 n=5+5)
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -148,6 +148,33 @@ func (c *RuntimeConfig) WithFeatureMultiValue(enabled bool) *RuntimeConfig {
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#semantic-phases%E2%91%A0
 type CompiledCode struct {
 	module *wasm.Module
+	// cachedEngines maps wasm.Engine to []*wasm.Module which originate from this .module.
+	// This is necessary to tracks which engine caches *Module where latter might be different
+	// from .module via import replacement config.
+	cachedEngines map[wasm.Engine]map[*wasm.Module]struct{}
+}
+
+// Close releases all the allocated resources for this CompiledCode.
+func (c *CompiledCode) Close() {
+	for engine, modules := range c.cachedEngines {
+		for module := range modules {
+			engine.ReleaseCompilationCache(module)
+		}
+	}
+}
+
+func (c *CompiledCode) addCacheEntry(module *wasm.Module, engine wasm.Engine) {
+	if c.cachedEngines == nil {
+		c.cachedEngines = map[wasm.Engine]map[*wasm.Module]struct{}{}
+	}
+
+	cache, ok := c.cachedEngines[engine]
+	if !ok {
+		cache = map[*wasm.Module]struct{}{}
+		c.cachedEngines[engine] = cache
+	}
+
+	cache[module] = struct{}{}
 }
 
 // ModuleConfig configures resources needed by functions that have low-level interactions with the host operating system.
--- a/config.go
+++ b/config.go
@@ -149,8 +149,8 @@ func (c *RuntimeConfig) WithFeatureMultiValue(enabled bool) *RuntimeConfig {
 type CompiledCode struct {
 	module *wasm.Module
 	// cachedEngines maps wasm.Engine to []*wasm.Module which originate from this .module.
-	// This is necessary to tracks which engine caches *Module where latter might be different
-	// from .module via import replacement config.
+	// This is necessary to track which engine caches *Module where latter might be different
+	// from .module via import replacement config (ModuleConfig.WithImport).
 	cachedEngines map[wasm.Engine]map[*wasm.Module]struct{}
 }
 
--- a/internal/wasm/jit/engine.go
+++ b/internal/wasm/jit/engine.go
@@ -378,24 +378,20 @@ func (e *engine) NewModuleEngine(name string, module *wasm.Module, importedFunct
 		importedFunctionCount: imported,
 	}
 
+	for idx, f := range importedFunctions {
+		cf, ok := e.getCompiledFunction(f)
+		if !ok {
+			return nil, fmt.Errorf("import[%d] func[%s]: uncompiled", idx, f.DebugName)
+		}
+		me.compiledFunctions = append(me.compiledFunctions, cf)
+	}
+
 	if cached, ok := e.getCachedCompiledFunctions(module); ok { // cache hit.
-		for i, c := range cached {
-			if i >= len(importedFunctions) {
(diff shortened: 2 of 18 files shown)


mathetake:
ok the left is to implement cache for interpreter as well!

config.go
```diff
@@ -148,6 +148,33 @@ func (c *RuntimeConfig) WithFeatureMultiValue(enabled bool) *RuntimeConfig {
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#semantic-phases%E2%91%A0
 type CompiledCode struct {
 	module *wasm.Module
+	// cachedEngines maps wasm.Engine to []*wasm.Module which originate from this .module.
+	// This is necessary to tracks which engine caches *Module where latter might be different
```
codefromthecrypt, line 152:
```suggestion
	// This is necessary to track which engine caches *Module where latter might be different
```

config.go
```diff
@@ -148,6 +148,33 @@ func (c *RuntimeConfig) WithFeatureMultiValue(enabled bool) *RuntimeConfig {
...
+	// from .module via import replacement config.
```
codefromthecrypt, line 153:
```suggestion
	// from .module via import replacement config (ModuleConfig.WithImport).
```

internal/wasm/jit/engine.go
```diff
@@ -193,6 +194,19 @@ type (
 	compiledFunctionStaticData = [][]byte
 )
 
+func (c *compiledFunction) clone(newSourceInstance *wasm.FunctionInstance) *compiledFunction {
+	// Note: we don't need to set finalizer to mummap the code segment since it is
+	// already a target of mummap by the finalizer set on the original compiledFunction `c`.
```
codefromthecrypt, line 199:
```suggestion
	// already a target of munmap by the finalizer set on the original compiledFunction `c`.
```

internal/wasm/jit/engine_test.go
```diff
@@ -477,3 +471,114 @@ func TestJIT_SliceAllocatedOnHeap(t *testing.T) {
 		})
 	}
 }
+
+func TestEngine_CachedCompiledFunctionsPerModule(t *testing.T) {
```
codefromthecrypt, line 459:
```suggestion
// TODO: move most of this logic to enginetest.go so that there is less drift between interpreter and jit
func TestEngine_CachedCompiledFunctionsPerModule(t *testing.T) {
```

internal/wasm/jit/engine_test.go
```diff
@@ -477,3 +471,114 @@ func TestJIT_SliceAllocatedOnHeap(t *testing.T) {
...
+	for i := range actual {
+		require.Equal(t, exp[i], actual[i])
+	}
+
+	e.deleteCachedCompiledFunctions(m)
+	_, ok = e.getCachedCompiledFunctions(m)
+	require.False(t, ok)
+}
+
+func TestEngine_NewModuleEngine_cache(t *testing.T) {
```
codefromthecrypt, line 481:
```suggestion
// TODO: move most of this logic to enginetest.go so that there is less drift between interpreter and jit
func TestEngine_NewModuleEngine_cache(t *testing.T) {
```

mathetake:
ps hopefully we should do compilation in CompileModule, but due to the current limitation for code being coupled with runtime representation of typeID, we do compilation during instantiation. This will be changed after making typeID deterministic across run.

tests/engine/adhoc_test.go
```diff
@@ -338,3 +340,41 @@ func testCloseInFlight(t *testing.T, r wazero.Runtime) {
...
+	// Instantiate multiple modules with the same source (*CompiledCode).
+	for i := 0; i < 100; i++ {
+		module, err := r.InstantiateModuleWithConfig(compiled, wazero.NewModuleConfig().WithName(strconv.Itoa(i)))
+		require.NoError(t, err)
+		defer module.Close()
+
+		// Ensure that compilation cache doesn't cause race on memory instance.
+		before, ok := module.Memory().ReadUint64Le(1)
+		require.True(t, ok)
+		// Value must be zero as the memory must not be affected by the previously insantiated modules.
```
codefromthecrypt, line 365:
```suggestion
		// Value must be zero as the memory must not be affected by the previously instantiated modules.
```

wasm.go
```diff
@@ -72,6 +72,7 @@ type Runtime interface {
 	// Ex.
 	//	r := wazero.NewRuntime()
 	//	code, _ := r.CompileModule(source)
+	//  defer code.Close()
```
codefromthecrypt, line 75:
```suggestion
	//	defer code.Close()
```

codefromthecrypt (approved):
Looks like a great win for memory usage!

I made some comments for later cleanup. It isn't necessary to spend too much time on enginetest right now.

internal/wasm/jit/engine.go
```diff
@@ -193,6 +194,19 @@ type (
 	compiledFunctionStaticData = [][]byte
 )
 
+func (c *compiledFunction) clone(newSourceInstance *wasm.FunctionInstance) *compiledFunction {
+	// Note: we don't need to set finalizer to mummap the code segment since it is
```
mathetake, line 198:
```suggestion
	// Note: we don't need to set finalizer to munmap the code segment since it is
```

PR #460, opened 2022-04-14
Isolates testify to one file, so that it is easier to remove

codefromthecrypt:
This starts the process of removing all dependencies from wazero, by
isolating all assertions we use into a single file. This allows us to
port those assertions as we have time, and when twitchy is gone, the
project literally has no dependencies except go!

### DIFF
--- a/internal/testing/require/require.go
+++ b/internal/testing/require/require.go
@@ -1,0 +1,102 @@
+// Package require includes test assertions that fail the test immediately. This is like to testify, but without a
+// dependency.
+//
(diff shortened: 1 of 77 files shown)


internal/testing/require/require.go
```diff
@@ -0,0 +1,102 @@
+// Package require includes test assertions that fail the test immediately. This is like to testify, but without a
```
codefromthecrypt, line 1: this is the important file

mathetake:
love it

mathetake (approved):

codefromthecrypt:
ps we'll have to be careful until this is complete to not accidentally bring testify back

PR #462, opened 2022-04-14
Removes testify dependency

codefromthecrypt:
This implements our own assertion library so that we can remove the
testify dependency. I've changed a lot of call sites so that we only
have to maintain a minimal amount of assertions.

One tradeoff here besides the code added.. While this does not bring
in a diff dependency, it also doesn't do diffs. In practice, I think this is
ok, but we could consider adding diffs if our IDEs aren't smart enough
later.

### DIFF
--- a/internal/testing/require/require.go
+++ b/internal/testing/require/require.go
@@ -6,72 +6,199 @@
 package require
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
-	"testing"
-
-	"github.com/stretchr/testify/require"
+	"path"
+	"reflect"
+	"runtime"
+	"strings"
+	"unicode"
+	"unicode/utf8"
 )
+
+// TestingT is an interface wrapper of functions used in TestingT
+type TestingT interface {
+	Fatal(args ...interface{})
+}
 
 // TODO: implement, test and document each function without using testify
 
-func Contains(t *testing.T, s interface{}, contains interface{}, msgAndArgs ...interface{}) {
-	require.Contains(t, s, contains, msgAndArgs)
+// Contains fails if `s` does not contain `substr` using strings.Contains.
+//
+//  * formatWithArgs are optional. When the first is a string that contains '%', it is treated like fmt.Sprintf.
+func Contains(t TestingT, s, substr string, formatWithArgs ...interface{}) {
+	if !strings.Contains(s, substr) {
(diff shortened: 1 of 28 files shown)


internal/testing/require/require.go
```diff
@@ -89,14 +101,41 @@ func CapturePanic(panics func()) (err error) {
...
 }
 
-func Zero(t *testing.T, i interface{}, msgAndArgs ...interface{}) {
-	require.Zero(t, i, msgAndArgs)
+func Zero(t TestingT, i interface{}, formatWithArgs ...interface{}) {
+	require.Zero(t, i, formatWithArgs)
+}
+
+// fail tries to treat the formatWithArgs as fmt.Sprintf parameters or joins on space.
+func fail(t TestingT, message string, formatWithArgs ...interface{}) {
```
codefromthecrypt, line 122: ps this is actually a bit smarter (IMHO) than testify as it detects a better message option

internal/testing/require/require.go
```diff
@@ -89,14 +208,137 @@ func CapturePanic(panics func()) (err error) {
...
+	}
+}
+
+// failStack returns the stack leading to the require fail, without test infrastructure.
+//
+// Note: This is similar to assert.CallerInfo in testify
+// Note: This is untested because it is a lot of work to do that. The rationale to punt is this is a test-only internal
+// type which returns optional info. Someone can add tests, but they'd need to do that as an integration test in a
+// different package with something stable line-number-wise.
+func failStack() (fs []string) {
```
codefromthecrypt, line 295: I verified this manually by making another test fail and verifying that the stack was clickable.

codefromthecrypt:
rebased

internal/testing/require/require_test.go
```diff
@@ -50,3 +52,620 @@ func TestCapturePanic(t *testing.T) {
...
+		},
+		{
+			name: "Equal fails on not equal: uint64", // ensure we don't use multi-line output!
+			require: func(t TestingT) {
+				Equal(t, uint64(12), uint64(13))
+			},
+			expectedLog: `expected 12, but was 13`,
+		},
+		{
+			name: "Equal fails on not equal: []byte",
```
codefromthecrypt, line 226: I added long form comparisons for structs and binary while keeping short for normal things like errors, numbers and strings.

mathetake (approved):

PR #466, opened 2022-04-15
Adds import-go example

### LINKED ISSUE #464: How do I wrap Go function and use it as import function in my WASM module?
I have WASM module that imports the `log_func` function. I'd like to wrap it similarly to the wastime-go API. How do we do that in wazero?

```go
func run() error {
	store := wasmtime.NewStore(wasmtime.NewEngine())
	module, err := wasmtime.NewModuleFromFile(store.Engine, "how_old_with_import.wasm")
	if err != nil {
		return err
	}
	item := wasmtime.WrapFunc(store, func(i int32) {
		fmt.Printf("%v\n", i)
	})
	instance, err := wasmtime.NewInstance(store, module, []wasmtime.AsExtern{
		item,
	})
	if err != nil {
		return err
	}
	f := instance.GetFunc(store, "log_how_old")
	_, err = f.Call(store, 2022, 2000)
	if err != nil {
		return err
	}
	return nil
```
```wat
(module
  (func $log (import "imports" "log_func") (param i32))

  (func $how_old (param $year_now i32) (param $year_born i32) (result i32)
    local.get $year_now
    local.get $year_born
    i32.sub)

  (func $log_how_old (param $year_now i32) (param $year_born i32)
    local.get $year_now
    local.get $year_born
    call $how_old
    call $log)

  (export "how_old" (func $how_old))
  (export "log_how_old" (func $log_how_old))
)
```

codefromthecrypt:
This shows how to define, export and import functions written in Go.

Fixes #464

### DIFF
--- a/examples/import-go/age-calculator_test.go
+++ b/examples/import-go/age-calculator_test.go
@@ -1,4 +1,4 @@
-package main
+package age_calculator
 
 import "os"
 
--- a/examples/import-go/age-calculator_test.go
+++ b/examples/import-go/age-calculator_test.go
@@ -1,0 +1,25 @@
+package main
+
+import "os"
+
+// Example_main ensures the following will work:
+//
+//	go build age-calculator.go
+//	./age-calculator 2000
+func Example_main() {
+
+	// Save the old os.Args and replace with our example input.
+	oldArgs := os.Args
+	_ = os.Setenv("CURRENT_YEAR", "2021")
+	os.Args = []string{"age-calculator", "2000"}
+	defer func() {
+		os.Args = oldArgs
+		_ = os.Unsetenv("CURRENT_YEAR")
+	}()
+
+	main()
+
+	// Output:
+	// println >> 21
+	// log_i32 >> 21
+}
--- a/examples/import-go/age-calculator_test.go
+++ b/examples/import-go/age-calculator_test.go
@@ -4,8 +4,7 @@ import "os"
 
 // Example_main ensures the following will work:
 //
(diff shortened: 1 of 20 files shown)


examples/import-go/age-calculator_test.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import "os"
+
+// Example_main ensures the following will work:
+//
+//	go build age-calculator.go
+//	./age-calculator 2000
```
mathetake, line 8:
```suggestion
//	go run age-calculator.go 2000
```

codefromthecrypt:
Note: later I will update this to use a global for the current year instead of a function, just I need to make the text format do that first. I didn't focus too much on that as it wasn't asked for anyway. However, it would be best to show globals in this demo I think.

examples/import-go/age-calculator_test.go
```diff
@@ -0,0 +1,25 @@
+package main
```
mathetake, line 1: I would rename to `age_calculator_test.go`

codefromthecrypt: ok

codefromthecrypt: went for package name `age_calculator` because I still want to be able to call `main`! good idea to use a separate package per example.. I did that to all of them

mathetake: wanna rename others too? (without hyphens)

codefromthecrypt:
thanks for the help polishing!

PR #468, opened 2022-04-15
Changes how examples are tested, and fixes ExitError bug

codefromthecrypt:
Before, we tested the examples/ directory using "ExampleXX", but this is
not ideal because it literally embeds the call to `main` into example
godoc output. This stops doing that for a different infrastructure.

This also makes sure there's a godoc example for both the main package
and wasi, so that people looking at https://pkg.go.dev see something and
also a link to our real examples directory.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,0 +1,43 @@
+package wazero
+
+import (
+	_ "embed"
+	"fmt"
+	"log"
+)
+
+// This is an example of how to use WebAssembly via adding two numbers.
+//
+// See https://github.com/tetratelabs/wazero/tree/main/examples for more examples.
+func Example() {
+	// Create a new WebAssembly Runtime.
+	r := NewRuntime()
+
+	// Add a module to the runtime named "wasm/math" which exports one function "add", implemented in WebAssembly.
+	mod, err := r.InstantiateModuleFromCode([]byte(`(module $wasm/math
+    (func $add (param i32 i32) (result i32)
+        local.get 0
+        local.get 1
+        i32.add
+    )
+    (export "add" (func $add))
+)`))
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer mod.Close()
+
+	// Get a function that can be reused until its module is closed:
+	add := mod.ExportedFunction("add")
+
+	x, y := uint64(1), uint64(2)
+	results, err := add.Call(nil, x, y)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Printf("%s: %d + %d = %d\n", mod.Name(), x, y, results[0])
+
+	// Output:
+	// wasm/math: 1 + 2 = 3
+}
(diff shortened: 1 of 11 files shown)


example_test.go
```diff
@@ -0,0 +1,43 @@
+package wazero
+
+import (
+	_ "embed"
+	"fmt"
+	"log"
+)
+
+// This is an example of how to use WebAssembly via adding two numbers.
```
codefromthecrypt, line 9: Ex this is intentional overlap with the examples directory, only to seed one example into https://pkg.go.dev/github.com/tetratelabs/wazero

wasi/example_test.go
```diff
@@ -1,14 +1,49 @@
...
-var r = wazero.NewRuntime()
+	"github.com/tetratelabs/wazero"
+	"github.com/tetratelabs/wazero/sys"
+)
 
-// Example_InstantiateSnapshotPreview1 shows how to instantiate ModuleSnapshotPreview1, which allows other modules to
-// import functions such as "wasi_snapshot_preview1" "fd_write".
-func Example_instantiateSnapshotPreview1() {
-	wm, _ := InstantiateSnapshotPreview1(r)
+// This is an example of how to use WebAssembly System Interface (WASI) with its simplest function: "proc_exit".
```
codefromthecrypt, line 13: This also has overlap with the examples directory, but we should have it to seed one example into https://pkg.go.dev/github.com/tetratelabs/wazero/wasi

examples/replace-import/replace-import_test.go
```diff
@@ -1,12 +1,16 @@
...
+	"github.com/tetratelabs/wazero/internal/testing/maintester"
+	"github.com/tetratelabs/wazero/internal/testing/require"
+)
 
-	// Output:
-	// module "needs-import" closed with exit_code(255)
+// Test_main ensures the following will work:
+//
+//	go run replace-import.go
+func Test_main(t *testing.T) {
```
codefromthecrypt, line 13: this still tests the same thing, just doesn't pretend it is an "Example" test.

codefromthecrypt (commented):
Notes

wasm.go
```diff
@@ -218,6 +219,9 @@ func (r *runtime) InstantiateModuleWithConfig(code *CompiledCode, config *Module
 			continue
 		}
 		if _, err = start.Call(mod.WithContext(r.ctx)); err != nil {
+			if _, ok := err.(*sys.ExitError); ok {
```
codefromthecrypt, line 222: this fixed the bug

example_test.go
```diff
@@ -0,0 +1,43 @@
+package wazero
+
+import (
+	_ "embed"
```
mathetake, line 4: unused?

mathetake (approved):

example_test.go
codefromthecrypt: good eye!

PR #469, opened 2022-04-15
Make CompileModule actually compile

mathetake:
This commie makes it possible for functions to be compiled before instantiation.
Notably, this adds `CompileModule` method on Engine interface where we pass
`wasm.Module` (which is the decoded module) to engines, and engines compile 
all the module functions  and caches them keyed on `*wasm.Module`.

In order to achieve that, this stops the compiled native code from embedding typeID
which is assigned for all the function types in a store.

### DIFF
--- a/config.go
+++ b/config.go
@@ -148,15 +148,35 @@ func (c *RuntimeConfig) WithFeatureMultiValue(enabled bool) *RuntimeConfig {
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#semantic-phases%E2%91%A0
 type CompiledCode struct {
 	module *wasm.Module
-	// compiledEngine holds the engine on which this `module` is compiled.
-	compiledEngine wasm.Engine
+	// cachedEngines maps wasm.Engine to []*wasm.Module which originate from this .module.
+	// This is necessary to track which engine caches *Module where latter might be different
+	// from .module via import replacement config (ModuleConfig.WithImport).
+	cachedEngines map[wasm.Engine]map[*wasm.Module]struct{}
 }
 
 // Close releases all the allocated resources for this CompiledCode.
 //
 // Note: it is safe to call Close while having outstanding calls from Modules instantiated from this *CompiledCode.
 func (c *CompiledCode) Close() {
-	c.compiledEngine.DeleteCompiledModule(c.module)
+	for engine, modules := range c.cachedEngines {
+		for module := range modules {
+			engine.DeleteCompiledModule(module)
+		}
+	}
+}
+
+func (c *CompiledCode) addCacheEntry(module *wasm.Module, engine wasm.Engine) {
+	if c.cachedEngines == nil {
+		c.cachedEngines = map[wasm.Engine]map[*wasm.Module]struct{}{}
+	}
+
+	cache, ok := c.cachedEngines[engine]
+	if !ok {
+		cache = map[*wasm.Module]struct{}{}
+		c.cachedEngines[engine] = cache
+	}
+
+	cache[module] = struct{}{}
 }
 
 // ModuleConfig configures resources needed by functions that have low-level interactions with the host operating system.
--- a/config.go
+++ b/config.go
@@ -148,35 +148,15 @@ func (c *RuntimeConfig) WithFeatureMultiValue(enabled bool) *RuntimeConfig {
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#semantic-phases%E2%91%A0
 type CompiledCode struct {
 	module *wasm.Module
-	// cachedEngines maps wasm.Engine to []*wasm.Module which originate from this .module.
-	// This is necessary to track which engine caches *Module where latter might be different
-	// from .module via import replacement config (ModuleConfig.WithImport).
-	cachedEngines map[wasm.Engine]map[*wasm.Module]struct{}
+	// compiledEngine holds the engine on which this `module` is compiled.
+	compiledEngine wasm.Engine
 }
 
 // Close releases all the allocated resources for this CompiledCode.
 //
 // Note: it is safe to call Close while having outstanding calls from Modules instantiated from this *CompiledCode.
 func (c *CompiledCode) Close() {
-	for engine, modules := range c.cachedEngines {
-		for module := range modules {
-			engine.DeleteCompiledModule(module)
-		}
-	}
-}
-
-func (c *CompiledCode) addCacheEntry(module *wasm.Module, engine wasm.Engine) {
-	if c.cachedEngines == nil {
-		c.cachedEngines = map[wasm.Engine]map[*wasm.Module]struct{}{}
-	}
-
-	cache, ok := c.cachedEngines[engine]
-	if !ok {
-		cache = map[*wasm.Module]struct{}{}
-		c.cachedEngines[engine] = cache
-	}
-
-	cache[module] = struct{}{}
+	c.compiledEngine.DeleteCompiledModule(c.module)
 }
 
 // ModuleConfig configures resources needed by functions that have low-level interactions with the host operating system.
--- a/config.go
+++ b/config.go
@@ -155,6 +155,8 @@ type CompiledCode struct {
 }
 
 // Close releases all the allocated resources for this CompiledCode.
+//
+// Note: it is safe to call Close while having outstanding calls from Modules instantiated from this *CompiledCode.
 func (c *CompiledCode) Close() {
 	for engine, modules := range c.cachedEngines {
 		for module := range modules {
--- a/config.go
+++ b/config.go
@@ -158,7 +158,7 @@ type CompiledCode struct {
 func (c *CompiledCode) Close() {
 	for engine, modules := range c.cachedEngines {
 		for module := range modules {
-			engine.ReleaseCompilationCache(module)
+			engine.DeleteCompiledModule(module)
 		}
 	}
 }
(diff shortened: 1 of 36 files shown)


internal/wasm/host.go
```diff
@@ -67,6 +67,10 @@ func NewHostModule(
 	return
 }
 
+func (m *Module) IsHostMdule() bool {
```
anuraaga, line 70:
```suggestion
func (m *Module) IsHostModule() bool {
```

internal/wasm/interpreter/interpreter.go
```diff
@@ -161,22 +140,25 @@ type callFrame struct {
...
 type compiledFunction struct {
-	moduleEngine *moduleEngine
-	source       *wasm.FunctionInstance
-	body         []*interpreterOp
-	hostFn       *reflect.Value
+	body   []*interpreterOp
+	hostFn *reflect.Value
+}
+
+type compiledFunctionInstance struct {
```
codefromthecrypt, line 151: It isn't obvious why there are two types here, can you add a comment to each (also in JIT)?

note: least favorite naming suffix Is instance. I see why you are doing it because we have this Instance suffix in other types due to literal translation of spec names. It is like saying I have a string instance.

internal/wasm/interpreter/interpreter.go
```diff
@@ -188,57 +170,58 @@ type interpreterOp struct {
...
 }
 
+// CompileModule implements the same method as documented on wasm.Engine.
+func (e *engine) CompileModule(module *wasm.Module) error {
+	if _, ok := e.getCompiledFunctions(module); ok { // cache hit!
+		return nil
+	}
+
+	funcs := make([]*compiledFunction, 0, len(module.FunctionSection))
+	if module.IsHostMdule() {
```
codefromthecrypt, line 180: maybe comment like "host modules are already compiled, so there's nothing to do... wasm functions need to be translated from wasm binary format into wazero's intermediate representation (wazeroir).

internal/wasm/interpreter/interpreter.go
```diff
@@ -188,57 +170,58 @@ type interpreterOp struct {
...
-			compiled.moduleEngine = me
-			me.compiledFunctions = append(me.compiledFunctions, compiled)
-
-			// Add the compiled function to the store-wide engine as well so that
-			// the future importing module can refer the function instance.
-			e.addCompiledFunction(f, compiled)
-		}
+	compiledFunctions, ok := e.getCompiledFunctions(module)
+	if !ok {
+		return nil, fmt.Errorf("source module for %s must be compiled before instantiation", name)
```
codefromthecrypt, line 218: I think we don't need to introduce source since it is the same type still. Maybe since this is a bug on our side, say "BUG: didn't call CompileModule on module %s", name

internal/wasm/jit/engine.go
```diff
@@ -159,12 +159,12 @@ type (
...
 		returnStackBasePointer uint64
 		// Set when making function call to this function frame.
-		compiledFunction *compiledFunction
+		compiledFunction *compiledFunctionInstance
 		// _ is a necessary padding to make the size of callFrame struct a power of 2.
 		_ [8]byte
 	}
 
-	compiledFunction struct {
+	compiledFunctionInstance struct {
```
codefromthecrypt, line 167: same thing on the docs on compiledFunction vs compiledFunctionInstance as it is a little confusing

internal/wasm/jit/engine.go
```diff
@@ -359,137 +392,112 @@ func releaseCompiledFunction(compiledFn *compiledFunction) {
...
-			} else {
-				compiled, err = compileHostFunction(f)
+	if module.IsHostMdule() {
+		for funcIndex := range module.HostFunctionSection {
+			compiled, err := compileHostFunction(module.TypeSection[module.FunctionSection[funcIndex]])
+			if err != nil {
+				return fmt.Errorf("function[%d/%d] %w", funcIndex, len(module.FunctionSection)-1, err)
 			}
+
+			// As this uses mmap, we need a finalizer in case moduleEngine.Close was never called. Regardless, we need a
```
codefromthecrypt, line 420: drift as moduleEngine.Close is dead

internal/wazeroir/compiler.go
```diff
@@ -114,8 +114,17 @@ type compiler struct {
 		depth int
 	}
 	pc     uint64
-	f      *wasm.FunctionInstance
 	result CompilationResult
+
+	// Function information
+	body       []byte
```
codefromthecrypt, line 120: body is the WebAssembly binary format (since we also are writing native code it might help keep head straight). Also, anything here correlated with wasm.Module would be best to call out. particularly below here are types ( complete type section I guess) and funcs (the whole index namespace or just module-defined?)

these will help folks who haven't memorized the spec, yet ;)

If you already documented it in the same package, you can say "// XX holds the same data documented by CompilationResult.XX", so that someone can click navigate to the why.

wasm.go
```diff
@@ -207,6 +210,12 @@ func (r *runtime) InstantiateModuleWithConfig(code *CompiledCode, config *Module
 	}
 
 	module := config.replaceImports(code.module)
+	if module != code.module {
+		// IF the module changed, we have to Compile again before instantiation.
```
codefromthecrypt, line 214: only question is why we have to if we are only changing extern string names.. are the strings embedded in the code somehow?

```suggestion
		// If replacing imports had an effect, the module changed, so we have to recompile it.
```

codefromthecrypt (approved):
Thanks for doing this. Should we have a different dimension in our bench about re-instantiate a module? I think wapc-go should benefit a lot.

wasm.go
codefromthecrypt: maybe we might needed a "TODO: implement re-linking" as probably this is more about where the imported functions point to vs re-compiing the code sections, right?

internal/wasm/interpreter/interpreter.go
mathetake: so renamed to `compiledFunction` and `compiledFunctionInstance` to simply `code` and `function`.

config.go
```diff
@@ -158,7 +158,7 @@ type CompiledCode struct {
 func (c *CompiledCode) Close() {
```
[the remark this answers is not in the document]
mathetake: https://github.com/tetratelabs/wazero/pull/469/commits/90242f944d95c0268e324387f06775d03c8ede52 Added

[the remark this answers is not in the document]
mathetake: will add the test case later

internal/testing/enginetest/enginetest.go
```diff
@@ -220,9 +309,7 @@ func RunTestModuleEngine_Call_HostFn(t *testing.T, et EngineTester) {
 
 	e := et.NewEngine(wasm.Features20191205)
 
-	imported, importedMe, importing, importingMe := setupCallTests(t, e)
-	defer importingMe.Close()
```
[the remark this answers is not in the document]
mathetake: https://github.com/tetratelabs/wazero/pull/469/commits/1bdb3edb41bc28d8c8bcbb6c22732d35005ccb4c done, thanks!

internal/wasm/interpreter/interpreter.go
mathetake: well it is actually not a bug on ourside as some one might accidentally try to call `InstantiateModule` _after_ calling `CompiledCode.Close`

wasm.go
mathetake: good question.. the compiled native code itself is not coupled with names, but the function instances will be different between with and without re-exports

mathetake: I think we should move this replacing config into `CompileModule`

mathetake: https://github.com/tetratelabs/wazero/pull/469/commits/478b41b3d485e669cd6cad918ef3bc4ee2701b1c added todo

mathetake: oh well yeah it should be possible to reuse the code actually.. lemme fix

mathetake: https://github.com/tetratelabs/wazero/pull/469/commits/42965e243879699d1e3b6a26578959231918f973 fixed!

mathetake: hmm realized that we use `*wasm.Module` as a cache key so this makes it impossible to share code.. let me revert it and resolve it in another pr by introducing `ID` for wasm.Module

internal/wasm/engine.go
```diff
@@ -16,8 +19,8 @@ type Engine interface {
 	// due to reasons such as out-of-bounds.
 	NewModuleEngine(name string, module *Module, importedFunctions, moduleFunctions []*FunctionInstance, table *TableInstance, tableInit map[Index]Index) (ModuleEngine, error)
 
-	// ReleaseCompilationCache releases compilation caches for the given module (source).
-	ReleaseCompilationCache(module *Module)
+	// DeleteCompiledModule releases compilation caches for the given module (source).
```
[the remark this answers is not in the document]
mathetake: https://github.com/tetratelabs/wazero/pull/469/commits/ac19a19682d7fd01df41e6c0bf0203cbc05a4a90 added while in use close tests!

config.go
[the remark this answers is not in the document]
mathetake: https://github.com/tetratelabs/wazero/pull/469/commits/ac19a19682d7fd01df41e6c0bf0203cbc05a4a90 added cases as well!

mathetake:
ok applied comments, I'm landing the change.. if something comes up I will do followups! Thank you guys!

PR #473, opened 2022-04-16
wasi: flattens interface into struct functions

codefromthecrypt:
This flattens the unexported interface as there's no plans to export it.
That these are separate knowing that is not helpful.

### DIFF
--- a/wasi/wasi.go
+++ b/wasi/wasi.go
@@ -18,11 +18,9 @@ import (
 	"github.com/tetratelabs/wazero/internal/wasm"
 )
 
-const (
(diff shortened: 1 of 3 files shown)


wasi/wasi.go
```diff
@@ -33,18 +31,6 @@ func InstantiateSnapshotPreview1(r wazero.Runtime) (api.Module, error) {
 }
 
 const (
-	// functionStart is the name of the nullary function a module must export if it is a WASI Command Module.
```
codefromthecrypt, line 36: this diff changed no behavior, just moved the docs to the struct types and removed the interface

mathetake (approved):

PR #474, opened 2022-04-16
Add .gitignore

anuraaga:
I think I've seen @codefromthecrypt using GoLand too so hopefully this is broadly useful.

I copied in GitHub's [standard template](https://github.com/github/gitignore/blob/main/Go.gitignore) and added `.idea` to it


mathetake (approved):
🕺

PR #475, opened 2022-04-17
Adds allocation examples in Rust and TinyGo

### LINKED ISSUE #471: Add tinygo and rust examples that include memory allocation
Hi, thanks for the awesome project! I was taking a look over examples, but I didn't find an example of passing complex values. I'm sorry if I missed something, but do you have something like this?
https://wasmedge.org/book/en/embed/go/memory.html#pass-complex-parameters-to-wasm-functions

codefromthecrypt:
This adds memory allocation examples for Rust and TinyGo. e.g. How to pass strings in and out of WebAssembly functions.

```bash
$ go run greet.go wazero
wasm >> Hello, wazero!
go >> Hello, wazero!
```

* rust - Calls Wasm built with `cargo build --release --target wasm32-unknown-unknown`
* tinygo - Calls Wasm built with `tinygo build -o X.wasm -scheduler=none --no-debug -target=wasi X.go`

Note: The above are language-specific because there's no WebAssembly
specification for memory allocation. This mean exported functions are different
and how pointers map to parameters are different. That said, the examples are
as close to the same as possible with subtle differences and gotchas mentioned
in the respective README files.

Fixes #471


codefromthecrypt:
Need to all allocation/deallocation from Go to this and pull out the null-terminated strings thing into a separate PR

api/wasm.go
```diff
@@ -199,7 +199,11 @@ type Memory interface {
 	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#-hrefsyntax-instr-memorymathsfmemorysize%E2%91%A0
 	Size() uint32
 
-	// ReadByte reads a single byte from the underlying buffer at the offset in or returns false if out of range.
+	// IndexByte returns the index of the first instance of c in the underlying buffer at the offset or returns false if
+	// not found or out of range.
+	IndexByte(offset uint32, c byte) (uint32, bool)
```
codefromthecrypt, line 204: TODO: add an example codeblock about how to use this to print a null-terminated string

examples/rust/testdata/src/lib.rs
```diff
@@ -0,0 +1,38 @@
+extern crate alloc;
+extern crate wee_alloc;
+
+use alloc::vec::Vec;
+use core::mem;
+
+#[global_allocator]
```
codefromthecrypt, line 7: TODO: document all this, though I've never seen anyone do that :)

examples/rust/testdata/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+# wee_aloc is a WebAssembly optimized allocator, which is needed to use non-numeric types like strings.
+# See https://docs.rs/wee_alloc/latest/wee_alloc/
+wee_alloc = "0.4.5"
+
+# Below settings dramatically reduce wasm output size
+# See https://rustwasm.github.io/book/reference/code-size.html#optimizing-builds-for-code-sizewasm-opt -Oz -o
+# See https://doc.rust-lang.org/cargo/reference/profiles.html#codegen-units
+[profile.release]
+opt-level = "z"
+lto = true
```
codefromthecrypt, line 20: PS `lto = true` has the biggest impact in wasm size reduction

codefromthecrypt:
ok Rust is pretty ok I think. I'll add the TinyGo next before a data structure example

examples/rust/testdata/src/allocator.rs
```diff
@@ -0,0 +1,41 @@
...
+
+/// Allocates size bytes in linear memory and returns the offset where it starts.
+///
+/// In WebAssembly, the corresponding import will be:
+/// ```
+/// (import "xxx" "alloc" (func (param $size i32) (result i32)))
+///```
+#[no_mangle]
+pub extern "C" fn alloc(size: usize) -> *mut c_char {
+    // Allocate the amount of bytes we need, which is 1 more for the NUL suffix.
```
anuraaga, line 20:
```suggestion
    // Allocate the amount of bytes we need, which is 1 more than the requested size for the NUL suffix.
```

examples/rust/testdata/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Says hello to the name indicated by the pointer.
+///
+/// In WebAssembly, the corresponding import will be:
+/// ```
+/// (import "xxx" "say_hello" (func (param $ptr i32)))
+///```
+#[no_mangle]
+pub unsafe extern "C" fn say_hello(ptr: *mut c_char) {
+    // The input string may be re-used many times. Hence, we borrow a reference to it with
+    // CStr::from_ptr instead of taking ownership with CString::from_raw (which avails it for
```
anuraaga, line 31: Scary stuff

examples/rust/testdata/src/allocator.rs
anuraaga: Just want to confirm that we want to do the allocation within Rust, as opposed to passing in memory allocated in Go.

codefromthecrypt: it is a good question. the main concern is who controls allocation of the wasm memory.

Let's say we passed an arbitrary offset, like 8. When the rust-defined wasm initializes, it might have written to that area already (maybe some static data). We could choose a high number instead like 4000, but that also might collide with data allocated at runtime by rust-generated wasm.

The main thing we buy with using rust's allocator is that the responsibility of "picking offsets so that they don't collide" is moved there. I think this is a general win regardless of rust vs something else like TinyGo, as the problem would be the same. Basically, if the internal allocation impl isn't aware in any way of allocations (because we picked randomly outside with Go), it can easily collide with it.

The flip side is not all wasm is defined by a runtime that includes garbage collection. Ex if defining wasm with the text format, you have to manually track offsets or make a wasm library that does it and call that to manage allocation. Until things consistently use alloc/dealloc functions this sort of "library bias" will exist, and probably for a while considering garbage collection is "phase 2" https://github.com/WebAssembly/proposals#phase-2---proposed-spec-text-available-cg--wg

codefromthecrypt:
PS one thing noted in README etc is that we aren't trying to artificially make all types of memory patterns explained.

Ex CString just because Rust (and CGO) use it. Go code doesn't think in terms of CString, so our Go example shouldn't use it.

That said, I intentionally didn't use CString in Rust because doing so leaves the end of the byte range as a search function. That would make the example not really useful if you are trying to figure out how to pass back and forth other serialized data which isn't NUL terminated. Moreover, rust has a book on this stuff, so we don't need to attempt to replicate that.

examples/allocation/rust/testdata/src/lib.rs
```diff
@@ -0,0 +1,84 @@
+extern crate alloc;
```
codefromthecrypt, line 1: rust

examples/allocation/tinygo/testdata/hello.go
```diff
@@ -0,0 +1,42 @@
+package main
```
codefromthecrypt, line 1: tinygo

codefromthecrypt (commented):
pointers to the source that generates wasm

mathetake (commented):
Needs to update examples/README.md as well?

codefromthecrypt:
yeah I'll do that. Indeed I forgot

mathetake:
I'm still thinking about if we can know where the binary originates from before hand, how about having one single host code where we have if block for each language specific logics (finding allocator/deallocator)

codefromthecrypt:
> I'm still thinking about if we can know where the binary originates from before hand, how about having one single host code where we have if block for each language specific logics (finding allocator/deallocator)

hmm isn't this doing that quite explicitly both in directory and the link on the helloWasm variable?

I don't think mixing and matching rust and tinygo is a good idea frankly, it is wildly confusing what each do.

examples/allocation/rust/README.md
```diff
@@ -0,0 +1,22 @@
...
+in Rust, built with `cargo build --release --target wasm32-unknown-unknown`
+
+Ex.
+```bash
+$ go run say-hello.go wazero
+Hello, wazero!
+```
+
+Under the covers, [lib.rs](testdata/src/lib.rs) does a few things of interest:
+* Uses a WebAssembly-tuned memory allocator: [we_alloc](https://github.com/rustwasm/wee_alloc).
```
mathetake, line 13:
```suggestion
* Uses a WebAssembly-tuned memory allocator: [wee_alloc](https://github.com/rustwasm/wee_alloc).
```

mathetake (approved):

codefromthecrypt (commented):
ok I added an example to read back a value allocated inside wasm. This is very different TinyGo vs Rust, so I felt it important to add, especially as I've been paying attention and still lost hours on it!

PTAL @mathetake

examples/allocation/rust/testdata/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+/// and calls [`greet`].
+///
+/// Note: The input parameters were returned by [`allocate`]. This is not an
+/// ownership transfer, so the inputs can be reused after this call.
+#[cfg_attr(
+all(target_arch = "wasm32", target_os = "unknown"),
+export_name = "greet"
+)]
+#[no_mangle]
+pub unsafe extern "C" fn _greet(ptr: u32, len: u32) {
```
codefromthecrypt, line 48: Note: I used direct wasm types here as I noticed rust will sometimes compile different arity. For example, if you return multiple values it switches them into out params in wasm. Having this defined the way it looks in wasm is helpful to me as it prevents odd signature changes.

mathetake (approved):

PR #478, opened 2022-04-18
Use same CompiledCode for import replacement

mathetake:
This commit allows CompiledCode to be re-used regardless of 
the existence of import replacement configs for instantiation. 

In order to achieve this, we introduce `ModuleID`, which is sha256
checksum calculated on source bytes, as a key for module compilation 
cache. Previously, we used`*wasm.Module` as keys for caches which 
differ before/after import replacement.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -256,9 +256,7 @@ func (b *moduleBuilder) Build() (*CompiledCode, error) {
 		return nil, err
 	}
 
-	ret := &CompiledCode{module: module}
-	ret.addCacheEntry(module, b.r.store.Engine)
-	return &CompiledCode{module: module}, nil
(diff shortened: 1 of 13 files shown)


builder.go
```diff
@@ -256,9 +256,7 @@ func (b *moduleBuilder) Build() (*CompiledCode, error) {
 		return nil, err
 	}
 
-	ret := &CompiledCode{module: module}
-	ret.addCacheEntry(module, b.r.store.Engine)
-	return &CompiledCode{module: module}, nil
```
mathetake, line 261: 😄

codefromthecrypt (approved):
slick

PR #479, opened 2022-04-19
Centralizes logic used to call go-defined functions

codefromthecrypt:
This centralizes logic that gets params and results in and out of
Go-defined functions. This allows us to refactor and optimize easier as
well ensure value coercion is the same regardless of the engine type.

This also renames `ModuleContext` back to `CallContext` and
documents it.

### DIFF
--- a/internal/wasm/gofunc.go
+++ b/internal/wasm/gofunc.go
@@ -66,6 +66,9 @@ func PopGoFuncParams(f *FunctionInstance, popParam func() uint64) []uint64 {
 //
 // Note: the popper intentionally doesn't return bool or error because the caller's stack depth is trusted.
 func PopValues(count int, popper func() uint64) []uint64 {
+	if count == 0 {
+		return nil
+	}
 	params := make([]uint64, count)
 	for i := count - 1; i >= 0; i-- {
 		params[i] = popper()
--- a/internal/wasm/gofunc.go
+++ b/internal/wasm/gofunc.go
@@ -3,6 +3,7 @@ package wasm
 import (
 	"context"
 	"fmt"
+	"math"
 	"reflect"
 
 	"github.com/tetratelabs/wazero/api"
@@ -30,8 +31,8 @@ var moduleType = reflect.TypeOf((*api.Module)(nil)).Elem()
 var goContextType = reflect.TypeOf((*context.Context)(nil)).Elem()
 var errorType = reflect.TypeOf((*error)(nil)).Elem()
 
-// GetHostFunctionCallContextValue returns a reflect.Value for a context param[0], or nil if there isn't one.
-func GetHostFunctionCallContextValue(fk FunctionKind, ctx *ModuleContext) *reflect.Value {
+// getGoFuncCallContextValue returns a reflect.Value for a context param[0], or nil if there isn't one.
+func getGoFuncCallContextValue(fk FunctionKind, ctx *CallContext) *reflect.Value {
 	switch fk {
 	case FunctionKindGoNoContext: // no special param zero
 	case FunctionKindGoContext:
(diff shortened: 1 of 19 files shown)


internal/wasm/interpreter/interpreter.go
```diff
@@ -549,76 +550,31 @@ func (me *moduleEngine) Call(m *wasm.ModuleContext, f *wasm.FunctionInstance, pa
...
-	}
-	results = make([]uint64, len(f.Type.Results))
-	for i := range results {
-		results[len(results)-1-i] = ce.pop()
+		results = ce.callGoFunc(m, compiled, params)
 	}
 	return
 }
 
-func (ce *callEngine) callHostFunc(ctx *wasm.ModuleContext, f *function) {
```
codefromthecrypt, line 567: @anuraaga this should make your trace integration less blinding :D

internal/wasm/gofunc.go
```diff
@@ -46,6 +47,94 @@ func GetHostFunctionCallContextValue(fk FunctionKind, ctx *ModuleContext) *refle
...
+	return params
+}
+
+// CallGoFunc executes the FunctionInstance.GoFunc by converting params to Go types. The results of the function call
+// are converted back to api.ValueType.
+//
+// * callCtx is passed to the host function as a first argument.
+//
+// Note: ctx must use the caller's memory, which might be different from the defining module on an imported function.
+func CallGoFunc(callCtx *CallContext, f *FunctionInstance, params []uint64) []uint64 {
```
codefromthecrypt, line 82: this is the new centralized function. Before, it was copy/pasted for JIT and Interpreter. While not doing anything too fancy now, you can imagine we can extract a template-based call invoker here one day if we wanted.

internal/wasm/interpreter/interpreter.go
```diff
@@ -549,76 +550,31 @@ func (me *moduleEngine) Call(m *wasm.ModuleContext, f *wasm.FunctionInstance, pa
 		}
 	}()
 
-	for _, param := range params {
-		ce.push(param)
-	}
 	if f.Kind == wasm.FunctionKindWasm {
+		for _, param := range params {
```
codefromthecrypt, line 554: I chose to not attempt to normalize calling wasm as it Is less similar between JIT and interpreter. However, someone can try later and maybe it isn't worthless to attempt.

internal/wasm/gofunc.go
```diff
@@ -46,6 +47,97 @@ func GetHostFunctionCallContextValue(fk FunctionKind, ctx *ModuleContext) *refle
 	return nil
 }
 
+// PopGoFuncParams pops the correct number of parameters off the stack into a parameter slice for use in CallGoFunc
```
codefromthecrypt, line 50: besides tidying up the pop loop, the main change here to logic is to avoid allocating empty slices when there are no inputs or outputs.

internal/wasm/gofunc_test.go
```diff
@@ -133,3 +132,218 @@ func TestGetFunctionTypeErrors(t *testing.T) {
 		})
 	}
 }
+
+// stack simulates the value stack in a way easy to be tested.
```
codefromthecrypt, line 136: all the tests here are new as we never unit tested this logic before (it was covered, but tested indirectly)

internal/wasm/jit/engine.go
```diff
@@ -85,9 +84,9 @@ type (
 		callFrameStackPointer uint64
 	}
 
-	// moduleContext holds the per-function call specific module information.
+	// callContext holds the per-function call specific module information.
 	// This is subject to be manipulated from JITed native code whenever we make function calls.
-	moduleContext struct {
+	callContext struct {
```
mathetake, line 89: oh well this is not exactly callContex... moduleContext holds per-module instance specific information, and change whenever enters/exits  new functions.. so I would revert this

mathetake: and this is nothing to do with `wasm.ModuleContext` or `wasm.CallCotext` (sorry for confusing)

internal/wasm/jit/engine.go
```diff
@@ -739,14 +669,20 @@ jitentry:
 		case jitCallStatusCodeReturned:
 			// Meaning that all the function frames above the previous call frame stack pointer are executed.
 		case jitCallStatusCodeCallHostFunction:
-			calleeHostFunction := ce.callFrameTop().function.source
+			calleeHostFunction := ce.callFrameTop().function
 			// Not "callFrameTop" but take the below of peek with "callFrameAt(1)" as the top frame is for host function,
 			// but when making host function calls, we need to pass the memory instance of host function caller.
 			callercode := ce.callFrameAt(1).function
```
mathetake, line 675:
```suggestion
			callerFunction := ce.callFrameAt(1).function
```

internal/wasm/jit/engine.go
codefromthecrypt: whoops. refactor tool fail :D will revert!

mathetake (approved):

mathetake:
like it

internal/wasm/jit/engine.go
codefromthecrypt: ok rolled that back. thanks for noticing

codefromthecrypt:

PR #481, opened 2022-04-19
Adds random module generator for fuzzing the compiler

mathetake:
This commit adds the random module generator for compiler testing.
Notably, it only generates valid and (supposedly) compilable Wasm modules.
Generating such random sources/programs and using them for testing or 
fuzzing compilers is the widely known way of securing and hardening them [1,2].

This commit only creates random modules where functions do not have
any meaningful body. That will be resolved in a subsequent commit
as the implementation for generating random function body could be
huge.

Note that the implementation has already found a bug #485, and can be
easily integrated with the [fuzzing feature of Go 1.18](https://tip.golang.org/doc/fuzz/):

```go
func FuzzCompiler(f *testing.F) {
	r := wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfig().
		WithFeatureMultiValue(true))
	f.Fuzz(func(t *testing.T, seed []byte, numTypes, numFunctions, numImports, numExports, numGlobals, numElements, numData uint32, needStartSection bool) {
		// Generate a random WebAssembly module.
		m := Gen(seed, wasm.FeaturesFinished, numTypes, numFunctions, numImports, numExports, numGlobals, numElements, numData, needStartSection)
		// Encode the generated module (*wasm.Module) as binary.
		bin := binary.EncodeModule(m)
		// Pass the generated binary into our compilers.
		_, err := r.CompileModule(context.Background(), bin)
		require.NoError(t, err)
	})
}

```

[1] https://dl.acm.org/doi/fullHtml/10.1145/3363562
[2] https://en.wikipedia.org/wiki/Compiler_correctness

### DIFF
--- a/internal/modgen/modgen.go
+++ b/internal/modgen/modgen.go
@@ -389,6 +389,9 @@ func (g *generator) dataSection() {
 		}
 
 		init := make([]byte, g.nextRandom().Intn(min-offset))
+		if len(init) == 0 {
+			continue
+		}
 		_, err := g.nextRandom().Read(init)
 		if err != nil {
 			panic(err)
--- a/internal/modgen/modgen.go
+++ b/internal/modgen/modgen.go
@@ -293,6 +293,7 @@ func (g *generator) exportSection() {
 	numExports := g.nextRandom().Intn(g.size)
 	for i := 0; i < numExports; i++ {
 		target := possibleExports[g.nextRandom().Intn(len(possibleExports))]
+
 		g.m.ExportSection = append(g.m.ExportSection, &wasm.Export{
 			Type:  target.Type,
 			Index: target.Index,
--- a/internal/modgen/modgen.go
+++ b/internal/modgen/modgen.go
@@ -243,12 +243,18 @@ func (g *generator) newConstExpr() (*wasm.ConstantExpression, wasm.ValueType) {
 	case 2:
 		opcode = wasm.OpcodeF32Const
 		data = make([]byte, 4)
-		g.nextRandom().Read(data)
+		_, err := g.nextRandom().Read(data)
+		if err != nil {
+			panic(err)
+		}
 		valueType = wasm.ValueTypeF32
 	case 3:
 		opcode = wasm.OpcodeF64Const
 		data = make([]byte, 8)
-		g.nextRandom().Read(data)
+		_, err := g.nextRandom().Read(data)
+		if err != nil {
+			panic(err)
+		}
 		valueType = wasm.ValueTypeF64
 	case 4:
 		opcode = wasm.OpcodeGlobalGet
--- a/internal/modgen/modgen.go
+++ b/internal/modgen/modgen.go
@@ -375,11 +375,11 @@ func (g *generator) dataSection() {
 		panic("BUG:" + err.Error())
 	}
 
-	min := int(mem.Min * wasm.MemoryPageSize)
-	if mem == nil || min == 0 {
+	if mem == nil || mem.Min == 0 {
 		return
 	}
 
+	min := int(mem.Min * wasm.MemoryPageSize)
 	dataSectionSize := g.nextRandom().Intn(g.size)
 	for i := 0; i < dataSectionSize; i++ {
 		offset := g.nextRandom().Intn(min)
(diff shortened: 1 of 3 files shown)


mathetake:
ok the left is to adds tests for dataSection generator.

internal/modgen/modgen.go
```diff
@@ -0,0 +1,420 @@
...
+	// size holds the original size of the seed.
+	size int
+
+	// m is the resulting module.
+	m *wasm.Module
+}
+
+// random is the interface over methods of rand.Rand which are used by our generator.
+// This is only for testing the generator implmenetation.
+
```
codefromthecrypt, line 56:
```suggestion
```

internal/modgen/modgen.go
```diff
@@ -0,0 +1,420 @@
...
+	g.globalSection()
+	g.exportSection()
+	g.startSection()
+	g.elementSection()
+	g.codeSection()
+	g.dataSection()
+	return g.m
+}
+
+func (g *generator) typeSection() {
```
codefromthecrypt, line 87: I think these functions should have "gen" prefix to match "gen()" which does everything.

"getTypeSection creates a random number of types each with a random number of parameters and results."

internal/modgen/modgen.go
```diff
@@ -0,0 +1,420 @@
...
+		ret = wasm.ValueTypeF32
+	case 3:
+		ret = wasm.ValueTypeF64
+	default:
+		panic("BUG")
+	}
+	return
+}
+
+func (g *generator) importSection() {
```
codefromthecrypt, line 122: maybe document that this will always generate a or import a table and memory (vs randomly choose to import or define one of each). also it seems order dependent, so maybe say what must happen before calling this.

internal/modgen/modgen.go
```diff
@@ -0,0 +1,420 @@
...
+	Read(p []byte) (n int, err error)
+}
+
+func (g *generator) nextRandom() (ret random) {
+	ret = g.rands[g.nextRandIndex]
+	g.nextRandIndex = (g.nextRandIndex + 1) % len(g.rands)
+	return
+}
+
+func (g *generator) gen() *wasm.Module {
```
codefromthecrypt, line 71: docs can say punt to individual sections for details on what's generated.

each should summarize what's likely to be here. Ex will there always be a start function? will there always be memory. will there be a name section?

internal/modgen/modgen.go
```diff
@@ -0,0 +1,420 @@
...
+//
+// TODO: make it possible to explicitly control the size of each section.
+// TODO: have an option to have whether or not the result has imports.
+//
+// Note: "pseudo" here means the determinism of the generated results,
+// e.g. giving same seed returns exactly the same module for
+// the same code base in Gen.
+//
+// Note: this is only used for testing wazero runtime.
+func Gen(seed []byte) *wasm.Module {
```
codefromthecrypt, line 27: gen should take a features thing I think, so that it can control which features it will generate.

internal/modgen/modgen.go
```diff
@@ -0,0 +1,420 @@
...
+	"fmt"
+	"math"
+	"math/rand"
+	"strconv"
+
+	"github.com/tetratelabs/wazero/internal/leb128"
+	"github.com/tetratelabs/wazero/internal/wasm"
+)
+
+// Gen generates a pseudo random compilable module based on `seed`. The size of
```
codefromthecrypt, line 15: I think the docs are better spent on what's generated vs the relationship between the size and the seed. I suspect we'll pull out any indirect relationship between seed and size at some point, and meanwhile knowing what this results in could be a better primary choice to talk about.

Ex. the seed's most interesting and important thing is that it allows deterministic generation. I would even go so far to say not random, but seed-deterministic maybe

codefromthecrypt (approved):
Code looks good, only doc and style notes. Exciting!

I know we can't currently do a go 1.18 dependent integration, or haven't figured it out yet heh, but this will be great.

mathetake:
changed Gen signature so that it accepts not only Feature Flags but also it accepts the explicit size of each section (which makes it possible for fuzzer to work better).

Issue #451, opened 2022-04-11
Add hooks to allow trace logging


codefromthecrypt:
Even if only in interpreter mode, it would be helpful for someone to be able to trace by logging or call graph. This could be initially limited to entry-exit on host function boundaries, and possibly extended to allow tracing of each `call` or `call_indirect` even if solely implemented in wasm.

The hook should:
* have a context.Context parameter (allows propagation of a correlation ID)
* have a flag about the type of invocation it is (call root, wasm or host, indirect or not)
* intercept around a function call (allows timing and value captures, as well mutating the context)

The hook should be allocated up-front with the following parameters:
* the function debug name (module.function name)
* the function signature (ideally parameter names)

The hook should not:
* have a reference to something that can hold up GC
* include any 3rd party library dependency

To ensure implementors can make more robust implementation, we can have a proof of concept in the test directory showing how it can be used in basic form.

Particularly, this proof of concept could on each call to an exported function, add the following k/v to messages outputted to console.
* correlation ID scoped to the originating function call (random hex 64 or 128bit.. doesn't matter)
* function DebugName (module.function name)
* monotonic time

Note: We don't need to do anything fancy like span IDs in the proof of concept as nothing is async in wasm function invocation.

With this in place a more featureful logger could be made, which optionally dumps parameter values.

anuraaga:
Did some random hacking on this as a dive into the codebase and came up with this hacky POC.

https://github.com/tetratelabs/wazero/compare/main...anuraaga:testing

Some initial notes to remember

- We currently only allow configuring wasm features, but not engine features orthogonal to wasm, like an interceptor. Just some additional wiring
- `InstantiateModuleFromCodeWithConfig` probably needs to be allowed a `Context` because it can trigger a `.start`. Currently there doesn't seem to be a way to trace `.start` without the wazero initialization ending up in it. I see that `.Call()` maybe has the ability to pass a context in as part of the Module, but I'm not sure on that yet

codefromthecrypt:
interesting and good progress.

Your code made me notice that how things are called is engine-specific due to how the stacks work. We can continue with the interceptor pattern or go with a start/finish hook thing depending on what you think works best. Of course the onStart hook can accept a context and return an updated one to use for real. (it can also take the slice of params though can't effect them). For correlation between start and finish we can add another parameter as needed.

When this shapes up whatever base interface can go into the api package and that might allow some impl flexibility without circular deps tangling.

[...]

PR #482, opened 2022-04-19
Refactors API to ensure context propagation

codefromthecrypt:
This is an API breaking change that does a few things:

* Stop encouraging practice that can break context propagation:
  * Stops caching `context.Context` in `wazero.RuntimeConfig`
  * Stops caching `context.Context` in `api.Module`

* Fixes context propagation in function calls:
  * Changes `api.Function`'s arg0 from `api.Module` to `context.Context`
  * Adds `context.Context` parameter in instantiation (propagates to
    .start)

* Allows context propagation for heavy operations like compile:
  * Adds `context.Context` as the initial parameter of `CompileModule`

The design we had earlier was a good start, but this is the only way to
ensure coherence when users start correlating or tracing. While adding a
`context.Context` parameter may seem difficult, wazero is a low-level
library and WebAssembly is notoriously difficult to troubleshoot. In
other words, it will be easier to explain to users to pass (even nil) as
the context parameter vs try to figure out things without coherent
context.

## Example impact in normal usage

Users should pass non-nil context, but we allow nil to coerce to `context.Background()`
```diff
+        // Choose the context to use for function calls.
+        ctx := context.Background()
+
        // Read a WebAssembly binary containing an exported "fac" function.
        // * Ex. (func (export "fac") (param i64) (result i64) ...
        source, _ := os.ReadFile("./tests/bench/testdata/fac.wasm")
 
        // Instantiate the module and return its exported functions
-       module, _ := wazero.NewRuntime().InstantiateModuleFromCode(source)
+       module, _ := wazero.NewRuntime().InstantiateModuleFromCode(ctx, source)
        defer module.Close()
 
        // Discover 7! is 5040
-       fmt.Println(module.ExportedFunction("fac").Call(nil, 7))
+       fmt.Println(module.ExportedFunction("fac").Call(ctx, 7))
```

## Example impact defining host functions

Since modules no longer cache a `context.Context` to access the inbound context, you need to define arg0 as such:
```diff
-               ExportFunction("call_get_age", func(m api.Module) (age uint64) {
+               ExportFunction("call_get_age", func(ctx context.Context, m api.Module) (age uint64) {
                        resultOffsetAge := uint32(8) // arbitrary memory offset (in bytes)
-                       _, _ = m.ExportedFunction("get_age").Call(m, uint64(resultOffsetAge))
+                       _, _ = m.ExportedFunction("get_age").Call(ctx, uint64(resultOffsetAge))
                        age, _ = m.Memory().ReadUint64Le(resultOffsetAge)
                        return
-               }).Instantiate()
+               }).Instantiate(ctx)
```

### DIFF
--- a/internal/wasm/call_context.go
+++ b/internal/wasm/call_context.go
@@ -24,6 +24,11 @@ func NewCallContext(store *Store, instance *ModuleInstance, Sys *SysContext) *Ca
 // functionality like trace propagation.
 // Note: this also implements api.Module in order to simplify usage as a host function parameter.
 type CallContext struct {
+	// TODO: We've never found a great name for this. It is only used for function calls, hence CallContext, but it
+	// moves on a different axis than, for example, the context.Context. context.Context is the same root for the whole
+	// call stack, where the CallContext can change depending on where memory is defined and who defines the calling
+	// function. When we rename this again, we should try to capture as many key points possible on the docs.
+
 	module *ModuleInstance
 	// memory is returned by Memory and overridden WithMemory
 	memory api.Memory
--- a/internal/wasm/call_context.go
+++ b/internal/wasm/call_context.go
@@ -12,9 +12,9 @@ import (
 // compile time check to ensure CallContext implements api.Module
 var _ api.Module = &CallContext{}
 
-func NewCallContext(ctx context.Context, store *Store, instance *ModuleInstance, Sys *SysContext) *CallContext {
+func NewCallContext(store *Store, instance *ModuleInstance, Sys *SysContext) *CallContext {
 	zero := uint64(0)
-	return &CallContext{ctx: ctx, memory: instance.Memory, module: instance, store: store, Sys: Sys, closed: &zero}
+	return &CallContext{memory: instance.Memory, module: instance, store: store, Sys: Sys, closed: &zero}
 }
 
 // CallContext is a function call context bound to a module. This is important as one module's functions can call
@@ -24,8 +24,6 @@ func NewCallContext(ctx context.Context, store *Store, instance *ModuleInstance,
 // functionality like trace propagation.
 // Note: this also implements api.Module in order to simplify usage as a host function parameter.
 type CallContext struct {
-	// ctx is returned by Context and overridden WithContext
-	ctx    context.Context // TODO: remove in next PR
 	module *ModuleInstance
 	// memory is returned by Memory and overridden WithMemory
 	memory api.Memory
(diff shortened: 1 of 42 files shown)


internal/wasm/store_test.go
```diff
@@ -369,66 +366,13 @@ func TestCallContext_ExportedFunction(t *testing.T) {
 	})
 }
 
-func TestFunctionInstance_Call(t *testing.T) {
```
codefromthecrypt, line 372: this actually tested the mock :D

internal/wasm/call_context.go
```diff
@@ -163,15 +149,12 @@ func (f *FunctionInstance) ResultTypes() []api.ValueType {
 }
 
 // Call implements the same method as documented on api.Function
-func (f *FunctionInstance) Call(m api.Module, params ...uint64) (ret []uint64, err error) {
-	mod := f.Module
-	modCtx, ok := m.(*CallContext)
-	if ok {
-		// TODO: check if the importing context is correct
```
codefromthecrypt, line 170: cool thing is we no longer have this TODO! before, we allowed the user (by accident) to change the module input. We did that only to allow the context to get through. Now that we separated the types, we no longer have that bomb.

internal/wasm/gofunc_test.go
```diff
@@ -10,6 +10,9 @@ import (
 	"github.com/tetratelabs/wazero/internal/testing/require"
 )
 
+// testCtx is an arbitrary, non-default context. Non-nil also prevents linter errors.
+var testCtx = context.WithValue(context.Background(), struct{}{}, "arbitrary")
```
codefromthecrypt, line 14: ps if wondering I added this once per package instead of exporting a test utility. If we feel strongly about it we can do that in the testing package.

wasm.go
```diff
@@ -43,65 +45,75 @@ type Runtime interface {
 	//  * Improve performance when the same module is instantiated multiple times under different names
 	//  * Reduce the amount of errors that can occur during InstantiateModule.
 	//
+	// Note: when `ctx` is nil, it defaults to context.Background.
```
mathetake, line 48: q: how do you envision ctx is used for compilation? timeout ?

codefromthecrypt: both timeout (cancel) and also tracing. For example, compilation should be in a different span. We don't have a hook yet for that, but having the context allows us to add that and have it with a coherent parent.

For example, if there's an RPC hook that compiles a module on-demand, compile can end up connected to the right trace.

mathetake: super cool

mathetake (approved):

codefromthecrypt:
thanks for noticing this problem @anuraaga! we're looking forward to your trace goodies #451

PR #485, opened 2022-04-20
Fix validate start section with imports

mathetake:
Previously, when searching the signature of imported function start section,
we haven't skipped non-function imports for indexes.


mathetake:
ps this is the first bug I found via #481

mathetake:
```
(module
  (func $g (import "env" "foo"))
  (global $x (import "env" "glob") i32)
  (func $f (import "env" "bar"))
  (start $f)
)

```

This is the valid module (wasm-tools validate succeeds), but our validation fails.

codefromthecrypt (approved):
good!

PR #489, opened 2022-04-20
Adds Memory.IndexByte and memory grow example

codefromthecrypt:
This adds Memory.IndexByte which allows efficent scanning for a
delimiter, ex NUL(0) in null-terminated strings.

This also adds an ad-hoc test to ensure we can export memory functions
such as grow. While this is implicitly in the spectests, it is easier to
find in the ad-hoc tests.


mathetake (approved):

codefromthecrypt:
thanks for the look again!

PR #493, opened 2022-04-21
Consolidates vs benchmark code

codefromthecrypt:
This makes it easier to add the next bench which I think should be allocation per engine!


codefromthecrypt:
I will make one tweak to make it more fair, basically formalize the input and output arity

codefromthecrypt:
made it more fair by having no-one externally allocate slices for inputs and outputs.

codefromthecrypt:
PTAL I'm done with this phase, basically it now separates out runtimes into different files. This should allow us to add runners that will likely not build anywhere conditional on build flags, such as wasmedge. I'll not add another runner in this PR though as it is already big.

mathetake (approved):

PR #496, opened 2022-04-22
Create an internal EngineConfig type for currently holding wasm featu…

anuraaga:
…res and in the future more.

I would like to add the ability for a user to register a function listener to trace function calls. Currently we only allow configuring wasm opt-in features. To add configurability I could come up with some approaches

- Create a `EngineConfig`, currently only containing the wasm features knob, but also allowing adding non-wasm related engine configuration in the future
- Extend the function signature of `NewEngine` for new parameters such as the function listeners
- Adding `With` type of functions to both engines to forward additional configuration through

This PR implements the first idea as a strawman's proposal.

### DIFF
--- a/config.go
+++ b/config.go
@@ -15,23 +15,27 @@ import (
 
 // RuntimeConfig controls runtime behavior, with the default implementation as NewRuntimeConfig
 type RuntimeConfig struct {
-	enabledFeatures wasm.Features
-	newEngine       func(wasm.Features) wasm.Engine
-	memoryMaxPages  uint32
+	engineConfig   wasm.EngineConfig
+	newEngine      func(config wasm.EngineConfig) wasm.Engine
+	memoryMaxPages uint32
 }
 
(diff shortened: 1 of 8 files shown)


mathetake (approved):
looks good and reasonable to me, deferring to @codefromthecrypt

codefromthecrypt:
tl'dr' Let's see a POC of intercepter in a PR first and once that's in the right shape, if the PR is too big, split the change into here!

long answer: I'd like to hold off on this until we know the shape of the integration of interceptors, since this is the only motivation for change. What I often do is get a PR together that has what I need in it then "shed code" into PRs like this after it is in the right overall shape. This is a less tech-debt creating approach as you already know things will generally work before making changes that affect possibly unrelated parts. While true this is internal and we can roll it back anyway, it is still nicer I think to see the integrated picture in case there's something about that which could be put to another place.

anuraaga:
Will send another PR with the function listener

PR #498, opened 2022-04-22
Fix build.examples so it builds all examples

anuraaga:
`$<` as we currently have only returns one of the matched files, meaning that currently `build.examples` doesn't build all the matched examples. `$^` returns all of them, so I added a for loop through that so they're all built.


codefromthecrypt (approved):
Thanks!

PR #502, opened 2022-04-24
Propagates context to all api interface methods that aren't constant

codefromthecrypt:
This prepares for exposing operations like Memory.Grow while keeping the
ability to trace what did that, by adding a `context.Context` initial
parameter. This adds this to all API methods that mutate or return
mutated data.

Before, we made a change to trace functions and general lifecycle
commands, but we missed this part. Ex. We track functions, but can't
track what closed the module, changed memory or a mutable constant.
Changing to do this now is not only more consistent, but helps us
optimize at least the interpreter to help users identify otherwise
opaque code that can cause harm. This is critical before we add more
functions that can cause harm, such as Memory.Grow.

### DIFF
--- a/internal/wasm/call_context.go
+++ b/internal/wasm/call_context.go
@@ -80,6 +80,8 @@ func (m *CallContext) Close(ctx context.Context) (err error) {
 
 // CloseWithExitCode implements the same method as documented on api.Module.
 func (m *CallContext) CloseWithExitCode(_ context.Context, exitCode uint32) (err error) {
+	// Note: If you use the context.Context param, don't forget to coerce nil to context.Background()!
+
 	closed := uint64(1) + uint64(exitCode)<<32 // Store exitCode as high-order bits.
 	if !atomic.CompareAndSwapUint64(m.closed, 0, closed) {
 		return nil
--- a/internal/wasm/call_context.go
+++ b/internal/wasm/call_context.go
@@ -74,12 +74,12 @@ func (m *CallContext) String() string {
 }
 
 // Close implements the same method as documented on api.Module.
-func (m *CallContext) Close() (err error) {
-	return m.CloseWithExitCode(0)
+func (m *CallContext) Close(ctx context.Context) (err error) {
+	return m.CloseWithExitCode(ctx, 0)
 }
 
 // CloseWithExitCode implements the same method as documented on api.Module.
-func (m *CallContext) CloseWithExitCode(exitCode uint32) (err error) {
+func (m *CallContext) CloseWithExitCode(_ context.Context, exitCode uint32) (err error) {
 	closed := uint64(1) + uint64(exitCode)<<32 // Store exitCode as high-order bits.
 	if !atomic.CompareAndSwapUint64(m.closed, 0, closed) {
 		return nil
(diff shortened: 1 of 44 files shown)


codefromthecrypt:
One trivia is that while currently nothing can really track mutable updates of exported globals, these hooks allow it. Thankfully, the primary use case of tracking "memory.grow" will already work as both interpreter and JIT have a context-aware callsite to it.

internal/wasm/jit/engine.go
```diff
@@ -740,10 +740,10 @@ func (ce *callEngine) builtinFunctionGrowCallFrameStack() {
 	ce.globalContext.callFrameStackElementZeroAddress = stackSliceHeader.Data
 }
 
-func (ce *callEngine) builtinFunctionMemoryGrow(mem *wasm.MemoryInstance) {
+func (ce *callEngine) builtinFunctionMemoryGrow(ctx context.Context, mem *wasm.MemoryInstance) {
 	newPages := ce.popValue()
 
-	res := mem.Grow(uint32(newPages))
+	res := mem.Grow(ctx, uint32(newPages))
```
codefromthecrypt, line 746: ex saved by the bell that even if we can't track most things in JIT due to it being native code, we can track grow!

internal/wasm/call_context.go
```diff
@@ -74,12 +74,12 @@ func (m *CallContext) String() string {
...
 // Close implements the same method as documented on api.Module.
-func (m *CallContext) Close() (err error) {
-	return m.CloseWithExitCode(0)
+func (m *CallContext) Close(ctx context.Context) (err error) {
+	return m.CloseWithExitCode(ctx, 0)
 }
 
 // CloseWithExitCode implements the same method as documented on api.Module.
-func (m *CallContext) CloseWithExitCode(exitCode uint32) (err error) {
+func (m *CallContext) CloseWithExitCode(_ context.Context, exitCode uint32) (err error) {
```
anuraaga, line 82: Do you think it's worth adding a comment along the lines of `// Currently context isn't used, but if using in the future, don't forget to convert nil to context.Background`?

anuraaga: Though we probably don't need a comment if we add a unit test now passing in `nil` which would presumably fail if used  in the future. Not sure if I see the test but maybe just missing it.

codefromthecrypt: sure I can do that

codefromthecrypt: I both changed unit tests to pass nil and not nil, as well added the note you mentioned. Good idea!

internal/wasm/interpreter/interpreter.go
```diff
@@ -1670,6 +1669,17 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, callCtx *wasm.CallCont
 	ce.popFrame()
 }
 
+// popMemoryOffset takes a memory offset off the stack for use in load and store instructions.
+// As the top of stack value is 64-bit, this ensures it is in range before returning it.
+func (ce *callEngine) popMemoryOffset(op *interpreterOp) uint32 {
+	// TODO: Document what 'us' is and why we expect to look at value 1.
```
codefromthecrypt, line 1675: @mathetake I want to backfill a test for this, but before I do, can you help me answer this?

mathetake: that's because we holds the offset constant there for memory operations 
https://github.com/tetratelabs/wazero/blob/main/internal/wasm/interpreter/interpreter.go#L357-L393

mathetake (approved):

Issue #321, opened 2022-03-03
Add opt-in support for post 1.0 bulk-memory-operations


codefromthecrypt:
Looks like Grain defaults to use bulk-memory-operations, though this can be disabled with `--no-bulk-memory` per https://github.com/grain-lang/grain/pull/819

Meanwhile, our crash isn't so pretty:

```
(41/323): invalid instruction 0xfc
```

cc @pims

codefromthecrypt:
This one is luckily not required by grain (as it can be controlled by a flag). Luckily, as it requires a fair amount of analysis to implement:

https://github.com/WebAssembly/spec/blob/main/proposals/bulk-memory-operations/Overview.md
https://github.com/WebAssembly/spec/commit/7fa2f20a6df4cf1c114582c8cb60f5bfcdbf1be1

PR #504, opened 2022-04-25
bulk memory: `memory.{fill, copy, init}` and `data.drop`

mathetake:
This commit implements 4 instructions from bulk memory proposal;
`memory.fill`, `memory.copy`, `memory.init` and `data.drop`.

`table.copy`, `table.fill` and `elem.drop` will be implemented in
the followup PRs.

Note that the implementation is pretty naive for-loop which copies
a byte one by one and not performant at all, leaving the optimization
to the future PRs.

part of #321


mathetake:
pretty much wip :D

mathetake:
ok finished arm64.. start working on amd64! 👯

internal/integration_test/spectest/spectest.go
```diff
@@ -0,0 +1,516 @@
+package spectest
+
```
mathetake, line 2: extracted the spec tests code into the public package so that they can be used for bulk memory spectests.

internal/wasm/interpreter/interpreter.go
```diff
@@ -500,6 +500,20 @@ func (e *engine) lowerIR(ir *wazeroir.CompilationResult) (*code, error) {
...
+			op.us = make([]uint64, 1)
+			op.us[0] = uint64(o.DataIndex)
+		case *wazeroir.OperationMemoryCopy:
+		case *wazeroir.OperationMemoryFill:
+		case *wazeroir.OperationTableInit:
+			panic("TODO: table.init unimplemented")
+		case *wazeroir.OperationElemDrop:
+			panic("TODO: elem.drop unimplemented")
+		case *wazeroir.OperationTableCopy:
+			panic("TODO: table.copy unimplemented")
```
mathetake, line 516: these are the TODOs in the next PR which will complete the bulk memory proposal.

internal/wasm/jit/engine.go
```diff
@@ -288,7 +293,7 @@ const (
 
 // jitCallStatusCode represents the result of `jitcall`.
 // This is set by the jitted native code.
-type jitCallStatusCode byte
+type jitCallStatusCode uint32
```
mathetake, line 296: changed a byte to uint32 so that there's no ambiguity in the offset (by Go's alignment adjustment)

internal/integration_test/asm/arm64_debug/impl_test.go
```diff
@@ -705,9 +708,6 @@ func TestAssemblerImpl_EncodeConstToRegister(t *testing.T) {
 								err = a.EncodeConstToRegister(&asm_arm64.NodeImpl{Instruction: tc.inst, SrcConst: c, DstReg: r})
 								require.NoError(t, err)
 
-								fmt.Println(hex.EncodeToString(expected))
```
codefromthecrypt, line 708: 😢

mathetake:
talking offline, and assigning @codefromthecrypt for test infra refactoring 😄

internal/integration_test/asm/arm64_debug/impl_test.go
mathetake: lol

internal/wasm/binary/data.go
```diff
@@ -10,18 +10,34 @@ import (
...
-		return nil, fmt.Errorf("read memory index: %v", err)
+		return nil, fmt.Errorf("read data segment prefix: %w", err)
 	}
 
-	if d != 0 {
-		return nil, fmt.Errorf("invalid memory index: %d", d)
-	}
+	var expr *wasm.ConstantExpression
+	switch dataSegmentPrefx {
+	case 0x0, 0x2:
```
codefromthecrypt, line 20: constants of some kind?

internal/wasm/interpreter/interpreter.go
```diff
@@ -1671,6 +1686,49 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, callCtx *wasm.CallCont
...
+			}
+			frame.pc++
+		case wazeroir.OperationKindMemoryFill:
+			fillSize := ce.popValue()
+			value := byte(ce.popValue())
+			offset := ce.popValue()
+			if fillSize+offset > uint64(len(memoryInst.Buffer)) {
+				panic(wasmruntime.ErrRuntimeOutOfBoundsMemoryAccess)
+			} else if fillSize != 0 {
+				// Uses the copy trcik for faster filling buffer.
```
codefromthecrypt, line 1723:
```suggestion
				// Uses the copy trick for faster filling buffer.
```

internal/wasm/jit/compiler.go
```diff
@@ -337,4 +337,24 @@ type compiler interface {
 	// compileSignExtend64From32 adds instruction to sign-extends the first 32-bits of 64-bit in as signed 64-bit int.
 	// See wasm.OpcodeI64Extend32S
 	compileSignExtend64From32() error
+	// compileMemoryInit adds instructions to perform operations corresponding to the memory.init instruction in
+	// the bulk-memory-operations proposal.
+	//
+	// https://github.com/WebAssembly/spec/blob/main/proposals/bulk-memory-operations/Overview.md
```
codefromthecrypt, line 343: use the spec 2.0 link to memory.init (here and below)? https://www.w3.org/TR/2022/WD-wasm-core-2-20220419/ ...

internal/wasm/jit/jit_impl_amd64.go
```diff
@@ -3357,6 +3358,305 @@ func (c *amd64Compiler) compileMemorySize() error {
 	return nil
 }
 
+// compileStore implements compiler.compileMemoryInit for the amd64 architecture.
```
codefromthecrypt, line 3361:
```suggestion
// compileMemoryInit implements compiler.compileMemoryInit for the amd64 architecture.
```

internal/wasm/jit/jit_impl_amd64.go
```diff
@@ -3357,6 +3358,305 @@ func (c *amd64Compiler) compileMemorySize() error {
...
+	c.assembler.CompileNoneToRegister(amd64.INCQ, copySize.register)
+	c.assembler.CompileJump(amd64.JMI).AssignJumpTarget(beginCopyLoop)
+
+	c.locationStack.markRegisterUnused(copySize.register, sourceOffset.register,
+		destinationOffset.register, dataInstanceAddr, tmp)
+	c.assembler.SetJumpTargetOnNext(skipJump)
+	return nil
+}
+
+// compileStore implements compiler.compileDataDrop for the amd64 architecture.
```
codefromthecrypt, line 3448:
```suggestion
// compileDataDrop implements compiler.compileDataDrop for the amd64 architecture.
```

internal/wasm/jit/jit_impl_amd64.go
```diff
@@ -3357,6 +3358,305 @@ func (c *amd64Compiler) compileMemorySize() error {
...
+	// dst = &moduleInstance.DataInstances[0] + dst
+	//     = &moduleInstance.DataInstances[0] + dataIndex*24
+	//     = &moduleInstance.DataInstances[dataIndex]
+	c.assembler.CompileMemoryToRegister(amd64.ADDQ,
+		amd64ReservedRegisterForCallEngine, callEngineModuleContextDataInstancesElement0AddressOffset,
+		dst,
+	)
+}
+
+// compileStore implements compiler.compileMemoryCopy for the amd64 architecture.
```
codefromthecrypt, line 3477:
```suggestion
// compileMemoryCopy implements compiler.compileMemoryCopy for the amd64 architecture.
```

internal/wasm/jit/jit_impl_amd64.go
```diff
@@ -3357,6 +3358,305 @@ func (c *amd64Compiler) compileMemorySize() error {
...
+		c.assembler.CompileJump(amd64.JMI).AssignJumpTarget(beginCopyLoop)
+	}
+
+	c.locationStack.markRegisterUnused(copySize.register, sourceOffset.register,
+		destinationOffset.register, tmp)
+	c.assembler.SetJumpTargetOnNext(skipJump, endJump)
+	return nil
+}
+
+// compileStore implements compiler.compileMemoryCopy for the amd64 architecture.
```
codefromthecrypt, line 3593:
```suggestion
// compileMemoryFill implements compiler.compileMemoryFill for the amd64 architecture.
```

internal/wasm/jit/jit_impl_arm64.go
```diff
@@ -2861,6 +2861,372 @@ func (c *arm64Compiler) compileFloatConstant(is32bit bool, value uint64) error {
 	return nil
 }
 
+// compileStore implements compiler.compileMemoryInit for the amd64 architecture.
```
codefromthecrypt, line 2864:
```suggestion
// compileMemoryInit implements compiler.compileMemoryInit for the arm64 architecture.
```

internal/wasm/jit/jit_impl_arm64.go
```diff
@@ -2861,6 +2861,372 @@ func (c *arm64Compiler) compileFloatConstant(is32bit bool, value uint64) error {
...
+
+	c.markRegisterUnused(copySize.register, sourceOffset.register,
+		destinationOffset.register, dataInstanceAddr)
+
+	c.assembler.SetJumpTargetOnNext(skipCopyJump)
+
+	return nil
+}
+
+// compileDataDrop implements compiler.compileDataDrop for the amd64 architecture.
```
codefromthecrypt, line 2978:
```suggestion
// compileDataDrop implements compiler.compileDataDrop for the arm64 architecture.
```

internal/wasm/jit/jit_impl_arm64.go
```diff
@@ -2861,6 +2861,372 @@ func (c *arm64Compiler) compileFloatConstant(is32bit bool, value uint64) error {
...
+		arm64ReservedRegisterForTemporary,
+	)
+
+	// dst = arm64ReservedRegisterForTemporary + dst
+	//     = &moduleInstance.DataInstances[0] + dataIndex*24
+	//     = &moduleInstance.DataInstances[dataIndex]
+	c.assembler.CompileRegisterToRegister(arm64.ADD, arm64ReservedRegisterForTemporary, dst)
+}
+
+// compileStore implements compiler.compileMemoryCopy for the arm64 architecture.
```
codefromthecrypt, line 3010:
```suggestion
// compileMemoryCopy implements compiler.compileMemoryCopy for the arm64 architecture.
```

internal/wasm/jit/jit_impl_arm64.go
```diff
@@ -2861,6 +2861,372 @@ func (c *arm64Compiler) compileFloatConstant(is32bit bool, value uint64) error {
...
+	}
+
+	// Mark all of the operand registers.
+	c.markRegisterUnused(copySize.register, sourceOffset.register, destinationOffset.register)
+
+	c.assembler.SetJumpTargetOnNext(skipCopyJump, endJump)
+	return nil
+}
+
+// compileStore implements compiler.compileMemoryCopy for the arm64 architecture.
```
codefromthecrypt, line 3149:
```suggestion
// compileMemoryFill implements compiler.compileMemoryFill for the arm64 architecture.
```

internal/wasm/module.go
```diff
@@ -662,6 +664,15 @@ type DataSegment struct {
 	Init             []byte
 }
 
+// IsPassive returns true if this data segment is "passive" in the sense that memory offset and
+// index is determined at runtime and used by "memory.init" instruction in the bulk memory
+// operations proposal.
+//
+// See https://github.com/WebAssembly/spec/blob/main/proposals/bulk-memory-operations/Overview.md
```
codefromthecrypt, line 671: use 2.0 links except when discussing the feature flag or something that is more general than an op like "memory.init" PS use XXXName instead of quoted constants where possible.

internal/wasm/store.go
```diff
@@ -73,6 +73,12 @@ type (
 		// TypeIDs is index-correlated with types and holds typeIDs which is uniquely assigned to a type by store.
 		// This is necessary to achieve fast runtime type checking for indirect function calls at runtime.
 		TypeIDs []FunctionTypeID
+
+		// DataInstances holds data segments bytes of the module.
+		// This is only used by bulk memory operations.
+		//
+		// https://webassembly.github.io/spec/core/exec/runtime.html#data-instances
```
codefromthecrypt, line 80: permalink to 2.0?

codefromthecrypt (approved):
Looks good. If you can, take a second pass to see if there are any code temptable between amd64 and arm64 to reduce code drift or copy/paste. If trying that makes things worse than now, that's ok, just try.

There are some places where we are using drifty links, so try to switch those to permalink ex about the spec as 2.0 draft is out now.

If you don't mind, add anything you think is interesting enough to mandate a wazeroir compiler_test to that, at least one op if there aren't more than one that feel needed.

After this, I'll give a try to decouple this feature from the spectest infra as I'm hoping we have another way out. there may not be such a way, but I want to give it a try!

mathetake:
done. I won't have shared code between amd64 and arm64 as the optimized code would be totally different from each other as in musl libc for example. (memcpy in [arm64](https://github.com/ziglang/zig/blob/master/lib/libc/musl/src/string/aarch64/memcpy.S) vs  [amd64](https://github.com/ziglang/zig/blob/master/lib/libc/musl/src/string/x86_64/memcpy.s)

codefromthecrypt:
> done. I won't have shared code between amd64 and arm64 as the optimized code would be totally different from each other as in musl libc for example. (memcpy in [arm64](https://github.com/ziglang/zig/blob/master/lib/libc/musl/src/string/aarch64/memcpy.S) vs [amd64](https://github.com/ziglang/zig/blob/master/lib/libc/musl/src/string/x86_64/memcpy.s)

Cool thanks for checking. I was mainly playing devil's advocate

internal/integration_test/post1_0/bulk-memory-operations/spec_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+			// The first segment is active, so we expect it to be readable.
+			bytes, ok := mod.Memory().Read(testCtx, 0, 5)
+			require.True(t, ok)
+			require.Equal(t, "hello", string(bytes))
+
+			// As the value of the global was zero, we don't expect "memory.init" to have copied the passive segment.
+			bytes, ok = mod.Memory().Read(testCtx, 16, 7)
+			require.True(t, ok)
+			// TODO: require.Equal(t, "goodbye", string(bytes))
```
codefromthecrypt, line 81: @mathetake FYI I left this until it is implemented

mathetake: oh? is this table related?

mathetake: everything memory.fill copy and init is ensured to be implemented so this should work

codefromthecrypt:
```
(module $bulk-memory-operations
   (import "a" "global" (global i32))  ;; global 0
   (memory 1)
   (data (i32.const 0) "hello")   ;; data segment 0, is active so always copied
   (data "goodbye")               ;; data segment 1, is passive

   (func $start
     (if (global.get 0)

       ;; copy data segment 1 into memory 0 (the 0 is implicit)
       (memory.init 1
         (i32.const 16)    ;; target offset
         (i32.const 0)     ;; source offset
         (i32.const 7))    ;; length

       ;; The memory used by this segment is no longer needed, so this segment can
       ;; be dropped.
       (data.drop 1))
   )
 )
```

the if should trigger and invoke the memory.init thingy.. wasn't sure why that didn't run assumed something wasn't done, yet. Let me double-check the wasm generated includes it.

codefromthecrypt: yeah seems to

```
$ wasm2wat  bulk_memory_operations.wasm 
(module
  (type (;0;) (func))
  (import "a" "global" (global (;0;) i32))
  (func (;0;) (type 0)
    global.get 0
    if  ;; label = @1
      i32.const 16
      i32.const 0
      i32.const 7
      memory.init 1
    else
      data.drop 1
    end)
  (memory (;0;) 1)
  (data (;0;) (i32.const 0) "hello")
  (data (;1;) "goodbye"))
``

mathetake: not following, but this only contains ones implemented so should work

codefromthecrypt: feature overview forgot to make the start function start. wasn't a bug on our side except copying the bug :D

codefromthecrypt:
I'm going to land more features in the text compiler so that making the spectests around memory and element segments are easier to do. I think we should merge this as it already passed "brute force" spectests. I think it is better than committing those to Git just to revert them later.

mathetake:
good to go!

codefromthecrypt:
nice work @mathetake!

PR #505, opened 2022-04-25
experimental: FunctionListenerFactory can decide whether to intercept function start/finish

anuraaga:
For #451 (maybe fixes if this is enough functionality for now)

This allows registering a `FunctionListenerFactory` to a runtime. Anytime a module is instantiated, the factory is invoked while building the functions within the module. If the factory returns a listener, it is invoked anytime the defined function is called with `context`. The `context` can be used to propagate stack across function invocations in a stack.

We'll keep the hooks internal for now to let us bake within our own tests, so registration is defined using an internal interface implemented by `runtime`.


examples/function-listener/print-trace.go
```diff
@@ -0,0 +1,139 @@
...
+//
+// See README.md for a full description.
+func main() {
+	// Choose the context to use for function calls.
+	ctx := context.Background()
+
+	// Create a new WebAssembly Runtime.
+	r := wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfigInterpreter())
+
+	// Sets the FunctionListenerFactory using an internal accessor.
```
anuraaga, line 48: While this example was useful for me to see the listening, realized it's not useful to anyone else since it can only be used in this repo. Guess I should remove the example before getting ready to merge

internal/wasm/module.go
```diff
@@ -528,6 +533,39 @@ func (m *Module) buildMemory() (mem *MemoryInstance) {
...
+	}
+
+	for _, nm := range m.NameSection.LocalNames {
+		if nm.Index == funcIdx {
+			for _, n := range nm.NameMap {
+				if int(n.Index) < len(params) {
+					params[n.Index].Name = n.Name
+				} else {
+					resIdx := int(n.Index) - len(params)
+					// Only malformed program would have an index out of bounds here.
```
anuraaga, line 553: Just first thought, need to write test cases to understand this logic better

anuraaga (commented):
I think this is ready for a first look before backfilling tests

internal/wasm/module.go
```diff
@@ -895,3 +933,66 @@ func ExternTypeName(et ExternType) string {
...
 }
+
+// A FunctionListener can be registered for any function via FunctionListenerFactory to
+// be notified when the function is called.
+type FunctionListener interface {
+	// Before is invoked before a function is called. ctx is the context of the caller function.
+	// The returned context will be used as the context of this function call. To add context
+	// information for this function call, add it to ctx and return the updated context. If
+	// no context informationis needed, return ctx as is.
+	Before(ctx context.Context) context.Context
```
anuraaga, line 944: For now values aren't passed here. I don't know how useful values would be in general given they're usually some opaque ABI-related value or pointer.

examples/function-listener/print-trace.go
codefromthecrypt: you could move it to internal/examples for now

internal/wasm/module.go
codefromthecrypt: maybe extract this sort of thing to listener.go/listener_test.go

codefromthecrypt:
I would recommend squashing and trying to contain changes as much as possible to new files. since a lot is in the same package, that should be possible. That way this low priority change doesn't have so much drift when we try to focus on things important to complete in the next 2 weeks

internal/wasm/host.go
```diff
@@ -176,6 +176,26 @@ func (m *Module) buildHostFunctions(moduleName string) (functions []*FunctionIns
...
+			params := make([]ValueInfo, len(f.ParamTypes()))
+			for i, p := range f.ParamTypes() {
+				params[i].Type = p
+			}
+			results := make([]ValueInfo, len(f.ResultTypes()))
+			for i, r := range f.ResultTypes() {
+				results[i].Type = r
+			}
+
+			// Cannot access parameter names for host functions.
```
codefromthecrypt, line 189: ps I plan to change the ModuleBuilder to supply param names as often these should match something else, like wasi definitions

codefromthecrypt:
This is great, @anuraaga!

As discussed offline, I squashed this draft into one commit then will add another to make it completely internal (ex no leak to wazero package) so we can iterate without effecting anyone.

codefromthecrypt:
@anuraaga if you are keen, mark ready for review

internal/experimental/examples/function-listener/README.md
```diff
@@ -0,0 +1,20 @@
...
+This example shows how to define a function listener to trace function calls. Function listeners are currently
+only implemented for interpreter mode.
+
+If the current year is 2022, and we give the argument 2000, [age-calculator.go](age-calculator.go) should output 22.
+```bash
+$ go run age-calculator.go 2000
+println >> 21
+log_i32 >> 21
+```
+
```
anuraaga, line 12:
```suggestion
```

Sorry my drift

codefromthecrypt (approved):
@mathetake if you are keen click mergerooo.. otherwise comment and we'll sort it

mathetake (approved):

mathetake (approved):

codefromthecrypt:
nice work @anuraaga!

PR #508, opened 2022-04-26
benchmark: removes go-wasm3

codefromthecrypt:
The project we were using for comparison with wasm3 was archived. There
was a good alternative mentioned by @birros which was to do benchmarks
in C directly, but I don't think it is important enough for us to start adding C
to this project over.

The main point of these benchmarks wasn't to compare against anything
that can run wasm, rather runtimes end users can use in Go. Mainly, we
wanted to understand how they are used and where we stand in impl, for
better or for worse. In other words, even though end users can write their
own CGO interface to a library that doesn't have its own go binding, ex
wasm3, it is probably better to leave that comparison out as it is neither
relevant nor time conservative to the project maintainers.

Anyway, it was fun while it lasted!

See https://github.com/birros/go-wasm3/issues/5#issuecomment-1109691123


codefromthecrypt:
revised desc as I was a bit more awkward than I wanted to be

codefromthecrypt:
ps on C, we'll probably have it in the examples directory at some point, maybe using zig to show building an example wasm using C. just we're trying to get towards api stability and adding C for the purposs of launch benchmarks vs other projects isn't probably not a wise use of time right now, and possibly not the right repo either.

PR #513, opened 2022-04-28
Adds experimental package to expose listener and sys APIs

### LINKED ISSUE #491: wasi: expose setters for timeNow and randSource
I have a somewhat niche use-case that requires strict non-determinism, and from what I can tell the `wasi` implementation mostly supports this except for 1) time and 2) randomness. I see the implementation has already provided hooks to override these in tests, but it would be very helpful for me to set these outside of the package.

Totally understand if this is deemed outside the scope of normal usage and not a priority. 

```
		timeNowUnixNano: func() uint64 {
		randSource: func(p []byte) error {
```

codefromthecrypt:
This adds an experimental package to expose two work-in-progress
features:
* FunctionListener - for tracing etc.
* Sys - to control random number generators

Both the functionality and the names of the features above are
not stable. However, this should help those who can tolerate drift a
means to test things out.

Fixes #491


codefromthecrypt:
ps @anuraaga I rejigged some things as I was a little worried about relying on our debug formatting externally.

internal/experimental/examples/function-listener/print-trace.go
```diff
@@ -8,30 +8,39 @@ import (
 
 	"github.com/tetratelabs/wazero"
 	"github.com/tetratelabs/wazero/api"
-	experimentalapi "github.com/tetratelabs/wazero/internal/experimental/api"
+	"github.com/tetratelabs/wazero/experimental"
+	"github.com/tetratelabs/wazero/internal/wasmdebug"
 )
 
 type stackKey struct{}
```
codefromthecrypt, line 15: ps I left this example in the really internal area as I'm not confident this is the practice we want, yet. I put a much smaller example in the listener_test which is also easier to follow for now.

experimental/sys.go
```diff
@@ -0,0 +1,13 @@
+package experimental
+
+// SysKey is a context.Context Value key. Its associated value should be a Sys.
+type SysKey struct{}
+
+// Sys controls experimental aspects currently only used by WASI.
+type Sys interface {
```
codefromthecrypt, line 7: ps naming.. ugh. We'll likely have more sysconfig at some point. Also, I don't want to necessarily say this is only for WASI. Since this is internal, we don't also need to force this name to be great, yet, but ideas welcome.

experimental/sys_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Try to read 4 bytes of random data.
+	if bytes, ok := mod.Memory().Read(ctx, 0, 4); !ok {
+		log.Fatalf("Memory.Read(0, 4) out of range of memory size %d", mod.Memory().Size(ctx))
+	} else {
+		fmt.Println(hex.EncodeToString(bytes))
+	}
+
+	// Output:
+	// 538c7f96
```
codefromthecrypt, line 67: if the deterministic stuff worked, this will never flake :D

experimental/listener.go
```diff
@@ -24,41 +24,32 @@ type FunctionListener interface {
 	// The returned context will be used as the context of this function call. To add context
 	// information for this function call, add it to ctx and return the updated context. If
 	// no context information is needed, return ctx as is.
-	Before(ctx context.Context) context.Context
+	Before(context.Context, []uint64) context.Context
```
codefromthecrypt, line 27: probably should label this params and the other results. This may break in later work, but again.. it is experimental, so that's ok!

codefromthecrypt:
FYI I consolidated the listener examples as it is awkward otherwise.

experimental/listener.go
```diff
@@ -0,0 +1,64 @@
...
+	// After is invoked after a function is called. ctx is the context of this function call.
+	// The err parameter is nil on success.
+	After(ctx context.Context, err error, resultValues []uint64)
+}
+
+// TODO: We need to add tests to enginetest to ensure contexts nest. A good test can use a combination of call and call
+// indirect in terms of depth and breadth. The test could show a tree 3 calls deep where the there are a couple calls at
+// each depth under the root. The main thing this can help prevent is accidentally swapping the context internally.
+
+// TODO: Errors aren't handled, and the After hook should accept one along with the result values.
```
anuraaga, line 40: `After` looks like it has an `error` now so guess this was already DOed?

internal/wasm/interpreter/interpreter.go
```diff
@@ -103,6 +104,20 @@ func (ce *callEngine) popValue() (v uint64) {
 	return
 }
 
+// peekValues pops api.ValueType values from the stack and returns them in reverse order.
```
anuraaga, line 107:
```suggestion
// peekValues peeks api.ValueType values from the stack and returns them in reverse order.
```

mathetake (approved):

anuraaga (commented):
Just a couple of tiny nits

experimental/listener.go
codefromthecrypt: It has a parameter, but "errors aren't handled" is still true. There's neither code nor tests to do that.

codefromthecrypt:
okie dokie. merge time!

PR #514, opened 2022-04-29
Adds Runtime.WithCapacityPages to avoid allocations during runtime.

### LINKED ISSUE #500: Add option to change Memory capacity
That is similar to https://github.com/tetratelabs/wazero/issues/483, but it's capacity, not length.

It's possible to define a maximum memory, https://github.com/tetratelabs/wazero/pull/419, however, each `grow` function can potentially cause reallocs on the host-side/wazero. It puts more pressure on GC and makes things slower, since allocs are expensive.

-----------

I'm using the following code:

```
	rConfig := wazero.NewRuntimeConfigJIT()
	rConfig.WithMemoryMaxPages(10_485_760 / 65_536)
```

The memory limit is ~10MB. So, I would also like to set:

```
	rConfig.WithMemoryCapacity(10_485_760)  // Bad naming
```

Internally, Wazero should do: `make([]byte, MemorySector.Mininum, 10_485_760)`. In that case, the wasm program can call `grow` without creating new allocs, since all the space available is already allocated and reserved.

codefromthecrypt:
`Runtime.WithMemoryCapacityPages` is a function that determines memory
capacity in pages (65536 bytes per page). The inputs are the min and
possibly nil max defined by the module, and the default is to return
the min.

Ex. To set capacity to max when exists:
```golang
c.WithMemoryCapacityPages(func(minPages uint32, maxPages *uint32) uint32 {
	if maxPages != nil {
		return *maxPages
	}
	return minPages
})
```

Note: This applies at compile time, ModuleBuilder.Build or Runtime.CompileModule.

Fixes #500

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -30,8 +30,10 @@ import (
 //	env2, _ := r.InstantiateModuleWithConfig(ctx, env, NewModuleConfig().WithName("env.2"))
 //	defer env2.Close(ctx)
 //
-// Note: Builder methods do not return errors, to allow chaining. Any validation errors are deferred until Build.
-// Note: Insertion order is not retained. Anything defined by this builder is sorted lexicographically on Build.
+// Notes:
+// * ModuleBuilder is mutable. WithXXX functions return the same instance for chaining.
+// * WithXXX methods do not return errors, to allow chaining. Any validation errors are deferred until Build.
+// * Insertion order is not retained. Anything defined by this builder is sorted lexicographically on Build.
 type ModuleBuilder interface {
 	// Note: until golang/go#5860, we can't use example tests to embed code in interface godocs.
 
@@ -91,7 +93,7 @@ type ModuleBuilder interface {
 	//	// (memory (export "memory") 1)
 	//	builder.ExportMemory(1)
 	//
-	// Note: This is allowed to grow to RuntimeConfig.WithMemoryMaxPages (4GiB). To bound it, use ExportMemoryWithMax.
+	// Note: This is allowed to grow to RuntimeConfig.WithMemoryLimitPages (4GiB). To bound it, use ExportMemoryWithMax.
 	// Note: If a memory is already exported with the same name, this overwrites it.
 	// Note: Version 1.0 (20191205) of the WebAssembly spec allows at most one memory per module.
 	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#memory-section%E2%91%A0
@@ -103,7 +105,7 @@ type ModuleBuilder interface {
 	//	// (memory (export "memory") 1 1)
 	//	builder.ExportMemoryWithMax(1, 1)
 	//
-	// Note: maxPages must be at least minPages and no larger than RuntimeConfig.WithMemoryMaxPages
+	// Note: maxPages must be at least minPages and no larger than RuntimeConfig.WithMemoryLimitPages
 	ExportMemoryWithMax(name string, minPages, maxPages uint32) ModuleBuilder
 
 	// ExportGlobalI32 exports a global constant of type api.ValueTypeI32.
@@ -195,7 +197,7 @@ func (b *moduleBuilder) ExportFunctions(nameToGoFunc map[string]interface{}) Mod
 
 // ExportMemory implements ModuleBuilder.ExportMemory
 func (b *moduleBuilder) ExportMemory(name string, minPages uint32) ModuleBuilder {
-	mem := &wasm.Memory{Min: minPages, Max: b.r.memoryMaxPages}
+	mem := &wasm.Memory{Min: minPages, Max: b.r.memoryLimitPages}
 	mem.Cap = b.r.memoryCapacityPages(mem.Min, nil)
 	b.nameToMemory[name] = mem
 	return b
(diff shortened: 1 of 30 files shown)


codefromthecrypt:
I have a couple tests to backfill, but want to ensure the approach I settled on is ok before doing that.

The main reason we need a function as opposed to an absolute value, is this is set on RuntimeConfig which doesn't know the memory values yet. This is unlike MemoryMaxPages, which is a threshold so is constant regardless of memory size.

internal/wasm/memory.go
```diff
@@ -193,17 +195,25 @@ func MemoryPagesToBytesNum(pages uint32) (bytesNum uint64) {
...
+	if newPages > m.Max {
 		return 0xffffffff // = -1 in signed 32-bit integer.
-	} else {
-		// Otherwise, grow the memory.
-		m.Buffer = append(m.Buffer, make([]byte, MemoryPagesToBytesNum(newPages))...)
+	} else if newPages > m.Cap { // grow the memory.
+		m.Buffer = append(m.Buffer, make([]byte, MemoryPagesToBytesNum(delta))...)
+		m.Cap = newPages
+		return currentPages
+	} else { // We already have the capacity we need.
```
codefromthecrypt, line 214: this is the main optimization as it is the only way I know of to reset the length of a slice w/o calling append or copy

config.go
```diff
@@ -69,6 +72,24 @@ func (c *RuntimeConfig) WithMemoryMaxPages(memoryMaxPages uint32) *RuntimeConfig
...
+// Ex. To set capacity to max when exists:
+//	c.WithMemoryCapacityPages(func(minPages uint32, maxPages *uint32) uint32 {
+//		if maxPages != nil {
+//			return *maxPages
+//		}
+//		return minPages
+//	})
+//
+// Note: This applies at compile time, ModuleBuilder.Build or Runtime.CompileModule.
+func (c *RuntimeConfig) WithMemoryCapacityPages(maxCapacityPages func(minPages uint32, maxPages *uint32) uint32) *RuntimeConfig {
```
codefromthecrypt, line 87: once another teammate is keen on this approach, I'll backfill test on this

mathetake:
Looks good to me

internal/wasm/module.go
```diff
@@ -559,9 +559,15 @@ func paramNames(localNames IndirectNameMap, funcIdx uint32, paramLen int) []stri
...
 	memSec := m.MemorySection
 	if memSec != nil {
+		min := MemoryPagesToBytesNum(memSec.Min)
+		capacity := MemoryPagesToBytesNum(memSec.Cap)
+		if capacity == 0 && min != 0 {
+			capacity = min // Internal code may avoid setting Cap. Default to min.
+		}
 		mem = &MemoryInstance{
-			Buffer: make([]byte, MemoryPagesToBytesNum(memSec.Min)),
+			Buffer: make([]byte, min, capacity),
```
mathetake, line 568: if capacity < min, than this could be panic I guess?

wasm.go
```diff
@@ -175,6 +178,15 @@ func (r *runtime) CompileModule(ctx context.Context, source []byte) (*CompiledCo
 		return nil, err
 	}
 
+	// Determine the correct memory capacity, if a memory was defined.
+	if mem := internal.MemorySection; mem != nil {
+		var max *uint32
+		if mem.IsMaxEncoded {
+			max = &mem.Max
+		}
+		mem.Cap = r.memoryCapacityPages(mem.Min, max)
```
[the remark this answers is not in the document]
codefromthecrypt: PTAL sorry I should have mentioned it wasn't just backfilling tests I was going to do next, but also the corner cases.

builder.go
```diff
@@ -91,7 +93,7 @@ type ModuleBuilder interface {
 	//	// (memory (export "memory") 1)
 	//	builder.ExportMemory(1)
 	//
-	// Note: This is allowed to grow to RuntimeConfig.WithMemoryMaxPages (4GiB). To bound it, use ExportMemoryWithMax.
+	// Note: This is allowed to grow to RuntimeConfig.WithMemoryLimitPages (4GiB). To bound it, use ExportMemoryWithMax.
```
codefromthecrypt, line 96: renamed because we use max twice and it is confusing

internal/wasm/module.go
codefromthecrypt: so basically these values are validated now before you'd get to instantiation (during compile)

internal/wasm/module.go
```diff
@@ -559,9 +559,15 @@ func paramNames(localNames IndirectNameMap, funcIdx uint32, paramLen int) []stri
 func (m *Module) buildMemory() (mem *MemoryInstance) {
 	memSec := m.MemorySection
 	if memSec != nil {
+		min := MemoryPagesToBytesNum(memSec.Min)
+		capacity := MemoryPagesToBytesNum(memSec.Cap)
+		if capacity == 0 && min != 0 {
```
codefromthecrypt, line 564: this is a little bomb, I'll remove it. Basically spectests aren't going through normal code paths, but it is better to fix them vs leave this around as it confuses that we've already validated.

mathetake (approved):

codefromthecrypt:
thanks for the eagle eyes @mathetake !

Issue #484, opened 2022-04-19
Support WebAssembly Core Specification 2.0


codefromthecrypt:
Right now, we support WebAssembly Core Specification 1.0 and a few features that were marked finished, but not in a spec yet. While not recommendation phase, yet, there is a draft for 2.0 which helps us group which features will be in it. I *think* we can assume no more or less features will end up in that version.

https://www.w3.org/TR/2022/WD-wasm-core-2-20220419/appendix/changes.html#release-1-1

This issue description should be rewritten with a tick box list of steps left to accomplish.

codefromthecrypt:
PS if anyone needs 2.0, please comment what tools you are using that target 2.0 and what use case and priority (timeline) it is for you. If you can workaround meanwhile with one of the features (as opposed to all of them), please note that too!

codefromthecrypt:
ok revised the description

codefromthecrypt:
TL;DR: we audited the new value types: externref and i128 and decided to continue use of encoding any and all value types as uint64 (for the purpose of `api.Function.Call`). This means anyone (if anyone) uses vector ops from a host function, they'd need to split the 128bit vector into two 64bit vectors before invoking the function.

---

We had lots of chats over the last couple days on how to handle the value types added to WebAssembly 2.0. The concern is the host function boundaries, and that can be split into two

* define host function: How to declare parameters for host functions using new value types
* call function: How to call exported functions whose signature includes new value types.

The new types are:
* [externref][1] which is a possibly nil opaque reference to a host pointer. In go, 64 bits should work.
* [I128][2] currently only used for vectors (ex. [v128.const][3]) which are N (possibly 1) values packed into 128 bits. Ex. [i32x4][4].

The biggest deal here is how important the i128bit type is to be represented as a single unit in our exported API, as opposed to split into two uint64s, representing the high and low 64bits. Remember for this discussion, the only use of this in WebAssembly 2.0 is vector instructions. There are no other proposals included in WebAssembly 2.0 that would use these.

The first question: is this a problem internally?, and it isn't. Function types are knowable before use, so it is not difficult to track how many units of 64bits should be pushed or popped from a stack. We also don't forsee any problem in JIT accessing these values.

[...]

PR #517, opened 2022-05-02
Adds support for multi tables

mathetake:
This commit adds support for multiple tables per module.
Notably, if the `WithFeatureReferenceTypes` is enabled,
`call_indirect`, `table.init` and `table.copy` instructions 
can reference non-zero indexed tables.

part of #484


mathetake:
so the interpreter done... will start on JIT in a few hours

internal/integration_test/spectest/spectest.go
```diff
@@ -308,6 +308,9 @@ func Run(t *testing.T, testDataFS embed.FS, newEngine func(wasm.Features) wasm.E
 		require.NoError(t, json.Unmarshal(raw, &base))
 
 		wastName := basename(base.SourceFile)
+		if wastName != "table_init.wast" {
```
mathetake, line 311: oh man

config.go
```diff
@@ -27,7 +27,7 @@ type RuntimeConfig interface {
 	//
 	// Here are the notable effects:
 	// * Adds `memory.fill`, `memory.init`, `memory.copy` and `data.drop` instructions.
-	// * Adds `table.fill`, `table.init`, `table.copy` and `elem.drop` instructions.
```
mathetake, line 30: table.init is actually depending on the reference type (as it takes reference typed value as an argument) so essentially in ReferenceType proposal

mathetake:
ok now backfilling more cases..

config.go
codefromthecrypt: maybe the feature should say it is dependent on another (ex enabling this enables the other, and visa versa, disabling retypes disables this as it is not functional without it)

mathetake:
ok ready!

config.go
```diff
@@ -71,6 +72,31 @@ type RuntimeConfig interface {
...
+	//   * `table.get`
+	//   * `table.grow`
+	//   * `table.set`
+	//   * `table.size`
+	// * Support for multiple tables per module:
+	//   * `call_indirect`, `table.init`, `table.copy` and `elem.drop` instructions can take non-zero table index.
+	//   * Element segments can take non-zero table index.
+	//
+	//
+	// Note: "reference-types" is mixed with the "bulk-memory-operations" proposal
```
codefromthecrypt, line 92: oops ;)

internal/wasm/binary/section.go
```diff
@@ -240,8 +250,11 @@ func encodeCodeSection(code []*wasm.Code) []byte {
 //
 // See encodeTable
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#table-section%E2%91%A0
-func encodeTableSection(table *wasm.Table) []byte {
-	contents := append([]byte{1}, encodeTable(table)...)
+func encodeTableSection(tables []*wasm.Table) []byte {
+	var contents []byte
+	for _, table := range tables {
+		contents = append([]byte{1}, encodeTable(table)...)
```
codefromthecrypt, line 256: comment why 1

internal/wasm/func_validation_test.go
```diff
@@ -486,119 +486,153 @@ func TestModule_ValidateFunction_BulkMemoryOperations(t *testing.T) {
...
+			},
+			{
+				body:        []byte{OpcodeMiscPrefix, OpcodeMiscTableCopy, 3},
+				flag:        FeatureBulkMemoryOperations | FeatureReferenceTypes,
+				tables:      []*Table{{}, {}, {}, {}},
+				expectedErr: "failed to read source table index for table.copy: EOF",
+			},
+			{
+				body:        []byte{OpcodeMiscPrefix, OpcodeMiscTableCopy, 0, 3},
+				flag:        FeatureBulkMemoryOperations,
```
codefromthecrypt, line 610:
```suggestion
				flag:        FeatureBulkMemoryOperations, // Multiple tables require FeatureReferenceTypes.
```

internal/wasm/table.go
```diff
@@ -41,9 +57,15 @@ const (
 type ElementSegment struct {
 	// OffsetExpr returns the table element offset to apply to Init indices.
 	// Note: This can be validated prior to instantiation unless it includes OpcodeGlobalGet (an imported global).
-	// Note: This is nil if the Mode is either passive or declarative.
+	// Note: This is not nil if and only if the Mode is active.
```
codefromthecrypt, line 60:
```suggestion
	// Note: This is only set when Mode is active.
```

codefromthecrypt (approved):
Looks good. Maybe add one toehold wazeroir compiler test that uses externref and/or multiple table (ex maybe call_indirect)

Issue #516, opened 2022-05-01
data count section not supported as feature "bulk-memory-operations" is disabled


inkeliz:
I'm using `WithFinishedFeatures()`:

	rConfig := wazero.NewRuntimeConfigJIT()
	rConfig.WithFinishedFeatures()

	runtime := wazero.NewRuntimeWithConfig(rConfig)
	if _, err := wasi.InstantiateSnapshotPreview1(context.Background(), runtime); err != nil {
		b.Error(err)
	}

	config := &wazero.ModuleConfig{}
	config.WithStdout(os.Stdout)

	wasifile, err := os.ReadFile(filepath.Join("testdata", "wasi", FileWasm))
	if err != nil {
		b.Error(err)
	}

	wasm, err := runtime.CompileModule(context.Background(), wasifile)
	if err != nil {
		b.Error(err)
	}

That is failing:

```
main_wasi_test.go:94: data count section not supported as feature "bulk-memory-operations" is disabled
```

In that case, the line is:  

	wasm, err := runtime.CompileModule(context.Background(), wasifile)
	if err != nil {
		b.Error(err)
	}

inkeliz:
I need to use it like:

```
wazero.NewRuntimeConfigJIT().WithFinishedFeatures()
```

So `wazero.NewRuntimeConfigJIT()` returns an pointer, and also `WithFinishedFeatures()` takes an pointer, but doesn't modify the existent struct, instead it clone the existent one and return. I don't think it makes sense, since it's an pointer.

codefromthecrypt:
Sorry about the confusion. Incidentally, I just recently changed the godoc on that to clarify it is immutable (and the other config isn't).

We are currently using only pointer receivers on our types as the reverse causes more problems. Basically review becomes more mechanical except a couple special cases, and also linters don't yell as much etc. It is true if someone was looking at the implementation of it being a pointer, this is confusing.

XXXConfig aren't quite "final" in their shape. An interface may be an outcome of this and the internal code wouldn't change (as a RuntimeConfig backed by runtimeConfig struct would likely still be implemented internally with pointers.)

Thanks for raising the mental glitch, I'm not sure what outcome we'll take at this moment, but appreciate the feedback, and it helps regardless. In any case, sorry for the glitch as it caused you distraction and that's not desirable for anyone.

codefromthecrypt:
[...]

PR #518, opened 2022-05-02
Changes RuntimeConfig to an interface and exposes WithWasmCore2

codefromthecrypt:
WebAssembly Core Working Draft 1 recently came out. Before that, we had
a toe-hold feature bucked called FinishedFeatures. This replaces
`RuntimeConfig.WithFinishedFeatures` with `RuntimeConfig.WithWasmCore2`.
This also adds `WithWasmCore1` for those who want to lock into 1.0
features as opposed to relying on defaults.

This also fixes some design debt where we hadn't finished migrating
public types that require constructor functions (NewXxx) to interfaces.
By using interfaces, we prevent people from accidentally initializing
key configuration directly (via &Xxx), causing nil field refs. This also
helps prevent confusion about how to use the type (ex pointer or not) as
there's only one way (as an interface).

See https://www.w3.org/blog/news/archives/9509
See https://github.com/tetratelabs/wazero/issues/516

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,8 +203,13 @@ as well as how to classify a request for a feature we don't yet support.
 
 ### WebAssembly Core
 wazero conforms with spectests [7] defined alongside WebAssembly Core
-Specification [1.0][1]. There is also [work in progress][14] towards release
-[2.0][2], despite it not being a Web Standard, yet.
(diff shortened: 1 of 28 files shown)


codefromthecrypt:
added README about this

mathetake (approved):

PR #521, opened 2022-05-03
Documents why we don't use Option pattern

### LINKED ISSUE #480: Use Option pattern for configuration (e.g., RuntimeConfig)
Currently `RuntimeConfig` uses a fluent builder pattern. I have seen many Go projects use an Option pattern where all the options are passed into the factory

https://github.com/openzipkin/zipkin-go/blob/master/tracer_options.go#L42

I guess this is is fairly idiomatic. I think this is because it allows returning an error for bad parameters.

```go
r, err := wazero.NewRuntime(
  wazero.WithInterceptor(...),
  wazero.WithCompilationCache("/tmp/wazero/cache")
)
if err != nil {
  return fmt.Errorf("did I forget to set cache directory to writable: %w", err)
}
```

The only way to do any parameter validation with fluent is to basically lose fluentness
```go
r, err := wazero.NewRuntime()
r, err = r.WithInterceptor(...)
if err != nil {}
r, err = r.WithCompilationCache(...)
if err != nil {}
```

The mid ground of only returning `(RuntimeConfig, error)` on specific methods that have possible errors is probably even worse for losing symmetry.

Currently, none of the configuration seems to really need validation but I propose switching to the Option pattern now to allow adding options that deserve validation in the future.

codefromthecrypt:
It has happened a couple times where folks have suggested options pattern, probably guessing we didn't have it in the past or explicitly remove it. All the API decisions were quite thought through, and particularly options, but that wasn't obvious if you read the rationale. This adds the rationale into the repo to help conserve time. Instead of spending time re-explaining why options isn't used here, we can spend the same time on more pressing and important design or maintenance issues.

Closes #480

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -4,7 +4,8 @@
 wazero uses internal packages extensively to balance API compatability desires for end users with the need to safely
 share internals between compilers.
 
-End-user packages include `wazero`,with `Config` structs, and `api`, with shared types. Everything else is internal.
+End-user packages include `wazero`, with `Config` structs, `api`, with shared types, and the built-in `wasi` library.
+Everything else is internal.
 
 ### Internal packages
 Most code in wazero is internal, and it is acknowledged that this prevents external implementation of facets such as
@@ -39,7 +40,7 @@ The above guarantees no cyclic dependencies at the cost of having to re-define s
 For example, if `wasm.Store` is a type the user needs access to, it is narrowed by a cover type in the `wazero`:
 
 ```go
-type Runtime struct {
+type runtime struct {
 	s *wasm.Store
 }
 ```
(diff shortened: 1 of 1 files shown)


RATIONALE.md
```diff
@@ -61,6 +62,131 @@ Ex. There is no exported symbol for the `[]byte` representing the `CodeSection`
...
+
+### Why aren't configuration assigned with option types?
+The option pattern is a familiar one in Go. For example, someone defines a type `func (x X) err` and uses it to update
+the target. For example, you could imagine wazero could choose to make `ModuleConfig` from options vs chaining fields.
+
+Ex instead of:
+```go
+type ModuleConfig interface {
+	WithName(string) ModuleConfig
+    WithFS(fs.FS) ModuleConfig
```
mathetake, line 155: tab vs spaces 😄 
```suggestion
	WithName(string) ModuleConfig
	WithFS(fs.FS) ModuleConfig
```

mathetake (approved):

PR #524, opened 2022-05-04
examples: switches tinygo example to use slice header

codefromthecrypt:
We get undefined behavior, at least in code that re-slices, unless
capacity is set. This ensures the buffer under the string has a capacity
set to the same length as its size. This also shows more explicitly the
problems in TinyGo (ex some type mismatch you have to ignore until
fixed).


mathetake (approved):

examples/allocation/tinygo/testdata/greet.go
```diff
@@ -56,13 +56,14 @@ func _greeting(ptr, size uint32) (ptrSize uint64) {
...
-	// Here, we want to get a string represented by the ptr and size. If we
-	// wanted a []byte, we'd use reflect.SliceHeader instead.
-	strHdr := (*reflect.StringHeader)(unsafe.Pointer(&ret))
-	strHdr.Data = uintptr(ptr)
-	strHdr.Len = uintptr(size)
-	return
+func ptrToString(ptr uint32, size uint32) string {
+	// Get a slice view of the underlying bytes in the stream. We use SliceHeader, not StringHeader
+	// as it allows us to fix the capacity to what was allocated.
+	return *(*string)(unsafe.Pointer(&reflect.SliceHeader{
```
codefromthecrypt, line 62: cc @wuhuizuo just in case you've not run into this, yet

Issue #483, opened 2022-04-19
Is it possible to grow api.Memory programmatically?


danielpacak:
There are scenarios where we may grow shared memory programmatically:

```go
module, _ := v.runtime.InstantiateModuleFromCode(ctx, policy)
module.Memory().Grow(1) // error: unresolved reference 'Grow'
```

However, the `Grow` method is not defined on the `api.Memory` interface. It's only defined on the internal `MemoryInstance` struct:

```go
func (m *MemoryInstance) Grow(newPages uint32) (result uint32) {
	// [...]
}
```

Is it intentionally designed like that, or there is another API to grow shared memory?

codefromthecrypt:
memory.grow is an instruction supported in wasm, just we haven't exported it into the host api.

Generally, we try to keep the API relevant by waiting for people to ask for things that aren't obvious. Can you mention the specific use cases you have in mind (curious)?

@mathetake is there anything tracking memory in JIT that exposing this as a host api would endanger? I guess the main thing is that if we put it in the host api it could be called outside the scope of a function..

codefromthecrypt:
ps the current workaround would be to make a function defined in %.wat that does memory.grow and export that, you could then call that from your host function. That said, I think we'll likely be able to add this, just need a second brain on if out-of-function scope will be a problem or not.

mathetake:
no, this isn't about how we implement, but growing memory must only be done by Wasm's memory.grow instruction. For example, otherwise language runtime won't be able to track the current number of pages in memory.

For example, in tinygo https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/src/runtime/arch_tinygowasm.go#L55 runtime's `growMemory()` is the only callsite of memory.grow instruction and the runtime tracks the current pages size there. If the pages size was changed by host, TinyGo's runtime can't know the current pages and won't be able to do allocation correctly. 


That said, growing memory in host API should't be allowed IMO.

codefromthecrypt:
@mathetake interesting points. So, I guess one issue is if you don't have a runtime, then there isn't a problem per-se (%.wat), it is mainly that if there is an allocation library (rust or tinygo), you should defer to that for memory.grow. In other words export an allocation function and use that instead of manually growing pages? Do I get it right?

mathetake:
yep, but even if there's no runtime issue, I can't find any rationale for host function being allowed to grow memory in the specification. looks like growing memory is only described with `memory.grow`.

[...]

PR #527, opened 2022-05-05
Documents Memory.Read as a way to share memory between Go and Wasm

### LINKED ISSUE #525: Less memory copies with Memory.MapSlice(offset, []byte)
Folks are doing more work moving data into the memory sandbox addressable by Wasm.

For example, @knqyf263 is working on [structured encoding][1] and @inkeliz is working on [random-access encoding][2]. Structs aside, the problem exists even with strings. You need to pass them into wasm (currently a copy) and its sandbox shouldn't be violated in the process..

Some alternatives solve this by providing unsafe access to the entire memory (ex a []byte). However, doing this breaks the sandbox and also any access controls you may have around boundaries. Furthermore, it may confuse who owns allocation, and typically that's the wasm code itself (ex TinyGo or rust malloc/free functions). Finally, it cements the underlying structure of memory, making changing it, ex [to pages of byte arrays][3], difficult later.

Another way could be to scope a memory overlay, given an offset and a range. For example, you could call an exported function to allocate memory, then apply an overlay which reads from that instead of the underlying source.

```go
var thing []byte = // something to manage in go

// Reserve linear memory using ex TinyGo's allocator.
results, _ := malloc.Call(ctx, len(thing))
ptr := results[0]
defer free.Call(ctx, ptr)

// map the slice so that until unmap is called, any writes to this region are write-through regardless
// of whether it is updated by Go or Wasm. Since this block is after malloc and before free, TinyGo
// shouldn't attempt to clear it.
unmap := mod.Memory().MapSlice(ptr, thing)
defer unmap()
```

Note this pattern could be accomplished many ways. For example if there is global mutable state,
the size could be indicated by a Global which is allocated during an "init" function. The actual
mapping could be done in a host function, which makes it possible to do in a .start function.
Regardless of how this is done, it could allow more precise handing over control of a region to
the host vs exposing the entire memory.

To do this internally may be trickier as there could be many stacks of these layers, so I'm not
entirely sure we want to even do this (think JIT). However, I figured we can think about it a while
and gather feedback about options.

[1]: #523
[2]: https://github.com/tetratelabs/wazero/issues/523#issuecomment-1117634602
[3]: #288

codefromthecrypt:
By testing the side-effects of Memory.Read, we ensure users who control
the underlying memory capacity can use the returned slice for
write-through access to Wasm addressible memory. Notably, this allows a
shared fixed length data structure to exist with a pointer on the Go
side and a memory offset on the Wasm side.

Thanks to @inkeliz for this idea and help on the code.

Fixes #525

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -215,6 +215,20 @@ type Memory interface {
 	ReadFloat64Le(ctx context.Context, offset uint32) (float64, bool)
 
 	// Read reads byteCount bytes from the underlying buffer at the offset or returns false if out of range.
+	//
+	// This returns a view of the underlying memory, not a copy. This means any writes to the slice returned are visible
+	// to Wasm, and any updates from Wasm are visible reading the returned slice.
+	//
+	// For example:
+	//	buf, _ = memory.Read(ctx, offset, byteCount)
+	//	buf[1] = 'a' // writes through to memory, meaning Wasm code see 'a' at that position.
+	//
+	// If you don't desire this behavior, make a copy of the returned slice before affecting it.
+	//
+	// Note: The returned slice is no longer shared on a capacity change. For example, `buf = append(buf, 'a')` results
+	// in a slice that is no longer shared. The same exists Wasm side. For example, if Wasm changes its memory capacity,
+	// ex via "memory.grow"), the host slice is no longer shared. Those who need a stable view must set Wasm memory
+	// min=max, or use wazero.RuntimeConfig WithMemoryCapacityPages to ensure max is always allocated.
 	Read(ctx context.Context, offset, byteCount uint32) ([]byte, bool)
 
 	// WriteByte writes a single byte to the underlying buffer at the offset in or returns false if out of range.
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -225,7 +225,7 @@ type Memory interface {
 	//
 	// If you don't desire this behavior, make a copy of the returned slice before affecting it.
 	//
-	// Note: The returned slice is no longer shared on a capacity change. For example, `buf = append(buf, 'a')` results
+	// Note: The returned slice is no longer shared on a capacity change. For example, `buf = append(buf, 'a')` might result
 	// in a slice that is no longer shared. The same exists Wasm side. For example, if Wasm changes its memory capacity,
 	// ex via "memory.grow"), the host slice is no longer shared. Those who need a stable view must set Wasm memory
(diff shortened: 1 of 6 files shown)


internal/testing/enginetest/enginetest.go
```diff
@@ -506,6 +506,89 @@ wasm stack trace:
...
+
+	hostPhrase := "Goodbye, cruel world. I'm off to join the circus." // Intentionally slightly longer.
+	hostPhraseSize := uint32(len(hostPhrase))
+
+	// Copy over the buffer, which should stop short.
+	copy(buf, hostPhrase)
+	require.Equal(t, "Goodbye, cruel world. I'm off to join the circ", string(buf))
+
+	// Append to the buffer we got from Wasm
+	buf = append(buf, 'u', 's', '.')
```
codefromthecrypt, line 576: @inkeliz per chat on #525, I started tests and notice this leaks. I've not started to fix it, yet. My guess is we need to use SliceHeader to change the capacity to the bytecount, before returning from Memory.Read.

inkeliz: You can set the capacity.

So, `Read` must return:

```
	return m.Buffer[offset : offset+byteCount : offset+byteCount], true
```

Notice, there's two `:`, the second is the capacity.

codefromthecrypt: ah thanks!

api/wasm.go
```diff
@@ -215,6 +215,20 @@ type Memory interface {
...
+	// This returns a view of the underlying memory, not a copy. This means any writes to the slice returned are visible
+	// to Wasm, and any updates from Wasm are visible reading the returned slice.
+	//
+	// For example:
+	//	buf, _ = memory.Read(ctx, offset, byteCount)
+	//	buf[1] = 'a' // writes through to memory, meaning Wasm code see 'a' at that position.
+	//
+	// If you don't desire this behavior, make a copy of the returned slice before affecting it.
+	//
+	// Note: The returned slice is no longer shared on a capacity change. For example, `buf = append(buf, 'a')` results
```
mathetake, line 228: precisely append within a capacity won't change the underlying slice so....

mathetake: maybe? 
```suggestion
	// Note: The returned slice is no longer shared on a capacity change. For example, `buf = append(buf, 'a')` might result
```

mathetake (approved):

codefromthecrypt:
@danielpacak after this, we have the right underlying tests to implement #483 (exposing api.Memory Grow)

PR #528, opened 2022-05-05
Fix bytes capacity on Read

inkeliz:
Before that change it was possible to overwrite content out of specified range.

See #527


codefromthecrypt (approved):
nice!

codefromthecrypt:
Thanks for contributing the commit!

PR #529, opened 2022-05-05
Adds ad-hoc test on bounds check with memory capacity config

mathetake:
Previously, we haven't had the tests to verify that memory bounds check work correctly 
with memory capacity configuration. Notably, this ensures that memory access offsets 
larger than size but less than capacity actually result in memory out of bounds access.


codefromthecrypt (approved):
LGTM

PR #534, opened 2022-05-09
Adds Runtime.Close

### LINKED ISSUE #382: Add Runtime.Close
I noticed that users of wazero including ourselves often have a single imported module (WASI) and the one they made. To close requires them closing those in reverse order. I wonder if we can't have `Runtime.Close` which can do the same?

An alternative would be if WASI could be special cased to never need closing. I don't know off the top of my head if we could pre-compile any glue needed and/or use offsets based on a go:embed FS or any other tricks to make WASI be safe to always be around.

anuraaga:
Fixes #382

It can be tedious to close many modules that have been instantiated, particularly when they are not instantiated in a scope where it is appropriate to use `defer`. `Runtime.Close` closes all the modules in a runtime's store, in reverse order of instantiation. I am wondering whether we should provide a version of this that takes an exit code like the `Module` version or if that would be overkill. It wouldn't be hard to add that if we need it here.

To keep track of order, `moduleNames` changes from a map to a slice - with a huge number of modules this could reduce instantiation performance. If this seems like it needs to be avoided, I can use a separate field instead, which would be something like inlining a linked map of modules into the store.

Also because Go locks aren't reentrant, I had to provide an internal module close codepath that doesn't remove from the store. There is a bit of circularity in that a module calls into store to remove itself, but when closing the store it is the other way around. Perhaps we may consider changing `Module.Close` to `Runtime.CloseModule(Module)` as a random idea.

### DIFF
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -413,6 +413,7 @@ func (s *Store) deleteModule(moduleName string) {
 	s.mux.Lock()
 	defer s.mux.Unlock()
 	delete(s.modules, moduleName)
+	// remove this module name
 	for i, n := range s.moduleNames {
 		if n == moduleName {
 			s.moduleNames = append(s.moduleNames[:i], s.moduleNames[i+1:]...)
@@ -627,13 +628,14 @@ func (s *Store) getFunctionTypeID(t *FunctionType) (FunctionTypeID, error) {
 	return id, nil
 }
 
-func (s *Store) Close(ctx context.Context) error {
+// CloseWithExitCode implements the same method as documented on wazero.Runtime.
+func (s *Store) CloseWithExitCode(ctx context.Context, exitCode uint32) (err error) {
 	s.mux.Lock()
 	defer s.mux.Unlock()
-	var err error
 	// Close modules in reverse initialization order.
 	for i := len(s.moduleNames) - 1; i >= 0; i-- {
-		if _, e := s.modules[s.moduleNames[i]].CallCtx.close(ctx, 0); e != nil && err == nil {
+		// If closing this module errs, proceed anyway to close the others.
+		if _, e := s.modules[s.moduleNames[i]].CallCtx.close(ctx, exitCode); e != nil && err == nil {
 			err = e // first error
 		}
 	}
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -633,10 +633,11 @@ func (s *Store) Close(ctx context.Context) error {
 	var err error
 	// Close modules in reverse initialization order.
 	for i := len(s.moduleNames) - 1; i >= 0; i-- {
-		_, err = s.modules[s.moduleNames[i]].CallCtx.close(ctx, 0)
+		if _, e := s.modules[s.moduleNames[i]].CallCtx.close(ctx, 0); e != nil && err == nil {
+			err = e // first error
+		}
 	}
 	s.moduleNames = nil
 	s.modules = map[string]*ModuleInstance{}
-	// Only returns last error.
 	return err
 }
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -33,7 +33,7 @@ type (
 		Engine Engine
 
 		// moduleNames ensures no race conditions instantiating two modules of the same name
-		moduleNames map[string]struct{} // guarded by mux
+		moduleNames []string // guarded by mux
 
 		// modules holds the instantiated Wasm modules by module name from Instantiate.
 		modules map[string]*ModuleInstance // guarded by mux
@@ -309,7 +309,7 @@ func NewStore(enabledFeatures Features, engine Engine) *Store {
 	return &Store{
 		EnabledFeatures:  enabledFeatures,
 		Engine:           engine,
-		moduleNames:      map[string]struct{}{},
+		moduleNames:      nil,
 		modules:          map[string]*ModuleInstance{},
 		typeIDs:          map[string]FunctionTypeID{},
 		functionMaxTypes: maximumFunctionTypes,
@@ -413,7 +413,12 @@ func (s *Store) deleteModule(moduleName string) {
 	s.mux.Lock()
 	defer s.mux.Unlock()
 	delete(s.modules, moduleName)
-	delete(s.moduleNames, moduleName)
+	for i, n := range s.moduleNames {
+		if n == moduleName {
+			s.moduleNames = append(s.moduleNames[:i], s.moduleNames[i+1:]...)
+			break
+		}
+	}
 }
 
 // requireModuleName is a pre-flight check to reserve a module.
@@ -421,10 +426,12 @@ func (s *Store) deleteModule(moduleName string) {
 func (s *Store) requireModuleName(moduleName string) error {
 	s.mux.Lock()
 	defer s.mux.Unlock()
-	if _, ok := s.moduleNames[moduleName]; ok {
-		return fmt.Errorf("module %s has already been instantiated", moduleName)
+	for _, n := range s.moduleNames {
+		if n == moduleName {
+			return fmt.Errorf("module %s has already been instantiated", moduleName)
+		}
 	}
-	s.moduleNames[moduleName] = struct{}{}
+	s.moduleNames = append(s.moduleNames, moduleName)
 	return nil
 }
 
@@ -618,4 +625,18 @@ func (s *Store) getFunctionTypeID(t *FunctionType) (FunctionTypeID, error) {
 		s.typeIDs[key] = id
 	}
 	return id, nil
+}
+
+func (s *Store) Close(ctx context.Context) error {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	var err error
+	// Close modules in reverse initialization order.
+	for i := len(s.moduleNames) - 1; i >= 0; i-- {
+		_, err = s.modules[s.moduleNames[i]].CallCtx.close(ctx, 0)
+	}
+	s.moduleNames = nil
+	s.modules = map[string]*ModuleInstance{}
+	// Only returns last error.
+	return err
 }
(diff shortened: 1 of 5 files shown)


internal/wasm/store.go
```diff
@@ -619,3 +626,17 @@ func (s *Store) getFunctionTypeID(t *FunctionType) (FunctionTypeID, error) {
...
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	var err error
+	// Close modules in reverse initialization order.
+	for i := len(s.moduleNames) - 1; i >= 0; i-- {
+		_, err = s.modules[s.moduleNames[i]].CallCtx.close(ctx, 0)
+	}
+	s.moduleNames = nil
+	s.modules = map[string]*ModuleInstance{}
+	// Only returns last error.
```
anuraaga, line 640: I couldn't find a pattern for returning multiple errors so did this as an initial try. Let me know if there's a way we would want to combine the errors or go with it.

codefromthecrypt: Probably we've not been as explicit in our code here as func-e, but we should use trailing comments and return first err (vs try to combine etc).

ex. from func-e.. I'm pretty sure we do this somewhere in wazero, but didn't trailing comment
```
	var err error
	go func() {
		if e := r.Run(ctx, args); e != nil && err == nil {
			err = e // first error
		}
		ran <- true
	}()
```

internal/wasm/store.go
```diff
@@ -413,18 +413,25 @@ func (s *Store) deleteModule(moduleName string) {
 	s.mux.Lock()
 	defer s.mux.Unlock()
 	delete(s.modules, moduleName)
-	delete(s.moduleNames, moduleName)
+	for i, n := range s.moduleNames {
```
codefromthecrypt, line 416:
```suggestion
	// remove this module name
	for i, n := range s.moduleNames {
```

internal/wasm/store.go
```diff
@@ -619,3 +626,18 @@ func (s *Store) getFunctionTypeID(t *FunctionType) (FunctionTypeID, error) {
 	}
 	return id, nil
 }
+
+func (s *Store) Close(ctx context.Context) error {
```
codefromthecrypt, line 630:
```suggestion
// Close implements the same method as documented on wazero.Runtime.
func (s *Store) Close(ctx context.Context) error {
```

internal/wasm/store.go
```diff
@@ -619,3 +626,18 @@ func (s *Store) getFunctionTypeID(t *FunctionType) (FunctionTypeID, error) {
...
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	var err error
+	// Close modules in reverse initialization order.
```
codefromthecrypt, line 634: ❤️

internal/wasm/store.go
```diff
@@ -619,3 +626,18 @@ func (s *Store) getFunctionTypeID(t *FunctionType) (FunctionTypeID, error) {
 	}
 	return id, nil
 }
+
+func (s *Store) Close(ctx context.Context) error {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	var err error
```
codefromthecrypt, line 633: probably you can name the result err for one less line :D

internal/wasm/store.go
```diff
@@ -619,3 +626,18 @@ func (s *Store) getFunctionTypeID(t *FunctionType) (FunctionTypeID, error) {
...
+	// Close modules in reverse initialization order.
+	for i := len(s.moduleNames) - 1; i >= 0; i-- {
+		if _, e := s.modules[s.moduleNames[i]].CallCtx.close(ctx, 0); e != nil && err == nil {
```
codefromthecrypt, line 636:
```suggestion
		// If closing this module errs, proceed anyway to close the others.
		if _, e := s.modules[s.moduleNames[i]].CallCtx.close(ctx, 0); e != nil && err == nil {
```

internal/wasm/store_test.go
```diff
@@ -182,6 +182,38 @@ func TestStore_CloseModule(t *testing.T) {
 	}
 }
 
+func TestStore_CloseStore(t *testing.T) {
```
codefromthecrypt, line 185: maybe make this two tests.. ex `t.Run("nothing closed", func...` and `t.Run("partially closed", func...` where the latter closed the module importing but not the imported. this would be the case if someone forgot to close wasi

wasm.go
```diff
@@ -92,6 +92,18 @@ type Runtime interface {
...
+	// An error is returned if any module returns an error when closed.
+	//
+	// Ex.
+	//	ctx := context.Background()
+	//  r := wazero.NewRuntime()
+	//  defer r.Close(ctx)
+	//	_, _ = r.InstantiateModuleFromCode(ctx, sourceLib)
+	//	mod, _ := r.InstantiateModuleFromCode(ctx, sourceApp)
+	//
+	Close(ctx context.Context) error
```
codefromthecrypt, line 106:
```suggestion
	Close(context.Context) error
```

wasm.go
```diff
@@ -92,6 +92,18 @@ type Runtime interface {
...
 	InstantiateModule(ctx context.Context, compiled CompiledModule, config ModuleConfig) (api.Module, error)
+
+	// Close closes all the modules that have been initialized in this Runtime with an exit code of 0.
+	// An error is returned if any module returns an error when closed.
+	//
+	// Ex.
+	//	ctx := context.Background()
+	//  r := wazero.NewRuntime()
+	//  defer r.Close(ctx)
+	//	_, _ = r.InstantiateModuleFromCode(ctx, sourceLib)
```
codefromthecrypt, line 103:
```suggestion
	//	_, _ = wasi.InstantiateSnapshotPreview1(ctx, r)
```

codefromthecrypt, line 103: make this more likely, ex they are using wasi and don't care :D

```suggestion
	//	_, _ = wasi.InstantiateSnapshotPreview1(ctx, r)
```

codefromthecrypt (commented):
Looks good! I have a couple requests.

1. implement `CloseWithExitCode` for parity with module. Ex this could help if someone's env.abort handler decides to close everything.

2. Make sure CompileCode.Close is done on everything (as that's the only other thing that can leak)

codefromthecrypt:
this is gonna be a *lot* less code in [wapc-go](https://github.com/wapc/wapc-go). can't wait! cc @pkedy

wasm_test.go
```diff
@@ -529,6 +529,72 @@ func TestInstantiateModule_ExitError(t *testing.T) {
...
+		if tc.exitCode == 0 {
+			err = r.Close(testCtx)
+		} else {
+			err = r.CloseWithExitCode(testCtx, tc.exitCode)
+		}
+		require.NoError(t, err)
+
+		// Modules closed so calls fail
+		_, err = func1.Call(testCtx)
+		require.Error(t, err, sys.NewExitError("mod1", tc.exitCode).Error())
```
anuraaga, line 574: Tried `ErrorIs` but it didn't work, can debug more if that seems unexpected

anuraaga:
Thanks @codefromthecrypt think I addressed all the comments

wasm.go
```diff
@@ -92,6 +92,30 @@ type Runtime interface {
 	// Note: Config is copied during instantiation: Later changes to config do not affect the instantiated result.
 	// Note: When the context is nil, it defaults to context.Background.
 	InstantiateModule(ctx context.Context, compiled CompiledModule, config ModuleConfig) (api.Module, error)
+
+	// Close closes all the modules that have been initialized in this Runtime with an exit code of 0.
+	// An error is returned if any module returns an error when closed.
+	//
+	// Ex.
+	//	ctx := context.Background()
+	//  r := wazero.NewRuntime()
```
codefromthecrypt, line 101:
```suggestion
	//	r := wazero.NewRuntime()
```

wasm.go
```diff
@@ -92,6 +92,30 @@ type Runtime interface {
...
+	//  defer r.Close(ctx)
```
codefromthecrypt, line 102:
```suggestion
	//	defer r.Close(ctx)
```

wasm.go
```diff
@@ -92,6 +92,30 @@ type Runtime interface {
...
+	//	mod, _ := r.InstantiateModuleFromCode(ctx, source)
+	//
+	Close(context.Context) error
+
+	// CloseWithExitCode closes all the modules that have been initialized in this Runtime with the provided exit code.
+	// An error is returned if any module returns an error when closed.
+	//
+	// Ex.
+	//	ctx := context.Background()
+	//  r := wazero.NewRuntime()
```
codefromthecrypt, line 113:
```suggestion
	//	r := wazero.NewRuntime()
```

wasm.go
```diff
@@ -92,6 +92,30 @@ type Runtime interface {
...
+	//  defer r.CloseWithExitCode(ctx, 2)
```
codefromthecrypt, line 114:
```suggestion
	//	defer r.CloseWithExitCode(ctx, 2)
```

wasm_test.go
codefromthecrypt, line 574:
```suggestion
 		// TODO: This could be neater with ErrorIs
		require.Error(t, err, sys.NewExitError("mod1", tc.exitCode).Error())
```

wasm_test.go
```diff
@@ -529,6 +529,72 @@ func TestInstantiateModule_ExitError(t *testing.T) {
...
+
+		_, err = func2.Call(testCtx)
+		require.Error(t, err, sys.NewExitError("mod2", tc.exitCode).Error())
+	}
+}
+
+func TestCloseClosesCompiledModules(t *testing.T) {
```
codefromthecrypt, line 581:
```suggestion
func TestClose_ClosesCompiledModules(t *testing.T) {
```

codefromthecrypt (approved):
only formatting nits. thanks a lot!

mathetake (approved):
🕺

wasm.go
anuraaga: I really need to prioritize figuring out how to insert a tab with my Emacs keybindings...

mathetake:
@anuraaga merge it once you feel ready!

PR #536, opened 2022-05-09
Sets lf line endings in gitattributes.

anuraaga:
Also removes unused linguist setting for stats

When developing on Windows, git will often give warnings like

`warning: CRLF will be replaced by LF in internal/wasm/text/decoder.go.`

and `make format` touches all files causing a huge "no-op diff" where git is seeing different line endings, but they don't actually show up in a PR.

I've generally not been able to have good results on developing on Windows without this `.gitattributes` line to force LF-always, which is appropriate for any modern codebase (even Windows ones :-) )

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,2 @@
-# Ignore the auto-generated Wasm spectest files from language statistics of wazero repo.
-# https://stackoverflow.com/questions/19052834/is-it-possible-to-exclude-files-from-git-language-statistics
-wasm/internal/spectests/cases/* linguist-vendored
+# Improves experience of commands like `make format` on Windows
+* text=auto eol=lf


.gitattributes
```diff
@@ -1,3 +1,5 @@
+* text=auto eol=lf
```
codefromthecrypt, line 1: maybe revise this comment as necessary

```suggestion
# prevents windows runners from failing on lint
* text=auto eol=lf
```

codefromthecrypt (approved):
thanks!

.gitattributes
```diff
@@ -1,3 +1,5 @@
...
+
 # Ignore the auto-generated Wasm spectest files from language statistics of wazero repo.
 # https://stackoverflow.com/questions/19052834/is-it-possible-to-exclude-files-from-git-language-statistics
 wasm/internal/spectests/cases/* linguist-vendored
```
mathetake, line 5: unrelated: I just realized that this drifted but somehow it's still working 😄

codefromthecrypt: cute.. I guess we should delete it. stats stuff are opaque anyway

.gitattributes
anuraaga: Tweaked the comment and added - indeed I was surprised that `make lint` seems to not fail, but there seem to be some other issues related to the setting

mathetake (approved):

codefromthecrypt:
thx!

PR #541, opened 2022-05-11
Check-in v2 spectest cases

mathetake:
This commit checks-in all the specification test cases for v2 Wasm spec,
including SIMD. Currently, we do not run the tests actually as we need to
fix several places to pass spectests plus we haven't implemented SIMD yet.

I wanted to have this PR separate from removing post1_0 tests and fixes 
as that would make it impossible for reviews to review on the actual code.


mathetake:
LOL on 200k loc

internal/integration_test/spectest/v2/spec_test.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+//go:embed testdata/*.wasm
+//go:embed testdata/*.json
+var testcases embed.FS //nolint:unused
+
+const enabledFeatures = wasm.Features20220419
+
+func TestJIT(t *testing.T) {
+	t.Skip() // TODO!
```
mathetake, line 20: skipped the actual test for now as I found that I need to patch several places in a different PRs.

codefromthecrypt (approved):
sound plan

PR #542, opened 2022-05-11
Pass all non-SIMD v2 core specification tests 

mathetake:
This commit enables WebAssembly 2.0 Core Specification tests.
In order to pass the tests, this fixes several places mostly on the
validation logic. 

Also, this removes pos1_0 integration tests as they are duplication of
spectets, though they are better in terms of devx.

Note that SIMD instructions are not implemented yet.

part of #484


internal/integration_test/spectest/v2/testdata/spec_test.go
```diff
@@ -1,30 +0,0 @@
-package spectest
```
mathetake, line 1: this is garbaged accidentally checked in

.github/workflows/spectest.yaml
```diff
@@ -84,7 +84,7 @@ jobs:
       - name: Build scratch container
         run: |
           echo 'FROM scratch' >> Dockerfile
-          echo 'CMD ["/test"]' >> Dockerfile
+          echo 'CMD ["/test", "-test.v"]' >> Dockerfile
```
mathetake, line 87: make it verbose!

internal/wasm/module.go
```diff
@@ -319,18 +319,73 @@ func (m *Module) validateFunctions(enabledFeatures Features, functions []Index,
...
+		if err := m.validateFunction(enabledFeatures, Index(idx), functions, globals, memory, tables, declaredFuncIndexes); err != nil {
 			return fmt.Errorf("invalid %s: %w", m.funcDesc(SectionIDFunction, Index(idx)), err)
 		}
 	}
 	return nil
 }
 
+// declaredFunctionIndexes returns a set of function indexes that can be used as an immediate for OpcodeRefFunc instruction.
+//
+// The criteria for which function indexes can be avaiable for that instruction is vague in the spec described as:
```
codefromthecrypt, line 341:
```suggestion
// The criteria for which function indexes can be available for that instruction is vague in the spec:
```

internal/wasm/module.go
```diff
@@ -319,18 +319,73 @@ func (m *Module) validateFunctions(enabledFeatures Features, functions []Index,
...
+//
+//  * "References: the list of function indices that occur in the module outside functions and can hence be used to form references inside them."
+//   * https://www.w3.org/TR/2022/WD-wasm-core-2-20220419/valid/conventions.html#contexts
+//  * "Ref is the set funcidx(module with functions=ε, start=ε) , i.e., the set of function indices occurring in the module, except in its functions or start function."
+//   * https://www.w3.org/TR/2022/WD-wasm-core-2-20220419/valid/modules.html#valid-module
+//
+// That said, the logic here is derived from the specification test case in WebAssembly 2.0:
```
codefromthecrypt, line 348:
```suggestion
// To clarify, we reverse-engineer logic required to pass the WebAssembly Core specification 2.0 test suite:
```

internal/wasm/module_test.go
```diff
@@ -828,3 +829,77 @@ func TestModule_validateDataCountSection(t *testing.T) {
 		}
 	})
 }
+
+func TestModule_declaredFunctionIndexes(t *testing.T) {
```
codefromthecrypt, line 833: thanks for putting in unit tests for this! can you make a test for each of the main special cases noted in the comment (include test name field please)

codefromthecrypt (approved):
This is great. This means we are running both spectests 1.0 and 2.0 on the same codebase. I'm not sure any other runtime does this, even if historically they may have. Regardless, it is good to set a good example, and things like this are setting a good example. Bravo!

internal/integration_test/vs/bench_factorial.go
```diff
@@ -8,17 +8,15 @@ import (
...
+	// catFS is an embedded filesystem limited to test.txt
+	//go:embed testdata/fac.wasm
+	factorialWasm   []byte
+	factorialParam  = uint64(30)
+	factorialResult = uint64(9682165104862298112)
+	factorialConfig *RuntimeConfig
 )
 
 func init() {
-	factorialWasm = readRelativeFile(factorialWasmPath)
```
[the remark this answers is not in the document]
mathetake: still used by tinygo allocation vs 😄

PR #543, opened 2022-05-11
Fixes coverage on externref/uintptr

codefromthecrypt:
thx @anuraaga for noticing the bug

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -327,6 +327,18 @@ type Memory interface {
 	Write(ctx context.Context, offset uint32, v []byte) bool
 }
 
+// EncodeExternref encodes the input as a ValueTypeExternref.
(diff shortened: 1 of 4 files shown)


api/wasm.go
```diff
@@ -327,6 +327,18 @@ type Memory interface {
 	Write(ctx context.Context, offset uint32, v []byte) bool
 }
 
+// EncodeExternref encodes the input as a ValueTypeExternref.
+// See DecodeExternref
+func EncodeExternref(input uintptr) uint64 {
```
codefromthecrypt, line 332: while this is trivial, it formalizes the documentation on ValueTypeExternref where uintptr is how we represent externtype

mathetake (approved):

PR #545, opened 2022-05-12
Implements errors.Is for sys.ExitError

anuraaga:
Allows use with test assertions like `require.ErrorIs`, which delegates to `errors.Is` both here and in testify.

Cleans up a related TODO.

### DIFF
--- a/sys/sys.go
+++ b/sys/sys.go
@@ -42,3 +42,10 @@ func (e *ExitError) ExitCode() uint32 {
 func (e *ExitError) Error() string {
 	return fmt.Sprintf("module %q closed with exit_code(%d)", e.moduleName, e.exitCode)
 }
+
+func (e *ExitError) Is(err error) bool {
+	if target, ok := err.(*ExitError); ok {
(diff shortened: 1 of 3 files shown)


sys/sys.go
```diff
@@ -42,3 +42,10 @@ func (e *ExitError) ExitCode() uint32 {
 func (e *ExitError) Error() string {
 	return fmt.Sprintf("module %q closed with exit_code(%d)", e.moduleName, e.exitCode)
 }
+
+func (e *ExitError) Is(err error) bool {
```
codefromthecrypt, line 46:
```suggestion
// Is allows use via errors.Is
func (e *ExitError) Is(err error) bool {
```

codefromthecrypt (approved):
looks good. maybe take a look at the godoc on ExitError to see if it should be revised or not!

anuraaga:
Thanks for the reminder - I'll leave it as is since it'll be common to only check the exit code instead of comparing to a full error outside of tests.

codefromthecrypt:
thx again!

PR #553, opened 2022-05-13
Adds test delegating to iotest against a wasi fs.Fs implementation

### LINKED ISSUE #408: wasi: add fs.FS integration test
Right now, we mostly focus on unit tests with a little ad-hoc here and there wrt file systems. It should be possible to add a `fs.FS` impl that calls WASI. Doing this would allow us to re-use existing tests for parity, like `iotest.TestReader`, as well show use cases like root vs workdir file usage operates with less clutter.

anuraaga:

Fixes #408

This implements `fs.Fs` and `io.Reader` / `io.Seeker` by calling into exported wasm functions that delegate to wasi. This let's use add integration tests that target `fs.Fs` as a generic mechanism for accessing files - in this PR we execute `iotest.TestReader` which runs some tests verifying the operation of a `Reader`. Because we haven't implemented `fd_stat` in wasi yet, this does not implement `File.Stat()` yet.

https://cs.opensource.google/go/go/+/master:src/testing/iotest/reader.go;l=136

### DIFF
--- a/internal/integration_test/fs/fs_test.go
+++ b/internal/integration_test/fs/fs_test.go
@@ -2,12 +2,12 @@ package fs
 
 import (
 	"context"
+	_ "embed"
 	"fmt"
 	"io"
 	"io/fs"
-	"os"
-	"path/filepath"
 	"testing"
+	"testing/fstest"
 	"testing/iotest"
 
 	"github.com/tetratelabs/wazero"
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
good job!

mathetake (approved):

PR #554, opened 2022-05-13
Adds a Closer interface to encapsulate the API for closing a re…

### LINKED ISSUE #551: Consider not returning api.Module when instantiating wasi
I noticed that the module returned from wasi instantiation isn't usable as a normal module - notably it doesn't have `Memory` or `Sys`, so trying to call exported functions doesn't work. wasi module is meant to be a supporting module imported by another wasm module. `Module` is likely returned to allow `Close` - now that we have `Runtime.Close` we could consider not returning a module when initializing wasi to prevent users (like me :-) ) from trying to use it as a normal module. It does mean that `Runtime.Close` stops being an alternative to closing modules but is required when using wasi - this may be natural because people often call `Close` anytime there is a `Close` method I guess.

It could also be ok to leave behavior as is if there is a reasonable runtime message. Unfortunately, wasi functions often requiring the host to write into `Memory` before calling. This means that the error can easily be a nil reference exception in the user code when trying to use `Memory`, without a good error message.

anuraaga:
…source with modules.

Fixes #551 

`api.Module` provides APIs for using a wasm module directly such as `Memory()` and `ExportedFunction`. Built-in modules such as wasi do not expose such functionality, but do still need to be closeable. `Closer` provides the functionality for closing a module and can be returned in these cases. It is also a natural fit with `Runtime`, which has the ability to close modules despite not being a module itself.

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -147,20 +147,8 @@ type Module interface {
 	// ExportedGlobal a global exported from this module or nil if it wasn't.
 	ExportedGlobal(name string) Global
 
-	// Closer closes this module.
-	Closer
-}
-
-// Closer closes a resource. Any modules closed with it will be provided an exit code.
-//
-// Note: This is an interface for decoupling, not third-party implementations. All implementations are in wazero.
-type Closer interface {
-	// Close is a convenience that invokes CloseWithExitCode with zero.
-	// Note: When the context is nil, it defaults to context.Background.
-	Close(context.Context) error
-
-	// CloseWithExitCode releases resources. The exit code will be provided to any released Modules.
-	// Use a non-zero exitCode parameter to indicate a failure to ExportedFunction callers.
+	// CloseWithExitCode releases resources allocated for this Module. Use a non-zero exitCode parameter to indicate a
+	// failure to ExportedFunction callers.
 	//
 	// The error returned here, if present, is about resource de-allocation (such as I/O errors). Only the last error is
 	// returned, so a non-nil return means at least one error happened. Regardless of error, this module instance will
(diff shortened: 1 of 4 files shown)


api/wasm.go
```diff
@@ -162,6 +146,30 @@ type Module interface {
...
 	ExportedGlobal(name string) Global
+
+	// ModuleCloser closes this module.
+	ModuleCloser
+}
+
+// ModuleCloser closes a module with an exit code.
+//
+// Note: This is an interface for decoupling, not third-party implementations. All implementations are in wazero.
+type ModuleCloser interface {
```
codefromthecrypt, line 157: hmm how about Closer? this also closes the runtime. shouldn't have any conflicts as `api.Closer` is different than say `io.Closer`

api/wasm.go
```diff
@@ -162,6 +146,30 @@ type Module interface {
...
+	// Close is a convenience that invokes CloseWithExitCode with zero.
+	// Note: When the context is nil, it defaults to context.Background.
+	Close(context.Context) error
+
+	// CloseWithExitCode releases resources allocated for a Module. Use a non-zero exitCode parameter to indicate a
```
codefromthecrypt, line 162: the main clarification when this is Closer not ModuleCloser is while `Runtime.CloseXXX` closes other things (including code under a module), when it is closing a module those will get this exit code.

anuraaga: Yeah forgot we're closing more than modules here.

anuraaga: Tweaked the doc to try to point out it can release many things, but when they're Modules they'll get the exit code

wasi/wasi.go
```diff
@@ -36,8 +36,8 @@ const ModuleSnapshotPreview1 = "wasi_snapshot_preview1"
 //	mod, _ := r.InstantiateModuleFromCode(ctx, source)
 //
 // Note: All WASI functions return a single Errno result, ErrnoSuccess on success.
-// Note: Closing the wazero.Runtime closes any api.Module it instantiated.
-func InstantiateSnapshotPreview1(ctx context.Context, r wazero.Runtime) (api.Module, error) {
+// Note: Closing the wazero.Runtime will release wasi as well.
```
anuraaga, line 39: Since we don't return `api.Module` here anymore, it didn't seem clear to talk about it in the doc

codefromthecrypt, line 39:
```suggestion
// Note: Closing the wazero.Runtime closes this instance of WASI as well.
```

codefromthecrypt (approved):
nit mainly to use the word close consistently

mathetake (approved):

PR #555, opened 2022-05-13
Adds SIMD opcodes and feature flag

mathetake:
part of #484 

### DIFF
--- a/config.go
+++ b/config.go
@@ -105,6 +105,12 @@ type RuntimeConfig interface {
 	// See https://github.com/WebAssembly/spec/blob/main/proposals/sign-extension-ops/Overview.md
 	WithFeatureSignExtensionOps(bool) RuntimeConfig
 
+	// WithFeatureSIMD enables the vector value type and vector instructions (aka SIMD).  This defaults to false
+	// as the feature was not in WebAssembly 1.0.
+	//
+	// See https://github.com/WebAssembly/spec/blob/main/proposals/simd/SIMD.md
+	WithFeatureSIMD(bool) RuntimeConfig
+
 	// WithWasmCore1 enables features included in the WebAssembly Core Specification 1.0. Selecting this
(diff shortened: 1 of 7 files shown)


config.go
```diff
@@ -105,6 +105,12 @@ type RuntimeConfig interface {
 	// See https://github.com/WebAssembly/spec/blob/main/proposals/sign-extension-ops/Overview.md
 	WithFeatureSignExtensionOps(bool) RuntimeConfig
 
+	// WithFeatureSIMD enables the vector value type and vector instructions (aka SIMD).  This defaults to false
+	// as the feature was not in WebAssembly 1.0.
+	//
+	// See https://github.com/WebAssembly/spec/blob/main/proposals/simd/SIMD.md
+	WithFeatureSIMD(bool) RuntimeConfig
```
codefromthecrypt, line 112: agree SIMD is the feature even if they discuss it as Vector instructions in the 2.0 changelog

internal/wasm/func_validation.go
```diff
@@ -1042,6 +1042,19 @@ func (m *Module) validateFunctionWithMaxStackValues(
 					valueTypeStack.push(r)
 				}
 			}
+		} else if op == OpcodeVecPrefix {
+			pc++
+			// Vector instructions come with two bytes which starts with OpcodeVecPrefix,
```
anuraaga, line 1047:
```suggestion
			// Vector instructions come with two bytes where the first byte is always OpcodeVecPrefix,
```

codefromthecrypt (approved):
lots of instructions :D

thanks for putting this in as it helps people know if they are using the only feature in 2.0 not fully implemented.

anuraaga (commented):
Just small suggestion on comment, was a bit confused reading it at first

internal/wasm/func_validation.go
mathetake: will apply your comment shortly! :D

PR #556, opened 2022-05-13
Implements v128.const and adds support for vector value type. 

mathetake:
This commit implements the `v128.const`, `i32x4.add` and  `i64x2.add` in
interpreter mode and this adds support for the vector value types in the 
locals and globals.

Notably, the vector type values can be passed and returned by exported functions
as well as host functions via two-uint64 encodings as described in https://github.com/tetratelabs/wazero/issues/484#issuecomment-1116866086.

Note: implementation of these instructions on JIT will be done in subsequent PR.

part of #484

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -59,7 +59,7 @@ func ExternTypeName(et ExternType) string {
 //  * ValueTypeI64 - uint64(int64)
 //  * ValueTypeF32 - EncodeF32 DecodeF32 from float32
 //  * ValueTypeF64 - EncodeF64 DecodeF64 from float64
-//  * ValueTypeVector TODO:
+//  * ValueTypeV128 TODO:
 //  * ValueTypeExternref - unintptr(unsafe.Pointer(p)) where p is any pointer type in Go (e.g. *string)
 //
 // Ex. Given a Text Format type use (param i64) (result i64), no conversion is necessary.
@@ -85,10 +85,10 @@ const (
 	ValueTypeF32 ValueType = 0x7d
 	// ValueTypeF64 is a 64-bit floating point number.
 	ValueTypeF64 ValueType = 0x7c
-	// ValueTypeVector is a 128-bit vector value.
+	// ValueTypeV128 is a 128-bit vector value.
 	//
 	// TODO
-	ValueTypeVector ValueType = 0x7b
+	ValueTypeV128 ValueType = 0x7b
 	// ValueTypeExternref is a externref type.
 	//
 	// Note: in wazero, externref type value are opaque raw 64-bit pointers, and the ValueTypeExternref type
@@ -117,7 +117,7 @@ func ValueTypeName(t ValueType) string {
 		return "f32"
 	case ValueTypeF64:
 		return "f64"
-	case ValueTypeVector:
+	case ValueTypeV128:
 		return "i128"
 	case ValueTypeExternref:
 		return "externref"
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -118,7 +118,7 @@ func ValueTypeName(t ValueType) string {
 	case ValueTypeF64:
 		return "f64"
 	case ValueTypeV128:
-		return "i128"
+		return "v128"
 	case ValueTypeExternref:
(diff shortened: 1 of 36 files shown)


internal/integration_test/spectest/spectest.go
```diff
@@ -136,24 +146,71 @@ func (c command) String() string {
...
-		exps = append(exps, exp.toUint64())
+		exps = append(exps, exp.toUint64s()...)
+		fmt.Println("exps", exps)
 	}
 	return args, exps
 }
 
+func (c commandActionVal) toUint64s() (ret []uint64) {
+	if c.ValType == "v128" {
+		strValues, ok := c.Value.([]interface{})
```
mathetake, line 168:
```suggestion
		strValues, ok := c.Value.(simdVal)
```

mathetake:
so basic handleings are done... the next is to support vector values in locals, then globals!

internal/integration_test/spectest/v2/spec_test.go
```diff
@@ -22,9 +24,18 @@ func TestJIT(t *testing.T) {
...
+		return !strings.Contains(jsonname, "simd")
+	})
 }
 
 func TestInterpreter(t *testing.T) {
-	spectest.Run(t, testcases, interpreter.NewEngine, enabledFeatures)
+	spectest.Run(t, testcases, interpreter.NewEngine, enabledFeatures, func(jsonname string) bool {
+		// TODO: remove after SIMD proposal
+		if strings.Contains(jsonname, "simd") {
+			return path.Base(jsonname) == "simd_const.json"
```
mathetake, line 37: interpreter impl now passes simd_const.wast https://github.com/tetratelabs/wazero/blob/main/internal/integration_test/spectest/v2/testdata/simd_const.wast

mathetake:
fixing unit tests....

api/wasm.go
```diff
@@ -112,6 +117,8 @@ func ValueTypeName(t ValueType) string {
 		return "f32"
 	case ValueTypeF64:
 		return "f64"
+	case ValueTypeVector:
```
codefromthecrypt, line 120: ValueTypeV128 would be the most common naming convention though internally some use i128

api/wasm.go
```diff
@@ -84,6 +85,10 @@ const (
 	ValueTypeF32 ValueType = 0x7d
 	// ValueTypeF64 is a 64-bit floating point number.
 	ValueTypeF64 ValueType = 0x7c
+	// ValueTypeVector is a 128-bit vector value.
+	//
+	// TODO
```
codefromthecrypt, line 90: maybe take from here? Might have to fix wrapping


```suggestion
	// The type corresponds to a 128 bit vector of packed integer or floating-point data.
	// The packed data can be interpreted as signed or unsigned integers, single or double
	// precision floating-point values, or a single 128 bit type. The interpretation is
	//  determined by individual operations.
	//
	// See https://www.w3.org/TR/2022/WD-wasm-core-2-20220419/syntax/types.html#syntax-vectype
```

codefromthecrypt: also should mention that this occupies 2 uint64 and due to little endian encoding the lower 64 bits are first

codefromthecrypt:
ps when going for api encoding, I think it would be something like this to match https://www.w3.org/TR/2022/WD-wasm-core-2-20220419/text/instructions.html#vector-instructions

```
api.EncodeV128_F64x2(float64, float64) (uint64, uint64)
... for the others.
```

Could be a TODO

codefromthecrypt:
by deferring the encoding to a utility, basically it allows things like Function.Call and Global to only deal with the problem of 128bit type (split across 2 uint64), not the complexity of V128 encoding of it.

codefromthecrypt (approved):
The way this PR is organized, there's very little external impact. While I'm not a fan of WebAssembly calling the type V128 (yet abstractly i128), it is what it is, and it is better to call it V128 vs VectorXXX (I checked the other impls on this too).

representing internally as low, high is sensible.

When you are ready I'm good with this.

PR #557, opened 2022-05-13
site: Adds first submenu for Community

codefromthecrypt:
This adds a submenu for community, filing GH and the history into it.
The same surgery should work for future submenus, for example,
compatibility and docs.

### DIFF
--- a/site/layouts/partials/menu.html
+++ b/site/layouts/partials/menu.html
@@ -1,0 +1,32 @@
+<nav class="menu">
+  <ul class="menu__inner menu__inner--desktop">
+    {{ range $.Site.Menus.main }}
+      {{ if not .HasChildren }}
+        <li><a href="{{ .URL }}">{{ .Name }}</a></li>
+      <!-- The default theme supports a drop-down for showMenuItems > 0. We re-use this class
+           for child menus instead. -->
+      {{ else }}
+        <ul class="menu__sub-inner">
+          <li class="menu__sub-inner-more-trigger">
+            {{ .Name }}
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
https://deploy-preview-557--wazero.netlify.app/

site/layouts/partials/menu.html
```diff
@@ -0,0 +1,32 @@
...
+              <li><a href="{{ .URL }}">{{ .Name }}</a></li>
+            {{ end }}
+          </ul>
+        </ul>
+      {{ end }}
+    {{ end }}
+  </ul>
+
+  <!-- This is the theme default layouts/partials/menu.html -->
+  <ul class="menu__inner menu__inner--mobile">
```
codefromthecrypt, line 25: there was no CSS for mobile submenus, so I decided to let this ride. another way could be to concatenate when children exist (ex About/GitHub About/History, etc)

anuraaga:
Looks like `Community` doesn't show on mobile just want to confirm that is intentional

image

codefromthecrypt:
wrt mobile not dropdown, well not so much intentional, but much more work as the underlying theme didn't have CSS etc for mobile. I can change to prefix it with a TODO for someone to make CSS

codefromthecrypt:
only issue with prefixing is that it will eventually become long. anyway this isn't worse for mobile, as before it truncated anything with children also. it fixes desktop though.

anuraaga (approved):

codefromthecrypt:
I did the inlining for now as at least that's easy. since mobile already makes a drop-down it wasn't so bad on my phone. good catch @anuraaga

PR #558, opened 2022-05-13
Move specs to the website

codefromthecrypt:


codefromthecrypt:
the drop-down only works once.. if you change the weight the same happens in the opposite order. maybe there's some state involved.

Screen Shot 2022-05-13 at 2 58 23 PM

https://deploy-preview-558--wazero.netlify.app/

the problem is likely in ./site/layouts/partials/logo.html or possibly the js generated by the theme (or both!)

codefromthecrypt:
Thanks for fixing this @anuraaga!

mathetake (approved):

PR #559, opened 2022-05-14
Adds badges on README

mathetake:


mathetake:
Screen Shot 2022-05-14 at 9 17 13

codefromthecrypt (approved):
cool badge. wonder if we can get one for each spec :) 1.0 vs 2.0?

mathetake:
Unfortunately badges are per workflow file so we can’t have separate ones for each version unless we have the separate workflow file for each😄

PR #562, opened 2022-05-16
Adds encoding API and support for host functions with v128 values

mathetake:


api/wasm.go
```diff
@@ -393,6 +393,92 @@ func DecodeF64(input uint64) float64 {
 	return math.Float64frombits(input)
 }
 
+// EncodeI8x16 encodes the input as a ValueTypeV128.
+func EncodeI8x16(i1, i2, i3, i4, i5, i6, i7, i8, i9, i10, i11, i12, i13, i14, i15, i16 int8) (low uint64, hi uint64) {
```
anuraaga, line 397: I suspect that when dealing with bytes, it will be more common to be working with `[]byte`. Would we provide another function for it?

anuraaga: Actually, I think for all of them (except maybe 64-bit), slice will be common, e.g. 

```go
var arr []byte
for i := 0; i < len(arr); i += 16 {
  low, hi := EncodeI8x16(arr[i:i+16])
  func.Call(low, hi)
}
```

```go
var arr []int16
for i := 0; i < len(arr); i += 8{
  low, hi := EncodeI16x18(arr[i:i+8])
  func.Call(low, hi)
}
```

Not sure if we should have both slice and non-slice versions but I think slice are important.

anuraaga: Ah I guess if `arr[i:i+16]...` works it's all good, oops

api/wasm_test.go
```diff
@@ -139,3 +139,122 @@ func TestEncodeCastI64(t *testing.T) {
...
+		{0xffffffff_ffffffff, 0},
+		{0xffffffff_ffffffff, 0xffffffff_ffffffff},
+		{0xffffffff_efffffff, 0xffffffff_ffffffff},
+		{0xffffffff_efffffff, 0xffffffff_efffffff},
+		{0xffff_ffff, 0xffff_ffff},
+		{0x1 << 4, 0x1 << 3},
+	} {
+		v := v
+		t.Run(fmt.Sprintf("%x", v), func(t *testing.T) {
+			low, hi := EncodeI8x16(DecodeI8x16(v[0], v[1]))
```
anuraaga, line 156: Related to my other comment, it may be worth having tests that create slices and call `Encode` directly to confirm UX for them. We've confirmed the UX of passing result of Decode into Encode but the unit tests can be good confirmation / documentation of expected usage patterns as well.

api/wasm.go
mathetake: yeah will change the signature!

api/wasm_test.go
mathetake: on it!

internal/wasm/store.go
```diff
@@ -511,7 +511,15 @@ func (s *Store) resolveImports(module *Module) (
 			importedFunction := imported.Function
 
 			actualType := importedFunction.Type
-			if !expectedType.EqualsSignature(actualType.Params, actualType.Results) {
+
+			var mismatch bool
+			if importedFunction.Kind != FunctionKindWasm {
+				// Host function takes two uint64 instead of a dedicated vector type in the signature.
+				mismatch = !expectedType.EqualsSignatureV128Flattened(actualType.Params, actualType.Results)
```
mathetake, line 518: So this seems weird special casing for host functions that want to take vector values as params or results ...

thoughts? @codefromthecrypt @anuraaga

mathetake: so basically with this we cannot judge, for example the host function `func myfunc(uint64, uint64)` should be imported for the signature either `(v128)` or `(i64, i64)` which we cannot disambiguate since we infer the host function's Wasm signature https://github.com/tetratelabs/wazero/blob/main/internal/wasm/gofunc.go#L240-L255

mathetake: though this is harmless :D

codefromthecrypt:
Vector types are prohibited in JavaScript, so there's no type binding there. Also V128 type != I128 (there's no i128 const). So, I want to check to see if there's any use case for vectorizing in host functions at all a sec, and that sounds suspicious as calling back to the host in a loop to use SIMD sounds backwards for perf. Anyway, I'll check and the result is that we may fail a signature rather than allow one.

codefromthecrypt:
Hi @mathetake I'm sorry that you did this work for possibly no reason. Can you pare this down so that the only thing we support publicly is global V128? probably high, low uint64 is the easiest way as unlike rust there's no u128 type in Go. Ex `Get128(ctx) (high, low uint64)`

Then inside host function documentation, mention that V128 is not a supported value type. In other words, we don't support binding, so we don't need to worry about marshaling of it until https://github.com/WebAssembly/spec/issues/1474

codefromthecrypt:
ps if you want to keep the code for internal reasons, that's ok, just let's make it internal and impossible to use it in host functions with the api until the mentioned issue resolves.

codefromthecrypt:
sorry to reverse on this, but the spec lead clarified that there's no rule restricting v128 in general for both global and host function https://github.com/WebAssembly/spec/issues/1474#issuecomment-1127768660

So, this leaves us with how to represent it. I think for starters in host functions use `[16]byte` for now, as unlikely anyone will use it anyway. Then, later someone could ask for a different representation. For globals, I think it is fine to use pair of uint64 as multi-return makes that fine.

My 2p

codefromthecrypt:
meanwhile the host-side vectorizing apis I really think we shouldn't add these as I can't imagine host side vectorization being anything except worse performance than without simd. I could be wrong and we could re-add it later if someone does that. Basically I think we can leave v128 opaque.

If folks feel strongly on keeping the host-side vectorization APIs, that's ok, just be careful to document this is unlikely to increase performance and could reverse it.

codefromthecrypt:
FWIW I searched as much as I could on github and could only find global v128 inside the wast itself. It makes me wonder really if it was the wast itself which was the use case leading to globals being supported. I may be jumping to conclusions also, just the lack of anything in GitHub is anchoring..

```webassembly
  (global $zero (mut v128) (v128.const i32x4 0 0 0 0))
  (func (export "v128.store8_lane_0")
    (param $address i32) (param $x v128) (result i64) (local $ret i64)
    (v128.store8_lane 0 (local.get $address) (local.get $x))
    (local.set $ret (i64.load (local.get $address)))
    (v128.store (local.get $address) (global.get $zero))    (local.get $ret))
;;snip
```
https://github.com/WebAssembly/spec/blob/main/test/core/simd/simd_store8_lane.wast#L6
ps why this is mutable no idea

Also, I found this source trying to figure out what the rules were:
https://github.com/WaterfoxCo/Waterfox/blob/current/js/src/jit-test/tests/wasm/simd/js-api.js

Here's the summary options we have (non exclusive)
* add global it because the spec is clear that globals v128 should be supported
* add vector intialization functions to the api (note: we can already pass a zero val!)
* add host function support regardless of if other impls such as wasmtime do

my feeling is stick with basic globals and wait for someone to ask for the vectorization logic.

codefromthecrypt:
last thing.. all the questioning above are about host defined extern types. obviously functions need to support v128 for vectorization to work. so, the main q here is about adding to our public api functions like `DecodeV128_I8x16` which agreed I asked about in the first place, just yeah the actual practical use is curious except matching parity with what you can do in wasm.

mathetake:
yeah agreed no usecase and wait unti someone asks for. Closing this and raise the Get128 API on the global in another PR!

PR #564, opened 2022-05-16
Renames JIT to Compiler and notes it is AOT

### LINKED ISSUE #560: terminology AOT vs JIT
I'm raising an issue not because it is vitally important, but because to do something about it is API affecting, and if we want to do it, we should do it now.

Right now, we conflate compiler with JIT, and one could argue how we compile is more like AOT(Ahead of Time) than JIT(Just in Time). This isn't to say others aren't also conflating, but I do think that even if others conflate this, we have a choice whether to or not.

A proposal I have is to rename "jit" to "compiler" resulting in runtimes being interpreter vs compiler. Then, when discussing the default compilation mode, we could say in docs it is AOT applied at `Runtime.Compile` instead of calling it JIT. In the future, we can introduce a JIT mode, and also we can easily transition to other AOT modes, such as being able to persist the result of Runtime.Compile for future use (See #179 and #421)

Since this has API impact, I'd like to get feedback both from other maintainers cc @mathetake @anuraaga as well end users cc @pims @pkedy @inkeliz @knqyf263 @sam-at-luther @efejjota @danielpacak @birros @wuhuizuo

The TL;DR; is rename JIT -> Compiler, and document it as AOT (Ahead of Time) compilation applied at `Runtime.CompileModule` until we support alternate modes, which would be indicated in RuntimeConfig or CompileConfig somehow.

My thinking is if there are a few thumbs up and no strong opinions against, renaming should happen. If there are no thumbs up, renaming shouldn't happen :D

codefromthecrypt:
This notably changes NewRuntimeJIT to NewRuntimeCompiler as well renames
packages from jit to compiler.

This clarifies the implementation is AOT, not JIT, at least when
clarified to where it occurs (Runtime.CompileModule). In doing so, we
reduce any concern that compilation will happen during function
execution. We also free ourselves to create a JIT option without
confusion in the future via CompileConfig or otherwise.

Fixes #560


codefromthecrypt:
ps I did my best on rewording JIT -> Compiler, but there will always be things we catch later also. Almost all changes are internal.

Makefile
```diff
@@ -3,19 +3,19 @@ golangci_lint := github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.0
 # sync this with netlify.toml!
 hugo          := github.com/gohugoio/hugo@v0.98.0
 
-ensureJITFastest := -ldflags '-X github.com/tetratelabs/wazero/internal/integration_test/vs.ensureJITFastest=true'
+ensureCompilerFastest := -ldflags '-X github.com/tetratelabs/wazero/internal/integration_test/vs.ensureCompilerFastest=true'
```
codefromthecrypt, line 6: ps still fastest :p

config.go
```diff
@@ -145,13 +145,22 @@ var engineLessConfig = &runtimeConfig{
 	enabledFeatures: wasm.Features20191205,
 }
 
-// NewRuntimeConfigJIT compiles WebAssembly modules into runtime.GOARCH-specific assembly for optimal performance.
+// NewRuntimeConfigCompiler WebAssembly modules into runtime.GOARCH-specific
```
codefromthecrypt, line 148: this is the only public API change

mathetake (approved):

PR #568, opened 2022-05-17
Renames asm packages to match folder name

anuraaga:

I was looking into possibly bringing in func-e's linting config. There are currently tons of failures so may proceed with that very gradually. One easy one that particularly spams the console is not liking package names with underscores. Here we are already aliasing in all the imports anyways so it seems OK to let the packages match the folder name.


mathetake (approved):
oh yeah was thinking the same :D

Issue #323, opened 2022-03-03
Add opt-in support for post 1.0 Sign-extension operators


nullpo-head:
Some use cases, such as Grain and AS need [Sign-extension operators](https://github.com/WebAssembly/sign-extension-ops), which is newly introduced feature in 1.1 specification.

```
invalid function (105/323): invalid instruction 0xc0
```

nullpo-head:
Sorry, this is duplicate of #66 . Let me close this issue.

PR #569, opened 2022-05-17
Adds assemblyscript host module and example

### LINKED ISSUE #552: Add AssemblyScript hooks and example
Currently, many users stub at least AssemblyScript's abort handler so that the code can run even if WASI isn't initialized. This is often partially stubbed because folks don't know how to handle the log data.

Ex. wapc-go
https://github.com/wapc/wapc-go/blob/master/engines/wazero/wazero.go#L153-L182

We should add an "assemblyscript" go module, and a basic instantiation like WASI:

```go
// Instantiates assemblyscript special imports under the module name "env", logging no-where
closerEnv, err := assemblyscript.Instantiate(ctx, r) 
```

There are choices to make, so I think we can make a builder, similar to ModuleBuilder, to sort that out.

```go
closerEnv, err := assemblyscript.NewBuilder(r).
WithAbortWriter(abort).
WithTraceWriter(trace).
Instantiate(ctx) // or Compile()
```

Note: the above could also do something like WithAbortToStdout() to plumb implicitly to defaults, which can make usage easier.

Finally, with this in, it would make sense to add examples/assemblyscript that shows basic usage of assemblyscript, but also tests abort and trace actually write (to stdout in the example test). This would be the right time in other words, to add an assemblyscript example because we have the hooks for it.

anuraaga:

A `assemblyscript` module is introduced which configures a wasm module, `env`, with AssemblyScript built-ins. While programs built with WASI will not use this module, other programs require it when instantiating the module. By default, `abort` is printed to stderr, `trace` is not printed, and `seed` returns a value based on `crypto/rand`.

An example is also added with two functions, one that performs simple math with an ignored call to `trace`, and another that throws an error, triggering `abort`. 

Fixes #552


assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,163 @@
...
+
+// WithSeedSource implements ModuleBuilder.WithSeedSource
+func (m *moduleBuilder) WithSeedSource(reader io.Reader) ModuleBuilder {
+	m.seedSource = reader
+	return m
+}
+
+// Instantiate implements ModuleBuilder.Instantiate
+func (m *moduleBuilder) Instantiate(ctx context.Context, runtime wazero.Runtime) (api.Closer, error) {
+	return runtime.NewModuleBuilder("env").
```
mathetake, line 85: `env` is not only used by AS but is kind of "default" host functions, so I'm not sure users want to use "env" module only for assembly script or not.. maybe not defining module builder, but instead just move all of the code in this PR under an example? @codefromthecrypt I think I need your inputs  here

anuraaga: Sorry, forgot to link the issue :) We do have some input from @codefromthecrypt in #552 though happy to change as needed.

examples/assemblyscript/assemblyscript_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/tetratelabs/wazero/internal/testing/maintester"
+	"github.com/tetratelabs/wazero/internal/testing/require"
+)
+
+// Test_main ensures the following will work:
+//
+// go run assemblyscript.go 7
+func Test_main(t *testing.T) {
+	if runtime.GOARCH == "riscv64" {
+		t.Skip("Uses bulk memory operations which are not implemented on riscv64 yet.")
```
mathetake, line 16: looking at the ci error, this is not due to bulk memory thingy (as the runtime passes all spec tests), but due the usage of AOT compiler on riscv...

assemblyscript/assemblyscript.go
mathetake: oh ic!

anuraaga: By the way, just as a reference, in AS the default for host declarations uses the file path as the module

https://www.assemblyscript.org/concepts.html#module-imports

So it may not be as idiomatic to use env as a catch-all host module for AS programs. Still does seem a bit concerning though

examples/assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,73 @@
...
+//
+// See README.md for a full description.
+func main() {
+
+	// Choose the context to use for function calls.
+	ctx := context.Background()
+
+	// Create a new WebAssembly Runtime. AssemblyScript enables bulk-memory operations by default
+	// so we enable it in the runtime too.
+	r := wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfigCompiler().
```
mathetake, line 30: I think this is the root cause of the failure :D 
```suggestion
	r := wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfig().
```

examples/assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,73 @@
...
+		WithFeatureBulkMemoryOperations(true))
```
mathetake, line 31: Some AS modules require sign-extensions proposals so maybe better enable Core2?
```suggestion
		WithWasmCore2())
```

mathetake: See #323

examples/assemblyscript/assemblyscript.go
anuraaga: Oops, thanks!

assemblyscript/assemblyscript.go
mathetake: thanks for the inputs!

examples/assemblyscript/assemblyscript.go
anuraaga: Good idea

examples/assemblyscript/testdata/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "hello-assemblyscript",
+  "version": "0.0.1",
+  "scripts": {
+    "build": "asc assemblyscript.ts --debug -b none -o assemblyscript.wasm -"
```
mathetake, line 5: wondering if we should verify the code is compilable in the examples workflow https://github.com/tetratelabs/wazero/blob/main/.github/workflows/examples.yaml

the same goes for rust examples @codefromthecrypt

assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,163 @@
...
+			if err != nil {
+				return
+			}
+			fn, err := readAssemblyScriptString(ctx, mod, fileName)
+			if err != nil {
+				return
+			}
+			_, _ = fmt.Fprintf(m.abortWriter, "%s at %s:%d:%d\n", msg, fn, lineNumber, columnNumber)
+		}).
+		ExportFunction("trace", func(ctx context.Context, mod api.Module, message uint32, nArgs uint32, arg0 float64, arg1 float64, arg2 float64, arg3 float64, arg4 float64) {
```
mathetake, line 97: lol the weird signature tbh :D

examples/assemblyscript/testdata/package.json
anuraaga: Added the AS build to the examples workflow

assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,163 @@
...
+				return float64(time.Now().UnixMilli())
+			}
+			bits := binary.LittleEndian.Uint64(b)
+			return math.Float64frombits(bits)
+		}).
+		Instantiate(ctx)
+}
+
+// readAssemblyScriptString reads a UTF-16 string created by AssemblyScript.
+func readAssemblyScriptString(ctx context.Context, m api.Module, pointer uint32) (string, error) {
```
mathetake, line 136: maybe adds unit tests for this?

assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// WithSeedSource implements ModuleBuilder.WithSeedSource
+func (m *moduleBuilder) WithSeedSource(reader io.Reader) ModuleBuilder {
+	m.seedSource = reader
+	return m
+}
+
+// Instantiate implements ModuleBuilder.Instantiate
+func (m *moduleBuilder) Instantiate(ctx context.Context, runtime wazero.Runtime) (api.Closer, error) {
```
mathetake, line 84: maybe I would separate out each function here and write unitests for each :D

assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,182 @@
...
+		})
+
+	// Optimize for our default which discards traces. No need to execute any logic in the host function.
+	if m.traceWriter != io.Discard {
+		mod.ExportFunction("trace", func(ctx context.Context, mod api.Module, message uint32, nArgs uint32, arg0 float64, arg1 float64, arg2 float64, arg3 float64, arg4 float64) {
+			trace(ctx, mod, message, nArgs, arg0, arg1, arg2, arg3, arg4, m.traceWriter)
+		})
+	} else {
+		mod.ExportFunction("trace", func(ctx context.Context, mod api.Module, message uint32, nArgs uint32, arg0 float64, arg1 float64, arg2 float64, arg3 float64, arg4 float64) {
+			// no-op
```
codefromthecrypt, line 100:
```suggestion
			// stub for no-op
```

codefromthecrypt (commented):
First pass is to rejig config.

This should default to what's set in ModuleConfig, and I think now's the right time to move experimental.Sys.RandSource to ModuleConfig then say it is used by default in WASI and AssemblyScript hooks. We should be able to read 8 bytes from the same source and that consolidates how things work.

For the trace writer vs abort writer, abort writer should default to ModuleConfig.StdErr (ack this will require using internal code to do that, but that's ok as it is in the same repo)

TraceWriter should as you had it default to io.Discard. When overriding either in AssemblyScript.ModuleBuilder, we should use same rationale notes as exist on ModuleConfig, or link saying Note: this has the same sandboxing implications as XXX

Thanks for the start on this!

assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,182 @@
...
+	//
+	// Defaults to os.Stderr.
+	WithAbortWriter(writer io.Writer) ModuleBuilder
+
+	// WithTraceWriter sets the io.Writer that should be used to write trace messages to.
+	// Trace messages are written when the built-in trace function is called within an
+	// AssemblyScript program.
+	//
+	// Defaults to io.Discard meaning trace messages are not written.
+	WithTraceWriter(writer io.Writer) ModuleBuilder
```
codefromthecrypt, line 39: Similarly, this shouldn't be assignable here as it is underneath the modules importing it. It should be io.Discard by default and allowed only to set to `WithTraceToStdout() or StdErr()` anything else should be an explicit user request as it complicates sandboxing as the real owner is the importing module.

assemblyscript/assemblyscript.go
```diff
@@ -0,0 +1,182 @@
...
+func Instantiate(ctx context.Context, r wazero.Runtime) (api.Closer, error) {
+	return NewModuleBuilder().Instantiate(ctx, r)
+}
+
+// ModuleBuilder allows configuring the module that will export functions used automatically by AssemblyScript.
+type ModuleBuilder interface {
+	// WithAbortWriter sets the io.Writer that should be used to write abort messages to.
+	// Abort messages are written when an error is thrown within an AssemblyScript program.
+	//
+	// Defaults to os.Stderr.
```
codefromthecrypt, line 31: so I probably caused this confusion earlier. I meant to default to ModuleConfig stderr, not os.stderr, as the former centralizes the sandbox config. So for example, if we set this manually here, this is set different than the module importing it. I think really the default should be AbortWriter -> importing ModuleConfig.Stderr (which is internally SysConfig). The only other option should be to turn it off entirely (WithAbortWriterDisabled())

anuraaga:
Thanks @codefromthecrypt - I have rewired this to use the sys context everywhere, including adding `RandSource` to `ModuleConfig` and using it both here and in wasi

examples/assemblyscript/testdata/assemblyscript.ts
```diff
@@ -0,0 +1,8 @@
+export function hello_world(arg: u32): u32 {
+  trace("hello world", 2, arg + 0.1, arg + 0.2);
+  return arg + 3;
+}
+
+export function goodbye_world(): void {
+  throw new Error("sad sad world");
```
codefromthecrypt, line 7: 😿

codefromthecrypt:
looks great. I'll push a polishing commit for some docs rather than have you parrot them into the code :D After that merge!

codefromthecrypt (approved):

examples/assemblyscript/testdata/package.json
codefromthecrypt: rust remains a TODO, but not in this PR!

codefromthecrypt:
I did a cleanup pass notably to ensure things error and not swallow, as well that the functions are extracted and documented. I ran out of time, but I think it is in good enough shape to merge now. Thanks for getting this together @anuraaga!

codefromthecrypt:
https://github.com/wapc/wapc-go/pull/29 uses this

Issue #563, opened 2022-05-16
Assign the file system when calling function


knqyf263:
I'd like to pass the different filesystem when calling a function as below.

```
	callback := mod.ExportedFunction("callback")
	filepath.WalkDir("root", func(path string, d fs.DirEntry, err error) error {
		if d.IsDir() && (path == "appA" || path == "appB") {
			callback.Call(ctx, wazero.WithFS(os.DirFS(path)))
		}
		return nil
	})
```

This is because we want wasm to be able to access only files under `appA` and `appB`. As far as I know, it is not supported now. It would be like the following in wazero.

```
	code, err := r.CompileModule(ctx, wasmCode, wazero.NewCompileConfig())
	if err != nil {...}

	filepath.WalkDir("root", func(path string, d fs.DirEntry, err error) error {
		if d.IsDir() && (path == "appA" || path == "appB") {
			rooted := os.DirFS(path)
			config := wazero.NewModuleConfig().WithFS(rooted)
			mod, err := r.InstantiateModule(ctx, code, config)
			if err != nil {...}
			defer mod.Close(ctx)

			callback := mod.ExportedFunction("callback")
			callback.Call(ctx)
		}
		return nil
	})
```

Is this the best way? I am afraid of the penalty of initializing the module every time since it is called frequently.

mathetake:
oh yeah, this sounds reasonable to me and should be possible. Assigning @codefromthecrypt as this is an API-related thing!

codefromthecrypt:
@knqyf263 would context-based configuration work for you? This feels like the solution here, or at least a workaround. Something like https://pkg.go.dev/github.com/tetratelabs/wazero/experimental#Sys

knqyf263:
Do you mean you will add a new function to the `Sys` interface?

codefromthecrypt:
I was thinking another key, actually experimental.FSKey *

What would happen is that any call in the stack would have this as its FS, unless overridden again. Ex. wasm would see this FS, so would any host function call, recursively, unless some host function overrode the key again.

```golang
fn.Call(context.WithValue(ctx, experimental.FSKey{}, os.DirFS(path)), ....)
```

* We will clean up the "experimental" namespace after implementing WASI snapshot 2 #263 and/or adding the new FS type #390. I think doing too much now would probably end up drifting later due to these anyway, so experimental is likely not the worst way to describe it and meanwhile we can this out to you and figure out how well it works.

codefromthecrypt:
[...]

PR #571, opened 2022-05-17
Allow passing fs.FS when calling functions

knqyf263:
## Issue
Close https://github.com/tetratelabs/wazero/issues/563

## ToDo
- [x] FSContext
- [x] FSConfig
- [x] FSKey
- [x] Override fs.FS in WASI
- [x] Add an example
- [x] Unit test
- [x] Example test

### DIFF
--- a/config.go
+++ b/config.go
@@ -10,6 +10,7 @@ import (
 	"github.com/tetratelabs/wazero/api"
 	"github.com/tetratelabs/wazero/internal/engine/compiler"
 	"github.com/tetratelabs/wazero/internal/engine/interpreter"
+	fs2 "github.com/tetratelabs/wazero/internal/fs"
 	"github.com/tetratelabs/wazero/internal/wasm"
 )
 
@@ -451,7 +452,7 @@ type moduleConfig struct {
 	// environKeys allow overwriting of existing values.
 	environKeys map[string]int
 
-	fs *wasm.FSConfig
+	fs *fs2.FSConfig
 }
 
 func NewModuleConfig() ModuleConfig {
@@ -459,7 +460,7 @@ func NewModuleConfig() ModuleConfig {
 		startFunctions: []string{"_start"},
 		environKeys:    map[string]int{},
 
-		fs: wasm.NewFSConfig(),
+		fs: fs2.NewFSConfig(),
 	}
 }
 
--- a/config.go
+++ b/config.go
@@ -3,7 +3,6 @@ package wazero
 import (
 	"context"
 	"errors"
-	"fmt"
 	"io"
 	"io/fs"
 	"math"
@@ -452,21 +451,15 @@ type moduleConfig struct {
 	// environKeys allow overwriting of existing values.
 	environKeys map[string]int
 
-	// preopenFD has the next FD number to use
-	preopenFD uint32
-	// preopens are keyed on file descriptor and only include the Path and FS fields.
-	preopens map[uint32]*wasm.FileEntry
-	// preopenPaths allow overwriting of existing paths.
-	preopenPaths map[string]uint32
+	fs *wasm.FSConfig
 }
 
 func NewModuleConfig() ModuleConfig {
 	return &moduleConfig{
 		startFunctions: []string{"_start"},
 		environKeys:    map[string]int{},
-		preopenFD:      uint32(3), // after stdin/stdout/stderr
-		preopens:       map[uint32]*wasm.FileEntry{},
-		preopenPaths:   map[string]uint32{},
+
+		fs: wasm.NewFSConfig(),
 	}
 }
 
@@ -493,7 +486,7 @@ func (c *moduleConfig) WithEnv(key, value string) ModuleConfig {
 // WithFS implements ModuleConfig.WithFS
 func (c *moduleConfig) WithFS(fs fs.FS) ModuleConfig {
 	ret := *c // copy
-	ret.setFS("/", fs)
+	ret.fs = ret.fs.WithFS(fs)
 	return &ret
 }
 
@@ -542,21 +535,8 @@ func (c *moduleConfig) WithRandSource(source io.Reader) ModuleConfig {
 // WithWorkDirFS implements ModuleConfig.WithWorkDirFS
 func (c *moduleConfig) WithWorkDirFS(fs fs.FS) ModuleConfig {
 	ret := *c // copy
-	ret.setFS(".", fs)
+	ret.fs = ret.fs.WithWorkDirFS(fs)
 	return &ret
-}
-
-// setFS maps a path to a file-system. This is only used for base paths: "/" and ".".
-func (c *moduleConfig) setFS(path string, fs fs.FS) {
-	// Check to see if this key already exists and update it.
-	entry := &wasm.FileEntry{Path: path, FS: fs}
-	if fd, ok := c.preopenPaths[path]; ok {
-		c.preopens[fd] = entry
-	} else {
-		c.preopens[c.preopenFD] = entry
-		c.preopenPaths[path] = c.preopenFD
-		c.preopenFD++
-	}
 }
 
 // toSysContext creates a baseline wasm.SysContext configured by ModuleConfig.
@@ -578,24 +558,9 @@ func (c *moduleConfig) toSysContext() (sys *wasm.SysContext, err error) {
 		environ = append(environ, key+"="+value)
 	}
 
-	// Ensure no-one set a nil FD. We do this here instead of at the call site to allow chaining as nil is unexpected.
-	rootFD := uint32(0) // zero is invalid
-	setWorkDirFS := false
-	preopens := c.preopens
-	for fd, entry := range preopens {
-		if entry.FS == nil {
-			err = fmt.Errorf("FS for %s is nil", entry.Path)
-			return
-		} else if entry.Path == "/" {
-			rootFD = fd
-		} else if entry.Path == "." {
-			setWorkDirFS = true
-		}
-	}
-
-	// Default the working directory to the root FS if it exists.
-	if rootFD != 0 && !setWorkDirFS {
-		preopens[c.preopenFD] = &wasm.FileEntry{Path: ".", FS: preopens[rootFD].FS}
+	preopens, err := c.fs.Preopens()
+	if err != nil {
+		return nil, err
 	}
 
 	return wasm.NewSysContext(math.MaxUint32, c.args, environ, c.stdin, c.stdout, c.stderr, c.randSource, preopens)
--- a/experimental/fs/fs.go
+++ b/experimental/fs/fs.go
@@ -1,0 +1,26 @@
+package fs
+
+import (
+	"context"
+	"io/fs"
+
+	"github.com/tetratelabs/wazero/api"
+	"github.com/tetratelabs/wazero/internal/wasm"
+)
+
+// FSKey is a context.Context Value key. It allows overriding fs.FS for WASI.
+//
+// See https://github.com/tetratelabs/wazero/issues/491
+type FSKey struct{}
+
+// WithFS overrides fs.FS in the context-based manner. Caller needs to take responsibility for closing the filesystem.
+func WithFS(ctx context.Context, fs fs.FS) (context.Context, api.Closer, error) {
+	fsConfig := wasm.NewFSConfig().WithFS(fs)
+	preopens, err := fsConfig.Preopens()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	fsCtx := wasm.NewFSContext(preopens)
+	return context.WithValue(ctx, FSKey{}, fsCtx), fsCtx, nil
+}
--- a/experimental/fs/fs_test.go
+++ b/experimental/fs/fs_test.go
@@ -1,0 +1,41 @@
+package fs_test
+
+import (
+	"context"
+	_ "embed"
+	"log"
+	"testing"
+	"testing/fstest"
+
+	"github.com/tetratelabs/wazero/experimental/fs"
+	"github.com/tetratelabs/wazero/internal/testing/require"
+	"github.com/tetratelabs/wazero/internal/wasm"
+)
+
+// This is a very basic integration of fs config. The main goal is to show how it is configured.
+func TestWithFS(t *testing.T) {
+	fileName := "animals.txt"
+	mapfs := fstest.MapFS{fileName: &fstest.MapFile{Data: []byte(`animals`)}}
+
+	// Set context to one that has experimental fs config
+	ctx, closer, err := fs.WithFS(context.Background(), mapfs)
+	if err != nil {
+		log.Panicln(err)
+	}
+	defer closer.Close(ctx)
+
+	v := ctx.Value(fs.FSKey{})
+	require.NotNil(t, v)
+	fsCtx, ok := v.(*wasm.FSContext)
+	require.True(t, ok)
+
+	entry, ok := fsCtx.OpenedFile(3)
+	require.True(t, ok)
+	require.Equal(t, "/", entry.Path)
+	require.Equal(t, mapfs, entry.FS)
+
+	entry, ok = fsCtx.OpenedFile(4)
+	require.True(t, ok)
+	require.Equal(t, ".", entry.Path)
+	require.Equal(t, mapfs, entry.FS)
+}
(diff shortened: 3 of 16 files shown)


codefromthecrypt:
taking a look now, thanks!

experimental/fs/fs.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Closer converts wasm.FSContext to api.Closer
+type Closer struct {
+	wasm.FSContext
+}
+
+// Close implements api.Closer
+func (c Closer) Close(context.Context) error {
+	return c.FSContext.Close()
+}
```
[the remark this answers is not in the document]
codefromthecrypt: I think you can make `FSContext.Close()` accept a Context and not use it. that's fine

experimental/fs/fs.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/tetratelabs/wazero/internal/wasm"
+)
+
+// FSKey is a context.Context Value key. It allows overriding fs.FS for WASI.
+//
+// See https://github.com/tetratelabs/wazero/issues/491
+type FSKey struct{}
+
+// Closer converts wasm.FSContext to api.Closer
+type Closer struct {
```
codefromthecrypt, line 17: probably after changing FSContext.Close you don't need this type anymore (if you did, it should be not exported)

experimental/fs/fs.go
```diff
@@ -0,0 +1,34 @@
...
+	wasm.FSContext
+}
+
+// Close implements api.Closer
+func (c Closer) Close(context.Context) error {
+	return c.FSContext.Close()
+}
+
+// WithFS overrides fs.FS in the context-based manner. Caller needs to take responsibility for closing the filesystem.
+func WithFS(ctx context.Context, fsConfig wazero.FSConfig) (context.Context, api.Closer, error) {
```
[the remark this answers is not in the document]
codefromthecrypt: I prefer WithFS simpler by default. then WithFSBuilder for adding WithWorkDirFS (citing ModuleConfig). I don't really like exposing wazero.FSConfig for this iotw, as it is incidental except this special case.

Ex.
```go
ctx, closer, err := WithFSBuilder(ctx, rootFS).
   WithWorkDirFS(workDir).
   Build()
```

internal/wasm/fs.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"io/fs"
+	"math"
+	"sync/atomic"
+)
+
+// FileEntry maps a path to an open file in a file system.
+//
+// Note: This does not introduce cycles because the types here are in the package "wasi" not "internalwasi".
+type FileEntry struct {
```
codefromthecrypt, line 12: if any tests exist for any of these symbols, move them to fs_test.go

wasi/wasi.go
```diff
@@ -1359,10 +1360,16 @@ func newSnapshotPreview1(ctx context.Context) *snapshotPreview1 {
...
 }
 
-func sysCtx(m api.Module) *wasm.SysContext {
+func sysCtx(ctx context.Context, m api.Module) *wasm.SysContext {
 	if internal, ok := m.(*wasm.CallContext); !ok {
 		panic(fmt.Errorf("unsupported wasm.Module implementation: %v", m))
 	} else {
+		// Override FSContext when it is passed via context
+		if fsCtx := ctx.Value(expfs.FSKey{}); fsCtx != nil {
+			sys := *internal.Sys // copy
```
codefromthecrypt, line 1369: hmm I don't think we want to do any copies here. maybe make an `fsCtx` function and use it where FS ops are used instead? you can leave this one alone. It is ok to call both sysCtx and also fsCtx in the same function body.. you could also make a special case that returns both, which can re-use the panic check and reduce the lookup overhead.

experimental/fs/fs.go
knqyf263: Thanks for confirming. I just didn't want to change the existing signature. Fixed https://github.com/tetratelabs/wazero/pull/571/commits/1aea8ea87c9cee34d33c392497348722732ce6be.

codefromthecrypt:
ping me when you are ready for the next round. I'll keep a look out

wasi/wasi.go
knqyf263: I separated `FSContext` from `SysContext` and define a function returning both as you suggested. It affects many places as expected, but hopefully I understand what you meant.
https://github.com/tetratelabs/wazero/pull/571/commits/5ec6e1ac97c74cf00860df3f6d74871442421f14

knqyf263:
@codefromthecrypt I think I finished reflecting your comments. There are a few concerns though.

1. I've added a test for WithFS to PathOpen only.
  - I'm not sure if it is enough or you may prefer another approach
2. FSConfig is still embedded into `ModuleConfig`
  - Do you think it should be extracted?

wasi/wasi.go
```diff
@@ -1359,11 +1360,19 @@ func newSnapshotPreview1(ctx context.Context) *snapshotPreview1 {
 	return &snapshotPreview1{sys: &defaultSys{}}
 }
 
-func sysCtx(m api.Module) *wasm.SysContext {
```
codefromthecrypt, line 1362: maybe keep sysCtx for the FS unrelated ops

internal/wasm/sys.go
```diff
@@ -4,44 +4,14 @@ import (
...
-	File fs.File
-}
-
 // SysContext holds module-scoped system resources currently only used by internalwasi.
 type SysContext struct {
 	args, environ         []string
 	argsSize, environSize uint32
 	stdin                 io.Reader
 	stdout, stderr        io.Writer
-
```
codefromthecrypt, line 28: I think maybe there was a misunderstanding. I think we meant to say we usually don't embed like this:

```
type SysContext struct {
*FSContext
}
```

and instead name the field, and only export it if actually used externally.
```
type SysContext struct {
FS *FSContext
}
```

Sorry I think this confusion was caused by me not being clear earlier. I do think that the SysContext should contain an FSContext field. Just if there's one in the go `context.Context` the field value should be ignored for that.

Make sense?

config.go
```diff
@@ -443,21 +442,15 @@ type moduleConfig struct {
 	// environKeys allow overwriting of existing values.
 	environKeys map[string]int
 
-	// preopenFD has the next FD number to use
-	preopenFD uint32
-	// preopens are keyed on file descriptor and only include the Path and FS fields.
-	preopens map[uint32]*wasm.FileEntry
-	// preopenPaths allow overwriting of existing paths.
-	preopenPaths map[string]uint32
+	*wasm.FSConfig
```
codefromthecrypt, line 445: Let's name fields instead of embedding. If this needs to be embedded, it would need a comment saying why.

```suggestion
	fs *wasm.FSConfig
```

config.go
```diff
@@ -443,21 +442,15 @@ type moduleConfig struct {
...
 
 func NewModuleConfig() ModuleConfig {
 	return &moduleConfig{
 		startFunctions: []string{"_start"},
 		environKeys:    map[string]int{},
-		preopenFD:      uint32(3), // after stdin/stdout/stderr
-		preopens:       map[uint32]*wasm.FileEntry{},
-		preopenPaths:   map[string]uint32{},
+
+		FSConfig: wasm.NewFSConfig(),
```
codefromthecrypt, line 453: This also shouldn't need to be exported I think..

```suggestion
		fs: wasm.NewFSConfig(),
```

codefromthecrypt (commented):
I wanted to reply quickly to you about some confusion I caused. I apologize for wasting some of your time. In general this looks good.

The main test needed would be to ensure that the context FS overrides vs adds to the one you expect. So one neat way to prove that is to have a test with the base FS having two files and the overlay having only one with a different content.

Meanwhile once done experimenting, maybe roll back the changes to the examples/wasi as we don't put experiemental things in that dir.

Finally, thanks again, I think after the adjustments here, this is getting to polish stage!

wasi/wasi.go
knqyf263: I kept it at the beginning, but most of the logic is the same as `fsCtx()` and it looks redundant. `sysFSCtx` returns both and we can just ignore `fsCtx`. Do you prefer having `sysCtx`, `fsCtx` and `sysFSCtx` though they are duplicated?

codefromthecrypt: no strong opinion, if anything sysCtx and the double one. We don't benchmark calls to env or args anyway, and if it were a problem we could refactor it later.

config.go
knqyf263: It is not exporting. Just pass the value to the field. Anonymous embedding requires the same name as the struct name for initialization.

codefromthecrypt: we don't tend to do anonymous embedding is the main point, what I meant is when that is removed, let's not assign an uppercased/exported field name as that would be distracting as we never do that.

internal/wasm/sys.go
knqyf263: oh...I see. It was a relatively big effort honestly😄 I should have asked you to make sure what you meant. I'll revert it then.

config.go
codefromthecrypt: fwiw ack this is bikeshed and also why we need to add rationale or some style notes.

we get some benefit of fast review and almost no accidental export accidents by always using unexported field names. we don't need to think too much if the field is the same name as anonymous or is anonymous when we have no anonymous. We also don't need to scan to see if the type is exported or not, and sometimes it is off diff anyway. Not saying all projects do this, but we are fairly consistent and do (should) comment when we do anything to the contrary. Once in the groove it is great because you can scan through a lot of changes fast without thinking so much all the time. OTOH projects don't always do this so it isn't a portable approach.

Anyway I thought it worth mentioning to you, as appreciate your help and new to the style here.

knqyf263: Yes, I don't have a plan to export it.

internal/wasm/sys.go
codefromthecrypt: well good news is I owe you one :D

codefromthecrypt:
@mathetake I think this is on the right track. I'm going to be mostly away this evening until tomorrow. Can you make sure this gets past the finish line. Cheers!

mathetake:
I think this is on the right track :D

experimental/fs/fs_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/tetratelabs/wazero/experimental/fs"
+	"github.com/tetratelabs/wazero/wasi"
+)
+
+// catWasm was compiled the TinyGo source testdata/cat.go
+//go:embed testdata/cat.wasm
+var catWasm []byte
+
+// This is a very basic integration of fs config. The main goal is to show how it is configured.
+func Example_fs() {
```
mathetake, line 20: I think this test case overlaps with the examples/wasi/** so maybe I would drop this and simply do unit-test WithFS function something like :
```
func Test_WithFS(t *testing.T) {
	testFs := ....
	ctx, api, err := fs.WithFS(context.Background(), testFs)
	// ...
	
	v := ctx.Value(fs.FSKey{})
	require.NotNil(t, v)
	fsCtx, ok := v.(*wasm.FSContext)
	require.True(t, ok)
	fsCtx.OpenedFile()
	// ...
}
```

the reason is that maybe we want to limit the places where tinygo is used in examples/ directory. @codefromthecrypt wdyt

codefromthecrypt: I'd like to offer my favor card by doing this part post merge ;)

mathetake: hah that's good so good as-is then!

mathetake (approved):
I'm good when you two feel good to go!

experimental/fs/fs_test.go
knqyf263: Actually, I had the same feeling. I just tried to be aligned with experimental/sys_test.go. Please let me fix it.

internal/wasm/sys.go
knqyf263: If we don't embed and export `FSContext`, we have to wrap all functions from `FSContext` in `SysContext`. That is exactly why I embedded `FSContext`. For instance,

```
func (c *SysContext) Close(ctx context.Context) error {
	return c.fs.Close(ctx)
}

func (c *SysContext) CloseFile(fd uint32) (bool, error) {
	return c.fs.CloseFile(fd)
}
```

It is applied to `FSConfig` as well. I was feeling like this wrapping is redundant. I hope I get @codefromthecrypt's point this time 😄 I will start defining the same functions in `SysContext`, but please stop me if I'm wrong, @mathetake @codefromthecrypt.

knqyf263: This is what it looks like. This commit fails tests, but I want to get your feedback in the early stage.
https://github.com/tetratelabs/wazero/pull/571/commits/21c6d0ced853eb8d2ab0adff1e3a69b545b20d4b

wasi/wasi.go
knqyf263: If we don't export `FSContext`, how can we extract `FSContext` from `CallContext` in `fsCtx`? We cannot access it like

```
func fsCtx(ctx context.Context, m api.Module) *wasm.FSContext {
	if internal, ok := m.(*wasm.CallContext); !ok {
		panic(fmt.Errorf("unsupported wasm.Module implementation: %v", m))
	} else {
		// Override FSContext when it is passed via context
		if fsValue := ctx.Value(expfs.FSKey{}); fsValue != nil {
			fsCtx, ok := fsValue.(*wasm.FSContext)
			if !ok {
				panic(fmt.Errorf("unsupported fs key: %v", fsValue))
			}
			return fsCtx
		}
		return internal.Sys.fs // Can't access `fs` as it is private.
	}
}
```

Do you want to define `type FSContext interface` and return `SysContext` as `FSContext`?

internal/wasm/sys.go
knqyf263: Or do you want to add 

```
func (c *SysContext) FS() FSContext {
	return c.fs
}
```

and just use it?

knqyf263: I also tried the above approach.
https://github.com/tetratelabs/wazero/pull/571/commits/8e361970044201633d8fbcb6653f11b65462b9c7

wasi/wasi.go
knqyf263: I've added `FS()`.
https://github.com/tetratelabs/wazero/pull/571/commits/8e361970044201633d8fbcb6653f11b65462b9c7

experimental/fs/fs_test.go
knqyf263: Done https://github.com/tetratelabs/wazero/pull/571/commits/849e48294cfc82898b92e33684f983f9cf71c935

config.go
knqyf263: Done https://github.com/tetratelabs/wazero/pull/571/commits/588f6f8e65ab479cb92ac40f1e25fc3dbfc20b1a

knqyf263:
>Meanwhile once done experimenting, maybe roll back the changes to the examples/wasi as we don't put experiemental things in that dir.

Reverted. https://github.com/tetratelabs/wazero/pull/571/commits/56714b8862959f661a9f408c8b61c774b2079af4

knqyf263:
>The main test needed would be to ensure that the context FS overrides vs adds to the one you expect. So one neat way to prove that is to have a test with the base FS having two files and the overlay having only one with a different content.

I wanted to do that, but I didn't find a good place to add the test. Also, it appeared to require somewhat larger changes. I probably missed something.

BTW, please feel free to add any commits. Sometimes it's faster to write code than to explain it in writing.

codefromthecrypt:
@knqyf263 thanks for all the work here. I'll be at the conference all day, but I can review and add commits to address any things, possibly today but more likely by tomorrow. In any case thanks for putting this together as you made it possible to land during a busy time 👍

codefromthecrypt:
ps I started fiddling with this, but might not push commits until tomorrow. Main thing I noticed was tension in the packages and made internal/fs instead. There are some tests in sys_test.go that should move into there also. I'll do it!

codefromthecrypt:
I landed the other PRs that would drift this, and ran out of time to finish this up (still at conference and should meet people :) ). I'll finish this soon

codefromthecrypt:
I force pushed a squash of the preceding work because the merges were getting hairy. I'll add a commit later with the changes I mentioned above

codefromthecrypt:
@mathetake I finished re-org, can you take a look and add a commit as needed? otherwise merge it.

mathetake:
@knqyf263 thank you so much for working to this! 🏅 👍

knqyf263:
@codefromthecrypt Thanks for your help!
I saw you added internal/fs package. It looks good!

PR #579, opened 2022-05-19
experimental: consolidates context patterns

codefromthecrypt:
This consolidates the pattern used for context overrides, notably
replacing clock overrides via experimental.WithTimeNowUnixNano
and making all context keys internal.

This also makes sure experimental example tests are handled the same
way, notably backfilling one for WithFS


codefromthecrypt:
cc @pkedy @sam-at-luther @knqyf263 on the experimental sig change. I think once we figure out what's going on with WASI 2, we'll end up with either a bunch of WithXXX functions or possibly a ContextBuilder to overlay certain parts of the system context per-function in one go. For now, we'll stick with a bunch of WithXXX and keep them named same as ModuleConfig.WithXXX if they are also there.

mathetake (approved):

PR #582, opened 2022-05-20
Use correct pattern for table tests everywhere

anuraaga:

One of these happened to be what I saw first when looking for a pattern for defining tests, but it doesn't seem to be the correct pattern. We need to fix these to prevent confusion when people add tests for the first time.


mathetake (commented):
I'm fine with the change, but does this make any difference? I'm not sure what "correct pattern" means here TBH...

mathetake (approved):

Issue #526, opened 2022-05-05
JIT: Exception 0xc0000005 0x8 0x0 0x0 


inkeliz:
I'm testing some code with Wazero, and I notice one odd crash, that I'm not sure if it's either my code (which is Zig) or Wazero. That is strange from previous errors, because it's one "Exception 0xc0000005 0x8 0x0 0x0". It's not an panic or something from `testing.Error(err)`.

**I'm using Windows/amd64.**

-----------

The log is:

```
 go test -tags wasi,km,zig -v -bench=. -benchmem -benchtime=5s -cpu 1
=== RUN   TestEncodeObjectAPI
Exception 0xc0000005 0x8 0x0 0x0                                                                                                                                              
PC=0x0                                                                                                                                                                        
                                                                                                                                                                              
github.com/tetratelabs/wazero/internal/wasm/jit.(*callEngine).execWasmFunction(0xc00074a240, {0x5dd008, 0xc00009e140}, 0xc000952e70, 0xc000952cf0)                            
        Z:/GOPATH/pkg/mod/github.com/tetratelabs/wazero@v0.0.0-20220430041858-abd1c79f3335/internal/wasm/jit/engine.go:699 +0x185 fp=0xc00095dbb8 sp=0xc00095dae0 pc=0x4d5305 
github.com/tetratelabs/wazero/internal/wasm/jit.(*moduleEngine).Call(0xc000954030?, {0x5dd008, 0xc00009e140}, 0xc000952e70, 0xc000494dd0, {0xc000954030, 0x1, 0x1})           
        Z:/GOPATH/pkg/mod/github.com/tetratelabs/wazero@v0.0.0-20220430041858-abd1c79f3335/internal/wasm/jit/engine.go:582 +0x48f fp=0xc00095dcb8 sp=0xc00095dbb8 pc=0x4d4b2f 
github.com/tetratelabs/wazero/internal/wasm.(*FunctionInstance).Call(0x53bca0?, {0x5dd008?, 0xc00009e140?}, {0xc000954030?, 0xc000022270?, 0x0?})                             
        Z:/GOPATH/pkg/mod/github.com/tetratelabs/wazero@v0.0.0-20220430041858-abd1c79f3335/internal/wasm/call_context.go:164 +0x63 fp=0xc00095dd08 sp=0xc00095dcb8 pc=0x4901c3
benchmark%2ekarmem%2eorg.(*Wasm).Run(...)
[...]

PR #583, opened 2022-05-20
compiler: fixes calling convention violation.

mathetake:
If the call_indirect's offset register is the same as the calling-convention
reserved register, the function call target address resulted in the wrong 
address and therefore segfault.

Fixes partially #526 

### DIFF
--- a/internal/engine/compiler/impl_amd64.go
+++ b/internal/engine/compiler/impl_amd64.go
@@ -4244,15 +4244,15 @@ func (c *amd64Compiler) allocateRegister(t registerType) (reg asm.Register, err 
 // Pass indexReg == asm.NilRegister to indicate that use addr argument as the source of target function's address.
 // Otherwise, the added code tries to read the function address from the register for indexReg argument.
 //
-// Note: this is the counter part for returnFunction, and see the comments there as well
+// Note: this is the counterpart for returnFunction, and see the comments there as well
 // to understand how the function calls are achieved.
-func (c *amd64Compiler) compileCallFunctionImpl(index wasm.Index, codeAddressRegister asm.Register, functype *wasm.FunctionType) error {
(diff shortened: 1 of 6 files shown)


mathetake:
will backfill the regression tests as a toehold!

internal/engine/compiler/impl_amd64.go
```diff
@@ -4420,14 +4421,19 @@ func (c *amd64Compiler) compileCallFunctionImpl(index wasm.Index, codeAddressReg
 	// So we increment the call frame stack pointer.
 	c.assembler.CompileNoneToMemory(amd64.INCQ, amd64ReservedRegisterForCallEngine, callEngineGlobalContextCallFrameStackPointerOffset)
 
+	if amd64CallingConventionModuleInstanceAddressRegister == targetFunctionAddressRegister {
```
mathetake, line 4424: This is the key, others are just styles

mathetake:
ok fixed arm64 as well as added unit tests!

internal/engine/compiler/impl_amd64.go
```diff
@@ -4420,14 +4421,22 @@ func (c *amd64Compiler) compileCallFunctionImpl(index wasm.Index, codeAddressReg
...
+		// This case we must move the value on targetFunctionAddressRegister to another register, otherwise
+		// the address (jump target below) will be modified and result in segfault.
+		// See $526.
```
anuraaga, line 4427: `#526` or consider actual url for this and others

internal/engine/compiler/impl_amd64.go
anuraaga: Just wondering, is the reason to only mov in this branch an optimization to save a register for other use?

anuraaga (approved):

internal/engine/compiler/impl_amd64.go
mathetake: yeah, optimization as not always end up in this situation, and the less instructions the better in general!

PR #584, opened 2022-05-23
compiler(arm64): allow large type IDs during compilation

mathetake:

### DIFF
--- a/internal/asm/arm64/impl.go
+++ b/internal/asm/arm64/impl.go
@@ -4,9 +4,8 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"math"
-
 	"github.com/tetratelabs/wazero/internal/asm"
+	"math"
 )
 
 type NodeImpl struct {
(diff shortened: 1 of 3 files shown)


internal/asm/arm64/impl.go
```diff
@@ -1583,10 +1583,10 @@ func (a *AssemblerImpl) encodeLoadOrStoreWithRegisterOffset(
 // In theory, offset can be any, but for simplicity of our homemade assembler, we limit the offset range
 // that can be encoded enough for supporting compiler.
 func validateMemoryOffset(offset int64) (err error) {
-	if offset > 255 && offset%8 != 0 {
-		// This is because we only have large offsets for load/store with Wasm value stack, and its offset
-		// is always multiplied by 8 (== the size of uint64, the type of value stack in Go).
-		err = fmt.Errorf("large memory offset (>255) must be a multiple of 8 but got %d", offset)
+	if offset > 255 && offset%4 != 0 {
+		// This is because we only have large offsets for load/store with Wasm value stack or reading type IDs, and its offset
+		// is always multiplied by 4.
```
anuraaga, line 1588: Can you add more background on where the value `4` comes from? The previous explanation with `uint64` for `8` was nice

internal/engine/compiler/compiler_controlflow_test.go
```diff
@@ -752,6 +752,65 @@ func TestCompiler_compileCallIndirect(t *testing.T) {
...
 }
 
+// TestCompiler_callIndirect_largeTypeIndex ensures that non-trivial large type index works well during call_indirect.
+func TestCompiler_callIndirect_largeTypeIndex(t *testing.T) {
+	env := newCompilerEnvironment()
+	table := make([]wasm.Reference, 1)
+	env.addTable(&wasm.TableInstance{References: table})
+	// Ensure that the module instance has the type information for targetOperation.TypeIndex,
+	// and the typeID  matches the table[targetOffset]'s type ID.
+	const typeIndex, typeID = 12345, 0
```
anuraaga, line 762: Is a large type ID anything more than 8 bits? 12345 seemed smaller than I was expecting :P

anuraaga (approved):

internal/engine/compiler/compiler_controlflow_test.go
mathetake: hehe yeah I think I need a comment why `12345` is considered large here :D

PR #585, opened 2022-05-23
wasi: returns fixed resolutin via ClockResGet

mathetake:
As pointed in https://github.com/tetratelabs/wazero/issues/526#issuecomment-1134034760, 
swift wasm binaries require clock_resg_get to return SUCCESS to run without unreachable.

For now, we temporarily fix the value `1000` to nanosec, and later we will do the overhaul on this
together with `ClockTimeGet` as these two implementations must be well aligned.


mathetake:
cc @inkeliz this will allow you to pass tests with Swift

anuraaga (approved):

PR #586, opened 2022-05-23
Follows go style for amd64 constants.

anuraaga:

This resolves `revive` linter errors in the amd64 package related to

- No ALL_CAPS constants (REG_AX -> RegAX)
- Doc on all exported members

All the code changes are automatic with GoLand's renaming functionality. The docs on the instruction names essentially map from go asm to amd64 asm (by reading https://www.felixcloutier.com/x86/index.html) which I think best explains how each instruction works.

@mathetake If this PR / pattern makes sense, I will send one for arm64 too.


mathetake (approved):
LGTM!

PR #587, opened 2022-05-23
Follows go style for arm64 constants.

anuraaga:

Similar to #586 

This also removes two instructions that were unused, `RNG` and `UXTW`, which didn't have corresponding instructions in the arm64 specification.


internal/asm/arm64/consts.go
```diff
@@ -34,305 +50,389 @@ const (
...
+	RegR26
+	// RegR27 is the R27 register
+	RegR27
+	// RegR28 is the R28 register
+	RegR28
+	// RegR29 is the R29 register
+	RegR29
+	// RegR30 is the R30 register
+	RegR30
+	// RegZERO is the RZR register (read-only, always returning zero)
```
anuraaga, line 115: Would it make sense to rename this to `RegRZR`?

internal/asm/arm64/consts.go
```diff
@@ -343,152 +443,296 @@ func RegisterName(r asm.Register) string {
...
+	// MOVBU loads a byte. https://developer.arm.com/documentation/dui0802/a/A64-Data-Transfer-Instructions/LDRB--register-
 	MOVBU
+	// MOVD loads or stores a register, in 64-bit mode. https://developer.arm.com/documentation/dui0802/a/A64-Data-Transfer-Instructions/LDR--register-
 	MOVD
+	// MOVH loads or stores a signed halfword. https://developer.arm.com/documentation/dui0802/a/A64-Data-Transfer-Instructions/LDRSH--register-
 	MOVH
+	// MOVHU loads a halfword. https://developer.arm.com/documentation/dui0802/a/A64-Data-Transfer-Instructions/LDRH--register-
 	MOVHU
+	// MOVW loads or stores a signed word, in 32-bit mode. https://developer.arm.com/documentation/dui0802/a/A64-Data-Transfer-Instructions/LDRSW--register-
 	MOVW
```
anuraaga, line 607: Go's docs seem to indicate this is 32-bit LDRSW

https://pkg.go.dev/cmd/internal/obj/arm64

> 32-bit variant str, stur, ldrsw => MOVW

But the arm docs don't seem to indicate a 32-bit mode for it 🤔 

https://developer.arm.com/documentation/dui0802/a/A64-Data-Transfer-Instructions/LDRSW--register-?lang=en

internal/asm/arm64/consts.go
```diff
@@ -343,152 +443,296 @@ func RegisterName(r asm.Register) string {
...
+	// VectorArrangement2S is an arrangement of 2 words (64-bit vector)
 	VectorArrangement2S
+	// VectorArrangement4S is an arrangement of 4 words (128-bit vector)
 	VectorArrangement4S
+	// VectorArrangement1D is an arrangement of 1 doubleword (64-bit vector)
 	VectorArrangement1D
+	// VectorArrangement2D is an arrangement of 2 doublewords (128-bit vector)
 	VectorArrangement2D
+
+	// Assign each vector size specifier to a vector arrangement ID.
```
anuraaga, line 724: I think this is what is going on here

internal/asm/arm64/consts.go
mathetake: yeah makesense ! go for it

internal/asm/arm64/consts.go
mathetake: as `LDRSW` always has loads 32-bit onto 64-bit register with sign-extension so there's not ambiguity there

mathetake (approved):
Thank you for your hard work on documentation... this would definitely save us (including me) future :D

PR #593, opened 2022-05-27
Small cleanups to docs

anuraaga:

- Adds doc to a few public APIs that didn't have them
- Moves docs about groups of items so that they don't get assigned to an item as its godoc
- Fixes a method name typo

### DIFF
--- a/config_supported.go
+++ b/config_supported.go
@@ -2,6 +2,7 @@
 
 package wazero
 
+// CompilerSupported returns whether the compiler is supported in this environment.
 const CompilerSupported = true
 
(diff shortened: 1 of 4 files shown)


mathetake (approved):

config_supported.go
```diff
@@ -2,6 +2,7 @@
 
 package wazero
 
+// CompilerSupported returns whether the compiler is supported in this environment.
 const CompilerSupported = true
```
codefromthecrypt, line 6: We accidentally exported this, so I moved it internal. Thanks for noticing as if you didn't we would have gone 1.0 with this exported without cause! https://github.com/tetratelabs/wazero/pull/597

PR #594, opened 2022-05-27
Adds a Make command to run tests with coverage

anuraaga:
I think it can be helpful to be able to check out coverage locally when writing tests so use this command for that. We could consider publishing to codecov in the future if` it seems useful.


codefromthecrypt:
probably sensible to start migrating in things from func-e vs slowly and differently introducing them here. We spent a good deal of time on the makefile there, including making sure there was re-use of expressions and that things work on windows.

Do you mind migrating in the coverage related stuff instead? The choices in the "coverage" target were made to allow use of code cov as a sort of coverage IDE without the spam. Please also add the "clean" target stuff, if you don't mind.

https://github.com/tetratelabs/func-e/blob/master/.github/workflows/commit.yaml#L88-L96
https://github.com/tetratelabs/func-e/blob/master/.gitignore#L2
https://github.com/tetratelabs/func-e/blob/master/Makefile#L85-L90

After this, it should be more clear where to start for adding more tools to the build, and yeah I think we spent several months so might as well start ahead!

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/596

Issue #426, opened 2022-03-31
Improve godocs


codefromthecrypt:
I think our package docs could be improved. Also code snippets in our godocs aren't tested. we should change to normal examples to ensure they don't drift Ex. https://pkg.go.dev/github.com/tetratelabs/wazero#StartWASICommand

If someone has a good example to compare to as what "good godoc" is, happy to have references!

See https://github.com/tetratelabs/wazero/issues/425

codefromthecrypt:
In most cases we can move code snippets embedded in godocs to Example tests like this https://github.com/golang/go/blob/master/src/os/exec/example_test.go

help wanted if someone has time, otherwise in a week or so I will do it.

codefromthecrypt:
fyi due to this ancient issue in go, there's limited ability to improve docs on interfaces, as they neither break down methods, nor embed examples in them. smell of a hackweek https://github.com/golang/go/issues/5860

codefromthecrypt:
a link to godocs is on the new website under API https://wazero.io/

codefromthecrypt:
closing as I think our godocs are pretty good right now

codefromthecrypt:
the formatting of godoc like below doesn't linefeed either. If someone complains, we should do something different (such as redo docs so that it isn't a problem), look for an issue there is in Go about formatting and cite it, or look to see a better formatted alternative to https://pkg.go.dev

```
// Note: All WASI functions return a single Errno result, ErrnoSuccess on success.
// Note: Closing the wazero.Runtime closes this instance of WASI as well.
```

Note: any godoc improvements should apply to all our public packages: wazero, wasi, assemblyscript, sys, api, and experimental. If we decide to do anything, we should do it to all the public packages, not just one of them.

codefromthecrypt:
I looked into it and while there are many ways to re-render godoc into
markdown, there are limited options on what to do in godoc itself. There
are tricks implicitly used where preformatting (tab indent) is used to
avoid having to add extra lines everywhere. This is done in gob, for
example. I raised #600 to fix all the files that are public

See https://go.dev/src/encoding/gob/doc.go
See https://github.com/fluhus/godoc-tricks/blob/master/doc.go

codefromthecrypt:
go 1.19 fixes some more common things we've worked around including lists and links! https://tip.golang.org/doc/comment

mathetake:
[...]

PR #595, opened 2022-05-27
Small cleanups to godoc

anuraaga:

Godoc compresses subsequent lines into one without adding empty lines. This causes lists to look like a blob of text.

- Adds empty lines between list items (e.g. https://pkg.go.dev/github.com/tetratelabs/wazero#CompiledModule). Unfortunately this causes each item to be a paragraph, not a line break, but there is no alternative from what I could find. This is better than blob of text though I think
- Documents the API for `NewRuntimeConfig`, not the behavior (which is arch specific). https://pkg.go.dev/github.com/tetratelabs/wazero#NewRuntimeConfig

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -10,7 +10,9 @@ import (
 // ExternType classifies imports and exports with their respective types.
 //
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#import-section%E2%91%A0
+//
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#export-section%E2%91%A0
+//
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#external-types%E2%91%A0
 type ExternType = byte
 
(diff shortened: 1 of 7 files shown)


mathetake (approved):

codefromthecrypt:
I prefer if fixing formatting to do it everywhere or nowhere, so making some comments especially as this makes the source look worse.

Here are some thoughts/questions
* Is there an alternative to https://pkg.go.dev for formatted docs which doesn't look bad?
* Are there any other ways to bullet used in other projects out there? seems odd there isn't even a google issue on this?
* How does this look (screenshot before after) with the extra spacing in pkg.go.dev
* Is this worse than pre-formatting (tab indented) to retain breaks?

Also, is there an alternative

codefromthecrypt:
If we end up in a situation of `Note:` `See:` having large line breaks, I might suggest we stop doing the `Note:` as it looks awkward in paragraph form, rather make a new paragraph. Just I feel like since this pattern is in hundreds of files, we can take some more effort to see how others handle this. #426 notes there are formatting limitations in godoc anyway, so I'm particularly curious about this.

codefromthecrypt:
giving a way out to retain time mutually..

knowing interfaces don't format anyway, and there are very few non-internal functions, let's either choose to not "fix" this and leave the source more readable, or "fix" this only for the things that would format and be used anyway (non-interface symbols in non-internal packages.) If "fixing" please take out the Note: prefix and just reword as it looks bad when separated with extra linebreak and asymmetric with the hundreds of places we do that without extra line break.

codefromthecrypt:
yep I think we should do things with research and diligence or not do it. good call to close this. I'll make a note of the fuzz in the other issue.

codefromthecrypt:
https://github.com/tetratelabs/wazero/issues/426#issuecomment-1139980708

PR #603, opened 2022-05-30
Tweaks doc for NewRuntimeConfig to describe API behavior

anuraaga:

Otherwise looks like this: https://pkg.go.dev/github.com/tetratelabs/wazero#NewRuntimeConfig

### DIFF
--- a/config_supported.go
+++ b/config_supported.go
@@ -2,7 +2,12 @@
 
 package wazero
 
-// NewRuntimeConfig returns NewRuntimeConfigCompiler
+// This file's implementation of the below function always returns compiler, but the doc for it should
+// describe the behavior of the API when the user calls it, not the implementation in this file which
+// may not be used based on build tags.
+
+// NewRuntimeConfig returns a RuntimeConfig using the compiler if it is supported in this environment,
+// or the interpreter otherwise.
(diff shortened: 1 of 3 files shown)


config_supported.go
```diff
@@ -2,7 +2,12 @@
 
 package wazero
 
-// NewRuntimeConfig returns NewRuntimeConfigCompiler
+// This file's implementation of the below function always returns compiler, but the doc for it should
```
codefromthecrypt, line 5: instead of copy/pasting stuff like this seems easier to solve this with an unexported function.

ex. move this to config.go
```
// NewRuntimeConfig returns a RuntimeConfig using the compiler if it is supported in this environment,
 // or the interpreter otherwise.
 func NewRuntimeConfig() RuntimeConfig {
  return newRuntimeConfig()
}
```

then you don't need to copy/paste these docs in the _supported and _unsupported files. instead, unexport NewRuntimeConfig to newRuntimeConfig

anuraaga: Yup fixed

mathetake (approved):

config_supported.go
codefromthecrypt: thanks tons!

Issue #599, opened 2022-05-28
Declaring exports functions on the runtime causes friction


prep:
**Is your feature request related to a problem? Please describe.**
`wazero` has us declare functions on the runtime via `NewModuleBuilder().ExportXXX()`. This is fine for functions that can operate in a vacuum who don't need to know about the instance from which the call is happening. For functions that do need that type of context, this approach is limiting.

**Describe the solution you'd like**
`wasmer` sets these functions up during instance creation and `wasmtime` does more or less the same by setting them up in the linker from which the instance is derived. 

I am not familiar with the inner workings of `wazero`. Reasoning purely from a API point of view, I'd love to be able to specify these functions when calling `InstantiateModule()`.

**Describe alternatives you've considered**
Of course, the workaround here is to create an runtime per instance. However, this takes away any benefit from the result of `CompileModule()` whose purpose is to speed up instance creation using `InstantiateModule()`.

codefromthecrypt:
Thanks for the description. If I understand correctly, this isn't so much about "Export" as wherever you do that, exported functions are how you accomplish an import binding.

It seems like you want the function binding to be a part of CompileModule. Ex to be able to supply a function import during wasm compilation. My guess is that doing so allows you to couple imports only used for one module. Is this right?

Keep in mind we won't likely achieve this the same way as the [deprecated linking proposal](https://github.com/WebAssembly/proposals/blob/main/inactive-proposals.md). Even if that still exists in wasmer, I think [wasmtime already deleted that](https://github.com/bytecodealliance/wasmtime-go/pull/125). However, we're interested in getting your problem solved.

As you may already notice, we have means in CompileConfig to rename imports so doing something like what you are saying seems reasonable (even if we may have tricks internally to make it work).

Main thing I want to know is if I understand you correctly, if you can mention concretely what you are doing (ex which type of imports you are satisfying.. wasi or custom ones), and what context you are using to decide exports, or if this is mainly just wanting to couple import function binding to compile module (for convenience)

codefromthecrypt:
Finally, I'm assuming this is best to do in [CompileConfig](https://pkg.go.dev/github.com/tetratelabs/wazero#CompileConfig) similar to how we plug in [ImportRenamer](https://pkg.go.dev/github.com/tetratelabs/wazero@v0.0.0-20220528010030-f1467e545350/api#ImportRenamer). 

[...]

PR #604, opened 2022-05-30
Adds Runtime.NewNamespace to allow intentional name collisions

codefromthecrypt:
`Runtime.NewNamespace` creates an empty namespace which won't conflict with any other namespace including the default. This is more efficient than multiple runtimes, as namespaces share a compiler cache.

In simplest case, a namespace won't conflict if another has a module with the same name:
```go
	b := assemblyscript.NewBuilder(r)
	m1, _ := b.InstantiateModule(ctx, r.NewNamespace(ctx))
	m2, _ := b.InstantiateModule(ctx, r.NewNamespace(ctx))
```

This is also useful for different modules that import the same module name (like "env"), but need different
configuration or state. For example, one with trace logging enabled and another disabled:
```go
	b := assemblyscript.NewBuilder(r)

	// m1 has trace logging disabled
	ns1 := r.NewNamespace(ctx)
	_ = b.InstantiateModule(ctx, ns1)
	m1, _ := ns1.InstantiateModule(ctx, compiled, config)

	// m2 has trace logging enabled
	ns2 := r.NewNamespace(ctx)
	_ = b.WithTraceToStdout().InstantiateModule(ctx, ns2)
	m2, _ := ns2.InstantiateModule(ctx, compiled, config)
```

Note: to facilitate this there's one breaking change below:

`ModuleBuilder.Instantiate(context.Context)` -> `Instantiate(context.Context, Namespace)`

This also fixes a resource leak when modules implicitly compiled code (ex `Runtime.InstantiateModuleFromCode`).

See #599


builder.go
```diff
@@ -302,11 +302,6 @@ func (b *moduleBuilder) Instantiate(ctx context.Context) (api.Module, error) {
 	if compiled, err := b.Compile(ctx, NewCompileConfig()); err != nil {
 		return nil, err
 	} else {
-		if err = b.r.store.Engine.CompileModule(ctx, compiled.(*compiledCode).module); err != nil {
```
codefromthecrypt, line 305: this was redundant

codefromthecrypt:
last step is adding an example. the example will show instantiate multiple stateful host modules under the same name using namespaces.

ex. each namespace includes "env" created from ModuleBuilder and a wasm source module that imports it. A function in "env" mutates its own map without function parameters needing to include a reference to it (externref or a uintptr). 

This should prove the isolation needed for #599 and also demonstrate how new sorts of module-scoped state not already in SysContext can be achieved without the use of go context (context.Context). cc @prep

codefromthecrypt:
Last thing I'll do is backfill tests to show that when there's implicit compilation of a module, the compiled code is deleted when the module closes. This was unrelated but easy to fix while here.

internal/wasm/engine.go
```diff
@@ -11,6 +11,14 @@ type Engine interface {
 	// CompileModule implements the same method as documented on wasm.Engine.
 	CompileModule(ctx context.Context, module *Module) error
 
+	// CompiledModuleCount is exported for testing, to track the size of the compilation cache.
```
mathetake, line 14: I see this is only used in tests, so maybe adds a comment about that?

mathetake (approved):

codefromthecrypt:
thanks for taking a break from simd to have a look @mathetake .. now back to simd 🧑‍🚒 ;)

PR #605, opened 2022-05-30
Implements clock ID support for getting time/res in wasi

anuraaga:

#### Supported clock IDs

We only support clock IDs REALTIME and MONOTONIC, while some runtimes support all IDs, not all. The next version of WASI removes the other IDs anyways

https://github.com/WebAssembly/WASI/pull/197
https://github.com/bytecodealliance/wasmtime/blob/8b48ce7fb79c7e36748e2e1a90be3bdf7d65cbce/crates/wasi-common/src/snapshots/preview_1.rs#L249

#### Mockable

Monotonic is not mockable, unlike realtime. I had first toyed with the idea of having the mock return `time.Time` instead of the unixnow value itself to support both but realized that there is no way to create a static monotonic `time.Time` in Go anyways. So allowing that would probably require instead to use the current function but provide it the `clockID` parameter. This didn't feel very useful so I left it out but happy to add it if it seems to make sense.

#### clock_time_get

`precision` is ignored, I looked at the other implementations and they seem to all ignore it. This is understandable since despite the WASI docs saying this is meant to be similar to POSIX `clock_gettime`, that function doesn't accept a precision. I suppose the main target of the parameter is mobile, not server-side. Let me know if it makes sense to commit to that understanding and remove the `TODO`

https://github.com/WasmEdge/WasmEdge/blob/bee4ad9831dec02d3224b9005d69cf30e3e973f6/lib/host/wasi/clock-linux.cpp#L26
https://github.com/wasmerio/wasmer/blob/3604debecc753c1a13701559035d75b30b77d912/lib/wasi/src/syscalls/unix/mod.rs#L36
https://github.com/bytecodealliance/cap-std/blob/024a7491b28de27d4a2dc6eea9e4372ac6001508/cap-time-ext/src/monotonic_clock.rs#L30

#### clock_get_res

`clock_get_res` should return a "real" resolution of a clock, which reflects the actual hardware and OS running the code. As such, it requires a syscall, but Go doesn't provide any function for doing this.

The other runtimes use libc or direct syscall

https://github.com/bytecodealliance/cap-std/blob/024a7491b28de27d4a2dc6eea9e4372ac6001508/cap-time-ext/src/monotonic_clock.rs#L2
https://github.com/WasmEdge/WasmEdge/blob/f77e67dfb298576b41eb7874df4f399a0de0e808/test/host/wasi/wasi.cpp#L17
https://github.com/wasmerio/wasmer/blob/3604debecc753c1a13701559035d75b30b77d912/lib/wasi/src/syscalls/unix/mod.rs#L3

For now, we just return a static value, 1 micro for realtime, and 1 nano for monotonic time, assuming realtime tends to have lower precision than monotonic. I couldn't find any actual usage in stdlib of tinygo or assemblyscript to better inform this decision.

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -351,6 +351,26 @@ See https://github.com/bytecodealliance/wasmtime/blob/2ca01ae9478f199337cf743a6a
 
 Their semantics match when `pathLen` == the length of `path`, so in practice this difference won't matter match.
 
+### ClockResGet
+
+A clock's resolution is hardware and OS dependent so requires a system call to retrieve an accurate value.
+Go does not provide a function for getting resolution, so without CGO we don't have an easy way to get an actual
+value. For now, we return fixed values of 1us for realtime and 1ns for monotonic, assuming that realtime clocks are
+often lower precision than monotonic clocks. In the future, this could be improved by having OS+arch specific assembly
+to make syscalls.
+
+For example, Go implements time.Now for linux-amd64 with this [assembly](https://github.com/golang/go/blob/f19e4001808863d2ebfe9d1975476513d030c381/src/runtime/time_linux_amd64.s).
+Because retrieving resolution is not generally called often, unlike getting time, it could be appropriate to only
+implement the fallback logic that does not use VDSO (executing syscalls in user mode). The syscall for clock_getres
+is 229 and should be usable. https://pkg.go.dev/syscall#pkg-constants.
+
+If implementing similar for Windows, [mingw](https://github.com/mirror/mingw-w64/blob/6a0e9165008f731bccadfc41a59719cf7c8efc02/mingw-w64-libraries/winpthreads/src/clock.c#L77
+) is often a good source to find the Windows API calls that correspond
+to a POSIX method.
+
+Writing assembly would allow making syscalls without CGO, but comes with the cost that it will require implementations
+across many combinations of OS and architecture.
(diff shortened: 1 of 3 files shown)


mathetake (approved):
LGTM,  deferring to @codefromthecrypt for the second pass

codefromthecrypt (approved):
As discussed offline, I think you can hueristically tell if the clock supports monotonic time via looking at its String val

```
// For debugging, the result of t.String does include the monotonic
// clock reading if present. If t != u because of different monotonic clock readings,
// that difference will be visible when printing t.String() and u.String().
```

so, ex. `m=` is the key
```
2022-05-30 16:54:47.337143 +0800 CST m=+0.000824695
```

When that's not the case, return ErrNosys

We can leave out other things around mock ability for a follow-up I can manage. In wazero, it isn't so much about mock ability of `Time` rather controlling the results of a function. What I'll do is similar to our context override for the wall clock, add one for the nano clock.

Note, WASI (both of them) aren't the only users of clock. The other ABI (ex go's) also use these two clocks, so I'll add plumbing in SysContext for that in the same change.

Meanwhile, feel free to merge this one when you have things organized as you see fit. It would be helpful to do that today as I have at least one PR that will drift yours and want to land today.

codefromthecrypt (approved):
looks good thanks!

PR #606, opened 2022-05-31
Adds clock_res_get to the finished list and removes `fs.FS`

codefromthecrypt:


codefromthecrypt:
oops I think I got this wrong. it looks like a copy of Emscripten's WASI polyfill...

codefromthecrypt:
I'll roll this back to just the formatting change and saying we're done. I don't know of a non-browser thing and we shouldn't hunt arbitrarily.

site/content/specs.md
```diff
@@ -76,53 +76,53 @@ your use case (ex which language you are using to compile, a.k.a. target Wasm).
...
-| fd_fdstat_set_flags     |   ❌   |                |
-| fd_fdstat_set_rights    |   ❌   |                |
-| fd_filestat_get         |   ❌   |                |
-| fd_filestat_set_size    |   ❌   |                |
-| fd_filestat_set_times   |   ❌   |                |
-| fd_pread                |   ❌   |                |
-| fd_prestat_get          |   ✅   | TinyGo,`fs.FS` |
-| fd_prestat_dir_name     |   ✅   | TinyGo         |
-| fd_pwrite               |   ❌   |                |
-| fd_read                 |   ✅   | TinyGo,`fs.FS` |
```
codefromthecrypt, line 101: removed the `fs.FS` thing I added at one point as we don't actually have an `fs.FS` abstraction. It is better to focus on language compilers.

mathetake (approved):

Issue #607, opened 2022-05-31
Add compiler support for FreeBSD


knqyf263:
I tried to compile wazero for FreeBSD, but got an error.

```
../../../../pkg/mod/github.com/tetratelabs/wazero@v0.0.0-20220529032218-029a79476b4f/internal/engine/compiler/mmap.go:74:16: undefined: syscall.Mprotect
```

Looks like `Mprotect` is not implemented in `syscall`.
https://github.com/golang/go/blob/aac1d3a1b12a290805ca35ff268738fb334b1ca4/src/syscall/zsyscall_freebsd_amd64.go

It is implemented in `golang.org/x/sys/unix`. I confirmed I could compile it after replacing `syscall` with `golang.org/x/sys/unix`.
https://github.com/golang/sys/blob/bc2c85ada10aa9b6aa9607e9ac9ad0761b95cf1d/unix/zsyscall_freebsd_amd64.go#L296-L308

If wazero needs to keep zero dependency (otherwise, need to rename it to `waone`?), can't we even use `golang.org/x`? Do you want to re-implement syscalls for other platforms in wazero?

codefromthecrypt:
hi @knqyf263 and thanks for reporting. yes it is indeed the goal to have zero deps. I think what you are hitting is lack of precision in our engine detection. It doesn't switch on a variant of linux so you are getting a later error like this (for the compiler). I suspect the interpreter would work, for example.

We have already added code for mmap for windows, based on demand, and ensure it with a test dimension. That allowed it to use the compiler instead of the interpreter, which doesn't have this issue.

We could add a runtimeconfig for BYOMMap easily enough so that folks can opt-into that for mmap. However, it may not be the only issue in the compiler. Our assembler has some platform specific nuance in it also. Let's first see what's going on in FreeBSD and get a test dimension up?

codefromthecrypt:
I'll clarify something we'll have to put into docs somehow.

The compiler has components that aren't OS-specific (the assembler) and also parts that are: the engine impl has a little and also the mmap function.

Since FreeBSD had an issue with one of these parts, we should have a test dimension to ensure it works. This doesn't mean we can all operating systems, but in this case it sounds like we should as it hit an issue and you need it and it probably won't be hard to support!

knqyf263:
> It doesn't switch on a variant of linux so you are getting a later error like this (for the compiler). I suspect the interpreter would work, for example.

[...]

PR #608, opened 2022-06-01
Extracts platform-specific runtime code into its own package

codefromthecrypt:
This moves the platform-specific runtime code (currently only used by
the compiler) into its own package. Specifically, this moves the mmap
logic, and in doing so makes it easier to test, for example new
operating systems.

This also backfills missing RATIONALE about x/sys and hints at a future
possibility to allow a plugin. However, the next step is to get FreeBSD
working natively on the compiler without any additional dependencies.

See #607

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -1,5 +1,50 @@
 # Notable rationale of wazero
 
+## Zero dependencies
+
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
cc @knqyf263 and sorry we didn't document this before

mathetake (approved):

PR #609, opened 2022-06-01
Backfill func validation unit tests for SIMD load, store, and lane manipulations

mathetake:


codefromthecrypt (approved):
great improvement thanks!

PR #612, opened 2022-06-01
Narrows compiler support to operating systems currently known to work

codefromthecrypt:
This narrows to what the `internal/platform` package supports, which is
currently bound by Go SDK source that include `Mprotect` or windows.

Ex. `zsyscall_linux_amd64.go` includes `Mprotect`, but
`zsyscall_freebsd_amd64.go` does not.

This should prevent errors like below, by allowing `wazero.NewRuntime()`
to properly fallback to the interpreter.

```
.../mmap.go:74:16: undefined: syscall.Mprotect
```

A later change will implement FreeBSD. This is just here to ensure users
don't crash on unexpected OS.

See #607

### DIFF
--- a/config_supported.go
+++ b/config_supported.go
@@ -1,4 +1,14 @@
-//go:build amd64 || arm64
+// Note: The build constraints here are about the compiler, which is more
+// narrow than the architectures supported by the assembler.
+//
+// For example, the internal/platform package has dependencies such as
+// syscall.Mprotect when not on Windows. Go's SDK only supports this on darwin
+// and linux. Constraints here may loosen as the platform package supports
+// other runtime.GOOS, or a plugin becomes available to accomplish the same.
+//
+// Meanwhile, users who know their runtime.GOOS can operate with the compiler
+// may choose to use NewRuntimeConfigCompiler explicitly.
+//go:build (amd64 || arm64) && (darwin || linux || windows)
 
 package wazero
 
--- a/config_supported.go
+++ b/config_supported.go
@@ -1,10 +1,7 @@
 // Note: The build constraints here are about the compiler, which is more
 // narrow than the architectures supported by the assembler.
 //
-// For example, the internal/platform package has dependencies such as
-// syscall.Mprotect when not on Windows. Go's SDK only supports this on darwin
-// and linux. Constraints here may loosen as the platform package supports
-// other runtime.GOOS, or a plugin becomes available to accomplish the same.
+// Constraints here must match platform.IsSupported.
 //
 // Meanwhile, users who know their runtime.GOOS can operate with the compiler
 // may choose to use NewRuntimeConfigCompiler explicitly.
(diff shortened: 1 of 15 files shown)


mathetake:
I believe this is not enough as Go's compiler resolves all symbols in the package regardless of the actual compiler pass
```
$  GOOS=freebsd go build ./...
# github.com/tetratelabs/wazero/internal/platform
internal/platform/mmap.go:55:16: undefined: syscall.Mprotect
```

mathetake:
need platform/unsupported.go to define symbols

mathetake:
(maybe I'm wrong though :D)

codefromthecrypt:
cool I can fix platform also!

config_supported.go
```diff
@@ -1,4 +1,14 @@
...
+// narrow than the architectures supported by the assembler.
+//
+// For example, the internal/platform package has dependencies such as
+// syscall.Mprotect when not on Windows. Go's SDK only supports this on darwin
+// and linux. Constraints here may loosen as the platform package supports
+// other runtime.GOOS, or a plugin becomes available to accomplish the same.
+//
+// Meanwhile, users who know their runtime.GOOS can operate with the compiler
+// may choose to use NewRuntimeConfigCompiler explicitly.
+//go:build (amd64 || arm64) && (darwin || linux || windows)
```
anuraaga, line 11: Should this line also explicitly reference linux and Darwin instead of not window?

https://github.com/tetratelabs/wazero/blob/main/internal/platform/mmap.go#L1

I think the syscall problem is a compile failure, not a runtime one so given how eager the compiler can be not sure if it wouldn't still try to compile the code referencing Mprotect even on freebsd, even though it's not used at runtime.

anuraaga: Sorry think I just repeated @mathetake

codefromthecrypt: same general problem area as takeshi's note. yep.

internal/platform/platform.go
```diff
@@ -9,6 +9,26 @@ import (
 	"runtime"
 )
 
+// IsSupported is exported for tests and includes constraints here and also the assembler.
+func IsSupported() bool {
+	if _, ok := map[string]struct{}{
```
mathetake, line 14: I would prefer switch rather than map as this might end up having allocating map each time or not
```go
switch runtime.GOOS {
case "darwin", "windows", "linux":
default: return false
}
```

mathetake: this way everything can be compile time const so friendlier for Go compiler 😉

mathetake (approved):

internal/platform/platform.go
codefromthecrypt: sure it was test but you are right this is cleaner anyway!

codefromthecrypt:
thanks for the eagle-eyes @mathetake and @anuraaga I forgot we can override runtime.GOOS to test compilation, though sadly GitHub Actions for freebsd won't be so easy 😅

PR #621, opened 2022-06-04
Adds gojs for Go generated Wasm

### LINKED ISSUE #432: Support GOOS=js GOARCH=wasm generated code
I was looking at @wuhuizuo's test which uses this source and compiles it like `GOOS=js GOARCH=wasm go build -o wasm.wasm`

```go
package main

import (
	"syscall/js"

	"github.com/wuhuizuo/go-wasm-go/provider/native"
)

func main() {
	js.Global().Set("Fibonacci", js.FuncOf(Wrap(native.Fibonacci)))
	js.Global().Set("RequestHTTP", js.FuncOf(Wrap(native.RequestHTTP)))
	js.Global().Set("FileIO", js.FuncOf(Wrap(native.FileIO)))
	js.Global().Set("MultiThreads", js.FuncOf(Wrap(native.MultiThreads)))
	js.Global().Set("BytesTest", js.FuncOf(Wrap(native.BytesTest)))
	js.Global().Set("InterfaceTest", js.FuncOf(Wrap(native.InterfaceTest)))
	js.Global().Set("ErrTest", js.FuncOf(Wrap(native.ErrTest)))

	select {}
}
```
https://github.com/wuhuizuo/go-wasm-go/blob/main/provider/wasm-go/main.go

The resulting wasm isn't exporting those functions. "run" "resume" "getsp" and "mem" are externalized.

You can identify, for example "Fibonacci" if you use the custom name section and get "github.com_wuhuizuo_go_wasm_go_provider_native.Fibonacci". However, again that function is neither the short name, nor is it exported. There's probably some approach to getting the function prefix somewhere

If we want to support this tech we'd need to:
1. add an API for non-exported functions
2. somehow detect the go-generated module path prefix to the function or add a utility to scan for it
 * possibly, we could add a second index when this is detected for the short name, and index it twice when doing so doesn't clobber something else. In go-generated it isn't currently possible.

I'm not sure how many use wasm produced by go vs tinygo, but raising it as I noticed there's no way to lookup these functions and use them, yet. We probably need some 👍 and people to say why/how they are using this, so that we don't displace time on things only used for demo projects.

codefromthecrypt:
This adds an experimental package `gojs` which implements the host side of Wasm compiled by `GOARCH=wasm GOOS=js go build -o X.wasm X.go`

This includes heavy disclaimers, in part inherited by Go's comments https://github.com/golang/go/blob/go1.19/src/syscall/js/js.go#L10-L11
Due to this many will still use TinyGo instead.

That said, this is frequently asked for and has interesting features including reflection and HTTP client support.

Ex.
```go
// Grant the compiled wasm access to the default HTTP Transport.
ctx = gojs.WithRoundTripper(ctx, http.DefaultTransport)

// Execute the "run" function with the binary you compiled via
// `GOARCH=wasm GOOS=js go build -o X.wasm X.go`
err = gojs.Run(ctx, r, compiled, config)
	log.Panicln(err)
} else if !ok {
	log.Panicln(err)
}
```

Fixes #432


wasm_exec/wasm_exec.go
```diff
@@ -0,0 +1,364 @@
...
+// See https://github.com/golang/go/blob/4170084ad12c2e14dc0485d2a17a838e97fee8c7/src/runtime/sys_wasm.s#L196
+func (j *jsWasm) clearTimeoutEvent(id uint32) {
+	if t := j.removeEvent(id); t != nil {
+		if !t.Stop() {
+			<-t.C
+		}
+	}
+}
+
+// _getRandomData converts the GOARCH=wasm stack to be compatible with
```
codefromthecrypt, line 247: @prep fyi here are my notes about go's use of getRandomData, though none of this is tested (hence draft)

codefromthecrypt:
note: I'm off for a week, so that's why this will be unusually stale for me

wasm_exec/REFERENCE.md
```diff
@@ -0,0 +1,136 @@
...
+type: `(func (param $sp i32))`. `$sp` is the base memory offset to read and
+write parameters to the stack (at 8 byte strides even if the value is 32-bit).
+
+So, implementors need to read the actual parameters from memory. Similarly, if
+there are results, the implementation must write those to memory.
+
+For example, `func walltime() (sec int64, nsec int32)` writes its results to
+memory at offsets `sp+8` and `sp+16` respectively.
+
+## Go-defined exported functions
```
codefromthecrypt, line 89: @prep haven't forgotten, just settling back in and will progress more over this week!

mathetake:
oh sorry mistakenly pushed a commit.. feel free to remove it

codefromthecrypt:
I'm going to finish this because it is too old

internal/wasm/host.go
```diff
@@ -156,6 +172,21 @@ func addFuncs(
...
+			proxiedIdx := len(funcNames)
+			if proxiedIdx > 255 {
+				return errors.New("TODO: proxied funcidx larger than one byte")
+			}
+			nameToFunc[pf.Proxied.Name] = pf.Proxied
+			funcNames = append(funcNames, pf.Proxied.Name)
+
+			proxyBody := make([]byte, len(pf.Proxy.Code.Body))
+			copy(proxyBody, pf.Proxy.Code.Body)
+			proxyBody[pf.CallBodyPos] = byte(proxiedIdx)
```
mathetake, line 185: shoulnd't this be encoded as unsigned 32-bit leb128 ?

wasm_exec/internal/spfunc/spfunc.go
```diff
@@ -0,0 +1,186 @@
...
+			panic(errors.New("TODO: param " + api.ValueTypeName(t)))
+		}
+	}
+
+	// Now that all parameters are on the stack, call the function
+	callFuncPos := len(code) + 1
+	if debugMode {
+		fmt.Printf("\n%s 0\n", wasm.OpcodeCallName)
+	}
+	code = append(code, wasm.OpcodeCall, 0)
```
mathetake, line 94: I think I want comment on this - zero is a place holder and will be replaced later by the actual index

internal/wasm/host.go
```diff
@@ -156,6 +172,21 @@ func addFuncs(
 		if hf, ok := v.(*HostFunc); ok {
 			nameToFunc[hf.Name] = hf
 			funcNames = append(funcNames, hf.Name)
+		} else if pf, ok := v.(*ProxyFunc); ok {
+			proxiedIdx := len(funcNames)
+			if proxiedIdx > 255 {
```
mathetake, line 177: if the MSB is set, the leb128 unsigned doesn't match the literal byte
```suggestion
			if proxiedIdx > 127 {
```

internal/wasm/host.go
codefromthecrypt: I was going to limit it to what fits in one byte. I'll panic if over noting the msg thing you mentioned. There aren't enough host functions to go over a byte even doubled.

codefromthecrypt:
next step is to backfill tests for errors on all the functions that are easy enough to err (like files not found, http errors etc)

wasm_exec/internal/compiler_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/tetratelabs/wazero/internal/wasm/binary"
+	"github.com/tetratelabs/wazero/wasm_exec"
+	. "github.com/tetratelabs/wazero/wasm_exec/internal"
+)
+
+// testCtx is an arbitrary, non-default context. Non-nil also prevents linter errors.
+var testCtx = context.WithValue(context.Background(), struct{}{}, "arbitrary")
+
+// Test_compileJsWasm ensures the infrastructure to generate wasm on-demand works.
+func Test_compileJsWasm(t *testing.T) {
```
mathetake, line 28: can this work on scratch containers?

codefromthecrypt: no that's why there's some logic later to skip the test. scratch doesn't have the "go" command.

codefromthecrypt: as this isn't a compiler per-se we shouldn't need to run also on scratch anyway. the main thing is to use multiple versions of go. This way, we can check the output wasm in multiple versions, which ideally aren't terribly different per os, or different at all!

codefromthecrypt: ps in the example test, I simply bail out when go compile fails. the instructions tell users to do it.

codefromthecrypt:
hmm the tests on CI take 40s which I assume is due to the amount of time it takes to compile. I might make a different approach where the tests use a single binary switched on arg. should reduce time and also flake

codefromthecrypt:
ok actually the execution speed of wasm is also a reason why it was slower. Regardless this is now compiling once for the package and re-using the same binary for parallel tests. Finishes in 1/4 the time on my laptop now.

examples/gojs/stars/main.go
```diff
@@ -0,0 +1,36 @@
+package main
```
codefromthecrypt, line 1: @mathetake fyi this might be a decent one to use for some perf curiosity tests as it makes 7.7MB of wasm that doesn't execute very fast due to so many host functions (and allocations within them)

codefromthecrypt:
Actually, the runtime performance is so bad I'm kindof wondering if we should just dump this.

Ex the below is literally returning a mock response

```bash
$ time go run stars.go 
wazero has 9999999 stars. Does that include you?

real    0m4.523s
user    0m8.657s
sys     0m0.563s
```

Also a fake main takes a second to execute:
```go
func main() {
	fmt.Println("hello")
}
```

@prep @paralin do you have a working example on https://github.com/prep/wasmexec that can compare execution time across its 3 runtimes? I want to double-check there's not something wrong with the impl here.

iotw I was expecting slow, but not second+ slow, and I'm not sure what use cases can be met by that slow as even docker is faster.

codefromthecrypt:
@paralin was thinking the same thing. I'll add a real benchmark which shows compile vs (instantiate/run). Instantiate and run are coupled as `GOOS=js` uses host state (host side values) coupled with wasm state (data segments etc).

codefromthecrypt:
ok @paralin's thoughts were right. Since @mathetake recently implemented compilation cache. Using `GOARCH=wasm GOOS=js` without it is incredibly slow due to the size and the complexity of the Wasm generated.

```bash
$ cd examples/gojs
$ go run .
2022/08/26 09:09:12 CompileModule took 4576ms with 36062KB cache
wazero has 9999999 stars. Does that include you?
2022/08/26 09:09:12 gojs.Run took 20ms

$ go run .
2022/08/26 09:09:15 CompileModule took 145ms with 36074KB cache
wazero has 9999999 stars. Does that include you?
2022/08/26 09:09:15 gojs.Run took 28ms
```

While still slow at runtime vs TinyGo, it isn't slow enough to endanger our CI anymore.

I think people building apps than embed wazero that runs `GOARCH=wasm GOOS=js` in docker...
They should use a cache dir, possibly doing a one-time run as a part of building the image, as otherwise people using the container will suffer and extraordinary first hit penalty every time.

codefromthecrypt:
@mathetake PTAL I think we are good. I'm not going to make a full on bench for gojs right now as already way over time budget and there is too much to review as it is. However, I think things are in good enough shape to review for merge.

mathetake (approved):

codefromthecrypt:
One thing I noticed, but won't dig into at the moment, is this has no ability to collaborate with ad-hoc JavaScript at the moment (ex define ad-hoc functions in JS like [spicedb](https://github.com/authzed/spicedb/blob/main/pkg/development/wasm/main.go#L59) cc @josephschorr)

This was out of scope for reasons of doing something like that needing a JavaScript engine which isn't a zero dependency thing in go, and makes the PR bigger also.

Currently, the only js functions supported are those Go defines in their own source code, so basically no dynamic javascript parsing. This makes sense because as wazero is typically used for non-browser use, starting with only what go uses allows folks to at least use normal go, even if they can't inject dynamic functions into it.

codefromthecrypt:
The other thing left out is that Go allows via .s files the ability to import custom functions. This was left out as the calling convention Go uses is completely unlike the rest of WebAssembly and making something to implement that especially before people ask for it would be odd. https://wazero.io/languages/go/#user-defined-host-functions

paralin:
Looks cool, will test!

PR #622, opened 2022-06-05
Replaces Memory.IndexByte with stronger documentation on Read

codefromthecrypt:
Memory.IndexByte is unneeded with better understanding of write-through
on Memory.Read. Removing this also helps usher folks into the myriad of
Go utilities that are compatable with []byte.

This also reduces the complexity of WASI which didn't need to re-buffer
random reads (also due to above).

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -254,18 +254,18 @@ type Memory interface {
 	// See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#-hrefsyntax-instr-memorymathsfmemorysize%E2%91%A0
 	Size(context.Context) uint32
 
-	// Grow increases memory by the delta in pages (65536 bytes per page). The return val is the previous memory size in
-	// pages, or false if the delta was ignored as it exceeds max memory.
+	// Grow increases memory by the delta in pages (65536 bytes per page).
+	// The return val is the previous memory size in pages, or false if the
+	// delta was ignored as it exceeds max memory.
 	//
-	// Note: This is the same as the "memory.grow" instruction defined in the WebAssembly Core Specification, except
-	// returns false instead of -1 on failure
+	// Notes
 	//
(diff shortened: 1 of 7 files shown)


wasi_snapshot_preview1/wasi.go
```diff
@@ -1382,34 +1382,34 @@ func (a *wasi) SchedYield(m api.Module) Errno {
 //
 // Note: importRandomGet shows this signature in the WebAssembly 1.0 (20191205) Text Format.
 // See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-random_getbuf-pointeru8-bufLen-size---errno
-func (a *wasi) RandomGet(ctx context.Context, m api.Module, buf uint32, bufLen uint32) (errno Errno) {
-	randomBytes := make([]byte, bufLen)
```
codefromthecrypt, line 1386: cc @clarkmcc here's the interesting part, which is you don't need to make a separate buffer to write to memory. Instead of make(buf), xx.Read(buf), mem.Write. We can do buf = mem.Read then xx.Read(buf)

api/wasm.go
```diff
@@ -293,21 +293,37 @@ type Memory interface {
 	// See math.Float64bits
 	ReadFloat64Le(ctx context.Context, offset uint32) (float64, bool)
 
-	// Read reads byteCount bytes from the underlying buffer at the offset or returns false if out of range.
+	// Read reads byteCount bytes from the underlying buffer at the offset or
+	// returns false if out of range.
+	//
+	// For example, to search for a NUL-terminated string:
```
codefromthecrypt, line 299: here's the example, which would be the same for integrating any other utility library that takes a []byte

mathetake (approved):

codefromthecrypt:
thx for the look @mathetake!

PR #623, opened 2022-06-06
wazeroir: backfill vector instruction unittets

mathetake:
This increases the test coverage on wazeroir/compiler.go by adding unttests which 
touch all lines related to V128 instructions. 


internal/wasm/func_validation_test.go
```diff
@@ -2709,8 +2709,7 @@ func TestModule_funcValidation_SIMD(t *testing.T) {
 	vi2v := func(vec OpcodeVec) (ret []byte) {
 		ret = addV128Const(ret)
 		return append(ret,
-			OpcodeI32Const,
-			1, 1, 1, 1,
+			OpcodeI32Const, 1,
```
mathetake, line 2712: needs leb128 encoding but yeah somehow this worked since before but fixed it 😄

codefromthecrypt: ah cool

codefromthecrypt (approved):
thanks for fixing the broken window!

Issue #626, opened 2022-06-13
Bindgen package for easily calling functions with arbitrary parameters


clarkmcc:
**Is your feature request related to a problem? Please describe.**
Calling guest functions and passing non-number values is a bit of a pain as noted [here](https://github.com/tetratelabs/wazero/issues/258#issuecomment-1145370290) (in issue https://github.com/tetratelabs/wazero/issues/258). It involves allocating guest memory from the host, then passing the pointers to the guest, the guest reads the data at those pointers, does some work, then allocates a response and the host then reads the response as raw bytes. I am proposing we introduce a bindgen package that simplifies the host -> guest interaction. Later we can look at introducing something to assist in guest -> host calls.

To start, I will be working on a PR that supports this for Rust-based guest WASM modules, and in the future, there may be a way we can provide the same integration for TinyGo guests.

**Describe the solution you'd like**
Most of the heavy lifting (and thinking) has already been done in [wasmedge-bindgen](https://github.com/second-state/wasmedge-bindgen). This existing repository relies on two guest functions and two host functions to help pass arguments around properly:
- (Host) `return_result` - Allows the guest to return a pointer and length of the result data.
- (Host) `return_error` - Allows the guest to return a pointer and length of the error message.
- (Guest) `allocate` - Allows the host to allocate guest memory
- (Guest) `deallocate` - Allows the host to deallocate guest memory

I propose we add the host functions to a new "wazero-bindgen" module. The Rust library providing the bindgen macros will optionally compile the two guest functions (allocate and deallocate) if a [feature](https://doc.rust-lang.org/cargo/reference/features.html) is enabled.

It would just be a matter of:
* Creating a new Go package like [this](https://github.com/second-state/wasmedge-bindgen/blob/main/host/go/bindgen.go) and providing a method on the `Module` interface so that we can call a function and pass in the arbitrary arguments.
* Creating a new Rust repository (perhaps `tetratelabs/wazero-bindgen-rs`) that contains the guest side of the bindgen like [this](https://github.com/second-state/wasmedge-bindgen/tree/main/bindgen/rust).

**Questions**
I plan on implementing this for a personal project, if this is of interest in the community, then it might make sense to get feedback on how the interfaces should look before I get started.
[...]

PR #627, opened 2022-06-13
Added proof-of-concept for Rust bindgen with wazero

clarkmcc:


codefromthecrypt:
breaks my heart to close PRs, but it is better to let you know as soon as possible. Please report back when you have a repo and I will start watching it.

Here's the long answer on why close https://github.com/tetratelabs/wazero/issues/626#issuecomment-1155820278

PR #630, opened 2022-06-17
Makes fake clocks increment and fixes mutability bug

codefromthecrypt:
This ensures fake clocks increment so that compilers that implement
sleep with them don't spin.

This also fixes a mutability bug in config where we weren't really doing
clone properly because map references are shared.

### DIFF
--- a/internal/platform/time.go
+++ b/internal/platform/time.go
@@ -2,19 +2,39 @@ package platform
 
 import (
 	"context"
+	"sync/atomic"
 	"time"
+
+	"github.com/tetratelabs/wazero/sys"
 )
 
-const FakeEpochNanos = int64(1640995200000000000) // midnight UTC 2022-01-01
+const (
+	ms = int64(time.Millisecond)
+	// FakeEpochNanos is midnight UTC 2022-01-01 and exposed for testing
+	FakeEpochNanos = 1640995200000 * ms
+)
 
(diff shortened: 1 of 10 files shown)


internal/platform/time.go
```diff
@@ -2,19 +2,39 @@ package platform
...
+// NewFakeWalltime implements sys.Walltime with FakeEpochNanos that increases by 1ms each reading.
+// See /RATIONALE.md
+func NewFakeWalltime() *sys.Walltime {
+	// AddInt64 returns the new value. Adjust so the first reading will be FakeEpochNanos
+	t := FakeEpochNanos - ms
+	var wt sys.Walltime = func(context.Context) (sec int64, nsec int32) {
+		wt := atomic.AddInt64(&t, ms)
+		return wt / 1e9, int32(wt % 1e9)
+	}
+	return &wt
```
mathetake, line 26: I think this is static, so maybe extract as the normal function rather than closure?

mathetake (approved):

internal/platform/time.go
codefromthecrypt: I can rework it to top-level the functions. they would still need a pointer receiver.

mathetake: huh, ic.. up to you!

codefromthecrypt: actually reworking it became more gnarly.. I'll leave it for now!

PR #632, opened 2022-06-17
arm64: implement vector comparisons 

mathetake:
This implements vector comparison instructions for 
arm64 backend.

part of #484

### DIFF
--- a/internal/asm/arm64/impl.go
+++ b/internal/asm/arm64/impl.go
@@ -3049,39 +3049,74 @@ var advancedSIMDThreeSame = map[asm.Instruction]struct {
 		return
 	}},
 	// https://developer.arm.com/documentation/ddi0596/2021-12/SIMD-FP-Instructions/UMAXP--Unsigned-Maximum-Pairwise-?lang=en
-	UMAXP: {U: 0b1, Opcode: 0b10100, qAndSizeResolver: func(arrangement VectorArrangement) (Q, size byte, err error) {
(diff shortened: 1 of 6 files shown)


internal/asm/arm64/impl.go
```diff
@@ -3049,41 +3049,76 @@ var advancedSIMDThreeSame = map[asm.Instruction]struct {
...
+		case VectorArrangement2D:
+			size, Q = 0b01, 1
+		default:
+			err = fmt.Errorf("unsupported arrangement %s for %s", arrangement.String(), InstructionName(FCMGE))
+		}
 		return
 	}},
 }
 
+func advancedSIMDThreeSameDefaultResolver(arrangement VectorArrangement) (Q, size byte, err error) {
```
codefromthecrypt, line 3117: curious why not rename the base function if this is just a dispatch? If there's a reason, maybe comment.

internal/engine/compiler/compiler_vec_test.go
```diff
@@ -2994,11 +2994,6 @@ func f64x2(f1, f2 float64) (ret [16]byte) {
 }
 
 func TestCompiler_compileV128Cmp(t *testing.T) {
-	if runtime.GOARCH != "amd64" {
```
codefromthecrypt, line 2997: cool

codefromthecrypt (approved):
good!

PR #629, opened 2022-06-16
wasi: Implements wasi_snapshot_preview1.poll_oneoff for relative clock events

codefromthecrypt:
This implements wasi_snapshot_preview1.poll_oneoff for relative clock events,
and in doing so stubs `Nanosleep` which defaults to noop, but can be configured
to `time.Sleep`.

PR #634, opened 2022-06-18
arm64: FP SIMD arithmetic/rounding instructions

mathetake:
This implements floating point SIMD arithmetics and rounding
instructions for arm64 backend.

Notably, now arm64 passes `simd_f32x4_arith.json`, `simd_f64x2_arith`,
`simd_f64x2`, `simd_f32x4`, `simd_f32x4_rounding`, `simd_f64x2_rounding`
spectests.

### DIFF
--- a/internal/asm/arm64/impl.go
+++ b/internal/asm/arm64/impl.go
@@ -2899,283 +2899,260 @@ func (a *AssemblerImpl) EncodeStaticConstToVectorRegister(n *NodeImpl) (err erro
 // advancedSIMDAcrossLanes holds information to encode instructions as "Advanced SIMD two-register miscellaneous" in
 // https://developer.arm.com/documentation/ddi0596/2021-12/Index-by-Encoding/Data-Processing----Scalar-Floating-Point-and-Advanced-SIMD?lang=en
 var advancedSIMDTwoRegisterMisc = map[asm.Instruction]struct {
-	U, Opcode byte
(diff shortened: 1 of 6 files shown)


internal/asm/arm64/impl.go
```diff
@@ -2899,283 +2899,260 @@ func (a *AssemblerImpl) EncodeStaticConstToVectorRegister(n *NodeImpl) (err erro
 // advancedSIMDAcrossLanes holds information to encode instructions as "Advanced SIMD two-register miscellaneous" in
 // https://developer.arm.com/documentation/ddi0596/2021-12/Index-by-Encoding/Data-Processing----Scalar-Floating-Point-and-Advanced-SIMD?lang=en
 var advancedSIMDTwoRegisterMisc = map[asm.Instruction]struct {
-	U, Opcode byte
-	// TODO: extract common implementation of qAndSizeResolver.
-	qAndSizeResolver func(arrangement VectorArrangement) (Q, Size byte, err error)
```
mathetake, line 2904: I just did the refactoring again, but nothing interesting... just decided to use map instead of anonymous functions
as this was getting hairy :D

mathetake:
`Test_Nanosleep` is failing but irrelevant as it is the flake introduced in #629

codefromthecrypt (approved):
Looks good and sensible naming also

PR #641, opened 2022-06-20
arm64: implement SIMD Integer arithmetics

mathetake:
This implements SIMD integer arithmetic instructions for arm64 backend.
Notably, now the engine passes  `simd_**_arith`, `simd_**_arith2`, and
`simd_**_sat_arith` spectests.

part of #484

### DIFF
--- a/internal/integration_test/asm/arm64_debug/impl_test.go
+++ b/internal/integration_test/asm/arm64_debug/impl_test.go
@@ -765,51 +765,6 @@ func TestAssemblerImpl_EncodeConstToRegister(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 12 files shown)


internal/integration_test/asm/arm64_debug/impl_test.go
```diff
@@ -765,51 +765,6 @@ func TestAssemblerImpl_EncodeConstToRegister(t *testing.T) {
 	}
 }
 
-func TestAssemblerImpl_EncodeSIMDByteToSIMDByte(t *testing.T) {
```
mathetake, line 768: these are migrated to asm/arm64/impl_test.go

codefromthecrypt (approved):
I like the map arrangment better! keep it coming!

Issue #618, opened 2022-06-03
Serialize a compiled module


clarkmcc:
**Describe the solution you'd like**
Is it possible to replicate [this feature](https://pkg.go.dev/github.com/wasmerio/wasmer-go/wasmer#Module.Serialize) from wasmer-go? Compiling a module results in a significant number of allocations, so I'd like to compile a batch of modules ahead of time to reduce the runtime memory overhead.

codefromthecrypt:
I guess what we are really talking about here is externalizing the compilation cache. plus possibly some guard to make sure if the structure of the code changes the cache is invalidated.

```go
		codes           map[wasm.ModuleID][]*code // guarded by mutex.
```

I think we can look into this after SIMD is done, wdyt @mathetake?

PS, related, but not exactly the same as this: https://github.com/tetratelabs/wazero/issues/179

F21:
I think this will also help us improve `mjml-go` performance: https://github.com/Boostport/mjml-go#benchmarks

Currently, spinning up a new worker using `InstantiateModule()` is quite expensive, so if we can clone a module, it would drive a lot of performance improvements.

codefromthecrypt:
@F21 I think you already have an issues about `InstantiateModule()` https://github.com/tetratelabs/wazero/issues/602 which is unrelated to this because Compile happens before that. Just setting expectations.

F21:
Oops, I must have misread. This feature would still potentially be useful for us. We're currently shipping a `.wasm` compressed with brotli, which is decompressed and compiled in `init()`. If the file size from serializing a compiled module is acceptable, we can just ship that directly in our library and remove the `.wasm` completely.

mathetake:
Note that this comes with severe security concerns -- as this means that we allow users to directly execute _any_ native code without validation passes (which are applied during `wazero.CompileBinary`). That means we also provide some ways to binary signing or something like that. 

Regardless, I will work on this after finishing SIMD instructions (== completion of Wasm 2.0 Draft). Stay tuned!

codefromthecrypt:
thanks for the feedback folks, this is great. don't worry too much about issue classification as we can sort it out.

mathetake:
I think all is set to implement the externalization of compiled native code...

mathetake:
[...]

PR #650, opened 2022-06-22
compiler: remove embedding of pointers of jump tables

mathetake:
This removes the embedding of pointers of jump tables (uintptr of []byte)
used by BrTable operations. That is the last usage of unsafe.Pointer in
compiler implementations.
Alternatively, we treat jump tables as asm.StaticConst and emit them 
into the constPool already implemented and used by various places.

Notably, now the native code compiled by compilers can be reusable
across multiple processes, meaning that they are independent of 
any runtime pointers.

part of #618

### DIFF
--- a/internal/asm/amd64/impl_staticconst.go
+++ b/internal/asm/amd64/impl_staticconst.go
@@ -40,12 +40,8 @@ func (a *AssemblerImpl) maybeFlushConstants(isEndOfFunction bool) {
 		}
 
 		for _, c := range a.pool.Consts {
-			offset := uint64(a.Buf.Len())
-			c.OffsetInBinary = offset
+			c.SetOffsetInBinary(uint64(a.Buf.Len()))
 			a.Buf.Write(c.Raw)
-			for _, callback := range c.OffsetFinalizedCallbacks {
-				callback(offset)
-			}
 		}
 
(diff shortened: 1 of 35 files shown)


internal/asm/amd64/impl_staticconst.go
```diff
@@ -8,69 +8,47 @@ import (
 	"github.com/tetratelabs/wazero/internal/asm"
 )
 
-type constPool struct {
```
mathetake, line 11: consolidated constPool impls into asm/assembler.go

internal/asm/assembler.go
```diff
@@ -53,11 +53,63 @@ type ConstantValue = int64
 
 // StaticConst represents an arbitrary constant bytes which are pooled and emitted by assembler into the binary.
 // These constants can be referenced by instructions.
-type StaticConst = []byte
+type StaticConst struct {
+	Raw []byte
```
codefromthecrypt, line 57: Do I understand that raw means we don't know the encoding because it could be anything? ex string iee754 etc?

codefromthecrypt (approved):
nice cleanup!

internal/asm/assembler.go
mathetake: Yeah can be any

PR #651, opened 2022-06-22
Clarify that wazero.Runtime can be assed as Namespace when instantiat…

anuraaga:
…ing a module

Not sure if this is the correct wording, but currently I had to look at the examples to find out how to initialize a module with host functions and no namespace.


codefromthecrypt:
can you please move this to the notes section? clarifying the type hierarchy of Runtime is more a note I think.

I'll correct the title meanwhile

builder.go
```diff
@@ -173,6 +173,18 @@ type ModuleBuilder interface {
 	Compile(context.Context, CompileConfig) (CompiledModule, error)
 
 	// Instantiate is a convenience that calls Compile, then Namespace.InstantiateModule.
+	// Ex.
```
codefromthecrypt, line 176:
```suggestion
	//
	// Ex.
```

codefromthecrypt (approved):
thanks, interfaces I think are still broken in godoc as examples don't embed as they should. Hence, we'll have to do more copy/pasting like this.

PR #654, opened 2022-06-24
arm64: remove CompileJumpToMemory

mathetake:
This was necessary for golang-asm compatibility but now unneeded 


codefromthecrypt (approved):
thanks for the shredding!

PR #655, opened 2022-06-24
arm64: remove golang-asm integeration test

mathetake:


codefromthecrypt (approved):
splitting to ensure IDE and GH doesn't crash works for me!

PR #657, opened 2022-06-26
asm: unexport arm64 pkg implementations

mathetake:


codefromthecrypt (approved):
sure. easier to find dead code!

PR #658, opened 2022-06-27
amd64: fix unsigned extension of i32 globals

### LINKED ISSUE #656: i64.extend_i32_u does signed shift on globals
Extending a global i32 returns a wrong result in the **compiled** version, the interpreter is correct.

This is go1.18.3 windows/amd64 with the current wazero 1489a9f19f39.

f.wat:
```
(module
(global $n i32 (i32.const -1))
(func $f (export "f") (result i64)
 global.get $n
 i64.extend_i32_u)
)
```

main.go:
```go
package main

import (
        "context"
        "fmt"
        "os"

)

func main() {
        file := os.Args[1]
        b, e := os.ReadFile(file)
        fatal(e)

        fmt.Printf("%s interpreted>\n", file)
        do(b, true)

        fmt.Printf("%s compiled>\n", file)
        do(b, false)
}
func do(wasm []byte, interpreted bool) {
        ctx := context.Background()

        var r wazero.Runtime
        if interpreted {
                r = wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfigInterpreter().WithWasmCore2())
        } else {
                r = wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfig().WithWasmCore2())
        }
        defer r.Close(ctx)

        m, e := r.InstantiateModuleFromBinary(ctx, wasm)
        fatal(e)

        j, e := m.ExportedFunction("f").Call(ctx)
        fatal(e)
        fmt.Printf("%x\n", j[0])
}
func fatal(e error) {
        if e != nil {
                panic(e)
        }
}
```

**output**:
```
j.wasm interpreted>
ffffffff
j.wasm compiled>
ffffffffffffffff
```

mathetake:
Fixes #656 


internal/engine/compiler/impl_amd64.go
```diff
@@ -2772,7 +2772,7 @@ func (c *amd64Compiler) compileExtend(o *wazeroir.OperationExtend) error {
 	if o.Signed {
 		inst = amd64.MOVLQSX // = MOVSXD https://www.felixcloutier.com/x86/movsx:movsxd
 	} else {
-		inst = amd64.MOVQ
+		inst = amd64.MOVL
```
mathetake, line 2775: this was a bug _if_ global i32's higher bits are set (which is also a kind of bug)

internal/wasm/module.go
```diff
@@ -555,7 +555,7 @@ func (m *Module) buildGlobals(importedGlobals []*GlobalInstance) (globals []*Glo
 		g := &GlobalInstance{Type: gs.Type}
 		switch v := executeConstExpression(importedGlobals, gs.Init).(type) {
 		case int32:
-			g.Val = uint64(v)
+			g.Val = uint64(uint32(v))
```
mathetake, line 558: previously this was sign-extending the i32 integers so higher bits were unnecessarily set for i32 globals.

codefromthecrypt (approved):
great

PR #659, opened 2022-06-27
asm: remove integration tests with golang-asm

mathetake:


mathetake:
lol 10k 😄

mathetake:
ok less than 10k

internal/asm/amd64/impl.go
```diff
@@ -10,148 +10,146 @@ import (
 	"github.com/tetratelabs/wazero/internal/asm"
```
mathetake, line 10: all the change in this file is only un-exporting so nothing fun

codefromthecrypt (approved):
cool, so what's left on golang-asm removal?

mathetake:
nothing left! removed completely... will change the title

PR #660, opened 2022-06-27
Removes WithWorkDirFS and "." resolution

codefromthecrypt:
This removes `WithWorkDirFS` and any other attempts to resolve the current directory (".") in host functions. This is a reaction to reality of compilers who track this inside wasm (not via host functions). One nice side effect is substantially simpler internal implementation of file-systems.

This also allows `experimental.WithFS` to block file access via passing nil.

---

Before, wazero's API included a `WithWorkDirFS`, which allowed control over which file a relative path such as "./config.yml" resolved to, independent of the root file system. This intended to help separate concerns like mutability of files, but it didn't work and is removed here.

Compilers that target wasm act differently with regard to the working directory. For example, while `GOOS=js` uses host functions to track the working directory, WASI host functions do not. wasi-libc, used by TinyGo, tracks working directory changes in compiled wasm instead: initially "/" until code calls `chdir`.

### DIFF
--- a/experimental/fs.go
+++ b/experimental/fs.go
@@ -8,16 +8,14 @@ import (
 	internalfs "github.com/tetratelabs/wazero/internal/sys"
 )
 
-// WithFS overrides fs.FS in the context-based manner. Caller needs to take responsibility for closing the filesystem.
+// WithFS overrides fs.FS in the context-based manner. Caller needs to take
+// responsibility for closing the filesystem.
 //
-// Note: This has the same effect as the same function name on wazero.ModuleConfig.
-func WithFS(ctx context.Context, fs fs.FS) (context.Context, api.Closer, error) {
-	fsConfig := internalfs.NewFSConfig().WithFS(fs)
-	preopens, err := fsConfig.Preopens()
-	if err != nil {
-		return nil, nil, err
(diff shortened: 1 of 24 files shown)


experimental/fs.go
```diff
@@ -8,16 +8,14 @@ import (
...
+// responsibility for closing the filesystem.
 //
-// Note: This has the same effect as the same function name on wazero.ModuleConfig.
-func WithFS(ctx context.Context, fs fs.FS) (context.Context, api.Closer, error) {
-	fsConfig := internalfs.NewFSConfig().WithFS(fs)
-	preopens, err := fsConfig.Preopens()
-	if err != nil {
-		return nil, nil, err
+// Note: This has the same effect as the same function on wazero.ModuleConfig.
+func WithFS(ctx context.Context, fs fs.FS) (context.Context, api.Closer) {
```
codefromthecrypt, line 15: @knqyf263 heads up. this removes the error return as it is no longer possible to err. Also, you can block filesystem access by passing nil.

mathetake (approved):
good

experimental/fs.go
knqyf263: Cool. Thanks for sharing.

PR #663, opened 2022-06-28
asm(amd64): resolve NOP padding TODO after golang-asm removal

mathetake:
Plus deleted out-dated comments


codefromthecrypt (approved):
thanks again

PR #664, opened 2022-06-28
spectest: precise assertion on (canonical,arithmetic) NaNs

mathetake:
This commit fixes interpreter and spectests so that we precisely match the 
semantics around NaN values. Previously, we round all the NaN values into
`math.NaN()` and only assert with `math.IsNaN` in the spectests. But now
we check the underlying bit pattern of NaN values and do assertion while 
differentiating arithmetic NaNs vs the canonical NaN.


mathetake:
whew glad that we don't need to modify compiler at all!

mathetake:
lol on again Go has different semantics for math.Trunc, Nearest, Ceil and Floor on RISCV vs others

mathetake:
almost all changes are in the re-generation of spectests so nothing special most of lines 😄

.github/workflows/spectest.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
       - name: Build scratch container
         run: |
           echo 'FROM scratch' >> Dockerfile
-          echo 'CMD ["/test", "-test.v"]' >> Dockerfile
```
mathetake, line 95: get rid of unnecessary verbosity :D

codefromthecrypt (approved):
great! less perl is best :D

PR #666, opened 2022-06-28
compiler: always allocate register to save conditional values

### LINKED ISSUE #665: failed to compile: invalid register [nil]
**Describe the bug**
Wazero fails to compile a wasm module I've compiled from Rust, which is compiled and executed successfully with Wasmtime and Wasmer.

**To Reproduce**
wasm module: https://github.com/olivierlemasle/wasm-text2png/releases/download/v0.1.0/wasm-text2png.wasm
It is a WebAssembly+WASI module compiled from Rust; its source code is in https://github.com/olivierlemasle/wasm-text2png (it simply uses the pure-Rust crate [text-to-png](https://crates.io/crates/text-to-png/) to format a text given as parameter to PNG.)
It is compiled with target `wasm32-wasi` using `cargo-wasi`.

I can compile and execute this wasm module successfully with Wasmtime and Wasmer.

My host code:

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/tetratelabs/wazero/wasi_snapshot_preview1"
)

func main() {
	if len(os.Args) < 2 {
		fmt.Println("Required arg: WASM file path")
		return
	}
	file := os.Args[1]

	ctx := context.Background()

	rConfig := wazero.NewRuntimeConfig().WithWasmCore2()
	r := wazero.NewRuntimeWithConfig(rConfig)
	defer r.Close(ctx)

	if _, err := wasi_snapshot_preview1.Instantiate(ctx, r); err != nil {
		log.Panicln(err)
	}

	wasm, err := os.ReadFile(file)
	if err != nil {
		log.Panicln(err)
	}

	_, err = r.CompileModule(ctx, wasm, wazero.NewCompileConfig())
	if err != nil {
		log.Panicln(err)
	}
	log.Println("Successfully compiled")
}

```

When executing with `go run . wasm-text2png.wasm`, it fails with:

```
panic: function[42/1870] failed to compile: invalid register [nil]
```

**Expected behavior**
I would expect this module to compile, as with Wasmtime or Wasmer, or fail with a more helpful message / debugging info.

- Go version: 1.18.3
- wazero Version: 2fa67b83bf6adca7b19905e5b0de7eea8e35f757
- Host architecture: amd64
- Runtime mode: compiler

mathetake:
Previously I had the assumption that we have always free registers for conditional value 
saving in compileLoadConditionalRegisterToGeneralPurposeRegister, but that is not the case 
when the conditional flag was created by floating-point comparison instructions.

Fixes #665 

### DIFF
--- a/internal/engine/compiler/impl_amd64.go
+++ b/internal/engine/compiler/impl_amd64.go
@@ -364,7 +364,9 @@ func (c *amd64Compiler) branchInto(target *wazeroir.BranchTarget) error {
 			// We can only re-use register state if when there's a single call-site.
 			// Release existing values on registers to the stack if there's multiple ones to have
 			// the consistent value location state at the beginning of label.
-			c.compileReleaseAllRegistersToStack()
+			if err := c.compileReleaseAllRegistersToStack(); err != nil {
+				return err
+			}
 		}
 		// Set the initial stack of the target label, so we can start compiling the label
 		// with the appropriate value locations. Note we clone the stack here as we maybe
@@ -459,7 +461,9 @@ func (c *amd64Compiler) compileBrIf(o *wazeroir.OperationBrIf) error {
 			// We can only re-use register state if when there's a single call-site.
 			// Release existing values on registers to the stack if there's multiple ones to have
 			// the consistent value location state at the beginning of label.
-			c.compileReleaseAllRegistersToStack()
+			if err := c.compileReleaseAllRegistersToStack(); err != nil {
+				return err
+			}
 		}
 		// Set the initial stack of the target label, so we can start compiling the label
 		// with the appropriate value locations. Note we clone the stack here as we maybe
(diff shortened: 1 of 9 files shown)


internal/engine/compiler/impl_amd64.go
```diff
@@ -4184,20 +4208,26 @@ func (c *amd64Compiler) compileLoadValueOnStackToRegister(loc *runtimeValueLocat
...
+	return
 }
 
 // loadConditionalRegisterToGeneralPurposeRegister saves the conditional register value
 // to a general purpose register.
-func (c *amd64Compiler) compileLoadConditionalRegisterToGeneralPurposeRegister(loc *runtimeValueLocation) {
-	// Get the free register.
-	reg, _ := c.locationStack.takeFreeRegister(registerTypeGeneralPurpose)
+func (c *amd64Compiler) compileLoadConditionalRegisterToGeneralPurposeRegister(loc *runtimeValueLocation) error {
+	reg, err := c.allocateRegister(loc.getRegisterType())
```
mathetake, line 4225: this is the fix: previously I had the assumption that we have always free registers for conditional value saves in `compileLoadConditionalRegisterToGeneralPurposeRegister`, but that is not the case when the conditional flag was created by floating-point comparison instructions. 

This is not only the case for amd64, but also on amd64.

mathetake:
I will backfill the unit test but this fixes the original bug already :D

codefromthecrypt (approved):
thanks and agreed that the extra context will help in case something similar occurs later!

olivierlemasle:
I confirm that this fixes my issue #665, thanks.

mathetake:
Thank you for confirming @olivierlemasle :D

PR #667, opened 2022-06-29
compiler: fix stackPointerCeil calculation

mathetake:

### DIFF
--- a/internal/engine/compiler/compiler_value_location.go
+++ b/internal/engine/compiler/compiler_value_location.go
@@ -202,6 +202,8 @@ func (v *runtimeValueLocationStack) push(loc *runtimeValueLocation) {
 		v.stack[v.sp] = loc
 	}
 	v.sp++
+	// stackPointerCeil must be set after sp is incremented since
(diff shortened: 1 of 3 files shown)


internal/engine/compiler/compiler_value_location.go
```diff
@@ -201,10 +201,10 @@ func (v *runtimeValueLocationStack) push(loc *runtimeValueLocation) {
 	} else {
 		v.stack[v.sp] = loc
 	}
+	v.sp++
```
mathetake, line 204: so stackPointerCeil must be set after sp is incremented :D

mathetake: as we skip the stack grow if `len(valueStack) >=  basePointer+stackPointerCeil`.

codefromthecrypt: sounds like a comment for the code :D

mathetake: done

codefromthecrypt (approved):
nice

PR #670, opened 2022-06-30
compiler: cleanup arch specific files

mathetake:


internal/engine/compiler/arch_amd64.go
```diff
@@ -16,11 +18,6 @@ type archContext struct{}
 // newArchContextImpl implements newArchContext for amd64 architecture.
 func newArchContextImpl() (ret archContext) { return }
 
-func init() {
```
mathetake, line 19: somehow there was two `init` 😄

codefromthecrypt (approved):
easily sensible!

PR #671, opened 2022-07-01
Adds support for FreeBSD

### LINKED ISSUE #607: Add compiler support for FreeBSD

codefromthecrypt:
FreeBSD was disabled due to lack of testing. This works around the
compilation problems. Note: We can't currently test arm64 automatically!

Notes:

* GitHub Actions doesn’t support FreeBSD, and may never.
* We could use Travis to run FreeBSD, but it would split our CI config.
* Using Vagrant directly is easier to debug than vmactions/freebsd-vm.
* GitHub Actions only supports virtualization on MacOS.
* GitHub Actions removed vagrant from the image starting with macos-11.
* Since VirtualBox doesn't work on arm64, freebsd/arm64 is untestable.

Fixes #607


mathetake (approved):
Awesome (glad it works without changing anything except mmap)

codefromthecrypt:
heh let's see.. I accidentally missed the build flags to actually use it with compiler 🤞

codefromthecrypt:
I can't currently reproduce the failure locally with vagrant even if deleting the VM and doing it again.

```bash
$ vagrant destroy
$ go version
go version go1.18.3 darwin/amd64
$ GOVERSION=$(go env GOVERSION) GOARCH=$(go env GOARCH) vagrant up --provision
--snip--
$ vagrant ssh -c "cd wazero && make spectest.v1"
ok  	github.com/tetratelabs/wazero/internal/integration_test/spectest/v1	1.854s
Connection to 127.0.0.1 closed.
```

codefromthecrypt:
can't try anymore because vagrant rate limits (via HTTP 429)

will park this meanwile

mathetake:
cool passed!

internal/engine/compiler/impl_amd64.go
```diff
@@ -1229,7 +1229,7 @@ func (c *amd64Compiler) compileCtz(o *wazeroir.OperationCtz) error {
 		return err
 	}
 
-	if runtime.GOOS != "darwin" {
+	if runtime.GOOS != "darwin" && runtime.GOOS != "freebsd" {
```
mathetake, line 1232: maybe this is not need and i guess only change on ^ `compileClz ` was necessary but not sure

PR #672, opened 2022-07-04
Enable wasm 2.0 support in tinygo examples to allow running with wasm…

anuraaga:
… built with latest

Noticed a new version of TinyGo with this commit, and confirmed building with it causes the tests to fail:

https://github.com/tinygo-org/tinygo/commit/b5c5d95b689979780ea01ed4c8d4fdb40602535b#diff-5016a7fcbfb8eb2b847f01d8a8490ce9720a2ee0af38b284743e18ed0ceda563R4


mathetake (approved):

codefromthecrypt:
looks good. can you update examples.yaml also, the version there is to ensure things work also.

codefromthecrypt (approved):
merge on green after examples.yaml bump!

Issue #601, opened 2022-05-29
Add emscripten_notify_memory_growth import? 


inkeliz:
**Is your feature request related to a problem? Please describe.**
Compiling using emscripten requires `emscripten_notify_memory_growth`, defined at https://emscripten.org/docs/api_reference/emscripten.h.html#abi-functions.

This is one single function and that is useless for Wazero. That is require for `malloc`.

Minimal code:

```
# include "stdint.h"
# include "stdlib.h"
# include "stdio.h"

uint8_t *InputMemory;
uint8_t *OutputMemory;

int main() {
    OutputMemory = (uint8_t *) malloc(8000000);
    InputMemory = (uint8_t *) malloc(8000000);

    if (InputMemory == NULL || OutputMemory == NULL) {
        printf("Erro");
        return 1;
    } else {
        printf("Okay");
        return 0;
    }
}
```

Compile with:
```
emcc wasm.c -o wasm.wasm -O3 -s ALLOW_MEMORY_GROWTH
```

Requires `emscripten_notify_memory_growth`.

**Describe the solution you'd like**
Maybe Wazero can define such import, so any emscripten compiled wasm will work without declaring additional modules.

**Describe alternatives you've considered**
It's possible to fix by using:

```
    env, _ := host.NewModuleBuilder("env").
        ExportFunction("emscripten_notify_memory_growth", func(_ int32) {}).
        Instantiate(ctx)
```

That is very small footprint, but very annoying to declare it everytime.

**Additional context**
Maybe we can do something similar of AssemblyScript and create one new Emscripten package? However, it's very small, maybe can be declared on WASI? But, it's not WASI anyway. :\

codefromthecrypt:
funny there's only one, and I guess it is multi-memory ready as the index is the memory index right? (always 0 at the moment)

Adding something for emscripten sounds fine as it can consolidate the docs and stop people from having to type out what you added, solely to ignore the callback. Of course, even in Go, you may want to pay attention to it, if there's an unsafe view of memory.

We can limit this to just the module part (ex similar to the /assemblyscript directory/package) and not dig into an example, yet. The default could be no-op and the builder could allow overriding via a  `func(context.Context, api.Module, uint32)` as I expect if some callback were needed, you'd need to grab the memory. Taking care to add godoc comment of that emscripten link as it is a handy reference.

Are you keen to give this a go?

inkeliz:
[...]

PR #673, opened 2022-07-05
Panics caller on exit error

codefromthecrypt:
This changes the AssemblyScript abort handler and WASI proc_exit
implementation to panic the caller which eventually invoked close.

This ensures no code executes afterwards, For example, LLVM inserts
unreachable instructions after calls to exit.

See https://github.com/emscripten-core/emscripten/issues/12322
See #601


codefromthecrypt:
need to backfill some unit tests so that we don't have this happen again.

internal/wasm/store.go
```diff
@@ -451,13 +452,17 @@ func (s *Store) instantiate(
...
+	m.CallCtx = NewCallContext(ns, m, sysCtx)
 
 	// Execute the start function.
 	if module.StartSection != nil {
 		funcIdx := *module.StartSection
 		f := m.Functions[funcIdx]
-		if _, err = f.Module.Engine.Call(ctx, m.CallCtx, f); err != nil {
+		_, err = f.Module.Engine.Call(ctx, m.CallCtx, f)
+
+		if exitErr, ok := err.(*sys.ExitError); ok { // Don't wrap an exit error!
```
anuraaga, line 463: Just wondering, would it make sense to not propagate code == 0 error here? Looking at the doc for `ExitError` I do see it mention the code == 0 case possibly being ok

https://github.com/tetratelabs/wazero/blob/main/sys/error.go#L15

But I never noticed it 😅 With Go muscle memory I generally assume that `err != nil` is a failure of some sort. `io.EOF` is similar so there's precedent, but same muscle memory always trips me up on that too.

codefromthecrypt: I was wondering the same thing this morning. Here's the only rationale I can think of that works.

* For a no return function, there's no issue except if you want to know if someone called exit or not. For example, it isn't always expected to call exit.
* For a function with a return, there's an issue as now you don't have a return value that you need.

What we could do to split the difference is that for the "start functions" we can swallow exit code zero, but otherwise not. Ex if you call a function with 1 return and that eventually calls exit, don't interfere.

If someone really wants to see exit code zero they can tell us between now and 1.0 even after beta starts end of aug.

codefromthecrypt: exit code zero is uncommon even if emscripten will return it. I checked tinygo and zig compiled wasi-libc. Since it invalidates the module (causing future functions to be broke), I think the best thing is to document the rare case chance of an implicit exit zero and hope it stays rare!

codefromthecrypt:
next step: test wasi with rust to verify the exit code zero thing.. then polish (a couple missing tests and RATIONALE)

codefromthecrypt:
ok leaving out a rust variant of wasi as it is massive and uses wasi-libc which the other example in this PR already does.

https://github.com/rust-lang/rust/issues/73432

PR #674, opened 2022-07-06
Remove cfg_attr from rust example since doesn't do anything

anuraaga:
Noticed these seem to be noise so thought the example can be clearer without them. Or alternatively, maybe it's a nice example to use snake_case method name and camel case export name to show how to map a rust convention name to a different one?

For some random context, I was hoping these would allow me to export functions as `malloc` / `free` when targeting `wasm32-wasi`, but it didn't work, even though the wasi target doesn't actually export malloc / free it still seems to give a duplicate symbol error even with the renaming... :( I don't think there is a way to export allocation functions from tinygo with any other name since `runtime.` methods are private, so my current pattern for allocations from host to try to support both is `if ExportedFunction("wazero_malloc") else ExportedFunction("malloc")`


codefromthecrypt:
FYI the main reason I put these here was less about the ability to rename, more to clarify that these are for wasm.

I think the "doesn't do anything" part is dependent on the fact that we currently don't build the same file for anything except wasm (ex we have instructions to select the wasm target and we only use that)

Other code that does multi-platform will use these attributes, which is why I added them (and where I added them from)
https://github.com/search?l=Rust&q=target_arch+wasm32&type=Code

Anyway, I'm ok with removing this, but I'd re-title this to remove confusion. They are valid attributes it is just that if you follow our instructions and use the indicated cargo args, they are implicit.

codefromthecrypt:
The below actually does work as it exports the function `_greet` as the name `greet`.

```rust
#[cfg_attr(
all(target_arch = "wasm32", target_os = "unknown"),
export_name = "greet"
)]
#[no_mangle]
pub unsafe extern "C" fn _greet(ptr: u32, len: u32) {
    greet(&ptr_to_string(ptr, len));
}
```

We have similar example of how to export as a different name in the tinygo example.
```go
//export greet
func _greet(ptr, size uint32) {
```

I'd suggest closing this or revising it to simplify like this (the target_os doesn't matter)

```
#[cfg_attr(all(target_arch = "wasm32"), export_name = "greet")]
```

I think it is a separate topic if rust allows renaming the internal exports it uses for allocation, and not necessarily something to optimize for. However, if you can't rename otherwise we do have `api.ImportRenamer` for that.

PR #676, opened 2022-07-07
examples: adds rust build and WASI example

### LINKED ISSUE #619: Rust/WASI example
I've copied the WASI example nearly exactly but I can't seem to get it to work with a Rust guest, for example, my scripts panic when I try to use `time::Instant`, and the `.WithArgs` aren't available when calling `std::env::args()`. Is there an example or anything special I need to do to get it working? If I were just shooting in the dark, I'd say that Rust just isn't expecting those host functions to go into the `wasi_snapshot_preview1` module name, but again, I'm just spitballing.

An example of using Rust and WASI.

codefromthecrypt:
This fixes where our pull requests didn't check the rust source in
examples were valid. It also adds an example of wasi with rust.

This uses cargo-wasi because the default target creates almost 2MB of
wasm for a simple cat program.

Fixes #619


examples/wasi/README.md
```diff
@@ -8,20 +8,51 @@ $ go run cat.go /test.txt
 greet filesystem
 ```
 
-If you do not set the environment variable `WASM_COMPILER`, main defaults
+If you do not set the environment variable `TOOLCHAIN`, main defaults
 to use Wasm built with "tinygo". Here are the included examples:
 
+* [cargo-wasi](testdata/cargo-wasi) - Built via `cargo wasi build --release; mv ./target/wasm32-wasi/release/cat.wasm .`
 * [tjnygo](testdata/tinygo) - Built via `tinygo build -o cat.wasm -scheduler=none --no-debug -target=wasi cat.go`
```
codefromthecrypt, line 15:
```suggestion
* [tinygo](testdata/tinygo) - Built via `tinygo build -o cat.wasm -scheduler=none --no-debug -target=wasi cat.go`
```

examples/wasi/README.md
```diff
@@ -8,20 +8,51 @@ $ go run cat.go /test.txt
...
+
+### cargo-wasi (Rustlang)
+
+The [Rustlang source][testdata/cargo-wasi] uses [cargo-wasi][1] because the
+normal release target `wasm32-wasi` results in almost 2MB for a simple program.
+
+Concretely, if you use `cargo build --target wasm32-wasi --release`, the binary
+`./target/wasm32-wasi/release/cat.wasm` is over 1.9MB. This is too large to
+check into source control. While `cargo wasi build --release` is still much
+larger than similar code in TinyGo or C, it 85KB is a lot more reasonable to
```
codefromthecrypt, line 41:
```suggestion
larger than similar code in TinyGo or C, 85KB is a lot more reasonable to
```

codefromthecrypt:
@mathetake FYI I finally added the make rule for rust, rather than kicking it down the road any longer. Also, I switched the example job to actually run tests as the wasm source could compile without error, but still not work in the example.

PR #677, opened 2022-07-07
Fix computation of offset in arm64 compiler

anuraaga:
I have a wasm binary (9.3MB rust wasi debug build for reference) that works on amd64 but gives an error on compile on arm64

`instantiating plugin: function[1915/3143] failed to compile: relative jump offset 102451936/4 must be within -33554432 and 33554431`

While I'm not that familiar with what these lines are doing, based on the comments and error message string, the change seems like what's needed :yolo:

I could confirm this change allowed my wasm to compile, though I'm not sure how to write a unit test for this.


codefromthecrypt (approved):
This looks correct, as we can see by the match on the error message.

I think we are missing error tests in TestAssemblerImpl_EncodeRelativeJump as well what you are noticing here, which is an offset value that is out-of bounds unless divided by 4. Please backfill these into the existing test matrixes if you don't mind (notice there's also a typo in that file "forward" not "forward"

codefromthecrypt:
only nit if you want is that we don't tend to use sprintf in test expectations as that has a risk of copy/pasta which drifts in ways difficult to track down. string literals are easier to know will keep working.

internal/asm/arm64/impl_6_test.go
```diff
@@ -2,13 +2,42 @@ package arm64
...
 	"encoding/hex"
+	"fmt"
 	"testing"
 
 	"github.com/tetratelabs/wazero/internal/asm"
 	"github.com/tetratelabs/wazero/internal/testing/require"
 )
 
 func TestAssemblerImpl_EncodeRelativeJump(t *testing.T) {
+	t.Run("offset limits", func(t *testing.T) {
```
anuraaga, line 13: I couldn't come up with a way to test these in the existing tests because it would mean having many megabytes of `expHex` making the test unwieldy.

For similar reasons, I couldn't come up with a way to test the negative cases here, since I think it requires a backwards jump, which then ends up again needing megabytes of code due to this line

https://github.com/tetratelabs/wazero/blob/main/internal/asm/arm64/impl.go#L756

Let me know if this is OK or any ideas for improving

codefromthecrypt (approved):

mathetake:
oh my bad ... thank you for this @anuraaga !

PR #678, opened 2022-07-08
Adds Emscripten integration

### LINKED ISSUE #601: Add emscripten_notify_memory_growth import?

codefromthecrypt:
This adds a toehold integration for emscripten users, so they don't have to define the memory growth function.

Ex. if you aren't using "env" for anything:

```go
_, err := emscripten.Instantiate(ctx, r)
```

Otherwise, you can add emscripten to an "env" builder:
```go
// Construct your own module builder for "env" with any functions you need.
envBuilder := r.NewModuleBuilder("env").
	ExportFunction("get_int", func() uint32 { return 1 })

// Now, add Emscripten special function imports into it.
emscripten.NewFunctionExporter().ExportFunctions(envBuilder)

// Finally, instantiate env.
_, err := envBuilder.Instantiate(ctx, r)
```

Fixes #601


examples/allocation/emscripten/testdata/greet.c
```diff
@@ -0,0 +1,42 @@
+/* Note: EMSCRIPTEN_KEEPALIVE is a macro that exports the function and prevents
+   dead-code elimination on it. */
+#include <emscripten.h>
+#include <stdlib.h> // for malloc/free
+#include <string.h> // for strncat
+#include <stdint.h> // for uint64_t
+
+// _log is a WebAssembly import which prints a string (linear memory offset,
+// byteCount) to the console.
```
mathetake, line 9: maybe I would use `/* */` for all comments

codefromthecrypt:
update I have a port to cpp for this locally will resume on monday or sooner

examples/allocation/emscripten/testdata/greet.c
codefromthecrypt: I removed the allocation example as it became really big as emscripten also requires WASI. For starters, just focused on the issue asked about.

emscripten/emscripten_test.go
```diff
@@ -0,0 +1,45 @@
...
+	_, err = Instantiate(ctx, r)
+	require.NoError(t, err)
+
+	// Emscripten exits main with zero by default
+	_, err = r.InstantiateModuleFromBinary(ctx, growWasm)
+	require.Error(t, err)
+	require.Zero(t, err.(*sys.ExitError).ExitCode())
+
+	// We expect the memory no-op memory growth hook to be invoked.
+	require.Contains(t, log.String(), "==> env.emscripten_notify_memory_growth(memory_index=0)")
```
codefromthecrypt, line 44: here shows the memory growth hook was called (which means the test cc file is correct)

Makefile
```diff
@@ -57,10 +57,36 @@ build.examples.tinygo: $(tinygo_sources)
 c_sources := $(wildcard examples/*/testdata/*.c examples/*/*/testdata/*.c examples/*/testdata/*/*.c)
 .PHONY: build.examples.zig-cc
 build.examples.zig-cc: $(c_sources)
-	for f in $^; do \
+	@for f in $^; do \
 	    zig cc --target=wasm32-wasi -O3 -o $$(echo $$f | sed -e 's/\.c/\.wasm/') $$f; \
 	done
 
+# Here are the emcc args we use:
```
codefromthecrypt, line 64: aka 100 browser tabs

Makefile
```diff
@@ -57,10 +57,36 @@ build.examples.tinygo: $(tinygo_sources)
...
+#
+# * `-Oz` - most optimization for code size.
+# * `--profiling` - adds the name section.
+# * `-s STANDALONE_WASM` - ensures wasm is built for a non-js runtime.
+# * `-s EXPORTED_FUNCTIONS=_malloc,_free` - export allocation functions so that
```
codefromthecrypt, line 69: this and below aren't needed for our current example, but are needed for things like imports. Since they take a while to figure out, I left them in.

Ex. this would require the WARM_ON_UNDEFINED_SYMBOLS=0 thing:
```c++
/* _log is a WebAssembly import which prints a string (linear memory offset,
byteCount) to the console. */
EM_IMPORT(log) extern "C" void _log(const char* ptr, uint32_t size);
```

PR #680, opened 2022-07-11
Removes api.ImportRenamer for a different approach to "env"

codefromthecrypt:
Before, we introduced a type `api.ImportRenamer` to resolve conflicts
where the "env" module was shared between AssemblyScript and
user-defined functions. This API was never used in GitHub, and is
complicated.

AssemblyScript also isn't the only ABI to share the "env" module, as
other web APIs like Emscripten do also. The less complicated approach is
to have packages that need to share "env" use
`ModuleBuilder.ExportFunctions` instead, and use namespaces as needed if
there is overlap.

Ex.
```go
// First construct your own module builder for "env"
envBuilder := r.NewModuleBuilder("env").
	ExportFunction("get_int", func() uint32 { return 1 })

// Now, add AssemblyScript special function imports into it.
envBuilder.ExportFunctions(assemblyscript.NewFunctionExporter().
	WithAbortMessageDisabled().ExportFunctions())
```


assemblyscript/assemblyscript_example_test.go
```diff
@@ -0,0 +1,45 @@
...
+	// This adds the "env" module to the runtime, with AssemblyScript's special
+	// function imports.
+	if _, err := assemblyscript.Instantiate(ctx, r); err != nil {
+		log.Panicln(err)
+	}
+
+	// Output:
+}
+
+// This shows how to instantiate AssemblyScript's special imports when you also
```
codefromthecrypt, line 28: This is the important part, and will end up embedded in the godoc

mathetake (approved):
like it better

PR #681, opened 2022-07-12
Adds ExportedFunctions API on CompiledModule.

### LINKED ISSUE #637: Expose more information from CompiledModule
After having created a `CompiledModule`, I'd like to infer some information from it. For example, I'd like to know what type of functions it wants imported and what kind of functions it is exporting, amongst other things.

`CompiledModule` (or actually, it's underlaying struct `compiledModule{}`) seems to embed `internal/wasm.Module{}` which appears to hold that information. Perhaps we could bubble up that information to the public API?

What I can do right now is create an instance from the `CompiledModule` and offer it all the imports available to me in the hope that it satisfies what the module needs. Then I can poke around with functions like `ExportedFunction()` and poll for information.

Obviously, this is not a great solution and this solution doesn't help me figure out what imports a module needs.

mathetake:
Closes #637 

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -173,7 +173,28 @@ type Closer interface {
 	Close(context.Context) error
 }
 
-// Function is a WebAssembly 1.0 (20191205) function exported from an instantiated module (wazero.Runtime InstantiateModule).
+// ExportedFunction is a WebAssembly function exported in a module (wazero.CompiledModule).
+//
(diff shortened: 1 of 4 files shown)


codefromthecrypt (approved):
Thanks. Actually I think this is enough to close https://github.com/tetratelabs/wazero/issues/637

I also needed this in dapr, so happy to see it!

mathetake:
Thanks for the look!

PR #682, opened 2022-07-12
ir: correctly ignore unreachable br/br_if

mathetake:
This is the first bug found by wazero-fuzz repository:

```wat
(module
  (type (func))
  (func (type 0)
    br 0 ;; return the function, therefore the following instructions are unreachable.
    block
      br 1 ;; <------------ this is unreachable and must be ignored, not causing panic :D 
    end))
```


codefromthecrypt (approved):
excellent

PR #683, opened 2022-07-12
ir: fixes a range for drop operation on vector values

mathetake:
Fixes the second bug found by wazero-fuzz. Previously, Drop instructions on the
vector values were wrongly translated and resulted in panic both on amd64 and arm64,
though that is not covered by spectes and practically doesn't appear in the real world program. 


codefromthecrypt (approved):
great progress!

PR #684, opened 2022-07-12
binary: allow externref in element segments

mathetake:
Even though this is in practice will never be used as of 2.0 since the only thing they could do is to add null references into the table (which is by default), this is allowed in spec but not fully covered. I found this via wazero-fuzz.


codefromthecrypt (approved):
treasure trove of bugs :D

PR #686, opened 2022-07-13
Top-levels FunctionDefinition to allow access to all function metadata

### LINKED ISSUE #637: Expose more information from CompiledModule

codefromthecrypt:
This top-levels `api.FunctionDefinition` which was formerly
experimental, and also adds import metadata to it. Now, it holds all
metadata known at compile time.

Here are the public API visible changes:
* api.ExportedFunction - replaced with api.FunctionDefinition as it is
  usable for all types of functions.
* api.Function - `.ParamTypes/ResultTypes()` are replaced with
  `.Definition().
* api.FunctionDefinition - extracted from experimental and adds
  `.Import()` to get the any imported module and function name.
* experimental.FunctionDefinition - replaced with
  api.FunctionDefinition.
* experimental.FunctionListenerFactory - adds first arg of the
  instantiated module name, as it can be different than compiled.
* wazero.CompiledModule - Adds `.ImportedFunctions()` and changes result
  type of `.ExportedFunctions()` to api.FunctionDefinition.

Internally, logic to create function definition are consolidated between
host and wasm-defined functions, notably wasm.Module now includes
`.BuildFunctionDefinitions()` which reduces duplication in
wasm.ModuleInstance `.BuildFunctions()`,

This obviates #681 by deleting the `ExportedFunction` type which
overlaps with this information.

This fixes #637 as it includes more metadata including imports.


internal/wasm/function_definition.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	return
+}
+
+// BuildFunctionDefinitions generates function metadata that can be parsed from
+// the module. This must be called after all validation.
+//
+// Note: This is exported for tests who don't use wazero.Runtime or
+// NewHostModule to compile the module.
+func (m *Module) BuildFunctionDefinitions() *Module {
```
mathetake, line 32: do we need to return the self pointer?

mathetake (commented):
now Is it possible to enhance the compilation error message with the actual name, not only the index here? https://github.com/tetratelabs/wazero/blob/c00afaa37e735049f360ab2595423d9410b7ee9c/internal/engine/compiler/engine.go#L444

internal/wasm/function_definition.go
```diff
@@ -0,0 +1,151 @@
...
+func (f *functionDefinition) Name() string {
+	return f.name
+}
+
+// Import implements the same method as documented on api.FunctionDefinition.
+func (f *functionDefinition) Import() (moduleName, name string, isImport bool) {
+	if importDesc := f.importDesc; importDesc != nil {
+		return importDesc[0], importDesc[1], true
+	}
+	return "", "", false
```
mathetake, line 130: less 
```suggestion
	if importDesc := f.importDesc; importDesc != nil {
		moduleName, name, isImport = importDesc[0], importDesc[1], true
	}
	return
```

mathetake (approved):
much better thanks!

internal/wasm/function_definition.go
codefromthecrypt: no, just did it for convenience when inlining wasm modules

codefromthecrypt:
> now Is it possible to enhance the compilation error message with the actual name, not only the index here? 

ok now we use debugname everywhere

internal/engine/compiler/compiler_test.go
```diff
@@ -19,6 +20,104 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+// Ensures that the offset consts do not drift when we manipulate the target
+// structs.
+//
+// Note: This is a package initializer as many tests could fail if these
+// constants are misaligned, hiding the root cause.
+func init() {
```
codefromthecrypt, line 28: this is new and allows an early break on offset problem

codefromthecrypt:

PR #691, opened 2022-07-13
Read unsigned leb128 for the index of ref.func in global initializer

mathetake:

### DIFF
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -627,10 +627,10 @@ func executeConstExpression(importedGlobals []*GlobalInstance, expr *ConstantExp
 			v = GlobalInstanceNullFuncRefValue
 		}
 	case OpcodeRefFunc:
-		// For ref.func const expresssion, we temporarily store the index as value,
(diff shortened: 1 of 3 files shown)


internal/wasm/store.go
```diff
@@ -627,10 +627,10 @@ func executeConstExpression(importedGlobals []*GlobalInstance, expr *ConstantExp
 			v = GlobalInstanceNullFuncRefValue
 		}
 	case OpcodeRefFunc:
-		// For ref.func const expresssion, we temporarily store the index as value,
+		// For ref.func const expression, we temporarily store the index as value,
 		// and if this is the const expr for global, the value will be further downed to
 		// opaque pointer of the engine-specific compiled function.
-		v, _, _ = leb128.DecodeInt32(r)
```
mathetake, line 633: `ref.func` instruction's immediate index is encoded as unsigned leb128, not signed one!

codefromthecrypt (approved):
keep on fuzzing!

PR #693, opened 2022-07-13
func_validation: do not modify original types in unreachable state

mathetake:


codefromthecrypt (approved):
thanks for the 🧹

PR #694, opened 2022-07-13
Adds test-time only build tag for assertions in main code

mathetake:


internal/engine/interpreter/interpreter.go
```diff
@@ -848,6 +848,21 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, callCtx *wasm.CallCont
 	bodyLen := uint64(len(frame.f.body))
 	for frame.pc < bodyLen {
 		op := frame.f.body[frame.pc]
+
+		var checkPc bool
+		var prevPc uint64
```
mathetake, line 853: I strongly believe that these vars are optimized out as well if unused if IsTest == false

mathetake:
talked offline, I will follow the path of having a feature flag for testing rather than build flag when another necessity comes up

mathetake:
I think I will close this :D

PR #695, opened 2022-07-14
compiler(amd64),interpreter: signed-extend to 32-bit in V128ExtractLane.

mathetake:


mathetake:
ok turns out that amd64 also fails :D

mathetake:
fixed amd64 ... which wrongly extracted the i8x16 to 64-bit in signed version

codefromthecrypt (approved):
so much good

PR #696, opened 2022-07-14
Support for select instructions on vector values

mathetake:
This adds support for select instructions whose operands are vector values.
This hasn't been covered by the specification tests.


mathetake:
ok done with interpreter and arm64...

codefromthecrypt (approved):
feels like we are making the more precise variant of spectests!

PR #698, opened 2022-07-14
Makes ExportedFunctions unique

codefromthecrypt:
Exported functions are easier to use as a map vs making the callers do
it.

### DIFF
--- a/config.go
+++ b/config.go
@@ -267,15 +267,24 @@ func (c *runtimeConfig) WithWasmCore2() RuntimeConfig {
 //
 // Note: Closing the wazero.Runtime closes any CompiledModule it compiled.
 type CompiledModule interface {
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
here's the call site from dapr

```go
// preflightCheckWasm check that the wasm contains functions exported by TinyGo.
func (rh *wazeroRequestHandler) preflightCheckWasm() error {
	exportedFunctions := rh.compiled.ExportedFunctions()

	for _, fn := range []string{wasmFuncMalloc, wasmFuncFree, wasmFuncRewrite} {
		if _, ok := exportedFunctions[fn]; !ok {
			return fmt.Errorf("module[%s] didn't export function[%s]", rh.compiled.Name(), fn)
		}
	}

	return nil
}
```

mathetake (approved):

PR #699, opened 2022-07-14
ir: rename NeedsAccess* -> Has*, and make them module-scoped

mathetake:
Previously, `HasDataInstances` and `HasElementInstances` have been
function-scoped and defined depending on if the function uses them or not.

In a compiler, we use these booleans to decide to initialize the callEngine's 
fields, but the initialization only happens if the called module instance changes.

Therefore, if function A and function B are in the same module but B only access
these instances, and A calls B, then there was an invalid memory access in B
when it invokes `data.init` instruction for example.


internal/wazeroir/compiler.go
```diff
@@ -206,10 +206,6 @@ type CompilationResult struct {
 	HasMemory bool
 	// HasTable is true if the module from which this function is compiled has table declaration.
 	HasTable bool
-	// NeedsAccessToDataInstances is true if the function needs access to data instances via memory.init or data.drop instructions.
-	NeedsAccessToDataInstances bool
```
mathetake, line 210: this was the culprit -- this is determined by per-function scope here, but was used
in the `compileModuleContextInitialization` which is module scope.

mathetake:
oh well lemme fix a bit

mathetake:
ok renamed the title and description! @codefromthecrypt

codefromthecrypt (approved):

Issue #512, opened 2022-04-28
compiler: optimize bulk memory operations


mathetake:
Currently the assembly implementations of `memory.copy` and `memory.fill` (and table equivalents) are pretty native 
and simply do the for-loop copying one by one.

amd64: https://github.com/tetratelabs/wazero/blob/266320e257e6a0496781aa740c79e61833d7d83a/internal/wasm/jit/jit_impl_amd64.go#L3536

arm64:
https://github.com/tetratelabs/wazero/blob/266320e257e6a0496781aa740c79e61833d7d83a/internal/wasm/jit/jit_impl_arm64.go#L3071

Therefore, there's plenty of rooms for improvements wrt performance as we can see in, for example, musl libs' assembly:

### `memory.copy`, `table.copy`, `memory.init` and `table.init`
- arm64: https://git.musl-libc.org/cgit/musl/tree/src/string/aarch64/memcpy.S?id=v1.2.1
- amd64: https://git.musl-libc.org/cgit/musl/tree/src/string/x86_64/memcpy.s?id=v1.2.1

### `memory.fill`
- arm64: https://git.musl-libc.org/cgit/musl/tree/src/string/aarch64/memset.S?id=v1.2.1
- amd64: https://git.musl-libc.org/cgit/musl/tree/src/string/x86_64/memset.s?id=v1.2.1

codefromthecrypt:
@wdvxdr1123 interested/have time to give this a try?

dranikpg:
Hi @mathetake!
One more valuable source must be GOlangs memmove implementation: https://go.dev/src/runtime/memmove_amd64.s
It has some comments on why particular decisions where made.
I've made some x86 experiments today on my own to find a fast and easy to implement memcpy function. You can check the sources out at https://github.com/dranikpg/memcpy-perf-research/. 

Wazero currently uses a simple copy loop (orange "loop" on the plot). While it is really hard to be more performant on small sizes, advanced implementations are over several times faster on mid and large sized arrays. Opt - is a combination of "loop 8byte" and "movs 8byte" with some additional cases for small values. 

I'd personally choose "loop 8byte", because there won't be any hassle with the currently not present REP, MOVSX instructions and special register cases. Movs of all kinds are already implemented in wazeros assembler. I'm no expert on amd64, but I'll guess an 8byte loop will perform similarly well. What do you think?

[...]

PR #700, opened 2022-07-14
Optimize memory bulk operations: copy on x86 

dranikpg:
First part for #512

Huh, this has been somewhat difficult to implement and debug at times 😅 

Because of the additional checks and rep instruction startup cost, the new implementation should be a little slower for sizes 0-16, but multiple times faster for larger sizes.

Resulting benchmarks: https://github.com/tetratelabs/wazero/pull/700#issuecomment-1185570467

Brief assembly changes for compileMemoryCopy:

A single byte copy loop:

```asm
.loop:
MOVB tmp, src[size]
MOVB dest[size], tmp
DEC size
CMP size, 0
JNE .loop
```

became an efficient `REP MOVSQ` + a copy loop for the remaining %8 bytes

```asm
.loop:
TEST size, 0b0111  ; = %8
JZ .end
MOVB tmp, src
MOVB dest, tmp
INC src
INC dst
JMP .loop

SHR size, 3 ; /= 8

; exchange (dest, src, size) with (rdi, rsi, rcx)

REP MOVSQ

; exchange registers back
```

For compileTableCopy, an 8-byte copy loop became an `REP MOVSQ`, which is of equal performance. A bit slower for small sizes, somewhat faster for larger sizes. https://github.com/tetratelabs/wazero/issues/512#issuecomment-1173166362


dranikpg (commented):
Some minor decisions left before being really "ready for review"

internal/asm/amd64/impl.go
```diff
@@ -2115,6 +2125,17 @@ func (a *AssemblerImpl) encodeConstToRegister(n *nodeImpl) (err error) {
 		} else {
 			a.WriteConst(n.srcConst, 32)
 		}
+	case TESTQ:
```
mathetake, line 2128: let's add tests for this!

internal/asm/amd64/impl.go
```diff
@@ -786,6 +786,14 @@ func (a *AssemblerImpl) encodeNoneToNone(n *nodeImpl) (err error) {
...
 		// https://mudongliang.github.io/x86/html/file_module_x86_id_318.html
 		_, err = a.buf.Write([]byte{0x0f, 0x0b})
+	case REP_MOVSB:
+		_, err = a.buf.Write([]byte{0xf3, 0xa4})
+	case REP_MOVSQ:
+		_, err = a.buf.Write([]byte{0xf3, RexPrefixW, 0xa5})
+	case STD:
+		_, err = a.buf.Write([]byte{0xfd})
+	case CLD:
+		_, err = a.buf.Write([]byte{0xfc})
```
mathetake, line 796: same for these. let's add unit tests

internal/engine/compiler/impl_amd64.go
```diff
@@ -3716,63 +3703,74 @@ func (c *amd64Compiler) compileCopyImpl(isTable bool, srcTableIndex, dstTableInd
 			c.assembler.CompileRegisterToRegister(amd64.ADDQ, amd64ReservedRegisterForMemory, sourceOffset.register)
 		}
 
-		beginCopyLoop := c.assembler.CompileStandAlone(amd64.NOP)
+		// handle copySize % 8 in manual mov loop
+		if !isTable {
```
mathetake, line 3707: I think maybe it's time to stop having `compileCopyImpl` as it's getting hairy and complex... could you just implement compileMemoryCopy and compileMemoryTable?

mathetake (commented):
Things to do before mark this ready for review!
* increase unit test more coverage for assembler changes 
* split compileCopyImpl into `compileMemoryCopy` and `compileTableCopy`
* Pass all the unit tests locally `make test` and `make spectests`

Plus I would like you to add the benchmark in the https://github.com/tetratelabs/wazero/tree/main/internal/integration_test/bench where you
* check in `memory.wat` and it's compiled wasm `memory.wasm` into internal/integration_test/bench/testdata/*
* use the wasm binary and adds the benchmark `BenchmarkBulkMemory` in https://github.com/tetratelabs/wazero/blob/main/internal/integration_test/bench/bench_test.go

internal/engine/compiler/impl_amd64.go
```diff
@@ -3638,66 +3638,53 @@ func (c *amd64Compiler) compileCopyImpl(isTable bool, srcTableIndex, dstTableInd
...
+			c.assembler.CompileMemoryToRegister(amd64.MOVQ, amd64ReservedRegisterForCallEngine, callEngineModuleContextTablesElement0AddressOffset, tmp)
+			c.assembler.CompileMemoryToRegister(amd64.MOVQ, tmp, int64(srcTableIndex*8), tmp)
+			// Compare length.
+			c.assembler.CompileMemoryToRegister(amd64.CMPQ, tmp, tableInstanceTableLenOffset, reg)
+		} else {
+			c.assembler.CompileMemoryToRegister(amd64.CMPQ,
+				amd64ReservedRegisterForCallEngine, callEngineModuleContextMemorySliceLenOffset,
+				reg)
+		}
+		return c.assembler.CompileJump(amd64.JCC)
```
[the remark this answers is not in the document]
mathetake: pls define this as top-level function, not the closure!

dranikpg:
Fixed your comments

The benchmark currently gives the following results (ns/op):


| Size/Direction  | Old | New | Ratio |
| :---        |  :---   |  :--- | :---: |
|8 F|7.911|11.22|0.71|
|8 B|7.674|10.98|0.7|
|24 F|16.15|12.25|1.32|
|24 B|12.23|12.21|1.0|
|64 F|34.45|12.78|2.7|
|64 B|23.01|12.92|1.78|
|128 F|72.47|14.87|4.87|
|128 B|44.38|14.47|3.07|
|1000 F|482.7|33.62|14.36|
|1000 B|256.3|105.1|2.44|
|4000 F|2101.0|83.95|25.03|
|4000 B|1194.0|308.5|3.87|
|16000 F|9097.0|293.2|31.03|
|16000 B|3842.0|1051.0|3.66|
|64000 F|30117.0|1332.0|22.61|
|64000 B|14831.0|4195.0|3.54|

We're 30% down for 8 bytes 🥲 This was the worst case for me (rep startup + checks)
At about 24 bytes the gap starts to increase and it gets multiple times faster for 64Kib (a full memory page)

Yet I don't understand why the old loop copies backwards faster than forwards 🤔 . Maybe because increment/comparison is handled in one instruction?https://github.com/tetratelabs/wazero/blob/ff4a7ff4f9ee1c75bd52353e8f5a5fd000a78d03/internal/engine/compiler/impl_amd64.go#L3809-L3810

internal/engine/compiler/impl_amd64.go
```diff
@@ -3598,21 +3598,42 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
-// TODO: the compiled code in this function should be reused and compile at once as
-// the code is independent of any module.
-func (c *amd64Compiler) compileCopyImpl(isTable bool, srcTableIndex, dstTableIndex uint32) error {
-	outOfBoundsErrorStatus := nativeCallStatusCodeMemoryOutOfBounds
-	if isTable {
-		outOfBoundsErrorStatus = nativeCallStatusCodeInvalidTableAccess
+	// prepare registers for REP MOVSQ: copy from rsi to rdi, rcx times
+	c.assembler.CompileRegisterToRegister(amd64.XCHGQ, destinationOffset, amd64.RegDI)
+	c.assembler.CompileRegisterToRegister(amd64.XCHGQ, sourceOffset, amd64.RegSI)
+	c.assembler.CompileRegisterToRegister(amd64.XCHGQ, copySize, amd64.RegCX)
```
dranikpg, line 3609: It was silly for me to assume this will work. If we have DI/SI/CX among our used registers this might cause a double `XCHG` where we get wrong values.
We might either push them on the stack (slow) or invent something clever. I bet any case can be resolved with 3 XCHGS or less 🤔

mathetake:
been busy for the next few days, but will do the second pass for sure by then

internal/engine/compiler/impl_amd64.go
```diff
@@ -3632,150 +3660,95 @@ func (c *amd64Compiler) compileCopyImpl(isTable bool, srcTableIndex, dstTableInd
...
-			c.assembler.CompileMemoryToRegister(amd64.MOVQ, amd64ReservedRegisterForCallEngine, callEngineModuleContextTablesElement0AddressOffset, tmp)
-			c.assembler.CompileMemoryToRegister(amd64.MOVQ, tmp, int64(srcTableIndex*8), tmp)
-			c.assembler.CompileMemoryToRegister(amd64.ADDQ, tmp, tableInstanceTableOffset, sourceOffset.register)
+	// if source + size < dest, we can copy forwards
+	c.assembler.CompileRegisterToRegister(amd64.MOVQ, destinationOffset.register, tmp)
+	c.assembler.CompileRegisterToRegister(amd64.SUBQ, copySize.register, tmp)
+	c.assembler.CompileRegisterToRegister(amd64.CMPQ, sourceOffset.register, tmp)
+	sourceBoundLowerThanDestJump := c.assembler.CompileJump(amd64.JLS)
+
+	assembleCopySection := func(backwards bool) {
```
mathetake, line 3698: As i mentioned before, please do not define the closures unless _really_ necessary

internal/engine/compiler/impl_amd64.go
```diff
@@ -3598,21 +3598,49 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
+	c.compileSwapRegisters(copySize.register, amd64.RegCX)
+
+	c.compileRestorePreventedCrossing(rvalues, swaps)
+
+	c.assembler.SetJumpTargetOnNext(emptyEightGroupsJump)
+	c.assembler.CompileStandAlone(amd64.NOP)
+}
+
+// compileMemoryCopy implements compiler.compileMemoryCopy for the amd64 architecture.
+func (c *amd64Compiler) compileMemoryCopy() error {
```
mathetake, line 3643: Could you add the detailed doc about the logic in this function so that others can understand what kind of native code will be generated here without looking at the function body? e.g. Pseudo assembly etc.

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5074,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
 	}
 	return
 }
+
+// compileSwapRegisters swaps two registers if they're not equal
+func (c *amd64Compiler) compileSwapRegisters(reg1, reg2 asm.Register) {
```
mathetake, line 5079:
```suggestion
func (c *amd64Compiler) compileMaybeSwapRegisters(reg1, reg2 asm.Register) {
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5074,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+	}
+}
+
+// compilePreventCrossedTargetRegisters ensures that neither runtimeValueLocations target register
+// is present among the other runtimeValueLocations. To prevent this, it swaps misplaced registers.
+//
+// This function makes it possible to safely exchange one set of registers with another, where a register might be in both sets.
+// Each register will correspond either to another register not present in its set or to itself
+func (c *amd64Compiler) compilePreventCrossedTargetRegisters(locs []*runtimeValueLocation, targets []asm.Register) []int {
+	swaps := make([]int, 0, 3)
```
mathetake, line 5091: comment on what 3 means

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5074,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+	for i := range locs {
+		targetLocation := -1
+		for j := range locs {
+			if locs[j].register == targets[i] {
+				targetLocation = j
+				break
+			}
+		}
+		if targetLocation != -1 && targetLocation != i {
```
mathetake, line 5100: what is -1?

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5074,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+		c.assembler.CompileRegisterToRegister(amd64.XCHGQ, reg1, reg2)
+	}
+}
+
+// compilePreventCrossedTargetRegisters ensures that neither runtimeValueLocations target register
+// is present among the other runtimeValueLocations. To prevent this, it swaps misplaced registers.
+//
+// This function makes it possible to safely exchange one set of registers with another, where a register might be in both sets.
+// Each register will correspond either to another register not present in its set or to itself
+func (c *amd64Compiler) compilePreventCrossedTargetRegisters(locs []*runtimeValueLocation, targets []asm.Register) []int {
```
mathetake, line 5090: want to write unit tests?

internal/engine/compiler/impl_amd64.go
```diff
@@ -3598,21 +3598,49 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
-//
-// TODO: the compiled code in this function should be reused and compile at once as
-// the code is independent of any module.
-func (c *amd64Compiler) compileCopyImpl(isTable bool, srcTableIndex, dstTableIndex uint32) error {
-	outOfBoundsErrorStatus := nativeCallStatusCodeMemoryOutOfBounds
-	if isTable {
-		outOfBoundsErrorStatus = nativeCallStatusCodeInvalidTableAccess
+	rvalues := []*runtimeValueLocation{destinationOffset, sourceOffset, copySize}
+
+	// prepare registers for swaps. There will never be more than 3 XCHGs in total
```
mathetake, line 3608: Please make sure that _all_ of your line comments start with Captial letter and end with period. See other comments. (though trailing comments are not the same)

```suggestion
	// Prepare registers for swaps. There will never be more than 3 XCHGs in total.
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -3598,21 +3598,49 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
 	)
 }
 
-// compileMemoryCopy implements compiler.compileMemoryCopy for the amd64 architecture.
-func (c *amd64Compiler) compileMemoryCopy() error {
-	return c.compileCopyImpl(false, 0, 0)
-}
+func (c *amd64Compiler) compileCopyLoopImpl(destinationOffset, sourceOffset, copySize *runtimeValueLocation, backwards bool, bwOffset uint8) {
```
mathetake, line 3601: Missing go doc

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5074,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+			c.compileSwapRegisters(locs[i].register, locs[targetLocation].register)
+			locs[i].register, locs[targetLocation].register = locs[targetLocation].register, locs[i].register
+			swaps = append(swaps, i, targetLocation)
+		}
+	}
+	return swaps
+}
+
+// compileRestorePreventedCrossing can be applied after compilePreventCrossedTargetRegisters to restore registers
+func (c *amd64Compiler) compileRestorePreventedCrossing(locs []*runtimeValueLocation, swaps []int) {
```
mathetake, line 5110: needs unit test

internal/engine/compiler/compiler_post1_0_test.go
```diff
@@ -238,6 +244,35 @@ func TestCompiler_compileMemoryCopy(t *testing.T) {
...
+				} else {
+					compiler.compileConstI32(&wazeroir.OperationConstI32{Value: 777}) // dest
+					compiler.compileConstI32(&wazeroir.OperationConstI32{Value: 1})   // src
+				}
+				compiler.compileConstI32(&wazeroir.OperationConstI32{Value: size})
+				compiler.compileMemoryCopy()
+				compiler.(compilerImpl).compileReturnFunction()
+				code, _, _ := compiler.compile()
+
+				env.execBench(code, b)
```
mathetake, line 270: one common mistake in benchmarks is to foger to check the result, and just run code. Could you check the memory and the resulting state is actually expected one?

mathetake (changes requested):
Thanks, good progress!

I would appreciate it if you cloud write the assembly in the PR description so that we can see the before and after of machine code.

internal/engine/compiler/impl_amd64.go
dranikpg: Already did 😎 (in `impl_amd64_test` for prevent/restore)

internal/engine/compiler/impl_amd64.go
dranikpg: Not found. Its the intiial value. I'll document it if its not clear

dranikpg:
Fixed benchmark, fixed comments

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -391,6 +391,42 @@ func TestAmd64Compiler_readInstructionAddress(t *testing.T) {
...
+	tests := []struct {
+		desired, expected []asm.Register
+		swaps             []int
+	}{
+		{desired: []asm.Register{3, 2, 1}, expected: []asm.Register{3, 2, 1}, swaps: []int{0, 2}},
+		{desired: []asm.Register{3, 1, 2}, expected: []asm.Register{3, 1, 2}, swaps: []int{0, 2, 1, 2}},
+		{desired: []asm.Register{5, 1, 4}, expected: []asm.Register{2, 1, 3}, swaps: []int{1, 0}},
+		{desired: []asm.Register{4, 5, 6}, expected: []asm.Register{1, 2, 3}, swaps: []int{}},
+		{desired: []asm.Register{1, 2, 3}, expected: []asm.Register{1, 2, 3}, swaps: []int{}},
+		{desired: []asm.Register{3, 5, 1}, expected: []asm.Register{3, 2, 1}, swaps: []int{0, 2}},
```
mathetake, line 417: could you use `amd64.Reg***`, rather than 3, 2, 1 or whatever?

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -391,6 +391,42 @@ func TestAmd64Compiler_readInstructionAddress(t *testing.T) {
 	})
 }
 
+func TestAmd64Compiler_preventCrossedTargetdRegisters(t *testing.T) {
+	env := newCompilerEnvironment()
+	compiler := env.requireNewCompiler(t, newAmd64Compiler, nil).(*amd64Compiler)
+
+	rv := []*runtimeValueLocation{{register: 1}, {register: 2}, {register: 3}}
```
mathetake, line 398: ditto

internal/engine/compiler/compiler_test.go
```diff
@@ -242,6 +244,22 @@ func (j *compilerEnv) exec(codeSegment []byte) {
...
+
+	b.StartTimer()
+	for i := 0; i < b.N; i++ {
+		j.ce.callFrameStack[j.ce.globalContext.callFrameStackPointer] = callFrame{function: f}
+		j.ce.globalContext.callFrameStackPointer++
+		nativecall(
+			uintptr(unsafe.Pointer(&codeSegment[0])),
+			uintptr(unsafe.Pointer(j.ce)),
+			uintptr(unsafe.Pointer(j.moduleInstance)),
+		)
```
mathetake, line 258: we should check the result in each loop, rather than after the timer.

internal/engine/compiler/compiler_post1_0_test.go
```diff
@@ -238,6 +245,65 @@ func TestCompiler_compileMemoryCopy(t *testing.T) {
...
+
+				for i := 0; i < b.N; i++ {
+					copy(testMem[destOffset:destOffset+size], testMem[sourceOffset:sourceOffset+size])
+				}
+
+				if !bytes.Equal(mem, testMem) {
+					b.FailNow()
+				}
```
mathetake, line 301: plz move this to inside the bench iteration loop

internal/engine/compiler/compiler_test.go
dranikpg: What do we benchmark then 🤔 ? We'll have a lot of interference from go functions (copy, equal)

mathetake: no these are not interference at at all, but it depends how expensive they are. What we should really care is the case like this: the logic only peformes correctly at 1000th iteration, and the other cases, it just does the NOP in the loop. That would be wrong benchmarks result and that's the _really_ common mistake in benchmark.

internal/engine/compiler/impl_amd64.go
```diff
@@ -3598,21 +3598,95 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
+	if backwards {
+		// Reset direction.
+		c.assembler.CompileStandAlone(amd64.CLD)
+	}
+
+	// Restore registers.
+	c.compileMaybeSwapRegisters(destinationOffset.register, amd64.RegDI)
+	c.compileMaybeSwapRegisters(sourceOffset.register, amd64.RegSI)
+	c.compileMaybeSwapRegisters(copySize.register, amd64.RegCX)
+	c.compileRestorePreventedCrossing(rvalues, swaps)
```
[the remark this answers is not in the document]
dranikpg: I've seen it when searching for similar solutions. But the questions stays the same. What if one of (Dest, Src, Size) is on (DI, SI, CX)? If we just MOV registers like in the example you linked, we might overwrite ourselves if (Dest, Src, Size) is among (DI, SI, CX). For example, we move Dest on DI, but DI stored our Size. So we'll have to read it from the stack 
Isn't that a lot of memory operations? (Whereas xchgs are almost free). At least those were my thoughts when going for it 🤔 Or is that okay?

internal/engine/compiler/compiler_test.go
mathetake: the other way you can do is to write unit tests on the logic performed in the each loop iteration -- you can extract it as the separate function, and then write unit test for it.

internal/engine/compiler/impl_amd64.go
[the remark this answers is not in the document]
mathetake: will do another review pass tomorrow based on your comment!

dranikpg:
I moved the check into the bench loop. While I get the point of checks in benchmarks, those are quite expensive here compared to the operation.

This is not a real benchmark anymore in a sense that the numbers don't mean anything 😕. For _correctness_, we have to check the full memory page (what if it wrote more than requested). But this is too expensive for the smaller-sized memcpy operations. It can't be used to compare different implementations or see how changes affect performance.  Whats the point of the benchmark then 🤔 

We could check only destination sections, but this is a half-baked solution. We don't catch all possible bugs and we get distorted results.

> What we should really care is the case like this: the logic only peformes correctly at 1000th iteration, and the other cases, it just does the NOP in the loop.

Yeah, sure, but this is highly unlikely. We have unittest for that. If we benchmark a function, we should assume it already works correctly.

> The other way you can do is to write unit tests on the logic performed in the each loop iteration

Unit tests for benchmark code 😄

mathetake:
will get back next monday!

internal/engine/compiler/compiler_post1_0_test.go
```diff
@@ -238,6 +245,62 @@ func TestCompiler_compileMemoryCopy(t *testing.T) {
...
+				err = compiler.compileConstI32(&wazeroir.OperationConstI32{Value: size})
+				requireBenchmarkNoError(b, err)
+				err = compiler.compileMemoryCopy()
+				requireBenchmarkNoError(b, err)
+				err = compiler.(compilerImpl).compileReturnFunction()
+				requireBenchmarkNoError(b, err)
+				code, _, err := compiler.compile()
+				requireBenchmarkNoError(b, err)
+
+				env.execBench(b, code, func() {
```
mathetake, line 293: Ok you got me... let's delete the checkFunc, and I'm fine without checking the result after each iteration

internal/asm/amd64/consts.go
```diff
@@ -215,6 +217,10 @@ const (
 	PSRLD
 	// PSRLQ is the PSRLQ instruction. https://www.felixcloutier.com/x86/psrlw:psrld:psrlq
 	PSRLQ
+	// REPMOVSB is a combination of REP MOVSB instructions. https://www.felixcloutier.com/x86/movs:movsb:movsw:movsd:movsq https://www.felixcloutier.com/x86/rep:repe:repz:repne:repnz
```
mathetake, line 220:
```suggestion
	// REPMOVSB is the REP MOVSB instruction with 8-bit mode. https://www.felixcloutier.com/x86/rep:repe:repz:repne:repnz
```

internal/asm/amd64/consts.go
```diff
@@ -215,6 +217,10 @@ const (
...
+	REPMOVSB
+	// REPMOVSB is a combination of REP MOVSQ instructions. https://www.felixcloutier.com/x86/movs:movsb:movsw:movsd:movsq https://www.felixcloutier.com/x86/rep:repe:repz:repne:repnz
```
mathetake, line 222:
```suggestion
	// REPMOVSB is the REP MOVSQ instruction with 64-bit mode. https://www.felixcloutier.com/x86/rep:repe:repz:repne:repnz
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -3598,21 +3598,95 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
+	} else {
+		c.assembler.CompileNoneToRegister(amd64.INCQ, sourceOffset.register)
+		c.assembler.CompileNoneToRegister(amd64.INCQ, destinationOffset.register)
+	}
+
+	c.assembler.CompileNoneToRegister(amd64.DECQ, copySize.register)
+	c.assembler.CompileJump(amd64.JMP).AssignJumpTarget(beginLoop)
+
+	c.assembler.SetJumpTargetOnNext(breakLoop)
+	// copySize /= 8
```
mathetake, line 3679: Comment on why we divide 8 (for anyone can understand)

internal/engine/compiler/impl_amd64.go
```diff
@@ -3598,21 +3598,95 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
+		c.assembler.CompileRegisterToRegister(amd64.SUBQ, copySize.register, sourceOffset.register)
+		c.assembler.CompileRegisterToRegister(amd64.SUBQ, copySize.register, destinationOffset.register)
 	}
 
+	// destinationOffset += memory buffer's absolute address.
+	c.assembler.CompileRegisterToRegister(amd64.ADDQ, amd64ReservedRegisterForMemory, destinationOffset.register)
+	// sourceOffset += memory buffer's absolute address.
+	c.assembler.CompileRegisterToRegister(amd64.ADDQ, amd64ReservedRegisterForMemory, sourceOffset.register)
+
+	beginLoop := c.assembler.CompileStandAlone(amd64.NOP)
```
mathetake, line 3658: Comment what happens inside this loop like `Copy the remainder bytes divided by 8 so that we can copy the content in the 8 byte loop below`

mathetake:
ok exellent! Thanks @dranikpg for getting through this complex compiler code and put things together.

Once you resolved the conflict and some miner nits, than good to go

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5082,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+	}
+}
+
+// compilePreventCrossedTargetRegisters ensures that neither runtimeValueLocations target register
+// is present among the other runtimeValueLocations. To prevent this, it swaps misplaced registers.
+//
+// This function makes it possible to safely exchange one set of registers with another, where a register might be in both sets.
+// Each register will correspond either to another register not present in its set or to itself
+func (c *amd64Compiler) compilePreventCrossedTargetRegisters(locs []*runtimeValueLocation, targets []asm.Register) []int {
+	swaps := []int{}
```
mathetake, line 5099:
```suggestion
	var swaps []int{}
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5082,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+
+// compileMaybeSwapRegisters swaps two registers if they're not equal.
+func (c *amd64Compiler) compileMaybeSwapRegisters(reg1, reg2 asm.Register) {
+	if reg1 != reg2 {
+		c.assembler.CompileRegisterToRegister(amd64.XCHGQ, reg1, reg2)
+	}
+}
+
+// compilePreventCrossedTargetRegisters ensures that neither runtimeValueLocations target register
+// is present among the other runtimeValueLocations. To prevent this, it swaps misplaced registers.
```
mathetake, line 5094: document what is the return of this function

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5082,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+//
+// This function makes it possible to safely exchange one set of registers with another, where a register might be in both sets.
+// Each register will correspond either to another register not present in its set or to itself
+func (c *amd64Compiler) compilePreventCrossedTargetRegisters(locs []*runtimeValueLocation, targets []asm.Register) []int {
```
mathetake, line 5098: I would suggest returning the restore function rather than the mystic []int.

```suggestion
func (c *amd64Compiler) compilePreventCrossedTargetRegisters(locs []*runtimeValueLocation, targets []asm.Register) (restore func ()) {
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -3598,21 +3598,95 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
+// compileCopyLoopImpl implements a REP MOVSQ memory copy for the given range with support for both directions.
+func (c *amd64Compiler) compileCopyLoopImpl(destinationOffset, sourceOffset, copySize *runtimeValueLocation, backwards bool, bwOffset uint8) {
+	// skip if nothing to copy
+	c.assembler.CompileRegisterToConst(amd64.CMPQ, copySize.register, 0)
+	emptyEightGroupsJump := c.assembler.CompileJump(amd64.JEQ)
+
+	rvalues := []*runtimeValueLocation{destinationOffset, sourceOffset, copySize}
+
+	// Prepare registers for swaps. There will never be more than 3 XCHGs in total.
+	swaps := c.compilePreventCrossedTargetRegisters(rvalues, []asm.Register{amd64.RegDI, amd64.RegSI, amd64.RegCX})
```
mathetake, line 3610: As suggested, let's return the restore func, and then defer it so that we don't forget to invoke later on (as that requires the context later vs just defer here.

```suggestion
	restore := c.compilePreventCrossedTargetRegisters(rvalues, []asm.Register{amd64.RegDI, amd64.RegSI, amd64.RegCX})
	defer restore()
```

internal/engine/compiler/impl_amd64.go
mathetake, line 3636:
```suggestion
	restoreCrossing()
```

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -391,6 +391,42 @@ func TestAmd64Compiler_readInstructionAddress(t *testing.T) {
...
+			out[i] = rv[i].register
+		}
+		return out
+	}
+
+	tests := []struct {
+		desired, expected []asm.Register
+		swaps             []int
+	}{
+		{desired: []asm.Register{amd64.RegDX, amd64.RegCX, amd64.RegAX}, expected: []asm.Register{amd64.RegDX, amd64.RegCX, amd64.RegAX}, swaps: []int{0, 2}},
```
mathetake, line 412: Could you break the lines?

```suggestion
		{
			desired: []asm.Register{amd64.RegDX, amd64.RegCX, amd64.RegAX}, 
			expected: []asm.Register{amd64.RegDX, amd64.RegCX, amd64.RegAX}, 
			swaps: []int{0, 2},
		},
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5082,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
 }
+
+// compileMaybeSwapRegisters swaps two registers if they're not equal.
+func (c *amd64Compiler) compileMaybeSwapRegisters(reg1, reg2 asm.Register) {
+	if reg1 != reg2 {
+		c.assembler.CompileRegisterToRegister(amd64.XCHGQ, reg1, reg2)
+	}
+}
+
+// compilePreventCrossedTargetRegisters ensures that neither runtimeValueLocations target register
```
mathetake, line 5093: this sentence is hard to understand. Could you elaborate more on what `runtimeValueLocations target register` means and please add the example here.

Also,use exactly the same name as in function params `locs` and `targets` in the documentation. That's how we reduce the overhead to understand how the function performs without looking at the function body. Please follow the same rule for all of the your implementation.

internal/engine/compiler/impl_amd64.go
```diff
@@ -5013,3 +5082,43 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+		for j := range locs {
+			if locs[j].register == targets[i] {
+				targetLocation = j
+				break
+			}
+		}
+		if targetLocation != -1 && targetLocation != i {
+			c.compileMaybeSwapRegisters(locs[i].register, locs[targetLocation].register)
+			locs[i].register, locs[targetLocation].register = locs[targetLocation].register, locs[i].register
+			swaps = append(swaps, i, targetLocation)
```
mathetake, line 5111: I see  from this, `swaps []int` is more of `[]struct{srcIndex, dstIndex}`? Please use the named struct, rather than flatted []int as it's a bit complex to understand for others.

internal/engine/compiler/compiler_post1_0_test.go
```diff
@@ -238,6 +245,62 @@ func TestCompiler_compileMemoryCopy(t *testing.T) {
...
 }
 
+func requireBenchmarkNoError(b *testing.B, err error) {
+	if err != nil {
+		b.Fatal(err)
+	}
+}
+
+func BenchmarkCompiler_compileMemoryCopy(b *testing.B) {
+	sizes := []uint32{8, 64, 128, 1000, 16000, 64000}
```
mathetake, line 255: adds the case where size cannot be divided by 8?

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -391,6 +391,42 @@ func TestAmd64Compiler_readInstructionAddress(t *testing.T) {
 	})
 }
 
+func TestAmd64Compiler_preventCrossedTargetdRegisters(t *testing.T) {
+	env := newCompilerEnvironment()
+	compiler := env.requireNewCompiler(t, newAmd64Compiler, nil).(*amd64Compiler)
+
+	rv := []*runtimeValueLocation{{register: amd64.RegAX}, {register: amd64.RegCX}, {register: amd64.RegDX}}
+
+	currentRegisters := func() []asm.Register {
```
mathetake, line 400: I don't think we need this as I see this just adds the complexity.

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -391,6 +391,42 @@ func TestAmd64Compiler_readInstructionAddress(t *testing.T) {
...
+		{desired: []asm.Register{amd64.RegDX, amd64.RegR9, amd64.RegAX}, expected: []asm.Register{amd64.RegDX, amd64.RegCX, amd64.RegAX}, swaps: []int{0, 2}},
+	}
+
+	for _, tt := range tests {
+		swaps := compiler.compilePreventCrossedTargetRegisters(rv, tt.desired)
+		require.Equal(t, tt.swaps, swaps)
+		require.Equal(t, tt.expected, currentRegisters())
+		compiler.compileRestorePreventedCrossing(rv, swaps)
+		require.Equal(t, []asm.Register{amd64.RegAX, amd64.RegCX, amd64.RegDX}, currentRegisters())
+	}
```
mathetake, line 426: Comment on why this is the expected state.

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -391,6 +391,42 @@ func TestAmd64Compiler_readInstructionAddress(t *testing.T) {
...
+		{desired: []asm.Register{amd64.RegAX, amd64.RegCX, amd64.RegDX}, expected: []asm.Register{amd64.RegAX, amd64.RegCX, amd64.RegDX}, swaps: []int{}},
+		{desired: []asm.Register{amd64.RegDX, amd64.RegR9, amd64.RegAX}, expected: []asm.Register{amd64.RegDX, amd64.RegCX, amd64.RegAX}, swaps: []int{0, 2}},
+	}
+
+	for _, tt := range tests {
+		swaps := compiler.compilePreventCrossedTargetRegisters(rv, tt.desired)
+		require.Equal(t, tt.swaps, swaps)
+		require.Equal(t, tt.expected, currentRegisters())
+		compiler.compileRestorePreventedCrossing(rv, swaps)
+		require.Equal(t, []asm.Register{amd64.RegAX, amd64.RegCX, amd64.RegDX}, currentRegisters())
```
mathetake, line 425: Comment on why this is the expected state.

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -391,6 +391,42 @@ func TestAmd64Compiler_readInstructionAddress(t *testing.T) {
 	})
 }
 
+func TestAmd64Compiler_preventCrossedTargetdRegisters(t *testing.T) {
+	env := newCompilerEnvironment()
+	compiler := env.requireNewCompiler(t, newAmd64Compiler, nil).(*amd64Compiler)
+
+	rv := []*runtimeValueLocation{{register: amd64.RegAX}, {register: amd64.RegCX}, {register: amd64.RegDX}}
```
mathetake, line 398: Shoulnd't this be defined inside each case? what if the each case has side effects on this and somehow passes the assertions in a wrong way?

internal/asm/amd64/consts.go
```diff
@@ -299,6 +307,8 @@ const (
 	XORPS
 	// XORQ is the XOR instruction in 64-bit mode. https://www.felixcloutier.com/x86/xor
 	XORQ
+	// XCHGQ is the XCHG instruction. https://www.felixcloutier.com/x86/xchg
```
mathetake, line 310:
```suggestion
	// XCHGQ is the XCHG instruction with 64-bit mode. https://www.felixcloutier.com/x86/xchg
```

internal/asm/amd64/impl_2_test.go
```diff
@@ -1161,7 +1161,9 @@ func TestAssemblerImpl_EncodeRegisterToRegister(t *testing.T) {
 		{name: "XORQ/src=AX/dst=R8/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegAX, dstReg: RegR8, arg: 0x0}, exp: []byte{0x49, 0x31, 0xc0}},
 		{name: "XORQ/src=R8/dst=AX/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegR8, dstReg: RegAX, arg: 0x0}, exp: []byte{0x4c, 0x31, 0xc0}},
 		{name: "XORQ/src=R8/dst=R8/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegR8, dstReg: RegR8, arg: 0x0}, exp: []byte{0x4d, 0x31, 0xc0}},
-		{name: "PXOR/src=X0/dst=X0/arg=0", n: &nodeImpl{instruction: PXOR, srcReg: RegX0, dstReg: RegX0, arg: 0x0}, exp: []byte{0x66, 0xf, 0xef, 0xc0}},
+		{name: "XCHG/src=R8/dst=AX/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegR8, dstReg: RegAX, arg: 0x0}, exp: []byte{0x4c, 0x87, 0xc0}},
+		{name: "XCHG/src=AX/dst=R8/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegAX, dstReg: RegR8, arg: 0x0}, exp: []byte{0x49, 0x87, 0xc0}},
+		{name: "XCHG/src=R8/dst=R9/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegR9, dstReg: RegR8, arg: 0x0}, exp: []byte{0x4d, 0x87, 0xc8}},
```
mathetake, line 1166:
```suggestion
		{name: "XCHGQ/src=R8/dst=R9/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegR9, dstReg: RegR8, arg: 0x0}, exp: []byte{0x4d, 0x87, 0xc8}},
```

internal/asm/amd64/impl_2_test.go
```diff
@@ -1161,7 +1161,9 @@ func TestAssemblerImpl_EncodeRegisterToRegister(t *testing.T) {
 		{name: "XORQ/src=AX/dst=R8/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegAX, dstReg: RegR8, arg: 0x0}, exp: []byte{0x49, 0x31, 0xc0}},
 		{name: "XORQ/src=R8/dst=AX/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegR8, dstReg: RegAX, arg: 0x0}, exp: []byte{0x4c, 0x31, 0xc0}},
 		{name: "XORQ/src=R8/dst=R8/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegR8, dstReg: RegR8, arg: 0x0}, exp: []byte{0x4d, 0x31, 0xc0}},
-		{name: "PXOR/src=X0/dst=X0/arg=0", n: &nodeImpl{instruction: PXOR, srcReg: RegX0, dstReg: RegX0, arg: 0x0}, exp: []byte{0x66, 0xf, 0xef, 0xc0}},
+		{name: "XCHG/src=R8/dst=AX/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegR8, dstReg: RegAX, arg: 0x0}, exp: []byte{0x4c, 0x87, 0xc0}},
+		{name: "XCHG/src=AX/dst=R8/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegAX, dstReg: RegR8, arg: 0x0}, exp: []byte{0x49, 0x87, 0xc0}},
```
mathetake, line 1165:
```suggestion
		{name: "XCHGQ/src=AX/dst=R8/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegAX, dstReg: RegR8, arg: 0x0}, exp: []byte{0x49, 0x87, 0xc0}},
```

internal/asm/amd64/impl_2_test.go
```diff
@@ -1161,7 +1161,9 @@ func TestAssemblerImpl_EncodeRegisterToRegister(t *testing.T) {
 		{name: "XORQ/src=AX/dst=R8/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegAX, dstReg: RegR8, arg: 0x0}, exp: []byte{0x49, 0x31, 0xc0}},
 		{name: "XORQ/src=R8/dst=AX/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegR8, dstReg: RegAX, arg: 0x0}, exp: []byte{0x4c, 0x31, 0xc0}},
 		{name: "XORQ/src=R8/dst=R8/arg=0", n: &nodeImpl{instruction: XORQ, srcReg: RegR8, dstReg: RegR8, arg: 0x0}, exp: []byte{0x4d, 0x31, 0xc0}},
-		{name: "PXOR/src=X0/dst=X0/arg=0", n: &nodeImpl{instruction: PXOR, srcReg: RegX0, dstReg: RegX0, arg: 0x0}, exp: []byte{0x66, 0xf, 0xef, 0xc0}},
+		{name: "XCHG/src=R8/dst=AX/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegR8, dstReg: RegAX, arg: 0x0}, exp: []byte{0x4c, 0x87, 0xc0}},
```
mathetake, line 1164:
```suggestion
		{name: "XCHGQ/src=R8/dst=AX/arg=0", n: &nodeImpl{instruction: XCHGQ, srcReg: RegR8, dstReg: RegAX, arg: 0x0}, exp: []byte{0x4c, 0x87, 0xc0}},
```

internal/engine/compiler/impl_amd64_test.go
dranikpg: I'll move it to a separate function then, because I need to collect registers three times

internal/engine/compiler/impl_amd64.go
```diff
@@ -5030,3 +5100,45 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+// corresponding register with the same index from targets is occupied by some other runtimeValueLocation from locs. It returns a
+// closure to restore the original register placement.
+//
+// This function makes it possible to safely exchange one set of registers with another, where a register might be in both sets.
+// Each register will correspond either to itself or another register not present in its own set.
+//
+// For example, if we have locs = [AX, BX, CX], targets = [BX, SI, AX], then it'll do two swaps
+// to make locs = [BX, CX, AX]
+func (c *amd64Compiler) compilePreventCrossedTargetRegisters(locs []*runtimeValueLocation, targets []asm.Register) (restore func()) {
+	type Swap struct{ srcIndex, dstIndex int }
```
mathetake, line 5121:
```suggestion
	type swap struct{ srcIndex, dstIndex int }
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -5030,3 +5100,45 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+			}
+		}
+		if targetLocation != -1 && targetLocation != i {
+			c.compileMaybeSwapRegisters(locs[i].register, locs[targetLocation].register)
+			locs[i].register, locs[targetLocation].register = locs[targetLocation].register, locs[i].register
+			swaps = append(swaps, Swap{i, targetLocation})
+		}
+	}
+	return func() {
+		for i := len(swaps) - 1; i >= 0; i -= 1 {
```
mathetake, line 5138: why do we need to do reverse-order? trailing comment would be helpful for others.

internal/engine/compiler/impl_amd64.go
```diff
@@ -3642,150 +3717,53 @@ func (c *amd64Compiler) compileCopyImpl(isTable bool, srcTableIndex, dstTableInd
...
-		// For tables, we move 8 bytes at once.
-		memToReg = amd64.MOVQ
-		regToMem = memToReg
-		scale = 8
-	} else {
-		memToReg = amd64.MOVBQZX
-		regToMem = amd64.MOVB
-		scale = 1
-	}
+	// if source + size < dest, we can copy forwards
```
mathetake, line 3749:
```suggestion
	// If source + size < dest, we can copy forwards
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -3608,21 +3608,96 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
+
+	// compileCopyLoopImpl counts in groups of 8 bytes, so we have to divide the copySize by 8.
+	c.assembler.CompileConstToRegister(amd64.SHRQ, 3, copySize.register)
+
+	c.compileCopyLoopImpl(destinationOffset, sourceOffset, copySize, backwards, 7)
+}
+
+// compileMemoryCopy implements compiler.compileMemoryCopy for the amd64 architecture.
+//
+// compileMemoryCopy uses efficient `REP MOVSQ` instructions to copy in quadword (8 bytes) batches. The remaining bytes
```
mathetake, line 3698:
```suggestion
// This uses efficient `REP MOVSQ` instructions to copy in quadword (8 bytes) batches. The remaining bytes
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -3901,9 +3879,101 @@ func (c *amd64Compiler) compileTableInit(o *wazeroir.OperationTableInit) error {
...
+
+	// Skip zero size.
+	c.assembler.CompileRegisterToConst(amd64.CMPQ, copySize.register, 0)
+	skipJump := c.assembler.CompileJump(amd64.JEQ)
+
+	// If dest < source, we can copy forwards
+	c.assembler.CompileRegisterToRegister(amd64.CMPQ, destinationOffset.register, sourceOffset.register)
+	destLowerThanSourceJump := c.assembler.CompileJump(amd64.JLS)
+
+	// if source + size < dest, we can copy forwards
```
mathetake, line 3959:
```suggestion
	// If source + size < dest, we can copy forwards.
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -3901,9 +3879,101 @@ func (c *amd64Compiler) compileTableInit(o *wazeroir.OperationTableInit) error {
...
+	c.assembler.CompileMemoryToRegister(amd64.CMPQ, tmp, tableInstanceTableLenOffset, destinationOffset.register)
+	destinationBoundOKJump := c.assembler.CompileJump(amd64.JCC)
+	c.compileExitFromNativeCode(nativeCallStatusCodeInvalidTableAccess)
+	c.assembler.SetJumpTargetOnNext(destinationBoundOKJump)
+
+	// Skip zero size.
+	c.assembler.CompileRegisterToConst(amd64.CMPQ, copySize.register, 0)
+	skipJump := c.assembler.CompileJump(amd64.JEQ)
+
+	// If dest < source, we can copy forwards
```
mathetake, line 3955:
```suggestion
	// If dest < source, we can copy forwards.
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -3901,9 +3879,101 @@ func (c *amd64Compiler) compileTableInit(o *wazeroir.OperationTableInit) error {
...
+	c.assembler.CompileMemoryToRegister(amd64.MOVQ, tmp, int64(o.SrcTableIndex*8), tmp)
+	c.assembler.CompileMemoryToRegister(amd64.ADDQ, tmp, tableInstanceTableOffset, sourceOffset.register)
+
+	c.compileCopyLoopImpl(destinationOffset, sourceOffset, copySize, backwards, 8)
+}
+
 // compileTableCopy implements compiler.compileTableCopy for the amd64 architecture.
+//
+// It uses efficient `REP MOVSB` instructions for optimized copying. It uses backward copying for
+// overlapped segments
```
mathetake, line 3908:
```suggestion
// overlapped segments.
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -5030,3 +5100,45 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+
+// compilePreventCrossedTargetRegisters swaps registers in such a way, that for neither runtimeValueLocation from locs its
+// corresponding register with the same index from targets is occupied by some other runtimeValueLocation from locs. It returns a
+// closure to restore the original register placement.
+//
+// This function makes it possible to safely exchange one set of registers with another, where a register might be in both sets.
+// Each register will correspond either to itself or another register not present in its own set.
+//
+// For example, if we have locs = [AX, BX, CX], targets = [BX, SI, AX], then it'll do two swaps
+// to make locs = [BX, CX, AX]
```
mathetake, line 5119:
```suggestion
// to make locs = [BX, CX, AX].
```

mathetake (approved):
Cool, awesome work @dranikpg and thank you for dutifully getting though my reviews. 

I Ieft some editorial nits, so could you apply them? Then we are good to merge!

internal/engine/compiler/impl_amd64_test.go
mathetake: cool

internal/engine/compiler/impl_amd64.go
```diff
@@ -3608,21 +3608,96 @@ func (c *amd64Compiler) compileLoadDataInstanceAddress(dataIndex uint32, dst asm
...
+		c.assembler.CompileRegisterToRegister(amd64.SUBQ, copySize.register, sourceOffset.register)
+		c.assembler.CompileRegisterToRegister(amd64.SUBQ, copySize.register, destinationOffset.register)
 	}
 
+	// destinationOffset += memory buffer's absolute address.
+	c.assembler.CompileRegisterToRegister(amd64.ADDQ, amd64ReservedRegisterForMemory, destinationOffset.register)
+	// sourceOffset += memory buffer's absolute address.
+	c.assembler.CompileRegisterToRegister(amd64.ADDQ, amd64ReservedRegisterForMemory, sourceOffset.register)
+
+	// Copy copySize % 8 bytes in loop to allow copying in 8 byte groups.
```
mathetake, line 3668:
```suggestion
	// Copy copySize % 8 bytes in loop to allow copying in 8 byte groups afterwards.
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -5030,3 +5100,45 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+
+// compileMaybeSwapRegisters swaps two registers if they're not equal.
+func (c *amd64Compiler) compileMaybeSwapRegisters(reg1, reg2 asm.Register) {
+	if reg1 != reg2 {
+		c.assembler.CompileRegisterToRegister(amd64.XCHGQ, reg1, reg2)
+	}
+}
+
+// compilePreventCrossedTargetRegisters swaps registers in such a way, that for neither runtimeValueLocation from locs its
+// corresponding register with the same index from targets is occupied by some other runtimeValueLocation from locs. It returns a
```
mathetake, line 5112:
```suggestion
// compilePreventCrossedTargetRegisters swaps registers in such a way, that none of runtimeValueLocation from `locs` and its
// corresponding register with the same index in `targets` are occupied by some other runtimeValueLocation from `locs`. It returns a
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -5030,3 +5100,45 @@ func (c *amd64Compiler) compileEnsureOnRegister(loc *runtimeValueLocation) (err
...
+// closure to restore the original register placement.
+//
+// This function makes it possible to safely exchange one set of registers with another, where a register might be in both sets.
```
mathetake, line 5115: good comment

mathetake:
oh well I will request @codefromthecrypt's review as well before merge for the editorial stuff :D

codefromthecrypt:
I'm excited about this going in as bulk-memory is now in use (ex in TinyGo)!

However I want to avoid comment storm, so I'll do a review once Vladislav is finished with the above comments. Mine will be mostly editorial in nature, unless I somehow catch something Takeshi hasn't.

@dranikpg ping me when you are ready!

dranikpg:
@codefromthecrypt ready

internal/engine/compiler/compiler_post1_0_test.go
```diff
@@ -238,6 +245,65 @@ func TestCompiler_compileMemoryCopy(t *testing.T) {
 	}
 }
 
+func requireBenchmarkNoError(b *testing.B, err error) {
```
codefromthecrypt, line 248: renaming will be a little less distracting as it looks more like the assertion and still won't name conflict. please move this to the bottom of the file.

```suggestion
func requireNoError(b *testing.B, err error) {
```

internal/engine/compiler/compiler_post1_0_test.go
```diff
@@ -238,6 +245,65 @@ func TestCompiler_compileMemoryCopy(t *testing.T) {
...
+				err = compiler.compileConstI32(&wazeroir.OperationConstI32{Value: size})
+				requireBenchmarkNoError(b, err)
+				err = compiler.compileMemoryCopy()
+				requireBenchmarkNoError(b, err)
+				err = compiler.(compilerImpl).compileReturnFunction()
+				requireBenchmarkNoError(b, err)
+				code, _, err := compiler.compile()
+				requireBenchmarkNoError(b, err)
+
+				env.execBench(b, code)
```
codefromthecrypt, line 293: I would revert extraction of this to a separate file as it is only used once. This being here forces navigation and splits any follow-up maintenance across multiple files despite being only for one benchmark.

internal/engine/compiler/compiler_post1_0_test.go
```diff
@@ -238,6 +245,65 @@ func TestCompiler_compileMemoryCopy(t *testing.T) {
...
 }
 
+func requireBenchmarkNoError(b *testing.B, err error) {
+	if err != nil {
+		b.Fatal(err)
+	}
+}
+
+func BenchmarkCompiler_compileMemoryCopy(b *testing.B) {
+	sizes := []uint32{5, 8, 17, 64, 128, 1000, 16000, 64000}
```
codefromthecrypt, line 255: please reduce the dimensions unless you are noticing something very surprising between the edges.

codefromthecrypt (commented):
Thanks for the help. I enjoyed reading through the actual code very much.

The only thing I'd like to see changed is the benchmark setup as maintenance prior to merge is better than post merge. Mainly, we tend to separate files for benchmarks, so make a _benchmark_test file. Later, one of us might find time to restructure this code so that the setup is unit tested, which will simplify how it looks when executed. This will be easier when not split across multiple files.

internal/engine/compiler/compiler_post1_0_test.go
dranikpg: Separate file and not function? So I should just move execBench to the new `compiler_bench_test.go`?

codefromthecrypt: yeah please separate it into another file. we do that almost always here. the filename sounds good!

codefromthecrypt (approved):
:shipit:

mathetake:
@dranikpg Thank you for your contribution! 🎖️

dranikpg:
Great! I'll move to part 2: memset on x86 in the near future. Sorry for small issues all over, I'll try to be more attentive next time 😅

PR #701, opened 2022-07-15
interpreter: signed-extend to 32-bit in SignExtend32

mathetake:


codefromthecrypt (approved):
love it

PR #703, opened 2022-07-15
validation: not modify loop block type on unreachable br_table

mathetake:


codefromthecrypt (approved):
I'm starting to wonder how useful something is if only passes spectests!

PR #704, opened 2022-07-15
interpreter: fixes i32x4/i16x8 bit mask

mathetake:


codefromthecrypt (approved):
wow these are so subtle

PR #705, opened 2022-07-15
moremath: align NaN handling with amd64/arm64 compilers

mathetake:
This makes moremath library align with the arm64/amd64 and the reference 
interpreter in terms of NaN handling.


mathetake:
found the fix, but no idea why it fixes... will figure out next week

mathetake:
I think I will repurpose this to aline our moremath with the corresponding one in the reference interpreter, especially in terms of NaN value handling.

internal/moremath/moremath.go
```diff
@@ -186,90 +141,131 @@ func WasmCompatNearestF32(f float32) float32 {
...
+
+func f32IsNaN(v float32) bool {
+	return v != v // this is how NaN is defined.
+}
+
+func f64IsNaN(v float64) bool {
+	return v != v // this is how NaN is defined.
+}
+
+// returnF32UniOp returns the result of 32-bit unary operation. This accepts `original` which is the operand and `result` which is its result,
```
codefromthecrypt, line 221: nit I've been recently changing docs to 80 column as it is easier to read them without scrolling if you have big fonts.

codefromthecrypt (approved):
cool this was less nasty than I would have expected :D

PR #707, opened 2022-07-19
Supports mix of wasm and go funcs in the same module

codefromthecrypt:
This removes the constraint of a module being exclusively wasm or host
functions. Later pull requests can optimize special imports to be
implemented in wasm, particularly useful for disabled logging callbacks.

### DIFF
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -367,14 +367,6 @@ func (s *Store) instantiate(
 	}
 	globals, memory := module.buildGlobals(importedGlobals), module.buildMemory()
 
-	// If there are no module-defined functions, assume this is a host module.
-	var funcSection SectionID
-	if module.HostFunctionSection == nil {
-		funcSection = SectionIDFunction
-	} else {
-		funcSection = SectionIDHostFunction
-	}
-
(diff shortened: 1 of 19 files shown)


internal/wasm/store.go
```diff
@@ -420,7 +412,7 @@ func (s *Store) instantiate(
 		if exitErr, ok := err.(*sys.ExitError); ok { // Don't wrap an exit error!
 			return nil, exitErr
 		} else if err != nil {
-			return nil, fmt.Errorf("start %s failed: %w", module.funcDesc(funcSection, funcIdx), err)
+			return nil, fmt.Errorf("start %s failed: %w", module.funcDesc(SectionIDFunction, funcIdx), err)
```
mathetake, line 415: curious if this line is covered in store_test.go

mathetake (approved):

internal/wasm/store.go
codefromthecrypt: yeah, what happened was it didn't test the error message difference before for SectionIDHostFunction. that's why changing this worked.

codefromthecrypt:
Thanks for the look @mathetake!

PR #709, opened 2022-07-21
interpreter: fixes V128FloatPromote to use lower 64-bits.

mathetake:


internal/engine/interpreter/interpreter.go
```diff
@@ -3831,9 +3831,9 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, callCtx *wasm.CallCont
 			ce.pushValue(retHi)
 			frame.pc++
 		case wazeroir.OperationKindV128FloatPromote:
-			hi, lo := ce.popValue(), ce.popValue()
+			_, lo := ce.popValue(), ce.popValue()
```
codefromthecrypt, line 3834: maybe rename `lo` to `toPromote` because when you see lo without hi it makes you wonder why hi is dropped. ack you can read the docs on the instruction, but might be better to just rename the var.

codefromthecrypt (approved):

internal/engine/interpreter/interpreter.go
mathetake: sg

PR #710, opened 2022-07-21
Refactors host function tests to stub with wasm

codefromthecrypt:
This refactors host functions with no-op or constant returns to be
implemented with wasm instead of the host function bridge. This allows
better performance.

This also breaks up and makes WASI tests consistent, in a way that shows
parameter name drifts easier.


mathetake:
> +3,342 −3,839 
LOL

wasi_snapshot_preview1/wasi_test.go
```diff
@@ -1909,93 +32,46 @@ func maskMemory(t *testing.T, ctx context.Context, mod api.Module, size int) {
...
-  (memory 1 1)  ;; just an arbitrary size big enough for tests
-  (export "memory" (memory 0))
-  (export "%[1]s" (func $wasi.%[1]s))
-)`, wasiFunction, wasiImport))
-	require.NoError(t, err)
+	r := wazero.NewRuntimeWithConfig(wazero.NewRuntimeConfigInterpreter())
 
-	compiled, err := r.CompileModule(ctx, binary, wazero.NewCompileConfig())
+	compiled, err := (&builder{r}).moduleBuilder().
+		ExportMemoryWithMax("memory", 1, 1).
```
codefromthecrypt, line 44: one interesting thing is that I'm exporting memory in tests that use the wasi module. This is much simpler than the import/export dance.

assemblyscript/assemblyscript.go
```diff
@@ -178,34 +180,41 @@ func (a *assemblyscript) abort(
 	panic(sys.NewExitError(mod.Name(), exitCode))
 }
 
-// trace implements the same named function in AssemblyScript. Ex.
+// traceNoop implements trace in wasm to avoid host call overhead on no-op.
+var traceNoop = &wasm.Func{
```
codefromthecrypt, line 184: this is declaring a wasm definition including its inlined type. this is only used by HostModule and is how functions can end up compiled to wasm even in a host module.

internal/wasm/host.go
```diff
@@ -9,6 +9,16 @@ import (
 	"github.com/tetratelabs/wazero/internal/wasmdebug"
 )
 
+// Func is a function with an inlined type, typically used for NewHostModule.
+type Func struct {
+	// Type is the equivalent function in the SectionIDType.
+	// This will resolve to an existing or new element.
+	Type *FunctionType
+
+	// Type is the equivalent function in the SectionIDCode.
```
mathetake, line 18: Type?

codefromthecrypt: typo

internal/wasm/gofunc.go
```diff
@@ -148,8 +148,19 @@ func newModuleVal(m api.Module) reflect.Value {
 	return val
 }
 
+// RequireFunctionType returns the function type corresponding to the function
+// signature or panics if invalid.
+func RequireFunctionType(fn interface{}) *FunctionType {
```
mathetake, line 153: Go folks tend to prefix this type of function `Must` (eg regex.MustCompile), and I feel this would belong to testing etc.

internal/engine/interpreter/interpreter.go
```diff
@@ -111,18 +111,13 @@ func (ce *callEngine) popValue() (v uint64) {
...
 }
 
-// peekValues peeks api.ValueType values from the stack and returns them in reverse order.
+// peekValues peeks api.ValueType values from the stack and returns them.
 func (ce *callEngine) peekValues(count int) []uint64 {
 	if count == 0 {
 		return nil
 	}
 	stackLen := len(ce.stack)
-	peeked := ce.stack[stackLen-count : stackLen]
```
mathetake, line 120: nice :D

internal/wasm/gofunc.go
codefromthecrypt: sgtm

internal/wasm/host.go
```diff
@@ -9,6 +9,16 @@ import (
 	"github.com/tetratelabs/wazero/internal/wasmdebug"
 )
 
+// Func is a function with an inlined type, typically used for NewHostModule.
```
mathetake, line 12: what does "inlined type" mean here?

codefromthecrypt: normally a module defines the function data in 3 places, the type section, the function section (points to type) then code section. This does the same thing in struct as allowed in the text format, where you define the type inline without needing to know its ultimate position in the type index namespace. this allows flexibility as we can declare it without any regards to what module it eventually is compiled into.

mathetake: got it!

mathetake (approved):
Brilliant!

internal/wasm/host.go
codefromthecrypt: for others.. this part https://www.w3.org/TR/wasm-core-1/#abbreviations%E2%91%A6

codefromthecrypt:
cool. it is exciting to now have hybrid modules!

PR #711, opened 2022-07-22
Fixes compileDropRange when all registers are used up by live values

mathetake:
This consolidates the logic on compileDrop, and fixes the bug happinng when 
the live values (values above the drop range) include the one allocated on the stack, 
not on registers and all registers are also used up by the live values.

This only happens with large multiple returns (regardless of whether it is block of function type), therefore unlikely to happen


internal/engine/compiler/impl_amd64.go
```diff
@@ -829,46 +838,7 @@ func (c *amd64Compiler) compileCallIndirect(o *wazeroir.OperationCallIndirect) e
 
 // compileDrop implements compiler.compileDrop for the amd64 architecture.
 func (c *amd64Compiler) compileDrop(o *wazeroir.OperationDrop) error {
-	return c.emitDropRange(o.Depth)
-}
-
-func (c *amd64Compiler) emitDropRange(r *wazeroir.InclusiveRange) error {
```
mathetake, line 835: decided to implement in an arch-independent way

mathetake:
ok done... will backfill the unit tests later (maybe Monday next week)

PR #713, opened 2022-07-24
Allows wasm-defined host functions to use memory in interpreter

codefromthecrypt:
Before, we allowed stubbed host functions to be defined in wasm instead
of Go. This improves performance and reduces a chance of side-effects vs
Go. In fact, any pure function was supported in wasm, provided it only
called pure functions.

This changes internals so that a wasm-defined host function can use
memory. Notably, host functions use the caller's memory, so this is
simpler to initially support in the interpreter.

This is needed to simplify and reduce performance hit of GOARCH=wasm,
GOOS=js code, which perform a lot of memory reads and do not have
idiomatic signatures.

Note: wasm-defined host functions remain internal until we gain
experience, at least conclusion of the wasm_exec host module.


internal/engine/interpreter/interpreter_test.go
```diff
@@ -117,31 +123,49 @@ func TestInterpreter_Engine_NewModuleEngine_InitTable(t *testing.T) {
...
+	<-- (1)
+<-- (1)
+--> importing.call_import->call->div_by.go(1)
+	--> imported.call->div_by.go(1)
+		--> host.div_by.wasm(1)
+		<-- (1)
+	<-- (1)
 <-- (1)
-==> host.host_div_by(1)
+--> host.read_mem.wasm()
```
codefromthecrypt, line 150: this is log evidence that reading memory can work (though the unit tests also don't fail)

"==>" implies a go-defined function call and "-->" a wasm-defined one

codefromthecrypt: ps I'm ok removing log checking from this test, as there are enough tests out there that we don't really need to check the whole thing.

internal/testing/enginetest/enginetest.go
```diff
@@ -362,19 +365,29 @@ func runTestModuleEngine_Call_HostFn_ModuleContext(t *testing.T, et EngineTester
...
 		require.Equal(t, uint64(3), results[0])
-		require.Same(t, memory, mMemory)
 	})
 }
 
 func RunTestModuleEngine_Call_HostFn(t *testing.T, et EngineTester) {
-	runTestModuleEngine_Call_HostFn_ModuleContext(t, et) // TODO: refactor to use the same test interface.
+	t.Run("wasm", func(t *testing.T) {
+		runTestModuleEngine_Call_HostFn(t, et, hostDivByWasm)
+		if !et.IsCompiler() { // TODO: Support host wasm func that uses caller memory in compiler.
```
codefromthecrypt, line 377: This is the TODO which is likely a lot harder, to have the compiler look at ir.IsHostFunction && ir.UsesMemory to allow wasm to read the memory from the caller not the defining module.

codefromthecrypt:
I will remove the excess logging tests and backfill a couple more tests.

internal/testing/enginetest/enginetest.go
```diff
@@ -326,55 +334,75 @@ func RunTestEngine_NewModuleEngine_InitTable(t *testing.T, et EngineTester) {
...
+	hostMemoryVal := uint64(3)
+	host.Memory = &wasm.MemoryInstance{Buffer: u64.LeBytes(hostMemoryVal), Min: 1, Cap: 1, Max: 1}
+	importingMemoryVal := uint64(6)
+	importing.Memory = &wasm.MemoryInstance{Buffer: u64.LeBytes(importingMemoryVal), Min: 1, Cap: 1, Max: 1}
 
-	m := &wasm.Module{
-		TypeSection:     []*wasm.FunctionType{sig},
-		FunctionSection: []wasm.Index{0},
-		CodeSection:     []*wasm.Code{{GoFunc: &host}},
+	tests := []struct {
```
codefromthecrypt, line 347: rewrote this test which only covered one case, to cover direct and indirect host function calls that access memory. notably, this includes a host wasm function which calls another host wasm function, and proves the nested call uses the caller memory, not the host.

mathetake (approved):

PR #715, opened 2022-07-25
compiler: save conditional values at table.size

mathetake:


internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -224,3 +226,31 @@ func Test709(t *testing.T) {
 		})
 	}
 }
+
+func Test715(t *testing.T) {
+	if !platform.CompilerSupported() {
```
codefromthecrypt, line 231: nit I think should filter the test cases not the entire test, but otoh there's not much impact to early exit here.

codefromthecrypt (approved):

PR #716, opened 2022-07-25
Makes it possible to implement FunctionListener and Factory

codefromthecrypt:
This simplifies FunctionListener definition by making it possible to
implement both interfaces without intermediate state. Passing the
function definition to the before/after callbacks is the key.

This also continues efforts towards Go 1.19 doc formatting.


experimental/listener_example_test.go
```diff
@@ -0,0 +1,92 @@
...
+	sort.Strings(ret)
+	return ret
+}
+
+// NewListener implements FunctionListenerFactory.NewListener
+func (u uniqGoFuncs) NewListener(def api.FunctionDefinition) FunctionListener {
+	if def.GoFunc() == nil {
+		return nil // only track go funcs
+	}
+	return u
```
codefromthecrypt, line 42: here's the trick. since this listener only needs stack parameters, it can return itself after filtering what to listen to.

internal/asm/arm64/consts.go
```diff
@@ -692,9 +692,9 @@ const (
 	// VCNT is the CNT instruction. https://developer.arm.com/documentation/dui0802/a/A64-Advanced-SIMD-Vector-Instructions/CNT--vector-
 	VCNT
 	// VMOV has different semantics depending on the types of operands:
-	//	* LDR(SIMD&FP) if the src is memory and dst is a vector: https://developer.arm.com/documentation/ddi0596/2020-12/SIMD-FP-Instructions/LDR--immediate--SIMD-FP---Load-SIMD-FP-Register--immediate-offset--
```
mathetake, line 695: oh curious why weren't they formatted in #714 ?

mathetake (approved):

internal/asm/arm64/consts.go
codefromthecrypt: no idea

PR #718, opened 2022-07-26
asm(arm64): fixes unaligned vector stores

mathetake:


codefromthecrypt (approved):
wow so subtle

PR #723, opened 2022-07-27
Disallow direct call of host functions 

mathetake:
This has been supported for our testing purpose, and useless for end-users, as calling a host function
directly means exactly the same as calling that Go function directly.

In reality, this not only complicates the engine implementations, but also this is beyond the Wasm 
specification, and fairly specific to wazero.

This PR disables such direct calls, and reactors the unit tests accordingly.


internal/testing/proxy/proxy.go
```diff
@@ -0,0 +1,74 @@
...
+	binaryformat "github.com/tetratelabs/wazero/internal/wasm/binary"
+)
+
+// GetProxyModuleBinary creates the proxy module to proxy a function call against
+// all the exported functions in `proxyTarget`, and returns its encoded binary.
+// The resulting module exports the proxy functions whose names are exactly the same
+// as the proxy destination.
+//
+// This is used to test host call implementations.
+func GetProxyModuleBinary(moduleName string, proxyTarget wazero.CompiledModule) []byte {
```
codefromthecrypt, line 17: I love this thing. If you have a chance before or after merge, a local to this package unit test may help with future sanity.

codefromthecrypt (approved):
Thanks so much for this. It is important before we cut beta to clarify the API. I appreciate there was a lot of labor, much appreciated!

codefromthecrypt:
merging as I have work that will conflict otherwise.. one suggestion I made can happen any time.

Thanks tons @mathetake!

PR #725, opened 2022-07-27
interpreter,compiler(arm64): clears higher bits in i32.load_8/16_s

mathetake:
This also refactors the arm64 asm package and disambiguate the `MOVD` `MOVW` etc
symbols which had been used both as load and store kind instructions.


internal/engine/interpreter/interpreter.go
```diff
@@ -967,20 +967,25 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, callCtx *wasm.CallCont
 			}
 
 			switch wazeroir.SignedInt(op.b1) {
-			case wazeroir.SignedInt32, wazeroir.SignedInt64:
+			case wazeroir.SignedInt32:
+				ce.pushValue(uint64(uint32(int8(val))))
```
mathetake, line 971: these are the actual fix on interpretr

internal/engine/compiler/impl_arm64.go
```diff
@@ -2642,17 +2641,16 @@ func (c *arm64Compiler) compileLoad8(o *wazeroir.OperationLoad8) error {
 	var vt runtimeValueType
 	switch o.Type {
 	case wazeroir.SignedInt32:
-		// TODO used 32-bit variant.
-		loadInst = arm64.MOVB
+		loadInst = arm64.LDRSBW
```
mathetake, line 2644: this and below for i32.load16_s are the fix for arm64 compiler. 

Other changes are mechanical according to the asm package refactoring.

codefromthecrypt (approved):
great job!

PR #726, opened 2022-07-28
amd64: fixes i64x2.shr_s on CX register allocation

mathetake:


internal/engine/compiler/impl_vec_amd64.go
```diff
@@ -835,22 +835,25 @@ func (c *amd64Compiler) compileV128ShrImpl(o *wazeroir.OperationV128Shr) error {
...
+			c.compileMoveConditionalToGeneralPurposeRegister(s, shiftCountRegister)
+		} else { // already on register.
+			old := s.register
+			c.assembler.CompileRegisterToRegister(amd64.MOVL, old, shiftCountRegister)
+			s.setRegister(shiftCountRegister)
+			c.locationStack.markRegisterUnused(old)
+		}
+	}
+
+	c.locationStack.markRegisterUsed(shiftCountRegister)
```
mathetake, line 856: This was a bug... previosuly mistakenly this was "Unused" and therefore, 
if the next free register is CX, then the shift amount (on CX) was overridden by the
following preparation before the shift.

codefromthecrypt (approved):
thanks again!

PR #727, opened 2022-07-30
example: adds missing error check 

mathetake:
Missed in #723

### DIFF
--- a/examples/multiple-results/multiple-results.go
+++ b/examples/multiple-results/multiple-results.go
@@ -108,8 +108,8 @@ func multiValueWasmFunctions(ctx context.Context, r wazero.Runtime) (api.Module,
 //go:embed testdata/multi_value_imported.wasm
 var multiValueFromImportedHostWasm []byte
 
-// multiValueWasmFunctions return the WebAssembly which imports the Go-defined "get_age" function which
-// returns multiple results. The source is in testdata/multi_value_imported.wat
+// multiValueFromImportedHostWasmFunctions return the WebAssembly which imports the Go-defined "get_age" function.
+// The imported "get_age" function returns multiple results. The source is in testdata/multi_value_imported.wat
 func multiValueFromImportedHostWasmFunctions(ctx context.Context, r wazero.Runtime) (api.Module, error) {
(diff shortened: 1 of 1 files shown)


examples/multiple-results/multiple-results.go
```diff
@@ -108,13 +108,19 @@ func multiValueWasmFunctions(ctx context.Context, r wazero.Runtime) (api.Module,
 //go:embed testdata/multi_value_imported.wasm
 var multiValueFromImportedHostWasm []byte
 
+// multiValueWasmFunctions return the WebAssembly which imports the Go-defined "get_age" function which
+// returns multiple results. The source is in testdata/multi_value_imported.wat
```
mathetake, line 112: which .. which ... :D

mathetake: @codefromthecrypt need your editorial fix :D

codefromthecrypt (approved):
thanks for the follow-up

examples/multiple-results/multiple-results.go
```diff
@@ -108,13 +108,19 @@ func multiValueWasmFunctions(ctx context.Context, r wazero.Runtime) (api.Module,
...
 func multiValueFromImportedHostWasmFunctions(ctx context.Context, r wazero.Runtime) (api.Module, error) {
-	r.NewModuleBuilder("multi-value/host").
+	// Instantiate the host module with the exported `get_age` function which returns multiple results.
+	if _, err := r.NewModuleBuilder("multi-value/host").
```
mathetake, line 115: this error hasn't been check, but no idea why our linter didn't catch

codefromthecrypt: not sure.. wonder if because it is in the examples dir?

examples/multiple-results/multiple-results.go
codefromthecrypt: "multiValueWasmFunctions returns a wasm-defined module that imports a host function. The host module is defined using wazero.ModuleBuilder, and the wasm module with multiValueFromImportedHostWasm."

the note about the source of multiValueFromImportedHostWasm should be on multiValueFromImportedHostWasm, not here.

PR #729, opened 2022-07-30
Updates v2 spectest version

mathetake:


codefromthecrypt (approved):
all pass huh?

Is this a new draft or just new spec test version

mathetake:
Just additional spectests so nothing weird all psss without change 🏄‍♂️

PR #730, opened 2022-08-02
amd64: correctly select sign of zeros on f32x4/f64x2 min/max

mathetake:
Previously, the vector min/max operations on float lanes were not compliant
with the specification with respect to the sign bits of zerso.
If the two operands are both zeros and the sign bits are different, positive or
negative zero must be returned for max or min instructions respectively.


mathetake:
Ok turned out that I misread the amd64 manual which says `If both source operands are equal to zero` on maxpd and minpd... that is not compatible with the sign handling in wasm's min/max.

codefromthecrypt (approved):
great comments

PR #731, opened 2022-08-02
amd64: correctly select sign of zeros on f64/32 min/max

mathetake:
Similar bug to #730 


codefromthecrypt (approved):
I vote we go back in time and make negative zero and 50 shades of NaN invalid :p

PR #732, opened 2022-08-02
Set floor on Go 1.18

mathetake:
Now that Go 1.19 has been released officially, it's time to
set our floor version of Go to 1.18 following our support policy.


codefromthecrypt (approved):
oh cool, that happened faster than I expected

PR #733, opened 2022-08-04
amd64: fixes huge memory offset consts handling

mathetake:


internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -298,3 +298,35 @@ func Test730(t *testing.T) {
...
+		name := "out of bounds"
+		t.Run(name, func(t *testing.T) {
+			f := mod.ExportedFunction(name)
+			require.NotNil(t, f)
+			_, err = f.Call(ctx)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), "out of bounds memory")
+		})
+
+		// Note: this case uses large memory space, so can be slow.
```
codefromthecrypt, line 316: clarify how slow in some unit (as example) so we know slow like 10s or slow like 10m

codefromthecrypt (approved):

PR #735, opened 2022-08-04
Optimize memory bulk operations: fill on x86

dranikpg:
Nobody answered me in #512 🥲, so I decided on my own what implementation to use.

Instead of a copy loop:
```asm
.loop:
MOVB [dst], value
DEC size
CMP size, 0
INC dst
JNE .loop
```

Its now a copy loop for the first %15 bytes, and then `REP STOSQ` instructions for efficient byte copies.
See https://github.com/tetratelabs/wazero/issues/512#issuecomment-1198659163 for benchmarks.

```asm
.loop:
TEST size, 15
JZ .end
MOVB [dst], value
DEC size
INC dst
JMP .loop
.end:

; divide by 8
SHR size, 8

; replicate byte on full register
IMUL value, 0x01010101010101

; exchange registers rdi, rax, rcx
REP STOSQ
; exhcange registers back
```


dranikpg:
The performance improvements are great. We have an about 5% performance decrease for small sizes up to 16 bytes, **but are up to 30 times faster for multi kilobyte ranges** and multiple times faster for several hundred bytes 🔥 

| Test  | Old | New | Ratio |
| :---        |  :---   |  :--- | :---: |
|5|6.823|7.219|0.95|
|17|11.7|10.52|1.11|
|128|67.8|13.6|4.99|
|10000|5066.0|159.9|31.68|
|64000|29764.0|1021.0|29.15|

codefromthecrypt:
howdy, and apologies about neglect on your question earlier. Both of us had looked at it, started to think and became distracted by personal things and international travel. I promised to reply and I didn't so I apologize.

Good news is we're together even physically next week and can review this when you're ready for review (assuming you'll do both arm64 also in this PR)

mathetake:
get back to you next monday or tuesday !

internal/asm/amd64/consts.go
```diff
@@ -717,6 +721,8 @@ func InstructionName(instruction asm.Instruction) string {
 		return "MULL"
 	case MULQ:
 		return "MULQ"
+	case IMULQ:
+		return "IMUL"
```
mathetake, line 725:
```suggestion
		return "IMULQ"
```

internal/engine/compiler/compiler_post1_0_test.go
```diff
@@ -318,6 +323,7 @@ func TestCompiler_compileMemoryFill(t *testing.T) {
 				// Check the status code and the destination memory region.
 				require.Equal(t, nativeCallStatusCodeReturned, env.compilerStatus())
 				require.Equal(t, exp, mem[:checkCeil])
+
```
mathetake, line 326:
```suggestion
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -3770,8 +3770,46 @@ func (c *amd64Compiler) compileMemoryCopy() error {
...
+	c.assembler.CompileStandAlone(amd64.REPSTOSQ)
+
+	// Restore registers.
+	c.compileMaybeSwapRegisters(destinationOffset.register, amd64.RegDI)
+	c.compileMaybeSwapRegisters(value.register, amd64.RegAX)
+	c.compileMaybeSwapRegisters(fillSize.register, amd64.RegCX)
+	restoreCrossing()
+
+	c.assembler.SetJumpTargetOnNext(emptyEightGroupsJump)
+	c.assembler.CompileStandAlone(amd64.NOP)
```
mathetake, line 3804:
```suggestion
```

mathetake (commented):
Excellent! are you ready to mark this ready?

dranikpg:
Added fixes & marked as ready. Anything left?

internal/engine/compiler/impl_amd64.go
```diff
@@ -3833,35 +3870,36 @@ func (c *amd64Compiler) compileFillImpl(isTable bool, tableIndex uint32) error {
...
+
 	} else {
 		// destinationOffset += memory buffer's absolute address.
 		c.assembler.CompileRegisterToRegister(amd64.ADDQ, amd64ReservedRegisterForMemory, destinationOffset.register)
-		scale, movInst = 1, amd64.MOVB
-	}
 
-	// Negate the counter.
-	c.assembler.CompileNoneToRegister(amd64.NEGQ, copySize.register)
+		// Copy first %15 bytes with simple MOVB instruction.
```
mathetake, line 3886: comment on where 15 comes from?

mathetake (approved):

mathetake:
@dranikpg thanks! awesome work!

codefromthecrypt:
@dranikpg thanks so much for helping on this, we're going to have our first beta at the end of the month and it is awesome you are making it more efficient.

PR #736, opened 2022-08-06
Switches default random source to deterministic

codefromthecrypt:
This changes the default random source to provide deterministic values
similar to how nanotime and walltime do. This also prevents any worries
about if wasm can deplete the host's underlying source of entropy.

### DIFF
--- a/internal/platform/crypto.go
+++ b/internal/platform/crypto.go
@@ -1,0 +1,17 @@
+package platform
+
+import (
+	"io"
+	"math/rand"
+)
+
+// seed is a fixed seed value for NewFakeRandSource.
+//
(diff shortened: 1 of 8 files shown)


internal/platform/crypto.go
```diff
@@ -0,0 +1,17 @@
+package platform
+
+import (
+	"io"
+	"math/rand"
+)
+
+// seed is a fixed seed value for NewFakeRandSource.
+//
+// Trivia: While arbitrary, 42 was chosen as it is the "Ultimate Answer" in
```
codefromthecrypt, line 10: a bit wordy, but I suspect not everyone knew why we have 42 in popping up in source files

mathetake (approved):

Issue #738, opened 2022-08-07
Add Zig example


knqyf263:
Do you think it would be worthwhile to maintain an example of Zig in the wazero repository? I extended the allocation example for Zig. I can open a PR if you have an interest. Please feel free to close the issue if it is not the case.
https://github.com/knqyf263/zig-wasm-example

codefromthecrypt:
I think so. I've also at times ad-hoc defined a zig project.

Please raise a PR and also don't forget to update the GitHub examples workflow and the Makefile accordingly!

knqyf263:
Thanks for confirming. I opened #741.

PR #741, opened 2022-08-09
Add Zig example

knqyf263:
Close https://github.com/tetratelabs/wazero/issues/738


examples/allocation/zig/testdata/greet.zig
```diff
@@ -0,0 +1,79 @@
...
+pub export fn malloc(size: usize) usize {
+    var memory = allocator.alloc(u8, size) catch unreachable;
+    return @ptrToInt(memory.ptr);
+}
+
+pub export fn free(ptr: usize) void {
+    var s: []u8 = undefined;
+    s.ptr = @intToPtr([*]u8, ptr);
+    allocator.free(s);
+}
```
knqyf263, line 27: Not sure it works properly...

codefromthecrypt: @Luukdegram @inkeliz can either of you do a pass on this to make sure we are suggesting good behavior for externalizing zig allocation functions in wasm?

examples/allocation/zig/README.md
```diff
@@ -0,0 +1,16 @@
+## Zig allocation example
```
codefromthecrypt, line 1: looks nice. can you update the summary README one level higher, too?

codefromthecrypt (approved):
I'll leave this open a day or two in case folks have some feedback about "free"

Thanks for raising this. indeed, we've had some routine interest in zig, including ourselves.

examples/allocation/zig/testdata/greet.zig
mathetake: I _think_ this needs to set the size of []u8 as well..

Luukdegram: As mathetake mentioned, the size on the slice is missing. 
Here's a godbolt with an idiomatic implementation of malloc/free: https://godbolt.org/z/M7vWM6rhc

Some notes: You shouldn't really have to use `@ptrToInt` at all, pointers are nothing more than an offset into memory under the hood. By using `@ptrToInt` and `@intToPtr` you're doing what the compiler already does for you, yet removing all type information for the compiler which then cannot apply optimizations to it.
This is why you should use `?[*]u8` which represents an optional many-item pointer. This can be assigned `null` (really just offset 0 in Wasm), in case of an error, as you'd expect from `malloc`.

examples/allocation/zig/testdata/greet.zig
```diff
@@ -0,0 +1,79 @@
...
+// WebAssembly 1.0.
+pub export fn greeting(ptr: u32, size: u32) u64 {
+    const name = ptrToString(ptr, size);
+    const g = _greeting(name) catch unreachable;
+    return stringToPtr(g);
+}
+
+// ptrToString returns a string from WebAssembly compatible numeric types
+// representing its pointer and length.
+pub fn ptrToString(ptr: u32, size: u32) []u8 {
```
Luukdegram, line 67: This function shouldn't be needed as rather than accepting `u32` functions such as `greeting` can accept a many-item pointer (`[*]u8`) and can then be sliced to create a slice like as follows:
```zig
pub export fn greeting(message: [*]const u8, size: usize) u64 {
    const g = _greeting(messages[0..size]) catch return 0;
    return stringToPtr(g);
}
```

examples/allocation/zig/testdata/greet.zig
```diff
@@ -0,0 +1,79 @@
...
+pub fn _greeting(name: []u8) ![]u8 {
+    return try std.fmt.allocPrint(
+        allocator,
+        "Hello, {s}!",
+        .{name},
+    );
+}
+
+// _greet prints a greeting to the console.
+pub fn _greet(name: []u8) void {
```
Luukdegram, line 38: It would be more idiomatic to make this accept `[]const u8` as you're not mutating the slice.

examples/allocation/zig/testdata/greet.zig
```diff
@@ -0,0 +1,79 @@
+const std = @import("std");
+const allocator = std.heap.page_allocator;
+
+// main is required for Zig to compile to enable WASI.
+pub fn main() anyerror!void {
+    return;
+}
+
+extern "env" fn log(ptr: u32, size: u32) void;
```
Luukdegram, line 9: Although `u32` is technically correct, you're accepting a pointer and a pointer-size, so a more correct function signature would be something like: `extern "env" fn log(ptr: [*]const u8, size: usize) void;`.

Then in your `_log` function it can be further simplified to:
```zig
pub fn _log(message: []const u8) void {
    log(message.ptr, message.len);
}
```

Luukdegram (commented):
I wrote a few review comments to display what an idiomatic implementation looks like. I'm not asking you to make those changes as that's up to the maintainers of this project to decide, but are purely for demonstration purposes.

examples/allocation/zig/testdata/greet.zig
knqyf263: Thanks for reviewing, all! I assumed that the allocated size was stored next to the heap pointer (technically wasm memory in this case) and I didn't have to pass the length.
Also, it is super cool to take `[*]u8` as an argument. It is much simpler. I've fixed it according to your comment. Thanks.
https://github.com/tetratelabs/wazero/pull/741/commits/ea5fa0ec27dec0e2ec1a7f0f87f2123ecd1f429f

examples/allocation/zig/README.md
knqyf263: Done
https://github.com/tetratelabs/wazero/pull/741/commits/cb48cbec5edfc6bc38876ea6f76f62ee6b5d2c36

examples/allocation/zig/testdata/greet.zig
knqyf263: Brilliant. Fixed.
I couldn't use `[*]const u8` because `const` doesn't allow slicing, maybe? I'm new to Zig so please let me know if I'm wrong.
https://github.com/tetratelabs/wazero/pull/741/commits/2c904c3c1698a8ef53edcf523a159a90244f2bab

examples/allocation/zig/testdata/greet.zig
knqyf263: Done
https://github.com/tetratelabs/wazero/pull/741/commits/828c95303d4a37e3b0d0d6ffa8968e2630965787

examples/allocation/zig/testdata/greet.zig
knqyf263: Awesome.
https://github.com/tetratelabs/wazero/pull/741/commits/b191b90928b9de4c53cd4f485daf0d71658fcba2

examples/allocation/zig/testdata/greet.zig
knqyf263: One more question. If `alloc` in Zig returns an error, `?[*]u8` can be null. In that case, will the return value (`namePtr`) be 0 in Go? Should we return an error if `namePtr` is 0?
https://github.com/tetratelabs/wazero/blob/b191b90928b9de4c53cd4f485daf0d71658fcba2/examples/allocation/zig/greet.go#L86

I'm wondering if it could return 0 even though it exits successfully.

examples/allocation/zig/testdata/greet.zig
Luukdegram: > Brilliant. Fixed. I couldn't use `[*]const u8` because `const` doesn't allow slicing, maybe? I'm new to Zig so please let me know if I'm wrong. [9c349ee](https://github.com/tetratelabs/wazero/commit/9c349eef38afa1b6b6e6f508681f4539c20f4ba6)

It's because the function signature in `_greeting` accepts an `[*]u8` so to accept `[*]const u8` you'd have to update that also.

examples/allocation/zig/testdata/greet.zig
Luukdegram: Yes, the return value (`namePtr`) will be 0 in Go.

knqyf263: Do you mean we should add this sort of error handling?

```
if namePtr == 0 {
    return fmt.Errorf("malloc failed for %s", name)
}
```

examples/allocation/zig/testdata/greet.zig
knqyf263: It makes sense. Thanks.
https://github.com/tetratelabs/wazero/pull/741/commits/d964f1f9a23eb0d4e8ad6e8c1e9ccfc3acad048b

examples/allocation/zig/testdata/greet.zig
Luukdegram: Sure. As allocations can fail, we would set the right example by handling such case.

knqyf263: Thanks for confirming. I was concerned that 0 could be returned as the valid offset. Added.
https://github.com/tetratelabs/wazero/pull/741/commits/e5917dc7271d1cd919e788f59cbe01aa110f2a2a

knqyf263:
@codefromthecrypt @mathetake I've addressed all the comments.

examples/allocation/zig/testdata/build.zig
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    const run_step = b.step("run", "Run the app");
+    run_step.dependOn(&run_cmd.step);
+
+    const exe_tests = b.addTest("greet.zig");
+    exe_tests.setTarget(CrossTarget{ .cpu_arch = .wasm32, .os_tag = .wasi });
+    exe_tests.setBuildMode(mode);
+
+    const test_step = b.step("test", "Run unit tests");
```
mathetake, line 28: nit; Do we need to have these steps? I don’t see any tests in greet.zig

examples/allocation/zig/README.md
```diff
@@ -0,0 +1,16 @@
...
+Ex.
+```bash
+$ go run greet.go wazero
+wasm >> Hello, wazero!
+go >> Hello, wazero!
+```
+
+Under the covers, [greet.zig](testdata/greet.zig) does a few things of interest:
+* Uses `@ptrToInt` to change a Zig pointer to a numeric type.
+* Uses `@intToPtr` to build back a string from a pointer, len pair.
```
mathetake, line 14: This seems a stale comment

Makefile
```diff
@@ -46,6 +46,13 @@ build.bench:
 build.examples.as:
 	@cd ./examples/assemblyscript/testdata && npm install && npm run build
 
+.PHONY: build.examples.zig
+build.examples.zig: examples/allocation/zig/testdata/greet.wasm
+
+%.wasm: %.zig
+	@(cd $(@D); zig build)
+	@mv $(@D)/zig-out/bin/$(@F) $(@D)
```
mathetake, line 54: Can we add the zig-out/** into .gitignore?

examples/allocation/zig/greet.go
```diff
@@ -0,0 +1,142 @@
...
+	} else {
+		fmt.Println("go >>", string(bytes))
+	}
+
+	return nil
+}
+
+// Looks like Zig exits the module after calling each exported function
+// regardless of whether there was an error. We need to handle exit code 0
+// so that the subsequent functions can be called.
```
mathetake, line 128: This is resolved once we user `reactor` model which is available with `-wasi-exec-model` on the master branch of zig. Could you add TODO comment on that?

mathetake: FYI https://github.com/WebAssembly/WASI/blob/main/legacy/application-abi.md this is where what’s reactor is, though it’s deprecated after wasi2 will be released

examples/allocation/zig/testdata/greet.zig
```diff
@@ -0,0 +1,65 @@
...
+}
+
+extern "env" fn log(ptr: [*]const u8, size: u32) void;
+
+// _log prints a message to the console using log.
+pub fn _log(message: []const u8) void {
+    log(message.ptr, message.len);
+}
+
+pub export fn malloc(length: usize) ?[*]u8 {
```
mathetake, line 16: Curious how this optional is translated at the Wasm layer. Returning null just becomes returning 0?

examples/allocation/zig/greet.go
mathetake: Oh actually we are using the HEAD-ish zig so you can use the reactor I guess

examples/allocation/zig/testdata/greet.zig
```diff
@@ -0,0 +1,65 @@
+const std = @import("std");
+const allocator = std.heap.page_allocator;
+
+// main is required for Zig to compile to enable WASI.
+pub fn main() anyerror!void {
```
mathetake, line 5: Once you use the reactor, I believe we don’t need the main function at all

mathetake (approved):
Oh sorry forget about everything on reactor - this example is not necessarily WASI

examples/allocation/zig/testdata/greet.zig
Luukdegram: > Curious how this optional is translated at the Wasm layer. Returning null just becomes returning 0?

Yes, null becomes `0`. An optional pointer in Zig works the same as a pointer in C, therefore this type matches the C-ABI correctly. Whereas a regular pointer in Zig is much more strict and doesn't allow `null` as a value, meaning you wouldn't be able to convey an allocation failure.

examples/allocation/zig/testdata/build.zig
knqyf263: I just used the generated template as is. Removed. https://github.com/tetratelabs/wazero/pull/741/commits/5cfab1321b74dd1e8367aad6074db7b64d6758fd

examples/allocation/zig/README.md
knqyf263: Updated. `@ptrToInt` is still used.
https://github.com/tetratelabs/wazero/pull/741/commits/298914b368773bb36a0475957e2544285652f74f

Makefile
knqyf263: I added zig-out and zig-cache in my local so that I would not add those dirs accidentally, but didn't push it as I was not sure you like that. Done.
https://github.com/tetratelabs/wazero/pull/741/commits/c2aafcf362573739e76eade10b66459713eb1390

examples/allocation/zig/testdata/greet.zig
Luukdegram: The main is required because the `build.zig` file creates an executable (a runnable binary). If you want to simply call into the Wasm module's functions as a 'library', the `build.zig` file should create a dynamic library instead.
Alternatively, you can overwrite the startup logic by defining your own `_start` function instead like so:
```zig
pub fn export _start() void {
    // do stuff
}
```

examples/allocation/zig/greet.go
```diff
@@ -0,0 +1,142 @@
...
+func isFailureExit(err error) bool {
+	if exitErr, ok := err.(*sys.ExitError); !ok || exitErr.ExitCode() != 0 {
```
codefromthecrypt, line 130: hmm this is a problem I think because state in wazero is changed when the module exits as this is an example, I don't think we should encourage people to use a module after it is freed. Also it is quite surprising to have something exit on account of a random exported function use. Ack WASI isn't well specified, but this isn't de'facto behavior either.

I think the ways out are:
* Change this comment to a TODO, which warns the person this is not something they should do and will break
* Change to reactor model and remove this comment
* Change zig to not exit after each exported function call, as I don't think any other lib does this so it can lead to problems.
* Add a hack option in wazero to ignore exit and use that here

wdyt @mathetake @Luukdegram

examples/allocation/zig/testdata/greet.zig
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// _greet prints a greeting to the console.
+pub fn _greet(name: []const u8) void {
+    const g = _greeting(name);
+    const s = std.fmt.allocPrint(
+        allocator,
+        "wasm >> {s}",
+        .{g},
+    ) catch unreachable;
```
codefromthecrypt, line 40: why are we catching unreachable? if there's an allocation failure we should fail all the way out, right? none of the other examples swallow errors.

one thing to keep in mind is if we are doing something in the example, it is indeed an example, so we are saying this is the right way to code. If catching unreachable is normal, then here is the right thing, though probably needs a line comment or in the README why we catch unreachable.

examples/allocation/zig/README.md
```diff
@@ -0,0 +1,16 @@
...
+This example shows how to pass strings in and out of a Wasm function defined in Zig, built with `zig build`.
+
+Ex.
+```bash
+$ go run greet.go wazero
+wasm >> Hello, wazero!
+go >> Hello, wazero!
+```
+
+Under the covers, [greet.zig](testdata/greet.zig) does a few things of interest:
```
codefromthecrypt, line 12: make sure these are up to date please, and any other notes of interest in a paragraph like `Note: `. I think there are a couple surprising things, so whatever's left post review should either be a line comment or in the ad-hoc note thing (thinking about exit on each export function unless that's going to be sorted soon)

codefromthecrypt (commented):
I think this is in good shape except the swallowing of unreachable and also module exit. both need explanation so that we don't steer newbies into dodgy things.

examples/allocation/zig/greet.go
mathetake: I recall this can be avoided by compiling this as the shared library instead of exe, but my memory on zig is fuzzy now

examples/allocation/zig/testdata/greet.zig
mathetake: that's good. thank you for the confirmation

examples/allocation/zig/testdata/greet.zig
mathetake: `catch unreachable` here in zig means `if there's an error, panics`, so this doesn't swallow!

examples/allocation/zig/greet.go
mathetake: or, as I said and limit this example to wasi, we can use `reactor` https://github.com/mathetake/proxy-wasm-zig-sdk/blob/main/build.zig#L16

Luukdegram: The reason why the module exits after execution is because the `build.zig` file is creating an executable. We use the 'command' execution model by default in such case. This means there's 3 options that can be used to solve this:

- Use the reactor model instead.
- Build a shared library in the `build.zig` file: `b.addSharedLibrary("name", .unversioned);`
- Overwrite the startup function by exporting your own _start using `pub export fn _start() void{}`.

Of those 3 options, the most correct one would be to build a shared library as those examples don't necessarily use WASI.

examples/allocation/zig/testdata/greet.zig
Luukdegram: > `catch unreachable` here in zig means `if there's an error, panics`, so this doesn't swallow!

This is incorrect. That's only the [case for safe building](https://ziglang.org/documentation/master/#unreachable) modes such as debug and release-safe. Considering we're building for Wasm, people tend to use release-small to get the smallest binary size. In such a case, this would lead into undefined behavior. `unreachable` should only be used when the author knows it *should* never be hit, unless a developer makes a mistake.

The correct way to solve this is to handle the error appropriately and notify the runtime. For example purposes, you could also use `@panic()` instead.

examples/allocation/zig/greet.go
knqyf263: Thank you, guys. I've switched to lib.
https://github.com/tetratelabs/wazero/pull/741/commits/b9018858415e81bb1f0aa4473748624b5585772d

examples/allocation/zig/testdata/greet.zig
knqyf263: @Luukdegram Thanks for your input. I didn't know `unreachable` behaved differently depending on modes.
`@panic()` takes `[]u8` and I had no idea what is an idiomatic implementation here, so I just switched and returned `[]u8` according to the returned error. Please let me know if there is a better way.
https://github.com/tetratelabs/wazero/pull/741/commits/398dc4f0bf6bc056dd044e52fc4e9e9b8f082573

examples/allocation/zig/greet.go
inkeliz: I think it's also possible to use `exe.wasi_exec_model = .reactor;` on build.zig to use reactor and prevent exit. _If I remember correctly, was @mathetake who add that feature._ 😅

examples/allocation/zig/README.md
knqyf263: Sure. Added a note about `unreachable`.
https://github.com/tetratelabs/wazero/pull/741/commits/398dc4f0bf6bc056dd044e52fc4e9e9b8f082573

codefromthecrypt:
Fantastic collaboration. @Luukdegram you're the best!

mathetake:
Thank you so much @Luukdegram! 🙌

knqyf263:
Thank you, @codefromthecrypt @mathetake! Also, I'm so honored to have learned that much from @Luukdegram though I'm new to Zig. I really appreciate it.
As we had some discussion about WASI, adding a Zig WASI example might also be useful.

Luukdegram:
Glad I could help! Feel free to message me anytime in the future :)

codefromthecrypt:
@knqyf263 you feel up to adding a wasi example in zig? no issues on our side if you have time. Zig is well loved by us.

knqyf263:
@codefromthecrypt I won't have time until the first half of next week. If you are not in hurry, I can add the WASI example later next week.

codefromthecrypt:
no hurry at all, thanks!

codefromthecrypt:
@knqyf263 once you have the WASI example up, I'll offer to put together a zig page like these. https://wazero.io/languages/

knqyf263:
Nice. I'm sorry to be late. I came up with an idea for Go plugin system over WebAssembly and had been developing that in my spare time 😄 It uses wazero under the food. You define proto files for Protocol Buffers and my protoc plugin generates TinyGo Wasm SDK like [this](https://github.com/tetratelabs/wazero/blob/df93d70654276ff7839f491516478f1179e61c1f/examples/allocation/tinygo/testdata/greet.go). The proto file would be as below.

```
service Greeter {
  // Sends a greeting
  rpc Greet(GreetRequest) returns (GreetReply) {}
}

// The request message containing the user's name.
message GreetRequest {
  string name = 1;
}

// The response message containing the greetings
message GreetReply { string message = 1; }
```

Hopefully, I will wrap it up this week and move back to the Zig work.

mathetake:
> You define proto files for Protocol Buffers and my protoc plugin generates TinyGo Wasm SDK ...

heh that sounds great!

evacchi:
hey, I have updated the Zig section on the wazero website here, please do chime in https://github.com/tetratelabs/wazero/pull/1051

PR #742, opened 2022-08-11
amd64: do not load/store higher bits of 32-bit int/float

mathetake:
Previously, regardless of the value types, i've been using MOVQ to load/store the value from/to stack.
That was unnecessarily costly as i32 and f32 values don't need to move higher 32-bits.

### DIFF
--- a/internal/engine/compiler/impl_amd64.go
+++ b/internal/engine/compiler/impl_amd64.go
@@ -959,7 +959,7 @@ func (c *amd64Compiler) compilePick(o *wazeroir.OperationPick) error {
 		var inst asm.Instruction
 		if o.IsTargetVector {
 			inst = amd64.MOVDQU
-		} else if pickTarget.valueType == runtimeValueTypeI32 {
+		} else if pickTarget.valueType == runtimeValueTypeI32 { // amd64 cannot copy single-precisions between registers.
 			inst = amd64.MOVL
 		} else {
 			inst = amd64.MOVQ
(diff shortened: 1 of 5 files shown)


mathetake:
lol only rust example has started failing

internal/engine/compiler/impl_amd64.go
```diff
@@ -4357,9 +4376,12 @@ func (c *amd64Compiler) compileLoadValueOnStackToRegister(loc *runtimeValueLocat
 		inst = amd64.MOVDQU
 	case runtimeValueTypeV128Hi:
 		panic("BUG: V128Hi must be be loaded to a register along with V128Lo")
-	default:
+	case runtimeValueTypeI32, runtimeValueTypeF32:
```
mathetake, line 4379: this and the same diff in `compileReleaseRegisterToStack` below are the actual change in this patch

internal/engine/compiler/impl_amd64.go
```diff
@@ -274,19 +274,23 @@ func (c *amd64Compiler) compileGlobalGet(o *wazeroir.OperationGlobalGet) error {
...
 		vt = runtimeValueTypeI64
 	case wasm.ValueTypeF32:
+		inst = amd64.MOVL
 		vt = runtimeValueTypeF32
 		valueReg, err = c.allocateRegister(registerTypeVector)
 		if err != nil {
 			return err
 		}
 	case wasm.ValueTypeF64:
+		inst = amd64.MOVQ
```
codefromthecrypt, line 293: maybe add default panic BUG because this requires a close look in rare case a new value type

internal/engine/compiler/impl_amd64.go
```diff
@@ -4357,9 +4376,12 @@ func (c *amd64Compiler) compileLoadValueOnStackToRegister(loc *runtimeValueLocat
 		inst = amd64.MOVDQU
 	case runtimeValueTypeV128Hi:
 		panic("BUG: V128Hi must be be loaded to a register along with V128Lo")
-	default:
+	case runtimeValueTypeI32, runtimeValueTypeF32:
+		inst = amd64.MOVL
+	case runtimeValueTypeI64, runtimeValueTypeF64:
 		inst = amd64.MOVQ
 	}
```
codefromthecrypt, line 4383: again maybe default: panic BUG to ensure nothing runs away

codefromthecrypt (approved):
only nits. good job

PR #744, opened 2022-08-12
compiler: drop higher bits on table index

mathetake:
as of #743 #742 , store into the value stack only stores lower 32-bits,
so we have to drop the higher 32-bits.


codefromthecrypt (approved):
thanks for the nice unit test

PR #746, opened 2022-08-16
Enable comparative benchmark builds on darwin/aarch64 (ie Apple Silicon)

mt-inside:
Enable for Wasmer, as confirmed that the version we use has arm64 support.
Enable for Wasmtime after bumping to a version which supports arm64.
Don't enable for wasm-edge as they don't support arm64/darwin.


codefromthecrypt (approved):
cool thanks!

PR #747, opened 2022-08-17
Externalize compilation cache by compilers

### LINKED ISSUE #618: Serialize a compiled module

mathetake:
This adds the experimental support of the file system compilation cache.
Notably, `experimental.WithCompilationCacheDirName` allows users to configure
where the compiler writes the cache into.

Versioning/validation of binary compatibility has been done via the release tag
(which will be created from the end of this month). More specifically, the cache
file starts with header with the hardcoded wazero version.

Note the cache implementation doesn't do any security mitigation like signing/verification,
and that would be a future TODO once we reach some stability.

The end-to-end test will be added in a subsequent PR.

Fixes #618


internal/integration_test/bench/bench_test.go
```diff
@@ -50,11 +52,39 @@ func BenchmarkInitialization(b *testing.B) {
 	}
 }
 
-func runInitializationBench(b *testing.B, r wazero.Runtime) {
+func BenchmarkCompilation(b *testing.B) {
```
mathetake, line 55: Result: 

```
BenchmarkCompilation
BenchmarkCompilation/with_extern_cache
BenchmarkCompilation/with_extern_cache-10         	    2008	    526860 ns/op
BenchmarkCompilation/without_extern_cache
BenchmarkCompilation/without_extern_cache-10      	     165	   7080493 ns/op
```

internal/engine/compiler/engine_cache.go
```diff
@@ -0,0 +1,183 @@
...
+		c.indexInModule = wasm.Index(i)
+		c.sourceModule = module
+	}
+	return
+}
+
+var (
+	wazeroMagic = "WAZERO"
+	// version must be synced with the tag of the wazero library.
+	version         = "1.0.0-dev"
```
mathetake, line 97: wondering if this is the right place to place version... maybe a dedicated pkg?

experimental/compilation_cache.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"context"
+
+	"github.com/tetratelabs/wazero/internal/compilationcache"
+)
+
+// WithCompilationCacheDirName configures the destination directory of the compilation cache.
+// Regardless of the usage of this, the compiled functions are cached in memory, but its lifetime is
+// bound to the lifetime of wazero.Runtime or wazero.CompiledModule.
+// With the given non-empty directory, wazero persists the cache into the directory and that caches
```
codefromthecrypt, line 12:
```suggestion
//
// With the given non-empty directory, wazero persists the cache into the directory and that cache
```

experimental/compilation_cache.go
```diff
@@ -0,0 +1,26 @@
...
+// will be used as long as the running wazero version match the version of compilation wazero.
+//
+// Note: this is an experimental, and currently no validation on cache is applied.
```
codefromthecrypt, line 15:
```suggestion
// A cache is only valid for use in one wazero.Runtime at a time. Concurrent use
// of a wazero.Runtime is supported, but multiple runtimes must not share the
// same directory.
//
// Note: The embedder must safeguard this directory from external changes.
```

internal/compilationcache/compilationcache.go
```diff
@@ -0,0 +1,44 @@
...
+	// ok=false with err=nil. content.Close() is automatically called by
+	// the caller of this Get.
+	//
+	// Note: the returned content won't go through the validation pass of Wasm binary
+	// which is applied when the binary is compiled from scratch without cache hit.
+	// Its implication is that the implementors of ExternCache might want to have
+	// their own validation phases. For example, sign the binary passed to Add, and
+	// verify the signature of the stored cache before returning it via Get, etc.
+	Get(key Key) (content io.ReadCloser, ok bool, err error)
+	// Add is called when the runtime is trying to add the new cache entry.
```
codefromthecrypt, line 31:
```suggestion
	//
	// Add is called when the runtime is trying to add the new cache entry.
```

internal/compilationcache/compilationcache.go
```diff
@@ -0,0 +1,44 @@
...
+	// The given `content` must be un-modified, and returned as-is in Get method.
+	//
+	// Note: the `content` is ensured to be safe through the validation phase applied on the Wasm binary.
+	Add(key Key, content io.Reader) (err error)
+	// Delete is called when the cache on the `key` returned by Get is no longer usable, and
```
codefromthecrypt, line 36:
```suggestion
	//
	// Delete is called when the cache on the `key` returned by Get is no longer usable, and
```

internal/engine/compiler/engine_cache.go
codefromthecrypt: yeah maybe like func-e we should show the flag that ends up used also by goreleaser to assign this.

```
// version is the string representation of globals.GlobalOpts
// We can't use debug.ReadBuildInfo because it doesn't get the last known version properly
// See https://github.com/golang/go/issues/37475
var version = "dev"
```

maybe by now we can use ReadBuildInfo instead?

codefromthecrypt (commented):
unleashing comments to allow you to drift things while I read the rest

internal/engine/compiler/engine_cache.go
mathetake: so wazero is library so we cannot use gorelease, but maybe `ReadBuildInfo` sounds promising

codefromthecrypt: oh yeah of course. 🤞 on ReadBuildInfo esp as we have a higher floor go version maybe can have some luck

mathetake: cool if wazero is used as the dependency, and inject the following code somewhere in wazero

```
	info, _ := debug.ReadBuildInfo()
	for _, dep := range info.Deps {
		if strings.Contains(dep.Path, "wazero") {
			fmt.Println(dep.Version)
		}
	}
```

then we get like `v0.0.0-20220818054749-076d3245e307`

internal/compilationcache/compilationcache.go
```diff
@@ -0,0 +1,44 @@
...
+// here means that the compiled binary cache by compilers. Regardless of the usage of
+// ExternCache, the compiled functions are cached in memory, but its lifetime is
+// bound to the lifetime of wazero.Runtime or wazero.CompiledModule.
+// Usually, the compilation of Wasm binary is time-consuming. Therefore, you might
+// want to cache the compilation result across the processes of wazero users.
+//
+// See NewFileCache for the example implementation.
+type Cache interface {
+	// Get is called when the runtime is trying to get the cached content.
+	// Implementations are supposed to return `content` which can be used to
```
codefromthecrypt, line 18: this is too wordy, maybe refer to map for an example of how this works? since content is only used for compilation of functions seems we can be more concrete generally

internal/compilationcache/compilationcache.go
```diff
@@ -0,0 +1,44 @@
...
+	// Note: the `content` is ensured to be safe through the validation phase applied on the Wasm binary.
+	Add(key Key, content io.Reader) (err error)
+	// Delete is called when the cache on the `key` returned by Get is no longer usable, and
+	// must be purged. Specifically, this is called happens when the wazero's version has been changed.
+	// For example, that is when there's a difference between the version of compiling wazero and the
+	// version of the currently used wazero.
+	Delete(key Key) (err error)
+}
+
+// Key represents the 256-bit unique identifier assigned to each cache content.
```
codefromthecrypt, line 43: cache entry? prefer cached function if that's what this is for. let's be concrete unless the cache is used for multiple types of things

codefromthecrypt (commented):
more comments

codefromthecrypt (approved):
I think exposing this as an experimental hook simply with a dir is a great way to start. There's some more polishing to do on internal code, but this can allow users to start giving feedback meanwhile.

PR #748, opened 2022-08-18
Pass context into NewRuntime

mathetake:


codefromthecrypt (approved):
oops! can't believe we forgot this. Can you do a search of all files even markdown and godoc to ensure no drift?

We need this in before we cut the beta the end of month, as we don't want to drift like this later.

PR #751, opened 2022-08-19
Use ReadBuildInfo to embed wazero version

mathetake:


internal/version/test/main_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+	"github.com/tetratelabs/wazero/internal/testing/require"
+	"github.com/tetratelabs/wazero/internal/version"
+)
+
+// TestGetWazeroVersion ensures that GetWazeroVersion returns the version of wazero in the go.mod in the
+// downstream wazero users.
+func TestGetWazeroVersion(t *testing.T) {
+	// This matches the one in the "replace" statement in the go.mod.
+	const exp = "v0.0.0-20220818123113-1948909ec0b1"
```
codefromthecrypt: looks too much like a real version, maybe set it to v0.0.0? we are wazero after all

codefromthecrypt (approved):

internal/version/version.go
```diff
@@ -1,4 +1,32 @@
...
+//
+// Note: this is tested in ./testdata/main_test.go with a separate go.mod to pretend as the wazero user.
+func GetWazeroVersion() (ret string) {
+	ret = "dev"
+	info, ok := debug.ReadBuildInfo()
+	if ok {
+		for _, dep := range info.Deps {
+			// Note: here's the assumption that wazero is imported as github.com/tetratelabs/wazero.
+			if strings.Contains(dep.Path, "github.com/tetratelabs/wazero") {
+				fmt.Println(dep.Sum)
```
mathetake, line 26: garbage

internal/version/test/main_test.go
mathetake: this must be a real one otherwise go test ./... will fail saying that wazero not found regardless of the existence of replace statement in go.mod

PR #753, opened 2022-08-19
Adds test to verify cross-process file system cache works 

mathetake:


internal/integration_test/filecache/filecache_test.go
```diff
@@ -0,0 +1,66 @@
...
+		// This case, this is the parent of the test.
+		cacheDir = t.TempDir()
+
+		// Before running test, no file should exist in the directory.
+		files, err := os.ReadDir(cacheDir)
+		require.NoError(t, err)
+		require.True(t, len(files) == 0)
+
+		// Get the executable path of this test.
+		testExecutable, err := os.Executable()
```
mathetake, line 37: so this is the key - with this, we can os.Exec this test itself recursively regardless of this is executed via `go test ./...` or as the stand alone binary as we do with scratch containers.

codefromthecrypt: cool!

codefromthecrypt (approved):
slick

PR #754, opened 2022-08-19
Disallows nil context and fixes linters

codefromthecrypt:
staticcheck linters broke until recent golangci-lint. Now, normal
behaviour of enforcing no nil context works again. Ex.
```
assemblyscript/assemblyscript_example_test.go:16:25: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
	r := wazero.NewRuntime(nil)
```

Since default lint already checks for nil context, this removes our
permission of nil context args. The original reason we permitted nil is
no longer valid: we once allowed context to be stashed in config, and
removed that as it caused bugs. We forgot to undo allowing nil
explicitly.

Note: this doesn't particularly check in our code for nil context,
similar as we don't particularly check in our code for nil anything
else. End users should use linters as none of our parameters should be
nil anyway.


mathetake (approved):

Makefile
```diff
@@ -4,10 +4,10 @@ comma := ,
 space :=
 space +=
 
-goimports := golang.org/x/tools/cmd/goimports@v0.1.10
-golangci_lint := github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2
+goimports := golang.org/x/tools/cmd/goimports@v0.1.12
+golangci_lint := github.com/golangci/golangci-lint/cmd/golangci-lint@v1.48.0
```
codefromthecrypt, line 8: this version finally works without warnings in go 1.19

PR #756, opened 2022-08-22
vs: adds the benchmark to see the cost of host function calls

mathetake:
This adds the new vs target to measure the cost of host function calls.

```wat

(module
  ;; env.f must be a host function for benchmarks on the cost of host calls which cross the Wasm<>Go boundary.
  (func $host_func (import "env" "f") (param i64) (result i64))
  ;; call_host_func calls "env.f" and returns the resut as-is.
  (func (export "call_host_func") (param i64) (result i64)
    local.get 0
    call $host_func
  )
)

```

Notably, I can see that wazero is roughly 2x to 4x times faster than CGO-based 
runtimes in terms of host call boundary crossing. One implication here is that
we can just focus on the native code generation rather than how to organize the
Go function calls. For example, it's not prioritized to call Go functions directly 
from the native code.

### amd64

```
make bench.check
goos: linux
goarch: amd64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/clock

wazero-compiler 1257.63 ns/op
wasmedge        2276.99 ns/op

wazero-compiler 1234.30 ns/op
wasmer          2676.20 ns/op


wazero-compiler 1248.29 ns/op
wasmtime        4362.91 ns/op
```

### arm64

```
make bench.check
goos: darwin
goarch: arm64

wazero-compiler 522.03  ns/op
wasmer          1300.86 ns/op

wazero-compiler 533.03  ns/op
wasmtime        1897.03 ns/op
```


mathetake:
fyi @anuraaga  hostcall itself is already fast enough so I won't implement direct Go function calls from machine code for now

internal/integration_test/vs/runtime.go
```diff
@@ -86,6 +89,11 @@ func (r *wazeroRuntime) Compile(ctx context.Context, cfg *RuntimeConfig) (err er
 			ExportFunction("log", r.log).Compile(ctx, wazero.NewCompileConfig()); err != nil {
 			return err
 		}
+	} else if cfg.EnvFReturnValue != 0 {
+		if r.env, err = r.runtime.NewModuleBuilder("env").
+			ExportFunction("f", func(uint64) uint64 { return cfg.EnvFReturnValue }).Compile(ctx, wazero.NewCompileConfig()); err != nil {
```
anuraaga, line 94: It would be good to get numbers for various number of arguments, and especially with `context.Context` and `api.Module` included

anuraaga (commented):
It would probably also be good to have a benchmark of the same function implemented in wasm to compare direct invoke vs crossing host boundary.

Everything is priority so no worries if there's other stuff to work on, but I think people like wazero as a runtime well enough compared to others so the conversation is probably shifting not wazero vs other runtimes, but wazero (wasm) vs native. For various reasons, people expect native performance from wasm and get disappointed when it commonly isn't - if optimizing host calls makes it closer to native then in the long term it'll probably be a much appreciated optimization but now may not be the time.

internal/integration_test/vs/runtime.go
codefromthecrypt: agreed. maybe we can to keep dimensions under wraps, compare one more with a memory write (offset/len)? This would use the context/mod arg and is extremely typical, probably more typical than pure function.

codefromthecrypt (approved):
I like this, though I would compare to wasm as that way we aren't just seeing how much faster we are than others, we can see how much slower it is than wasm.

ex. two cases:
increment i64 (or add or swap) both in wasm vs hostcall
memory read or write both in wasm vs host call

mathetake:
I would add the host vs wasm benchmarks in another PR

internal/integration_test/vs/runtime.go
mathetake: oh ok

mathetake: turns out the same result with this https://github.com/tetratelabs/wazero/pull/756/commits/c0d7682a9eb8b0fe7296233a7b826fbbf1fbfad3 

I just default it here, and will do the comparison (with/without api.Module and context) in wasm vs host bench in another PR

internal/integration_test/vs/runtime.go
```diff
@@ -86,6 +89,16 @@ func (r *wazeroRuntime) Compile(ctx context.Context, cfg *RuntimeConfig) (err er
 			ExportFunction("log", r.log).Compile(ctx, wazero.NewCompileConfig()); err != nil {
 			return err
 		}
+	} else if cfg.EnvFReturnValue != 0 {
+		if r.env, err = r.runtime.NewModuleBuilder("env").
+			ExportFunction("f",
+				// Note: accepting (context.Context, api.Module) is the slowest type of host function with wazero.
+				func(context.Context, api.Module, uint64) uint64 {
```
codefromthecrypt, line 96: I think probably perf is same as we don't optimize much for the new `UsesMemory` yet, or we do and it doesn't matter :)

PR #758, opened 2022-08-23
bench: compare host call costs on Wasm-impl vs Go-impl

mathetake:
Follow-up on #756 

This adds the benchmarks to measure the cost of host calls
between Wasm implemented vs Host implemented.

#### arm64

```
goos: darwin
goarch: arm64
BenchmarkHostFunctionCall
BenchmarkHostFunctionCall/call_go_host
BenchmarkHostFunctionCall/call_go_host-10         	 1328292	       904.5 ns/op
BenchmarkHostFunctionCall/call_wasm_host
BenchmarkHostFunctionCall/call_wasm_host-10       	 5341951	       221.4 ns/op
```

#### amd64

```
goos: linux
goarch: amd64
BenchmarkHostFunctionCall/call_go_host-32         	  495591	      2356 ns/op	    1464 B/op	      14 allocs/op
BenchmarkHostFunctionCall/call_wasm_host-32       	 2382007	       503.0 ns/op	    1232 B/op	       5 allocs/op
```

### DIFF
--- a/internal/integration_test/bench/hostfunc_bench_test.go
+++ b/internal/integration_test/bench/hostfunc_bench_test.go
@@ -1,0 +1,218 @@
+package bench
+
+import (
+	"context"
+	_ "embed"
+	"encoding/binary"
+	"github.com/tetratelabs/wazero/api"
+	"github.com/tetratelabs/wazero/internal/engine/compiler"
+	"github.com/tetratelabs/wazero/internal/platform"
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
cool and interesting.

one brain teaser is how we eventually measure perf of wasm -> wasm vs wasm -> host

PR #762, opened 2022-08-24
site: Addresses feedback from TinyGo folks

codefromthecrypt:
Thanks to @aykevl @dkegel-fastly @mathetake @anuraaga and jimmysl lee
for all the feedback. Hopefully, this gets most of it.


codefromthecrypt:
https://deploy-preview-762--wazero.netlify.app/languages/tinygo/

mathetake (approved):

site/content/languages/tinygo.md
```diff
@@ -165,19 +258,85 @@ in [runtime_wasm_wasi.go][13]. `syscall.Chdir` is implemented with the same
...
+wasm anyway, there may be limited effect to removing goroutine support.
+
+## Optimizations
+
+Below are some commonly used configurations that allow optimizing for size or
+performance vs defaults. Note that sometimes one sacrifices the other.
+
+### Binary size
+
+Those who aren't concerned about binary (`%.wasm` file) size should not
```
anuraaga, line 306:
```suggestion
Those who are concerned about binary (`%.wasm` file) size should not
```

anuraaga: Or perhaps it was about being "extremely converned" about binary size. But this didn't read well for me

codefromthecrypt: confusing double-negative I made. I'll add fixing that to another change

codefromthecrypt: maybe

> Those unconcerned about binary (`%.wasm` file) size should use TinyGo's
defaults. Those with size constraints can reduce the wasm binary size by
changing `tinygo` flags. This can result in binaries less than 100KB.

anuraaga: I think my tension is with the fact that those unconcerned about binary size may be best using `-O2`.

Maybe

> TinyGo's defaults produce fairly small binaries. Those particularly concerned with binary size can make them even smaller by...

codefromthecrypt: I guess as  defaults can approach a meg for simple code, I have some difficulty saying they are small.

I'll just ax the intro sentence and say the latter.

PR #767, opened 2022-08-26
exp: create the compilation cache dir when not exist

mathetake:


codefromthecrypt (approved):
LGTM when this becomes non-experimental, we'd defer error till compile, but for now looks completely right.

PR #768, opened 2022-08-26
vs: adds in-Wasm loop inside the memory bench

mathetake:
Without this, CGO cost becomes dominant in each literation, and therefore couldn't measure the actual
cost of memory  load/store.


internal/integration_test/vs/bench_memory.go
```diff
@@ -14,6 +14,7 @@ const (
 	i32ValueMemoryOffset = 32
 	i64                  = "i64"
 	i64ValueMemoryOffset = 64
+	inWasmIteration      = 100
```
mathetake, line 17: setting this to >=1000, we can see the actual diff, and I found that there's room for optimization in our memory access code

codefromthecrypt (approved):
makes sense

PR #770, opened 2022-08-26
Add some more context on why 3 works as the fd for the root of a file…

anuraaga:
…system

Need these notes for when cribbing from wazero for implementing in other languages :D

### DIFF
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -68,7 +68,6 @@ var emptyFSContext = &FSContext{
 // NewFSContext returns a mutable context if the fs is not EmptyFS. The file
 // descriptor used in the case of non-empty is 3.
 //
-//
 // Why file descriptor 3?
 //
 // While not specified, the most common WASI implementation, wasi-libc, expects
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -73,7 +73,9 @@ func NewFSContext(fs fs.FS) *FSContext {
 	return &FSContext{
 		fs: fs,
 		openedFiles: map[uint32]*FileEntry{
-			3: {Path: "/"}, // after STDERR
+			// after STDERR. wasi-libc always searches for preopened files starting with the value 3
+			// https://github.com/WebAssembly/wasi-libc/blob/2057ce9262f76f7ef5a2002fa16da219e2176896/libc-bottom-half/sources/preopens.c#L228
+			3: {Path: "/"},
 		},
 		lastFD: 3,
 	}
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -65,7 +65,19 @@ var emptyFSContext = &FSContext{
 	lastFD:      2,
 }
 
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
Added more notes since indeed it is curious without them. feel free to merge if ok with them!

codefromthecrypt:
when you merge please kill the excess commit notes.. I made some fuzz

internal/sys/fs.go
```diff
@@ -65,15 +65,26 @@ var emptyFSContext = &FSContext{
...
-// NewFSContext returns a mutable context if the fs is not EmptyFS.
+// NewFSContext returns a mutable context if the fs is not EmptyFS. The file
+// descriptor used in the case of non-empty is 3.
+//
+// Why file descriptor 3?
+//
+// While not specified, the most common WASI implementation, wasi-libc, expects
+// POSIX style file descriptor allocation, where the lowest available number is
+// used to open the next file. Since 1 and 2 are taken by stdout and stderr
+// respectively, this means the `fs` parameter will be 3.
```
anuraaga, line 76:
```suggestion
// respectively, this means the `fd` parameter will be 3 in the first call to `fd_prestat_get`.
```

Mentioning fd_prestat_get here might be TMI, fs I guess is a typo though

internal/sys/fs.go
```diff
@@ -65,15 +65,26 @@ var emptyFSContext = &FSContext{
 	lastFD:      2,
 }
 
-// NewFSContext returns a mutable context if the fs is not EmptyFS.
+// NewFSContext returns a mutable context if the fs is not EmptyFS. The file
+// descriptor used in the case of non-empty is 3.
```
anuraaga, line 69:
```suggestion
// descriptor for the always-open root file, `/`, is 3.
```

The file descriptor of what? is what I wondered when reading the comment so tried to clarify.

Though I didn't use the word preopened here and went with always-open - I had never known the concept of preopened files until diving deep into these functions, but basically it seems file system roots are all preopened, and any other path isn't. "always-open root file" seems to convey both the semantics and the concept, maybe

anuraaga (commented):
Thanks! Added a couple more suggested tweaks, will merge with them if they seem OK

internal/sys/fs.go
codefromthecrypt: the only parameter is `fs`, so I think I'll tweak the doc a bit further.

It should be clear and unambiguous that the only parameter here is the file system and so the only file descriptor we are talking about is what that file system becomes. Good point it is root and maybe the param should be renamed as such.

meanwhile this is internal docs. The actual parameter is described in ModuleConfig

codefromthecrypt:
OK I tried to re-iterate that the only parameter is the one we are talking about the FD for. good enough? cc @mathetake

mathetake (approved):

codefromthecrypt:
ok merging to budget time as we've a lot of docs to write and the ones end-users can see (wazero package, site) have the higher priority with first release in <1week.

That said, thanks for the help and def this is better!

PR #771, opened 2022-08-26
site: promotes languages and tinygo

codefromthecrypt:
We've had feedback the languages page is hard to find. We also notice
the majority of wazero users use TinyGo and probably no one uses
primarily `%.wat`. This redoes the home page to reflect that.

Note: This doesn't add a drop-down for language yet, as to do that we
need to fix mobile. Currently mobile concatenates all drop-downs, so it
would fill up the screen as we add new languages.

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -37,18 +37,17 @@ googleAnalytics = "G-LG7LD9G4ZB"
           name = "History"
           weight = 2
       [[languages.en.menu.main]]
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
here's the preview url https://deploy-preview-771--wazero.netlify.app/

mathetake (approved):

PR #772, opened 2022-08-29
filecache: include GOOS/GOARCH in the key

mathetake:
This allows us to avoid crashes with `GOARCH=amd64 make test` after running `make test` on M1 Mac


codefromthecrypt:
need go version? or are we ok

mathetake:
no need go version as the assembly independent of it :D

codefromthecrypt (approved):

PR #774, opened 2022-08-29
amd64: remove redundant ADD in GlobalGet/Set

mathetake:


codefromthecrypt (approved):
cool more efficient global gets should pay off as they are often used.

PR #775, opened 2022-08-29
Change OperationSwap to OperationSet

mathetake:
This commit changes `wazeroir.OperationSwap` to `OperationSet` which is the 
combination of OperationSwap and Drop in the previous implementation.

Previously, `OperationSwap` was always followed by `OperationDrop` on the swapped value on top.
Because of that, we had a redundant register allocation in Swap.

As a result, we use only one register in `OperationSet` which is a part of translations of
`local.tee` and `local.set`.

### DIFF
--- a/internal/wazeroir/compiler.go
+++ b/internal/wazeroir/compiler.go
@@ -826,15 +826,13 @@ operatorSwitch:
 			c.emit(
 				// +2 because we already popped the operands for this operation from the c.stack before
 				// called localDepth ^^,
-				&OperationSwap{Depth: depth + 2, IsTargetVector: isVector},
-				&OperationDrop{Depth: &InclusiveRange{Start: 0, End: 1}},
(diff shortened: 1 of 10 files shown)


internal/wazeroir/compiler.go
```diff
@@ -847,14 +845,12 @@ operatorSwitch:
 		if isVector {
 			c.emit(
 				&OperationPick{Depth: 1, IsTargetVector: isVector},
-				&OperationSwap{Depth: depth + 2, IsTargetVector: isVector},
-				&OperationDrop{Depth: &InclusiveRange{Start: 0, End: 1}},
```
mathetake, line 851: `OperationSwap` was always followed by Drop, and because of that, we had a redundant register allocation in Swap. That is what this commit resolves.

codefromthecrypt (approved):
thanks for the stellar PR desc

PR #781, opened 2022-08-30
compiler: hold stack base pointers in bytes, not index to the stack slice

mathetake:
To reduce the redundant instructions and register usage on the hot path.

### DIFF
--- a/internal/engine/compiler/impl_amd64.go
+++ b/internal/engine/compiler/impl_amd64.go
@@ -4608,7 +4608,7 @@ func (c *amd64Compiler) compileCallFunctionImpl(index wasm.Index, functionAddres
 
 		c.assembler.CompileRegisterToMemory(amd64.MOVQ, tmpRegister,
 			// "rb.1" is BELOW the top address. See the above example for detail.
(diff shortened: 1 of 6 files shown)


internal/engine/compiler/impl_amd64.go
```diff
@@ -4970,19 +4972,9 @@ func (c *amd64Compiler) compileReservedStackBasePointerInitialization() {
...
-
 	// next we move the base pointer (callEngine.stackBasePointer) to the tmp register.
-	c.assembler.CompileMemoryToRegister(amd64.MOVQ,
-		amd64ReservedRegisterForCallEngine, callEngineValueStackContextStackBasePointerOffset,
-		tmpReg,
-	)
-
-	c.assembler.CompileMemoryWithIndexToRegister(
-		amd64.LEAQ,
-		amd64ReservedRegisterForStackBasePointerAddress, 0, tmpReg, 8,
```
mathetake, line 4985: these two instructions are now consolidated to one ADDQ instruction, plus we don't need tmpReg anymore!

codefromthecrypt (approved):
thanks!

PR #782, opened 2022-08-30
compiler: simplify calling convention

mathetake:
This simplifies the calling convention, and consolidates the call frame stack
and value stack into a single stack.

As a result, the cost of function calls decreases because we now don't need
to check the boundary twice (value and call frame stacks) at each function call.

The following is the result of the bench mark for recursive fibonnaci
function in integration_test/bench/testdata/case.go, and it shows that
this actually improves the performance of function calls.

### amd64
```
$ benchstat fib_old.txt fib_new.txt
name                               old time/op  new time/op  delta
Invocation/compiler/fib_for_5-32    109ns ± 3%    81ns ± 1%  -25.86%  (p=0.008 n=5+5)
Invocation/compiler/fib_for_10-32   556ns ± 3%   473ns ± 3%  -14.99%  (p=0.008 n=5+5)
Invocation/compiler/fib_for_20-32  61.4µs ± 2%  55.9µs ± 5%   -8.98%  (p=0.008 n=5+5)
Invocation/compiler/fib_for_30-32  7.41ms ± 3%  6.83ms ± 3%   -7.90%  (p=0.008 n=5+5)
```

### arm64
```
$ benchstat old.txt new.txt                    
name                               old time/op    new time/op    delta
Invocation/compiler/fib_for_5-10     67.7ns ± 1%    60.2ns ± 1%  -11.12%  (p=0.000 n=9+9)
Invocation/compiler/fib_for_10-10     487ns ± 1%     460ns ± 0%   -5.56%  (p=0.000 n=10+9)
Invocation/compiler/fib_for_20-10    58.0µs ± 1%    54.3µs ± 1%   -6.38%  (p=0.000 n=10+10)
Invocation/compiler/fib_for_30-10    7.12ms ± 1%    6.67ms ± 1%   -6.31%  (p=0.000 n=10+9)
```

### DIFF
--- a/internal/engine/compiler/compiler_value_location.go
+++ b/internal/engine/compiler/compiler_value_location.go
@@ -388,8 +388,8 @@ func (v *runtimeValueLocationStack) getCallFrameLocations(sig *wasm.FunctionType
 	return v.stack[offset], v.stack[offset+1], v.stack[offset+2]
 }
 
-// pushCallFrame pushes a call frame's runtime locations onto the stack when the function call parameters
-// are already pushed there.
+// pushCallFrame pushes a call frame's runtime locations onto the stack assuming that
+// the function call parameters are already pushed there.
 //
 // See the diagram in callEngine.stack.
 func (v *runtimeValueLocationStack) pushCallFrame(callTargetFunctionType *wasm.FunctionType) (
--- a/internal/engine/compiler/compiler_value_location.go
+++ b/internal/engine/compiler/compiler_value_location.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/tetratelabs/wazero/internal/asm"
+	"github.com/tetratelabs/wazero/internal/wasm"
 )
 
 var (
@@ -14,10 +15,6 @@ var (
 	// unreservedVectorRegisters contains unreserved vector registers.
 	unreservedVectorRegisters []asm.Register
 )
-
-func isNilRegister(r asm.Register) bool {
-	return r == asm.NilRegister
-}
 
 func isGeneralPurposeRegister(r asm.Register) bool {
 	return unreservedGeneralPurposeRegisters[0] <= r && r <= unreservedGeneralPurposeRegisters[len(unreservedGeneralPurposeRegisters)-1]
@@ -212,7 +209,7 @@ func (v *runtimeValueLocationStack) push(loc *runtimeValueLocation) {
 	}
 	v.sp++
 	// stackPointerCeil must be set after sp is incremented since
-	// we skip the stack grow if len(valueStack) >= basePointer+stackPointerCeil.
+	// we skip the stack grow if len(stack) >= basePointer+stackPointerCeil.
 	if v.sp > v.stackPointerCeil {
 		v.stackPointerCeil = v.sp
 	}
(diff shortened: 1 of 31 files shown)


mathetake:
ok progress 20% 😄

mathetake:
ok I would say 50% :D

srenatus:
image
😄

mathetake:
ok now amd64 has started passing all tests in engine/compiler pkg

mathetake:
ok passed all tests on amd64!

mathetake:
cool clearly the function call gets faster! 

```
benchstat fib_old.txt fib_new.txt
name                               old time/op  new time/op  delta
Invocation/compiler/fib_for_5-32    109ns ± 3%    81ns ± 1%  -25.86%  (p=0.008 n=5+5)
Invocation/compiler/fib_for_10-32   556ns ± 3%   473ns ± 3%  -14.99%  (p=0.008 n=5+5)
Invocation/compiler/fib_for_20-32  61.4µs ± 2%  55.9µs ± 5%   -8.98%  (p=0.008 n=5+5)
Invocation/compiler/fib_for_30-32  7.41ms ± 3%  6.83ms ± 3%   -7.90%  (p=0.008 n=5+5)
```

^ is the bench result for
```go
func fibonacci(in uint32) uint32 {
	if in <= 1 {
		return in
	}
	return fibonacci(in-1) + fibonacci(in-2)
}

```

in integration_test/bench/testdata/case.go

mathetake:
cool done with arm64... now fuzzing phase!

internal/engine/compiler/impl_amd64.go
```diff
@@ -4487,220 +4443,122 @@ func (c *amd64Compiler) allocateRegister(t registerType) (reg asm.Register, err
...
+			loc.valueType = runtimeValueTypeI64
+		case wasm.ValueTypeF32:
+			loc.valueType = runtimeValueTypeF32
+		case wasm.ValueTypeF64:
+			loc.valueType = runtimeValueTypeF64
+		case wasm.ValueTypeV128:
+			loc.valueType = runtimeValueTypeV128Lo
+			hi := c.locationStack.pushRuntimeValueLocationOnStack()
+			hi.valueType = runtimeValueTypeV128Hi
+		}
```
codefromthecrypt, line 4560: check here and the rest of the diff switches that don't include ref types. add case: for them with a comment why we drop out, or panic with TODO or "unexpected" if there should never be any

internal/engine/compiler/impl_amd64.go
```diff
@@ -4717,118 +4575,64 @@ func (c *amd64Compiler) compileReturnFunction() error {
...
-
-	// Alias these free tmp registers for readability.
-	decrementedCallFrameStackPointerRegister, callFrameStackTopAddressRegister, tmpRegister := regs[0], regs[1], regs[2]
 
-	// Since we return from the function, we need to decrement the callframe stack pointer.
-	c.assembler.CompileNoneToMemory(amd64.DECQ, amd64ReservedRegisterForCallEngine, callEngineGlobalContextCallFrameStackPointerOffset)
+	returnAddress, callerStackBasePointerInBytes, callerFunction := c.locationStack.getCallFrameLocations(c.ir.Signature)
 
-	// next, get the decremented callframe stack pointer into decrementedCallFrameStackPointerRegister.
+	// If the return address is zero, meaning that we return from the execution.
```
codefromthecrypt, line 4591:
```suggestion
	// A zero return address means return from the execution.
```

internal/wazeroir/compiler_test.go
```diff
@@ -2754,9 +2754,167 @@ func TestCompile_select_vectors(t *testing.T) {
...
+	const v128 = wasm.ValueTypeV128
+	tests := []struct {
+		name                               string
+		sig                                *wasm.FunctionType
+		functionLocalTypes                 []wasm.ValueType
+		callFrameStackSizeInUint64         int
+		expLocalIndexToStackHeightInUint64 map[uint32]int
+	}{
+		{
+			name: "no function local, args>results",
```
codefromthecrypt, line 2774: maybe here and other matrixes add args=results?

codefromthecrypt (approved):
LGTM

I did a more thorough review, just to make sure docs are understandable by someone else. please read the comments from bottom to top instead of the other way.

One thing I'm noticing is that `callFrameDataSizeInUint64` is used in main and test code as it if it is a special number. I would recommend a variable if it is like a baseOffset or something, or a function, if there's a common calculation that better uses it. Otherwise, when people read it, they'll have to remember why `callFrameDataSizeInUint64` is special for a specific context.

internal/engine/compiler/compiler_value_location.go
```diff
@@ -343,3 +340,75 @@ func (v *runtimeValueLocationStack) takeStealTargetFromUsedRegister(tp registerT
...
+			loc.valueType = runtimeValueTypeI64
+		case wasm.ValueTypeF32:
+			loc.valueType = runtimeValueTypeF32
+		case wasm.ValueTypeF64:
+			loc.valueType = runtimeValueTypeF64
+		case wasm.ValueTypeV128:
+			loc.valueType = runtimeValueTypeV128Lo
+			hi := v.pushRuntimeValueLocationOnStack()
+			hi.valueType = runtimeValueTypeV128Hi
+		default:
```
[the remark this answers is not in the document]
mathetake: yes no ref

PR #783, opened 2022-08-30
compiler: zero-initializes moduleInstanceAddress of call engine

mathetake:
This fixes a bug introduced in #761 where the reuse of the api.Function
causes some trouble around module context initialization.

### DIFF
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -557,13 +557,12 @@ func (ce *callEngine) Call(ctx context.Context, callCtx *wasm.CallContext, param
 	// and we have to make sure that all the runtime errors, including the one happening inside
 	// host functions, will be captured as errors, not panics.
 	defer func() {
-		// If the module closed during the call, and the call didn't err for another reason, set an ExitError.
-		if err == nil {
-			err = callCtx.FailIfClosed()
-		}
-		// TODO: ^^ Will not fail if the function was imported from a closed module.
-
 		err = ce.deferredOnCall(recover())
+		if err == nil {
+			// If the module closed during the call, and the call didn't err for another reason, set an ExitError.
+			err = callCtx.FailIfClosed()
+			// TODO: ^^ Will not fail if the function was imported from a closed module.
+		}
 	}()
 
 	for _, v := range params {
@@ -575,7 +574,7 @@ func (ce *callEngine) Call(ctx context.Context, callCtx *wasm.CallContext, param
 }
 
 // deferredOnCall takes the recovered value `recovered`, and wraps it
-// with the call frame stack traces if v != nil. Also, reset
+// with the call frame stack traces when not nil. This also resets
 // the state of callEngine so that it can be used for the subsequent calls.
 //
 // This is defined for testability.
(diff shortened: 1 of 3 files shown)


internal/engine/compiler/engine.go
```diff
@@ -573,19 +574,23 @@ func (ce *callEngine) Call(ctx context.Context, callCtx *wasm.CallContext, param
...
+			def := ce.callFrameStack[ce.callFrameStackPointer-1-i].function.source.FunctionDefinition
+			builder.AddFrame(def.DebugName(), def.ParamTypes(), def.ResultTypes())
+		}
+		err = builder.FromRecovered(recovered)
 	}
-	err = builder.FromRecovered(v)
 
 	// Allows the reuse of CallEngine.
-	ce.stackPointer, ce.callFrameStackPointer = 0, 0
+	ce.stackBasePointerInBytes, ce.stackPointer, ce.callFrameStackPointer, ce.moduleInstanceAddress = 0, 0, 0, 0
```
mathetake, line 593: I should've reset `moduleInstanceAddress` as well in #761

mathetake:
will backfill more tests, but this should fix https://github.com/inkeliz/karmem/pull/85 and https://github.com/aquasecurity/trivy/pull/2791 @codefromthecrypt @inkeliz @knqyf263

mathetake:
ok verified that at least on my local machine, the PRs ^^ pass all tests. will do the backfilling

internal/engine/compiler/engine.go
```diff
@@ -573,19 +574,23 @@ func (ce *callEngine) Call(ctx context.Context, callCtx *wasm.CallContext, param
...
-// recoverOnCall takes the recovered value `recoverOnCall`, and wraps it
-// with the call frame stack traces. Also, reset the state of callEngine
-// so that it can be used for the subsequent calls.
-func (ce *callEngine) recoverOnCall(v interface{}) (err error) {
-	builder := wasmdebug.NewErrorBuilder()
-	for i := uint64(0); i < ce.callFrameStackPointer; i++ {
-		def := ce.callFrameStack[ce.callFrameStackPointer-1-i].function.source.FunctionDefinition
-		builder.AddFrame(def.DebugName(), def.ParamTypes(), def.ResultTypes())
+// deferredOnCall takes the recovered value `recovered`, and wraps it
+// with the call frame stack traces if v != nil. Also, reset
```
codefromthecrypt, line 578:
```suggestion
// with the call frame stack traces when not nil. This also resets
```

codefromthecrypt (approved):
Looks good once the usage pattern that led to this in trivy and karmem is backfilled as a test!

internal/integration_test/engine/adhoc_test.go
```diff
@@ -558,50 +558,53 @@ func testCloseInFlight(t *testing.T, r wazero.Runtime) {
 
 func testMemOps(t *testing.T, r wazero.Runtime) {
```
mathetake, line 559: so this was exactly the same pattern in trivy and karmen - calling exported functions which _might_ grow memory. But in order to catch this bug, I had to reuse the same api.Function, so I fixed it in that way.

mathetake: and verified that without the patch, this test fails!

PR #784, opened 2022-08-31
Moves host function imports into their own directory

codefromthecrypt:
Our root directory is getting crowded and it is also difficult to
organize the "host imports" concept due to this.

This moves common and language-specific imports into their own
directory. This will break go import signatures on the next release, but
is more sustainable overall.

Ex.
```diff
-	"github.com/tetratelabs/wazero/wasi_snapshot_preview1"
+	"github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1"
```


imports/README.md
```diff
@@ -0,0 +1,40 @@
+## wazero imports
```
codefromthecrypt, line 1: Here's a new document which was hard to place before this change.

Makefile
```diff
@@ -107,7 +102,7 @@ build.examples.emscripten: $(emscripten_sources)
 %/cat.wasm : cargo_target := wasm32-wasi
 
 .PHONY: build.examples.rust
-build.examples.rust: examples/allocation/rust/testdata/greet.wasm examples/wasi/testdata/cargo-wasi/cat.wasm
+build.examples.rust: examples/allocation/rust/testdata/greet.wasm imports/wasi_snapshot_preview1/example/testdata/cargo-wasi/cat.wasm
 
 # Builds rust using cargo normally, or cargo-wasi.
 %.wasm: %.rs
```
mathetake, line 108: not relevant to this PR, but I realized that `%.wasm` defined twice :D

mathetake (approved):

Makefile
codefromthecrypt: yeah good point

PR #791, opened 2022-09-01
Redirects users to tested example

codefromthecrypt:
Rather than clutter both the README and home page with details that can
drift and have caused maintenance, this directs users to the tested
basic example.

This also adds a FAQ entry about TinyGo's main, thanks to @basvanbeek
for the help.


mathetake (commented):
nit `examples/basic/testdata/add.wasm` is changed to 755 permission but do we need the execution bit?

mathetake (approved):

codefromthecrypt:
weird on chmod. fixed it!

codefromthecrypt:
fyi https://deploy-preview-791--wazero.netlify.app/

codefromthecrypt:
need to sort out the example test

Issue #789, opened 2022-08-31
Published version v1.0.0-beta1 takes precedence over v1.0.0-beta.2


ncruces:
I don't know what caused this, since I see no tag `v1.0.0-beta1`, but according to semver `v1.0.0-beta1` (notice the missing dot) will always take precedence over anything after a dot (like `v1.0.0-beta.2`).

https://pkg.go.dev/github.com/tetratelabs/wazero?tab=versions

ncruces:
The solution is to withdraw `v1.0.0-beta1` (which I don't know how to do) or to not use the dot going forward.

codefromthecrypt:
yeah that tag was an accident and go.mod didn't like it.. I'll look into seeing if there's a way to sort that out

codefromthecrypt:
https://go.dev/ref/mod#go-mod-file-retract will try this

codefromthecrypt:
later fixed up to 

```
// Wrong naming convention
retract v1.0.0-beta1

// Has memory bug and quickly fixed in v1.0.0-beta.2.
retract v1.0.0-beta.1
```

ncruces:
Note that #790 is not sufficient.

You now need to tag this retraction as `v1.0.0-beta.3` (and get it into the module cache), or it never gets "published", and we never get to see the retraction.

Hence the section in https://go.dev/ref/mod#go-mod-file-retract about a retraction retracting itself.
This… is complicated.

codefromthecrypt:
or maybe I won't need to say the exact beta after that.. let's try first

codefromthecrypt:
I guess in any case we have to wait for propagation to take place. Meanwhile, I'll keep #792 to ensure folks don't paste a command to get the broken tag.

codefromthecrypt:
beta 3 shows up but the other tag still does too :) https://pkg.go.dev/github.com/tetratelabs/wazero?tab=versions

codefromthecrypt:
oh... I hope it isn't what I think it is, which is that the redacted version will stay until a release has a semver higher than it..

codefromthecrypt:
ok so problem is `beta` is dead. we can't really use the broken version pattern because if we do, it is invalid from go's mod version syntax.

We need to switch to a word lexicographically later than `beta`

codefromthecrypt:
[...]

PR #792, opened 2022-09-01
Switches from "beta" to "pre"

codefromthecrypt:
This switches our pre-release nomenclature from "beta" to pre-release
("pre") as we are otherwise unable to rank higher than a broken tag and
still be compatible with go mod's naming conventions (semver).

See https://go.dev/blog/module-mirror-launch

Fixes https://github.com/tetratelabs/wazero/issues/789


mathetake (approved):
so start with v1.0.0-pre.1, right

PR #793, opened 2022-09-01
languages/go: fix wasmexport proposal link

srenatus:
There're no branches or tags on that repo. Must have slipped in accidentally.


mathetake (approved):
thanks @srenatus! (glad to see you again here in wazero repo :D)

srenatus:
Those Go+Wasm notes are a very thorough, thanks for compiling and sharing them 🎈

PR #795, opened 2022-09-06
Adds gojs benchmark

codefromthecrypt:
As noted in the godoc, this is not because we expect it to be fast,
rather for folks doing internal work to quickly tell if something might
help or not.

### DIFF
--- a/imports/go/example/stars_test.go
+++ b/imports/go/example/stars_test.go
@@ -1,16 +1,20 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"os"
 	"os/exec"
 	"path"
 	"testing"
-	"time"
 
+	"github.com/tetratelabs/wazero"
+	gojs "github.com/tetratelabs/wazero/imports/go"
+	"github.com/tetratelabs/wazero/internal/platform"
 	"github.com/tetratelabs/wazero/internal/testing/maintester"
 	"github.com/tetratelabs/wazero/internal/testing/require"
+	"github.com/tetratelabs/wazero/sys"
 )
 
 // Test_main ensures the following will work:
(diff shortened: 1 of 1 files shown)


imports/go/example/stars_test.go
```diff
@@ -56,3 +61,50 @@ func compileFromGo() error {
...
+	compiled, err := r.CompileModule(ctx, bin, wazero.NewCompileConfig())
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	// Instead of making real HTTP calls, return fake data.
+	ctx = gojs.WithRoundTripper(ctx, &fakeGitHub{})
+	cfg := wazero.NewModuleConfig()
+
+	b.Run("CompileModule", func(b *testing.B) {
```
mathetake, line 100: CompileModule?
```suggestion
	b.Run("Run", func(b *testing.B) {
```

mathetake (approved):

imports/go/example/stars_test.go
codefromthecrypt: renamed!

codefromthecrypt:
I added this, but it didn't show any differences, but that's ok as I was just curious. Here are the commands for the next time!

```bash
# old branch
for i in 1 2 3 4 5 ; do go test -run=NONE -bench=Benchmark_main/gojs.Run github.com/tetratelabs/wazero/imports/go/example >> old.txt; done
# new branch
for i in 1 2 3 4 5 ; do go test -run=NONE -bench=Benchmark_main/gojs.Run github.com/tetratelabs/wazero/imports/go/example >> new.txt; done

# install benchstat
go get golang.org/x/perf/cmd/benchstat
go install golang.org/x/perf/cmd/benchstat
# compare results
benchstat old.txt new.txt
```

PR #798, opened 2022-09-06
Add private methods to config interfaces to prevent compilation of cu…

anuraaga:
…stom implementations

@jcchavezs pointed out this idea in my coraza config exploration, curious what do you think of it? Unfortunately it doesn't seem to be without compromise, the godoc gets a mysterious line at the bottom

image

### DIFF
--- a/config.go
+++ b/config.go
@@ -140,6 +140,10 @@ type RuntimeConfig interface {
 	//
 	// See https://www.w3.org/TR/2022/WD-wasm-core-2-20220419/
 	WithWasmCore2() RuntimeConfig
+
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
TL'DR' I think all our interfaces should have a doc on them like where not already:

```
// # Notes
//
//   - This is an interface for decoupling, not third-party implementations. All implementations are in wazero.
```

I think this and a runtime fail is enough, because doing something odd like this will only work when they are in the same package, and major interfaces like `api.Module` are not in the same package as the Implementation. IOTW, I think this is clever, but not really solving the whole problem, and has as you notice some awkward side effects. In summary, maybe trying too hard.

anuraaga:
Cool, sounds reasonable

PR #800, opened 2022-09-06
Top-levels CoreFeatures and defaults to 2.0

codefromthecrypt:
While compilers should be conservative when targeting WebAssembly Core
features, runtimes should be lenient as otherwise people need to
constantly turn on all features. Currently, most examples have to turn
on 2.0 features because compilers such as AssemblyScript and TinyGo use
them by default. This matches the policy with the reality, and should
make first time use easier.

This top-levels an internal type as `api.CoreFeatures` and defaults to
2.0 as opposed to 1.0, our previous default. This is less cluttered than
the excess of `WithXXX` methods we had prior to implementing all
planned WebAssembly Core Specification 1.0 features.

Finally, this backfills rationale as flat config types were a distinct
decision even if feature set selection muddied the topic.


codefromthecrypt:
one interesting thing in the diff, is that this allows defaults to work for AssemblyScript, Go, TinyGo and Zig

mathetake (approved):
Love turning on 2.0 by default!

codefromthecrypt:
passerby's should note that we still individually test specs of both versions. turning this default doesn't risk breaking that.

PR #803, opened 2022-09-08
Delete buildoptions pkg

mathetake:
This has existed since the beginning, but I have never used recently 
since it's used everywhere and stdout gets pretty messy if enabled.


codefromthecrypt (approved):
yep

PR #806, opened 2022-09-14
site: adds instructions on how to hack go

codefromthecrypt:
One reality of wasm compilers is that they are incomplete and often have bugs. While I don't want our site to become overwhelmed with instructions, it is helpful I think to have tips on how to do routine changes to the compiler, at least how to test them.

This is for Go, and I'll also add one on TinyGo as in both cases, it might not be obvious that you need may need to rebuild the binary in order to test changes, and if you do it may be difficult to figure out how.

### DIFF
--- a/site/content/languages/go.md
+++ b/site/content/languages/go.md
@@ -241,6 +241,52 @@ signature unrelated to the source, more care is needed implementing the host
 side, to ensure the proper count of parameters are read and results written to
 the Go stack.
 
+## Hacking
+
+If you run into an issue where you need to change Go's sourcecode, the first
+thing you should do is read the [contributing guide][20], which details how to
+confirm an issue exists and a fix would be accepted. Assuming they say get a
+yes, the next step is to ensure you can build and test go.
+
+### Setting up your GOROOT
+
+First, clone upstream or your fork of golang/go and make a branch off `master`
+for your work, as GitHub pull requests are against that branch.
+
+```bash
+$ git clone https://github.com/golang/go.git
+$ cd go
+$ export GOROOT=$PWD
+$ git checkout -b my-fix
+```
+
+### Build a branch-specific `go` binary
+
+While your change may not affect the go binary itself, there are checks inside
+go that require version matching. Build a go binary from source to avoid these:
+```bash
+$ cd src
+$ GOOS=js GOARCH=wasm ./make.bash
+Building Go cmd/dist using /usr/local/go. (go1.19 darwin/amd64)
+Building Go toolchain1 using /usr/local/go.
+--snip--
+$ cd ..
+$ bin/go version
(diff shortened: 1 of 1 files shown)


site/content/languages/go.md
```diff
@@ -241,6 +241,52 @@ signature unrelated to the source, more care is needed implementing the host
...
+$ git checkout -b my-fix
+```
+
+### Build a branch-specific `go` binary
+
+While your change may not affect the go binary itself, there are checks inside
+go that require version matching. Build a go binary from source to avoid these:
+```bash
+$ cd src
+$ GOOS=js GOARCH=wasm ./make.bash
```
anuraaga, line 269: Do these variables work? They look like they would create a compiler that runs in the browser, not on the computer

codefromthecrypt: PTAL redid it

anuraaga: Cool - also found `GOHOSTARCH` so makes sense now https://github.com/golang/go/blob/master/src/make.bash#L13

anuraaga (approved):

codefromthecrypt:
@anuraaga thanks for the tip. I think the above works. do you mind running it on your own?

site/content/languages/go.md
```diff
@@ -241,6 +241,73 @@ signature unrelated to the source, more care is needed implementing the host
...
+confirm an issue exists and a fix would be accepted. Assuming they say yes, the
+next step is to ensure you can build and test go.
+
+### Make a branch for your changes
+
+First, clone upstream or your fork of golang/go and make a branch off `master`
+for your work, as GitHub pull requests are against that branch.
+
+```bash
+$ git clone https://github.com/golang/go.git
```
anuraaga, line 257: Consider `--depth=1`

site/content/languages/go.md
```diff
@@ -241,6 +241,73 @@ signature unrelated to the source, more care is needed implementing the host
...
+```
+
+### Build a branch-specific `go` binary
+
+While your change may not affect the go binary itself, there are checks inside
+go that require version matching. Build a go binary from source to avoid these:
+
+```bash
+$ cd src
+$ GOOS=js GOARCH=wasm ./make.bash
```
anuraaga, line 269: Perhaps it differs too much from the norm to suggest, but JFYI, I didn't have xcode tools set up yet it seems, but following the error message's suggestion of `CGO_ENABLED=0` worked, and the below `syscall/js` test ran fine. But I don't know what problems it might cause in other development (race detector uses cgo so won't work with this build, not sure what else).

anuraaga (approved):
Commands worked for me

site/content/languages/go.md
codefromthecrypt: just in case it helps others, this and other things messed up for me today when I updated OS/x until I did `xcodebuild -runFirstLaunch`

codefromthecrypt:
thanks for the help @anuraaga!

PR #813, opened 2022-09-28
Initial wazero CLI to run a standalone Wasm binary

anuraaga:
This is a first iteration of a wazero CLI for running Wasm binaries. Future iterations will need to add more options, such as mapping filesystem paths, but for now this just provides args, std streams, rand, time.


Makefile
```diff
@@ -182,16 +182,11 @@ spectest.v1:
...
-
-$(golangci_lint_path):
-	@go install $(golangci_lint)
-
 golangci_lint_goarch ?= $(shell go env GOARCH)
 
 .PHONY: lint
 lint: $(golangci_lint_path)
-	@GOARCH=$(golangci_lint_goarch) CGO_ENABLED=0 $(golangci_lint_path) run --timeout 5m
+	@GOARCH=$(golangci_lint_goarch) CGO_ENABLED=0 go run $(golangci_lint) run --timeout 5m
```
anuraaga, line 189: I had some incompatible version installed. Using `go run` avoids interaction with whatever's already installed

wazero/testdata/wasi_arg.wat
```diff
@@ -0,0 +1,61 @@
+;; $wasi_arg is a WASI command which copies null-terminated args to stdout.
```
anuraaga, line 1: Copied as is from wasi test, let me know if this makes sense

wazero/wazero.go
```diff
@@ -0,0 +1,94 @@
+package main
```
anuraaga, line 1: The folder is named `wazero`, from what I understand it is the way to allow `go install` to install a binary named `wazero`. It does mean `github.com/tetratelabs/wazero/wazero` though, stuttering but don't know if there's another way to support `go install` but may have missed it

Makefile
anuraaga: Reverted for now though since had bad interaction with CI

examples/cli/cli_test.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+	}
+
+	wasmPath := filepath.Join(t.TempDir(), "cli.wasm")
+	require.NoError(t, os.WriteFile(wasmPath, cliWasm, 0755))
+
+	for _, tc := range tests {
+		tt := tc
+		t.Run(strings.Join(tt.args, " "), func(t *testing.T) {
+			goCmd := filepath.Join(runtime.GOROOT(), "bin", "go")
```
anuraaga, line 40: Oops guess this strategy won't work for the docker tests. I can try disabling these in the docker tests or something else after a first review pass

wazero/wazero.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	wasmExe := filepath.Base(wasmPath)
+
+	ctx := context.Background()
+	rt := wazero.NewRuntime(ctx)
+	defer rt.Close(ctx)
+
+	// Because we are running a binary directly rather than embedding in an application,
+	// we default to wiring up standard streams and args.
```
codefromthecrypt, line 64: do we want to plumb time and RNG? or wait till someone asks.

codefromthecrypt (approved):
nice thanks for getting this started cc @dgryski who was the last, but not the first, to ask for this.

examples/cli/testdata/cli.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func main() {
+	var sub bool
+	flag.BoolVar(&sub, "sub", false, "whether to subtract arguments instead of add")
+
+	flag.Parse()
+
+	if flag.NArg() < 2 {
+		fmt.Println("bad arguments")
```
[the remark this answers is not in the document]
anuraaga: Neat trick

wazero/README.md
```diff
@@ -0,0 +1,21 @@
...
+### Usage
+
+The wazero CLI accepts a single argument, the path to a WebAssembly binary.
+Arguments can be passed to the WebAssembly binary itself after `--`.
+
+```bash
+wazero calc.wasm -- 1 + 2
+```
+
+In addition to arguments, the WebAssembly binary has access to stdout, stderr,
```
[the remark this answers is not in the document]
anuraaga: Yeah switched to `run` think it's good to do that now

wazero/testdata/wasi_arg.wat
```diff
@@ -0,0 +1,61 @@
...
+(module $wasi_arg
```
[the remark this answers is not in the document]
anuraaga: I think that would prevent working in Docker without more tweaks.

Would it make sense to map the repo into a folder in the docker test to allow readfile with relative paths to work?

wazero/wazero.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		fmt.Fprintf(stdErr, "error reading wasm binary: %v\n", err)
+		printUsage(stdErr)
+		exit(1)
+	}
+
+	wasmExe := filepath.Base(wasmPath)
+
+	ctx := context.Background()
+	rt := wazero.NewRuntime(ctx)
```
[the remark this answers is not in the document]
anuraaga: Think you already did that :)

https://github.com/tetratelabs/wazero/blob/main/config.go#L30

wazero/wazero.go
anuraaga: Good call think it's good to plumb them now.

anuraaga:
Thanks @codefromthecrypt think I got everything except for https://github.com/tetratelabs/wazero/pull/813#discussion_r982042772

wazero/testdata/wasi_arg.wat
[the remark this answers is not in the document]
codefromthecrypt: I see what you mean. let's drop this comment, as ex I use lima and mounts don't work :D

wazero/testdata/wasi_arg.wat
codefromthecrypt: maybe just add a comment that this is a copy of (where)

wazero/wazero.go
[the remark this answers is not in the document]
codefromthecrypt: /me whistles

wazero/testdata/wasi_arg.wat
anuraaga: Sure

codefromthecrypt (approved):
We probably have a decision after the fact, which is whether or not to add binary release config (like nfpm etc, like what we have in func-e or car) on the next one, or wait for folks to ask. Regardless, it would be after this.

PR #815, opened 2022-09-28
Replaces MemorySizer and CompileConfig with RuntimeConfig

codefromthecrypt:
We formerly introduced `MemorySizer` as a way to control capacity independently of size. This was the first and only feature in `CompileConfig`. While possibly used privately, `MemorySizer` has never been used in public GitHub code.

These APIs interfere with how we do caching of compiled modules. Notably, they can change the min or max defined in wasm, which invalidates some constants. This has also had a bad experience, forcing everyone to boilerplate`wazero.NewCompileConfig()` despite that API never being used in open source.

This addresses the use cases in a different way, by moving configuration to `RuntimeConfig` instead. This allows us to remove `MemorySizer` and `CompileConfig`, and the problems with them, yet still retaining functionality in case someone uses it.

* `RuntimeConfig.WithMemoryLimitPages(uint32)`: Prevents memory from growing to 4GB (spec limit) per instance.
  * This works regardless of whether the wasm encodes max or not. If there is no max, it becomes effectively this value.
* `RuntimeConfig.WithMemoryCapacityFromMax(bool)`: Prevents reallocations (when growing).
  * Wasm that never sets max will grow from min to the limit above.

Note: Those who want to change their wasm (ex insert a max where there was none), have to do that externally, ex via compiler settings or post-build transformations such as [wabin](https://github.com/tetratelabs/wabin)

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -280,8 +280,6 @@ space. It is accepted that the options pattern is common in Go, which is the mai
 wazero's configuration types cover the three main scopes of WebAssembly use:
 * `RuntimeConfig`: This is the broadest scope, so applies also to compilation
   and instantiation. Ex. This controls the WebAssembly Specification Version.
-* `CompileConfig`: This affects any compilation related concerns not defined in
-  the binary format. Ex. This controls the allocation of linear memory.
 * `ModuleConfig`: This affects modules instantiated after compilation and what
   resources are allowed. Ex. This defines how or if STDOUT is captured.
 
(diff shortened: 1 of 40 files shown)


anuraaga (approved):

codefromthecrypt:
rewrote the description based on feedback from @inkeliz which clarified the original intents.

codefromthecrypt:
thanks @inkeliz for helping clarify and @anuraaga for the review

PR #816, opened 2022-09-29
cli: add mount option to expose paths to wasm app

anuraaga:
This is mostly implemented but missing unit tests for the mount flag itself (there is an e2e test using `go run` which doesn't allow step debugging).

I'm wondering between

- Write another `.wat` file using filesystem
- Add `testdata.go` that embeds files in an importable way to reuse other test cases without copying (would uncopy `wasi_arg.wat` if doing that)

Also the mount syntax is similar to docker's `volume`. For reference, I found docker seems to not recommend that flag anymore for a more complicated one

https://docs.docker.com/engine/reference/commandline/run/#add-bind-mounts-or-volumes-using-the---mount-flag

Without a difference between bind/volume there may not be as much need for key/value pair type complicated syntax here, but it would be a way to fix the colon issue. Another option in mind is to allow escaping `\:`


codefromthecrypt:
so far so good. I'm not going to cut release until this is in, even if we are a day after the month turns

anuraaga:
Was able to verify with TinyGo's wasi unit tests, all pass except for some that create temp files as wazero will support that later. Did run across a hairy issue which also stems from the `fs.FS` abstraction I think

https://github.com/tetratelabs/wazero/pull/816/files#diff-178d93f32aae083cbe1e7913856b8938f9891ca91382eabd678628c1eb81ce84R16

While this was needed for TinyGo's unit tests I can forbid `.` as a mount for now until that's supported if it makes sense.

Also let me know any suggestion on the unit test question in https://github.com/tetratelabs/wazero/pull/816#issue-1390142390

anuraaga:
Think this is finally ready

codefromthecrypt (approved):
splendid!

PR #817, opened 2022-09-29
Adds MemoryDefinition to CompiledModule and Memory

codefromthecrypt:
It is more often the case that projects are enabling a freestanding target, and that may or may not have an exporting memory depending on how that's interpreted. This adds the ability to inspect memories similar to how you can already inspect compiled code prior to instantiation. For example, you can enforce an ABI constraint that "memory" must be exported even if WASI is not in use.

Here's an example of validating both a function and a memory are exported prior to instantiating the guest.
```go
const (
	FuncHandle = "handle"
	Memory = "memory"
)

// validateGuestModule ensures the guest exports features the host ABI requires.
func validateGuestModule(guest wazero.CompiledModule) error {
	if handle, ok := guest.ExportedFunctions()[FuncHandle]; !ok {
		return fmt.Errorf("wasm: guest doesn't export func[%s]", FuncHandle)
	} else if len(handle.ParamTypes()) != 0 || len(handle.ResultTypes()) != 0 {
		return fmt.Errorf("wasm: guest exports the wrong signature for func[%s]. should be nullary", FuncHandle)
	} else if _, ok = guest.ExportedMemories()[Memory]; !ok {
		return fmt.Errorf("wasm: guest doesn't export memory[%s]", Memory)
	}
	return nil
}
```

See https://github.com/tinygo-org/tinygo/pull/3072

cc @hunjixin

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -179,20 +179,56 @@ type Closer interface {
 	Close(context.Context) error
 }
 
-// FunctionDefinition is a WebAssembly function exported in a module (wazero.CompiledModule).
+// ExportDefinition is a WebAssembly type exported in a module
+// (wazero.CompiledModule).
 //
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#exports%E2%91%A0
-type FunctionDefinition interface {
+type ExportDefinition interface {
 	// ModuleName is the possibly empty name of the module defining this
-	// function.
+	// export.
 	//
 	// Note: This may be different from Module.Name, because a compiled module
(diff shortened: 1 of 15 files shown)


codefromthecrypt:
@hunjixin glad! The main thing is that the idea of "freestanding" isn't standard :) so for example, someone else might decide to do no "memory" or export as "mem" instead. This feature allows the embedder to verify memory is exported especially in cases where there's no formal ABI. Otherwise, errors would wait until runtime and be harder to troubleshoot.

PR #821, opened 2022-10-04
Use repo-scoped image for internal build image

anuraaga:
I randomly noticed this action start failing on my fork after I pushed main to it recently to verify some other workflows. Having repository-scoped ghcr images is relatively recent so maybe this can be migrated to it - it would allow forks to behave the same as this repo. Or if intentionally using global don't have to make this change and I could send another PR with an `if repo == tetratelabe/wazero` condition to the job instead


codefromthecrypt (approved):
worth a shot!

(ps func-e has some like this too)

anuraaga:
Oops thanks - worth a shot actually reminded me to run on my fork.

https://github.com/anuraaga/wazero/actions/runs/3179525557/jobs/5182111405

Issue #823, opened 2022-10-17
Remove wasm.CallGoFunc use reflect.Value


hafeidejiangyou:
**Is your feature request related to a problem? Please describe.**
When I try wazero, I found wasm.CallGoFunc takes a lot of CPU in my case with BenchMarks,  It is show a lot of time is spent on reflections, so I want to remove the reflective call of the host function.

**Describe the solution you'd like**
To remove reflections, the host function signature must be same. And we can use uint64 slice or interface{} slice as the param type, so the step will be:
1. add host function with Module, it will be like this
``` go
_, err := r.NewHostModuleBuilder("env").
ExportFunction("log_i32", func(params []uint64) ([]uint64, error){
	fmt.Println("log_i32 >>", int32(params[0]))
	return nil, nil
}).
ExportFunction("current_year", func(_ []uint64) ([]uint64, error) {
	if envYear, err := strconv.ParseUint(os.Getenv("CURRENT_YEAR"), 10, 64); err == nil {
		return []uint64{uint64(uint32(envYear))}, nil // Allow env-override to prevent annual test maintenance!
	}
	return []uint64{uint64(uint32(time.Now().Year()))}, nil
}).
Instantiate(ctx, r)
```

2. when compile, use reflect to judge the FunctionKind for whether it contain `context.Context` or `apt.Module`
3. when call wasm.CallGoFunc, get the FunctionKind, and use the `params []uint64` as one of the call param
4. call host function 

**Describe alternatives you've considered**
This raw solution can not check the signature of host function, It can be point out when write export function by user.

**Additional context**
-

codefromthecrypt:
Thanks for raising this issue @hafeidejiangyou. I'm also wondering if we can do something about this. Perhaps we can attempt to cast the input to `func([]uint64) ([]uint64, error)` and skip reflection if so? It might require a different code path.

Would you be willing to try to prove the performance gain by making modifications on your fork in a way that the current and the optimized/raw approach can co-exist? Then update the below bench to show the results?

https://github.com/tetratelabs/wazero/blob/b01effc8a91f624c51ce09698d964b0b5b77819f/internal/integration_test/bench/hostfunc_bench_test.go

Note that this experiment is mainly to prove the point, and might need to change a little more if we decide to go forward on it.

codefromthecrypt:
[...]

PR #825, opened 2022-10-20
HostFunc: remove reflect for host function call

hafeidejiangyou:
This PR remove reflect for wasm host function call, the benchmark is:

**Before：**
``` 
goos: darwin
goarch: amd64
BenchmarkHostFunctionCall
BenchmarkHostFunctionCall/call_go_host
BenchmarkHostFunctionCall/call_go_host-8                  939739              1262 ns/op
BenchmarkHostFunctionCall/call_wasm_host
BenchmarkHostFunctionCall/call_wasm_host-8              22812769                48.90 ns/op
PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/bench   2.658s
```

**After:**

```
goos: darwin
goarch: amd64
pkg: github.com/tetratelabs/wazero/examples/import-go
BenchmarkHostFunctionCall
BenchmarkHostFunctionCall/call_go_host
BenchmarkHostFunctionCall/call_go_host-8                 9719632               125.6 ns/op
BenchmarkHostFunctionCall/call_wasm_host
BenchmarkHostFunctionCall/call_wasm_host-8              24260326                48.20 ns/op
PASS
ok      github.com/tetratelabs/wazero/examples/import-go        2.988s
```

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"reflect"
 )
 
 // ExternType classifies imports and exports with their respective types.
@@ -102,6 +101,13 @@ const (
 	// Note: The usage of this type is toggled with api.CoreFeatureBulkMemoryOperations.
 	ValueTypeExternref ValueType = 0x6f
 )
+
+// HostFuncSignature is the signature of host function
+type HostFuncSignature struct {
+	Fn     func(context.Context, Module, ...uint64) ([]uint64, error)
+	NumIn  []ValueType
+	NumOut []ValueType
+}
 
 // ValueTypeName returns the type name of the given ValueType as a string.
 // These type names match the names used in the WebAssembly text format.
@@ -260,7 +266,7 @@ type FunctionDefinition interface {
 	// different from its defining module.
 	//
 	// See https://www.w3.org/TR/wasm-core-1/#host-functions%E2%91%A0
-	GoFunc() *reflect.Value
+	GoFunc() *HostFuncSignature
 
 	// ParamTypes are the possibly empty sequence of value types accepted by a
 	// function with this signature.
--- a/examples/import-go/age-calculator.go
+++ b/examples/import-go/age-calculator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"github.com/tetratelabs/wazero/api"
 	"log"
 	"os"
 	"strconv"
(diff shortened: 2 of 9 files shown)


hafeidejiangyou:
#823

codefromthecrypt:
Thanks I see what you are up to. Are you interested in proceeding? If so, I can wait until it is all compiling and then perform review and/or help polish.

hafeidejiangyou:
OK, I will fix all the test first.

codefromthecrypt:
ps my advice would be to implement HostFuncSignature with reflection by default. Ex if the input function is not a HostFuncSignature, then make a HostFuncSignature which uses the existing reflective logic.

hafeidejiangyou:
@codefromthecrypt  You mean support two way to write host function at the same time? 

>  if the input function is not a HostFuncSignature, then make a HostFuncSignature which uses the existing reflective logic.

make a HostFuncSignature for call the non HostFuncSignature function is ok, but it will use reflect to do the final call.
In this way, the benefit is user can write host function with more go type not just uint64, is my understanding correct?

codefromthecrypt:
@hafeidejiangyou let's call the type `api.HostFunc`

ex.
```go
package api

type HostFunc func(ctx context.Context, mod api.Module, in ...uint64) ([]uint64, error)
```

We then move the field `wasm.Code.GoFunc` to `wasm.HostFunc` and add `wasm.Code.HostFunc` as the above type.
The main idea is to make sure all code paths don't use reflection.

We add `wasm.HostFunc.GoFunc` for transition purposes explained below

We add a `HostFuncBuilder` which collects this and the metadata it needs.
```go
package wazero

type HostFuncBuilder interface {
	WithParams(api.ValueType...) HostFuncBuilder

	WithResults(api.ValueType...) HostFuncBuilder

	Build() interface{} // or any, but not the internal type.
}

type hostFuncBuilder struct {
	fn api.HostFunc
	params, values []api.ValueType
}

func NewHostFuncExporter(fn HostFunc) {
--snip--

func (b *hostFuncBuilder) Build() {
	return &wasm.HostFunc{Code: wasm.Code{HostFunc: b.fn ...} // note the new field HostFunc
}
```

HostModuleBuilder is left alone.. Let's take example of two calls to  `HostModuleBuilder.ExportFunction`
```
		ExportFunction("add_1_reflect", func(in uint32) uint32 {
				return in+1
			}).
		ExportFunction("add_1_hostfunc", wazero.NewHostFuncBuilder(
			func(_context.Context,_ api.Module, params ...uint64) ([]uint64, error) {
				return []uint64{params[0]+1}, nil
			}).
                        WithParams(api.ValueTypeI32).
                        WithResults(api.ValueTypeI32).
                        Build()).
```

So HostModuleBuilder.Build() ends up with two entries in the map: 

* "add_1_reflect": the anonymous func instance
* "add_1_hostfunc": an instance of wasm.HostFunc which has a new field of `api.HostFunc` set.
  * Note: we already have internal code that add  `wasm.HostFunc` to the builder. ex assemblyscript!

So all the same logic is used... this map is passed to `wasm.NewHostModule` same as now, but the logic in `addFuncs` changes.

Where it has this.. add something to adapt as necessary
```
if hf, ok := v.(*HostFunc); ok {
	if gf := hf.GoFunc; gf != nil {
		hf.Code.HostFunc = parseGoFunc(gf)
		...
		hf.GoFunc = nil // done
```

The main thing is to change the logic here to make sure that if HostFunc.GoFunc is present it is adapted to an api.GoFunc (using reflection), but making sure the engine is agnostic to this.

codefromthecrypt:
@hafeidejiangyou lemme know by tomorrow if you are up to finishing this. I have a really long flight so might go for it otherwise.

hafeidejiangyou:
@codefromthecrypt  I think you have wrote almost all the code, to reduce communication costs and review costs, I think you can modify it directly, if you think it is needed.

codefromthecrypt:
@hafeidejiangyou cool in any case I appreciate your efforts bringing this forward. there's still time to improve it before our monthly release!

PR #826, opened 2022-10-21
Adds test to ensure custom contexts are usable

hafeidejiangyou:
I want to hold the my Context and use it in the host funtion call, So I add a holder funtion to hold the Context instance.
I think this is a feature, Please review it.

### DIFF
--- a/runtime_test.go
+++ b/runtime_test.go
@@ -5,6 +5,7 @@ import (
 	_ "embed"
 	"errors"
 	"testing"
+	"time"
 
 	"github.com/tetratelabs/wazero/api"
 	"github.com/tetratelabs/wazero/internal/leb128"
@@ -20,6 +21,21 @@ var (
 	// testCtx is an arbitrary, non-default context. Non-nil also prevents linter errors.
 	testCtx = context.WithValue(context.Background(), struct{}{}, "arbitrary")
 )
+
+var _ context.Context = &HostContext{}
+
+// HostContext contain the content will be used in host function call
+type HostContext struct {
+	Content string
+}
+
+func (h *HostContext) Deadline() (deadline time.Time, ok bool) { return }
+
+func (h *HostContext) Done() <-chan struct{} { return nil }
+
+func (h *HostContext) Err() error { return nil }
+
+func (h *HostContext) Value(key interface{}) interface{} { return nil }
 
 func TestNewRuntimeWithConfig_version(t *testing.T) {
 	cfg := NewRuntimeConfig().(*runtimeConfig)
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
Hi! because context is already propagated, you can use normal context keys instead to add state for host function calls. That's how we do it internally already for host imports like wasi.


First define the context state you need.
```go
// requestStateKey is a context.Context value associated with a requestState
// pointer to the current request.
type requestStateKey struct{}

type requestState struct {
  magic uint64
}
```

Next, define your host function using this state
```go
func getMagic(ctx context.Context) uint64 {
	return requestStateFromContext(ctx).magic
}

func requestStateFromContext(ctx context.Context) *requestState {
	return ctx.Value(requestStateKey{}).(*requestState)
}
```

Finally, when something calls in, pass the same context
```go
	 ctx := context.WithValue(context.Background(), requestStateKey{}, s)
     	_, err = mod.ExportedFunction("get_magic").Call(ctx)
```

hafeidejiangyou:
I had try, benchmark show use context.Context is so slow in host function call, so I change it.

codefromthecrypt:
@hafeidejiangyou I think if you want a faster way, you can find a faster one or make your own context. We don't decorate the context internally in any way.

Ex. Define a custom context.
```
var _ context.Context = &MyContext{}

type MyContext struct {
	magic uint64
}

func (c *MyContext) Deadline() (deadline time.Time, ok bool) {
	return
}

func (c *MyContext) Err() (err error) {
	return 
}

func (c *MyContext) Value(any)(val any) {
	return
}

func (f *fastCtx) Done() (done <-chan struct{}) {
	return
}
```

Use it directly in your functions:
```
func getMagic(ctx context.Context) uint64 {
	return ctx.(*MyContext).magic
}
```

I think this is the best way to improve performance vs making a new type to pass around.

hafeidejiangyou:
OK, I agree it is a better way. 
Does the context propagate will be intended for other uses when `ExportedFunction.Call(ctx, args)` is called？ I mean the `ctx` passed to `Call` can not be decorate like `newCtx = context.WithValue(ctx, "hello", "World")`, because it will panic with cast `ctx.(*MyContext)` in host function

codefromthecrypt:
@hafeidejiangyou we currently don't decorate it, but of course if you are making a shared library and they decorate externally that is another issue. If you want to make sure this works, maybe change your PR to a runtime_test that does exactly this. this way, it will break our build if any problem later.

hafeidejiangyou:
Ok, I will add a test for it.

runtime_test.go
```diff
@@ -533,6 +549,75 @@ func TestRuntime_CloseWithExitCode(t *testing.T) {
 	}
 }
 
+func TestCallHostWithSelfImplContext(t *testing.T) {
```
codefromthecrypt, line 552:
```suggestion
func TestHostFunctionWithCustomContext(t *testing.T) {
```

codefromthecrypt:
thanks for this!

PR #828, opened 2022-10-27
Adds HostFunctionBuilder to enable high performance host functions

codefromthecrypt:
This PR follows @hafeidejiangyou advice to not only enable end users to
avoid reflection when calling host functions, but also use that approach
ourselves internally. The performance results are staggering and will be
noticable in high performance applications.

Before
```
BenchmarkHostCall/Call
BenchmarkHostCall/Call-16            	 1000000	      1050 ns/op
Benchmark_EnvironGet/environGet
Benchmark_EnvironGet/environGet-16         	  525492	      2224 ns/op
```

Now
```
BenchmarkHostCall/Call
BenchmarkHostCall/Call-16            	14807203	        83.22 ns/op
Benchmark_EnvironGet/environGet
Benchmark_EnvironGet/environGet-16         	  951690	      1054 ns/op
```

To accomplish this, this PR consolidates code around host function
definition and enables a fast path for functions where the user takes
responsibility for defining its WebAssembly mappings. Existing users
will need to change their code a bit, as signatures have changed.

For example, we are now more strict that all host functions require a
context parameter zero. Also, we've replaced
`HostModuleBuilder.ExportFunction` and `ExportFunctions` with a new type
`HostFunctionBuilder` that consolidates the responsibility and the
documentation.

```diff
 ctx := context.Background()
-hello := func() {
+hello := func(context.Context) {
         fmt.Fprintln(stdout, "hello!")
 }
-_, err := r.NewHostModuleBuilder("env").ExportFunction("hello", hello).Instantiate(ctx, r)
+_, err := r.NewHostModuleBuilder("env").
+        NewFunctionBuilder().WithFunc(hello).Export("hello").
+        Instantiate(ctx, r)
```

Power users can now use `HostFunctionBuilder` to define functions that
won't use reflection. There are two choices of interfaces to use
depending on if that function needs access to the calling module or not:
`api.GoFunction` and `api.GoModuleFunction`. Here's an example defining
one.

```go
builder.WithGoFunction(api.GoFunc(func(ctx context.Context, params []uint64) []uint64 {
	x, y := uint32(params[0]), uint32(params[1])
	sum := x + y
	return []uint64{sum}
}, []api.ValueType{api.ValueTypeI32, api.ValueTypeI32}, []api.ValueType{api.ValueTypeI32})
```
As you'll notice and as documented, this approach is more verbose and
not for everyone. If you aren't making a low-level library, you are
likely able to afford the 1us penalty for the convenience of reflection.
However, we are happy to enable this option for foundational libraries
and those with high performance requirements (like ourselves)!

Fixes https://github.com/tetratelabs/wazero/pull/825

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"reflect"
 )
 
 // ExternType classifies imports and exports with their respective types.
@@ -95,9 +94,10 @@ const (
 	//	(func (import "env" "f") (param externref) (result externref))
 	//
 	// This can be defined in Go as:
-	//  r.NewHostModuleBuilder("env").ExportFunctions(map[string]interface{}{
-	//    "f": func(externref uintptr) (resultExternRef uintptr) { return },
-	//  })
+	//  r.NewHostModuleBuilder("env").
+	//		NewFunctionBuilder().
+	//		WithFunc(func(context.Context, _ uintptr) (_ uintptr) { return }).
+	//		Export("f")
 	//
 	// Note: The usage of this type is toggled with api.CoreFeatureBulkMemoryOperations.
 	ValueTypeExternref ValueType = 0x6f
@@ -251,16 +251,11 @@ type FunctionDefinition interface {
 	// and not the imported function name.
 	DebugName() string
 
-	// GoFunc is present when the function was implemented by the embedder
-	// (ex via wazero.HostModuleBuilder) instead of a wasm binary.
+	// GoFunction is non-nil when implemented by the embedder instead of a wasm
+	// binary, e.g. via wazero.HostModuleBuilder
 	//
-	// This function can be non-deterministic or cause side effects. It also
-	// has special properties not defined in the WebAssembly Core
-	// specification. Notably, it uses the caller's memory, which might be
-	// different from its defining module.
-	//
-	// See https://www.w3.org/TR/wasm-core-1/#host-functions%E2%91%A0
-	GoFunc() *reflect.Value
+	// The expected results are nil, GoFunction or GoModuleFunction.
+	GoFunction() any
 
 	// ParamTypes are the possibly empty sequence of value types accepted by a
 	// function with this signature.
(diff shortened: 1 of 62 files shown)


api/wasm.go
```diff
@@ -301,6 +295,65 @@ type Function interface {
...
+// This function can be non-deterministic or cause side effects. It also
+// has special properties not defined in the WebAssembly Core specification.
+// Notably, this uses the caller's memory (via Module.Memory). See
+// https://www.w3.org/TR/wasm-core-1/#host-functions%E2%91%A0
+//
+// Most end users will not define functions directly with this, as they will
+// use reflection or code generators instead. These approaches are more
+// idiomatic as they can map go types to ValueType. This type is exposed for
+// those willing to trade usability and safety for performance.
+type GoModuleFunction interface {
```
codefromthecrypt, line 311: these two are the main thing. I audited and almost all functions can be covered with these two patterns, definitely all of our imports.

builder_test.go
```diff
@@ -12,10 +13,10 @@ import (
 func TestNewHostModuleBuilder_Compile(t *testing.T) {
 	i32, i64 := api.ValueTypeI32, api.ValueTypeI64
 
-	uint32_uint32 := func(uint32) uint32 {
+	uint32_uint32 := func(context.Context, uint32) uint32 {
```
codefromthecrypt, line 16: note: this drops support for host functions who don't add a context arg0.

imports/wasi_snapshot_preview1/clock.go
```diff
@@ -42,36 +42,45 @@ const (
...
-		case clockIDMonotonic:
-			resolution = uint64(sysCtx.NanotimeResolution())
-		default:
-			return ErrnoInval
-		}
-		if !mod.Memory().WriteUint64Le(ctx, resultResolution, resolution) {
-			return ErrnoFault
-		}
-		return ErrnoSuccess
+var clockResGet = &wasm.HostFunc{
```
codefromthecrypt, line 54: here's an example internal conversion

internal/engine/interpreter/interpreter.go
```diff
@@ -834,13 +833,20 @@ func (ce *callEngine) callFunction(ctx context.Context, callCtx *wasm.CallContex
...
 func (ce *callEngine) callGoFunc(ctx context.Context, callCtx *wasm.CallContext, f *function, params []uint64) (results []uint64) {
-	callCtx = callCtx.WithMemory(ce.callerMemory())
 	if f.source.Listener != nil {
 		ctx = f.source.Listener.Before(ctx, f.source.Definition, params)
 	}
 	frame := &callFrame{f: f}
 	ce.pushFrame(frame)
-	results = wasm.CallGoFunc(ctx, callCtx, f.source, params)
+
+	fn := f.source.GoFunc
```
codefromthecrypt, line 842: ex we only reset the call context if the function needs to

codefromthecrypt:
the integrated performance results are phenominally better for fine-grained ABI:

ex. https://github.com/http-wasm/http-wasm-host-go/pull/38 run via https://github.com/http-wasm/http-wasm-guest-tinygo benchmarks, TinyGo req/response logging went down from 43.5us/op to 14.5us/op

Before
```
Benchmark/example_router_host_response/TinyGo-16     	  352765	      3334 ns/op
Benchmark/example_router_host_response/wat-16        	  363664	      3171 ns/op
Benchmark/example_router_wasm_response/wat-16        	  232335	      4712 ns/op
Benchmark/example_router_wasm_response/TinyGo-16     	  225828	      4966 ns/op
Benchmark/example_wasi/TinyGo-16                     	   27218	     43631 ns/op
Benchmark/example_wasi/wat-16                        	   36333	     31908 ns/op
```

After
```
Benchmark/example_router_host_response/TinyGo-16  	 1148056	       979.8 ns/op
Benchmark/example_router_host_response/wat-16     	 1269897	       886.1 ns/op
Benchmark/example_router_wasm_response/wat-16     	 1098566	      1014 ns/op
Benchmark/example_router_wasm_response/TinyGo-16  	  802510	      1416 ns/op
Benchmark/example_wasi/TinyGo-16                  	   77762	     14477 ns/op
Benchmark/example_wasi/wat-16                     	  176056	      5794 ns/op
```

codefromthecrypt:
I'm going to merge this as we have a release in the next few days. I want to raise some tentative PRs meanwhile to ensure things work as well as they are so far. I'm happy to re-adjust and deal with drift on those PR if needed.

codefromthecrypt:
removed enforcement of context param when using reflection later per #832

PR #829, opened 2022-10-27
Special cases io.Discard on fd_write

codefromthecrypt:
This avoids memory copying in default configuration where stdout/stderr are discarded. This helps avoid performance impact when console logging happens, but intentionally discarded.


codefromthecrypt:
merging as I don't expect this controversial and my next PR will severely drift it!

PR #833, opened 2022-11-01
Extends build compatability to one version behind Go's support policy

codefromthecrypt:
This enforces that wazero will build and operate one version behind Go's support policy, making wazero's Go policy effectively three versions.

This is to allow libraries with wider Go policies to be able to use wazero. Specifically, mosn is the first to need this.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -14,7 +14,7 @@ on:
       - 'netlify.toml'
 
 env:  # Update this prior to requiring a higher minor version in go.mod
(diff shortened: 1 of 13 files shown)


.github/workflows/commit.yaml
```diff
@@ -24,12 +24,6 @@ jobs:
   check:
     name: Pre-commit check, Go-${{ matrix.go-version }}
```
codefromthecrypt, line 25: intentionally not inheriting the "keep old" version in the pre-commit check as it effects which tooling can be used (ex linters)

mathetake (approved):

PR #836, opened 2022-11-02
Implement WASI path_filestat_get

jerbob92:


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -744,12 +744,87 @@ var pathCreateDirectory = stubFunction(
...
+	wasiFiletypeCharacterDevice              = 2
+	wasiFiletypeDirectory                    = 3
+	wasiFiletypeRegularFile                  = 4
+	wasiFiletypeSocketDgram                  = 5
+	wasiFiletypeSocketStream                 = 6
+	wasiFiletypeSymbolicLink                 = 7
 )
 
+func pathFilestatGetFn(ctx context.Context, mod api.Module, params []uint64) []uint64 {
+	fd := uint32(params[0])
```
codefromthecrypt, line 773: so I think here fd is what to use to lookup from, and path is the relative one to that.

ex. fd=3 for '/' and "foo/bar" as path, resulting in `path_filestat_get(3, "foo/bar"...)` returning stat on /foo/bar, not "/"

jerbob92: I should probably have added some description in this PR.
`fstat` doesn't have these parameters so I didn't implement them. I think it's to support `fstat` and `fstatat` with one WASI call. I'm fine with adding support for these params, but for `fstat` the current implementation works.

codefromthecrypt: Roger. in any case this isn't about fstat it is about the wasi thing which is quasi similar. We won't be able to merge something that ignores its path param.. hope you understand https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-path_filestat_getfd-fd-flags-lookupflags-path-string---errno-filestat

jerbob92: @codefromthecrypt See my last edit, I'll close this PR and keep it like it is for future reference, will implement `fd_filestat_get` now.

Issue #835, opened 2022-11-01
Optimize `wasm.executeConstExpression`


ckaznocha:
**Is your feature request related to a problem? Please describe.**
I've been doing a bit of profiling. I noticed while instantiating my wasm module, my program is spending a fair bit of time in [`wasm.executeConstExpression`](https://github.com/tetratelabs/wazero/blob/d108ce4c43c6a2062a5ac723c5635163c3a18645/internal/wasm/store.go#L535). Most of the time is spent creating a `*bytes.Reader`. The `*bytes.Reader` is created every time `wasm.executeConstExpression` is called however there are three code paths (`OpcodeRefNull`, `OpcodeVecV128Const`, and an implicit default) which do not use the `*bytes.Reader`.
<img width="1332" alt="Screenshot 2022-11-01 at 12 47 50" src="https://user-images.githubusercontent.com/1851985/199325507-836fd16b-fca9-45fa-9dcb-62b3b471e186.png">

**Describe the solution you'd like**
Only create a `*bytes.Reader` for code paths that use it.

**Describe alternatives you've considered**
Taking a look at a few of the `Decode{Numeric Type}` funcs that the `*bytes.Reader` gets passed to, it seems using an `io.Reader` or a `*bytes.Reader` is an unnecessary indirection. Passing a `[]byte` will eliminate the performance penalty and make for simpler code. All of these functions look to be under `internal` so changing their signatures should be safe.

**Additional context**
Happy to submit a PR!

codefromthecrypt:
I agree that the best is to get rid of the reader entirely from this part. You got it right about internal.. that's why we use it!

There may be other functions that use leb etc which can also use a byte slice instead. If you see other opportunities for the same code pattern, please put it in the same PR.

Most of all, thanks for offering to help!

mathetake:
#837

PR #837, opened 2022-11-02
add ieee754 and leb128 byte slice funcs

ckaznocha:
This resoles #835 by factoring out the ieee754 and leb128 binary decoding, creating functions to decode from byte slices in addition to `io.ByteReader`s.

Once I started looking closer at how these functions are used I realized there are several places where passing the reader made sense. I've called the functions that decode from a byte slice `Load{numeric type}` and swapped them out anywhere a `*bytes.Reader` was being created for a single use.

Time spent creating readers no longer shows up in my profile.
Screenshot 2022-11-02 at 16 31 30

### DIFF
--- a/internal/leb128/leb128.go
+++ b/internal/leb128/leb128.go
@@ -1,14 +1,25 @@
 package leb128
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
+	"io"
 )
 
 const (
 	maxVarintLen32 = 5
+	maxVarintLen33 = maxVarintLen32
 	maxVarintLen64 = 10
+
+	int33Mask  int64 = 1 << 7
+	int33Mask2       = ^int33Mask
+	int33Mask3       = 1 << 6
+	int33Mask4       = 8589934591 // 2^33-1
+	int33Mask5       = 1 << 32
+	int33Mask6       = int33Mask4 + 1 // 2^33
+
+	int64Mask3 = 1 << 6
+	int64Mask4 = ^0
 )
 
 var (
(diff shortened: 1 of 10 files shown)


internal/leb128/leb128.go
```diff
@@ -141,11 +182,19 @@ func DecodeUint64(r *bytes.Reader) (ret uint64, bytesRead uint64, err error) {
 	return 0, 0, errOverflow64
 }
 
-func DecodeInt32(r *bytes.Reader) (ret int32, bytesRead uint64, err error) {
+func DecodeInt32(r io.ByteReader) (ret int32, bytesRead uint64, err error) {
+	return decodeInt32(byteReaderNext{r})
```
codefromthecrypt, line 186: clever way to re-use

codefromthecrypt (commented):
looking nice! looks like a glitch not found in normal tests. try `make spectest`

also, if you don't mind resolving the DCO bot.

excited about landing this!

ps. if you have instructions on how to see a profile result like you have, worthwhile sharing for others.

ckaznocha:
Thanks for taking a pass @codefromthecrypt. I'll get all the bots and checks happy shortly.

The flame graph is from pprof but the profile is from an project in a private repo. I can write up instructions to generate a similar one. I'll do that in another PR. Suggestions for where that would make the most sense to live?

codefromthecrypt:
> The flame graph is from pprof but the profile is from an project in a private repo. I can write up instructions to generate a similar one. I'll do that in another PR. Suggestions for where that would make the most sense to live?

Maybe make a new HACKING.md? I was thinking CONTRIBUTING, but it isn't really required for all change to do this.

codefromthecrypt (approved):
thanks!

PR #838, opened 2022-11-03
Implement WASI FS fd_filestat_get

jerbob92:
This is the proper version of #836 

Things that have not been fully implemented of `__wasi_filestat_t`:

- dev: device, id of device, don't know if we can get this from somewhere?
- ino: inode, file serial number, don't know if we can get this from somewhere?
- nlink: linkcount, number of hard links to the file, don't know if we can get this from somewhere?
- atim: timestamp, last data access timestamp, we just have the mod time so I made it use that, maybe we should leave this 0?
- ctim: timestamp, last file status change timestamp, we just have the mod time so I made it use that, maybe we should leave this 0?


codefromthecrypt:
@jerbob92 what you've done looks like a good start. I would move the questions in your PR desc to comments in the impl as this being a virtual filesystem and wasi being underspecified.. there aren't perfect choices. Later, we'll have to add a 

next steps would be to add unit test similar to the existing ones, and also vet through a couple compilers that it works. The latter might be easier as so many use wasi-libc.

Maybe try tinygo rust or zig as they are already in the example project.  This will help ensure that the partial implementation is usable (often is!) What I mean is you can play around in imports/wasi_snapshot_preview1/example/testdata easier as it has Makefile targets to compile the sources. This may help in ad-hoc testing to ensure it works'ish. I don't mean to change the cat example, just use the infra there to ensure the implementation is usable.

Thanks!

jerbob92:
@codefromthecrypt done, let me know what you think :)

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -194,15 +194,130 @@ var fdFdstatSetRights = stubFunction(
...
+		wasiFileMode = wasiFiletypeCharacterDevice
+	} else if fileMode&fs.ModeDir != 0 {
+		wasiFileMode = wasiFiletypeDirectory
+	} else if fileMode&fs.ModeType == 0 {
+		wasiFileMode = wasiFiletypeRegularFile
+	} else if fileMode&fs.ModeSymlink != 0 {
+		wasiFileMode = wasiFiletypeSymbolicLink
+	}
+
+	if !mod.Memory().WriteByte(ctx, buf+16, uint8(wasiFileMode)) {
```
codefromthecrypt, line 298: if you feel up to making this more efficient, go for it! I noticed @ckaznocha hunting down some perf and we might as well make a good example forward even if the rest of this file isn't. If this is beyond your budget one of us can do in a follow-up as generally this sort of thing is the fun part. you did the hard part :)

The key idea is that when writing fixed width data, we can use
`buf, ok = memory.Read(ctx, offset, byteCount)` to get a write-through buffer of the range. You may want to zero out any padding.

Then, we can get read the mtime only once like and convert it to a slice like this, then write it down to the slice. that or add functions in u64 to write the slice instead of making one.

e.g. in current code:
```
mtim := fileStat.ModTime().UnixNano()
mtimSlice := u64.LeBytes(mtim)

// now copy it a few times to buf you made.
```



u64

imports/wasi_snapshot_preview1/example/testdata/tinygo/cat.go
```diff
@@ -10,12 +12,27 @@ import (
...
+		}
+
+		defer file.Close()
+
+		stat, err := file.Stat()
+		if err != nil {
+			os.Exit(1)
+		}
+
+		os.Stdout.Write([]byte(fmt.Sprintf("Size: %d\n", stat.Size())))
```
codefromthecrypt, line 27: cool I trust this works so you can revert it now!

codefromthecrypt (approved):
Looks great and thanks for the rigor. I think main thing is to satisfy the bot and make sure tests pass. I'll squash on way in unless you care to prior.

imports/wasi_snapshot_preview1/fs.go
ckaznocha: Hey, hope I'm not over stepping but since I heard my name I took a peek 😉.

Alternately:
```suggestion
	mtim := uint64(fileStat.ModTime().UnixNano())
    binary.LittleEndian.PutUin64(buf[offset+40:], mtim)
    binary.LittleEndian.PutUin64(buf[offset+48:], mtim)
    binary.LittleEndian.PutUin64(buf[offset+56:], mtim)
```

I'm honestly not sure if it would be faster to allocate the new slice and copy it 3 times vs writing directly to the write-through buffer but doing the bitwise operations 3 times. Either way write-through buffer seems like a good call!

codefromthecrypt: I agree with your approach! in any case a bench could be added here to see the impact. I used this earlier to validate the significant perf impact when we stopped using reflective signatures https://github.com/tetratelabs/wazero/blob/2173f3075878b89162884bf5ce280d59f6212c69/imports/wasi_snapshot_preview1/wasi_bench_test.go

imports/wasi_snapshot_preview1/example/testdata/tinygo/cat.go
jerbob92: Don't you want to keep this in here to make sure that stat keeps working? Or is it actually supposed to be an example and not some sort of test?

imports/wasi_snapshot_preview1/fs.go
jerbob92: Sounds good, will make the changes later! Would zero'ing out the padding do anything? Isn't the memory already zero'ed when it's being allocated?

imports/wasi_snapshot_preview1/example/testdata/tinygo/cat.go
codefromthecrypt: the example is not a full scale test. Main thing since things aren't documented is to make sure something works with it. After that's the case normal unit tests should be enough.

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: I'm not terribly concerned about the zeroing. good point!

codefromthecrypt:
I plan to contribute to polishing this tomorrow or earlier

codefromthecrypt:
rebased

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: done!

codefromthecrypt (approved):
@jerbob92 looks good to me. once you ack I'll merge

jerbob92:
> @jerbob92 looks good to me. once you ack I'll merge

Ack

PR #841, opened 2022-11-05
Optimize wasm.(*ModuleInstance).BuildFunctions

ckaznocha:
Before:
```sh
❯ go test -benchmem -memprofile memprofile.out -cpuprofile profile.out -run=^$ -bench ^BenchmarkInitialization/compiler$ github.com/tetratelabs/wazero/internal/integration_test/bench
goos: darwin
goarch: arm64
pkg: github.com/tetratelabs/wazero/internal/integration_test/bench
BenchmarkInitialization/compiler-10                69037             18685 ns/op          145951 B/op        128 allocs/op
PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/bench   2.850s
```
Screenshot 2022-11-05 at 13 36 14

After:
```sh
❯ go test -benchmem -memprofile memprofile.out -cpuprofile profile.out -run=^$ -bench ^BenchmarkInitialization/compiler$ github.com/tetratelabs/wazero/internal/integration_test/bench
goos: darwin
goarch: arm64
pkg: github.com/tetratelabs/wazero/internal/integration_test/bench
BenchmarkInitialization/compiler-10                64510             17417 ns/op          145950 B/op        128 allocs/op
PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/bench   1.600s
```
Screenshot 2022-11-05 at 13 39 00

Not a massive win but simpler code IMO

PR #842, opened 2022-11-05
Optimize compiler.(*engine).NewModuleEngine

ckaznocha:
Before:
```sh
❯ go test -benchmem -memprofile memprofile.out -cpuprofile profile.out -run=^$ -bench ^BenchmarkInitialization/compiler$ github.com/tetratelabs/wazero/internal/integration_test/bench
goos: darwin
goarch: arm64
BenchmarkInitialization/compiler-10                64510             17417 ns/op          145950 B/op        128 allocs/op
PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/bench   1.600s
```
Screenshot 2022-11-05 at 14 06 24

After:
```sh
❯ go test -benchmem -memprofile memprofile.out -cpuprofile profile.out -run=^$ -bench ^BenchmarkInitialization/compiler$ github.com/tetratelabs/wazero/internal/integration_test/bench
goos: darwin
goarch: arm64
BenchmarkInitialization/compiler-10                59250             17296 ns/op          145950 B/op        128 allocs/op
PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/bench   1.493s
```
Screenshot 2022-11-05 at 14 15 46

Another pretty minor win. Includes commits from #841, if that one can get merged first this will be a clean diff.

### DIFF
--- a/internal/integration_test/bench/bench_test.go
+++ b/internal/integration_test/bench/bench_test.go
@@ -89,6 +89,8 @@ func runCompilation(b *testing.B, r wazero.Runtime) wazero.CompiledModule {
 func runInitializationBench(b *testing.B, r wazero.Runtime) {
 	compiled := runCompilation(b, r)
 	defer compiled.Close(testCtx)
+	// Configure with real sources to avoid performance hit initializing fake ones. These sources are not used
+	// in the benchmark.
 	config := wazero.NewModuleConfig().WithSysNanotime().WithSysWalltime().WithRandSource(rand.Reader)
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
--- a/internal/integration_test/bench/bench_test.go
+++ b/internal/integration_test/bench/bench_test.go
@@ -2,9 +2,9 @@ package bench
 
 import (
 	"context"
+	"crypto/rand"
 	_ "embed"
 	"fmt"
-	"math/rand"
 	"runtime"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


ckaznocha:
Not sure why the ci timed out, it passes when I run it on my laptop
```sh
❯ make spectest
ok      github.com/tetratelabs/wazero/internal/integration_test/spectest/v1     2.197s
ok      github.com/tetratelabs/wazero/internal/integration_test/spectest/v2     3.307s
```

internal/integration_test/bench/bench_test.go
```diff
@@ -89,9 +89,10 @@ func runCompilation(b *testing.B, r wazero.Runtime) wazero.CompiledModule {
 func runInitializationBench(b *testing.B, r wazero.Runtime) {
 	compiled := runCompilation(b, r)
 	defer compiled.Close(testCtx)
+	config := wazero.NewModuleConfig().WithSysNanotime().WithSysWalltime().WithRandSource(rand.Reader)
```
mathetake, line 92: what's the reason for this change? looks irrelevant and I suppose this will end up introducing non-determinism in the test, which is bad.

mathetake:
freebsd tends to timeout occationally, so you can just ignore the ci error on that one

internal/integration_test/bench/bench_test.go
ckaznocha: The reason for moving `wazero.NewModuleConfig()` out here and storing it in a variable to be re-used is because typically you'd want to exclude setup code from your benchmark to limit the measurements to the function under test (in this case `r.InstantiateModule`).

The reason for adding `.WithSysNanotime().WithSysWalltime().WithRandSource(rand.Reader)` is because `platform.NewFakeRandSource()`, `platform.NewFakeWalltime()`, and `platform.NewFakeNanotime()` are all costly to call during each benchmark. I'm just amortizing the config to reduce irrelevant noise in the profile. Because this benchmark is just measuring initialization it didn't seem like loading a real clock and rand source would introduce any non-determinisim but if you prefer it would be just as effective to setup the config with the same fake versions before resetting the timer. Happy to make that change if it's something that's important to you.

codefromthecrypt:
```suggestion
	// Configure with real sources to avoid performance hit initializing fake ones. These sources are not used
	// in the benchmark.
	config := wazero.NewModuleConfig().WithSysNanotime().WithSysWalltime().WithRandSource(rand.Reader)
```

ckaznocha:
I regret trying to stack changes now. If this one is acceptable, it can just get merged and #841 should auto close or it can be manually closed.

mathetake (approved):
cool, thanks!

PR #845, opened 2022-11-08
gojs: gc fix refcount issue

robbertvanginkel:
I was attempting to use gojs with a binary that read from stdin, did some
transformations, then printed to stdout. This worked or small inputs, but for 
larger values would panic like:

	runtime: g 5: unexpected return pc for syscall/js.Value.Call called from 0x18ca120
	<memory addresses & stack>
	syscall/js.Value.Call({{}, 0x1720028, 0x141c0001}, {0x0, 0x0}, {0x0, 0x0, 0x0})
	        /usr/local/go/src/syscall/js/js.go:384 +0x35 fp=0x8327d0 sp=0x832710 pc=0x16e50035
	created by runtime.createfing
	        /usr/local/go/src/runtime/mfinal.go:157 +0x5

After some debugging, I traced it to gojs.FinalizeRef being called for idJsFs
when a GC is triggered after stdout/stderr is used. I've included a small test
that triggers the same issue (and and is fixed by the same patch). The 
testcase fails with:

	runtime error: index out of range [4294967288] with length 5 (recovered by wazero)
        wasm stack trace:
                go.syscall/js.finalizeRef(i32)
                go.syscall/js.finalizeRef.proxy(i32)

The 4294967288 index results from uint32 underflow after subracting `nextID`
(18) from 10 (the value for `idJsfs`). Looking at the code paths that reach
refcount increment, it seems like decrement for values below nextID should be a
nop or decrement should not be called at all.

### DIFF
--- a/internal/gojs/state.go
+++ b/internal/gojs/state.go
@@ -195,6 +195,7 @@ func (j *values) increment(v interface{}) uint32 {
 }
 
 func (j *values) decrement(id uint32) {
+	// Special IDs are not refcounted.
 	if id < nextID {
 		return
 	}
--- a/internal/gojs/state.go
+++ b/internal/gojs/state.go
@@ -195,6 +195,9 @@ func (j *values) increment(v interface{}) uint32 {
 }
 
 func (j *values) decrement(id uint32) {
+	if id < nextID {
+		return
+	}
 	id -= nextID
 	j.goRefCounts[id]--
 	if j.goRefCounts[id] == 0 {
--- a/internal/gojs/misc_test.go
+++ b/internal/gojs/misc_test.go
@@ -40,3 +40,13 @@ func Test_stdio(t *testing.T) {
 	require.Equal(t, "println stdin\nStderr.Write", stderr)
 	require.Equal(t, "Stdout.Write", stdout)
 }
+
+func Test_gc(t *testing.T) {
+	t.Parallel()
+
+	stdout, stderr, err := compileAndRun(testCtx, "gc", wazero.NewModuleConfig())
+
+	require.EqualError(t, err, `module "" closed with exit_code(0)`)
+	require.Equal(t, "", stderr)
+	require.Equal(t, "before gc\nafter gc\n", stdout)
+}
--- a/internal/gojs/testdata/gc/main.go
+++ b/internal/gojs/testdata/gc/main.go
@@ -1,0 +1,12 @@
+package gc
+
+import (
+	"fmt"
+	"runtime"
+)
+
+func Main() {
+	fmt.Println("before gc")
+	runtime.GC()
+	fmt.Println("after gc")
+}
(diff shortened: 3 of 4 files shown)


internal/gojs/state.go
```diff
@@ -195,6 +195,9 @@ func (j *values) increment(v interface{}) uint32 {
 }
 
 func (j *values) decrement(id uint32) {
+	if id < nextID {
```
mathetake, line 198: > it seems like decrement for values below nextID should be a
nop or decrement should not be called at all.

I would love to put a comment with the permalink for your research here
```suggestion
	// Decrement  for values below nextID should be a
nop.
	// See https://github.com/golang/go/....
	if id < nextID {
```

robbertvanginkel: This wasn't based on research into the golang repo's `wasm_exec.js`, but purely
based on wazero's gojs implementation. I couldn't actually find any references
in the upstream implementation to nextID or the reason the `value` structs's
get/increment/decrement in `state.go` uses it to offset IDs before lookup.

Looking at what increments the refcount, it only happens in `storeRef`.
However, it does not increment for values that are *jsVal:
https://github.com/tetratelabs/wazero/blob/d5664ed5d1f5af2836c29ff731a1f13192bf255d/internal/gojs/state.go#L133-L134
which jsfs is:
https://github.com/tetratelabs/wazero/blob/d5664ed5d1f5af2836c29ff731a1f13192bf255d/internal/gojs/fs.go#L21.
based on the switch statement in
https://github.com/tetratelabs/wazero/blob/d5664ed5d1f5af2836c29ff731a1f13192bf255d/internal/gojs/state.go#L55,
all ids < nextID are special cased and not stored in values.

I think the values struct was only meant to to handle ids >= nextID, and that
finalize was not expected to be called for ids < nextID. I did not find the
reason why it is called on a low id, but since `wasm_exec.js` doesn't differentiate
among the ids on the finalizer it seems reasonable that it could. If wazero
special cases the load/store for these ids, it probably also should the
finalize.
 
Happy to commenting something along these lines or moving the id check to the
finalize call instead, but I don't have an external source to cite.

mathetake: ok ic, thank you for clarifying. I think I defer this to @codefromthecrypt as this is beyond my knowledge on gojs.

codefromthecrypt:
```suggestion
	// Special IDs are not refcounted.
	if id < nextID {
```

codefromthecrypt (approved):
excellent detective work. thanks!

internal/gojs/state.go
codefromthecrypt: I added a comment which while imperfect is better than nothing. The overall thing is that we have more "special IDs" than go has, to avoid constant thrash on things that are commonly requested, and won't change for the same module instance. the guest is unaware of these special IDs so they will try to call finalize etc for them. special casing is the correct thing to do here!

PR #846, opened 2022-11-08
Adds wasi_snapshot_preview1.NewFunctionExporter

codefromthecrypt:
This adds `wasi_snapshot_preview1.NewFunctionExporter` for the following two use cases:

* Overriding a builtin function with an alternate implementation.
  * Allows developers like @jerbob92 to test new WASI functions without forking wazero.
* Exporting functions to the module "wasi_unstable" for legacy code.
  * Needed by mosn to support existing wasm which may never go away.


codefromthecrypt:
cc @antJack @taoyuanyuan from mosn. This is the main change needed that lets wazero power old proxy-wasm ABI

codefromthecrypt:
verified wazero can now implement old proxy-wasm in mosn with this https://github.com/mosn/proxy-wasm-go-host/compare/main...codefromthecrypt:wazero

codefromthecrypt:
actually, this won't help too much with WASI because most things need internal types. anyway it would when we expose them (which isn't soon as we have to re-evaluate fs.FS)

imports/wasi_snapshot_preview1/usage_test.go
```diff
@@ -16,18 +16,17 @@ var wasiArg []byte
...
 func TestInstantiateModule(t *testing.T) {
 	r := wazero.NewRuntime(testCtx)
+	defer r.Close(testCtx)
 
 	stdout := bytes.NewBuffer(nil)
 
 	// Configure WASI to write stdout to a buffer, so that we can verify it later.
 	sys := wazero.NewModuleConfig().WithStdout(stdout)
-	wm, err := Instantiate(testCtx, r)
+	_, err := Instantiate(testCtx, r)
```
anuraaga, line 25: While cleaning, can consider switching to `MustInstantiate`

anuraaga (approved):

imports/wasi_snapshot_preview1/usage_test.go
codefromthecrypt: good idea

Issue #847, opened 2022-11-08
Too large offset for ADR


clarkmcc:
**Describe the bug**
I'm compiling a production wasm module that has many, very large regular expressions (around 9kb) and I'm getting the following error when compiling in debug mode only
```
error compiling wasm func[.<regex::exec::ExecNoSync as regex::re_trait::RegularExpression>::captures_read_at::hdbe37ea0d10b827b]: failed to compile: BUG: too large offset for ADR
```
In release mode, I'm not getting the same error
```
[profile.release]
lto = "thin"
opt-level = 's'
strip = "symbols"
```

**To Reproduce**
Sadly, the wasm module is a closed-source project. If it is helpful, I can try to come up with a reproduction, but I wanted to check if there was a quick answer first.

**Environment (please complete the relevant information):**
- Go version: go version go1.19 darwin/arm64
- wazero Version: v1.0.0-pre.3
- Host architecture: arm64
- Runtime mode: compiler

codefromthecrypt:
looks like indeed this is a temporary limitation in the arm64 variant

```
		offset := targetNode.OffsetInBinary() - n.OffsetInBinary()
		if offset > math.MaxUint8 {
			// We could support up to 20-bit integer, but byte should be enough for our impl.
			// If the necessity comes up, we could fix the below to support larger offsets.
			return fmt.Errorf("BUG: too large offset for ADR")
		}
```

codefromthecrypt:
hi @clarkmcc can you run off latest main? Curious how big an ADR offset you have in your wasm, as that's the main thing we need to consider.

mathetake:
oh i didn't think there would be a real example to hit this branch... :D

mathetake:
glad I captured this as an error rather than horrible runtime crash though!

clarkmcc:
I'm having trouble replicating 🙄. I should have set aside that compiled version for further troubleshooting. If I can figure out how to replicate it, I'll post results.

mathetake:
my memory on this has gotten fuzzy so I'm not sure too, but I imagine a huge `br_table` branch might cause this, but yeah not sure

mathetake:
In anyway I don't think we need a reproducer, and we can simply change the encoding of `ADR` instruction https://github.com/tetratelabs/wazero/blob/main/internal/asm/arm64/impl.go#L1844-L1857 here and can add support for maximum 20-bit offset.

mathetake:
[...]

PR #850, opened 2022-11-09
Adds more context for ADR offset overrun

codefromthecrypt:
This is mainly so we can tell if changing this will work. e.g. if 20 bits are enough for #847


codefromthecrypt:
I'm guessing an offset >20bits will be near impossible, but yeah just curious.

anuraaga (approved):

PR #851, opened 2022-11-09
arm64: adds support for large in-block jumps with ADR

### LINKED ISSUE #847: Too large offset for ADR

mathetake:
Fixes #847 

### DIFF
--- a/internal/asm/arm64/impl.go
+++ b/internal/asm/arm64/impl.go
@@ -1833,11 +1833,11 @@ func (a *AssemblerImpl) encodeADR(n *nodeImpl) (err error) {
 		}
 
 		offset := targetNode.OffsetInBinary() - n.OffsetInBinary()
-		if offset > 1<<20 {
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
I love how you supported more bits and the main code looks simpler as a result!

Issue #662, opened 2022-06-27
Ability/example to implement custom host functions defined by WITX


grokify:
**Is your feature request related to a problem? Please describe.**

I'm interested in implementing custom host functions defined via WITX in wazero, which can be accessed by a Wasm guest.

The host function WITX files for this project are from the Fastly Compute@Edge ABI defined here. This is currently implemented on a Rust runtime and it would be great to have this implemented on the wazero zero dependency Go runtime as well.

* [`compute-at-edge.witx`](https://github.com/fastly/Viceroy/blob/b37b89fc07ecd5845f1103380c5e4fae9cc13b30/lib/compute-at-edge-abi/compute-at-edge.witx)
* [`typenames.witx`](https://github.com/fastly/Viceroy/blob/b37b89fc07ecd5845f1103380c5e4fae9cc13b30/lib/compute-at-edge-abi/typenames.witx)

**Describe the solution you'd like**

The ability to implement a WITX defined host call in a wazero host implementation, ideally with an example.

**Describe alternatives you've considered**

I've looked at the counter example here. It would be nice to see how this maps to a WITX definition.

https://github.com/tetratelabs/wazero/blob/main/examples/namespace/counter.go

I'm unsure if this can be implemented given the lack of support for user-defined types.

https://github.com/tetratelabs/wazero/issues/420

**Additional context**

None at the moment.

codefromthecrypt:
quick context for starters.

* Host function signatures map directly to wasm signatures, so first step is to get a format to map to that (ex via %.wat)
* As far as I know, the witx format is deprecated for another format. ex https://github.com/WebAssembly/wasi-nn/pull/17
* I don't think user-defined types is the concern here as the same issue occurs in wasi even if they switched off witx mostly.

What's ultimately needed is code generator support in something like https://github.com/jedisct1/witx-codegen See if you can find one that can emit imports in a way convertible to %.wat is maybe the best start?

In other words it is probably a good first step to port a tool that already generates code from witx to generate go (specifically wazero host functions). That reduces the surface area to conversions and also buys time to figure out if witx is really deprecated or only deprecated in the WebAssembly orgs..

codefromthecrypt:
@aturon Do you know any plans for fastly and witx vs its successor? Also, it would be nice to be able to support this API even regardless of what might be migrating in terms of ABI formats. If you know something that can dump the ABI as %.wat imports I can help guide a manual impl meanwhile.

grokify:
[...]

PR #855, opened 2022-11-21
Adds rationale for api.ValueType encoding

codefromthecrypt:
We made some decisions around encoding last year, but we didn't detail why. This leads to questions.

e.g. https://gophers.slack.com/archives/C040AKTNTE0/p1668883638982119

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -46,6 +46,7 @@ x/sys as a platform plugin without forcing all users to maintain that
 dependency.
 
 ## Project structure
+
 wazero uses internal packages extensively to balance API compatability desires for end users with the need to safely
 share internals between compilers.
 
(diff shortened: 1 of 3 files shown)


RATIONALE.md
```diff
@@ -112,6 +115,65 @@ Here are some examples:
 
 Besides security, this practice prevents other bugs and allows centralization of validation logic such as decoding Wasm.
 
+## API Design
+
+### Why are Go function signature parameters and results uint64?
+
+Go-defined function signatures include parameters and results expressed as a
```
mathetake, line 122: Isn't this only for the Call API? Host functions can naturally have any types other than u64

mathetake (approved):

RATIONALE.md
codefromthecrypt: good point. I guess I didn't write what I meant.. we have two apis to call functions from go..

I'll try to rewrite a bit as this is about invocation from go

codefromthecrypt: PTAL!

mathetake (approved):

codefromthecrypt:
ps @jerbob92 mentions

> It makes sense but still think it's confusing for developers that you're expecting something with the size of a uint64, and not a literal uint64. 

imho agreed it is confusing, and hopefully the ecosystem will use/make things like [go-plugin](https://github.com/knqyf263/go-plugin) or a future [wit/witx binder](#662) to reduce exposure of end users. meanwhile we can expect some questions, but at least RATIONALE is here for folks in the community to help answer them

PR #856, opened 2022-11-22
Adds emscripten invoke_xxx functions needed for PDFium

codefromthecrypt:
This adds host functions that work on dynamic function tables. These are only used by emscripten, but require some infrastructure to support it. I added the least possible to due the task. This also only handles i32 and void returns with up to four parameters as that covers the needs of PDFium. Future integrations may need more parameters or a mix of floats. Such use cases should be addressed as they come as otherwise it is a lot of work for the cartesian product of all combinations.

See https://github.com/WebAssembly/binaryen/blob/1b0d724fd5938931e6924941641dd8924ad49938/test/passes/post-emscripten.wast

See https://github.com/jerbob92/go-pdfium-wasm


mathetake (commented):
are these functions stable over emcc versions, or can be broken at any time?

codefromthecrypt:
> are these functions stable over emcc versions, or can be broken at any time?

the convention is stable so far. as noted by @jerbob92 on slack, a future emscripten user may need 70+ variants, and when we run into that we may need to make a dynamic generator based on a guest module. However, we don't currently have a way to define host modules that have 1-1 relationship with a guest, so punting it. When we get there, the tech needed will be in a better place to justify.

PR #857, opened 2022-11-24
implements fd_pread in wasi and adds zig wasi example

codefromthecrypt:
This implements fd_pread which is required when you don't buffer reads in zig (particularly when your buffer isn't >= file length).

Advice on zig code welcome as it is a new language to me.

Note: The approach here is same as we use internall for GOOS=js as that has an fdPread syscall. Just, we never implemented it for wasi, yet.


imports/wasi_snapshot_preview1/example/testdata/zig/cat.zig
```diff
@@ -0,0 +1,41 @@
+const std = @import("std");
+const io = std.io;
+const os = std.os;
+const allocator = std.heap.page_allocator;
+const PreopenList = std.fs.wasi.PreopenList;
+const stdout = std.io.getStdOut();
+const warn = std.log.warn;
+
+pub fn main() !void {
```
codefromthecrypt, line 9: cc @Luukdegram @knqyf263 @mathetake for wasi+zig style hints I found pre-opens the most unintuitive as I figured "/" would somehow be easier to map

mathetake: I expected that we don't need to access WASi specifics like PreopenList...

mathetake: ok so I imagine the reason behind this is Zig stdlib maintains the preopens in memory and therefore it requires the allocation, and it turns out that we need to call WASI specific APIs in the main to avoid hidden allocations. sounds plausible

codefromthecrypt: yeah I should clarify the best comment and cite https://ziglang.org/documentation/0.10.0/#WASI

codefromthecrypt: still feels off that we can't avoid adding wasi-specific code

imports/wasi_snapshot_preview1/example/testdata/zig/cat.zig
```diff
@@ -0,0 +1,37 @@
...
+    // Allocate arguments from the the operating system.
+    const args = try std.process.argsAlloc(allocator);
+    defer std.process.argsFree(allocator, args);
+
+    // Pre-opened files are not available by default. When using WASI, you must
+    // manually populate and initialize pre-opens to prevent openFile crashes.
+    var preopens = PreopenList.init(allocator);
+    defer preopens.deinit();
+    try preopens.populate(null);
```
mathetake, line 18: what's this for? I would like a comment on it

Makefile
```diff
@@ -53,9 +53,12 @@ build.examples.as:
 
 # Use -fstage1 to avoid bugs in the new compiler
 # https://github.com/ziglang/zig/wiki/Self-Hosted-Compiler-Upgrade-Guide#is-it-time-to-upgrade
+%.wasm: %.zig
+	@(cd $(@D); zig build -fstage1)
```
mathetake, line 57: off topic:now that 0.10.0 becomes stable, maybe worth trying to  delete`-fstage1` in another PR

mathetake (approved):

imports/wasi_snapshot_preview1/example/testdata/zig/cat.zig
codefromthecrypt: no idea, I was wildly trying to get things to not crash by looking at other sources. I noticed I can remove it so just did!

imports/wasi_snapshot_preview1/example/testdata/zig/cat.zig
codefromthecrypt: added comment

Makefile
codefromthecrypt: good idea. there are some unrelated things that break here so we can try it in another PR

codefromthecrypt:
ok I'll wait till tomorrow in case someone from ziglang can clarify anything. the pre-opens still feel very uncomfortable and I prefer zig to be as comfortable as a couch at the end of a long day cc @andrewrk

codefromthecrypt:
I'll merge this and handle any feedback post-commit as I need this infra to test #858

codefromthecrypt:

PR #858, opened 2022-11-24
wasi: adds path_filestat_get

codefromthecrypt:
This adds path_filestat_get which is needed for PDFium and zig's fstatatWasi.

https://github.com/ziglang/zig/blob/aea617c60be848ed372343e08465bb0ac6cb7b85/lib/std/os.zig#L4171

The original source of this was written by @jerbob92

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -7,7 +7,7 @@ import (
 	"io"
 	"io/fs"
 	"path"
-	"strings"
+	"syscall"
 
 	"github.com/tetratelabs/wazero/api"
 	internalsys "github.com/tetratelabs/wazero/internal/sys"
@@ -946,9 +946,6 @@ func pathFilestatGetFn(ctx context.Context, mod api.Module, params []uint64) []u
 		return errnoNametoolong
 	}
 	pathName := string(b)
-	if strings.Contains(pathName, "../") {
-		return errnoInval // don't allow escaping the root of the FD
-	}
 
 	if dir, ok := fsc.OpenedFile(ctx, fd); !ok {
 		return errnoBadf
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
having trouble triggering this with zig so I raised https://github.com/ziglang/zig/issues/13644

I'd like to verify at least two compilers use this coherently..

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -850,14 +853,91 @@ var pathCreateDirectory = stubFunction(
...
+	resultBuf := uint32(params[4])
+
+	// open_at isn't supported in fs.FS, so we check the path can't escape,
+	// then join it with its parent
+	b, ok := mod.Memory().Read(ctx, pathOffset, pathLen)
+	if !ok {
+		return errnoNametoolong
+	}
+	pathName := string(b)
+	if strings.Contains(pathName, "../") {
```
jerbob92, line 915: Isn't this the responsibility of the FS to validate this in OpenFile?

codefromthecrypt: you are totally right. I forgot that's a rule of fs.FS!

codefromthecrypt:
verified in zig by changing the workdir explicitly to "/". Note, this doesn't work in zig until https://github.com/ziglang/zig/pull/13646


```zig
const std = @import("std");
const allocator = std.heap.page_allocator;
const PreopenList = std.fs.wasi.PreopenList;

pub fn main() !void {
    var preopens = PreopenList.init(allocator);
    defer preopens.deinit();

    try std.os.initPreopensWasi(allocator, "/");
    try std.os.chdir("/");
}
```

```
==> wasi_snapshot_preview1.fd_prestat_get(fd=3,result.prestat=1040264)
<== ESUCCESS
==> wasi_snapshot_preview1.fd_prestat_dir_name(fd=3,path=1044384,path_len=1)
<== ESUCCESS
==> wasi_snapshot_preview1.fd_prestat_get(fd=4,result.prestat=1040264)
<== EBADF
==> wasi_snapshot_preview1.path_filestat_get(fd=3,flags=1,path=1049005,path_len=1,result.buf=1048512)
<== ESUCCESS
==> wasi_snapshot_preview1.proc_exit(rval=0)
```

mathetake (approved):

PR #860, opened 2022-11-25
Optimizes GoModuleFunction signature and ensures function result slices are unique

codefromthecrypt:
This fixes a bug where function call results used the same slice (in the compiler engine). This was due to it being a write-through of a shared stack. What happens is a user of a first result can corrupt the next, or possibly worse:

```go
results, err := fn.Call(ctx, 1)
results2, err := fn.Call(ctx, 2)
results[0] = 255 // affects results2[0]
```

The safe approach is to always copy-out (internally). However, stack re-use was an optimization to avoid copying. To pay for the copy-out, this optimizes the signatures of `api.GoFunction` and `api.GoModuleFunction` as described below. As these are low-level interfaces, most users will have no impact to this change. I'll help port things on github that use them, though.

### Design

The general idea is that if we want to make Function.Call safe (ex copy-out results), we should pay for that by reducing the amount of slices we allocate.

Currently, we require host functions to allocate a temporary slice to hold results when we will have to copy them out anyway. This changes the design so that low-level functions accept a slice representing the stack instead. The size is the max of the parameter or result count.

```diff
 type GoModuleFunction interface {
-       Call(ctx context.Context, mod Module, params []uint64) []uint64
+       Call(ctx context.Context, mod Module, stack []uint64)
 }
```

This approach reduces allocation and also copies as currently both impls needed to copy results back to the stack anyway.

### Porting 

Common functions who return a single value, like WASI, can be adapted using a custom type instead of `api.GoModuleFunction`.

Here's an example:
```go
// wasiFunc special cases that all WASI functions return a single Errno
// result. The returned value will be written back to the stack at index 0.
type wasiFunc func(ctx context.Context, mod api.Module, params []uint64) Errno

// Call implements the same method as documented on wasiFunction.
func (f wasiFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
	// Write the result back onto the stack
	stack[0] = uint64(f(ctx, mod, stack))
}
```

The diff then is mostly mechanical:
```diff
--- a/imports/wasi_snapshot_preview1/args.go
+++ b/imports/wasi_snapshot_preview1/args.go
@@ -52,11 +52,11 @@ var argsGet = &wasm.HostFunc{
        ResultTypes: []api.ValueType{i32},
        Code: &wasm.Code{
                IsHostFunction: true,
-               GoFunc:         api.GoModuleFunc(argsGetFn),
+               GoFunc:         wasiFunc(argsGetFn),
        },
 }
 
-func argsGetFn(ctx context.Context, mod api.Module, params []uint64) []uint64 {
+func argsGetFn(ctx context.Context, mod api.Module, params []uint64) Errno {
        sysCtx := mod.(*wasm.CallContext).Sys
        argv, argvBuf := uint32(params[0]), uint32(params[1])
        return writeOffsetsAndNullTerminatedValues(ctx, mod.Memory(), sysCtx.Args(), argv, argvBuf)
@@ -99,20 +99,21 @@ var argsSizesGet = &wasm.HostFunc{
        ResultTypes: []api.ValueType{i32},
        Code: &wasm.Code{
                IsHostFunction: true,
-               GoFunc:         api.GoModuleFunc(argsSizesGetFn),
+               GoFunc:         wasiFunc(argsSizesGetFn),
        },
 }
```

### DIFF
--- a/imports/wasi_snapshot_preview1/wasi.go
+++ b/imports/wasi_snapshot_preview1/wasi.go
@@ -241,7 +241,7 @@ func writeOffsetsAndNullTerminatedValues(ctx context.Context, mem api.Memory, va
 // result. The returned value will be written back to the stack at index zero.
 type wasiFunc func(ctx context.Context, mod api.Module, params []uint64) Errno
 
-// Call implements the same method as documented on api.GoModuleFunc.
+// Call implements the same method as documented on api.GoModuleFunction.
 func (f wasiFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 	// Write the result back onto the stack
 	stack[0] = uint64(f(ctx, mod, stack))
--- a/imports/wasi_snapshot_preview1/wasi.go
+++ b/imports/wasi_snapshot_preview1/wasi.go
@@ -215,39 +215,36 @@ func exportFunctions(builder wazero.HostModuleBuilder) {
 	exporter.ExportHostFunc(sockShutdown)
 }
 
-// Declare constants to avoid slice allocation per call.
-var (
-	errnoBadf        = []uint64{uint64(ErrnoBadf)}
-	errnoExist       = []uint64{uint64(ErrnoExist)}
-	errnoInval       = []uint64{uint64(ErrnoInval)}
-	errnoIo          = []uint64{uint64(ErrnoIo)}
-	errnoNoent       = []uint64{uint64(ErrnoNoent)}
-	errnoNotdir      = []uint64{uint64(ErrnoNotdir)}
-	errnoFault       = []uint64{uint64(ErrnoFault)}
-	errnoNametoolong = []uint64{uint64(ErrnoNametoolong)}
-	errnoSuccess     = []uint64{uint64(ErrnoSuccess)}
-)
-
-func writeOffsetsAndNullTerminatedValues(ctx context.Context, mem api.Memory, values []string, offsets, bytes uint32) []uint64 {
+func writeOffsetsAndNullTerminatedValues(ctx context.Context, mem api.Memory, values []string, offsets, bytes uint32) Errno {
 	for _, value := range values {
 		// Write current offset and advance it.
 		if !mem.WriteUint32Le(ctx, offsets, bytes) {
-			return errnoFault
+			return ErrnoFault
 		}
 		offsets += 4 // size of uint32
 
 		// Write the next value to memory with a NUL terminator
 		if !mem.Write(ctx, bytes, []byte(value)) {
-			return errnoFault
+			return ErrnoFault
 		}
 		bytes += uint32(len(value))
 		if !mem.WriteByte(ctx, bytes, 0) {
-			return errnoFault
+			return ErrnoFault
 		}
 		bytes++
 	}
 
-	return errnoSuccess
+	return ErrnoSuccess
+}
+
+// wasiFunc special cases that all WASI functions return a single Errno
+// result. The returned value will be written back to the stack at index zero.
+type wasiFunc func(ctx context.Context, mod api.Module, params []uint64) Errno
+
+// Call implements the same method as documented on wasiFunction.
+func (f wasiFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
+	// Write the result back onto the stack
+	stack[0] = uint64(f(ctx, mod, stack))
 }
 
 // stubFunction stubs for GrainLang per #271.
--- a/imports/wasi_snapshot_preview1/wasi.go
+++ b/imports/wasi_snapshot_preview1/wasi.go
@@ -241,7 +241,7 @@ func writeOffsetsAndNullTerminatedValues(ctx context.Context, mem api.Memory, va
 // result. The returned value will be written back to the stack at index zero.
 type wasiFunc func(ctx context.Context, mod api.Module, params []uint64) Errno
 
-// Call implements the same method as documented on wasiFunction.
+// Call implements the same method as documented on api.GoModuleFunc.
 func (f wasiFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 	// Write the result back onto the stack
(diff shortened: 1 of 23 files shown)


codefromthecrypt:
found a more efficient way, basically pass a re-slice of the stack as that's what engines use anyway. this ensures no allocation of slices (except the slice header for re-slicing)

```diff
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -307,6 +307,18 @@ type Function interface {
 // The Module parameter is the calling module, used to access memory or
 // exported functions. See GoModuleFunc for an example.
 //
+// The stack is includes any parameters encoded according to their ValueType.
+// Its length is the max of parameter or result length. When there are results,
+// write them in order beginning at index zero.
+//
+// Here's a typical way to read three parameters and write back one.
+//
+//     // read parameters off the stack in index order
+//     argv, argvBuf := uint32(stack[0]), uint32(stack[1])
+//
+//     // write results back to the stack in index order
+//     stack[0] = uint64(ErrnoSuccess)
+//
 // This function can be non-deterministic or cause side effects. It also
 // has special properties not defined in the WebAssembly Core specification.
 // Notably, this uses the caller's memory (via Module.Memory). See
@@ -317,25 +329,28 @@ type Function interface {
 // idiomatic as they can map go types to ValueType. This type is exposed for
 // those willing to trade usability and safety for performance.
 type GoModuleFunction interface {
-       Call(ctx context.Context, mod Module, params, results []uint64)
+       Call(ctx context.Context, mod Module, stack []uint64)
 }
 
 // GoModuleFunc is a convenience for defining an inlined function.
 //
 // For example, the following returns a uint32 value read from parameter zero:
 //
-//     api.GoModuleFunc(func(ctx context.Context, mod api.Module, params, results []uint64) {
-//             ret, ok := mod.Memory().ReadUint32Le(ctx, uint32(params[0]))
+//     api.GoModuleFunc(func(ctx context.Context, mod api.Module, stack []uint64) {
+//             offset := uint32(params[0]) // read the parameter from the stack
+//
+//             ret, ok := mod.Memory().ReadUint32Le(ctx, offset)
 //             if !ok {
 //                     panic("out of memory")
 //             }
-//             results[0] = uint64(ret)
-//     }
-type GoModuleFunc func(ctx context.Context, mod Module, params, results []uint64)
+//
+//             results[0] = uint64(ret) // add the result back to the stack.
+//     })
```

imports/wasi_snapshot_preview1/wasi.go
```diff
@@ -215,39 +215,36 @@ func exportFunctions(builder wazero.HostModuleBuilder) {
 	exporter.ExportHostFunc(sockShutdown)
 }
 
-// Declare constants to avoid slice allocation per call.
```
codefromthecrypt, line 218: these are no longer needed as we can write to the stack directly now.

internal/engine/interpreter/interpreter.go
```diff
@@ -827,6 +827,9 @@ func (ce *callEngine) call(ctx context.Context, m *wasm.CallContext, tf *functio
 
 	ce.callFunction(ctx, m, tf)
 
+	// This returns a safe copy of the results, instead of a slice view. If we
+	// returned a re-slice, the caller could accidentally or purposefully
+	// corrupt the stack of subsequent calls.
 	results = wasm.PopValues(ft.ResultNumInUint64, ce.popValue)
```
codefromthecrypt, line 833: interpreter already made a copy, so I just documented it.

internal/wasm/gofunc_test.go
```diff
@@ -267,15 +267,27 @@ func Test_callGoFunc(t *testing.T) {
...
 			case api.GoModuleFunction:
-				results = code.GoFunc.(api.GoModuleFunction).Call(testCtx, callCtx, tc.inputParams)
+				code.GoFunc.(api.GoModuleFunction).Call(testCtx, callCtx, stack)
 			default:
 				t.Fatal("unexpected type.")
 			}
+
+			var results []uint64
+			if resultLen > 0 {
+				results = stack[0:resultLen]
```
mathetake, line 289:
```suggestion
				results = stack[:resultLen]
```

imports/wasi_snapshot_preview1/wasi.go
```diff
@@ -215,39 +215,36 @@ func exportFunctions(builder wazero.HostModuleBuilder) {
...
 
-	return errnoSuccess
+	return ErrnoSuccess
+}
+
+// wasiFunc special cases that all WASI functions return a single Errno
+// result. The returned value will be written back to the stack at index zero.
+type wasiFunc func(ctx context.Context, mod api.Module, params []uint64) Errno
+
+// Call implements the same method as documented on wasiFunction.
```
mathetake, line 244: I can't find any interface named `wasiFunction`

mathetake (approved):

imports/wasi_snapshot_preview1/wasi.go
codefromthecrypt: I maka the pasta

```suggestion
// Call implements the same method as documented on api.GoModuleFunc.
```

imports/wasi_snapshot_preview1/wasi.go
```diff
@@ -215,39 +215,36 @@ func exportFunctions(builder wazero.HostModuleBuilder) {
...
 
-	return errnoSuccess
+	return ErrnoSuccess
+}
+
+// wasiFunc special cases that all WASI functions return a single Errno
+// result. The returned value will be written back to the stack at index zero.
+type wasiFunc func(ctx context.Context, mod api.Module, params []uint64) Errno
+
+// Call implements the same method as documented on api.GoModuleFunc.
```
codefromthecrypt, line 244:
```suggestion
// Call implements the same method as documented on api.GoModuleFunction.
```

codefromthecrypt:
looks like vagrant/freebsd is at least temporarily broke. let's see if this is consistent or not

codefromthecrypt:

PR #863, opened 2022-11-28
Update spectest v2 to latest

mathetake:


codefromthecrypt (approved):
Thanks for keeping wazero passing!

PR #865, opened 2022-11-28
wasi: adds fd_readdir

codefromthecrypt:
This adds an implementation of `fd_readdir` for WASI, which ensures a very large directory is not kept in host memory until its directory is closed.

Original implementation and test data are with thanks from @jerbob92


codefromthecrypt:
this is draft until polished and tested in a non-ad-hoc fashion. This is tricky as it is stateful in ways unlike all other WASI functions.

jerbob92:
@codefromthecrypt I think this implementation might be too strict. Let's imagine the following situation:
 - `readdir` needs to receive directory entries and calls the `getdents` syscall with a predefined buffer
 - `getdents` receives a buffer of 2770 bytes, every dirent is 277 bytes, so we can fit 10 entries in there
 - In the WASI implementation in Emscripten I create a buffer size that can also fit 10 items (24 per entry + 256 for each filename), making it 2800 bytes
 - In the WASI implementation it just fills up the 2800 opportunistically, and since it's able to use the buffer more efficiently than the C `dirent` struct, it's very good possible that it will fit way more than 10 items, let's say every entry has a short name, making it only use 100 bytes per entry, not 280, meaning we can now fit 28 entries in there, not 10
 - Back in Emscripten land, `getdents` loops through the returned entries by Wazero, but at item 10, we have exhausted our own buffer, so we return early and ignore the other 18, the cookie of our last entry is 10 (or 11? not sure)
 - Every `readdir` returns 1 of the 10 items from `getdents` until that has exhausted the buffer too, then it calls `getdents` again until it returns nothing
 - However, the cookie that `readdir`  just sent to `getdents` is now 10 (or 11), because that is the last item it processed, because the other 18 didn't fit into the buffer
 - `getdents` goes back to Wazero WASI with the cookie of 10 (or 11), and now it triggers this check:
 
```go
	// Validate the cookie sent is correct. We cannot seek backwards, so the
	// last position is the only valid one.
	if f.ReadDirCount != cookie {
		return
```

Hope it's clear what I wanted to explain :)

codefromthecrypt:
I'm trying hard to have our first implementation not more complicated than the other runtimes in the first commit. It is literally me who has to write all these tests. Is there some limit that can be placed to avoid buffering an unbounded directory and still work with your patch?

codefromthecrypt:
per slack what we're going to do is keep the last page of directory listings around in case someone asks for something besides the last cookie. We will still die if the cookie is before that.

I'll document the rationale for it including options if this becomes important later. However, most use of the directory list will just scroll forward and will unlikely need to lookback too many entries. That's the main assumption.

Other alternatives are hard to reach right now, especially without tossing fs.FS for some other model that works with seeking for all supported operating systems.

Thanks @jerbob92 for the tenacity on finding a way out here.

codefromthecrypt:
current status: impl is in a shape I think is as testable as it can get without spending too much time for 1pct gains

still needed: docs and warnings about wasi-libc stuff ex that they manipulate cookie manually which is crazy as well some notes on how to identify when something is looking too far back. these are notes for myself as I'll translate it into plain english as a part of the change

codefromthecrypt:
here is the wasi program I use for ad-hoc testing:

```c
#include <dirent.h>
#include <stdio.h>

int main(void) {
  DIR *d;
  struct dirent *dir;
  d = opendir(".");
  if (d) {
    while ((dir = readdir(d)) != NULL) {
      printf("%s\n", dir->d_name);
    }
    closedir(d);
  }
  return(0);
}
```

codefromthecrypt:
thanks, though I think there's still an incompatilibilty with wasi-libc as running your diff, I still end up with not enough entries. ex I do a list on a directory with 223 entries and this returns 119.

This is the same as the base commit, where when in the base commit as noted
`bufused >= direntsLen` changes to `bufused > direntsLen`, all files are read.

I would say for sure existing compilers are a must work, ex clang and zig, they can't break and these are the ones that must complete. A variant based on interpreting text or other repos, are a nice to have. I'll fix the bug I left last night and hopefully we can get to the emscripten patch after.

codefromthecrypt:
there's another important point, which is that if we do fill exactly bufLen with a truncated entry, which certainly I'll help do, it won't necessarily be conveniently 24 bytes. the remaining space could be 1-N bytes, where anything less than 8 is totally unusable. You can't even read the cookie from that. I'm just noting this to make sure no one tries to implement anything based on a truncated entry.

codefromthecrypt:
so next steps.. I'll get this working and most importantly place a test that uses clang code and works on a paginated listing. This will ensure any things we do to make non-standard pagination doesn't at the same time break clang. If I have time, I'll do zig as well. Meanwhile, any changes once that's done would have to be presented in ways that both compile and execute unit tests as well those integration tests. This will keep the theory interpretation down, as really the docs are extremely short.

https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-fd_readdirfd-fd-buf-pointeru8-buf_len-size-cookie-dircookie---errno-size

The only thing we can anchor to sustainably is existing compilers and to conserve time of everyone we shouldn't really interpret other people's code manually until the above tests are in and working, as "proof is in the pudding" as they say.

TL;DR;

NEXT STEP: get clang and maybe also zig integration tests working and a part of this PR so experiments on custom pagination don't break existing compilers by accident.

IDEAL FOR SOMEONE ELSE: implement readdir in tinygo :) https://github.com/tinygo-org/tinygo/issues/3087

codefromthecrypt:
on thing that makes no sense to me, but seems to be the missing document, is it feels like wasi fd_readdir is supposed to write past buf_len as needed to ensure an entry of the size dirent is there https://github.com/WebAssembly/wasi-libc/blob/659ff414560721b1660a19685110e484a081c3d4/libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c#L40

This is scary because this basically assumes allocation is linear and nothing is used after buf/buf_len as the extra bytes to do this could corrupt in that case. Also, it assumes you don't run out of memory attempting to write the over-memory value. A guest could defend against this by allocating a buffer of size buf_len+24 to ensure no corruption
or out of memory errors in these cases.

codefromthecrypt:
current status:

this works with ad-hoc wasi libc (via clang) and unit tests pass. this is using the janky "write past buf_len" approach  which seems to be needed due to wasi-libc

next step:

add clang and zig integration tests for a directory list that overflows their default buffer. This makes sure the edge cases work.

codefromthecrypt:
update: looks like truncating works as wasi-libc. I changed the code to ensure we never overrun buf-len

next  step: polish and add the integration tests

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -643,11 +643,243 @@ func fdRead_shouldContinueRead(n, l uint32, err error) (bool, Errno) {
...
+	if errno != ErrnoSuccess {
+		return errno
+	}
+
+	// expect a cookie only if we are continuing a read.
+	if cookie == uint64(0) {
+		if dir.CountRead > 0 {
+			return ErrnoInval // invalid as a cookie is minimally one.
+		}
+	}
```
mathetake, line 676:
```suggestion
	if cookie == uint64(0) && dir.CountRead > 0 {
		return ErrnoInval // invalid as a cookie is minimally one.
	}
```

codefromthecrypt:
still same status

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -643,11 +643,243 @@ func fdRead_shouldContinueRead(n, l uint32, err error) (bool, Errno) {
...
+
+	// Potentially truncate it
+	copy(dirents[pos:], dirent)
+}
+
+// writeDirent writes direntSize bytes
+func writeDirent(buf []byte, dNext uint64, dNamlen uint32, dType bool) {
+	binary.LittleEndian.PutUint64(buf, dNext)        // d_next
+	binary.LittleEndian.PutUint64(buf[8:], 0)        // no d_ino
+	binary.LittleEndian.PutUint32(buf[16:], dNamlen) // d_namlen
```
mathetake, line 862: calling PutUintXX three times might be costly, but instead how about directly putting integers into buf as we know that it is little endian

codefromthecrypt:
like our other wasi stuff, we mostly test with stdlib and normal code approach. ex do a web search and how to list directory, that kind of code. So the tests on the way are like the cat example, where they are normal stdlib code patterns.

This change will be ready to merge when two compilers of normal stdlib patterns work. This helps ensure the work can eventually complete. Remembering fd_readdir didn't work at all, at least normal stdlib is great progress. Change after that is welcome, but not while breaking these. I'm making this change block on additional tests due to demand to have different behaviours than are needed to satisfy stdlib things I've used so far.

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -643,11 +643,245 @@ func fdRead_shouldContinueRead(n, l uint32, err error) (bool, Errno) {
...
+	entryCount := len(dir.Entries)
+	if entryCount == 0 { // there was no prior call
+		if cookie != 0 {
+			errno = ErrnoInval // invalid as we haven't sent that cookie
+		}
+		return
+	}
+
+	countRead := int(dir.CountRead)
+	if int(cookie)-countRead == 1 {
```
mathetake, line 744: cookie is 64-bit so there's a possibility `int(cookie)` can be negative, though I'm not sure if we want to care about that :D

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -643,11 +643,245 @@ func fdRead_shouldContinueRead(n, l uint32, err error) (bool, Errno) {
...
+// lastDirEntries is broken out from fdReaddirFn for testability.
+func lastDirEntries(dir *internalsys.ReadDir, cookie uint64) (entries []fs.DirEntry, errno Errno) {
+	entryCount := len(dir.Entries)
+	if entryCount == 0 { // there was no prior call
+		if cookie != 0 {
+			errno = ErrnoInval // invalid as we haven't sent that cookie
+		}
+		return
+	}
+
```
mathetake, line 742: you maybe want to create `c := int(cookie)` to do casting only one time

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -643,11 +643,245 @@ func fdRead_shouldContinueRead(n, l uint32, err error) (bool, Errno) {
...
+			// When the next entry is truncated, only write the dirent.
+			if lenRemaining < direntSize {
+				truncatedEntryLen = direntSize - lenRemaining
+			}
+			break
+		}
+
+		// At this point, this entry can fit within bufLen.
+		bufused += entryLen
+		lenRemaining -= entryLen
```
mathetake, line 816: could you teach me why this won't be negative..

mathetake (approved):

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: maybe we can change the arg type to int64

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: we did `if lenRemaining < entryLen` before this line so worst case this is zero right?

codefromthecrypt:
status: added clang example, but it has a breaking test. one entry is repeated

codefromthecrypt:
I'll add zig after dinner then look into the broken test

codefromthecrypt:
current status: clang and zig stlib examples are in, and both fail on large directories. At least clang fails due to a dupe. Once these are solved, we can add more test cases that use stdlibs or merge.

next action: figure out why there are dupes and fix them!

note: I'm personally done for the day

codefromthecrypt:
update:

last commit fixes c thanks to @jerbob92 as well adds tests and clarifies what d_next is. zig doesn't work

zig seems to not handle the edge case when the last entry doesn't fit in buf_len. this needs some analysis cc @mathetake as maybe wasmer works by writing past the end (like an older version of this commit did, but we shouldn't have if only considering the spec). making a guess here

https://github.com/ziglang/zig/blob/209a0d2a83c4d60d776a6346752732f35d146a45/lib/std/fs.zig#L831
vs
https://github.com/WebAssembly/wasi-libc/blob/659ff414560721b1660a19685110e484a081c3d4/libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c#L44

next action:

handle review comments and add another language here, despite offline confirmation that emscripten works. ideally we have two here that work.. maybe rust is the only one that implements fd_readdir as tinygo doesn't yet.

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -643,11 +643,246 @@ func fdRead_shouldContinueRead(n, l uint32, err error) (bool, Errno) {
...
+	}
+
+	// expect a cookie only if we are continuing a read.
+	if cookie == uint64(0) && dir.CountRead > 0 {
+		return ErrnoInval // invalid as a cookie is minimally one.
+	}
+
+	// Ensure we have the max directory entries that can be serialized as
+	// dirents, knowing file names come after.
+	maxDirEntries := int(bufLen/direntSize + 1)
```
mathetake, line 678: what if bufLen is not a multiple of direntSize?

codefromthecrypt: I'll change the comment. These are the amount of entries we read into the hosts' struct and we keep the leftovers. We need to over-fetch as we have no idea relationship between file count and size to encode them.

codefromthecrypt: updated!

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: made cookie signed

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: made cookie signed

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: it can be negative if the name size is a very large number. I think we can skip in that case, though testing it would be extremely expensive I think

codefromthecrypt: I think I have a way to test it.. we are only reading the size of each entry and nothing else

codefromthecrypt: got a simper way, nm

codefromthecrypt:
ok I think feedback is addressed @mathetake. I'm switching to try to get a rust example in here now.

codefromthecrypt:
ok swapped rust in for zig and we can try for zig (and tinygo) post merge. @mathetake PTAL!

codefromthecrypt:
finally green! woot

jerbob92:
Great this is in :partying_face: 

Now that I have all syscalls implemented in Emscripten that I need for pdfium (and validated that they work with Wazero), I can open the PR at Emscripten and get that in. 

I will add tests for Emscripten once that happened so that I can add examples that are actually reproducible with an official Emscripten build.

PR #866, opened 2022-11-28
Optimizes memory instance creation

mathetake:
some tests started failing, so this is wip


mathetake:
this is actually trying to violate the Wasm spec! Closing :D

PR #867, opened 2022-11-29
Adds logging.NewHostLoggingListenerFactory

codefromthecrypt:
This formalizes something I hand edit constantly, which is a listener that only shows logging when it is a host function or callable by the host (exported). This is important as some compilers create a large amount of overhead functions, which can make thousands of lines around a host call you are looking for.

Note: we do all exported functions, not just those actually called by the host, as we can't statically analyze implementations of functions implemented in go to see what they might call.


experimental/logging/log_listener_example_test.go
```diff
@@ -55,3 +56,43 @@ func Example_newLoggingListenerFactory() {
 	//	<== ESUCCESS
 	//<-- ()
 }
+
+// This example shows how to see all function calls, including between host
+// functions.
+func Example_newLoggingListenerFactory() {
+	// Set context to one that has an experimental listener
+	ctx := context.WithValue(context.Background(), experimental.FunctionListenerFactoryKey{}, logging.NewLoggingListenerFactory(os.Stdout))
```
mathetake, line 64: shouldn't this NewHostLoggingListenerFactory?

experimental/logging/log_listener_example_test.go
```diff
@@ -55,3 +56,43 @@ func Example_newLoggingListenerFactory() {
...
+
+	// Output:
+	//--> listener.rand(len=4)
+	//	--> listener.wasi_rand(len=4)
+	//		==> wasi_snapshot_preview1.random_get(buf=4,buf_len=4)
+	//		<== ESUCCESS
+	//		==> wasi_snapshot_preview1.random_get(buf=8,buf_len=4)
+	//		<== ESUCCESS
+	//	<-- ()
+	//<-- ()
```
mathetake, line 97: seems all logged

codefromthecrypt: yeah this is distracting but intentional. I wanted an integration test to show the old logger will show all calls still.

experimental/logging/log_listener_example_test.go
```diff
@@ -19,10 +19,11 @@ import (
...
 var listenerWasm []byte
 
-// This is a very basic integration of listener. The main goal is to show how it is configured.
-func Example_newLoggingListenerFactory() {
+// This is a very basic integration of listener. The main goal is to show how
+// it is configured.
+func Example_newHostLoggingListenerFactory() {
 	// Set context to one that has an experimental listener
-	ctx := context.WithValue(context.Background(), experimental.FunctionListenerFactoryKey{}, logging.NewLoggingListenerFactory(os.Stdout))
+	ctx := context.WithValue(context.Background(), experimental.FunctionListenerFactoryKey{}, logging.NewHostLoggingListenerFactory(os.Stdout))
```
codefromthecrypt, line 26: so this is using `logging.NewHostLoggingListenerFactory`, and it doesn't show the intermediate guest function.

experimental/logging/testdata/listener.wat
```diff
@@ -25,4 +16,9 @@
     i32.const 3 call_indirect (type 0) ;; element 3, func type 0
     drop ;; errno
   )
+
+  (func $rand (export "rand") (param $len i32)
+    local.get 0 ;; buf_len
+    call $wasi_rand
```
codefromthecrypt, line 22: this is the guest call which should be skipped when host only mode, though the fact `wasi.random_get` was initiated by the exported `rand` should show in both cases.

codefromthecrypt (commented):
adding missing notes. sorry!

experimental/logging/log_listener.go
```diff
@@ -12,17 +12,38 @@ import (
...
+
+type loggingListenerFactory struct {
+	writer   io.Writer
+	hostOnly bool
+}
 
 // NewListener implements the same method as documented on
 // experimental.FunctionListener.
 func (f *loggingListenerFactory) NewListener(fnd api.FunctionDefinition) experimental.FunctionListener {
+	if f.hostOnly && fnd.GoFunction() == nil && len(fnd.ExportNames()) == 0 {
```
mathetake, line 44: maybe need comment on this line especially on how `len(fnd.ExportNames()) == 0` relates to `functions called or implemented by the host`

mathetake (approved):

experimental/logging/log_listener.go
codefromthecrypt: good idea will do

PR #869, opened 2022-11-29
compiler: adds support for FunctionListeners

mathetake:

### DIFF
--- a/experimental/listener_example_test.go
+++ b/experimental/listener_example_test.go
@@ -58,7 +58,7 @@ func Example_customListenerFactory() {
 	// Set context to one that has an experimental listener
 	ctx := context.WithValue(context.Background(), FunctionListenerFactoryKey{}, u)
 
-	r := wazero.NewRuntimeWithConfig(ctx, wazero.NewRuntimeConfigInterpreter())
+	r := wazero.NewRuntime(ctx)
 	defer r.Close(ctx) // This closes everything this Runtime created.
 
 	wasi_snapshot_preview1.MustInstantiate(ctx, r)
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -125,7 +125,7 @@ type (
 		// passed to the Call API.
 		ctx context.Context
 		// contextStack is a stack of contexts which is pushed and popped by function listeners.
-		// This is used and modified when there's function listeners.
+		// This is used and modified when there are function listeners.
 		contextStack *contextStack
 	}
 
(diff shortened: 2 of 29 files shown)


internal/engine/compiler/engine.go
```diff
@@ -455,15 +464,21 @@ func (e *engine) CompileModule(ctx context.Context, module *wasm.Module) error {
 
 	importedFuncs := module.ImportFuncCount()
 	funcs := make([]*code, len(module.FunctionSection))
+	ln := len(listeners)
 	for i, ir := range irs {
+		var withListener bool
+		if i < ln {
+			withListener = listeners[i] != nil
+		}
+
```
mathetake, line 473: this makes it possible for us to assemble machine code on the function listeners calling only when a function has non-nil listeners, and in the machine code generation, we call the listener on preamble and function return as opposed to the case of interpreter where it's calling it on the caller side.

mathetake:
ok switched all tests using listener to compilers, and passed all of them on arm64!

mathetake:
passed all tests both on amd64 and arm64!

experimental/listener_example_test.go
```diff
@@ -58,7 +58,7 @@ func Example_customListenerFactory() {
 	// Set context to one that has an experimental listener
 	ctx := context.WithValue(context.Background(), FunctionListenerFactoryKey{}, u)
 
-	r := wazero.NewRuntimeWithConfig(ctx, wazero.NewRuntimeConfigInterpreter())
+	r := wazero.NewRuntime(ctx)
```
codefromthecrypt, line 61: yay

internal/engine/compiler/engine.go
```diff
@@ -118,6 +119,20 @@ type (
 
 		// initialFn is the initial function for this call engine.
 		initialFn *function
+
+		// ctx is the context.Context passed to all the host function calls.
+		// This is modified when there's a function listener call, otherwise it's always the context.Context
+		// passed to the Call API.
+		ctx context.Context
+		// contextStack is a stack of contexts which is pushed and popped by function listeners.
+		// This is used and modified when there's function listeners.
```
codefromthecrypt, line 128:
```suggestion
		// This is used and modified when there are function listeners.
```

codefromthecrypt (approved):
looking good. we might need some doc clarification.

So, I think listeners need to be in context for compile. In the compile engine, must the count match at runtime, or is the listener value only read at compile time and re-used thereafter?

Can we can tighten it up to only reading the listeners during compile and that's it, ex proved by passing context.Background to instantiate and also compile? If not should we or how much effort would take to do that?

codefromthecrypt:
clarification: looks like this is compile config, so we can add a test to prove it and clarify on the context key that it applies to CompileModule or things that imply that. ignore my other comment otherwise!

codefromthecrypt (approved):
coolio

PR #870, opened 2022-11-29
Ensures listeners only bound to compile time objects

mathetake:
This stops binding `experimental.FunctionListener` to wasm.FunctionIntance and removes the need to pass listeners to `Store.Instantiate`. This clarifies more that Function Listeners are compile-time options and prevents the accidental dependence on the instantiation-time objects.


codefromthecrypt (approved):
I really like this

PR #872, opened 2022-11-30
Merge wazero-fuzz into internal/integration_test

mathetake:
To make it easy for anyone interested to run fuzzing, not only myself, this merges the code in https://github.com/tetratelabs/wazero-fuzz repository into integration tests.

Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>

PR #873, opened 2022-11-30
Consolidates table init logics

mathetake:
This fixes a bug found by the fuzzer in #872.
As of the latest v2 spec, we also need to initialize tables with ref.null external references, and
therefore table initialization is no longer an engine-specific thing. Therefore, this conslidates
a table init logic scattered across interpreter and compiler, and do it in the internal/wasm pkg.


codefromthecrypt (approved):
less and more coherence code. thanks!

PR #874, opened 2022-11-30
Fixes applyTableInits for many extern null inits

mathetake:


internal/wasm/store.go
```diff
@@ -198,7 +198,8 @@ func (m *ModuleInstance) applyTableInits(tables []*TableInstance, tableInits []t
 	for _, init := range tableInits {
 		table := tables[init.tableIndex]
 		references := table.References
-		if int(init.offset)+len(init.functionIndexes) > len(references) {
+		if int(init.offset)+len(init.functionIndexes) > len(references) ||
+			int(init.offset)+init.nullExternRefCount > len(references) {
```
mathetake, line 202: this is the fix

codefromthecrypt: thanks for the note

codefromthecrypt (approved):
lots of great stuff lately. thanks!

PR #875, opened 2022-11-30
wasi: try to fix fd_readdir logic and tests

jerbob92:
This tries to fix up the logic as mush as possible. I will explain inline why I made the comments I did.

I could not get Zig working, I'm convinced this is due to their WASI implementation not being correct, they do not take into account that the entry might be truncated and this causes out bounds reads: https://github.com/ziglang/zig/blob/209a0d2a83c4d60d776a6346752732f35d146a45/lib/std/fs.zig#L831

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -710,11 +710,9 @@ func fdReaddirFn(ctx context.Context, mod api.Module, params []uint64) Errno {
 	// Now, write entries to the underlying buffer.
 	if bufused > 0 {
 
-		// Cookie is zero when there is no cookie. Now that we know we will
-		// write entries, ensure the first one is one not zero.
-		if cookie == uint64(0) {
-			cookie = 1
-		}
+		// Cookie is the index of the next file in the list, so it should
+		// always be one higher than the requested cookie.
+		cookie++
 
 		dirents, ok := mem.Read(ctx, buf, bufused)
 		if !ok {
@@ -780,15 +778,19 @@ const direntSize = uint32(24)
 //
 // See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#fd_readdir
 // See https://github.com/WebAssembly/wasi-libc/blob/659ff414560721b1660a19685110e484a081c3d4/libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c#L44
-func maxDirents(entries []fs.DirEntry, bufLen uint32) (bufused, direntCount, truncatedEntryLen uint32) {
+func maxDirents(entries []fs.DirEntry, bufLen uint32) (bufused, direntCount uint32, writeTruncatedEntry bool) {
 	lenRemaining := bufLen
 	for _, e := range entries {
 		nameLen := uint32(len(e.Name()))
 		entryLen := direntSize + nameLen
 
-		if lenRemaining == 0 {
-			// We've hit exactly the end of bufLen, so cannot write the next
+		if lenRemaining < direntSize {
+			// We don't have enough space in bufLen for another struct,
 			// entry. A caller who wants more will retry.
+
+			// bufused != bufLen means more entries exist, which is the case
+			// when the next entry is larger than bytes remaining.
+			bufused = bufLen
 			break
 		}
 
@@ -804,10 +806,9 @@ func maxDirents(entries []fs.DirEntry, bufLen uint32) (bufused, direntCount, tru
 			// when the next entry is larger than bytes remaining.
 			bufused = bufLen
 
-			// When the next entry is truncated, only write the dirent.
-			if lenRemaining < direntSize {
-				truncatedEntryLen = direntSize - lenRemaining
-			}
+			// We do have enough space to write the header, this value will be
+			// passed on to writeDirents to only write the header for this entry.
+			writeTruncatedEntry = true
 			break
 		}
 
@@ -825,7 +826,7 @@ func maxDirents(entries []fs.DirEntry, bufLen uint32) (bufused, direntCount, tru
 func writeDirents(
 	entries []fs.DirEntry,
 	entryCount uint32,
-	truncatedEntryLen uint32,
+	writeTruncatedEntry bool,
 	dirents []byte,
 	cookie uint64,
 ) {
@@ -842,7 +843,7 @@ func writeDirents(
 		cookie++
 	}
 
-	if truncatedEntryLen == 0 {
+	if !writeTruncatedEntry {
 		return
 	}
 
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -1062,6 +1062,7 @@ func Test_fdReaddir(t *testing.T) {
 		buf, bufLen     uint32
 		cookie          uint64
 		expectedMem     []byte
+		expectedMemSize int
 		expectedBufused uint32
 		expectedReadDir *internalsys.ReadDir
 	}{
@@ -1104,10 +1105,27 @@ func Test_fdReaddir(t *testing.T) {
 
 				return &internalsys.FileEntry{File: dir}
 			},
-			buf: 0, bufLen: 24, // length is too short
+			buf: 0, bufLen: 23, // length is too short for header
 			cookie:          0,
-			expectedBufused: 24, // == bufLen which is the size of the dirent
+			expectedBufused: 23, // == bufLen which is the size of the dirent
 			expectedMem:     nil,
+			expectedReadDir: &internalsys.ReadDir{
+				CountRead: 2,
+				Entries:   testDirEntries[:2],
+			},
+		},
+		{
+			name: "can't read name",
+			dir: func() *internalsys.FileEntry {
+				dir, err := fdReadDirFs.Open("dir")
+				require.NoError(t, err)
+
+				return &internalsys.FileEntry{File: dir}
+			},
+			buf: 0, bufLen: 24, // length is too short for name
+			cookie:          0,
+			expectedBufused: 24,                       // == bufLen which is the size of the dirent
+			expectedMem:     dirent1[:len(dirent1)-1], // header without name
 			expectedReadDir: &internalsys.ReadDir{
 				CountRead: 3,
 				Entries:   testDirEntries,
@@ -1147,7 +1165,7 @@ func Test_fdReaddir(t *testing.T) {
 				}
 			},
 			buf: 0, bufLen: 26, // length is long enough for one more.
-			cookie:          2,  // previous d_next
+			cookie:          1,  // previous d_next
 			expectedBufused: 26, // length to read exactly one more.
 			expectedMem:     dirent2,
 			expectedReadDir: &internalsys.ReadDir{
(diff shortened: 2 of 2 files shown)


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -710,18 +710,16 @@ func fdReaddirFn(ctx context.Context, mod api.Module, params []uint64) Errno {
...
 	if bufused > 0 {
 
-		// Cookie is zero when there is no cookie. Now that we know we will
-		// write entries, ensure the first one is one not zero.
-		if cookie == uint64(0) {
-			cookie = 1
-		}
+		// d_next is the index of the next file in the list, so it should
+		// always be one higher than the requested cookie.
+		d_next := cookie + 1
```
jerbob92, line 715: I renamed this to d_next so that it's more clear we're not talking about the initial cookie anymore, but about the d_next value of the entries that we're writing away. 

I changed this to always be +1 because d_next is the cookie of the next record, so if you ask cal fd_readdir with cookie 7, you should never get an entry that has a d_next value of 7.

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -780,15 +778,19 @@ const direntSize = uint32(24)
...
-func maxDirents(entries []fs.DirEntry, bufLen uint32) (bufused, direntCount, truncatedEntryLen uint32) {
+func maxDirents(entries []fs.DirEntry, bufLen uint32) (bufused, direntCount uint32, writeTruncatedEntry bool) {
 	lenRemaining := bufLen
 	for _, e := range entries {
 		nameLen := uint32(len(e.Name()))
 		entryLen := direntSize + nameLen
 
-		if lenRemaining == 0 {
-			// We've hit exactly the end of bufLen, so cannot write the next
+		if lenRemaining < direntSize {
```
jerbob92, line 787: If we do not have enough space for a new dirent header, we want to break out of the loop and say that we have exhausted the whole buffer, not just when we have exactly 0 bytes left in the buffer.

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -780,15 +778,19 @@ const direntSize = uint32(24)
...
+			// We don't have enough space in bufLen for another struct,
 			// entry. A caller who wants more will retry.
+
+			// bufused != bufLen means more entries exist, which is the case
+			// when the next entry is larger than bytes remaining.
+			bufused = bufLen
```
jerbob92, line 793: According to the docs we should do this when there was no space in the buffer for more entries.

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -804,10 +806,9 @@ func maxDirents(entries []fs.DirEntry, bufLen uint32) (bufused, direntCount, tru
 			// when the next entry is larger than bytes remaining.
 			bufused = bufLen
 
-			// When the next entry is truncated, only write the dirent.
-			if lenRemaining < direntSize {
-				truncatedEntryLen = direntSize - lenRemaining
-			}
+			// We do have enough space to write the header, this value will be
+			// passed on to writeDirents to only write the header for this entry.
+			writeTruncatedEntry = true
```
jerbob92, line 811: I changed this to a boolean because we don't really need to keep track of the remaining size. 
I removed the condition because we always want to write away the header if there is room enough for it. If there is no room for the header we already break out in the statement above.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1104,10 +1105,27 @@ func Test_fdReaddir(t *testing.T) {
 
 				return &internalsys.FileEntry{File: dir}
 			},
-			buf: 0, bufLen: 24, // length is too short
+			buf: 0, bufLen: 23, // length is too short for header
```
jerbob92, line 1108: I have changed this to 23, because 24 is actually enough to contain the header.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1104,10 +1105,27 @@ func Test_fdReaddir(t *testing.T) {
...
-			expectedBufused: 24, // == bufLen which is the size of the dirent
+			expectedBufused: 23, // == bufLen which is the size of the dirent
 			expectedMem:     nil,
+			expectedReadDir: &internalsys.ReadDir{
+				CountRead: 2,
+				Entries:   testDirEntries[:2],
+			},
+		},
+		{
+			name: "can't read name",
```
jerbob92, line 1118: I have added this test, because this is probably actually what the test above wanted to test? This tests if we can write away the header when we have exactly 24 bytes in the buffer.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1147,7 +1165,7 @@ func Test_fdReaddir(t *testing.T) {
 				}
 			},
 			buf: 0, bufLen: 26, // length is long enough for one more.
-			cookie:          2,  // previous d_next
+			cookie:          1,  // previous d_next
```
jerbob92, line 1168: I have changed this to 1 because if you send in cookie 1, you expect to get entry 2.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1172,9 +1190,60 @@ func Test_fdReaddir(t *testing.T) {
 				}
 			},
 			buf: 0, bufLen: 30, // length is longer than the second entry
-			cookie:          2,  // previous d_next
-			expectedBufused: 30, // length to read exactly one more.
-			expectedMem:     append(dirent2, dirent3[0:4]...),
+			cookie:          1,  // previous d_next
```
jerbob92, line 1193: I have changed this to 1 because if you send in cookie 1, you expect to get entry 2.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1172,9 +1190,60 @@ func Test_fdReaddir(t *testing.T) {
...
+			expectedBufused: 30, // length to read some more, but not enough for a header.
+			expectedMem:     append(dirent2),
+			expectedMemSize: len(dirent2), // we do not want to compare the full buffer since we don't know what the leftover 4 bytes will contain.
+			expectedReadDir: &internalsys.ReadDir{
+				CountRead: 3,
+				Entries:   testDirEntries[1:],
+			},
+		},
+		{
+			name: "read second and another header",
```
jerbob92, line 1203: I have added this check to validate the logic when there is space for 1 record but not enough for a second but it does have enough for the header of the second.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1172,9 +1190,60 @@ func Test_fdReaddir(t *testing.T) {
...
+			cookie:          1,  // previous d_next
+			expectedBufused: 50, // length to read exactly one more.
+			expectedMem:     append(dirent2, dirent3[0:24]...),
+			expectedReadDir: &internalsys.ReadDir{
+				CountRead: 3,
+				Entries:   testDirEntries[1:],
+			},
+		},
+		{
+			name: "read second and third",
```
jerbob92, line 1228: This test was added to validate that it works to get exactly 2 items.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1197,7 +1266,32 @@ func Test_fdReaddir(t *testing.T) {
 				}
 			},
 			buf: 0, bufLen: 27, // length is long enough for one more.
-			cookie:          3,  // previous d_next
+			cookie:          2,  // previous d_next
```
jerbob92, line 1269: I have changed this to 2 because if you send in cookie 2, you expect to get entry 3.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1197,7 +1266,32 @@ func Test_fdReaddir(t *testing.T) {
...
+			expectedBufused: 27, // length to read exactly one more.
+			expectedMem:     dirent3,
+			expectedReadDir: &internalsys.ReadDir{
+				CountRead: 3,
+				Entries:   testDirEntries[2:],
+			},
+		},
+		{
+			name: "read third and beyond",
```
jerbob92, line 1278: I have added this test to validate what happens if it hits the end of the entries with a larger buffer.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1237,7 +1331,10 @@ func Test_fdReaddir(t *testing.T) {
 			require.True(t, ok)
 
 			if tc.expectedMem != nil {
-				require.Equal(t, tc.expectedMem, mem)
+				if tc.expectedMemSize == 0 {
```
jerbob92, line 1334: I had to add/change this for the test case "read second and a little more", that results in a bufused of the complete buffer because it fits one item but not another header but since the last bytes of the buffer could be anything because it's not zero'ed out I had to come up with a way to exclude these last bytes from the comparison.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1146,9 +1164,9 @@ func Test_fdReaddir(t *testing.T) {
 					},
 				}
 			},
-			buf: 0, bufLen: 26, // length is long enough for one more.
-			cookie:          2,  // previous d_next
-			expectedBufused: 26, // length to read exactly one more.
+			buf: 0, bufLen: 26, // length is long enough for exactly second.
+			cookie:          1,  // d_next of first
```
jerbob92, line 1168: I have changed cookie 2 to 1, because when you send in cookie 1, you expect the second entry.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1171,10 +1189,61 @@ func Test_fdReaddir(t *testing.T) {
 					},
 				}
 			},
-			buf: 0, bufLen: 30, // length is longer than the second entry
-			cookie:          2,  // previous d_next
-			expectedBufused: 30, // length to read exactly one more.
-			expectedMem:     append(dirent2, dirent3[0:4]...),
+			buf: 0, bufLen: 30, // length is longer than the second entry, but not long enough for a header.
+			cookie:          1,  // d_next of first
```
jerbob92, line 1193: I have changed cookie 2 to 1, because when you send in cookie 1, you expect the second entry.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1196,9 +1265,34 @@ func Test_fdReaddir(t *testing.T) {
 					},
 				}
 			},
-			buf: 0, bufLen: 27, // length is long enough for one more.
-			cookie:          3,  // previous d_next
-			expectedBufused: 27, // length to read exactly one more.
+			buf: 0, bufLen: 27, // length is long enough for exactly third.
+			cookie:          2,  // d_next of second.
```
jerbob92, line 1269: I have changed cookie 3 to 2, because when you send in cookie 2, you expect the third entry.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1171,10 +1189,61 @@ func Test_fdReaddir(t *testing.T) {
...
+			expectedBufused: 30, // length to read some more, but not enough for a header, so buf was exhausted.
+			expectedMem:     append(dirent2),
+			expectedMemSize: len(dirent2), // we do not want to compare the full buffer since we don't know what the leftover 4 bytes will contain.
+			expectedReadDir: &internalsys.ReadDir{
+				CountRead: 3,
+				Entries:   testDirEntries[1:],
+			},
+		},
+		{
+			name: "read second and header of third",
```
jerbob92, line 1203: I have added this check to validate the logic when there is space for 1 record but not enough for a second but it does have enough for the header of the second.

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: great comment

codefromthecrypt:
excellent work. I'll merge this and we may have to pull zig from the PR to merge the parent. It may be wise to make a rust one instead so that two languages in this repo pass, even if zig doesn't yet. I'll own that

PR #877, opened 2022-12-01
binary: adds custom section decoder

mathetake:
Disabled by default, and will be only enabled true if DWARF trace option is on (which will be added later)


codefromthecrypt (approved):
cool. I think @pkedy has some custom sections in some of his things, though ack we shouldn't decode by default as it could burn time

ref https://github.com/tetratelabs/wabin/pull/4

mathetake:
oh yeah actually I did copy-pasta from the PR :D

PR #879, opened 2022-12-02
fuzz: adds validation target for maybe invalid module compilation

mathetake:

### DIFF
--- a/internal/integration_test/fuzz/fuzz/fuzz_targets/validation.rs
+++ b/internal/integration_test/fuzz/fuzz/fuzz_targets/validation.rs
@@ -1,0 +1,36 @@
+#![no_main]
+
+use libfuzzer_sys::arbitrary::{Result, Unstructured};
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    drop(run(data));
+});
+
+fn run(data: &[u8]) -> Result<()> {
+    // Create the random source.
+    let mut u = Unstructured::new(data);
+
+    // Generate the random module via wasm-smith, but MaybeInvalidModule.
+    let module: wasm_smith::MaybeInvalidModule = u.arbitrary()?;
+    let module_bytes = module.to_bytes();
+    let wat_bytes = wasmprinter::print_bytes(&module_bytes).unwrap();
+
+    // Pass the randomly generated module to the wazero library.
+    unsafe {
+        validate(
+            module_bytes.as_ptr(),
+            module_bytes.len(),
+            wat_bytes.as_ptr(),
+            wat_bytes.len(),
+        );
+    }
+
+    // We always return Ok as inside require_no_diff, we cause panic if the binary is interesting.
+    Ok(())
+}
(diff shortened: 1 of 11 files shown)


internal/wasm/func_validation.go
```diff
@@ -1464,6 +1464,9 @@ func (m *Module) validateFunctionWithMaxStackValues(
 				valueTypeStack.push(p)
 			}
 		} else if op == OpcodeEnd {
+			if len(controlBlockStack) == 0 {
```
mathetake, line 1467: found a bug actually :D

internal/integration_test/fuzz/wazerolib/nodiff.go
```diff
@@ -0,0 +1,211 @@
+package main
```
mathetake, line 1: this is just moved from lib.go

internal/wasm/func_validation.go
mathetake: oh maybe I need to add regression test case

internal/integration_test/fuzz/fuzz/fuzz_targets/validation.rs
```diff
@@ -0,0 +1,29 @@
...
+fuzz_target!(|data: &[u8]| {
+    drop(run(data));
+});
+
+fn run(data: &[u8]) -> Result<()> {
+    // Create the random source.
+    let mut u = Unstructured::new(data);
+
+    // Generate the random module via wasm-smith, but MaybeInvalidModule.
+    let module: wasm_smith::MaybeInvalidModule = u.arbitrary()?;
```
codefromthecrypt, line 15: so this isn't sure if it will make an invalid one or not.. interesting :D

codefromthecrypt (approved):
agree failing earlier is better especially as wasm gives us a lot of validation opportunity

internal/wasm/func_validation.go
codefromthecrypt: yeah thanks I should have nagged you, but you self-nagged. excellent

internal/integration_test/fuzz/wazerolib/validate_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"testing"
+)
+
+// TestReRunFailedValidateCase re-runs the failed case specified by WASM_BINARY_NAME in testdata directory.
+func TestReRunFailedValidateCase(t *testing.T) {
+	binaryPath := os.Getenv("WASM_BINARY_PATH")
+
+	wasmBin, err := os.ReadFile(binaryPath)
+	if err != nil {
+		t.Fatal(err)
```
codefromthecrypt, line 14: should this skip or fail?

internal/integration_test/fuzz/fuzz/fuzz_targets/validation.rs
mathetake: yeah not sure until we pass it to wazero :D

PR #883, opened 2022-12-03
validation: not panic with redundant Else instrs

mathetake:


codefromthecrypt (approved):
thanks for adding the test!

PR #881, opened 2022-12-02
Adds support for DWARF based stack traces

mathetake:
This implements the support for DWARF-based stack traces. This new feature is enabled by
the experimental API `experimental.DWARFBasedStackTraceEnabled` and passing that context
to the runtime at compilation time. 

For example, let's have the following program:
```go
package main

func main() { a() }
func a() { b() }
func b() { c() }
func c() { panic("NOOOOOOOOOOOOOOO") }
```

and use `experimental.DWARFBasedStackTraceEnabled` during the compilation:

```go
ctx := experimental.WithDWARFBasedStackTrace(ctx)
compiled, _ := r.CompileModule(ctx, wasmBinary)
```

Now, instantiate the compiled module, which results in a runtime error:
 
```go
_, err = r.InstantiateModule(context.Background(), compiled, wazero.NewModuleConfig())
require.Error(t, err)
fmt.Println(err)
```

and we have the output like the following:

```
module[] function[_start] failed: wasm error: unreachable
wasm stack trace:
	.runtime._panic(i32)
		0x16e2: /opt/homebrew/Cellar/tinygo/0.26.0/src/runtime/runtime_tinygowasm.go:73:6
	.c()
		0x190b: /Users/mathetake/wazero/internal/testing/dwarftestdata/testdata/main.go:19:7
	.b()
		0x1901: /Users/mathetake/wazero/internal/testing/dwarftestdata/testdata/main.go:14:3
	.a()
		0x18f7: /Users/mathetake/wazero/internal/testing/dwarftestdata/testdata/main.go:9:3
	.main.main()
		0x18ed: /Users/mathetake/wazero/internal/testing/dwarftestdata/testdata/main.go:4:3
	.runtime.run()
		0x18cc: /opt/homebrew/Cellar/tinygo/0.26.0/src/runtime/scheduler_none.go:26:10
	._start()
		0x18b6: /opt/homebrew/Cellar/tinygo/0.26.0/src/runtime/runtime_wasm_wasi.go:22:5
```

from which we can easily infer how the execution reached the runtime error because the trace comes with the source code locations.

closes #58

PR #884, opened 2022-12-04
Delete obsolete integration test from wat parser

mathetake:


codefromthecrypt (commented):
this shows how fast or slow our binary codec is, though. is there another reason besides wat to delete this?

mathetake:
`EncodeModule` is not maintained anymore AFAIK plus this example `newExample()` is too unrealistic to measure the perf of decoder. I would replace it with TinyGo one and delete `newExample()` instead. on it.

mathetake:
FYI `newExample()` is defined to line with Wat impl which is no longer existent.

mathetake:
done

mathetake:
@codefromthecrypt PTAL aa I don't want to fix this `TestExampleUpToDate` in #881

codefromthecrypt (approved):
cool I just checked and we don't use the encoder outside tests, so sounds good to me. in fact maybe we should delete the encoder.

mathetake:
lol just in time

codefromthecrypt:
maybe first we can move the encoder to a testing package..

PR #885, opened 2022-12-05
wasi: optimizes args/environ parsing

codefromthecrypt:
While most compilers will only read args/environ once, tools like WAGI make heavy use of environment, possibly dozens of long variables. This optimizes both args and environ for this reason and also to setup for optimizing other functions.

Here are the notable changes:
* eagerly coerce to byte slices instead of strings
* re-use null terminated length for writing values
* avoid loops that call mem.WriteXXX internally

Before:
```
Benchmark_ArgsEnviron
Benchmark_ArgsEnviron/args_get
Benchmark_ArgsEnviron/args_get-16         	 6147727	       187.9 ns/op	      72 B/op	       5 allocs/op
Benchmark_ArgsEnviron/args_sizes_get
Benchmark_ArgsEnviron/args_sizes_get-16   	11869351	       100.7 ns/op	      24 B/op	       2 allocs/op
Benchmark_ArgsEnviron/environ_get
Benchmark_ArgsEnviron/environ_get-16      	 6202508	       191.0 ns/op	      72 B/op	       5 allocs/op
Benchmark_ArgsEnviron/environ_sizes_get
Benchmark_ArgsEnviron/environ_sizes_get-16         	11792656	        99.71 ns/op	      24 B/op	       2 allocs/op
```

After:
```
Benchmark_ArgsEnviron
Benchmark_ArgsEnviron/args_get
Benchmark_ArgsEnviron/args_get-16         	 9542246	       119.2 ns/op	      24 B/op	       2 allocs/op
Benchmark_ArgsEnviron/args_sizes_get
Benchmark_ArgsEnviron/args_sizes_get-16   	11963056	       105.7 ns/op	      24 B/op	       2 allocs/op
Benchmark_ArgsEnviron/environ_get
Benchmark_ArgsEnviron/environ_get-16      	 9742885	       123.0 ns/op	      24 B/op	       2 allocs/op
Benchmark_ArgsEnviron/environ_sizes_get
Benchmark_ArgsEnviron/environ_sizes_get-16         	12100869	        97.06 ns/op	      24 B/op	       2 allocs/op
```

### DIFF
--- a/imports/wasi_snapshot_preview1/args.go
+++ b/imports/wasi_snapshot_preview1/args.go
@@ -59,7 +59,7 @@ var argsGet = &wasm.HostFunc{
 func argsGetFn(ctx context.Context, mod api.Module, params []uint64) Errno {
 	sysCtx := mod.(*wasm.CallContext).Sys
 	argv, argvBuf := uint32(params[0]), uint32(params[1])
-	return writeOffsetsAndNullTerminatedValues(ctx, mod.Memory(), sysCtx.Args(), argv, argvBuf)
+	return writeOffsetsAndNullTerminatedValues(ctx, mod.Memory(), sysCtx.Args(), argv, argvBuf, sysCtx.ArgsSize())
 }
 
 // argsSizesGet is the WASI function named functionArgsSizesGet that reads
@@ -108,7 +108,8 @@ func argsSizesGetFn(ctx context.Context, mod api.Module, params []uint64) Errno 
 	mem := mod.Memory()
 	resultArgc, resultArgvLen := uint32(params[0]), uint32(params[1])
 
-	// Write the Errno back to the stack
+	// argc and argv_len offsets are not necessarily sequential, so we have to
+	// write them independently.
 	if !mem.WriteUint32Le(ctx, resultArgc, uint32(len(sysCtx.Args()))) {
 		return ErrnoFault
 	}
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
ps for those interested in inlining, there was discussion but closed without doing anything on go:inline https://github.com/golang/go/issues/21536 If you build with `-gcflags="-m"` you can see things including what was inlined.

imports/wasi_snapshot_preview1/args.go
```diff
@@ -108,7 +108,8 @@ func argsSizesGetFn(ctx context.Context, mod api.Module, params []uint64) Errno
 	mem := mod.Memory()
 	resultArgc, resultArgvLen := uint32(params[0]), uint32(params[1])
 
-	// Write the Errno back to the stack
+	// argc and argv_len offsets are not necessarily sequential, so we have to
+	// write them independently.
 	if !mem.WriteUint32Le(ctx, resultArgc, uint32(len(sysCtx.Args()))) {
```
codefromthecrypt, line 113: ps this doesn't inline. I'll check if anything gets better by inlining manually in another commit

codefromthecrypt: difference is inside the error range. I think generally we'll only optimize when there's a WriteUint32Le inside a loop

mathetake (approved):

PR #886, opened 2022-12-05
fuzz: enable imports in differential fuzzing 

mathetake:


mathetake:
found the failing case with the locally run fuzz

```
(module
  (import "" "s" (memory (;0;) 0 5))
  (import "" "" (global (;0;) funcref))
  (global (;1;) (mut f64) f64.const -0x1.1ff6861000008p-652 (;=-0.00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006019210420568196;))
  (global (;2;) (mut funcref) global.get 0)
  (global (;3;) (mut i32) i32.const 1000)
  (export "\00\00\00\00\00\00\14\14\14\14\14" (memory 0))
  (export "" (global 0))
  (export "2" (global 1))
  (export "3" (global 2))
  (elem (;0;) declare externref (ref.null extern) (ref.null extern) (ref.null extern))
  (data (;0;) "\ff\ff\ff\ff")
)
```

mathetake:
rebased onto #888

mathetake:
ok now looks stable with the fix in #888

internal/integration_test/fuzz/wazerolib/nodiff.go
```diff
@@ -77,6 +114,107 @@ func requireNoDiff(wasmBin []byte, requireNoError func(err error)) {
 	}
 }
 
+func ensureDummyImports(r wazero.Runtime, origin *wasm.Module, requireNoError func(err error)) (skip bool) {
```
codefromthecrypt, line 117: what "dummy imports" means might be useful to document here.

codefromthecrypt (approved):
love to see death of TODO

PR #887, opened 2022-12-05
Adds ResultNames to HostFunctionBuilder and FunctionDefinition

codefromthecrypt:
This adds ResultNames to HostFunctionBuilder and FunctionDefinition which helps for multi-results or special-cased ones.

End users can access result names in `FunctionDefinition.ResultNames` or set for their own host functions via
`HostFunctionBuilder.WithResultNames`. This change adds them for all built-in functions where result names help.

Most notably, GOOS=js uses `ProxyFunc` to allow logging when a function returns multiple results. Before, the results were returned without names: e.g. `11231,1` and now they are named like `n=11231,ok=1`.

We soon plan to allow more visibility in WASI, for example, logging results that will write to memory offsets. This infrastructure makes it possible to do that.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -121,6 +121,12 @@ type HostFunctionBuilder interface {
 	// Note: When defined, names must be provided for all parameters.
 	WithParameterNames(names ...string) HostFunctionBuilder
 
+	// WithResultNames defines optional result names of the function
+	// signature, e.x. "errno"
+	//
+	// Note: When defined, names must be provided for all results.
+	WithResultNames(names ...string) HostFunctionBuilder
+
 	// Export exports this to the HostModuleBuilder as the given name, e.g.
 	// "random_get"
 	Export(name string) HostModuleBuilder
@@ -206,7 +212,7 @@ type hostModuleBuilder struct {
 	r            *runtime
 	moduleName   string
 	nameToGoFunc map[string]interface{}
-	funcToNames  map[string][]string
+	funcToNames  map[string]*wasm.HostFuncNames
 }
 
 // NewHostModuleBuilder implements Runtime.NewHostModuleBuilder
@@ -215,16 +221,17 @@ func (r *runtime) NewHostModuleBuilder(moduleName string) HostModuleBuilder {
 		r:            r,
 		moduleName:   moduleName,
 		nameToGoFunc: map[string]interface{}{},
-		funcToNames:  map[string][]string{},
+		funcToNames:  map[string]*wasm.HostFuncNames{},
 	}
 }
 
 // hostFunctionBuilder implements HostFunctionBuilder
 type hostFunctionBuilder struct {
-	b          *hostModuleBuilder
-	fn         interface{}
-	name       string
-	paramNames []string
+	b           *hostModuleBuilder
+	fn          interface{}
+	name        string
+	paramNames  []string
+	resultNames []string
 }
 
 // WithGoFunction implements HostFunctionBuilder.WithGoFunction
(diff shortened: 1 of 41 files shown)


builder.go
```diff
@@ -121,6 +121,12 @@ type HostFunctionBuilder interface {
 	// Note: When defined, names must be provided for all parameters.
 	WithParameterNames(names ...string) HostFunctionBuilder
 
+	// WithResultNames defines optional result names of the function
+	// signature, e.x. "errno"
+	//
+	// Note: When defined, names must be provided for all results.
+	WithResultNames(names ...string) HostFunctionBuilder
```
codefromthecrypt, line 128: here's how someone can add result names if they like

experimental/logging/log_listener.go
```diff
@@ -47,7 +47,18 @@ func (f *loggingListenerFactory) NewListener(fnd api.FunctionDefinition) experim
 		!exported { // not callable by the host
 		return nil
 	}
-	return &loggingListener{writer: f.writer, fnd: fnd, isWasi: fnd.ModuleName() == "wasi_snapshot_preview1"}
+
+	// special-case formatting of WASI error number until there's a generic way
```
codefromthecrypt, line 51: there's one wasi function which has no result, and future change will allow us to see multiple results. so I made the errno field a position instead of a boolean.

experimental/logging/log_listener.go
```diff
@@ -96,9 +109,9 @@ func (l *loggingListener) writeIndented(before bool, err error, vals []uint64, i
 		l.writeFuncEnter(&message, vals)
 	} else { // after
 		if l.fnd.GoFunction() != nil {
-			message.WriteString("<== ")
+			message.WriteString("<==")
 		} else {
-			message.WriteString("<-- ")
+			message.WriteString("<--")
```
codefromthecrypt, line 114: this stops writing `<-- ()` for no results in favor of `<--`

imports/assemblyscript/assemblyscript_test.go
```diff
@@ -132,8 +132,8 @@ func TestSeed(t *testing.T) {
 	require.Equal(t, `
 --> proxy.seed()
 	==> env.~lib/builtins/seed()
-	<== (4.958153677776298e-175)
-<-- (4.958153677776298e-175)
```
codefromthecrypt, line 136: this is another style change to not paren wrap commonplace of single results

imports/assemblyscript/assemblyscript.go
```diff
@@ -272,6 +272,7 @@ var seed = &wasm.HostFunc{
 	ExportNames: []string{functionSeed},
 	Name:        "~lib/builtins/seed",
 	ResultTypes: []api.ValueType{f64},
+	ResultNames: []string{"rand"},
```
codefromthecrypt, line 275: I put result names for anything somewhat interesting. For example, the function name here is `seed` but it might not be intuitive that the seed value is a random number (some use clocks etc).

imports/wasi_snapshot_preview1/args.go
```diff
@@ -44,25 +44,15 @@ const (
...
-	Name:        functionArgsGet,
-	ParamTypes:  []api.ValueType{i32, i32},
-	ParamNames:  []string{"argv", "argv_buf"},
-	ResultTypes: []api.ValueType{i32},
-	Code: &wasm.Code{
-		IsHostFunction: true,
-		GoFunc:         wasiFunc(argsGetFn),
-	},
-}
+var argsGet = newHostFunc(argsGetName, argsGetFn, []api.ValueType{i32, i32}, "argv", "argv_buf")
```
codefromthecrypt, line 47: to pay for the logic in this change, I refactored out a helper function to define the wasi funcs here

internal/gojs/runtime.go
```diff
@@ -114,9 +114,10 @@ func nanotime1(ctx context.Context, mod api.Module, stack []uint64) {
 //
 // See https://github.com/golang/go/blob/go1.19/src/runtime/sys_wasm.s#L188
 var Walltime = spfunc.MustCallFromSP(false, &wasm.HostFunc{
-	ExportNames: []string{functionWalltime},
-	Name:        functionWalltime,
+	ExportNames: []string{walltimeName},
+	Name:        walltimeName,
 	ResultTypes: []api.ValueType{i64, i32},
+	ResultNames: []string{"sec", "nsec"},
```
codefromthecrypt, line 120: currently gojs is the only thing that returns multiple results. However, a future change will do similar in WASI, and this is infrastructure for that.

codefromthecrypt (commented):
added some change highlights

experimental/logging/log_listener.go
mathetake: sg

mathetake (approved):

PR #888, opened 2022-12-06
Fixes FuncRef global initialization with imported globals 

mathetake:
A bug was found in #886 where if a module imports a FuncRef type global and uses it as an initial value for a locally defined FuncRef global.


internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -343,3 +345,38 @@ func Test874(t *testing.T) {
 		require.NoError(t, err)
 	})
 }
+
+func Test888(t *testing.T) {
```
codefromthecrypt, line 349: lucky number

codefromthecrypt (approved):
good when fixing something makes it simpler

internal/integration_test/fuzzcases/fuzzcases_test.go
mathetake: in japan lucky number is 777

codefromthecrypt: off by 111 problem

PR #890, opened 2022-12-06
wasi: proxy functions to that natural results are visible

codefromthecrypt:
Right now, developing and troubleshooting wasi host functions are difficult as they are required to be WebAssembly 1.0 compatible. Specifically, they can only return a single result, and that's always used for errno. Other results are passed as out-parameters (memory offsets to write results), and their values are not very useful in practice.

This change proxies functions so that we can see their results while still keeping compatibility with wasm signatures. It uses the same approach we use in gojs (`GOOS=js GOARCH=wasm go build...`). What happens is we proxy in wasm, so that the go function implementing the ABI function has natural multi-value returned.

Specifically, we get logging like this, which is better as you can see the values returned:
```diff
-       ==> wasi_snapshot_preview1.args_sizes_get(result.argc=1,result.argv_len=6)
-       <== ESUCCESS
+       --> wasi_snapshot_preview1.args_sizes_get(result.argc=1,result.argv_len=6)
+               ==> wasi_snapshot_preview1.args_sizes_get.go()
+               <== (argc=2,argv_len=5,ESUCCESS)
+       <-- ESUCCESS
```


codefromthecrypt:
will leave this draft as I refactor as many WASI functions as I can. In the process I imagine a lot of the code will generalize as opposed to re-defining N helpers manually.

imports/wasi_snapshot_preview1/wrapper.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// Determine the index of a temporary local used for i32 manipulation. This
+	// will be directly after the parameters.
+	localI32Temp := byte(len(proxy.ParamTypes))
+
+	// Get the current memory size in bytes, to validate parameters.
+	body := memBytes // stack: [memSizeBytes]
+
+	// Ensure param[0] and param[1] are valid memory offsets for storing i32.
```
codefromthecrypt, line 53: on my laptop, doing bounds validation in wasm makes it a few nanoseconds slower than before. However, I think the visibility is worth it.

An aside: not all wasi implementations actually check memory bounds, but at least one besides wazero does (wasmtime iirc does)

imports/wasi_snapshot_preview1/args_test.go
```diff
@@ -158,8 +160,8 @@ func Test_argsSizesGet_Errors(t *testing.T) {
 			argvLen: validAddress,
 			expectedLog: `
 --> proxy.args_sizes_get(result.argc=65536,result.argv_len=0)
-	==> wasi_snapshot_preview1.args_sizes_get(result.argc=65536,result.argv_len=0)
-	<== EFAULT
+	--> wasi_snapshot_preview1.args_sizes_get(result.argc=65536,result.argv_len=0)
```
codefromthecrypt, line 163: interesting thing here is you can see the validation of memory doesn't hit the host function (`-->` not `==>`)

codefromthecrypt:
next step: convert the function that proxies the function to be generic. ex claim N result parameters

codefromthecrypt:
made progress with a number of file-related calls. once I back port all the calls whose out params can fit in 8 bytes (i64), I'll go back and clean up. Particularly, I'll try to clean up logging so that results that are flags are formatted when in wasi mode, as well any obvious cruft cleanup opportunities in the main code. need another day.

mathetake:
how's the perf diff look like before vs this?

codefromthecrypt:
bench stat in progress!

```bash
$ git checkout main;go test -run='^$' -bench=. -count=5 > old.txt; git checkout wrap-wasi; go test -run='^$' -bench=. -count=5 >new.txt; benchstat old.txt new.txt
```

codefromthecrypt:
perf looks same across 5 runs. verified with bench stat because ad-hoc looked the same

```
name                               old time/op    new time/op    delta
_ArgsEnviron/args_get-16              119ns ± 2%     119ns ± 2%    ~     (p=1.000 n=5+5)
_ArgsEnviron/args_sizes_get-16        106ns ±16%      97ns ± 2%    ~     (p=0.056 n=5+5)
_ArgsEnviron/environ_get-16           123ns ± 4%     118ns ± 2%  -4.21%  (p=0.016 n=5+5)
_ArgsEnviron/environ_sizes_get-16    98.7ns ± 3%    96.7ns ± 2%    ~     (p=0.310 n=5+5)

name                               old alloc/op   new alloc/op   delta
_ArgsEnviron/args_get-16              24.0B ± 0%     24.0B ± 0%    ~     (all equal)
_ArgsEnviron/args_sizes_get-16        24.0B ± 0%     24.0B ± 0%    ~     (all equal)
_ArgsEnviron/environ_get-16           24.0B ± 0%     24.0B ± 0%    ~     (all equal)
_ArgsEnviron/environ_sizes_get-16     24.0B ± 0%     24.0B ± 0%    ~     (all equal)

name                               old allocs/op  new allocs/op  delta
_ArgsEnviron/args_get-16               2.00 ± 0%      2.00 ± 0%    ~     (all equal)
_ArgsEnviron/args_sizes_get-16         2.00 ± 0%      2.00 ± 0%    ~     (all equal)
_ArgsEnviron/environ_get-16            2.00 ± 0%      2.00 ± 0%    ~     (all equal)
_ArgsEnviron/environ_sizes_get-16      2.00 ± 0%      2.00 ± 0%    ~     (all equal)
```

imports/wasi_snapshot_preview1/wrapper.go
```diff
@@ -0,0 +1,234 @@
...
+		byteLength = 8
+	default:
+		panic(fmt.Errorf("TODO: unsupported outType: %v", outType))
+	}
+	body = append(body, wasm.OpcodeI32Const, byte(byteLength), wasm.OpcodeI32Sub) // stack: [memBytes-byteLength]
+	body = append(body, wasm.OpcodeLocalGet, byte(paramIdx))                      // stack: [memBytes-byteLength, $0]
+	body = append(body, wasm.OpcodeI32LtU, wasm.OpcodeIf, nilType)                // stack: []
+	body = append(body, wasm.OpcodeI32Const, fault)                               // stack: [efault]
+	body = append(body, wasm.OpcodeReturn, wasm.OpcodeEnd)                        // stack: []
+	return body
```
mathetake, line 215: maybe single call to append?

mathetake: same for elsewhere

codefromthecrypt: sure

imports/wasi_snapshot_preview1/wrapper.go
```diff
@@ -0,0 +1,234 @@
...
+		body = append(body, compileStore(localOffsetIdx, localValIdx, outType)...)
+	}
+
+	// Finally, add the success error code to the stack.
+	body = append(body, wasm.OpcodeI32Const, success) // stack: [success]
+	body = append(body, wasm.OpcodeEnd)
+
+	// Assign the wasm generated as the proxy's body
+	proxy.Code.Body = body
+	return &wasm.ProxyFunc{Proxy: proxy, Proxied: goFunc, CallBodyPos: fakeFuncIdxPos}
```
mathetake, line 181: just realized that we haven't checked if the function indexes replaced by ProxyFunc never exceed one-byte leb128 encoding. This PR makes it more likely to happen that we have 2-bytes long function indexes?

mathetake (approved):

imports/wasi_snapshot_preview1/wrapper.go
codefromthecrypt: It isn't enforced here, it is enforced where we make a host module. I don't think we'll run into this in the near term because this is internal and our biggest host module is wasi. We might run into it, if we make this non-internal later, or use it in emscripten which could make massive amounts of things.

from internal/wasm/host.go
```
			proxiedIdx := len(funcNames)
			if proxiedIdx > maxProxiedFuncIdx {
				return errors.New("TODO: proxied funcidx larger than one byte")
			}
```

mathetake: oh cool

codefromthecrypt: so concretely, old wasi is 45 functions, and many aren't implemented. Even if we could proxy them all it would reach 90 in that module. 127 is the max. I think yeah it would be emscripten that if we start using this approach would end up hitting it. The result of that would be some work in the host module code to insert bytes for the larger func index. We did something similar for the text format back in the day, so not a big deal to do when we get to it.

codefromthecrypt:
I caused an init performance regression as I was only looking at call performance before/after. I will fix this possibly with a different design that doesn't use proxy functions. For example, passing the module memory to the logging interceptor instead. Before I thought about this approach but hesitated due to risking an interceptor corrupting memory. I think this risk is better than higher overhead all the time.

```
name                       old time/op    new time/op    delta
Allocation/Compile-16        38.3ms ± 0%    38.3ms ± 0%      ~     (p=0.556 n=4+5)
Allocation/Instantiate-16     845µs ± 0%    1666µs ± 1%   +97.29%  (p=0.008 n=5+5)
Allocation/Call-16           4.04µs ± 2%    4.00µs ± 2%      ~     (p=0.310 n=5+5)

name                       old alloc/op   new alloc/op   delta
Allocation/Compile-16        20.3MB ± 0%    20.3MB ± 0%      ~     (p=0.421 n=5+5)
Allocation/Instantiate-16     568kB ± 0%    1043kB ± 0%   +83.70%  (p=0.008 n=5+5)
Allocation/Call-16            48.0B ± 0%     48.0B ± 0%      ~     (all equal)

name                       old allocs/op  new allocs/op  delta
Allocation/Compile-16          432k ± 0%      432k ± 0%      ~     (p=1.000 n=5+5)
Allocation/Instantiate-16     6.74k ± 0%    16.59k ± 0%  +146.09%  (p=0.008 n=5+5)
Allocation/Call-16             5.00 ± 0%      5.00 ± 0%      ~     (all equal)
```

Issue #889, opened 2022-12-06
Ideas for reducing heap memory usage


anuraaga:
As discussed on Slack, I was debugging a problem that runs into GC performance limits in low-CPU environments with Go's default settings. Go's settings can be tweaked and there will likely always be some patterns of usage that stress allocations and will require tweaking e.g. `GOMEMLIMIT`, in these cases, but it could still be possible to reduce GC to lower the chance of this happening or possibly enabling applications to perform better in general. These are just ideas, and may not be good ones.

The ideas come from running this test with memory profiler. I do this by loading in GoLand and hitting the green play button, selecting run with memory profiler.

https://github.com/anuraaga/re2-go/blob/main/exec2_test.go#L19

It is instantiating several million regex's, each instantiating and closing a wasm module. The module has about 2K functions and uses 3 wasm pages of memory.

### Function table

A huge number of the objects are `&function{}` and `&functionInstance`, allocated at these two locations (about 50/50 between the two)

https://github.com/tetratelabs/wazero/blob/main/internal/engine/compiler/engine.go#L543
https://github.com/tetratelabs/wazero/blob/main/internal/wasm/module.go#L609

The instances are always stored into a slice from what I can tell - I wonder if it would be possible to use contiguous slice, e.g. `[]function` instead of pointer slice, e.g. `[]*function` for these. If it is possible, it could dramatically reduce the number of heap objects while still accessing everything through the memory in these slices.

This would definitely make the code more fragile - most likely most logic operations will still need to work on pointer of function, with the goal being for those pointers being into a contiguous slice of memory. It can be easy in such a style to accidentally not use a pointer and break things mysteriously.

Objects:
![image](https://user-images.githubusercontent.com/198344/205823452-3acefda2-d5a3-4453-938a-7ac3d8cc4c9c.png)

### Module memory

The number of objects for memory was very low while the number of bytes is high - for context, the number of bytes in memory objects is about equal to the number of bytes in function objects (which, to me, was surprising, I didn't expect the function objects to be taking so much space but I guess it adds up). I'm not sure how Go's GC performs with respect to number of objects and bytes of objects, both should matter but I think generally GC's are more affected by the number. So optimizing module memory may not have as big an impact as function table.

[...]

PR #892, opened 2022-12-06
Store function objects in contiguous memory in heap

anuraaga:
Reference #889 

Currently, we allocate the object storing information on a function directly on the heap (using Go's `&struct{}` syntax), and store the pointers into a slice. Each of these ends up being a live object, with the number of live objects impacting the performance of the GC. We can instead allocate a slice of structs to store the information into and keep pointers into that instead. This reduces the number of live objects from equalling the number of functions to a constant one for the slice.

The biggest impact will be when the module has a lot of functions ([re2](https://github.com/anuraaga/re2-go/tree/main/wasm) compiled with clang has about 2K), but it looks like the existing initialization benchmarks show a reasonable change so hopefully that's enough to justify the PR. Creating a benchmark with many functions would probably require checking in a large wasm file and not be a great idea.

After
```
goos: darwin
goarch: arm64
BenchmarkInitialization
BenchmarkInitialization/compiler
BenchmarkInitialization/compiler-10         	   75512	     13915 ns/op
PASS
```

Before
```
goos: darwin
goarch: arm64
BenchmarkInitialization
BenchmarkInitialization/compiler
BenchmarkInitialization/compiler-10         	   44950	     25956 ns/op
PASS
```


codefromthecrypt (approved):
Nice find and comments. Interestingly we don't have any benchmarks for *heavy* wasm. Maybe we can find a way to add something like this to `vs` one day.

mathetake:
so no longer need `GOMAXPROC` with this?

mathetake (approved):

anuraaga:
@mathetake While this improved the runtime of the stress test on my laptop by about ~10% under normal conditions, unfortunately it isn't enough to allow the low `GOMAXPROCS` to pass still :( It is likely a step in that direction, but taking the performance improvement here without thinking too much about that test :) regex's are usually long lived outside my stress test, so it won't affect regex users that much either way, but hopefully will help users of the pattern that initializes a module per request in a server setting.

PR #893, opened 2022-12-06
Add a note about the offset requirement in moduleEngine struct

anuraaga:

If this code is too deep for such comments, feel free to close no worries. Happened to be something mysterious I ran into in first iterations of #892

### DIFF
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -39,6 +39,10 @@ type (
 		// name is the name the module was instantiated with used for error handling.
 		name string
 
+		// NOTE: The offset of the functions field is referenced from assembly, so
+		// adding any fields here will cause failure at runtime. They should be
+		// added to the end of the struct instead.
+
 		// functions are the functions in a module instances.
 		// The index is module instance-scoped. We intentionally avoid using map
 		// as the underlying memory region is accessed by assembly directly by using
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -39,9 +39,10 @@ type (
 		// name is the name the module was instantiated with used for error handling.
 		name string
 
-		// NOTE: The offset of the functions field is referenced from assembly, so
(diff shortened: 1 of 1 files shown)


internal/engine/compiler/engine.go
```diff
@@ -39,6 +39,10 @@ type (
 		// name is the name the module was instantiated with used for error handling.
 		name string
 
+		// NOTE: The offset of the functions field is referenced from assembly, so
+		// adding any fields here will cause failure at runtime. They should be
```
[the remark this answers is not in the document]
anuraaga: Ah that's much more precise, tweaked the comment to reference it

internal/engine/compiler/engine.go
```diff
@@ -39,6 +39,10 @@ type (
 		// name is the name the module was instantiated with used for error handling.
 		name string
 
+		// NOTE: The offset of the functions field is referenced from assembly based on
```
mathetake, line 42: so this is not moduleEngine specific thing but rather general thing about all the structs in this file.. so move this comment to https://github.com/tetratelabs/wazero/blob/main/internal/engine/compiler/engine.go#L25 ?

mathetake: basically like "be careful when you modify fields in this file!"

anuraaga: Interesting - I think if the comment isn't exactly where a problematic field would be added, it wouldn't have prevented my lengthyish debugging session 😅 But it makes sense so moved it up

mathetake: yeah understood - basically we should dup the same comment across all the fields

anuraaga: I tried adding some lightweight dup comments to increase the chance of safety without getting overblown, how does it look?

codefromthecrypt (approved):
I think this makes a lot of sense and former me probably would have noticed this. Thanks!

Issue #769, opened 2022-08-26
wazero support compile python?


mougeCM:
```go
package main

import (
	"context"
	"embed"
	_ "embed"
	"io/fs"
	"log"

	"github.com/tetratelabs/wazero"
	"github.com/tetratelabs/wazero/wasi_snapshot_preview1"
)

//go:embed testdata/env.py
var envPy embed.FS

//go:embed testdata/bin/python3.wasm
var greetWasm []byte

func main() {
	ctx := context.Background()

	// 新建一个 wasm 运行时
	r := wazero.NewRuntimeWithConfig(ctx, wazero.NewRuntimeConfig().
		WithWasmCore2())
	defer r.Close(ctx)

	// 实例化一个名为“env”的 Go 定义的模块，该模块导出一个函数以记录到控制台
	_, err := r.NewModuleBuilder("env").
		Instantiate(ctx, r)
	if err != nil {
		log.Panicln(err)
	}

	if _, err = wasi_snapshot_preview1.Instantiate(ctx, r); err != nil {
		log.Panicln(err)
	}

	code, err := r.CompileModule(ctx, greetWasm, wazero.NewCompileConfig())
	if err != nil {
		log.Panicln(err)
	}

	rooted, err := fs.Sub(envPy, "testdata")
	if err != nil {
		log.Panicln(err)
	}

	config := wazero.NewModuleConfig().WithFS(rooted)

	env := config.
		WithEnv("PYTHONHOME", "testdata/lib/python3.11").
		WithEnv("PYTHONPATH", "testdata/lib/python3.11")
	if _, err = r.InstantiateModule(ctx, code, env); err != nil {
		log.Panicln(err)
	}
}

```
I Already have a `python3.wasm`，I want to run the env.py on the `python3.wasm`

Use `wasmer run testdata/bin/python3.wasm --mapdir /::./ --env 'PYTHONHOME=testdata/lib/python3.11' --env 'PYTHONPATH=testdata/lib/python3.11' -- testdata/env.py` command is ok

But use wazero will appear error
```
2022/08/26 16:20:41 module "" closed with exit_code(1)
panic: module "" closed with exit_code(1)


goroutine 1 [running]:
log.Panicln({0xc001255e40?, 0x11b2bd0?, 0xc0000000d8?})
        /usr/local/go/src/log/log.go:399 +0x65
main.main()
[...]

PR #895, opened 2022-12-07
wasi: implements fd_fdstat_get and adds stdio support to fd_filestat_get

codefromthecrypt:
This implements fd_fdstat_get and furthers implementation of fd_fdstat_get. Notably, this does the following:

* Ensures stdio are treated as character devices.
* Ensures someone can stat the root FD (4).
* Fixes encoding as we didn't clear zeros when encoding numbers.

See #769


mathetake (commented):
maybe add integration test with a real language?

codefromthecrypt:
> maybe add integration test with a real language?

good idea, though python may be too big to check in

codefromthecrypt:
@mathetake PTAL I added a commit for stdlib tests for isatty which is the motivating change

imports/wasi_snapshot_preview1/testdata/cargo-wasi/wasi.rs
```diff
@@ -0,0 +1,39 @@
+use std::env;
+use std::fs;
+use std::io;
+use std::io::Write;
+use std::process::exit;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+
+    match args[1].as_str() {
```
codefromthecrypt, line 10: I merged these as otherwise our repo size will grow a lot as we test more. we do something similar in gojs where we compile a runner wasm once

mathetake (approved):

PR #899, opened 2022-12-07
Use shifted iota for wasiFdflags values

robbertvanginkel:
I did not find an authoratative reference in the wasi snapshot 1 docs, but based on the wasi-libc implementation [1] these should not be individual values but a bitmask.

 [1] https://github.com/WebAssembly/wasi-libc/blob/3a261b03809018ceae294e9062cde0758b55b61a/libc-bottom-half/headers/public/wasi/api.h#L891-L921

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -190,7 +190,7 @@ func fdFdstatGetFn(ctx context.Context, mod api.Module, params []uint64) Errno {
 
 type wasiFdflags = byte // actually 16-bit, but there aren't that many.
 const (
-	wasiFdflagsNone wasiFdflags = iota
+	wasiFdflagsNone wasiFdflags = 1<<iota - 1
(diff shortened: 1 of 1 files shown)


robbertvanginkel:
Small difference I noticed while looking at the implementation attached to https://github.com/tetratelabs/wazero/issues/769.

The existing tests should cover this, and probably went unnoticed because only `wasiFdflagsAppend` was used and those values are the same.

mathetake (approved):
Confirmed that other runtime does this
- https://github.com/wasmerio/wasmer/blob/19e7e7af2727792ffae1e092e99585dc486a05b2/lib/wasi-types/src/wasi/bindings.rs#L773-L789
- https://github.com/bytecodealliance/wasmtime/blob/918debfe59d3bf4b13b6f76fe32debbe1393a7ad/crates/wasi-common/src/file.rs#L159-L165

ack we definitely need to cover this in tests, though I couldn't find any reasonable way to test them. So merging

PR #902, opened 2022-12-08
Makes log listener default to signed interpretation

codefromthecrypt:
This changes the log listener to assume a value is signed instead of unsigned, as it leads to logs looking more correct than the other way around. Notably, values that are often negative, such as seek offset, look better without a lot of downside.

discussed with @anuraaga

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -410,7 +410,7 @@ var fdFilestatSetTimes = stubFunction(
 var fdPread = newHostFunc(
 	fdPreadName, fdPreadFn,
 	[]api.ValueType{i32, i32, i32, i64, i32},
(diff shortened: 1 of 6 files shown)


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -410,7 +410,7 @@ var fdFilestatSetTimes = stubFunction(
 var fdPread = newHostFunc(
 	fdPreadName, fdPreadFn,
 	[]api.ValueType{i32, i32, i32, i64, i32},
-	"fd", "iovs", "iovs_len", "offset", "result.size",
+	"fd", "iovs", "iovs_len", "offset", "result.nread",
```
codefromthecrypt, line 413: noticed this mistook the type for the result name

anuraaga (approved):

mathetake (approved):

PR #903, opened 2022-12-08
cmd: enable DWARF by default

mathetake:


codefromthecrypt (approved):
should this re-use the dwarf wasm elsewhere (to prove it works) in the tests?

PR #906, opened 2022-12-08
Simplify namespace

ckaznocha:
This change consolidates the two maps on the namespace into one while preserving the O(1) add/delete operations. I realized after introducing the linked list the second map was no longer needed.

I've also added an atomic to track the closed state and the namespace no longer allocates a new map when being closed per suggestion by @codefromthecrypt. `addModule` has been renamed to `setModule` and its signature updated to return an error because its no longer safe to add a module who's name has not already been created. The rest of the methods which did not return an error have become no-op after the namespace is closed. Happy to revise that if they should be updated to return errors.

Added test cases for closed namespaces to ensure namespace.go still has 100% coverage.

### DIFF
--- a/internal/wasm/namespace_test.go
+++ b/internal/wasm/namespace_test.go
@@ -51,7 +51,7 @@ func TestNamespace_setModule(t *testing.T) {
 	})
 
 	t.Run("error on closed", func(t *testing.T) {
-		ns.CloseWithExitCode(context.Background(), 0) //
+		require.NoError(t, ns.CloseWithExitCode(context.Background(), 0))
 		require.Error(t, ns.setModule(m1))
 	})
 }
@@ -80,7 +80,7 @@ func TestNamespace_deleteModule(t *testing.T) {
 	})
 
 	t.Run("noop on closed", func(t *testing.T) {
-		ns.CloseWithExitCode(context.Background(), 0)
+		require.NoError(t, ns.CloseWithExitCode(context.Background(), 0))
 		ns.deleteModule(m1.Name)
 
 		require.Zero(t, len(ns.nameToNode))
@@ -100,7 +100,7 @@ func TestNamespace_module(t *testing.T) {
 	})
 
 	t.Run("namespace closed", func(t *testing.T) {
-		ns.CloseWithExitCode(context.Background(), 0)
+		require.NoError(t, ns.CloseWithExitCode(context.Background(), 0))
 		require.Nil(t, ns.module(m1.Name))
 	})
 }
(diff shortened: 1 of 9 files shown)


internal/wasm/namespace_test.go
```diff
@@ -11,31 +12,47 @@ import (
...
+		ns.nameToNode[m2.Name] = &moduleListNode{name: m2.Name}
+		require.NoError(t, ns.setModule(m2))
+		require.Equal(t, map[string]*moduleListNode{m1.Name: {name: m1.Name, module: m1}, m2.Name: {name: m2.Name, module: m2}}, ns.nameToNode)
+
+		// Doesn't affect module names
+		require.Nil(t, ns.moduleList)
+	})
+
+	t.Run("error on closed", func(t *testing.T) {
+		ns.CloseWithExitCode(context.Background(), 0) //
```
codefromthecrypt, line 54:
```suggestion
		require.NoError(t, ns.CloseWithExitCode(context.Background(), 0))
```

internal/wasm/namespace_test.go
```diff
@@ -58,9 +75,16 @@ func TestNamespace_deleteModule(t *testing.T) {
...
 
-		require.Zero(t, len(ns.modules))
-		require.Zero(t, len(ns.moduleNamesSet))
-		require.Nil(t, ns.moduleNamesList)
+		require.Zero(t, len(ns.nameToNode))
+		require.Nil(t, ns.moduleList)
+	})
+
+	t.Run("noop on closed", func(t *testing.T) {
+		ns.CloseWithExitCode(context.Background(), 0)
```
codefromthecrypt, line 83:
```suggestion
		require.NoError(t, ns.CloseWithExitCode(context.Background(), 0))
```

mathetake:
fyi there's a bug in the GHA workflows, and now fixed on the main branch, so could you rebase onto main so we can run all the jobs on the CI?

codefromthecrypt (approved):
I agree with this change. The signature difference is not public API affecting, so no big deal. Thanks for testing.

internal/wasm/namespace_test.go
```diff
@@ -74,6 +98,11 @@ func TestNamespace_module(t *testing.T) {
 	t.Run("unknown", func(t *testing.T) {
 		require.Nil(t, ns.module("unknown"))
 	})
+
+	t.Run("namespace closed", func(t *testing.T) {
+		ns.CloseWithExitCode(context.Background(), 0)
```
codefromthecrypt, line 103:
```suggestion
		require.NoError(t, ns.CloseWithExitCode(context.Background(), 0))
```

internal/wasm/namespace_test.go
ckaznocha: 🤦

PR #909, opened 2022-12-09
wasi: optimize and benchmark fd_read

codefromthecrypt:
```bash
# before change
$ go test -run='^$' -bench '^Benchmark_fdRead$' -count=5 > old.txt
# after change
$ go test -run='^$' -bench '^Benchmark_fdRead$' -count=5 > new.txt
$ benchstat old.txt new.txt
name             old time/op  new time/op  delta
_fdRead/1x8-16    157ns ± 5%   133ns ± 1%  -15.34%  (p=0.008 n=5+5)
_fdRead/2x16-16   193ns ± 6%   155ns ± 1%  -19.77%  (p=0.008 n=5+5)
```

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -638,12 +638,13 @@ func fdReadOrPread(ctx context.Context, mod api.Module, stack []uint64, isPread 
 	}
 
 	var nread uint32
-	iovsBuf, ok := mod.Memory().Read(ctx, iovs, iovsCount<<3)
+	iovsStop := iovsCount << 3 // iovsCount * 8
+	iovsBuf, ok := mod.Memory().Read(ctx, iovs, iovsStop)
 	if !ok {
 		return 0, ErrnoFault
 	}
 
-	for iovsPos := uint32(0); iovsPos < (iovsCount << 3); iovsPos += 8 {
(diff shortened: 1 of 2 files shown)


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -638,16 +638,15 @@ func fdReadOrPread(ctx context.Context, mod api.Module, stack []uint64, isPread
...
-		iov := iovs + i*8
-		offset, ok := mem.ReadUint32Le(ctx, iov)
-		if !ok {
-			return 0, ErrnoFault
-		}
-		l, ok := mem.ReadUint32Le(ctx, iov+4)
-		if !ok {
-			return 0, ErrnoFault
-		}
+	iovsBuf, ok := mod.Memory().Read(ctx, iovs, iovsCount<<3)
```
mathetake, line 641: iovsCount<<3 can be defined as a variable so that we can reuse below?(though I hope Go compiler is wise enough to optimize that)

mathetake (approved):

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: I tried both ways and perf is roughly the same, so will change here and also fdWrite

PR #910, opened 2022-12-09
wasi: benchmarks fd_readdir for both real and fake file systems

codefromthecrypt:
It is likely we'll have to special-case real files in wasi, as wasi-libc makes behavioral choices based on presence of non-zero inode data. This adds a defensive benchmark, so that we know what our performance base case is. This adds both real and fake file systems as both patterns are in use today.

```bash
$ go test -run='^$' -bench '^Benchmark_fdReaddir$'
goos: darwin
goarch: amd64
pkg: github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1
Benchmark_fdReaddir/embed.FS-16         	  457398	      2416 ns/op
Benchmark_fdReaddir/os.DirFS-16         	   45662	     26271 ns/op
PASS
ok  	github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1	35.632s
```

See https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-17 and https://github.com/WebAssembly/wasi-filesystem/issues/65

cc @jerbob92

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -709,6 +709,12 @@ func fdReaddirFn(ctx context.Context, mod api.Module, stack []uint64) (uint32, E
 	// it in such a way that becomes negative.
 	cookie := int64(stack[3])
 
+	// The bufLen must be enough to write a dirent. Otherwise, the caller can't
+	// read what the next cookie is.
+	if bufLen < direntSize {
+		return 0, ErrnoInval
+	}
+
 	// Validate the FD is a directory
 	rd, dir, errno := openedDir(ctx, fsc, fd)
 	if errno != ErrnoSuccess {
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
I added a benchmark of a continued read. This reads the same amount of dirents, but some from memory and rest from disk.

```
Benchmark_fdReaddir
Benchmark_fdReaddir/embed.FS
Benchmark_fdReaddir/embed.FS-16                    	  505809	      2402 ns/op
Benchmark_fdReaddir/embed.FS_-_continued
Benchmark_fdReaddir/embed.FS_-_continued-16        	  586935	      2089 ns/op
Benchmark_fdReaddir/os.DirFS
Benchmark_fdReaddir/os.DirFS-16                    	   44936	     26617 ns/op
Benchmark_fdReaddir/os.DirFS_-_continued
Benchmark_fdReaddir/os.DirFS_-_continued-16        	  245131	      4880 ns/op
```

mathetake (approved):

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -707,14 +710,19 @@ func fdReaddirFn(ctx context.Context, mod api.Module, stack []uint64) (uint32, E
...
+	rd, dir, errno := openedDir(ctx, fsc, fd)
 	if errno != ErrnoSuccess {
 		return 0, errno
 	}
 
 	// expect a cookie only if we are continuing a read.
-	if cookie == 0 && dir.CountRead > 0 {
-		return 0, ErrnoInval // invalid as a cookie is minimally one.
+	if cookie == 0 {
+		// TODO: When bufLen < 8 we can't ever pass a cookie > 0 later, as the
```
jerbob92, line 720: I think it makes no sense to support partial structs write. The given bufLen should always give at least room for 1 struct (without the filename), so 24 bytes. If the room left is less than 24 bytes I don't think it should write away another struct (that's also what maxDirents does).

So in my opinion, bufLen < 24 = ErrnoInval

codefromthecrypt: sounds great to me!

codefromthecrypt:
Thanks for the help again @jerbob92. I'll implement the fstat next, since we can see how much it costs now.

PR #917, opened 2022-12-13
wasi: allows stat on root file descriptor (3)

codefromthecrypt:
Before, we didn't allow a real stat on the root file descriptor. Now, those that pass fs.ReadDirFS will return the stat of the root file (which is implemented by a open against ".").

This also simplifies logic as we always have a file representing root, even if faked.

### DIFF
--- a/experimental/fs.go
+++ b/experimental/fs.go
@@ -9,13 +9,16 @@ import (
 )
 
 // WithFS overrides fs.FS in the context-based manner. Caller needs to take
-// responsibility for closing the filesystem.
+// responsibility for closing the returned api.Closer.
 //
 // Note: This has the same effect as the same function on wazero.ModuleConfig.
-func WithFS(ctx context.Context, fs fs.FS) (context.Context, api.Closer) {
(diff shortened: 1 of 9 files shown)


experimental/fs.go
```diff
@@ -9,13 +9,16 @@ import (
 )
 
 // WithFS overrides fs.FS in the context-based manner. Caller needs to take
-// responsibility for closing the filesystem.
+// responsibility for closing the returned api.Closer.
 //
 // Note: This has the same effect as the same function on wazero.ModuleConfig.
-func WithFS(ctx context.Context, fs fs.FS) (context.Context, api.Closer) {
+func WithFS(ctx context.Context, fs fs.FS) (context.Context, api.Closer, error) {
```
codefromthecrypt, line 15: note this breaks the experimental API a bit, though we may in the future see if we can do without this. e.g. if instantiation performance is good enough, it would be faster for everyone to not have to check the Go context prior to any filesystem operation cc @knqyf263

mathetake (approved):

PR #919, opened 2022-12-13
wasi: enforce fd_preXXX functions only work on pre-opened files

codefromthecrypt:
At the moment, the only pre-open support we have is the file system itself (root a.k.a. / or file-descriptor 3). We may in the future add the ability to pre-open sockets, but in any case, this is where we are today.

This change hardens logic around fd_preXXX functions, ensuring they only work on actual pre-opens. This also fixes the path returned in filestat as we sometimes returned a full path, when typically the basename is the only part that can be returned.

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -596,7 +596,7 @@ returned by `fd_prestat_dir_name` for file descriptor 3 (STDERR+1).
 
 See
  * https://github.com/WebAssembly/wasi-libc/blob/a02298043ff551ce1157bc2ee7ab74c3bffe7144/libc-bottom-half/sources/preopens.c
- * https://github.com/ziglang/zig/blob/master/lib/std/fs/wasi.zig#L50-L53
+ * https://github.com/ziglang/zig/blob/9cb06f3b8bf9ea6b5e5307711bc97328762d6a1d/lib/std/fs/wasi.zig#L50-L53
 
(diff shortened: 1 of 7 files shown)


codefromthecrypt:
added some RATIONALE thanks @mathetake for zig stuff

RATIONALE.md
```diff
@@ -592,8 +594,9 @@ not used again after initialization.
 wazero currently supports only one pre-opened file, "/" and so that is the name
 returned by `fd_prestat_dir_name` for file descriptor 3 (STDERR+1).
 
-See https://github.com/WebAssembly/wasi-libc/blob/a02298043ff551ce1157bc2ee7ab74c3bffe7144/libc-bottom-half/sources/preopens.c
-See https://github.com/ziglang/zig/blob/master/lib/std/fs/wasi.zig#L50-L53
+See
+ * https://github.com/WebAssembly/wasi-libc/blob/a02298043ff551ce1157bc2ee7ab74c3bffe7144/libc-bottom-half/sources/preopens.c
+ * https://github.com/ziglang/zig/blob/master/lib/std/fs/wasi.zig#L50-L53
```
mathetake, line 599: master will change really fast, so permalink would be better

codefromthecrypt: true!

PR #920, opened 2022-12-13
std.fs.wasi.preopensAlloc is no longer needed

mathetake:


codefromthecrypt (approved):
excellent.. this was confusing and bloated the wasm

PR #926, opened 2022-12-15
Removes importedFn

mathetake:
I think this is a remnant since a while and no we don't need it 


codefromthecrypt (approved):
yeah I think you are right. it must have been back when we allowed direct calling of host functions

PR #934, opened 2022-12-16
Deletes encode caches

mathetake:
encoding is no longer something we optimize for 


codefromthecrypt (approved):
agree

PR #935, opened 2022-12-16
wasi: detect if stdio are char devices instead of assuming

### LINKED ISSUE #928: stdout always detected as a character device
After 78ac85cce1d `stdout` is always detected as a character device. Is this expected?
This breaks my usage of `wazero` in [RethinkRAW](https://rethinkraw.com).

**To Reproduce**

```sh
git clone --branch atty https://github.com/ncruces/wazero-bugs
cd wazero-bugs
go run main.go
```

This is the offending code in `dcraw`:
https://github.com/ncruces/dcraw/blob/8dcf1ff24ad0b5a09c8b0c1030d983c09d159284/dcraw.c#L10314-L10317

**Expected behavior**

IMO, `isatty` should not blindly return true in all cases. I'm capturing the output of a tool to a `byte.Buffer`. It's expected that it is binary. I can patch `dcraw`, but other tools (e.g. `gzip`) will similarly complain.

- Go version: go1.19.4
- wazero Version: 78ac85cce1d674c1c47c7ff2dacbfbc2e3a3e2b4
- Host architecture: amd64
- Runtime mode: compiler

codefromthecrypt:
This uses ioctl syscalls or appropriate alternative, to detect if stdin/out/err are character devices or not. This caches the result, to ensure performance is ok at runtime as executing stat can approach microsecond overhead.

Fixes #928

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -144,6 +144,8 @@ var fdFdstatGet = newHostFunc(fdFdstatGetName, fdFdstatGetFn, []api.ValueType{i3
 // fdFdstatGetFn cannot currently use proxyResultParams because fdstat is larger
 // than api.ValueTypeI64 (i64 == 8 bytes, but fdstat is 24).
 func fdFdstatGetFn(_ context.Context, mod api.Module, params []uint64) Errno {
+	fsc := mod.(*wasm.CallContext).Sys.FS()
+
 	fd, resultFdstat := uint32(params[0]), uint32(params[1])
 
 	// Ensure we can write the fdstat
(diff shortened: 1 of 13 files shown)


mathetake (approved):

codefromthecrypt:
RethinkRaw tests pass now. I will try python https://github.com/ncruces/wazero-bugs

codefromthecrypt:
python takes a while to build. I used our tty tests instead and it works.

```bash
 2022-12-16 16:03:58 ⌚  Adrians-MacBook-Pro in ~/oss/wazero/imports/wasi_snapshot_preview1/testdata/zig-cc
± |tty {15} ✓| → wazero run wasi.wasm stat
stdin isatty: true
stdout isatty: true
stderr isatty: true
/ isatty: false

 2022-12-16 16:04:02 ⌚  Adrians-MacBook-Pro in ~/oss/wazero/imports/wasi_snapshot_preview1/testdata/zig-cc
± |tty {15} ✓| → wazero run wasi.wasm stat 2>&-
stdin isatty: true
stdout isatty: true
stderr isatty: false
/ isatty: false
```

PR #940, opened 2022-12-17
Adds gojs.MustInstantiate to avoid conflicts

### LINKED ISSUE #939: Support wazero.Namespace with gojs
I'd like to be able to run multiple wasm modules concurrently and ideally have them share the compilation cache. I noticed that the docs indicate that a cache directory should only be accessed by a single runtime, but that runtime has the concept of `wazero.Namespaces` to instantiate and execute independent wasm modules without conflict.

I've figured out how to do this with `wasi_snapshot_preview1`, where I can build the wasi module with a runtime and Instantiate it into multiple namespaces:
```
var r wazero.Runtime
var config wazero.ModuleConfig
var module wazero.CompiledModule

ns := r.NewNamespace(ctx)
defer ns.Close(ctx) // todo: errcheck

// wasi_snapshot_preview1
if _, err := wasi_snapshot_preview1.NewBuilder(r).Instantiate(ctx, ns); err != nil {
	return err
}
_, err = ns.InstantiateModule(ctx, module, config)
```

However for `gojs`, there's only a `Run()` method which requires a `wazero.Runtime`:
```
// gojs
err = gojs.Run(ctx, r, module, config) // can't use a ns
```

gojs to have a similar builder interface, and support instantiating the module into a namespace:
```
if _, err := gojs.NewBuilder(r).Instantiate(ctx, ns); err != nil {
	return err
}
_, err = ns.InstantiateModule(ctx, module, config)
```

I currently have a mutex around my usage of  `gojs.Run`, which isn't great but does the job for now :). Alternatively I could create new runtimes for each gojs call, but not sharing compilation cache is a big downside.

**Additional context**
Looking at the code for gojs, I think it should be possible to give it an api surface similar to `wasi_snapshot_preview1` to allow using it with namespaces? This would entail some breaking changes (or just new APIs?) so I figured I'd open an issue to ask if making `gojs` support usage like `wasi_snapshot_preview1` makes sense before I look at making a PR!

codefromthecrypt:
This separate host from guest instantiation in ways similar to other imports such as emscripten. Doing so allows parallel use of gojs.Run, provided the ModuleConfig has been assigned a unique name (e.g. via an atomic number).

Fixes #939


codefromthecrypt:
@robbertvanginkel please give this a shot and let us know!

mathetake (approved):

codefromthecrypt:
rebased, and benchmarks look as expected where there is less overhead when the "go" module isn't instantiated every run.

```bash
# before change
$ go test -run='^$' -bench '^Benchmark_main$' -count=5 > old.txt
# after change
$ go test -run='^$' -bench '^Benchmark_main$' -count=5 > new.txt
$ benchstat old.txt new.txt
name               old time/op    new time/op    delta
_main/gojs.Run-16    9.65ms ± 4%    9.45ms ± 2%    ~     (p=0.548 n=5+5)

name               old alloc/op   new alloc/op   delta
_main/gojs.Run-16    23.1MB ± 0%    22.9MB ± 0%  -0.72%  (p=0.008 n=5+5)

name               old allocs/op  new allocs/op  delta
_main/gojs.Run-16     80.4k ± 0%     77.7k ± 0%    ~     (p=0.079 n=4+5)
```

PR #948, opened 2022-12-20
Adds benchmarks from lib sodium

codefromthecrypt:
This adds benchmarks from lib sodium, but the results are hard to interpret. We may need to recompile them using zig so that we can export a function instead of using WASI to do it. Right now, we can't really see performance of functions because other runtimes aren't designed to re-instantiate like ours.

Wasmer (old as wasmer-go isn't released this year)

```
BenchmarkShorthash
BenchmarkShorthash/Compile
BenchmarkShorthash/Compile-16         	  511204	      2119 ns/op	      11 B/op	       1 allocs/op
BenchmarkShorthash/Instantiate
BenchmarkShorthash/Instantiate-16     	      37	  29909106 ns/op	   66361 B/op	     475 allocs/op
```

Wasmtime (latest)

```
BenchmarkShorthash
BenchmarkShorthash/Compile
BenchmarkShorthash/Compile-16         	  179979	      6485 ns/op	       8 B/op	       1 allocs/op
BenchmarkShorthash/Instantiate
BenchmarkShorthash/Instantiate-16     	      43	  25012476 ns/op	     607 B/op	      46 allocs/op
```

Wazero (compiler)
```
BenchmarkShorthash
BenchmarkShorthash/Compile
BenchmarkShorthash/Compile-16         	      67	  16912063 ns/op	 7589444 B/op	  101568 allocs/op
BenchmarkShorthash/Instantiate
BenchmarkShorthash/Instantiate-16     	     564	   2061773 ns/op	 1422694 B/op	    3910 allocs/op
```


codefromthecrypt:
@jedisct1 wondering if you can export the loop function (between the calls to time) so that we can measure it precisely? then, I could use the wasm like normal. Right now, it is hard for us to benchmark against other runtimes as certain objects don't close.

For example, export "bench" which calls the stuff in a loop, and doesn't call time around it. Ack this isn't precisely what you need for your benches, but it would be handy for us especially as we can re-use the result and not accidentally build it differently.

jedisct1:
Yes, I can absolutely do that if it helps!

mathetake:
I think you can use `NewModule` API as a counterpart for our CompileModule API https://gist.github.com/kyleconroy/a50054984159918d78aa7c0448e56db5#file-main-go-L26-L30

mathetake:
also, there's such one in wasmer as well (though the doc on the function is confusing) https://github.com/wasmerio/wasmer-go/blob/master/examples/example_memory_test.go#L52-L53

codefromthecrypt:
@jedisct1 

>> Yes, I can absolutely do that if it helps!

yep it will help for sure, as then we can call the export and use go's benchmark system to run the .. well wasm benchmark ;) .. in isolation.

codefromthecrypt:
@mathetake thanks for the hint. Until we have non-wasi the only engine we can compare vs is wasmtime as that's the only one that can safely re-instantiate. I removed the others to focus.

```
# wazero compiler
BenchmarkShorthash/Instantiate-16     	     544	   2131448 ns/op	 1422624 B/op	    3909 allocs/op

# wasmtime
BenchmarkShorthash/Instantiate-16      	    1994	    573844 ns/op	     552 B/op	      41 allocs/op
```

I'll switch this once an export variant becomes available, but at least for now we can see for example perf difference amd64 arm64 for our single pass vs wasmtime latest default.

codefromthecrypt:
@mathetake thanks for the hint. Until we have non-wasi the only engine we can compare vs is wasmtime as that's the only one that can safely re-instantiate. I removed the others to focus.

```
# wazero compiler
BenchmarkShorthash/Instantiate-16     	     544	   2131448 ns/op	 1422624 B/op	    3909 allocs/op

# wasmtime
BenchmarkShorthash/Instantiate-16      	    1994	    573844 ns/op	     552 B/op	      41 allocs/op
```

I'll switch this once an export variant becomes available, but at least for now we can see for example perf difference amd64 arm64 for our single pass vs wasmtime latest default.

mathetake:
cool, though Instantiate contains the data-initialization time so not entirely execution time, but anyway good starting point!

internal/integration_test/vs/bench_shorthash.go
```diff
@@ -0,0 +1,33 @@
...
+func init() {
+	shorthashConfig = &RuntimeConfig{
+		ModuleName: "shorthash",
+		ModuleWasm: shorthashWasm,
+		NeedsWASI:  true, // runs as a _start function
+	}
+}
+
+func RunTestShorthash(t *testing.T, runtime func() Runtime) {
+	// not testCall as there are no exported functions except _start
```
mathetake, line 27: can we bench on _start?

codefromthecrypt: we can try ;) wasi initializes during start which is why we usually don't attempt to call start in a loop

mathetake (approved):

internal/integration_test/vs/bench_shorthash.go
codefromthecrypt: worth a shot but it failed on `unreachable` on the second call to `_start`

PR #949, opened 2022-12-21
Disable comp cache for host modules

mathetake:


codefromthecrypt (approved):
agree as host modules can't be wasm anymore (no more proxyfunc etc)

PR #952, opened 2022-12-21
internal/gojs: reuse single wazero.Runtime to fix flake in tests

mathetake:


mathetake:
Still, using the same path for multiple contexts and passing these contexts to multiple wazero.Runtime are not safe, but I think out of common usecase

mathetake:
now `go test ./internal/gojs/... -count=100` runs without crash

codefromthecrypt (approved):
yep this works!

PR #954, opened 2022-12-21
wasi: allows closing stdio file descriptors

### LINKED ISSUE #953: fclose <stdin>: Bad file descriptor
**Describe the bug**
I experiment with userland unix in Go and your project looks like a perfect match allowing me to run a unix utilities. After I build a few of them and tested them, I have noticed the `fclose <stdin>: Bad file descriptor` appeared. It happens for `os.Stdin`, `bytes.Buffer` or the `*io.PipeReader`. The `io.NopCloser` has no effect there. So I believe this is nothing to do with Go, but with the wasi integration of `fclose`.

**To Reproduce**
A small C reproducer

```c
#include <stdlib.h>
#include <stdio.h>

int main() {
    int ret = fclose(stdin);
    if (ret != 0) {
        perror("fclose <stdin>");
    }
    exit(ret);
}
```

To wasm - `zig cc --target=wasm32-wasi fclose.c -o fclose.wasm`

Run it via this program

```go
//go:embed testdata/fclose.wasm
var fcloseWasm []byte

func main() {
	ctx := context.Background()
	if err := run(ctx); err != nil {
		log.Fatal(err)
	}
}

func run(ctx context.Context) error {
	r := wazero.NewRuntime(ctx)
	defer r.Close(ctx)

		WithStdin(io.NopCloser(os.Stdin)).
		WithStdout(os.Stdout).
		WithStderr(os.Stderr)

	code, err := r.CompileModule(ctx, fcloseWasm)
	if err != nil {
		return err
	}

	if _, err = r.InstantiateModule(ctx, code, config.WithArgs("wasi")); err != nil {
			return err
		} else if !ok {
			return err
		}
	}
	return nil
}
```

```sh
go run main.go 
fclose <stdin>: Bad file descriptor
exit_code: 4294967295
2022/12/21 09:13:07 module "" closed with exit_code(4294967295)
exit status 1
```

**Expected behavior**
The stdin is closed, no error message appear and exit code is 0.

**Screenshots**

- Go version: go1.19.1 linux/amd64
- wazero Version: github.com/tetratelabs/wazero v1.0.0-pre.5
- Host architecture: amd64
- Runtime mode: compiler

codefromthecrypt:
This allows wasm to close stdio file descriptors such as STDOUT(1). This will not close the underlying host resource as that would break a lot of folks doing logging to the console.

Thanks a lot to @vyskocilm for making this easy to identify and fix!

```bash
$ cat > fclose.c <<EOF
#include <stdlib.h>
#include <stdio.h>

int main() {
    int ret = fclose(stdin);
    if (ret != 0) {
        perror("fclose <stdin>");
    }
    exit(ret);
}
EOF

$ zig cc --target=wasm32-wasi fclose.c -o fclose.wasm
$ go build -o wazero ./cmd/wazero/...
$ ./wazero run fclose.wasm
```

Fixes #953

### DIFF
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -3,6 +3,7 @@ package sys
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"io/fs"
 	"math"
@@ -19,7 +20,6 @@ const (
 	FdStdin uint32 = iota
 	FdStdout
 	FdStderr
-
 	// FdRoot is the file descriptor of the root ("/") filesystem.
 	//
 	// # Why file descriptor 3?
@@ -32,6 +32,11 @@ const (
 	//   - https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_14
 	//   - https://github.com/WebAssembly/wasi-libc/blob/wasi-sdk-16/libc-bottom-half/sources/preopens.c#L215
 	FdRoot
+)
+
+const (
+	modeDevice     = uint32(fs.ModeDevice | 0o640)
+	modeCharDevice = uint32(fs.ModeCharDevice | 0o640)
 )
 
 // EmptyFS is exported to special-case an empty file system.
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
ps I also double-checked that char device detection still works (ad-hoc)

mathetake (approved):

PR #956, opened 2022-12-23
Hides internal logging

codefromthecrypt:
This makes it easier to do maintenance by hiding the implementation details of using a proxy module to test host modules. What happens is we skip logging of these proxy modules, which makes it easier to see what log affects a change has.

### DIFF
--- a/experimental/logging/log_listener.go
+++ b/experimental/logging/log_listener.go
@@ -41,6 +41,10 @@ type loggingListenerFactory struct {
 // NewListener implements the same method as documented on
 // experimental.FunctionListener.
 func (f *loggingListenerFactory) NewListener(fnd api.FunctionDefinition) experimental.FunctionListener {
+	if fnd.ModuleName() == "internal/testing/proxy/proxy.go" {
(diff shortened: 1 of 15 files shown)


experimental/logging/log_listener.go
```diff
@@ -41,6 +41,10 @@ type loggingListenerFactory struct {
 // NewListener implements the same method as documented on
 // experimental.FunctionListener.
 func (f *loggingListenerFactory) NewListener(fnd api.FunctionDefinition) experimental.FunctionListener {
+	if fnd.ModuleName() == "internal/testing/proxy/proxy.go" {
```
mathetake, line 44: how about having this under testing/loggingListener and override NewListener rather than having this "test code" in the main code

codefromthecrypt: better idea indeed. lemme try!

mathetake: maybe under testing/proxy/ would be just good

mathetake (approved):
love it

codefromthecrypt:
thanks for making it better @mathetake!

PR #957, opened 2022-12-23
doc: document why it's safe against async preemption

### LINKED ISSUE #153: JIT: consider locking Goroutine thread until JIT exits
We could use `runtime.LockOSThread()` to block Goruntime from switching the goroutine via [async preemption](https://github.com/golang/proposal/blob/master/design/24543-non-cooperative-preemption.md) which can happen at any point of execution.

Even though we've tested the JIT execution is stable without locking (as in jit/engine_test.go ), there might be still some condition where the JIT would fail. 

This needs more through investigation on how Goroutine saves CPU registers when switches happens (at source code level), and more test cases to ensure whether or not we need thread blocking via `LockOSThread`.

mathetake:
closes #153 

### DIFF
--- a/internal/engine/compiler/RATIONALE.md
+++ b/internal/engine/compiler/RATIONALE.md
@@ -59,16 +59,22 @@ as a convenience layer to comply with the Go's official calling convention.
 We delegate the task to jump into the code segment to the Go assembler code.
 
 
-## Why it's safe to execute runtime-generated machine codes against Goroutine preemption
+## Why it's safe to execute runtime-generated machine codes against async Goroutine preemption
 
+Goroutine preemption is the mechanism of the Go runtime to switch goroutines contexts on an OS thread.
+There are two types of preemption: cooperative preemption and async preemption. The former happens, for example,
+when making a function call, and it is not an issue for our runtime-generated functions as they do not make
+direct function calls to Go-implemented functions. On the other hand, the latter, async preemption, can be problematic
+since it tries to interrupt the execution of Goroutine at any point of function, and manipulates CPU register states.
+
+Fortunately, our runtime-generated machine codes do not take the async preemption into account.
 All the assembly codes are entered via the trampoline implemented as Go Assembler Function (e.g. [arch_amd64.s](./arch_amd64.s)),
 and as of Go 1.20, these assembler functions are considered as _unsafe_ for async preemption:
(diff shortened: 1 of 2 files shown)


internal/engine/compiler/RATIONALE.md
```diff
@@ -58,6 +58,18 @@ And `nativecall` is actually implemented in [arch_amd64.s](./arch_amd64.s)
...
+
+## Why it's safe to execute runtime-generated machine codes against Goroutine preemption
+
+All the assembly codes are entered via the trampoline implemented as Go Assembler Function (e.g. [arch_amd64.s](./arch_amd64.s)),
+and as of Go 1.20, these assembler functions are considered as _unsafe_ for async preemption:
+- https://github.com/golang/go/blob/go1.20rc1/src/runtime/preempt.go#L406-L407
+- https://github.com/golang/go/blob/9f0234214473dfb785a5ad84a8fc62a6a395cbc3/src/runtime/traceback.go#L227
+
+From the Go runtime point of view, the execution of runtime-generated machine codes is considered as a part of
+that trampoline function. That means they are ensured to be unsafe for async preemption, so there won't be any
```
codefromthecrypt, line 70: shorten last sentence to "Therefore, runtime-generated machine code is also unsafe for async preemption."

internal/engine/compiler/RATIONALE.md
```diff
@@ -58,6 +58,18 @@ And `nativecall` is actually implemented in [arch_amd64.s](./arch_amd64.s)
 as a convenience layer to comply with the Go's official calling convention.
 We delegate the task to jump into the code segment to the Go assembler code.
 
+
+## Why it's safe to execute runtime-generated machine codes against Goroutine preemption
+
```
codefromthecrypt, line 63: there is a little jargon, so possibly an overview sentence or two can help.

Maybe start with the problem of goroutine premption, and then say "unsafe" means goroutines are not allowed to pre-empt.

codefromthecrypt (commented):
Thanks a lot!

internal/engine/compiler/RATIONALE.md
mathetake: `ensured to be unsafe` is the important part and means `correctly marked as unsafe`, so will modify it.

codefromthecrypt (approved):
nicely done!

PR #938, opened 2022-12-16
logging: extracts utilities to read parameters and results from memory

codefromthecrypt:
This changes the listener signature to accept context and calling
module, so that all possible parameters and results can be logged. This
also changes the logging listener to make parameters visible when
logging results.

This infrastructure supports some helpful use cases, such as logging
WASI result parameters, such as the prestat path, which is only knowable
after the function has been called. The context parameter supposed
reading results of gojs functions, which are stored host-side in a go
context object.

Future pull requests will complete this as well backfill unit tests.
This is raised independently mainly to keep the PR size down of the
upcoming filesystem logger.

PR #959, opened 2022-12-24
gojs: refactors GOOS and GOARCH specific code into their own packages

codefromthecrypt:
This refactors GOOS and GOARCH specific code into their own packages. This allows logging interceptors to be built without cyclic package dependencies.

This prepares for #938 which would allow us to see filesystem call arguments in logs, even with gojs.


internal/gojs/wasm/wasm.go
```diff
@@ -0,0 +1,526 @@
...
+	return &wasm.HostFunc{
+		ExportNames: []string{name},
+		Name:        name,
+		ParamTypes:  []wasm.ValueType{wasm.ValueTypeI32},
+		ParamNames:  []string{"sp"},
+		Code:        &wasm.Code{IsHostFunction: true, Body: []byte{wasm.OpcodeEnd}},
+	}
+}
+
+func NewFuncR_V(name string, goFunc FuncR_V, pName string) *wasm.HostFunc {
```
codefromthecrypt, line 72: from here down handle all the go signatures used by `GOARCH=wasm GOOS=js` derived from the SP parameter. There won't be more (I think) as inside the javascript "operating system" "valueCall" uses dynamic dispatch, which doesn't affect the stack except arg length.

internal/gojs/wasm/wasm.go
```diff
@@ -0,0 +1,526 @@
...
+	p2Len := le.Uint32(params[16:])
+
+	p2 := refSlice(mem, f.p2Name, p2Array, p2Len)
+
+	r1, r2 := f.f(ctx, mod, p1, p2)
+
+	refreshSPAndStoreResults(mem, mod, paramCount, r1, r2)
+}
+
+func refreshSPAndStoreResults(mem api.Memory, mod api.Module, paramCount uint32, r1 js.Ref, r2 bool) {
```
codefromthecrypt, line 470: the two functions that use this are gnarly "new" and "call", but now I like that even though the stack game including refreshing the pointer is gnarly, it isn't also mixed in with the business logic, like it was before.

codefromthecrypt:
found a way to reduce the amount of code.. WIP StackFunc

codefromthecrypt:
ok hobby PR from the weekend is ready for review ;)

internal/gojs/runtime.go
```diff
@@ -19,16 +20,21 @@ const (
...
+var Debug = goarch.StubFunction("debug")
+
 // WasmExit implements runtime.wasmExit which supports runtime.exit.
 //
 // See https://github.com/golang/go/blob/go1.19/src/runtime/sys_wasm.go#L28
-var WasmExit = newSPFunc(wasmExitName, wasmExit)
-
-func wasmExit(ctx context.Context, mod api.Module, sp []uint64) {
-	mem := mod.Memory()
+var WasmExit = goarch.NewFunc(wasmExitName, wasmExit,
```
codefromthecrypt, line 31: this file defines the functions of GOARCH=wasm and now has no dependency on anything related to js

internal/gojs/syscall.go
```diff
@@ -163,175 +138,150 @@ func valueSet(ctx context.Context, mod api.Module, sp []uint64) {
...
-var ValueSetIndex = stubFunction(valueSetIndexName)
+var ValueSetIndex = goarch.StubFunction(valueSetIndexName)
 
 // ValueCall implements js.valueCall, which is used to call a js.Value function
 // by name, e.g. `document.Call("createElement", "div")`.
 //
 // See https://github.com/golang/go/blob/go1.19/src/syscall/js/js.go#L394
 // and https://github.com/golang/go/blob/go1.19/misc/wasm/wasm_exec.js#L343-L358
-var ValueCall = newSPFunc(valueCallName, valueCall)
+var ValueCall = goos.NewFunc(valueCallName, valueCall,
```
codefromthecrypt, line 175: this function actually implements most syscalls as file i/o is mapped at the JS layer in GOARCH=wasm/GOOS=js.

In order to do logging, we need to see the parameters. We now, have natural visibility except the expected names of the args. Ex. the args to "mkdir". Adding those will be in a follow-up PR as it is too much to do right now. This PR doesn't change any behavior, just re-organizes the code so that the abstractions are separated and also the packages that handle them are decoupled.

internal/gojs/syscall.go
```diff
@@ -163,175 +138,150 @@ func valueSet(ctx context.Context, mod api.Module, sp []uint64) {
...
-		if n, ok := args[0].(float64); ok {
+		var result interface{}
+		a := args[0]
+		if n, ok := a.(float64); ok {
 			result = &byteArray{make([]byte, uint32(n))}
-		} else if n, ok := args[0].(uint32); ok {
-			result = &byteArray{make([]byte, n)}
-		} else if b, ok := args[0].(*byteArray); ok {
-			// In case of below, in HTTP, return the same ref
+		} else if _, ok := a.(*byteArray); ok {
```
codefromthecrypt, line 240: fyi I tried to get args to be able to unwrap byteArray, but failed a little, so Ieft it. In any case that can be done in a follow-up PR. Meanwhile I simplified the logic here a little bit and made the reference increment more explicit than before.

codefromthecrypt (commented):
notes

mathetake (approved):

PR #964, opened 2022-12-27
do not merge: demonstrate Go routine stack allocation and stack pointer manipulation during JIT code execution

mathetake:


mathetake:
ok verified that this works for amd64 and arm64 both!

PR #967, opened 2022-12-27
Fix fdPread: shouldn't move file offset

### LINKED ISSUE #958: WASI fd_pread incorrectly moves file offset
**Describe the bug**

The [`fd_pread`](https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-fd_preadfd-fd-iovs-iovec_array-offset-filesize---errno-size) WASI syscall incorrectly moves the file offset.

Implementation probably can't be shared with `fd_read`, and should use `io.ReaderAt` not `io.ReadSeeker`.
https://github.com/tetratelabs/wazero/blob/6b9d119a7b069561ccb7f536a6d0d97d3772f549/imports/wasi_snapshot_preview1/fs.go#L561-L624

**To Reproduce**

```sh
git clone https://github.com/ncruces/wazero-bugs.git
cd wazero-bugs

echo Expected output:
cc -o main main.c && ./main

echo Actual output:
go run ./cmd
```

**Expected behavior**

```
Expected output:
Line 1
Line 2
Line 1
Line 3
Actual output:
Line 1
Line 2
Line 1
Line 2
```

Note the difference in the last line.

- Go version: 1.19.4
- wazero Version: v1.0.0-pre.5
- Host architecture: amd64
- Runtime mode: compiler

**Additional context**

[`main.c`](https://github.com/ncruces/wazero-bugs/blob/main/main.c) and [`main.go`](https://github.com/ncruces/wazero-bugs/blob/main/cmd/main.go) should be pretty self explanatory.
`main.wasm` from the repo was compiled with `zig cc --target=wasm32-wasi` version `0.10.0`.

I tried running it with `wasmtime --dir . main.wasm` which produces the correct output.

ncruces:
Fixes #958.

This is breaking a zig test, I still haven't been able to figure out why.

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -592,6 +592,19 @@ operation will err if it needs to. This helps reduce the complexity of the code
 in wazero and also accommodates the scenario where the bytes read are enough to
 satisfy its processor.
 
+### fd_pread: io.Seeker fallback when io.ReaderAt is not supported
+
+`ReadAt` is the Go equivalent to `pread`: it does not affect, and is not
+affected by, the underlying file offset. Unfortunately, `io.ReaderAt` is not
+implemented by all `fs.File`, notably `embed.file`.
+
+The initial implementation of `fd_pread` instead used `Seek`. To avoid a
+regression we fallback to `io.Seeker` when `io.ReaderAt` is not supported.
+
+This requires obtaining the initial file offset, seeking to the intended read
+offset, and reseting the file offset the initial state. If this final seek
+fails, the file offset is left in an undefined state. This is not thread-safe.
+
 ### Pre-opened files
 
 WASI includes `fd_prestat_get` and `fd_prestat_dir_name` functions used to
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -523,8 +523,20 @@ func fdReadOrPread(mod api.Module, params []uint64, isPread bool) Errno {
 		return ErrnoBadf
 	}
 
+	read := r.Read
 	if isPread {
-		if s, ok := r.(io.Seeker); ok {
+		if ra, ok := r.(io.ReaderAt); ok {
+			read = func(p []byte) (int, error) {
+				n, err := ra.ReadAt(p, offset)
+				offset += int64(n)
+				return n, err
+			}
+		} else if s, ok := r.(io.Seeker); ok {
+			currentOffset, err := s.Seek(0, io.SeekCurrent)
+			if err != nil {
+				return ErrnoFault
+			}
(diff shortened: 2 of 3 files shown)


ncruces:
It seems, at the root of it, we have that [`embed.openfile`](https://github.com/golang/go/blob/15605ca827723d568f780402d03b29842fd20eec/src/embed/embed.go#L342-L350) does not implement `io.ReaderAt`. 😔

ncruces:
I could `Seek`, `Read` and then `Seek` back. But if any operation fails, I'm breaking the invariant of `pread`: _“The file offset is not changed.”_

codefromthecrypt:
@ncruces is it required for embed to implement this, or is it just due to our tests using embed?

ncruces:
No, not required. It's just due to tests using embed.

But it also means anyone using embed will start seeing `pread` failures. And it seems zig uses `pread` for most (all?) reads.

I came up with an alternative implementation that tries `io.ReaderAt` and falls back to `io.Seeker` doing multiple seeks. All tests pass. https://github.com/ncruces/wazero/commit/e12b34e2e3e8ab9e3d09d5c59afd63c28198712f

There's some prior to doing this. For instance, see Go's implementation of `pread` under Windows:
https://github.com/golang/go/blob/e870de9936a7efa42ac1915ff4ffb16017dbc819/src/internal/poll/fd_windows.go#L538-L546

codefromthecrypt:
thanks for the notes and the research leading to them!

codefromthecrypt:
If you want to go with an alternative fix, works for me, @achille-roussel will be starting some more abstraction work soon, and the important part is to have tests in to make sure this doesn't re-break later.

ncruces:
I'll go with this since at least it means no regression for zig folks.

The big issue is this leaving the file offset in an inconsistent state if the initial seeks succeed but the final one fails. It's still an improvement over the current situation (final offset is always wrong).

Other considerations don't really apply (`pread` is supposed to be thread-safe, but there's no concurrency in WASM). I can add some comments pointing this out.

I'll work on tests now.

codefromthecrypt:
@ncruces sounds good. we're going to cut the release in a couple days, and this is great to go in. You can add notes in RATIONALE.md and then code comment to there on the tradeoffs. you'll notice there are others.

codefromthecrypt:
I drifted you. for names, use export (capitalize first letter) and dot import like so `	. "github.com/tetratelabs/wazero/internal/wasi_snapshot_preview1"`

codefromthecrypt (approved):
great and professional work!

if you like I can fix the drift I caused.

codefromthecrypt:
thanks for hanging in there @ncruces!

ncruces:
No problem.
Thank you for wazero!

PR #971, opened 2022-12-29
wasi: unexports errno and re-uses constants for logging

codefromthecrypt:
We originally exported WASI errno as we originally supported invoking host functions directly (e.g. without a guest). This was an invalid call pattern, and we removed that. However, we left the errnos exported even though the caller of a guest won't ever see them. This prevented us from re-using them cleanly in features such as logging.

This moves all constants including function names and flag enums internal so that there is less duplication between logging and implementation of wasi functions. This also helps in reference searches, as we can analyze uses of a particular function name.

The only constant left exported is the module name, as there's a use case for that (overriding implementations via FunctionBuilder).


imports/wasi_snapshot_preview1/wasi_test.go
```diff
@@ -50,8 +52,8 @@ func TestNewFunctionExporter(t *testing.T) {
 		defer r.Close(testCtx)
 
 		// Export the default WASI functions
-		wasiBuilder := r.NewHostModuleBuilder(ModuleName)
-		NewFunctionExporter().ExportFunctions(wasiBuilder)
+		wasiBuilder := r.NewHostModuleBuilder(wasi_snapshot_preview1.ModuleName)
```
codefromthecrypt, line 55: this is the only constant left exported, as it is needed when you use FunctionExporter.

PR #972, opened 2022-12-29
Exposes experimental writefs.DirFS, but hides implementation

codefromthecrypt:
The type we use to expose write operations is still evolving. It might be a single writefs.FS interface, or similar to go where we have an interface per feature (e.g. writefs.MkdirFS). These choices are all implementation details for DirFS and won't be settled before the end of the month version cutoff. Instead, this only exposes the ability to create a DirFS, not an arbitrary implementation of writefs.FS. This does so by making `writefs.FS` an internal type.


codefromthecrypt:
@achille-roussel this will drift you, but it will allow end users to start with write features while we decide on a pluggability API. Those giving just a filepath don't need to wait for api shape to settle. We can finalize these things in january, in other words, but allow folks to begin using write support meanwhile.

PR #973, opened 2022-12-29
fs: splits unlink and rmdir from remove

codefromthecrypt:
This splits unlink and rmdir from remove, as it is not only more precise in GOOS=js, but it is also needed to implement wasi. I verified this works by running go unit tests with logging.

```
==> go.syscall/js.valueCall(fs.open(name=/tmp/TestErrIsNotExist1062486353,flags=,perm=----------))
<== (err=<nil>,fd=10)
==> go.syscall/js.valueCall(fs.fstat(fd=10))
<== (err=<nil>,stat={isDir=true,mode=-rwx------,size=96,mtimeMs=1672285985206})
==> go.syscall/js.valueCall(fs.readdir(name=/tmp/TestErrIsNotExist1062486353))
<== (err=<nil>,dirents=&{[001]})
==> go.syscall/js.valueCall(fs.unlink(path=/tmp/TestErrIsNotExist1062486353/001))
<== (err=is a directory,ok=true)
==> go.syscall/js.valueCall(fs.rmdir(path=/tmp/TestErrIsNotExist1062486353/001))
<== (err=<nil>,ok=false)
```

### DIFF
--- a/internal/gojs/fs.go
+++ b/internal/gojs/fs.go
@@ -34,7 +34,7 @@ var (
 		addFunction(custom.NameFsWrite, &jsfsWrite{}).
 		addFunction(custom.NameFsReaddir, &jsfsReaddir{}).
 		addFunction(custom.NameFsMkdir, &jsfsMkdir{}).
-		addFunction(custom.NameFsRmdir, &jsfsUnlink{}).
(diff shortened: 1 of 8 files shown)


mathetake (approved):

codefromthecrypt:
heh windows again, but luckily I have my laptop ready

PR #975, opened 2022-12-29
add internal/sys.FileTable

achille-roussel:
As discussed on Slack, this PR brings the file table from https://github.com/achille-roussel/wazero/tree/experimental/sys into upstream Wazero.

I added the data structure as `internal/sys.FileTable`, and I made it a generic type despite the single use in `FSContext` in anticipation that:
- I'm going to bring these changes in https://github.com/achille-roussel/wazero/tree/experimental/sys
- we may have to change the key or value type as we work towards full support of writefs
- we may have multiple instances of it for wasi and gojs (and maybe more with wasi2)

If these assumption turn out not to be true we can revisit and 1) unexport the type and 2) drop the generic type parameters.

Please take a look and let me know if anything needs to be changed!

### DIFF
--- a/internal/sys/file_table.go
+++ b/internal/sys/file_table.go
@@ -1,0 +1,132 @@
+package sys
+
+import "math/bits"
+
+// FileTable is a data structure mapping 32 bit file descriptor to file objects.
+//
+// The data structure optimizes for memory density and lookup performance,
+// trading off compute at insertion time. This is a useful compromise for the
+// use cases we employ it with: files are usually read or written a lot more
+// often than they are opened, each operation requires a table lookup so we are
+// better off spending extra compute to insert files in the table in order to
+// get cheaper lookups. Memory efficiency is also crucial to support scaling
+// with programs that open thousands of files: having a high or non-linear
+// memory-to-item ratio could otherwise be used as an attack vector by malicous
+// applications attempting to damage performance of the host.
+type FileTable struct {
+	masks []uint64
+	files []*FileEntry
+}
+
+// Len returns the number of files stored in the table.
+func (t *FileTable) Len() (n int) {
+	// We could make this a O(1) operation if we cached the number of files in
+	// the table. More state usually means more problems, so until we have a
+	// clear need for this, the simple implementation may be a better trade off.
+	for _, mask := range t.masks {
+		n += bits.OnesCount64(mask)
+	}
+	return n
+}
+
+// Grow ensures that t has enough room for n files, potentially reallocating the
+// internal buffers if their capacity was too small to hold this many files.
+func (t *FileTable) Grow(n int) {
+	// Round up to a multiple of 64 since this is the smallest increment due to
+	// using 64 bits masks.
+	n = (n*64 + 63) / 64
+
+	if n > len(t.masks) {
+		masks := make([]uint64, n)
+		copy(masks, t.masks)
+
+		files := make([]*FileEntry, n*64)
+		copy(files, t.files)
+
+		t.masks = masks
+		t.files = files
+	}
+}
+
+// Insert inserts the given file to the table, returning the fd that it is
+// mapped to.
+//
+// The method does not perform deduplication, it is possible for the same file
+// to be inserted multiple times, each insertion will return a different fd.
+func (t *FileTable) Insert(file *FileEntry) (fd uint32) {
+	offset := 0
+insert:
+	// Note: this loop could be made a lot more efficient using vectorized
+	// operations: 256 bits vector registers would yield a theoretical 4x
+	// speed up (e.g. using AVX2).
+	for index, mask := range t.masks[offset:] {
+		if ^mask != 0 { // not full?
+			shift := bits.TrailingZeros64(^mask)
+			index += offset
+			fd = uint32(index)*64 + uint32(shift)
+			t.files[fd] = file
+			t.masks[index] = mask | uint64(1<<shift)
+			return fd
+		}
+	}
+
+	offset = len(t.masks)
+	n := 2 * len(t.masks)
+	if n == 0 {
+		n = 1
+	}
+
+	t.Grow(n)
+	goto insert
+}
+
+// Lookup returns the file associated with the given fd (may be nil).
+func (t *FileTable) Lookup(fd uint32) (file *FileEntry, found bool) {
+	if i := int(fd); i >= 0 && i < len(t.files) {
+		index := uint(fd) / 64
+		shift := uint(fd) % 64
+		if (t.masks[index] & (1 << shift)) != 0 {
+			file, found = t.files[i], true
+		}
+	}
+	return
+}
+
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/tetratelabs/wazero
 
 // Floor Go version of wazero (current - 2)
-go 1.17
+go 1.18
```
codefromthecrypt, line 4: this will hold us back until feb/when go 1.20 is released. are you anticipating keeping this PR open until then, or do you want to work around it for now?

achille-roussel: Ah OK, I may have misunderstood your directions on Slack.

Without this change Go refuses to build the package since generics weren't supported in 1.17. I guess I'll change the type to use concrete implementation and we'll keep copies for now then 👍

internal/sys/fs.go
```diff
@@ -420,11 +392,13 @@ func (c *FSContext) CloseFile(fd uint32) bool {
...
-	for fd, entry := range c.openedFiles {
-		delete(c.openedFiles, fd)
+	c.openedFiles.Range(func(fd uint32, entry *FileEntry) bool {
 		if e := entry.File.Close(); e != nil {
 			err = e // This means err returned == the last non-nil error.
 		}
-	}
+		return true
+	})
+	// TODO: Required to pass TestEmptyFSContext but Reset would be better.
```
codefromthecrypt, line 401: a closed context can't be re-opened, why would we have a TODO to use reset?

internal/sys/fs_test.go
```diff
@@ -62,12 +62,12 @@ func TestNewFSContext(t *testing.T) {
 			require.NoError(t, err)
```
codefromthecrypt, line 62: probably in this file we need to add a test that FDs are re-used. This way a small change to the file table or internals don't invalidate the overall goal which is fd re-use.

ex. 4 == open(f1) 5 == open(f2) 6 == open(f1)  close(4) 4 == open(f2)

internal/sys/file_table.go
```diff
@@ -0,0 +1,141 @@
...
+					}
+				}
+			}
+		}
+	}
+}
+
+// Reset clears the content of the table.
+func (t *FileTable) Reset() {
+	for i := range t.masks {
```
codefromthecrypt, line 135: do you want to clear the references or truncate the slices instead? e.g. t.masks = t.masks[:]

internal/sys/file_table.go
```diff
@@ -0,0 +1,141 @@
...
+			t.masks[index] = mask & ^uint64(1<<shift)
+		}
+	}
+}
+
+// Range calls f for each file and its associated fd in the table. The function
+// f might return false to interupt the iteration.
+func (t *FileTable) Range(f func(uint32, *FileEntry) bool) {
+	for i, mask := range t.masks {
+		if mask != 0 {
```
codefromthecrypt, line 121: nit: prefer early continues vs deep nesting

internal/sys/file_table.go
```diff
@@ -0,0 +1,141 @@
...
+	for _, mask := range t.masks {
+		n += bits.OnesCount64(mask)
+	}
+	return n
+}
+
+// Grow ensures that t has enough room for n files, potentially reallocating the
+// internal buffers if their capacity was too small to hold this many files.
+func (t *FileTable) Grow(n int) {
+	if n = (n*8 + 7) / 8; n > len(t.masks) {
```
codefromthecrypt, line 50: maybe comment this arithmetic?

codefromthecrypt, line 50: since this file looks benchmark optimized you can consider shifting vs multiplying by powers of 2

e.g. `n << 3 // n * 8`

internal/sys/file_table.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Insert inserts the given file to the table, returning the fd that it is
+// mapped to.
+//
+// The method does not perform deduplication, it is possible for the same file
+// to be inserted multiple times, each insertion will return a different fd.
+func (t *FileTable) Insert(file *FileEntry) (fd uint32) {
+	offset := 0
+insert:
+	// TODO: this loop could be made a lot more efficient using vectorized
```
codefromthecrypt, line 70: does go support vectorized operations now or in the near future? If so, maybe not a TODO

internal/sys/file_table.go
```diff
@@ -0,0 +1,141 @@
+package sys
+
+import "math/bits"
+
+// FileTable is a data structure mapping 32 bit file descriptor to file objects.
```
codefromthecrypt, line 5: move the heft of this to /RATIONALE.md then link back to it. e.g. See /RATIONALE.md. this way the godoc can focus on what it is doing overall vs how and why it was made the way it was.

codefromthecrypt (approved):
Looks nice and professional. I have some feedback which are polish in nature!

internal/sys/fs.go
achille-roussel: I assumed it could, but if it cannot there's no need for this comment indeed 👍

internal/sys/file_table.go
achille-roussel: I think the convention on `Reset` methods is usually that they don't release the internal state. The data structure is in an "empty" state after clearing the bit masks so I think this is the intended behavior.

internal/sys/file_table.go
achille-roussel: The Go compiler will do arithmetic operations already (e.g. n*8 is translated to n<<3). The `Grow` method is not intended to be called frequently since the capacity grows exponentially, its cost is amortized over repeated calls to `Insert`.

Unless we have a performance profile showing the optimization would be worth making I'd rather keep the arithmetic in a form that is more readable to human maintainers if that's OK with you.

internal/sys/file_table.go
achille-roussel: I'll change this to just a note rather than a TODO.

achille-roussel:
Thanks for the feedback @codefromthecrypt, I believe that I addressed or followed up on all the comments you left.

I also realized that the email address that I had signed the commits with was outdated, so I rewrote the commit history to change signatures to the correct address.

Please let me know if you have any other feedback!

codefromthecrypt:
your feedback on feedback was all reasonable. thanks a lot @achille-roussel!

internal/sys/file_table.go
codefromthecrypt: this was more a nit as I noticed you had some benchmarks. there are other code where we found this improved the results, but agree that "grow" is indeed not one of those frequently called things and thanks for pointing it out!

internal/sys/file_table.go
codefromthecrypt: this was mainly an idea to get less code. ack it is different to retain the underlying slice vs retain the count of masks (ex clear it as you are). sgtm

PR #979, opened 2022-12-30
renames writefs to syscallfs and implements utimes in gojs

codefromthecrypt:
This renames the internal writefs package to syscallfs as it is largely dependent on syscall signatures. This also implements utimes in gojs. WASI will be a follow-up change as it requires more infrastructure. Notably, we also need non-TinyGo tests because TinyGo doesn't yet support os.Chtimes or corresponding syscalls.

### DIFF
--- a/internal/gojs/testdata/fs/main.go
+++ b/internal/gojs/testdata/fs/main.go
@@ -53,5 +53,5 @@ func testAdHoc() {
 	if err != nil {
 		log.Panicln(err)
 	}
-	fmt.Println("empty:", string(b))
+	fmt.Println("empty:" + string(b))
 }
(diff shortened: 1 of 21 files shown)


internal/gojs/testdata/fs/main.go
```diff
@@ -53,5 +53,5 @@ func testAdHoc() {
 	if err != nil {
 		log.Panicln(err)
 	}
-	fmt.Println("empty:", string(b))
+	fmt.Println("empty:" + string(b))
```
codefromthecrypt, line 56: this was to stop the IDE from correcting the trailing space all the time.

PR #980, opened 2022-12-30
gojs: fix memory out of bounds

codefromthecrypt:
When in a loop reading a lot of data, the interpreter passes, but the compiler fails after this change. I'd love a hand from @mathetake to figure out the latter!

### DIFF
--- a/internal/gojs/misc_test.go
+++ b/internal/gojs/misc_test.go
@@ -3,6 +3,7 @@ package gojs_test
 import (
 	"bytes"
 	"fmt"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -37,6 +38,7 @@ func Test_stdio(t *testing.T) {
 
 	input := "stdin\n"
 	stdout, stderr, err := compileAndRun(testCtx, "stdio", wazero.NewModuleConfig().
+		WithEnv("BUF_LEN", strconv.Itoa(len(input))).
 		WithStdin(strings.NewReader(input)))
 
 	require.Equal(t, "stderr 6\n", stderr)
@@ -50,6 +52,7 @@ func Test_stdio_large(t *testing.T) {
 	size := 2 * 1024 * 1024 // 2MB
 	input := make([]byte, size)
 	stdout, stderr, err := compileAndRun(testCtx, "stdio", wazero.NewModuleConfig().
+		WithEnv("BUF_LEN", strconv.Itoa(size)).
 		WithStdin(bytes.NewReader(input)))
 
 	require.EqualError(t, err, `module "" closed with exit_code(0)`)
--- a/internal/gojs/misc_test.go
+++ b/internal/gojs/misc_test.go
@@ -45,7 +45,6 @@ func Test_stdio(t *testing.T) {
 }
 
 func Test_stdio_large(t *testing.T) {
-	t.Skip("TODO: #980 memory out of bounds when run with compiler")
 	t.Parallel()
 
 	size := 2 * 1024 * 1024 // 2MB
--- a/internal/gojs/testdata/stdio/main.go
+++ b/internal/gojs/testdata/stdio/main.go
@@ -6,14 +6,16 @@ import (
 	"os"
 )
 
+var buf = make([]byte, 2*1024*1024)
+
 func Main() {
-	b, err := io.ReadAll(os.Stdin)
+	n, err := io.ReadFull(os.Stdin, buf)
 	if err != nil {
 		panic(err)
 	}
 
-	printToFile("stdout", os.Stdout, len(b))
-	printToFile("stderr", os.Stderr, len(b))
+	printToFile("stdout", os.Stdout, n)
+	printToFile("stderr", os.Stderr, n)
 }
 
 func printToFile(name string, file *os.File, size int) {
--- a/internal/gojs/testdata/stdio/main.go
+++ b/internal/gojs/testdata/stdio/main.go
@@ -4,11 +4,16 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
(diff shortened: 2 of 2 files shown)


mathetake:
Isn't the failure just timeout? How about increasing it in Makefile

mathetake:
I see only Go 1.17 failing and that somehow indicates it

codefromthecrypt:
actually the failure is severe on compiler. but yeah some glitches are on interpreter

internal/gojs/compiler_test.go
```diff
@@ -89,7 +89,7 @@ func TestMain(m *testing.M) {
 
 	// Seed wazero's compilation cache to see any error up-front and to prevent
 	// one test from a cache-miss performance penalty.
-	rt = wazero.NewRuntimeWithConfig(testCtx, wazero.NewRuntimeConfig())
+	rt = wazero.NewRuntimeWithConfig(testCtx, wazero.NewRuntimeConfigInterpreter())
```
codefromthecrypt, line 92: here's where I temporarily switched to interpreter

mathetake:
verified that the failure is not arch-specific therefore, it seems some discrepancy in the host function implementation

internal/gojs/compiler_test.go
mathetake: ic got it!

mathetake:
as @robbertvanginkel pointed out, I suspect this https://github.com/tetratelabs/wazero/blob/gojs_scheduledTimeouts/internal/gojs/runtime.go#L56-L57 is the root cause

mathetake:
checked the logging filter logs, and the executions are exactly the same between the two engines until face the out of bounds error.

mathetake:
Failed after resumed!

```
<== 
	==> go.syscall/js.valueCall(fs.read(fd=0,offset=0,byteCount=286720,fOffset=<nil>))
		--> .wasm_export_resume()
		<--
	<== (err=<nil>,n=286720)
    require.go:312: expected error "module "" closed with exit_code(0)", but was "wasm error: out of bounds memory access
```

codefromthecrypt:
I will pull out the change that implements the timeout event from this PR, to keep things simpler. That go is asking for a ref that is beyond memory, and that it has already grown memory.. makes me wonder if interpreter passing is more about the nature of how it runs (slower more go-side GC etc) vs anything about the impl. ps I also found that you can allocate a buf that size and not crash in either case. It may be possible to narrow this down to something not host function related, in other words. If we could, that would be great as it could end up being something we could get help from go people on. (e.g. if run in node.js it also crashes)

codefromthecrypt:
reduced the size of this change to show the failure. the commit this is based on adds more comments around memory tracking and such, as well makes it easier to run a test with the interpreter to compare.

codefromthecrypt:
added a commit to show that this isn't i/o related. rather something about how GC is going on, considering the buf used in io.ReadAll starts at 512. Will stop this for now as there are other things on the release I need to catch up.

```diff
+ var buf = make([]byte, 2*1024*1024)

func Main() {
-	b, err := io.ReadAll(os.Stdin)
+	n, err := io.ReadFull(os.Stdin, buf)
```

mathetake:
does that timeout executing "resume" thingy spawn another execution of the gojs Wasm? Then if one execution grows memory, another execution doesn't know if it's grown or not and therefore reach the memory out of bounds

mathetake:
that is safe for the interpreter as it always references *wasm.MemoryInstance rather than []byte in the compiler.

mathetake:
oh wait, how can I reproduce now?

mathetake:
> does that timeout executing "resume" thingy spawn another execution of the gojs Wasm? Then if one execution grows memory, another execution doesn't know if it's grown or not and therefore reach the memory out of bounds

yeah this comes back to implementing resetMemoryDataView 
 https://github.com/tetratelabs/wazero/blob/main/internal/gojs/runtime.go#L59-L61 which is necessary for compilers

codefromthecrypt:
you can reproduce by going back a commit before the workaround.

ps you seem to keep saying things like "yeah this comes back to implementing resetMemoryDataView" but how would that happen in a way that would do anything? for example, what state would the host change inside that implementation? For example, this function does not return any result back to wasm, so I don't know how a compiler will depend on it...

codefromthecrypt:
sorry I guess you meant by "compiler" our compiler engine. otoh wouldn't our compiler be aware that memory.grow was called? e.g. in emscripten the callback is invoked directly after memory.grow..

mathetake:
> sorry I guess you meant by "compiler" our compiler engine. otoh wouldn't our compiler be aware that memory.grow was called?

yeah I meant our compiler engine, and the engine is aware of memory.grow being called *if that memory.grow is called on the same callEngine*.

mathetake:
if another callEngine is executing memory.grow, then others are not aware of it and memory view becomes stale. That is what I am suspicious about.

mathetake:
so basically concurrently executing the same module instance which might come with memory.grow is dangerous without implementing resetMemoryDataView thingy for compiler

mathetake:
in anyway dangerous though 😉

codefromthecrypt:
ok I understand now. I had responsibilities partially understood, basically we may need internal sharing of grow even if host state isn't dependent on guest memory slices. So, my code comment only covers the front end, not concerns our compiler may need to deal with.

mathetake:
but on the other hand I inserted `runtime.Breakpoint()` in the timeout event, but the failing case doesn't hit the point... not sure if the concurrent thing is the real cause now 😞

mathetake:
ok maybe getting close to the fix..

mathetake:
superseded by #988

Issue #977, opened 2022-12-29
Deadlock in compilation cache


robbertvanginkel:
**Describe the bug**
Use of the compilation cache between wazero versions seems to be able to cause a deadlock when the cache is cleaned.

**To Reproduce**
```
$ go mod init example.com/wazero-dl
go: creating new go.mod: module example.com/wazero-dl
$ cat <<EOF > main.go
package main

import (
        "context"
        "log"
        "os"
        "time"

        "github.com/tetratelabs/wazero"
        "github.com/tetratelabs/wazero/experimental"
)

func main() {
        bin, err := os.ReadFile(os.Args[1])
        if err != nil {
                log.Panicln(err)
        }

        ctx, err := experimental.WithCompilationCacheDirName(context.Background(), ".build")
        if err != nil {
                log.Panicln(err)
        }
        r := wazero.NewRuntime(ctx)
        defer r.Close(ctx) // This closes everything this Runtime created.

        start := time.Now()
        if _, err = r.CompileModule(ctx, bin); err != nil {
                log.Panicln(err)
        }
        log.Printf("CompileModule took %dms", time.Since(start).Milliseconds())
}
EOF
$ go get github.com/tetratelabs/wazero@3f578ddac3897b1435fbb1fe7e807676f062e4b1
go: added github.com/tetratelabs/wazero v1.0.0-pre.5.0.20221229073945-3f578ddac389
$ echo '(module (func (export "main") (result i32) i32.const 42 return))' | wat2wasm - -o simple.wasm
$ go run . simple.wasm
2022/12/29 15:16:56 CompileModule took 0ms
$ go get github.com/tetratelabs/wazero@62be68dfc5f477d766d9dcfe422dcfc498c7bbfd
go: downloading github.com/tetratelabs/wazero v1.0.0-pre.5.0.20221229092947-62be68dfc5f4
go: upgraded github.com/tetratelabs/wazero v1.0.0-pre.5.0.20221229073945-3f578ddac389 => v1.0.0-pre.5.0.20221229092947-62be68dfc5f4
$ go run . simple.wasm
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [semacquire]:
sync.runtime_SemacquireMutex(0x104425620?, 0xa0?, 0x1400013db68?)
        /Users/robbert/.gimme/versions/go1.19.3.darwin.arm64/src/runtime/sema.go:77 +0x24
sync.(*RWMutex).Lock(0x101400013dba8?)
[...]

PR #982, opened 2022-12-30
compilationcache: uses version-specific subdirectory

codefromthecrypt:
This makes the version, arch, os tuple into a subdirectory to help troubleshooting and cache management in general. The version is left inside the binary key regardless.

Note: when installing via `go install ./cmd/wazero/...` the version ends up as `dev`. This helps make that obvious. For example.

```bash
$ wazero version
dev
$ ./build/tinygo test -target wasi -c -o os.wasm os
$ wazero run -cachedir=$HOME/.wazero -mount=.:/ -env=HOME=.  os.wasm -test.v
$ find $HOME/.wazero
/Users/adrian/.wazero
/Users/adrian/.wazero/wazero-dev-amd64-darwin
/Users/adrian/.wazero/wazero-dev-amd64-darwin/1f149f4bf475a33023ce33302780bee29ec08e89bd57cfbdf639c65c6009f1a4
```

See #977

### DIFF
--- a/experimental/compilation_cache.go
+++ b/experimental/compilation_cache.go
@@ -5,9 +5,12 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
+	"runtime"
 
 	"github.com/tetratelabs/wazero/internal/compilationcache"
+	"github.com/tetratelabs/wazero/internal/version"
 )
 
 // WithCompilationCacheDirName configures the destination directory of the compilation cache.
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
@mathetake in case you get bored in the next 18hrs ;)

codefromthecrypt:
I'll split the crash on tinygo tests into a different PR. for starters I think moving the cachedir makes a lot of sense, as it is not possible to debug version number skew inside a hashed string.

codefromthecrypt:
ps normal wasm seems to still work on cachedir, that's why I'm splitting it.

codefromthecrypt:
https://github.com/tetratelabs/wazero/issues/983 forked the tinygo with cache thing as while it is an issue it shouldn't block release as other wasm works with the cache

PR #987, opened 2022-12-31
Removes internal dependency on fs.FS

codefromthecrypt:
As noted in slack, we are unlikley to long term use fs.FS internally. This ensures we attempt to cast to syscallfs.FS for all I/O by panicing on fs.Open.

### DIFF
--- a/internal/syscallfs/syscallfs.go
+++ b/internal/syscallfs/syscallfs.go
@@ -2,9 +2,6 @@ package syscallfs
 
 import (
 	"io/fs"
-	"os"
-	"path"
-	"syscall"
 )
 
 // FS is a writeable fs.FS bridge backed by syscall functions needed for ABI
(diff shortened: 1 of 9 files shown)


internal/syscallfs/syscallfs.go
```diff
@@ -14,15 +11,20 @@ import (
...
 
 	// OpenFile is similar to os.OpenFile, except the path is relative to this
 	// file system.
 	OpenFile(name string, flag int, perm fs.FileMode) (fs.File, error)
+	// ^^ TODO: switch to syscall.X, not os.X
 
 	// Mkdir is similar to os.Mkdir, except the path is relative to this file
 	// system.
 	Mkdir(name string, perm fs.FileMode) error
+	// ^^ TODO: switch to syscall.X, not os.X
```
mathetake, line 27: what's os.X?

mathetake (approved):

internal/syscallfs/syscallfs.go
codefromthecrypt: os.Mkdir -> syscall.Mkdir except doing this we need to maybe end up mapping some syscall flags which go does for us when we use os. stuff

codefromthecrypt: made a better comment

PR #996, opened 2023-01-01
Ensures 32-bit platforms build

codefromthecrypt:
Trivy builds with linux_386, so this adds a check to ensure we cover 386 arch even if it may or may not work in interpreter (definitely not in compiler)

See https://github.com/aquasecurity/trivy/pull/3355

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,7 @@ format:
 .PHONY: check
 check:
 	@GOARCH=amd64 GOOS=dragonfly go build ./... # Check if the internal/platform can be built on compiler-unsupported platforms
+	@GOARCH=386 GOOS=linux go build ./... # Check if the internal/platform can be built on compiler-unsupported platforms
 	@$(MAKE) lint golangci_lint_goarch=arm64
 	@$(MAKE) lint golangci_lint_goarch=amd64
 	@$(MAKE) format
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -213,6 +213,7 @@ format:
 .PHONY: check
 check:
 	@GOARCH=amd64 GOOS=dragonfly go build ./... # Check if the internal/platform can be built on compiler-unsupported platforms
+	@GOARCH=386 GOOS=linux go build ./... # Check if the internal/platform can be built on compiler-unsupported platforms
```
codefromthecrypt, line 216: this is the missing check, though theoretically we could fail on the various flavors of 32-bit arm or powerpc

internal/syscallfs/dirfs.go
```diff
@@ -83,13 +83,12 @@ func (dir dirFS) Unlink(name string) error {
...
-func (dir dirFS) Utimes(name string, atimeSec, atimeNsec, mtimeSec, mtimeNsec int64) error {
+func (dir dirFS) Utimes(name string, atimeNsec, mtimeNsec int64) error {
 	if !fs.ValidPath(name) {
 		return syscall.EINVAL
 	}
-
 	return syscall.UtimesNano(path.Join(string(dir), name), []syscall.Timespec{
-		{Sec: atimeSec, Nsec: atimeNsec},
-		{Sec: mtimeSec, Nsec: mtimeNsec},
+		syscall.NsecToTimespec(atimeNsec),
```
codefromthecrypt, line 91: the key is using this which implicitly coerces 64 bit to 32bit as needed.

mathetake (approved):

PR #997, opened 2023-01-03
site: Adds notes about how GOARCH=js GOOS=wasm tests work

codefromthecrypt:
Go's tests are configured to require reading of files like /etc/passwd, which was surprising to me.

### DIFF
--- a/site/content/languages/go.md
+++ b/site/content/languages/go.md
@@ -327,14 +327,26 @@ like launching subprocesses on wasm, which won't likely ever support that.
 
 `TestStat` tries to read `/etc/passwd` due to a [runtime.GOOS default][21].
 As `GOARCH=wasm GOOS=js` is a virtualized operating system, this may not make
-sense, as it has no such files. Moreover, as of Go 1.19, tests don't pass
-through any configuration to hint at the real OS underneath the VM.
+sense, because it has no files representing an operating system.
 
-The reason tests in packages like `os` proceed is due to configuration of
-node.js in [wasm_exec_node.js][22]. Specifically, functions that implement file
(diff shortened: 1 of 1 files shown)


mathetake (approved):

codefromthecrypt:
redoing notes a little as noticed windows doesn't work anyway

```bash
$ GOOS=js GOARCH=wasm go test syscall/js
fork/exec C:\Users\fernc\AppData\Local\Temp\go-build2236168911\b001\js.test: %1 is not a valid Win32 application.
FAIL    syscall/js      0.034s
FAIL
```

site/content/languages/go.md
```diff
@@ -298,15 +298,56 @@ like wazero. In other words, go can't run the wasm it just built. Instead,
...
+[wasm_exec_node.js][22], doesn't actually work on Windows.
+
+```bash
+$ GOOS=js GOARCH=wasm go test os
+fork/exec C:\Users\fernc\AppData\Local\Temp\go-build2236168911\b001\os.test: %1 is not a valid Win32 application.
+FAIL    os      0.034s
+FAIL
+```
+
+Hosts like OS/x and Linux pass these tests because they include files like
```
mathetake, line 345: nit: OS/x is not common, and should be Darwin as we are referencing Linux (kernel)

mathetake (approved):

site/content/languages/go.md
codefromthecrypt: yep safe to assume people who are reading this should know OS/x is darwin

codefromthecrypt:
thanks for the look @mathetake!

PR #1002, opened 2023-01-04
Adds checks into the Makefile for 32-bit platforms

codefromthecrypt:
Developers don't commonly think about 32-bit platforms and in one instance we broke a release due to this (quickly fixed in pre.7). This ensures arm and 386, used by dapr and trivy respectively are tested. It also helps prove these are in use to avoid a later developer removing support by thinking they aren't.

Don't read into these results, as they are new releases. This is just an example that shows the output of the commands:
```bash
$ gh release view -R aquasecurity/trivy --json assets --jq 'first(.assets[] | select(.name| test("Linux-32bit.*tar.gz")) | {url, downloadCount})'
{"downloadCount":1,"url":"https://github.com/aquasecurity/trivy/releases/download/v0.36.0/trivy_0.36.0_Linux-32bit.tar.gz"
$ gh release view -R dapr/dapr --json assets --jq 'first(.assets[] | select(.name = "daprd_linux_arm.tar.gz") | {url, downloadCount})'
{"downloadCount":16,"url":"https://github.com/dapr/dapr/releases/download/v1.9.5/dapr-operator.yaml"}
```

apologies to @knqyf263 and @berndverst

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,7 +222,7 @@ check:
 # Ensure we build on linux arm for Dapr:
 #	gh release view -R dapr/dapr --json assets --jq 'first(.assets[] | select(.name = "daprd_linux_arm.tar.gz") | {url, downloadCount})'
 	@GOARCH=arm GOOS=linux go build ./...
-# Ensure we build on linux arm for Dapr:
+# Ensure we build on linux 386 for Trivy:
 #	gh release view -R aquasecurity/trivy --json assets --jq 'first(.assets[] | select(.name| test("Linux-32bit.*tar.gz")) | {url, downloadCount})'
 	@GOARCH=386 GOOS=linux go build ./...
 	@$(MAKE) lint golangci_lint_goarch=arm64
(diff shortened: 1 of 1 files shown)


mathetake (approved):

codefromthecrypt:
I think this check is valid for arm, but would love @berndverst to ack, just to make sure I didn't miss something. Since this runs on every PR, I'm hoping we can use it to trust contributors change wrt 32-bit compilation, without a lot of nervousness.

mathetake:
why this passes without any change while still failing to build on arm/linux?

codefromthecrypt:
long answer

pre.7 fixed platform code which didn't fall back to unsupported properly. The following is the fallback that's correct, basically the opposite of what compiler supports. so pre.7 already *should* work on arm, just this PR explicitly verifies it, at the cost of a longer `make check`

```go
//go:build !(amd64 || arm64) || !(darwin || linux || freebsd || windows)
```

codefromthecrypt:
aside: we aren't currently publishing the `wazero` CLI on release, but doing so would also catch that.. at least building it would. I don't expect we want to release the CLI until the syntax stabilizes in the next month or two

(reminder 1.0 target slipped from end of feb to end of march, so that's when all api drift and cli drift should be done)

codefromthecrypt:
I'll merge this as I think this was the only thing impacting 32-bit, and I also want to have verification as new syscall support are added. If we are missing something else, easy PR!

PR #1009, opened 2023-01-05
Clarifies at semantics and preopen semantics in WASI

codefromthecrypt:
This adds FS.Path which holds the pre-open path currently only used in WASI. It also fixes a TODO where we didn't know for sure if the FD parameter for `path_` functions must always be a pre-open. The TL;DR; is that usually it is, but it may not be (e.g. in our zig-cc example we can see any directory FD, not just pre-opens).

Finally, this fixes a bug in our path resolution where we mistook paths like "foo/foo" for "foo" because we only considered basenames instead of the full path from the pre-open root.

This also makes pre-open directory lookup lazy because I noticed in Trivy specifically, this is unnecessary for us to do eagerly, as they change the FS at runtime per-call. In other words, any value from init time is invalid later.

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -7,7 +7,7 @@ import (
 	"io/fs"
 	"math"
 	"os"
-	"path"
+	pathutil "path"
 
 	"github.com/tetratelabs/wazero/api"
 	"github.com/tetratelabs/wazero/internal/platform"
@@ -233,7 +233,12 @@ func fdFilestatGetFunc(mod api.Module, fd, resultBuf uint32) Errno {
 		return ErrnoFault
 	}
 
-	stat, err := sys.StatFile(fsc, fd)
+	f, ok := fsc.LookupFile(fd)
+	if !ok {
+		return ErrnoBadf
+	}
+
+	stat, err := f.Stat()
 	if err != nil {
 		return ToErrno(err)
 	}
@@ -353,16 +358,9 @@ func fdPrestatGetFn(_ context.Context, mod api.Module, params []uint64) Errno {
 	fsc := mod.(*wasm.CallContext).Sys.FS()
 	fd, resultPrestat := uint32(params[0]), uint32(params[1])
 
-	// Currently, we only pre-open the root file descriptor.
-	if fd != sys.FdRoot {
-		return ErrnoBadf
-	}
-
-	var name string
-	if f, ok := fsc.LookupFile(fd); !ok {
-		return ErrnoBadf // closed
-	} else {
-		name = f.Name
+	name, errno := preopenPath(fsc, fd)
+	if errno != ErrnoSuccess {
+		return errno
 	}
 
 	// Upper 32-bits are zero because...
@@ -414,16 +412,9 @@ func fdPrestatDirNameFn(_ context.Context, mod api.Module, params []uint64) Errn
 	fsc := mod.(*wasm.CallContext).Sys.FS()
 	fd, path, pathLen := uint32(params[0]), uint32(params[1]), uint32(params[2])
 
-	// Currently, we only pre-open the root file descriptor.
-	if fd != sys.FdRoot {
-		return ErrnoBadf
-	}
-
-	var name string
-	if f, ok := fsc.LookupFile(fd); !ok {
-		return ErrnoBadf // closed
-	} else {
-		name = f.Name
+	name, errno := preopenPath(fsc, fd)
+	if errno != ErrnoSuccess {
+		return errno
 	}
 
 	// Some runtimes may have another semantics. See /RATIONALE.md
@@ -931,10 +922,6 @@ func fdSeekFn(_ context.Context, mod api.Module, params []uint64) Errno {
 	whence := uint32(params[2])
 	resultNewoffset := uint32(params[3])
 
-	if fd == sys.FdRoot {
-		return ErrnoBadf // cannot seek a directory
-	}
-
 	var seeker io.Seeker
 	// Check to see if the file descriptor is available
 	if f, ok := fsc.LookupFile(fd); !ok {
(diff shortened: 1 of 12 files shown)


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -1344,32 +1317,47 @@ func pathOpenFn(_ context.Context, mod api.Module, params []uint64) Errno {
...
+// working directory function in WASI, so most assume CWD is "/". Notably, Zig
+// has different behavior which assumes it is whatever the first pre-open name
+// is.
 //
-// See https://github.com/WebAssembly/wasi-libc/blob/659ff414560721b1660a19685110e484a081c3d4/libc-bottom-half/sources/at_fdcwd.c#L24-L26
+// See https://github.com/WebAssembly/wasi-libc/blob/659ff414560721b1660a19685110e484a081c3d4/libc-bottom-half/sources/at_fdcwd.c
 // See https://linux.die.net/man/2/openat
-func atPath(fsc *sys.FSContext, mem api.Memory, dirFd, path, pathLen uint32) (string, Errno) {
-	if dirFd != sys.FdRoot { //nolint
-		// TODO: Research if dirFd is always a pre-open. If so, it should
```
codefromthecrypt, line 1355: this TODO is now complete

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -759,26 +758,35 @@ func Test_fdPrestatGet_Errors(t *testing.T) {
...
-			fd:            fd,
+			fd:            sys.FdPreopen,
 			resultPrestat: memorySize,
 			expectedErrno: ErrnoFault,
 			expectedLog: `
 ==> wasi_snapshot_preview1.fd_prestat_get(fd=3)
 <== (prestat=,errno=EFAULT)
 `,
 		},
-		// TODO: non pre-opened file == api.ErrnoBadf
```
codefromthecrypt, line 781: complete, but actually it is better as ErrnoInval

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -2109,54 +2133,30 @@ func Test_pathFilestatGet(t *testing.T) {
...
 			memory:         initialMemoryNotExists,
 			pathLen:        1,
 			resultFilestat: 2,
 			expectedErrno:  ErrnoNoent,
 			expectedLog: `
 ==> wasi_snapshot_preview1.path_filestat_get(fd=3,flags=,path=?)
 <== (filestat=,errno=ENOENT)
-`,
-		},
-		{
```
codefromthecrypt, line 2122: these tests aren't at the right abstraction, as they are in the filesystem.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -2109,54 +2133,30 @@ func Test_pathFilestatGet(t *testing.T) {
...
-<== (filestat=,errno=ENOENT)
 `,
 		},
 		{
 			name:          "path is out of memory",
-			fd:            sys.FdRoot,
+			fd:            sys.FdPreopen,
 			memory:        initialMemoryFile,
 			pathLen:       memorySize,
-			expectedErrno: ErrnoNametoolong,
```
codefromthecrypt, line 2151: wrong errno

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -2044,12 +2068,12 @@ func Test_pathFilestatGet(t *testing.T) {
...
-			memory:         initialMemoryFile,
-			pathLen:        1,
-			resultFilestat: 2,
+			fd:             sys.FdPreopen, // root
+			memory:         initialMemoryFileInDir,
+			pathLen:        uint32(len(initialMemoryFileInDir)) - 1,
+			resultFilestat: uint32(len(initialMemoryFileInDir)),
 			expectedMemory: append(
-				initialMemoryFile,
+				initialMemoryFileInDir,
```
codefromthecrypt, line 2076: this test was wrong, and the data now shows we can tell the difference between a subpath and a root path.

internal/sys/fs.go
```diff
@@ -19,46 +17,25 @@ const (
...
-	FdRoot
+	FdPreopen
 )
 
 const (
 	modeDevice     = uint32(fs.ModeDevice | 0o640)
 	modeCharDevice = uint32(fs.ModeCharDevice | 0o640)
 )
 
-// An emptyRootDir is a fake "/" directory.
```
codefromthecrypt, line 41: I traded the fake stuff for lazy, based on looking at how trivy works

internal/syscallfs/adapter.go
```diff
@@ -16,13 +16,20 @@ func Adapt(fs fs.FS) FS {
...
+}
 
 // Open implements the same method as documented on fs.FS
 func (ro *adapter) Open(name string) (fs.File, error) {
 	panic(fmt.Errorf("unexpected to call fs.FS.Open(%s)", name))
 }
 
+// Path implements FS.Path
+func (ro *adapter) Path() string {
+	return "/"
```
codefromthecrypt, line 30: these stub in "/" until a later PR which will make the pre-open path configurable.

internal/syscallfs/syscallfs.go
```diff
@@ -13,6 +13,25 @@ import (
 //
 // See https://github.com/golang/go/issues/45757
 type FS interface {
+	// Path is the name of the path the guest should use this filesystem for,
```
codefromthecrypt, line 16: added some interesting notes here

codefromthecrypt (commented):
As a lot of this is refactoring, I highlighted the important parts.

codefromthecrypt:
I will sort out the windows error prior to merge

codefromthecrypt:
🤞 windows should work in CI now (works on my laptop).

mathetake (approved):

PR #1010, opened 2023-01-05
Moves FreeBSD to compilation test only.

codefromthecrypt:
FreeBSD is not natively supported by GitHub actions or Docker. To test it, we had to run vagrant which is both time-consuming (~7minutes), and frequently fails to setup, resulting in false negatives. Meanwhile the code related to FreeBSD is quite limited and doesn't change.

This changes our approach to testing FreeBSD to the same as how we deal with 386, which is to do compilation tests only.


mathetake (approved):

Makefile
```diff
@@ -225,6 +225,9 @@ check:
 # Ensure we build on linux 386 for Trivy:
 #	gh release view -R aquasecurity/trivy --json assets --jq 'first(.assets[] | select(.name| test("Linux-32bit.*tar.gz")) | {url, downloadCount})'
 	@GOARCH=386 GOOS=linux go build ./...
+# Ensure we build on FreeBSD 386 for Trivy:
+#	gh release view -R aquasecurity/trivy --json assets --jq 'first(.assets[] | select(.name| test("FreeBSD-32bit.*tar.gz")) | {url, downloadCount})'
+	@GOARCH=386 GOOS=freebsd go build ./...
```
mathetake, line 230: oh no this should be amd64

codefromthecrypt: yep actually they build also on 386 but more important is amd64 agree

mathetake: as amd64 is  the original issue why we have started testing https://github.com/tetratelabs/wazero/issues/607

codefromthecrypt: cool. and verified the Makefile comment, also.

```bash
$ gh release view -R aquasecurity/trivy --json assets --jq 'first(.assets[] | select(.name| test("FreeBSD-64bit.*tar.gz")) | {url, downloadCount})'
{"downloadCount":13,"url":"https://github.com/aquasecurity/trivy/releases/download/v0.36.0/trivy_0.36.0_FreeBSD-64bit.tar.gz"}
```

PR #1014, opened 2023-01-06
add syscallfstest package

achille-roussel:
This PR is a follow up to a conversation @codefromthecrypt and I had on Slack about bringing a reusable test suite to eventually export the `syscallfs.FS` interface.

This is a first step which ports the `TestDirFS_*` tests to use the test suite, which should include all tests except for the windows-specific error handling:
```
--- PASS: TestDirFS (0.01s)
    --- PASS: TestDirFS/OpenFile (0.00s)
        --- PASS: TestDirFS/OpenFile/existing_directories_can_be_opened (0.00s)
        --- PASS: TestDirFS/OpenFile/existing_files_can_be_opened (0.00s)
        --- PASS: TestDirFS/OpenFile/files_can_be_created_in_the_file_system (0.00s)
        --- PASS: TestDirFS/OpenFile/opening_files_which_do_not_exist_fails_with_ENOENT (0.00s)
    --- PASS: TestDirFS/Mkdir (0.00s)
        --- PASS: TestDirFS/Mkdir/create_a_directory_in_place_where_a_file_exists_fails_with_EEXIST (0.00s)
        --- PASS: TestDirFS/Mkdir/create_a_directory_which_already_exists_fails_with_EEXIST (0.00s)
        --- PASS: TestDirFS/Mkdir/directories_can_be_created_in_the_file_system (0.00s)
    --- PASS: TestDirFS/Rmdir (0.00s)
        --- PASS: TestDirFS/Rmdir/empty_directories_can_be_removed (0.00s)
        --- PASS: TestDirFS/Rmdir/removing_a_directory_which_does_not_exist_fails_with_ENOENT (0.00s)
        --- PASS: TestDirFS/Rmdir/removing_a_directory_which_is_not_empty_fails_with_ENOTEMPTY (0.00s)
        --- PASS: TestDirFS/Rmdir/removing_a_file_which_is_not_a_directory_fails_with_ENOTDIR (0.00s)
    --- PASS: TestDirFS/Rename (0.00s)
        --- PASS: TestDirFS/Rename/a_directory_can_be_moved_to_itself (0.00s)
        --- PASS: TestDirFS/Rename/a_file_can_be_moved_to_itself (0.00s)
        --- PASS: TestDirFS/Rename/directories_can_be_moved_to_a_location_where_a_directory_already_exists (0.00s)
        --- PASS: TestDirFS/Rename/files_can_be_moved_to_a_location_where_a_file_already_exists (0.00s)
        --- PASS: TestDirFS/Rename/files_can_be_moved_to_an_existing_directory (0.00s)
        --- PASS: TestDirFS/Rename/moving_a_directory_to_a_location_where_a_file_exists_fails_with_ENOTDIR (0.00s)
        --- PASS: TestDirFS/Rename/moving_a_directory_to_a_location_which_does_not_exist_fails_with_ENOENT (0.00s)
        --- PASS: TestDirFS/Rename/moving_a_file_to_a_location_where_a_directory_exists_fails_with_EISDIR (0.00s)
        --- PASS: TestDirFS/Rename/moving_a_file_which_does_not_exist_fails_with_ENOENT (0.00s)
    --- PASS: TestDirFS/Unlink (0.00s)
        --- PASS: TestDirFS/Unlink/existing_files_can_be_unlinked_and_do_not_exist_in_the_directory_anymore_afterwards (0.00s)
        --- PASS: TestDirFS/Unlink/unlinking_a_directory_fails_with_EISDIR (0.00s)
        --- PASS: TestDirFS/Unlink/unlinking_a_file_which_does_not_exist_fails_with_ENOENT (0.00s)
    --- PASS: TestDirFS/Utimes (0.00s)
        --- PASS: TestDirFS/Utimes/changing_time_of_a_file_which_does_not_exist_fails_with_ENOENT (0.00s)
        --- PASS: TestDirFS/Utimes/times_of_existing_directories_can_be_changed (0.00s)
        --- PASS: TestDirFS/Utimes/times_of_existing_directories_can_be_set_to_zero (0.00s)
        --- PASS: TestDirFS/Utimes/times_of_existing_files_can_be_changed (0.00s)
        --- PASS: TestDirFS/Utimes/times_of_existing_files_can_be_set_to_zero (0.00s)
```

Please let me know if you would like to see anything revisisted, I am recording below the context I wanted to share with the pull request.

## Context

From a design perspective, I modeled the test suite following test suites like [testing/fstest](https://pkg.go.dev/testing/fstest#TestFS) and [x/net/nettest](https://pkg.go.dev/golang.org/x/net/nettest#TestConn), which also aim to validate the behavior of interface implementations.

To limit exposure of wazero internals in this test suite, I chose not to rely on the `internal/testing/require` package for error validation, and instead implement the tests as functions returning errors with this signature:
```go
func(syscallfs.FS) error
```
One of the advantages of this approach is that it makes each test a code snippet somewhat similar to a code example of how to use the `syscallfs.FS` interface, since the code looks a lot more like how a *regular* program would use the interface (regarding invocation and error handling).

The test suite is implemented as a function with this signature:
```go
func TestReadWriteFS(t *testing.T, makeFS MakeFS)
```
The `MakeFS` parameter is a function instantiating a file system intended to be used for the duration a single test. This provides a strong isolation mechanism to ensure that state mutations performed by tests do not overlap with one another.

I believe that it will be valuable to have a test suite for *read-only* file systems as well, which is why I am naming this function `TestReadWriteFS`, expecting to introduce a `TestReadOnlyFS` test suite later. Such test suites could be useful to validate implementations of `syscallfs.FS` which are backed by a `fs.FS` for example (which are read-only).

I have also exposed the building blocks of the file system test suite as global variables named `OpenFileTests`, `MkdirTests`, etc... With the intent being to allow external dependencies to reuse parts of the test suite in cases where some implementations cannot satisfy all the tests.

I used full english sentences asserting the expectation of the tests instead of the more common short descriptions that we can find in Wazero test suites. Since these tests are intended to be used outside of the Wazero code base, I the logs will show when running tests in external code bases, bringing as much context as we can close to the user will likely help create value from the test suite (as a user, I know that this is what I would like to have).

## Follow-up

While working on this, there are a few thoughts that I would like to record here to orient follow up work:

- I would like to encourage focusing on portability; considering for example the slight behavior differences between windows and other platforms when moving directories: windows disallows moving to an existing directory. However, could we emulate the POSIX behavior by first removing the target directory and then moving the source. This would make it a non-atomic operation in a multi-thread environment (which Wazero isn't yet) but it would ensure that we reduce the system-dependent behavior of applications run by Wazero. See https://github.com/achille-roussel/wazero/pull/1/files for a demonstration of how we could mimic the POSIX behavior on windows.

- I would like to suggest that we raise the abstraction level a bit: from an external use standpoint, dealing with syscall-level abstractions (which tend to be very platform-dependent) tend to be a lot more difficult than working with a higher-level API like what the standard `os` package provides for example. Here is a list of potential changes that we could make:
  - rename `syscallfs` to something like `sysfs`, the current name is very verbose and in my opinion, damages readability when repeated very frequently. Personally I still think that using the top-level `sys` package would be an even better option since it's an abbreviation which conveys a lot of meaning with reading code like `sys.FS` (no redundancy, clear understanding of what the scope and type are).
  - Change `Utimes` to `Chtimes`, and use `os.Chtimes` in the implementation. Go developers are familiar with using `time.Time` to manage time values, which also brings greater compile-time validation than using timestamps encoded as integers. I demonstrated this change in https://github.com/achille-roussel/wazero/pull/2/files
  - Use `fs.Err*` errors instead of `syscall.E*` errno codes: the former already matches against the syscall errno codes but is more familiar to Go developers and helps provide a level of abstraction to help adapt platform-specific variations. In cases where the `fs` package does not have error variables, we can provide them in Wazero. Overall, I think we should avoid leaking usage of the `syscall` package outside of Wazero, it should be possible to implement the `FS` interface without having to import `syscall`.
  - Wrap all errors except `io.EOF` in `fs.PathError`: while the context of the Go error cannot be carried through the WASI ABI, it is very helpful when developing and testing implementations of the file system interface. I had experimented with validating this as part of the test suite in https://github.com/achille-roussel/wazero/blob/sysfs/experimental/systest/fs.go#L58-L64 and found it really useful to detect variations across platforms or functions.

- I think it would be very helpful to define a `File` interface which declares the methods that must exist on files that can back WASM file systems, and have `OpenFile` return a value of this `File` type instead of `fs.File`. This would help developers use the Go type system to ensure that they are not missing methods (e.g. typo `Write` for `Wirte` -> if implementing `io.Writer` is optional, both the compiler and the test suite might be defeated to believe that the developer intended to not provide the method, such bugs might even reach production instead of being caught at compile time). I published an example of this change in https://github.com/achille-roussel/wazero/pull/3/files.

- The recent changes to the `syscallfs` package have broke compatiblity with `fs.FS`: the `fstest.TestFS` test suite used to pass in this branch a couple of days ago, it is not the case anymore because we accept paths starting with `/` now. I think it would be useful to satisfy the `fs.FS` interface for read-only use cases. We could then reuse the standard `fstest` test suite, and create more opportunities to integrate with the Go ecosystem.


internal/syscallfs/dirfs.go
```diff
@@ -24,7 +23,7 @@ type dirFS string
 
 // Open implements the same method as documented on fs.FS
 func (dir dirFS) Open(name string) (fs.File, error) {
-	panic(fmt.Errorf("unexpected to call fs.FS.Open(%s)", name))
+	return dir.OpenFile(name, 0, 0)
```
achille-roussel, line 26: I changed this when the `syscallfs.FS` type was still declaring that it implemented `fs.FS` so that the implementations could pass the `fstest.TestFS` test suite, but I can revert it if we don't want to validate this as part of the pull request.

internal/syscallfs/syscallfs_test.go
```diff
@@ -6,15 +6,32 @@ import (
...
+
+	testFS_Open_Read(t, tmpDir, testFS)
+
+	t.Run("path outside root valid", func(t *testing.T) {
+		_, err := testFS.OpenFile("../foo", os.O_RDONLY, 0)
+
+		// syscall.FS allows relative path lookups
+		require.True(t, errors.Is(err, fs.ErrNotExist))
+	})
+}
```
achille-roussel, line 33: I moved this test here for now so `dirfs_test.go` can be in the `syscallfs_test` package to avoid cyclic dependencies with `syscalltestfs` and `syscallfs`.

codefromthecrypt:
Thanks for the spike on this @achille-roussel TL;DR; I'll try to merge in what is used here, but nothing that tries to change the abstraction or make this an fs.FS underneath again.

This SPI is literally a syscall abstraction and outside implementations of this will be rare. fs.FS will be a first class implementation, and bugs on that are bugs. Another abstraction would imply routine use and usually the rule of abstraction is that there are multiple independent implementations. We should remain cautious about trying to abstract the plugin, as it is definitely premature considering we haven't event tested the real abstraction used by spis (syscall)

Here's an example. When errors don't match syscalls they crash wasm. This is behind me testing exactly what codes are returned! For example, there are mechanisms both in WASI and Go that only work when the error code matches. For this reason, moving from syscall errors to os. errors for the sake of portability is actually optimizing the wrong thing, at least in this stage of design. For this reason, at least until all the WASI tests are in place, using a non-syscall abstraction is an explicit non-goal. I'll make a RATIONALE.md comment to that effect.

WRT the break on fs.FS as well read-only things, let's keep those as issues with the adapter, not the syscallfs itself. As you'll notice there's less and less overlap with anything related to go's fs.FS the more we complete functionality used in WASI and GOOS=js.

achille-roussel:
Thanks for sharing the extra context.

I believe that in this PR there is nothing that currently changes the interface of `syscallfs.FS` nor its behavior, it's simply a refactor of the existing tests to be reusable and validate alternative implementations of the interface.

As to the use of the `syscall` package, I feel like I am probably missing some context to fully understand your comment:

> Here's an example. When errors don't match syscalls they crash wasm. This is behind me testing exactly what codes are returned! For example, there are mechanisms both in WASI and Go that only work when the error code matches.

From what I understand, WASI should define the set of error codes that each function of the ABI can return, and if Wazero returns the wrong code under some situation then it is undefined behavior land.

However, I don't think there is a guarantee that the WASI error codes will match the `syscall` error codes since the latter are system-dependent. WASI will define the values of each of its errors independently so that WASM applications remain portable across runtimes.

Taking this into account, Wazero has to provide a translation layer from error codes of the OS to error codes of either WASI or GOOS=js.

I think that what I am suggesting is that we make the errors returned `syscallfs.FS` standard across platforms so we have the conversion from OS->Wazero validated in the `FS` test suite, then Wazero is in charge of converting errors to the proper codes when returning them to the application.

This could be as simple as redefining the constants in the `syscallfs` package for the errors used by the `FS` interface (which is somewhat equivalent to declare error variables):
```go
type Errno = syscall.Errno

const (
  EINVAL Errno = syscall.EINVAL
  ...
)
```
Which would help make the abstraction layer portable and not require spreading dependency on the semi-deprecated and platform-dependent `syscall` package of the standard library.

> WRT the break on fs.FS as well read-only things, let's keep those as issues with the adapter, not the syscallfs itself. As you'll notice there's less and less overlap with anything related to go's fs.FS the more we complete functionality used in WASI and GOOS=js.

My take on this is we should default to being compatible with the standard `fs.FS` package, unless there it is truly incompatible with achieving the goals. From the spikes I did, I have not run into a case where it did not seem like a quality to maintain compatibility with the standard library.

As a Wazero maintainer, you will obviously have higher stakes in the design decision here, so you'll make the call, but I really think it's worth pursuing compatibility if we can.

achille-roussel:
Ah, and maybe here's where my confusion comes from:

> This SPI is literally a syscall abstraction and outside implementations of this will be rare. fs.FS will be a first class implementation, and bugs on that are bugs.

I'm working from the standpoint of: we eventually want to expose some type of `FS` abstraction in Wazero which allows users (like me :D) to modify the file system backend. I made the assumption that `syscallfs.FS` would eventually graduate to being this abstraction, but maybe that was the wrong idea. I don't know if the implementations will be rare, but I have multiple use cases in the pipeline already, so from where I stand it's not that rare!

We could also have another bridge layer which converts between the "public" `FS` and `syscallfs.FS`. For this to be possible, `syscallfs` needs to be somewhat exposed externally, at least in a function signature like:
```go
func NewSyscallFS(baseFS FS) syscallfs.FS { ... }
```

My intuition is this layer of abstraction seems a bit overkill: it's more interfaces to test, more code to maintain, one interface should be sufficient do the job? This is why I imagined `syscallfs.FS` would eventually turn into the public interface.

codefromthecrypt:
Thanks for the replies.

Main thing I can try to emphasize is that until implementation is complete, please consider the interface an implementation detail and tactical until we pass the tests we need to. Nothing is zero sum game with fs.FS, just that we know for sure that fs.FS doesn't cover many things that are needed underneath to implement syscalls. We also don't want translation everywhere at each ABI, and there will be minimally three ABIs, as wasi-filesystem will certainly complete and we don't want to have to re-make a new plugin when that happens.


Each time we circle around trying to imply that the internal implementation of fs.FS and os.File bindings is an abstraction, we lose focus. I have to respond back similarly to ideas that are on a tactical implementation. This focus loss comes at the cost of doing the work, as I'm already working evenings and weekends to do what is needed as fast as practical, but without skipping steps like windows or GOOS=js, etc. This is why I really request to please stop cycling about future design of abstraction things until we have both implemented and tested the functions needed to support FS abis, here in this repo. Many things will drift between now and then. syscall errors may or may not be possible to re-use, but only tests will prove that's the case, and at least they will be done to the point where we aren't reliant on ad-hoc manual tests to figure this out. Once the ground work is done, we can figure out what's reusable, what the concrete extension use cases are, and what to do about abstraction beyond os.File and fs.FS..

Is this fair? If not, and we have to re-discuss abstraction each time, I really think we need to not just push wazero out till march, but also maybe april. This is purely a function of time and focus and if we aren't able to implement prior to abstraction the impact is more time.

achille-roussel:
> Is this fair? If not, and we have to re-discuss abstraction each time, I really think we need to not just push wazero out till march, but also maybe april. This is purely a function of time and focus and if we aren't able to implement prior to abstraction the impact is more time.

Your response is totally fair!

I apologize if I am making things harder than they should be, please know that it was not my intention to steal focus away from your priorities. I want to help but without good visibility into your roadmap, it is difficult to find an angle to help move things forward as an external contributor.

I understand the iterative process of software development, and that what exists in Wazero will evolve over time. As a user, there are goals I would like to achieve with the tech you've built, which is why I am getting involved and offering my help.

I'm sorry to hear you are under heavy workload these days. I have free cycles to help so please let me know if there is something I can do! Otherwise I'll just hold on for now as it seems it might just not be the right time to bring the contributions I am trying to get into Wazero, I can simply continue working in parallel on a fork.

internal/syscallfs/dirfs.go
codefromthecrypt: I think for this we can add a test-only fs wrapper instead, since the purpose is only for tests using `fstest`. In that, we'd have to follow the fs.FS rules like cleaning the path and making sure it is relative as iirc they will test them.

We can add to that test wrapper each interface we use internally besides `fs.FS`. For example, right now, we don't use anything in adapter.go except `fs.FS`, but if we did use something else, we can implement that.

codefromthecrypt (commented):
In summary, there's some behavioral changes that need sifting through. Not to say they are wrong, just it will take time to inventory what is new, changed or deleted.

Here are the longer notes, and sadly I'm out of time to continue this right now. You can choose to park it as maybe I can get the existing tests into a closer shape to this offline, making your driftwork easier. Landing this isn't urgent so feel free to leave it stalled for a bit.


---

One thing the re-organization is interesting and feels nicer than before, at the cost of more abstraction which makes it a bit harder to understand what is tested vs before. I think in some cases the tests are different than before, not just rejigged to reduce copy/paste and increase elegance. Some of the existing tests were hours of work, so we can't delete them without consideration. For this reason, I need more analysis to figure out what is now missing, and once inventoried, how to move it back. Also, since this doesn't do the read-only side, I feel it will need to change again even after figuring it out.

It is interesting to use fstest as a secondary check even though our primary integration tests will be through os tests of zig and tinygo. We don't need to worry about features of fstest we don't use e.g. we don't use ReadDirFS as all reads are via an opened file descriptor. In any case I think adapting can allow us to use it if we want, at the cost of having to deal with the nuance that implies (like prefixes etc)

What I noticed first is it seems more abstracted than just making less repetition between the various FS backed implementations. For example, it is using the subject of test itself to also setup test files, notably creating new files and directories. This seems convenient but there's also a chicken egg issue, for example, concretely dirfs is real files. It seems we need to verify that they are in fact created properly before depending on those functions working prior to other tests. This will be more obvious when getting to read-only tests, where you can't use the filesystem to create files to test itself with.

Issue #1013, opened 2023-01-06
Expose the FS interface


wjam:
**Is your feature request related to a problem? Please describe.**
The way filesystems were accessed by WASM modules was recently changed (https://github.com/tetratelabs/wazero/commit/15dc3d7d371c86470c8bb0cd9f775ed728e766e7#diff-e1cd45f905e175c6cc7e0298d2fa2efcf732d7182465e7e73614c10132e5a3a5 I believe, but certainly between https://github.com/tetratelabs/wazero/compare/v1.0.0-pre.5...v1.0.0-pre.6). This change altered what was needed to get a writeable filesystem - needing to implement [the `FS` interface](https://github.com/tetratelabs/wazero/blob/8e78f3fb0488ea233472c6662280c1a6ec82c89f/internal/syscallfs/syscallfs.go#L15). For example, this interface would need to be implemented if you wanted to certain files to come from an embedded `fs.FS` or if you wanted certain files to be immutable.

The problem is that the wazero `FS` interface lives in an `internal/` package and so encourages people to _not_ implement this interface. 

**Describe the solution you'd like**
Move the interface somewhere more public so that it can be discovered by developers and implemented.

**Describe alternatives you've considered**
* Keep using a plain `fs.FS` - this won't work because attempting to create files is blocked by https://github.com/tetratelabs/wazero/blob/4197caa05bb6fd529eecae56a9eea6bcb1b1780e/internal/sys/fs.go#L312

**Additional context**

achille-roussel:
Hello @wjam!

I am not part of the core maintainer team of Wazero but I have been in touch with @codefromthecrypt in the #wazero-dev Slack channel on this topic.

I believe that the goal is eventually to expose the `FS` interface, but at this time the code in `internal/syscallfs` isn't stable enough yet to allow external programs to take a dependency on it.

codefromthecrypt:
@wjam thanks for the feedback!

Currently, the internal filesystem is very much not encouraged to be implemented. It is changing almost every day, so not ready to consider exposing. Moreover, no one should think too deeply about its current form given that it is changing.

### What is currently supported until our internal interface is done.

At a high level, we currently only support `fs.FS` and an existing directory as entrypoints. As you know `fs.FS` to create new files is a hack around a lack of an alternative. We can add an unsafe FS adapter meanwhile to help you along. I'll raise one even though the goal is to get the real filesystem together this month.

### Why we are focusing on syscalls and when do we think we are done our internal interface

[...]

PR #1015, opened 2023-01-07
fs: Adds tests to prove we allow fs.FS to create new files

codefromthecrypt:
This ensures those who cannot use `writefs.DirFS` can use a `fs.FS` implementation that creates or opens files as read/write. This does so by making a hacked fs.FS and testing it.

```go
type hackFS string

func (dir hackFS) Open(name string) (fs.File, error) {
	path := ensureTrailingPathSeparator(string(dir)) + name

	if f, err := os.OpenFile(path, os.O_RDWR, 0); err == nil {
		return f, nil
	} else if errors.Is(err, syscall.EISDIR) {
		return os.OpenFile(path, os.O_RDONLY, 0)
	} else {
		return nil, err
	}
}
```

This also removes type filtering to ensure hacked filesystems aren't accidentally masked. Instead we get more specific on error codes returned by functions such as `os.File.Write`.

---

Note: This is meant to extend the life of fs.FS for custom implementations, though we expect eventually to have our own API. However, such an API will remain internal until at least the end of the month due to a lot of prerequisite work. See #1013 cc @wjam

### DIFF
--- a/internal/syscallfs/dirfs.go
+++ b/internal/syscallfs/dirfs.go
@@ -4,17 +4,28 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
-	"path"
 	"syscall"
 )
 
 func NewDirFS(dir string) (FS, error) {
+	if dir == "" {
+		panic("empty dir")
+	}
+	// For easier OS-specific concatenation later, append the path separator.
+	dir = ensureTrailingPathSeparator(dir)
 	if stat, err := os.Stat(dir); err != nil {
 		return nil, syscall.ENOENT
 	} else if !stat.IsDir() {
 		return nil, syscall.ENOTDIR
 	}
 	return dirFS(dir), nil
+}
+
+func ensureTrailingPathSeparator(dir string) string {
+	if dir[len(dir)-1] != os.PathSeparator {
+		return dir + string(os.PathSeparator)
+	}
+	return dir
 }
 
 // dirFS currently validates each path, which means that input paths cannot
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
windows is every day :p

internal/syscallfs/dirfs.go
```diff
@@ -34,20 +45,16 @@ func (dir dirFS) Path() string {
 
 // OpenFile implements FS.OpenFile
 func (dir dirFS) OpenFile(name string, flag int, perm fs.FileMode) (fs.File, error) {
-	f, err := os.OpenFile(path.Join(string(dir), name), flag, perm)
+	f, err := os.OpenFile(dir.join(name), flag, perm)
 	if err != nil {
 		return nil, err
 	}
-
-	if flag == 0 || flag == os.O_RDONLY {
```
codefromthecrypt, line 42: instead of masking here, we are verifying the errno on Write. seems to work e.g. flags are honored to prohibit writes.

internal/syscallfs/readfs.go
```diff
@@ -28,10 +29,70 @@ func (r *readFS) Path() string {
...
+	if err != nil {
+		return nil, err
+	}
+	return maskForReads(f), nil
+}
+
+// maskForReads masks the file with read-only interfaces used by wazero.
+//
+// This technique was adapted from similar code in zipkin-go.
+func maskForReads(f fs.File) fs.File {
```
codefromthecrypt, line 46: this was moved as it is only used here now

mathetake (approved):

PR #1016, opened 2023-01-09
Introduces Cache API

mathetake:


cache.go
```diff
@@ -0,0 +1,29 @@
+package wazero
+
+import (
+	"context"
+
+	"github.com/tetratelabs/wazero/internal/wasm"
+)
+
+// NewCache returns a new Cache to be passed to RuntimeConfig.
+func NewCache() Cache {
```
mathetake, line 10: so the question is if we should add CacheConfig or not
```suggestion
func NewCache(CacheConfig) Cache {
```

mathetake: maybe we can just have WithFileCacheDir() to `type Cache interface` and then we don't need CacheConfig

mathetake:
no intention to neither delete namespace in this PR

cache.go
```diff
@@ -27,3 +62,46 @@ func (c *cache) Close(_ context.Context) (err error) {
...
 	return
 }
+
+// WithCompilationCacheDirName implements the same method on the Cache interface.
+func (c *cache) WithCompilationCacheDirName(dir string) error {
+	return c.withCompilationCacheDirName(dir, version.GetWazeroVersion())
+}
+
+// WithCompilationCacheDirName implements the same method on the Cache interface.
+func (c *cache) withCompilationCacheDirName(dir string, wazeroVersion string) error {
```
mathetake, line 72: this is moved from experimental/compilation_cache

cache_example_test.go
```diff
@@ -7,49 +7,47 @@ import (
 	"os"
 
 	"github.com/tetratelabs/wazero"
-	"github.com/tetratelabs/wazero/experimental"
 )
 
 // This is a basic example of using the file system compilation cache via WithCompilationCacheDirName.
 // The main goal is to show how it is configured.
-func Example_withCompilationCacheDirName() {
+func Example_withCache() {
```
mathetake, line 14: same - moved from experimentak pkg

cache_test.go
```diff
@@ -89,3 +93,72 @@ func getCacheSharedRuntimes(ctx context.Context, t *testing.T) (foo, bar *runtim
 	require.Equal(t, foo.cache, bar.cache)
 	return
 }
+
+func TestWithCompilationCacheDirName(t *testing.T) {
```
mathetake, line 97: same

cache.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/tetratelabs/wazero/api"
+	"github.com/tetratelabs/wazero/internal/compilationcache"
+	"github.com/tetratelabs/wazero/internal/version"
+	"github.com/tetratelabs/wazero/internal/wasm"
+)
+
+// Cache is the configuration for caching behavior of wazero.Runtime and can be passed to wazero.RuntimeConfig.
+type Cache interface {
+	// Closer can be used to clear the in-memory cache and this doesn't delete this cache directory when configured.
```
mathetake, line 20: I think this isn't very clear - maybe better to have separate Close APIs depending on deleting the file caches or not.

mathetake: having said that, given that so far no one has requested delete file behavior on the experimental.CompilationCache API, so maybe this is just fine

cache.go
```diff
@@ -0,0 +1,99 @@
...
+	// Regardless of the usage of this, the compiled functions are cached in memory, but its lifetime is
+	// bound to the lifetime of wazero.Cache.
+	//
+	// If the dirname doesn't exist, this creates the directory.
+	//
+	// With the given non-empty directory, wazero persists the cache into the directory and that cache
+	// will be used as long as the running wazero version match the version of compilation wazero.
+	//
+	// Note: The embedder must safeguard this directory from external changes.
+	WithCompilationCacheDirName(dir string) error
```
mathetake, line 33: maybe simply 
```suggestion
	WithCacheDir(dir string) error
```

codefromthecrypt:
thanks. I will look into this this evening!

cache.go
codefromthecrypt: do we expect this to be used for anything except compilation? If not, I think we should say CompileCache, as it matched CompileModule (or something like that). then when you link the godoc together it seems more intuitive

cache.go
codefromthecrypt: I don't think the runtime(closeable) object should have a function to change its directory.

We may be ok having `NewCompileCache` and `NewCompileDiskCache` instead?

codefromthecrypt: not a fan of the last name, basically it is a write-behind directory...

mathetake: perfect!

cache_example_test.go
```diff
@@ -7,49 +7,48 @@ import (
...
+	// Creates a new cache context with the file cache directory configured.
+	cache := wazero.NewCache()
+	defer cache.Close(ctx)
+	if err := cache.WithCompilationCacheDirName(cacheDir); err != nil {
+		log.Fatal(err)
 	}
 
+	// Creates a runtime configuration to create multiple runtimes.
+	config := wazero.NewRuntimeConfig().
+		WithCache(cache)
```
codefromthecrypt, line 33: I'd like the cache to not actually do anything until the runtime is created, if possible. e.g. don't create the directory until you have a runtime. ack there's some tension here and it implies some internal types

mathetake: the problem here is that NewRuntime doesn't return error.. so wdyt on handling the error on directory creation (or not a dir error)?

codefromthecrypt: maybe it is better to make NewRuntime return an error. It just was we had no reason to before.

Then the logic is similar to instantiate which bundles up any sort of problem leading to it.

codefromthecrypt: If we make NewRuntime return an error we can add MustRuntime to allow people to assume it won't and panic. it is an api break anyway when we remove namespace, and I can help with the rounds on it.

codefromthecrypt: or.. we accept if the ony thing that can fail is directory creation, we add error return to the NewCompileCacheFromDisk and leave it at that.. is that the only possible error? If so mainly this turns more of the work to defer using it until NewRuntime.

cache.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// The returned CompileCache can be used to share the in-memory compilation results across multiple instances of wazero.Runtime.
+func NewCache() CompileCache {
+	return &cache{}
+}
+
+// NewCompileCacheWithDir is the same as Cache returned by wazero.NewCache except that this also persists
+// the compilation results into the directory specified by `dirname` parameter.
+//
+// If the dirname doesn't exist, this creates the directory. And if the directory creation fails, this returns an error.
```
codefromthecrypt, line 32:
```suggestion
// If the dirname doesn't exist, this creates it or returns an error.
```

cache.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// If the given directory already exists, and it's a file, this returns an error.
```
codefromthecrypt, line 34: probably not needed to clarify this, we can spend the words on  why to use this instead.
```suggestion
Those running wazero as a CLI or frequently restarting a process using the same wasm should
use this feature to reduce time waiting to compile the same module a second time.
```

cache.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// With the given non-empty directory, wazero persists the cache into the directory and that cache
```
codefromthecrypt, line 36: maybe change to the higher level point "The contents written into `dirname` are wazero-version specific, meaning different versions of wazero will duplicate entries for the same input wasm."

cache.go
```diff
@@ -0,0 +1,98 @@
...
+	"path/filepath"
+	goruntime "runtime"
+
+	"github.com/tetratelabs/wazero/api"
+	"github.com/tetratelabs/wazero/internal/compilationcache"
+	"github.com/tetratelabs/wazero/internal/version"
+	"github.com/tetratelabs/wazero/internal/wasm"
+)
+
+// CompileCache is the configuration for compilation cache behavior of wazero.Runtime and can be passed to wazero.RuntimeConfig.
```
codefromthecrypt, line 18:
```suggestion
// CompileCache reduces time spent compiling (Runtime.CompileModule) the same wasm module.
//
// Instances of this can be reused across multiple runtimes, if configured via RuntimeConfig.
```

cache_example_test.go
```diff
@@ -7,49 +7,47 @@ import (
...
 
 	"github.com/tetratelabs/wazero"
-	"github.com/tetratelabs/wazero/experimental"
 )
 
-// This is a basic example of using the file system compilation cache via WithCompilationCacheDirName.
+// This is a basic example of using the file system compilation cache via wazero.NewCompileCacheWithDir.
 // The main goal is to show how it is configured.
-func Example_withCompilationCacheDirName() {
+func Example_withCache() {
```
codefromthecrypt, line 14:
```suggestion
func Example_compileCache() {
```

cache_example_test.go
```diff
@@ -7,49 +7,47 @@ import (
...
 
-	// Append the directory into the context for configuration.
-	ctx, err := experimental.WithCompilationCacheDirName(context.Background(), cacheDir)
+	ctx := context.Background()
+
+	// Creates a new cache context with the file cache directory configured.
+	cache, err := wazero.NewCompileCacheWithDir(cacheDir)
 	if err != nil {
-		log.Panicln(err)
+		log.Fatal(err)
```
codefromthecrypt, line 27:
```suggestion
		log.Panicln(err)
```

cache_example_test.go
```diff
@@ -7,49 +7,47 @@ import (
...
 	if err != nil {
 		log.Panicln(err)
 	}
 	defer os.RemoveAll(cacheDir)
 
-	// Append the directory into the context for configuration.
-	ctx, err := experimental.WithCompilationCacheDirName(context.Background(), cacheDir)
+	ctx := context.Background()
+
+	// Creates a new cache context with the file cache directory configured.
```
codefromthecrypt, line 24:
```suggestion
	// Create a runtime config which shares a compilation cache directory.
```

cache_example_test.go
```diff
@@ -7,49 +7,47 @@ import (
...
+	cache, err := wazero.NewCompileCacheWithDir(cacheDir)
 	if err != nil {
-		log.Panicln(err)
+		log.Fatal(err)
 	}
+	defer cache.Close(ctx)
+
```
codefromthecrypt, line 30:
```suggestion
```

cache_example_test.go
```diff
@@ -7,49 +7,47 @@ import (
...
+	// Creates a runtime configuration to create multiple runtimes.
```
codefromthecrypt, line 31:
```suggestion
```

cache_example_test.go
```diff
@@ -7,49 +7,47 @@ import (
...
+	config := wazero.NewRuntimeConfig().WithCompileCache(cache)
 
 	// Repeat newRuntimeCompileClose with the same cache directory.
-	newRuntimeCompileClose(ctx)
+	newRuntimeCompileClose(ctx, config)
 	// Since the above stored compiled functions to dist, below won't compile.
```
codefromthecrypt, line 36:
```suggestion
	// Since the above stored compiled functions to disk, below won't compile from scratch.
```

cache_example_test.go
```diff
@@ -7,49 +7,47 @@ import (
...
 	if err != nil {
-		log.Panicln(err)
+		log.Fatal(err)
 	}
+	defer cache.Close(ctx)
+
+	// Creates a runtime configuration to create multiple runtimes.
+	config := wazero.NewRuntimeConfig().WithCompileCache(cache)
 
 	// Repeat newRuntimeCompileClose with the same cache directory.
```
codefromthecrypt, line 34:
```suggestion
	// Use the same cache directory for multiple runtimes.
```

cache_test.go
```diff
@@ -0,0 +1,164 @@
...
+	goruntime "runtime"
+	"testing"
+
+	"github.com/tetratelabs/wazero/internal/testing/require"
+)
+
+//go:embed internal/integration_test/vs/testdata/fac.wasm
+var facWasm []byte
+
+func TestCache(t *testing.T) {
```
codefromthecrypt, line 18:
```suggestion
func TestCompileCache(t *testing.T) {
```

cache_test.go
```diff
@@ -0,0 +1,164 @@
...
+		// Also check it is actually cached.
+		require.Equal(t, uint32(1), cacheInst.eng.CompiledModuleCount())
+		barCompiled, err := bar.CompileModule(ctx, facWasm)
+		require.NoError(t, err)
+
+		// Ensures compiled modules are the same.
+		require.Equal(t, compiled, barCompiled)
+
+		// Two runtimes are completely separate except the compilation cache,
+		// therefore it should be ok to instantiate the same name module for each of them.
```
codefromthecrypt, line 37: this is key to getting rid of namespace!

config.go
```diff
@@ -101,6 +102,21 @@ type RuntimeConfig interface {
 	// DWARF "custom sections" that are often stripped, depending on
 	// optimization flags passed to the compiler.
 	WithDebugInfoEnabled(bool) RuntimeConfig
+
+	// WithCompileCache configures how runtime caches the compiled modules. In the default configuration, cached modules are
```
codefromthecrypt, line 106:
```suggestion
	// WithCompileCache configures how runtime caches the compiled modules. In the default configuration, compilation results are
```

config.go
```diff
@@ -101,6 +102,21 @@ type RuntimeConfig interface {
...
+	// only alive until Runtime.Close is closed, and are not shared by multiple Runtime.
```
codefromthecrypt, line 107:
```suggestion
	// only in-memory until Runtime.Close is closed, and not shareable by multiple Runtime.
```

config.go
```diff
@@ -101,6 +102,21 @@ type RuntimeConfig interface {
...
+	//
+	// Below defines the shared cache across multiple instances of Runtime:
+	//
+	//	// Creates the new Cache and the runtime configuration with it.
+	//	cache := wazero.NewCache()
```
codefromthecrypt, line 112:
```suggestion
	//	cache := wazero.NewCompileCache()
```

config.go
```diff
@@ -115,7 +131,8 @@ type runtimeConfig struct {
 	memoryCapacityFromMax bool
 	isInterpreter         bool
 	dwarfDisabled         bool // negative as defaults to enabled
-	newEngine             func(context.Context, api.CoreFeatures) wasm.Engine
+	newEngine             func(context.Context, api.CoreFeatures, compilationcache.Cache) wasm.Engine
```
codefromthecrypt, line 134: s/compilationcache/compilecache/g

internal/engine/compiler/engine.go
```diff
@@ -795,11 +804,11 @@ func (f *function) getSourceOffsetInWasmBinary(pc uint64) uint64 {
...
 }
 
-func NewEngine(ctx context.Context, enabledFeatures api.CoreFeatures) wasm.Engine {
-	return newEngine(ctx, enabledFeatures)
+func NewEngine(ctx context.Context, enabledFeatures api.CoreFeatures, fileCache compilationcache.Cache) wasm.Engine {
+	return newEngine(ctx, enabledFeatures, fileCache)
 }
 
-func newEngine(ctx context.Context, enabledFeatures api.CoreFeatures) *engine {
+func newEngine(ctx context.Context, enabledFeatures api.CoreFeatures, fileCache compilationcache.Cache) *engine {
```
codefromthecrypt, line 811: I think this isn't file, so skip it? of if compilationcache package is only for file, rename it to filecache?

internal/engine/compiler/engine.go
```diff
@@ -808,7 +817,7 @@ func newEngine(ctx context.Context, enabledFeatures api.CoreFeatures) *engine {
 		enabledFeatures: enabledFeatures,
 		codes:           map[wasm.ModuleID][]*code{},
 		setFinalizer:    runtime.SetFinalizer,
-		Cache:           compilationcache.NewFileCache(ctx),
+		Cache:           fileCache,
```
codefromthecrypt, line 820: unless fileCache will only be for files, then rename field to FileCompileCache

```suggestion
		CompileCache:           fileCache,
```

internal/engine/compiler/engine_cache_test.go
```diff
@@ -293,9 +292,7 @@ func TestEngine_getCodesFromCache(t *testing.T) {
 			e := engine{}
 			if tc.ext != nil {
 				tmp := t.TempDir()
-				e.Cache = compilationcache.NewFileCache(
-					context.WithValue(context.Background(), compilationcache.FileCachePathKey{}, tmp),
-				)
+				e.Cache = compilationcache.NewFileCache(tmp)
```
codefromthecrypt, line 295: if this is only for file `compilationcache.NewFileCache` -> `filecache.New` ?

codefromthecrypt (approved):
some editorial notes, and some naming decisions left.

Right now, we are saying CompileCache, like CompileModule.

we should be consistent whatever we use, so I made suggestions to do that, though in hindsight CompileCache sounds like a verb, so maybe rename everything to CompilationCache after all.

So, even though I made comments with CompileCache, you might decide to after applying them, globally replace CompileCache with CompilationCache case preserving!

PR #1018, opened 2023-01-10
Deletes namespace API

mathetake:
Formerly, we introduced `wazero.Namespace` to help avoid module name or import conflicts while still sharing the runtime's compilation cache. Now that we've introduced `CompilationCache` `wazero.Namespace` is no longer necessary. By removing it, we reduce the conceptual load on end users as well internal complexity. Since most users don't use namespace, the change isn't very impactful.

Users who are only trying to avoid module name conflict can generate a name like below instead of using multiple runtimes:

```go
moduleName := fmt.Sprintf("%d", atomic.AddUint64(&m.instanceCounter, 1))
module, err := runtime.InstantiateModule(ctx, compiled, config.WithName(moduleName))
```

For `HostModuleBuilder` users, we no longer take `Namespace` as the last parameter of `Instantiate` method: 

```diff
 	// log to the console.
 	_, err := r.NewHostModuleBuilder("env").
 		NewFunctionBuilder().WithFunc(logString).Export("log").
-		Instantiate(ctx, r)
+		Instantiate(ctx)
 	if err != nil {
 		log.Panicln(err)
 	}
```

The following is an example diff a use of namespace can use to keep compilation cache while also ensuring their modules don't conflict:

```diff

 func useMultipleRuntimes(ctx context.Context, cache) {
-	r := wazero.NewRuntime(ctx)
+	cache := wazero.NewCompilationCache()
 
 	for i := 0; i < N; i++ {
-		// Create a new namespace to instantiate modules into.
-		ns := r.NewNamespace(ctx) // Note: this is closed when the Runtime is
+		r := wazero.NewRuntimeWithConfig(ctx, wazero.NewRuntimeConfig().WithCompilationCache(cache))
 
 		// Instantiate a new "env" module which exports a stateful function.
 		_, err := r.NewHostModuleBuilder("env").
```


codefromthecrypt:
Thanks for starting this! I helped start the description for you.

pointers which you likely already know...

I think to ensure this works and also to get the nice diff for the description, we should make a tentative PR to https://github.com/http-wasm/http-wasm-host-go on this branch prior to merge. 

Probably the namespace example can be converted to a cache one, then all the godocs need to be scrubbed carefully.

Other visible users on GitHub are

* https://github.com/mosn/proxy-wasm-go-host
* https://github.com/cisco-open/nasp
* https://github.com/Pryz/wagi-go
* https://github.com/knqyf263/go-plugin

codefromthecrypt:
added a note to the desc about using generated names instead of switching from namespace to multiple runtimes. This can work for most folks, though the stack traces are less nice.

mathetake:
even mosn/proxy-wasm-go-host can do without multiple runtimes https://github.com/mosn/proxy-wasm-go-host/compare/main...mathetake:proxy-wasm-go-host:makesure?expand=1

mathetake:
I think good to go now

mathetake:
[cisco-open/nasp](https://github.com/cisco-open/nasp) is a fork of mosn, so should be fine as well

mathetake:
Pryz/wagi-go doesn't need namespace at all

mathetake:
go-plugin also doesn't need namespace with the atomic counter and a bit of change in the codegen accordingly.

mathetake:
ah forgot to convert namespace example to multiple runtimes

codefromthecrypt:
> even mosn/proxy-wasm-go-host can do without multiple runtimes 

it is behavior changing, but fine by me. basically we are instantiating host modules that aren't in use.

codefromthecrypt:
I think in mosn, the change you made is expecting that RegisterImports isn't called multiple times, as if it did, it would fail instantiating wasi_unstable.. in any case, one way out would be to defer Runtime creation to later, or to always instantiate wasi_unstable. Regardless, this shows it is missing a test ;)

mathetake:
done with the description and example!

examples/namespace/README.md
```diff
@@ -1,17 +0,0 @@
-## Stateful import example
```
codefromthecrypt, line 1: accidental delete instead of move/replace

imports/wasi_snapshot_preview1/wasi.go
```diff
@@ -45,30 +45,27 @@ func MustInstantiate(ctx context.Context, r wazero.Runtime) {
...
 type Builder interface {
-	// Compile compiles the ModuleName module that can instantiated in any
-	// namespace (wazero.Namespace).
+	// Compile compiles the ModuleName module that can be instantiated in wazero.Runtime.
 	//
 	// Note: This has the same effect as the same function on wazero.HostModuleBuilder.
 	Compile(context.Context) (wazero.CompiledModule, error)
 
-	// Instantiate instantiates the ModuleName module into the given namespace.
+	// Instantiate instantiates the ModuleName module into the given wazero.Runtime.
```
codefromthecrypt, line 65:
```suggestion
	// Instantiate instantiates the ModuleName module and returns a function to close it.
```

imports/wasi_snapshot_preview1/wasi.go
```diff
@@ -45,30 +45,27 @@ func MustInstantiate(ctx context.Context, r wazero.Runtime) {
...
 func Instantiate(ctx context.Context, r wazero.Runtime) (api.Closer, error) {
-	return NewBuilder(r).Instantiate(ctx, r)
+	return NewBuilder(r).Instantiate(ctx)
 }
 
 // Builder configures the ModuleName module for later use via Compile or Instantiate.
 type Builder interface {
-	// Compile compiles the ModuleName module that can instantiated in any
-	// namespace (wazero.Namespace).
+	// Compile compiles the ModuleName module that can be instantiated in wazero.Runtime.
```
codefromthecrypt, line 60:
```suggestion
	// Compile compiles the ModuleName module. Call this before Instantiate.
```

internal/wasm/module.go
```diff
@@ -53,7 +53,7 @@ type Module struct {
 	// ImportSection contains imported functions, tables, memories or globals required for instantiation
 	// (Store.Instantiate).
 	//
-	// Note: there are no unique constraints relating to the two-level namespace of Import.Module and Import.Name.
+	// Note: there are no unique constraints relating to the two-level space of Import.Module and Import.Name.
```
codefromthecrypt, line 56:
```suggestion
	// Note: there are no unique constraints relating to the two-level namespace of Import.Module and Import.Name.
```

internal/wasmdebug/debug.go
```diff
@@ -19,7 +19,7 @@ import (
 //
 //   - moduleName is the possibly empty name the module was instantiated with.
 //   - funcName is the name in the Custom Name section.
-//   - funcIdx is the position in the function index namespace, prefixed with
+//   - funcIdx is the position in the function index space, prefixed with
```
codefromthecrypt, line 22: bonus points for finding where "index space" could be "index" and mean the same thing

codefromthecrypt (approved):
cool beans!

mathetake:
https://github.com/tetratelabs/wazero/pull/1018/commits/70c0880a757761e5bdcd18bf1cf7fdd2f83ece31 revived readme

codefromthecrypt (approved):

PR #1019, opened 2023-01-10
fs: fixes windows fstest.FS and consolidates more

codefromthecrypt:
This works around a known glitch in windows where directory entry stat doesn't match its corresponding file stat (namely times don't). It consolidates more test files, in the process, to ensure we are more likely to trigger issues like this earlier.

Future work will finish the last couple places where we still use go's fstest.MapFS internally, as well introduce stat tests at the syscallfs abstraction: right now, most tests are still only defined in WASI.

### DIFF
--- a/internal/fstest/fstest.go
+++ b/internal/fstest/fstest.go
@@ -4,7 +4,10 @@
 //
 // Here's an example using this inside code that compiles to wasm.
 //
-//	if err := fstest.TestFS(os.DirFS); err != nil {
+//	if err := fstest.WriteTestFiles(tmpDir); err != nil {
+//		log.Panicln(err)
+//	}
+//	if err := fstest.TestFS(os.DirFS(tmpDir)); err != nil {
 //		log.Panicln(err)
 //	}
 //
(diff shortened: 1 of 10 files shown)


internal/fstest/fstest.go
```diff
@@ -62,24 +83,68 @@ func WriteTestFiles(tmpDir string) (err error) {
 			return
 		}
 	}
+
+	// The below is similar to code in os_test.go. In summary, Windows mtime
```
codefromthecrypt, line 87: this was why the tests failed on windows before

internal/syscallfs/syscallfs_test.go
```diff
@@ -62,24 +64,28 @@ func testOpen_Read(t *testing.T, tmpDir string, testFS FS) {
 		requireErrno(t, syscall.ENOENT, err)
 	})
 
-	t.Run("dir exists", func(t *testing.T) {
-		f, err := testFS.OpenFile(dir, os.O_RDONLY, 0)
+	t.Run(". opens root", func(t *testing.T) {
```
codefromthecrypt, line 67: added this as even if fs.FS doesn't like ".", we don't use that abstraction internally

internal/wasi_snapshot_preview1/logging/logging.go
```diff
@@ -138,7 +138,7 @@ func (i logFilestat) Log(_ context.Context, mod api.Module, w logging.Writer, pa
 		w.WriteString(",size=")              //nolint
 		writeI64(w, le.Uint64(buf[32:]))
 		w.WriteString(",mtim=") //nolint
-		writeI64(w, le.Uint64(buf[40:]))
+		writeU64(w, le.Uint64(buf[40:]))
```
codefromthecrypt, line 141: prevent logging showing negative time

mathetake: if the negative is found, doesn't that mean mtim can have the 63-th bit set? that sounds weird

mathetake (approved):

internal/wasi_snapshot_preview1/logging/logging.go
codefromthecrypt: you mean it is easier to see a problem if we leave it I64?

codefromthecrypt: anyway it is better for negative.. as that's for time before 1970. I'm removing it.

mathetake: oh I thought you observed the negative and that's why you changed...

codefromthecrypt: heh I did, but then it was while debugging something ;)

codefromthecrypt:
I'll sort out and comment why on the go 1.17 thing.. can't wait for 1.20 so we can kill that version

PR #1021, opened 2023-01-10
Fixes race on engine creation with Cache API

### LINKED ISSUE #1020: Unsafe assignment when sharing a cache across multiple runtimes
**Describe the bug**
There appears to be a potential race condition here: https://github.com/tetratelabs/wazero/blob/3fc5392570de4715e3c8d844879017efdad7a0b2/runtime.go#L119-L121

Since the eng is lazily instantiated, if multiple goroutines call `NewRuntimeWithConfig` with an uninitialized cache, it appears that multiple engines could be created/assigned and potentially not cleaned up properly.

**To Reproduce**
Haven't written a test to reproduce - just noticed via code inspection.

**Expected behavior**
If a cache shared by multiple runtimes are assumed to share a single engine, it should be assigned with proper locking to ensure no race conditions or memory leaks can occur.

**Screenshots**
n/a

- Go version: 1.19.4
- wazero Version: 3fc5392570de4715e3c8d844879017efdad7a0b2
- Host architecture: arm64
- Runtime mode: compiler

**Additional context**
n/a

mathetake:
Fixes #1020 which has been introduced recently

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -76,6 +76,11 @@ jobs:
           cache: true
 
       - run: make test
+        if: ${{ github.event_name  == 'pull_request' }}
+
+      # Run tests with -race only on main branch push.
+      - run: make test go_test_options='-race'
+        if: ${{ github.event_name  == 'push' }}
 
       - name: "Generate coverage report"  # only once (not per OS)
         if: runner.os == 'Linux'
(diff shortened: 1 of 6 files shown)


codefromthecrypt (approved):
Looks good. one question which probably not an issue, but what if someone uses same cache first with interpreter, then with compiler backend. does it ignore the interpreter?

mathetake:
it does ignore compiler. First wins!

PR #1022, opened 2023-01-11
Separate interpreter and engine per wazero.cache impl

mathetake:

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -79,7 +79,7 @@ jobs:
         if: ${{ github.event_name  == 'pull_request' }}
 
       # Run tests with -race only on main branch push.
-      - run: make test go_test_options='-race'
+      - run: make test go_test_options='-timeout 10m -race'
         if: ${{ github.event_name  == 'push' }}
 
       - name: "Generate coverage report"  # only once (not per OS)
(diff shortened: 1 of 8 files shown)


.github/workflows/commit.yaml
```diff
@@ -79,7 +79,7 @@ jobs:
         if: ${{ github.event_name  == 'pull_request' }}
 
       # Run tests with -race only on main branch push.
-      - run: make test go_test_options='-race'
+      - run: make test go_test_options='-timeout 10m -race'
```
mathetake, line 82: turns out that 2m was not enough for main branch race tests

cache_test.go
```diff
@@ -22,11 +24,16 @@ func TestCompilationCache(t *testing.T) {
 		foo, bar := getCacheSharedRuntimes(ctx, t)
 		cacheInst := foo.cache
 
+		eng := foo.cache.engs[engineKindInterpreter]
```
codefromthecrypt, line 27:
```suggestion
		// add interpreter first, to ensure compiler support isn't order dependent
		eng := foo.cache.engs[engineKindInterpreter]
```

config.go
```diff
@@ -145,6 +145,14 @@ var engineLessConfig = &runtimeConfig{
 	dwarfDisabled:         false,
 }
 
+type engineKind int
+
+const (
+	engineKindCompiler engineKind = iota
+	engineKindInterpreter
+	engineKindNum
```
codefromthecrypt, line 153: supernit: engineKindCount?

codefromthecrypt (approved):
thanks for doing this. sometimes we tell people to temporarily use interpreter so I got nervous before.

mathetake:
🤞  on the timeout stuff on main branch test

Issue #992, opened 2022-12-31
flakey test: gojs Test_stdio_large


codefromthecrypt:
```
--- FAIL: Test_stdio_large (0.50s)
[34](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:35)
    require.go:312: expected "stderr 2097152
[35](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:36)
        ", but was "fatal error: malloc deadlock
[36](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:37)
        
[37](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:38)
        runtime stack:
[38](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:39)
        runtime.throw({0x6b9ce, 0xf})
[39](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:40)
        	/opt/hostedtoolcache/go/1.17.13/x64/src/runtime/panic.go:1198 +0x7
[40](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:41)
        runtime.mallocgc(0x10, 0x42cc0, 0x1)
[41](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:42)
        	/opt/hostedtoolcache/go/1.17.13/x64/src/runtime/malloc.go:972 +0xcf
[42](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:43)
        runtime.newobject(0x42cc0)
[43](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:44)
        	/opt/hostedtoolcache/go/1.17.13/x64/src/runtime/malloc.go:1234 +0x4
[44](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:45)
        runtime.handleEvent()
[45](https://github.com/tetratelabs/wazero/actions/runs/3811321496/jobs/6483930063#step:4:46)
        	/opt/hostedtoolcache/go/1.17.13/x64/src/runtime/lock_js.go:238 +0x2
[...]

PR #1023, opened 2023-01-11
ci: adds `-skip` flag to the race tests

mathetake:


mathetake:
This at least works now modulo the flakiness we already are aware of #992

PR #1026, opened 2023-01-12
fs: Adds fd_sync

ncruces:
This adds the fd_sync syscall:
https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-fd_syncfd-fd---errno

### DIFF
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -1666,7 +1666,7 @@ func Test_fdSeek_Errors(t *testing.T) {
 	}
 }
 
-// Test_fdSync only tests it is stubbed for GrainLang per #271
+// Test_fdSync only tests that the call succeeds; it's hard to test its effectiveness.
 func Test_fdSync(t *testing.T) {
 	tmpDir := t.TempDir() // open before loop to ensure no locking problems.
 	pathName := "test_path"
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -1668,11 +1668,46 @@ func Test_fdSeek_Errors(t *testing.T) {
 
 // Test_fdSync only tests it is stubbed for GrainLang per #271
 func Test_fdSync(t *testing.T) {
-	log := requireErrnoNosys(t, FdSyncName, 0)
-	require.Equal(t, `
---> wasi_snapshot_preview1.fd_sync(fd=0)
-<-- errno=EBADF
-`, log)
+	tmpDir := t.TempDir() // open before loop to ensure no locking problems.
+	pathName := "test_path"
+	mod, fd, log, r := requireOpenFile(t, tmpDir, pathName, []byte{}, false)
+	defer r.Close(testCtx)
+
+	tests := []struct {
+		name          string
+		fd            uint32
+		expectedErrno Errno
+		expectedLog   string
+	}{
+		{
+			name:          "invalid fd",
+			fd:            42, // arbitrary invalid fd
+			expectedErrno: ErrnoBadf,
+			expectedLog: `
+==> wasi_snapshot_preview1.fd_sync(fd=42)
+<== errno=EBADF
+`,
+		},
+		{
+			name:          "valid fd",
+			fd:            fd,
+			expectedErrno: ErrnoSuccess,
+			expectedLog: `
+==> wasi_snapshot_preview1.fd_sync(fd=4)
+<== errno=ESUCCESS
+`,
+		},
+	}
+
+	for _, tt := range tests {
+		tc := tt
+		t.Run(tc.name, func(t *testing.T) {
+			defer log.Reset()
+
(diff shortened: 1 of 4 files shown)


ncruces:
I need this to get `test_demovfs.c` to work on [`github.com/ncruces/go-sqlite3`](https://github.com/ncruces/go-sqlite3).

The intent is to (eventually) replace this VFS by one that sidesteps WASI but, meanwhile, SQLite might be a nice stress test of WASI IO.

Unit tests here are far from ideal. Any ideas?

If you're curious about testing `go-sqlite3` you can do this:
```sh
git clone --branch file https://github.com/ncruces/go-sqlite3
cd go-sqlite3
./build_deps.sh
go run ./cmd
```

The `build_deps.sh` stage downloads and builds the SQLite amalgamation with `zig cc`.
`go run` creates a `test.db` which you can then inspect with the `sqlite3` console.

ncruces:
Tests are failing on Windows, because of (not sure how to fix it):
https://github.com/tetratelabs/wazero/blob/main/internal/syscallfs/syscall_windows.go#L96

ncruces:
Because of #1027, `go-sqlite3` doesn't really succeed if this gets merged.
I need to patch this on top of 859ca2451c12dc62145c52d5e4a7517570bf6cb2, just before bedde6dc7a3dd279f0e19ccd1a6d100570d996bd got merged.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1662,11 +1662,46 @@ func Test_fdSeek_Errors(t *testing.T) {
 
 // Test_fdSync only tests it is stubbed for GrainLang per #271
```
danp, line 1663: Perhaps this comment should be updated/removed?

ncruces: Obviously. I'll update it. Thanks!

codefromthecrypt:
@ncruces thanks for progressing this. I have a windows laptop with me and so can help on this PR once I finish the pre-open thing I'm working on. Fantastic use case, btw!

codefromthecrypt:
actually the pre-opens thing will take longer. will look at this now

codefromthecrypt:
TL;DR; I think you can for now make the test skip on windows with a TODO, and I can handle it later.

e.g. `if runtime.GOOS == "windows { t.Skip("TODO: windows currently fails on EBADFD") }`

I can't immediately think of a semantic test we can do for linux/darwin etc, besides just making sure it doesn't error.

---

My laptop is stuck in windows update and it is really slow, so I'll mention what I think so far.

As far as I can tell, go doesn't test `File.Sync()` or Fsync for platforms that end up calling it. Hence, I'm not sure how much we can do about this. What's odd is os.File is defined in file_posix.go and the latter is defined for windows. So, I think the Sync method exists, and the EBADFD is for another reason maybe.

Not directly related, but seems we have another mapping thing to do, if we start testing O_SYNC (or attempting to) https://github.com/golang/go/issues/35358

ncruces:
> As far as I can tell, go doesn't test `File.Sync()` or Fsync for platforms that end up calling it. Hence, I'm not sure how much we can do about this. What's odd is os.File is defined in file_posix.go and the latter is defined for windows. So, I think the Sync method exists, and the EBADFD is for another reason maybe.

Yes, on Windows `os.File.Sync()` exists (it exists on all platforms). The thing is, you're wrapping the file on that [anonymous struct](https://github.com/tetratelabs/wazero/blob/ecac3f529b8db1dae13a95933a0aa15b0ccdf530/internal/syscallfs/syscall_windows.go#L96) that only implements the [read](https://github.com/tetratelabs/wazero/blob/ecac3f529b8db1dae13a95933a0aa15b0ccdf530/internal/syscallfs/syscallfs.go#L134) and [write](https://github.com/tetratelabs/wazero/blob/ecac3f529b8db1dae13a95933a0aa15b0ccdf530/internal/syscallfs/syscall_windows.go#L109) side of it.

If no error translation is necessary for `Seek`, I think this is enough (but have no way to test it other than make the change and get your CI to do it):
```go
func maybeWrapFile(f file) file {
	return struct {
		readFile
		io.Writer
		syncer
	}{f, &windowsWriter{f}, f}
}

type syncer interface{ Sync() error }
```

ncruces:
I had to rebase on head to try to fix Windows, which means I'm now broken for #1027.

codefromthecrypt:
Good eye! 

Notes not about this, but more FYI: we will hopefully by the end of the month no longer be using fs.FS mask wrappers and instead our own file type. There's still some construction work left to do to pass multiple pre-opens to WASI which I'll be doing on a flight tomorrow. I will download #1027 to try to make sure at the same time, it fixes that.

codefromthecrypt (approved):
Thanks again for going beyond the call of duty. Ack there are a lot of broken eggs lately while we try to get test infra in place to guarantee wasi works generally, tested with tinygo, zig and wasi-testsuite. We're not there, yet and really appreciate your brain and hands meanwhile!

PR #1028, opened 2023-01-12
examples/basic: parse positional arguments with flags

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

Use `flag` to parse positional arguments (avoiding `os.Args` out of bounds panics) and return friendly errors:

```
$ go run .
2023/01/12 16:38:57 failed to read arguments: must specify two command line arguments
exit status 1

$ go run . 1 foo
2023/01/12 16:39:00 failed to read arguments: strconv.ParseUint: parsing "foo": invalid syntax
exit status 1

$ go run . 1 2
1 + 2 = 3
```

Edit: switched to log.Panicf per feedback below.

### DIFF
--- a/examples/basic/add.go
+++ b/examples/basic/add.go
@@ -3,9 +3,10 @@ package main
 import (
 	"context"
 	_ "embed"
+	"errors"
+	"flag"
 	"fmt"
 	"log"
-	"os"
 	"strconv"
 
 	"github.com/tetratelabs/wazero"
@@ -25,6 +26,9 @@ var addWasm []byte
 // Since addWasm was compiled with TinyGo's `wasi` target, we need to configure
 // WASI host imports.
 func main() {
+	// Parse positional arguments.
+	flag.Parse()
+
 	// Choose the context to use for function calls.
 	ctx := context.Background()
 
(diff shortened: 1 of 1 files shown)


examples/basic/add.go
```diff
@@ -44,27 +48,35 @@ func main() {
 	}
 
 	// Read two args to add.
-	x, y := readTwoArgs()
+	x, y, err := readTwoArgs(flag.Arg(0), flag.Arg(1))
+	if err != nil {
+		log.Fatalf("failed to read arguments: %v", err)
```
codefromthecrypt, line 53: we've been asked to put Panicln in the examples, as while they are in the form of a CLI, they are tinkering areas. panics show you the line that failed.

codefromthecrypt (commented):
using flags for the positional arguments sounds great!

Totally agree that if these were real CLIs (such as cmd/wazero), more attention should be paid to error messages to avoid panic. However, these are optimizing for people tinkering. We used to not use panics, and we had feedback to add them because they show which line failed. Hope you understand that while your opinion is correct and valid, in the context here it is grey area and that's the rationale.

codefromthecrypt:
so basically, the tl;dr; is yep lets make sure none of our examples panic for the wrong reason. (e.g. positional arg problem).

If possible, I would prefer to not have to go back and re-negotiate panicln when there is a runtime error as it sounded reasonable when we were first asked for that.

codefromthecrypt:
ps if you do feel strongly about this, I'll back you up, just please do this to all the examples and make sure there's a fmt.Errorf prefix that can help identify the line that failed. our example are tested in CI, so it helps to make sure we know roughly where in the source might have died and while panicln does that, a message can too!

examples/basic/add.go
mdlayher: Not a problem, I am happy to restore the panic. I just thought that running an example binary with no arguments and getting a panic seemed a little unfriendly.

Now that there are checks for empty positional arguments, it's more clear where things went wrong.

mdlayher:
@codefromthecrypt updated, ptal!

codefromthecrypt (approved):
lovely. thanks for fixing the broken windows!

PR #1033, opened 2023-01-13
ci: skip cargo install cargo-fuzz if cache hit

mathetake:
The `cargo-fuzz` binary is included in the cache, so no need to re-execute `cargo install`.
This avoids nasty long timeouts related to the cargo index regardless of the actual installation. 


mathetake:
working now

codefromthecrypt (approved):
thanks!

PR #1037, opened 2023-01-15
fs: extracts syscallfs.ReaderAtOffset for WASI and gojs

codefromthecrypt:
This extracts a utility `syscallfs.ReaderAtOffset()` to allow WASI and gojs to re-use the same logic to implement `syscall.Pread`.

What's different than before is that if WASI passes multiple iovecs an emulated `ReaderAt` will seek to the read position on each call to `Read` vs once per loop. This was a design decision to keep the call sites compatible between files that implement ReaderAt and those that emulate them with Seeker (e.g. avoid the need for a read-scoped closer/ defer function). The main use case for emulation is `embed.file`, whose seek function is cheap, so there's little performance impact to this.

```
goos: darwin
goarch: amd64
pkg: github.com/tetratelabs/wazero/internal/syscallfs
BenchmarkReaderAtOffset
BenchmarkReaderAtOffset/os.DirFS_io.File
BenchmarkReaderAtOffset/os.DirFS_io.File-16         	  549439	      2196 ns/op
BenchmarkReaderAtOffset/os.DirFS_readerAtOffset
BenchmarkReaderAtOffset/os.DirFS_readerAtOffset-16  	  612918	      1985 ns/op
BenchmarkReaderAtOffset/embed.FS_embed.file
BenchmarkReaderAtOffset/embed.FS_embed.file-16      	10059037	       126.1 ns/op
BenchmarkReaderAtOffset/embed.FS_seekToOffsetReader
BenchmarkReaderAtOffset/embed.FS_seekToOffsetReader-16         	 7055109	       179.0 ns/op
```


codefromthecrypt:
cc @ncruces as this moves around the stuff you made, and thanks for that!

ncruces:
I'll try to have a stab at convincing the Go folks to add `ReaderAt` to `embed`. There's no good reason to for it not too. But that's always at least a year off.

codefromthecrypt:
> I'll try to have a stab at convincing the Go folks to add ReaderAt to embed. There's no good reason to for it not too. But that's always at least a year off.

Thanks, I scratched my head looking at it, too, as it seems as easy as mapfs. You are right though, there is a socialization, review and landing concern. Thanks for offering to take that on which is much more effort than the code and tests!

internal/syscallfs/syscallfs_test.go
```diff
@@ -257,3 +260,100 @@ func (f *testFSAdapter) Open(name string) (fs.File, error) {
 func requireErrno(t *testing.T, expected syscall.Errno, actual error) {
 	require.True(t, errors.Is(actual, expected), "expected %v, but was %v", expected, actual)
 }
+
+var (
+	//go:embed testdata/wazero.txt
+	readerAtFS   embed.FS
+	readerAtFile = "wazero.txt"
```
mathetake, line 267: maybe better add empty file as well so that we won't crash when reading on the empty file?

mathetake (approved):

internal/syscallfs/syscallfs_test.go
codefromthecrypt: sure!

ncruces:
Got it merged, should be in 1.21:
https://github.com/golang/go/issues/57803#event-8989467802

PR #1042, opened 2023-01-17
fs: adds string for better error experience

codefromthecrypt:
This prepares for pseudo-root when the CLI doesn't provide one by improving the error messages in general, as well being consistent about parameter order.

### DIFF
--- a/internal/syscallfs/readfs.go
+++ b/internal/syscallfs/readfs.go
@@ -15,11 +15,20 @@ import (
 func NewReadFS(fs FS) FS {
 	if _, ok := fs.(*readFS); ok {
 		return fs
+	} else if _, ok = fs.(*adapter); ok {
+		return fs // fs.FS is always read-only
+	} else if _, ok = fs.(empty); ok {
+		return fs // empty is always read-only
 	}
 	return &readFS{fs}
 }
 
 type readFS struct{ fs FS }
+
+// String implements fmt.Stringer
+func (r *readFS) String() string {
+	return r.fs.String() + ":ro"
+}
 
 // Open implements the same method as documented on fs.FS
 func (r *readFS) Open(name string) (fs.File, error) {
@@ -28,7 +37,7 @@ func (r *readFS) Open(name string) (fs.File, error) {
 
 // GuestDir implements FS.GuestDir
 func (r *readFS) GuestDir() string {
-	return "/"
+	return r.fs.GuestDir()
 }
 
 // OpenFile implements FS.OpenFile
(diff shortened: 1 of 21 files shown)


internal/syscallfs/readfs.go
```diff
@@ -15,20 +15,29 @@ import (
...
+
 // Open implements the same method as documented on fs.FS
 func (r *readFS) Open(name string) (fs.File, error) {
 	panic(fmt.Errorf("unexpected to call fs.FS.Open(%s)", name))
 }
 
 // GuestDir implements FS.GuestDir
 func (r *readFS) GuestDir() string {
-	return "/"
+	return r.fs.GuestDir()
```
codefromthecrypt, line 40: this was a bug

internal/syscallfs/readfs_test.go
```diff
@@ -11,22 +11,50 @@ import (
 	"github.com/tetratelabs/wazero/internal/testing/require"
 )
 
-func TestReadFS_MkDir(t *testing.T) {
+func TestNewReadFS(t *testing.T) {
 	tmpDir := t.TempDir()
-	testFS := NewReadFS(Adapt("/", hackFS(tmpDir)))
 
-	err := testFS.Mkdir("mkdir", fs.ModeDir)
+	// Doesn't double-wrap file systems that are already read-only
```
codefromthecrypt, line 17: this is better than before where we accidentally wrapped things that didn't need it

internal/syscallfs/readfs_test.go
```diff
@@ -11,22 +11,50 @@ import (
...
+	require.Equal(t, writeable.GuestDir(), readFS.GuestDir())
+}
+
+func TestReadFS_String(t *testing.T) {
+	writeable, err := NewDirFS(".", "/tmp")
+	require.NoError(t, err)
+
+	readFS := NewReadFS(writeable)
+	require.NotEqual(t, writeable, readFS)
+	require.Equal(t, "./:/tmp:ro", readFS.String())
```
codefromthecrypt, line 36: This is a lot better in error messages as it matches our CLI syntax (like docker volume)

internal/syscallfs/rootfs.go
```diff
@@ -54,7 +53,10 @@ func NewRootFS(fs ...FS) (FS, error) {
 
 	// Ensure there is always a root match to keep runtime logic simpler.
 	if ret.rootIndex == -1 {
-		return nil, errors.New("you must supply a root filesystem")
+		// TODO: Make a fake root filesystem that can do a directory listing of
```
codefromthecrypt, line 56: note about what's actually a fair amount of work to implement

codefromthecrypt (commented):
notes

PR #1045, opened 2023-01-17
Updates Spectest to the latest (Dec 16, 2022)

mathetake:


mathetake:
lol we magically hit the implementation diff on riscv!

.github/workflows/commit.yaml
```diff
@@ -28,7 +28,8 @@ concurrency:
 jobs:
   check:
     name: Pre-commit check
-    runs-on: ubuntu-20.04
+    # wabt requires the latest libc.
+    runs-on: ubuntu-latest
```
codefromthecrypt, line 32: maybe 

```
    # wabt requires a later version of libc than what's installed on ubuntu-20.04
     runs-on: ubuntu-22.04
```

codefromthecrypt (approved):
only one nit. thanks!

PR #1046, opened 2023-01-18
fs: adds base UnimplementedFS type and unwraps PathError

codefromthecrypt:
This reduces some boilerplate by extracting UnimplementedFS from the existing FS implementations, such that it returns ENOSYS. This also removes inconsistency where some methods on FS returned syscall.Errno and others PathError.

Note: this doesn't get rid of all PathError, yet. We still need to create a syscallfs.File type which would be able to do that. This is just one preliminary cleanup before refactoring out the `fs.FS` embedding from `syscallfs.FS`.

P.S. naming convention is arbitrary, so I took UnimplementedXXX from grpc. This pattern is used a lot of places, also proxy-wasm-go-sdk, e.g. `DefaultVMContext`.


internal/syscallfs/unsupported.go
```diff
@@ -0,0 +1,56 @@
...
+	return "Unimplemented:/"
+}
+
+// Open implements the same method as documented on fs.FS
+func (UnimplementedFS) Open(name string) (fs.File, error) {
+	panic(fmt.Errorf("unexpected to call fs.FS.Open(%s)", name))
+}
+
+// GuestDir implements FS.GuestDir
+func (UnimplementedFS) GuestDir() string {
```
codefromthecrypt, line 24: ps another change I'm going to take out this entirely. basically config will have the ability to map filesystems, but the fs type itself doesn't need to carry this around.

PR #1048, opened 2023-01-18
CompiledModule: export Custom Sections

evacchi:

Add support to Custom Sections to the public API. 

Please do let me know if I put everything in the right places and especially if something is missing, I am justing getting started learning my way through the codebase :)

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -562,6 +562,14 @@ type Memory interface {
 	WriteString(offset uint32, v string) bool
 }
 
+// CustomSection contains the name and raw data of a custom section.
+type CustomSection interface {
+	// Name is the name of the custom section
+	Name() string
+	// Data is the raw data of the custom section
+	Data() []byte
+}
+
 // EncodeExternref encodes the input as a ValueTypeExternref.
 //
 // See DecodeExternref
--- a/config.go
+++ b/config.go
@@ -117,6 +117,16 @@ type RuntimeConfig interface {
 	//	foo := wazero.NewRuntimeWithConfig(context.Background(), config)
 	// 	bar := wazero.NewRuntimeWithConfig(context.Background(), config)
 	WithCompilationCache(CompilationCache) RuntimeConfig
+
+	// WithCustomSections toggles parsing of "custom sections". Defaults to false.
+	//
+	// When enabled, it is possible to retrieve custom sections from a CompiledModule:
+	//
+	//	config := wazero.NewRuntimeConfig().WithCustomSections(true)
+	//	r := wazero.NewRuntimeWithConfig(ctx, config)
+	//	c, err := r.CompileModule(ctx, wasm)
+	//	customSections := c.CustomSections()
+	WithCustomSections(bool) RuntimeConfig
 }
 
 // NewRuntimeConfig returns a RuntimeConfig using the compiler if it is supported in this environment,
@@ -135,6 +145,7 @@ type runtimeConfig struct {
 	dwarfDisabled         bool // negative as defaults to enabled
 	newEngine             newEngine
 	cache                 CompilationCache
+	storeCustomSections   bool
 }
 
 // engineLessConfig helps avoid copy/pasting the wrong defaults.
@@ -227,6 +238,13 @@ func (c *runtimeConfig) WithDebugInfoEnabled(dwarfEnabled bool) RuntimeConfig {
 	return ret
 }
 
+// WithCustomSections implements RuntimeConfig.WithCustomSections
+func (c *runtimeConfig) WithCustomSections(storeCustomSections bool) RuntimeConfig {
+	ret := c.clone()
+	ret.storeCustomSections = storeCustomSections
+	return ret
+}
+
 // CompiledModule is a WebAssembly module ready to be instantiated (Runtime.InstantiateModule) as an api.Module.
 //
 // In WebAssembly terminology, this is a decoded, validated, and possibly also compiled module. wazero avoids using
@@ -264,6 +282,10 @@ type CompiledModule interface {
 	// Note: As of WebAssembly Core Specification 2.0, there can be at most one
 	// memory.
 	ExportedMemories() map[string]api.MemoryDefinition
+
+	// CustomSections returns all the custom sections
+	// (api.CustomSection) in this module keyed on the section name.
+	CustomSections() []api.CustomSection
 
 	// Close releases all the allocated resources for this CompiledModule.
 	//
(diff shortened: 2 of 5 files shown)


jordan-rash:
Thanks @evacchi

config_example_test.go
```diff
@@ -0,0 +1,43 @@
...
+	//
+}
+
+func mustContain(ss []api.CustomSection, name string) {
+	for _, s := range ss {
+		if s.Name() == name {
+			return
+		}
+	}
+	log.Panicln("Could not find section named", name)
```
mathetake, line 42:
```suggestion
	log.Panicf("Could not find section named %s\n", name)
```

config_test.go
```diff
@@ -597,6 +606,53 @@ func Test_compiledModule_Name(t *testing.T) {
...
+	}{
+		{
+			name:     "no custom section",
+			input:    &compiledModule{module: &wasm.Module{}},
+			expected: nil,
+		},
+		{
+			name: "name",
+			input: &compiledModule{module: &wasm.Module{
+				CustomSections: []*wasm.CustomSection{
```
mathetake, line 623: [`name`](https://webassembly.github.io/spec/core/appendix/custom.html#name-section) and [`dwarf_**` (DWARF)](https://yurydelendik.github.io/webassembly-dwarf/) custom sections are handled differently and not stored in wasm.Module.CustomSections, though not sure the users want to access them directly. are we ok with it?

api/wasm.go
```diff
@@ -562,6 +562,12 @@ type Memory interface {
 	WriteString(offset uint32, v string) bool
 }
 
+// CustomSection contains the name and raw data of a custom section.
+type CustomSection interface {
+	Name() string
+	Data() []byte
```
mathetake, line 568: In Go in general, every single one of good public API *must* have Godoc! https://go.dev/doc/comment
```suggestion
	// Name ...
	Name() string
	// Data ...
	Data() []byte
```

config.go
```diff
@@ -117,6 +117,16 @@ type RuntimeConfig interface {
...
 	WithCompilationCache(CompilationCache) RuntimeConfig
+
+	// WithCustomSections toggles parsing of "custom sections". Defaults to false.
+	//
+	// When enabled, it is possible to retrieve custom sections from a CompiledModule:
+	//
+	//	config := wazero.NewRuntimeConfig().WithCustomSections(true)
+	//	r := wazero.NewRuntimeWithConfig(context.Background(), config)
+	//  c, err := r.CompileModule(ctx, wasm)
+	//  customSections := c.CustomSections()
```
mathetake, line 128: wrong indentation? 😉

config.go
```diff
@@ -117,6 +117,16 @@ type RuntimeConfig interface {
 	//	foo := wazero.NewRuntimeWithConfig(context.Background(), config)
 	// 	bar := wazero.NewRuntimeWithConfig(context.Background(), config)
 	WithCompilationCache(CompilationCache) RuntimeConfig
+
+	// WithCustomSections toggles parsing of "custom sections". Defaults to false.
+	//
+	// When enabled, it is possible to retrieve custom sections from a CompiledModule:
+	//
+	//	config := wazero.NewRuntimeConfig().WithCustomSections(true)
+	//	r := wazero.NewRuntimeWithConfig(context.Background(), config)
```
mathetake, line 126:
```suggestion
	//	r := wazero.NewRuntimeWithConfig(ctx, config)
```

config.go
```diff
@@ -318,6 +340,29 @@ func (c *compiledModule) ExportedMemories() map[string]api.MemoryDefinition {
...
+
+type customSection struct {
+	name string
+	data []byte
+}
+
+func (c *customSection) Name() string {
+	return c.name
+}
+
```
mathetake, line 361: // Data implements...

config.go
```diff
@@ -318,6 +340,29 @@ func (c *compiledModule) ExportedMemories() map[string]api.MemoryDefinition {
...
+	}
+	return ret
+}
+
+type customSection struct {
+	name string
+	data []byte
+}
+
+func (c *customSection) Name() string {
```
mathetake, line 358: // Name implements...

mathetake (commented):
Good work!

config_test.go
mathetake: cc @codefromthecrypt I think we are fine as long as document this behavior on the API like "DWARF and name sections are not retrieved" since I think no one wants them to access, though this seems weird asymmetry in the behavior.

codefromthecrypt: I feel like we can skip these sections and wait for someone to ask for them. if that happens we'd retain the raw bytes of them

config.go
```diff
@@ -318,6 +340,29 @@ func (c *compiledModule) ExportedMemories() map[string]api.MemoryDefinition {
 	return c.module.ExportedMemories()
 }
 
+// CustomSections implements CompiledModule.CustomSections
+func (c *compiledModule) CustomSections() []api.CustomSection {
+	var ret []api.CustomSection
+	for _, d := range c.module.CustomSections {
+		dd := &customSection{data: d.Data, name: d.Name}
+		ret = append(ret, dd)
```
mathetake, line 348: Let's use `make` to pre-reserve the memory so that we can avoid unnecessary memory allocation in `append` inside the loop. (whenever possible doing it is the norm in Go! 🕵️)

```suggestion
	ret := make([]api.CustomSection, len(c.module.CustomSections))
	for i, d := range c.module.CustomSections {
		ret[i] = &customSection{data: d.Data, name: d.Name}
```

config.go
```diff
@@ -318,6 +340,29 @@ func (c *compiledModule) ExportedMemories() map[string]api.MemoryDefinition {
...
+	}
+	return ret
+}
+
+type customSection struct {
```
mathetake, line 353:
```suggestion
// customSection implements ...
type customSection struct {
```

config_test.go
mathetake: cool good

evacchi: resolving

config.go
evacchi: I think I followed the style of the doc strings above it? the example is indented 1 tab

config.go
evacchi: BTW is it fine to implement it here or should it go some place else?

config.go
evacchi: FYI: this slightly alters the semantics, it will no longer return nil but an empty slice (which is fine, just FYI because I had to update `Test_compiledModule_CustomSections`!)

config.go
```diff
@@ -117,6 +117,16 @@ type RuntimeConfig interface {
...
 	WithCompilationCache(CompilationCache) RuntimeConfig
+
+	// WithCustomSections toggles parsing of "custom sections". Defaults to false.
+	//
+	// When enabled, it is possible to retrieve custom sections from a CompiledModule:
+	//
+	//	config := wazero.NewRuntimeConfig().WithCustomSections(true)
+	//	r := wazero.NewRuntimeWithConfig(ctx, config)
+	//  c, err := r.CompileModule(ctx, wasm)
+	//  customSections := c.CustomSections()
```
mathetake, line 128: @evacchi you closely compare these lines, you mixed spaces and tabs, that's what I meant the previous comment.

config.go
mathetake: i think for now it's fine to implement here (in any way we can change it later when necessary because it's not exported!)

mathetake (approved):
Great!

config.go
evacchi: oh my! :D

evacchi: fix pushed!

Issue #777, opened 2022-08-29
Will new wazeroir / optimized compilers be supported


hafeidejiangyou:
Now, wazeroir is a format similar to LightbeamIR, it is used for singlepass compiler. Does there a plan to have an IR for deeper optimization with higher execution performance？

mathetake:
Hi @hafeidejiangyou, that's a good q! When we first started designing our compiler (which used to be called JIT), I even had no idea whether it was possible to develop the dynamic compilation and execution engine purely in Go. So I didn't spare time to have the optimized infrastructure for the compilation but rather chose to develop the "baseline" compiler as fast as possible (given that CGO itself is slow).

Now that we are approaching 1.0 (#506), I plan to develop more time on the optimization later on. That of course comes with the redesign of wazeroir (sure it will be SSA style one!) and the introduction of a more optimized register allocator (vs the one we have now is a naive local register allocator). But it all depends on how we prioritize -- it's only me who works on the compiler backend in the wazero project, and therefore, our cycles are limited.

Anyway, stay tuned!

hafeidejiangyou:
Thanks for your reply, I'll keep an eye on it. By the way, I am interested in contributing in the future.

mathetake:
https://github.com/tetratelabs/wazero/pull/1049

codefromthecrypt:
Here's an update.

### We haven't had time to begin this

Until now (mid-2023), we've decided to not displace resources on an optimizing compiler. This has been to support wazero's 1.0 release, and also ensure we have people to champion the success of various things going on in the emerging `GOOS=wasip1`, as well support end users. That said, the existing codebase has been cleaned significantly and the existing non-optimizing compiler is more efficient than before.

### We aren't likely to open the internals for outside compiler implementations

It is technically possible to embed a compiler like cranelift, by compiling it down to wasm and then wazero embedding that and running it internally. However, it is in direct conflict with wazero's goal of being zero dependency. The resulting wasm would be far larger than wazero itself, and an alternative of bootstrapping it somehow makes the runtime fragile on that point and above all more complicated. Further, no one has experience contributing to crane lift and it would end up affecting how we maintain wazero. Meanwhile we would still need our existing compiler backend for the same reasons, so we would have more than twice the problems in reality as now.

### We can do an optimizing compiler based on prior art, but not now

[...]

PR #1049, opened 2023-01-19
Initial implementation of the new Cranelift backend

mathetake:
## Background

TODO

## Architecture

TODO

## Scope of this PR

Basically this PR ensures the base for the further development of this new backend, and is trying to
enable us to incrementally work on it by multiple people simultaneously.

- [x] Buildable (for all GOARCH target)
- [x] Automated test of Cranelift backend (`cargo test` in CI)
- [x] Automated tests (CI)
- [ ] Initial Arm64 implementation (Darwin and Linux)
    - [x] Calling convention bridge between Go and Cranelift-generated functions
    - [x] Locally-defined function calls
    - [x] Memory Access
    - [x] Imported function calls
    - [ ] Host Function calls
    - [x] Imported Memory Access

## Out of scope 
- Go->Wasm entry with many parameters (> parameter registers in the arm64 calling convention).
- Tables, element related instructions.
- Spec V2 features (SIMD, reference types etc).
- Trap handling
    - we cannot rely on the signal handler (as done by wasmtime), and therefore we have to translate it to the "exit function call" with the specific status. 
- Stack guard implementation
    - Eventually, we need to insert stack guard preamble at every function, and grow the stack in the Go world when necessary (that is what the our compiler backend does).
- Entire amd64 implementation


hafeidejiangyou:
Hi, I am glad to see this progress. As mentioned in #777 , I'd be happy to contribute this.
I have some quesions with this:
1. Although I know this is a good choose, but I still want to ask why use cranelift.
2. Compile the cranelift compiler as a wasm lib is a greate idea! But I don't know whether there have some problem in the compile with Rust code to wasm, like system call in the dep crate. And Is this way the final solution with Cranelift?

PR #1051, opened 2023-01-20
docs: add zig section 

evacchi:
Docs for the Zig section of the website. I am new to Zig so please do add your notes :)

### DIFF
--- a/examples/allocation/zig/README.md
+++ b/examples/allocation/zig/README.md
@@ -22,8 +22,3 @@ This example uses `@panic()` rather than `unreachable` to handle errors
 since `unreachable` emits a call to panic only in `Debug` and `ReleaseSafe`
 mode. In `ReleaseFast` and `ReleaseSmall` mode, it would lead into undefined
 behavior.
-
-If building wasm with a pre-release version of Zig 0.10.0, use `-fstage1` to
-avoid [bugs in the new compiler][1].
-
-[1]: https://github.com/ziglang/zig/wiki/Self-Hosted-Compiler-Upgrade-Guide#is-it-time-to-upgrade
--- a/site/content/languages/zig.md
+++ b/site/content/languages/zig.md
@@ -1,0 +1,196 @@
++++
+title = "Zig"
++++
+
+## Introduction
+
+Beginning with 0.4.0 [Zig][1] can generate `%.wasm` files instead of
+architecture-specific binaries through three targets:
+
+* `wasm32-emscripten`: mostly for browser (JavaScript) use.
+* `wasm32-freestanding`: for standalone use in or outside the browser.
+* `wasm32-wasi`: for use outside the browser.
+
+This document is maintained by wazero, which is a WebAssembly runtime that
+embeds in Go applications. Hence, our notes focus on targets used outside the
+browser, tested by wazero: `wasm32-freestanding` and `wasm32-wasi`.
+
+## Overview
+
+When Zig compiles a `%.zig` file with a `wasm32-*` target, the output `%.wasm`
+depends on a subset of features in the [WebAssembly 2.0 
+Core specification]({{< ref "/specs#core" >}}) and [WASI]({{< ref "/specs#wasi" >}}) host
+functions.
+
+Unlike some compilers, Zig also supports importing custom host functions and
+exporting functions back to the host.
+
+Here's a basic example of source in Zig:
+
+```zig
+export fn add(a: i32, b: i32) i32 {
+    return a + b;
+}
+```
+
+The following is the minimal command to build a Wasm file.
+
+```bash
+zig build-lib -dynamic -target wasm32-freestanding main.zig
+```
+
+The resulting Wasm `export`s the `add` function so that the embedding host can
+call it, regardless of if the host is written in Zig or not.
+
+Notice we are using `zig build-lib -dynamic`: this
+compiles the source as a library, i.e. without a `main` function.
+
+## Disclaimer
+
+This document includes notes contributed by the wazero community for Zig 0.10.1. 
+While wazero includes Zig examples, and maintainers contribute to Zig, this
+isn't a Zig official document. For more help, consider the [WebAssembly Documentation][4] 
+or joining the [#programming-discussion channel on 
+Zig's Discord][5]. 
+
+Meanwhile, please help us [maintain][6] this document and [star our GitHub
+repository][7], if it is helpful. Together, we can make WebAssembly easier on
+the next person.
+
+## Constraints
+
+Please read our overview of WebAssembly and
+[constraints]({{< ref "_index.md#constraints" >}}). In short, expect
+limitations in both language features and library choices when developing your
+software.
+
+## Memory
+
+The Zig language performs no memory management on behalf of the programmer. 
+However, Zig has no default allocator. Instead, functions which need to allocate 
+accept an `Allocator` parameter.
+
+### Host Allocations
+
+Sometimes a host function needs to allocate memory directly. For example, to write JSON
+of a given length before invoking an exported function to parse it.
+
+```zig
+pub export fn configure(ptr: [*]const u8, size: u32) void {
+    _configure(message[0..size]) catch |err| @panic(switch (err) {
+        error.OutOfMemory => "out of memory",
+    });
+}
+```
+
+The general flow is that the host allocates memory by calling an allocation
+function with the size needed. Then, it writes data, in this case JSON, to the
+memory offset (`ptr`). At that point, it can call a host function, ex
+`configure`, passing the `ptr` and `size` allocated. The guest Wasm (compiled
+from Zig) will be able to read the data. To ensure no memory leaks, the host
+calls a free function, with the same `ptr`, afterwards and unconditionally.
+
+Note: wazero includes an [example project][9] that shows this.
+
+The [zig example][9] does a few things of interest:
+* Uses `@ptrToInt` to change a Zig pointer to a numeric type
+* Uses `[*]u8` as an argument to take a pointer and slices it to build back a
+string
+* It also shows how to import a host function using the `extern` directive
+
+To allow the host to allocate memory, you need to define your own `malloc` and
+`free` functions:
+```webassembly
+(func (export "malloc") (param $size i32) (result (;$ptr;) i32))
+(func (export "free") (param $ptr i32) (param $size i32))
+```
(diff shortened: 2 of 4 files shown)


examples/allocation/zig/README.md
```diff
@@ -22,8 +22,3 @@ This example uses `@panic()` rather than `unreachable` to handle errors
 since `unreachable` emits a call to panic only in `Debug` and `ReleaseSafe`
 mode. In `ReleaseFast` and `ReleaseSmall` mode, it would lead into undefined
 behavior.
-
-If building wasm with a pre-release version of Zig 0.10.0, use `-fstage1` to
-avoid [bugs in the new compiler][1].
-
-[1]: https://github.com/ziglang/zig/wiki/Self-Hosted-Compiler-Upgrade-Guide#is-it-time-to-upgrade
```
evacchi, line 29: since the bootstrapped compiler is now the default and 0.10.1 has been released I am assuming this note is redundant now?

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+Notice we are using `zig build-lib -dynamic`: this
+compiles the source as a library, i.e. without a `main` function.
+
+## Disclaimer
+
+This document includes notes contributed by the wazero community. While wazero
+includes Zig examples, and maintainers contribute to Zig, this
+isn't a Zig official document. For more help, consider the [Zig
+WebAssembly Guide](#???) or joining the [#programming-discussion channel on 
+Zig's Discord][5].
```
evacchi, line 54: `#???` obviously calls for attention. Is there a proper guide we can link here? Is it ok to link the Discord invite here?

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+### Binary size
+
+Those with `%.wasm` binary size constraints can change their source, 
+e.g. picking a [different allocator][9b] or set `zig` flags to reduce it.
+
+Source changes:
+* [WebAssembly-only fast allocator][12] is a specialized allocator that 
+trades generality for simplicity, making however specific assumptions 
+on the runtime environment (e.g. single-threaded access to global variables, 
+it direcly calls Wasm-specific intrinsics and it never releases memory)
```
evacchi, line 175: I see this is still on `main`, so I am assuming this has not been folded back into the GPA yet (although the source says "//! This is intended to be merged into GeneralPurposeAllocator at some point.")

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+### Performance
+
+Those with runtime performance constraints can change their source or set
+`zig` flags to improve it.
+
+[`zig` flags][13]:
+* `-OReleaseFast`: Enable additional optimizations, possibly at the cost of 
+  increased binary size.
+
+## Frequently Asked Questions
```
evacchi, line 196: any other FAQs worth mentioning here?

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+```zig
+export fn add(a: i32, b: i32) i32 {
+    return a + b;
+}
+```
+
+The following is the minimal command to build a Wasm file.
+
+```bash
+zig build-lib main.zig -target wasm32-freestanding -dynamic
```
mathetake, line 39:
```suggestion
zig build-lib -dynamic main.zig -target wasm32-freestanding
```

let's group "build-lib -dynamic" together so that this lines with the below.

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+The resulting Wasm `export`s the `add` function so that the embedding host can
+call it, regardless of if the host is written in Zig or not.
+
+Notice we are using `zig build-lib -dynamic`: this
+compiles the source as a library, i.e. without a `main` function.
+
+## Disclaimer
```
mathetake, line 48: we should add the used version of Zig here since Zig changes extremely rapidly and can be easily broken

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+
+Please read our overview of WebAssembly and
+[constraints]({{< ref "_index.md#constraints" >}}). In short, expect
+limitations in both language features and library choices when developing your
+software.
+
+## Memory
+
+When Zig compiles a program into Wasm, it configures the WebAssembly linear memory
+to an initial size of 17 pages (1.1MB), and marks a position in that memory as
```
mathetake, line 70: > an initial size of 17 pages (1.1MB),

maybe add the permalink to this behavior would be helpful.

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+[2]: https://ziglang.org/documentation/master/#WASI
+[5]: https://discord.gg/gxsFFjE
+[6]: https://github.com/tetratelabs/wazero/tree/main/site/content/languages/zig.md
+[7]: https://github.com/tetratelabs/wazero/stargazers
+[9]: https://github.com/tetratelabs/wazero/tree/main/examples/allocation/zig
+[9b]: https://ziglang.org/documentation/0.10.1/#Memory
+[10]: https://github.com/tetratelabs/wazero/tree/main/imports/wasi_snapshot_preview1/example/testdata/zig
+[11]: https://github.com/tetratelabs/wazero/blob/main/imports/wasi_snapshot_preview1/example/testdata/zig/cat.zig
+[12]: https://github.com/ziglang/zig/pull/13513
+[13]: https://ziglang.org/documentation/master/#Build-Mode
```
mathetake, line 213: can this be a permalink rather than master?

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+[WASI]({{< ref "/specs#wasi" >}}).
+
+For example, `zig build-exe -target wasm32-wasi hello.zig` compiles the
+below `main` function into a WASI function exported as `_start`.
+
+```zig
+const std = @import("std");
+const io = std.io;
+const os = std.os;
+const allocator = std.heap.page_allocator;
```
mathetake, line 148: I think allocator is nothing to do with system calls (they do not make any system calls).

site/content/languages/zig.md
```diff
@@ -0,0 +1,213 @@
...
+an operating system.
+
+For functionality the operating system would otherwise provide, you must use
+the `wasm32-wasi` target. This imports host functions in
+[WASI]({{< ref "/specs#wasi" >}}).
+
+For example, `zig build-exe -target wasm32-wasi hello.zig` compiles the
+below `main` function into a WASI function exported as `_start`.
+
+```zig
```
mathetake, line 144: I think we don't need this block, but instead we can simply document something like "system call usage is abstracted on top of Zig's standard library APIs (e.g. `std.io`), and we can use them when compiling against wasm32-wasi target."

mathetake (commented):
excellent !

site/content/languages/zig.md
codefromthecrypt: I don't think we've accumulated any questions often enough for them to be frequent, yet :) Once the doc is in, they can be added, IMHO

site/content/languages/zig.md
codefromthecrypt: I think change the link name to "WebAssembly Documentation" and point to here https://ziglang.org/documentation/master/#WebAssembly

site/content/languages/zig.md
codefromthecrypt: If this isn't released or widely used, yet, we maybe don't need to add it. Are we using it in examples?

This is a first-time doc so we probably should hint to well used/tested things.

site/content/languages/zig.md
evacchi: Yeah I forgot to comment this line: TBH I just took a look at the generated `wasm` and I didn't find any references yet

site/content/languages/zig.md
evacchi: no we are not using it in examples and -- I have just checked -- it has not been released yet, so let's skip it for now

site/content/languages/zig.md
evacchi: I'll group all the switches together:

```
zig build-lib -dynamic -target wasm32-freestanding main.zig
```

site/content/languages/zig.md
evacchi: from further inspection. I think 17 (1088 KiB) may be because of [this behavior in LLVM's Wasm lld](https://lld.llvm.org/WebAssembly.html#cmdoption-initial-memory)

> **`--initial-memory=<value>`**
>     Initial size of the linear memory. Default: static data size.

even an "empty" WASI binary (nop `main`) contains a `data` section:

```
(data (i32.const 1048576) "InputOutput\00TruncatedInput...
```

i.e. 1024 KiB, which is about 17 pages

I don't know if it's worth documenting explicitly, though, it's basically a consequence of a consequence of a consequence :P

site/content/languages/zig.md
evacchi: so, remove the entire section?

site/content/languages/zig.md
evacchi: fixing `[2]` too

site/content/languages/zig.md
evacchi: (I am adding a note for now, let me know)

site/content/languages/zig.md
evacchi: changed to

> This document includes notes contributed by the wazero community *for Zig 0.10.1*.

site/content/languages/zig.md
evacchi: upon further inspection, that block of code is not very useful, so let's kill it

site/content/languages/zig.md
mathetake: I think this is dynamic and depends on the "data" of the compiled program - for example global variables are placed via data section's content and put into these memory. so tldr is that this size is not deterministic. so maybe we could just delete it.

mathetake: that's why this set by the linker - linker is the only component which knows the final size of data in the program.

codefromthecrypt (approved):
I'm happy with things like this, especially as zig's wasm stuff is under construction (some parts wasi-libc and others implemented directly). It is better to have the doc in, and add things upon need.

codefromthecrypt:
Thanks a lot @evacchi!

PR #1052, opened 2023-01-22
ci: adds TinyGo stdlib integration test 

### LINKED ISSUE #1039: Pass TinyGo wasi tests
TinyGo has a target `make test-tinygo-wasi` which ends up running the following via wasmtime. We should be able to do the same via wazero

```bash
$ tinygo test -target wasi -c compress/zlib container/heap container/list container/ring crypto/des crypto/internal/subtle crypto/md5 crypto/rc4 crypto/sha1 crypto/sha256 crypto/sha512 debug/macho embed/internal/embedtest encoding encoding/ascii85 encoding/base32 encoding/csv encoding/hex go/scanner hash hash/adler32 hash/crc64 hash/fnv html internal/itoa internal/profile math math/cmplx net net/http/internal/ascii net/mail os path reflect sync testing testing/iotest text/scanner unicode unicode/utf16 unicode/utf8  crypto/internal/nistec/fiat compress/bzip2 crypto/dsa index/suffixarray  ./tests/runtime_wasi
```

mathetake:
Closes #1039 

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -116,7 +116,7 @@ jobs:
 
       - name: Run standard library tests
         # On PRs, we only test the minimal std packages to test basic functionality.
-        run: tinygo test -target ./wasi_wazero.json os net text/template
+        run: tinygo test -target ./wasi_wazero.json os net math
 
       - name: Run all standard library tests
         # Running all tests is pretty slow (~8min), so only do it on the main branch.
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -44,7 +44,7 @@ jobs:
           path: |
             zig-cache
             zig-out
-          key: ${{ env.ZIG_BUILD_VERSION }}
+          key: integration-test-zig-${{ env.ZIG_BUILD_VERSION }}
 
       - name: Build Stdlib test binary
         # TODO: maybe uploading the built binary to GHA artifact would be helpful to reproduce the failure locally.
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yaml
```diff
@@ -70,3 +70,101 @@ jobs:
...
+          jq '.emulator = "wazero run -hostlogging=filesystem -mount=.:/ -mount={tmpDir}:/tmp {}"' \
+            $(tinygo env TINYGOROOT)/targets/wasi.json > ./wasi_wazero.json
+
+      - name: Run standard library tests
+        # On PRs, we only test the minimal std packages to test basic functionality.
+        run: tinygo test -target ./wasi_wazero.json os math hash
+
+      - name: Run all standard library tests
+        # Running all tests is pretty slow (~8min), so only do it on the main branch.
+        if: ${{ github.event_name  == 'push' }}
```
mathetake, line 124: already verified that this passes https://github.com/tetratelabs/wazero/actions/runs/3978659931/jobs/6820572028

.github/workflows/integration.yaml
```diff
@@ -70,3 +70,100 @@ jobs:
...
+
+      - name: Checkout wazero
+        uses: actions/checkout@v3
+
+      - name: Install wazero
+        run: go install ./cmd/wazero
+
+      - name: Generate target json for wazero.
+        run: |
+          jq '.emulator = "wazero run -hostlogging=filesystem -mount=.:/ -mount={tmpDir}:/tmp {}"' \
```
codefromthecrypt, line 114: maybe it will run faster without host logging? I think it is more helpful when debugging, but "shouldn't" be needed generally

```suggestion
          jq '.emulator = "wazero run -mount=.:/ -mount={tmpDir}:/tmp {}"' \
```

.github/workflows/integration.yaml
```diff
@@ -70,3 +70,100 @@ jobs:
...
+            $(tinygo env TINYGOROOT)/targets/wasi.json > ./wasi_wazero.json
+
+      - name: Run standard library tests
+        # On PRs, we only test the minimal std packages to test basic functionality.
+        run: tinygo test -target ./wasi_wazero.json os math hash
```
codefromthecrypt, line 119: any specific reason for math and hash? `os` is a great one as it makes a lot of sys calls.

One of the following would also be good, if affordable, as they use relative dir lookups. If needed we could swap out math or hash for one?
```
./src/compress/flate:/ ../testdata ../../testdata
./src/image:../testdata
./src/net:/ /etc:/etc
./src/text/template/:/ /tmp:/tmp
```

codefromthecrypt (approved):
This is a big milestone. Great!

.github/workflows/integration.yaml
mathetake: `image` and `text/template` are not even supported by TinyGo itself, and `ompress/flate` is not test target package for Wasm target in TinyGo. So, I swapped hash with the only left `net` .

PR #1054, opened 2023-01-23
fuzz: adds target on checking memory state

mathetake:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -254,4 +254,5 @@ fuzz_timeout_seconds ?= 10
 .PHONY: fuzz
 fuzz:
 	@cd internal/integration_test/fuzz && cargo fuzz run basic -- -max_total_time=$(fuzz_timeout_seconds)
+	@cd internal/integration_test/fuzz && cargo fuzz run memory_no_diff -- -max_total_time=$(fuzz_timeout_seconds)
 	@cd internal/integration_test/fuzz && cargo fuzz run validation -- -max_total_time=$(fuzz_timeout_seconds)
--- a/internal/integration_test/fuzz/fuzz/Cargo.toml
+++ b/internal/integration_test/fuzz/fuzz/Cargo.toml
@@ -14,6 +14,12 @@ wasm-smith = "0.11.4"
 wasmprinter = "0.2.39"
 
 [[bin]]
+name = "memory_no_diff"
+path = "fuzz_targets/memory_no_diff.rs"
+test = false
+doc = false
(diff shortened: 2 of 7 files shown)


mathetake:
now running this locally for an hour or so, and if it ends, I will make this ready

mathetake:
seems like found a bug in memory.fill

```
Failed WebAssembly Text:
(module
  (type (;0;) (func))
  (func (;0;) (type 0)
    (local v128 f32)
    global.get 1
    i32.eqz
    if  ;; label = @1
      unreachable
    end
    global.get 1
    i32.const 1
    i32.sub
    global.set 1
    table.size 1
    table.size 0
    i32.const -1
    i32.rotr
    table.size 1
    memory.fill
  )
  (func (;1;) (type 0)
    global.get 1
    i32.eqz
    if  ;; label = @1
      unreachable
    end
    global.get 1
    i32.const 1
    i32.sub
    global.set 1
  )
  (table (;0;) 128 376 externref)
  (table (;1;) 74 241 externref)
  (memory (;0;) 1 2)
  (global (;0;) (mut funcref) ref.func 1)
  (global (;1;) (mut i32) i32.const 1000)
  (export "" (func 0))
  (export "1" (func 1))
  (export "2" (table 0))
  (export "3" (table 1))
  (export "4" (memory 0))
  (export "5" (global 0))
  (start 1)
  (elem (;0;) (table 1) (i32.const 0) externref)
)
```

mathetake:
minimized by hand :) 

```
(module
  (type (;0;) (func))
  (func (;0;) (type 0)
    table.size 0
    table.size 0
    table.size 0
    memory.fill
  )
  (table (;1;) 1000 1000 externref)
  (memory (;0;) 1 2)
  (start 0)
)
```

mathetake:
ok it's nothing to do with table

```
(module
  (type (;0;) (func))
  (func (;0;) (type 0)
    i32.const 1000
    i32.const 1000
    i32.const 1000
    memory.fill
  )
  (memory (;0;) 1 2)
  (start 0)
)
```

mathetake:
extracted to #1055, and keep this draftt until we solve it

mathetake:
good news is that it seems like there's no other bugs found

mathetake:
so thanks to the fix by @evacchi, now 6 hours of fuzzing failed to find anything.  Marking this ready!

codefromthecrypt (approved):
LGTM

PR #1058, opened 2023-01-23
fs: decouples sysfs from fs.FS and consolidates guest path logic

codefromthecrypt:
This decouples sysfs.FS from fs.FS by introducing a temporary type FSHolder, which will be removed when we top-level FSConfig (shortly).

This further reduces complexity by consolidating guest path configuration into the only type that uses it: CompositeFS.

The next PR will add FSConfig and be a lot smaller due to extracting this stuff now.

### DIFF
--- a/internal/sysfs/dirfs.go
+++ b/internal/sysfs/dirfs.go
@@ -78,8 +78,9 @@ func (d *dirFS) Utimes(name string, atimeNsec, mtimeNsec int64) error {
 }
 
 func (d *dirFS) join(name string) string {
-	if name == "." {
-		return d.cleanedDir
+	switch name {
+	case "", ".", "/":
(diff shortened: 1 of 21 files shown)


codefromthecrypt:
I will fix the windows error tomorrow

internal/sysfs/dirfs.go
```diff
@@ -93,10 +78,11 @@ func (d *dirFS) Utimes(name string, atimeNsec, mtimeNsec int64) error {
 }
 
 func (d *dirFS) join(name string) string {
-	if name == "." {
-		return d.cleanedHostDir
+	switch name {
```
codefromthecrypt, line 81: windows is more sensitive to this than others, but basically there's no need to add a trailing delimiter when requesting the root path.

mathetake (approved):

PR #1061, opened 2023-01-25
fs: adds FSConfig to replace experimental writefs

codefromthecrypt:
This adds a new top-level type FSConfig, which is configured via `ModuleConfig.WithFSConfig(fcfg)`. This implements read-only and read-write directory mounts, something not formally supported before. It also implements `WithFS` which adapts a normal `fs.FS`. For convenience, we retain the old `ModuleConfig.WithFS` signature so as to not affect existing users much.

Here's an example:
```go
moduleConfig = wazero.NewModuleConfig().
	// Make the current directory read-only accessible to the guest.
	WithReadOnlyDirMount(".", "/")
	// Make "/tmp/wasm" accessible to the guest as "/tmp".
	WithDirMount("/tmp/wasm", "/tmp")
```

A new configuration for our emerging sysfs filesystem, `FSConfig.WithMount(sysfs, guestPath)` will happen later without breaking this API. This PR is an important milestone for API stability meanwhile.


codefromthecrypt:
@mathetake lemme know if you think the API is sound. if so, I'll backfill missing unit tests on fsconfig!

mathetake (approved):

codefromthecrypt:
cool I'll backfill w/in 18hrs

PR #1062, opened 2023-01-25
gha: add wasi-testsuite

evacchi:
Checkout `wasi-testsuite` from `tetratelabs/wasi-testsuite@wazero` branch then launch the tests using the wazero adapter. 

It looks fast enough it could run at each PR.

- Windows looks broken, but I think it may be an issue with `-mount` syntax (Windows paths may contain `:`) . I disabled it for now.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -167,3 +167,56 @@ jobs:
             unicode \
             unicode/utf16 \
             unicode/utf8
+
+
+  wasi-testsuite:
+    name: wasi-testsuite
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-12] #todo: windows-2022
+
+    steps:
+      - uses: actions/cache@v3
+        id: cache
+        with:
+          path:
+            ~/.cache/go-build
+          key: integration-test-wasi-testsuite-${{ env.GO_VERSION }}
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Checkout wazero
+        uses: actions/checkout@v3
+
+      - name: Install wazero
+        run: |
+          go install ./cmd/wazero
+        shell: bash
+
+      - name: Checkout wasi-testsuite
+        uses: actions/checkout@v3
+        with:
+          repository: tetratelabs/wasi-testsuite
+          ref: wazero
+
+      - name: Initialize Python environment
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10' # latest version of python 3
+          cache: pip
+
+      - name: Install dependencies
+        working-directory: test-runner
+        run: |
+          pip install -r requirements/dev.txt
+
+      - name: Run all wasi-testsuite
+        run: |
+              python3 test-runner/wasi_test_runner.py \
+                -t ./tests/assemblyscript/testsuite/ \
+                ./tests/c/testsuite/ \
+                -r adapters/wazero.sh
+


.github/workflows/integration.yaml
```diff
@@ -167,3 +167,40 @@ jobs:
             unicode \
             unicode/utf16 \
             unicode/utf8
+
+
+  wasi-testsuite:
+    name: wasi-testsuite
+    runs-on: ubuntu-20.04
```
mathetake, line 174: let's run mac and windows as well!

evacchi: oh! 😬

evacchi: is that     `runs-on: ${{ matrix.os }}` ?

mathetake: yep

evacchi: missing Python module on macOS:
- ModuleNotFoundError: No module named 'colorama'
[18](https://github.com/tetratelabs/wazero/actions/runs/4004990710/jobs/6874809700#step:7:19)

path syntax error on windows?
- go install ./cmd/wazero
[3](https://github.com/tetratelabs/wazero/actions/runs/4004990710/jobs/6874809815#step:5:3)

.github/workflows/integration.yaml
```diff
@@ -167,3 +167,43 @@ jobs:
             unicode \
             unicode/utf16 \
             unicode/utf8
+
+
+  wasi-testsuite:
```
mathetake, line 172: You have to install the python dependencies via `pip install -r requirements/dev.txt` and please configure the proper caching for them`

https://github.com/WebAssembly/wasi-testsuite/blob/e5bcece941f66f8f6c845a9add32f9f99e444a2c/.github/workflows/test-runner.yml#L31

evacchi: should be ok now

evacchi:
I think I managed to run the scripts on Windows as well, but it was not trivial. The "fix" I tried is to set an env variable to make sure the runners are able to find `bash` (which is not really on the PATH). see https://github.com/tetratelabs/wasi-testsuite/pull/1

It looks like some tests are not passing but it is likely to be the `-mount` directive not being parsed/passed correctly

.github/workflows/integration.yaml
```diff
@@ -167,3 +167,64 @@ jobs:
...
+    name: wasi-testsuite
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-12, windows-2022]
```
codefromthecrypt: switch windows-2022 to a TODO comment

.github/workflows/integration.yaml
```diff
@@ -167,3 +167,64 @@ jobs:
...
+      - name: Checkout wasi-testsuite
+        uses: actions/checkout@v3
+        with:
+          repository: evacchi/wasi-testsuite
+          ref: gha/wazero-win
+      
+      - name: Initialize Python environment
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
codefromthecrypt, line 208: comment if this can be any version of python (e.g. `. # latest version of python 3` if it isn't specific)

codefromthecrypt (approved):
ps when we get to windows, `echo "EXTRA_ENV_VAR=value" >> $GITHUB_ENV` in a conditional step will be better than repeating the latter.

codefromthecrypt:
actually I wonder if this would have worked on windows vs setting that ENV variable. In func-e we force bash for consistency https://github.com/tetratelabs/func-e/blob/master/.github/workflows/packaging.yaml#L34

codefromthecrypt:
Thanks for this!

evacchi:
Unfortunately no I tried that. Apparently the "right" bash is not on the PATH. An alternative fix may be to fiddle with PATH too but I wonder how much pain it is to do it in GHA, on Windows, x-platform. I tried to figure it out and it looked complicated

codefromthecrypt:
Thanks for trying! In any case if we default to bash we can use the append syntax to add special variables, even if we yeah still need to add those. e.g. `echo "$WIX\\bin" >> $GITHUB_PATH` works now in func-e's GHA

codefromthecrypt:
or we can use normal template variables. In any case, setting variables is the easier part.

PR #1064, opened 2023-01-26
Add -hostlogging=crypto

### LINKED ISSUE #1040: Add `-hostlogging=crypto`

For troubleshooting, I would like to see RNG access via a flag `-hostlogging=crypto`. This should help people understand when RNG that isn't real is in use, or otherwise debugging crypto setup issues.

Right now, we have `-hostlogging=filesystem`. For troubleshooting, I would like to see crypto access via a flag `-hostlogging=crypto`. This should be mixable with filesystem like `-hostlogging=filesystem,crypto`

The name crypto is arbitrary, but following the convention of WASI, even though this should also apply to GOOS=js (gojs)
https://github.com/WebAssembly/wasi-crypto.

To implement this involves a number of changes. Minimally get WASI working at the same time as the logging hooks, GOJS can be a different PR later, or at the same time, whichever is easier.

#### Logging API

* update internal/logging/logging.go 
  * Add LogScopes with two bitflag vals LogScopeFilesystem and LogScopeCrypto. 
  * Look at CoreFeatures for an example. We should be able to tell which are set and it should print pretty e.g. "filesystem|crypto"
* update experimental/logging/log_listener.go 
  * Change NewFilesystemLoggingListenerFactory to NewScopedLoggingListenerFactory and pass LogScopes into it instead
  * Use the LogScopes in NewListener where you see callouts to IsFilesystemFunction. e.g. switch to scope logic
* update cmd/wazero/wazero.go
  * change hostLoggingFlag to one that can parse multiple scopes, but fail if unknown. e.g. only "filesystem" "crypto" or "filesystem,crypto" (not empty!) are allowed.
  * Use that in maybeHostLogging to adjust for the new NewScopedLoggingListenerFactory

#### WASI
* update internal/wasi-snapshot-preview1/logging/logging.go
  * Change IsFilesystemFunction to IsInLogScope(scopes) e.g. when LogScopeFilesystem is enabled.
    * Add LogScopeCrypto to that when RandomGetName ("random_get")
 * Make an ad-hoc test that triggers "random_get" so that you can verify logging works with the scope is enabled, and that there is no logging when it is disabled.
 * There may be some finesse needed in the log format which you'll notice

#### gojs
* update internal/gojs/logging/logging.go
  * IsInLogScope(scopes) needs to be added, because crypto is used twice in gojs
    * This should return true for custom.NameSyscallValueCall if there is any scope enabled
    * It should also return true on NameRuntimeGetRandomData ("runtime.getRandomData") if LogScopeCrypto is enabled
  * gologging.Config needs to change
     * return a constant logger NameRuntimeGetRandomData if LogScopeCrypto is enabled
     * move logic from syscallValueCallParamSampler, syscallValueCallParamLogger and syscallValueCallResultLogger into a new type that is aware of the log scopes, and handle accordingly
     * This should handle filesystem as it is now, and additionally "crypto" "getRandomValues" should log when LogScopeCrypto is enabled. This may require adding a small file custom/crypto.go like js.go even though there's only one function used.

* Make an ad-hoc test that triggers "runtime.getRandomData" as well  "syscall/js.valueCall" with "crypto" "getRandomValues", so that you can verify logging works with the scope is enabled, and that there is no logging when it is disabled.
* There may be some finesse needed in the log format which you'll notice

e.g. make a new type that can see the log scopes, then you can return it if crypto, filesystem or both are enabled
```go
var _ logging.ParamSampler = syscallValueLogger
var _ logging.ParamLogger = syscallValueLogger
var _ logging.ResultLogger = syscallValueLogger
type syscallValueLogger logging.LogScopes
```

implement an "all" scope, but I don't think we should do that quite yet as many things are verbose, such as clocks!

**Additional context**

Currently, we only have one logging scope: `filesystem`, enabled via `-hostlogging=filesystem`. The name `filesystem` was taken from https://github.com/WebAssembly/wasi-filesystem, even though the logic applies also to gojs.

evacchi:
This allows you to specify multiple logging scopes, both in API and the CLI.

e.g for the CLI.
```bash
$ wazero run --hostlogging=crypto --hostlogging=filesystem --mount=.:/:ro cat.wasm
```

e.g. for Go
```go
	logging.NewHostLoggingListenerFactory(&log, logging.LogScopeCrypto|logging.LogScopeFilesystem))
```

fixes #1040


internal/logging/logging.go
```diff
@@ -33,6 +34,46 @@ const (
...
+)
+
+func scopeName(s LogScopes) string {
+	switch s {
+	case LogScopeFilesystem:
+		return "filesystem"
+	case LogScopeCrypto:
+		return "crypto"
+	default:
+		return "<unknown>"
```
codefromthecrypt, line 51: maybe format <unknown=%d>

experimental/logging/log_listener.go
```diff
@@ -91,7 +90,7 @@ func (f *loggingListenerFactory) NewListener(fnd api.FunctionDefinition) experim
 			return nil // not yet supported
 		}
 	default:
-		if f.fsOnly {
+		if f.scopes.IsInLogScope(logging.LogScopeFilesystem) {
```
codefromthecrypt, line 93: this is breaking my brain at the moment.. not sure it fits to say anything about FS anymore

internal/logging/logging.go
```diff
@@ -33,6 +34,46 @@ const (
 	ValueTypeString = 0xff
 )
 
+type LogScopes uint64
```
codefromthecrypt, line 37: don't forget to copy/pasta the tests

internal/wasi_snapshot_preview1/logging/logging.go
```diff
@@ -14,11 +14,13 @@ import (
 
 var le = binary.LittleEndian
 
-func IsFilesystemFunction(fnd api.FunctionDefinition) bool {
+func IsInLogScope(fnd api.FunctionDefinition, scopes logging.LogScopes) bool {
```
codefromthecrypt, line 17: I might keep ` IsFilesystemFunction` as `isFilesystemFunction` then dispatch to it from `IsInLogScope`. This should keep the logic easier.

so IsInLogScope would be a case statement for the known scopes, which does a `fallthrough` each time as they can compose (watch the edge case of filesystem|crypto)

experimental/logging/log_listener.go
evacchi: yeah I changed that to `f.scopes!=0` (and actually `f.scopes.Defined()`)

experimental/logging/testdata/listener.wat
```diff
@@ -1,3 +1,4 @@
+;; copied from experimental/logging/testdata/listener.wat
```
mathetake, line 1: self copying? 😄

cmd/wazero/wazero_test.go
```diff
@@ -251,6 +254,15 @@ func TestRun(t *testing.T) {
 <== (nread=5,errno=ESUCCESS)
 `,
 		},
+		{
```
mathetake, line 257: I believe we are missing the composed scope case `"--hostlogging=crypto,filesystem`

codefromthecrypt:
@evacchi since I'm working this weekend, I'm going to yank it for the last couple polishing steps. thanks tons for getting it this far!

cmd/wazero/wazero_test.go
```diff
@@ -285,13 +297,27 @@ func TestRun(t *testing.T) {
 <== (err=<nil>,ok=true)
 `, bearMode, bearMtime),
 		},
+		{
+			name:       "GOARCH=wasm GOOS=js hostlogging=crypto and filesystem",
+			wasm:       wasmCat,
+			wazeroOpts: []string{"--hostlogging=crypto", "--hostlogging=filesystem"},
```
codefromthecrypt, line 303: we only need to test combination once. this was the easiest place since `GOARCH=wasm GOOS=js` initializes with crypto functions.

codefromthecrypt:
TODO: add gojs crypto.getRandomValues and assemblyscript seed

codefromthecrypt:
updated summary comment. will add the missing host functions in a while

experimental/logging/log_listener.go
```diff
@@ -68,7 +82,7 @@ type flusher interface {
 // experimental.FunctionListener.
 func (f *loggingListenerFactory) NewListener(fnd api.FunctionDefinition) experimental.FunctionListener {
 	exported := len(fnd.ExportNames()) > 0
-	if (f.hostOnly || f.fsOnly) && // choose functions defined or callable by the host
+	if (f.hostOnly || f.scopes != LogScopeNone) && // choose functions defined or callable by the host
```
codefromthecrypt, line 85: LogScopeNone comparisons in this version of the change are wrong. We need as @evacchi suggested offline an all bits set LogScopeAny and do it the other way.

codefromthecrypt:
done. the last commit was really gnarly as the code wasn't setup to log anything except async functions (fs is async while crypto is not).

internal/gojs/logging/logging.go
```diff
@@ -90,42 +144,59 @@ func syscallValueCallResultLogger(ctx context.Context, mod api.Module, w logging
...
-			w.WriteByte('=')                          //nolint
-			w.WriteString(fmt.Sprintf("%v", vals[i])) //nolint
+			w.WriteByte(',') //nolint
+			writeVal(w, name, val)
 		}
 	}
 }
 
+func writeVal(w logging.Writer, name string, val interface{}) {
+	if b, ok := val.(*goos.ByteArray); ok {
```
codefromthecrypt, line 188: exposed this (still internal) as there was no other way to access it in the logging package.

internal/gojs/crypto_test.go
```diff
@@ -1,19 +1,34 @@
...
+	loggingCtx := context.WithValue(testCtx, experimental.FunctionListenerFactoryKey{},
+		logging.NewHostLoggingListenerFactory(&log, logging.LogScopeCrypto))
+
+	stdout, stderr, err := compileAndRun(loggingCtx, "crypto", wazero.NewModuleConfig())
 
 	require.Zero(t, stderr)
 	require.EqualError(t, err, `module "" closed with exit_code(0)`)
 	require.Equal(t, `7a0c9f9f0d
 `, stdout)
+	require.Equal(t, `==> go.runtime.getRandomData(r_len=32)
```
codefromthecrypt, line 27: this is where we can see the two places gojs can end up calling a crypto related host function

codefromthecrypt:
looks like go 1.17 initializes randoms in opposite order. I'll try to sort that out.

codefromthecrypt:
lol

```
FAIL	github.com/tetratelabs/wazero/internal/gojs [build failed]
[37](https://github.com/tetratelabs/wazero/actions/runs/4031228233/jobs/6930394563#step:4:38)
Error: internal\gojs\crypto_test.go:34:27: info.GoVersion undefined (type *debug.BuildInfo has no field or method GoVersion)
```

codefromthecrypt:
ok I'll just ignore the test until we can remove go 1.17 which is in a couple weeks

codefromthecrypt:
finally 💚

PR #1072, opened 2023-01-28
logging: adds poll scope

codefromthecrypt:
This allows you to specify the poll scope amongst existing logging scopes, both in API and the CLI.

e.g for the CLI.
```bash
$ wazero run --hostlogging=poll --hostlogging=filesystem --mount=.:/:ro cat.wasm
```

e.g. for Go
```go
	logging.NewHostLoggingListenerFactory(&log, logging.LogScopePoll|logging.LogScopeFilesystem))
```

This is particularly helpful to understand side-effects of GC. For example, in `GOOS=js` GC will trigger events and these have been tricky to troubleshoot in the past.

### DIFF
--- a/internal/gojs/misc_test.go
+++ b/internal/gojs/misc_test.go
@@ -2,11 +2,14 @@ package gojs_test
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"strings"
 	"testing"
 
 	"github.com/tetratelabs/wazero"
+	"github.com/tetratelabs/wazero/experimental"
+	"github.com/tetratelabs/wazero/experimental/logging"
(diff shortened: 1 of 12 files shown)


internal/gojs/misc_test.go
```diff
@@ -47,14 +50,27 @@ func Test_stdio(t *testing.T) {
 func Test_stdio_large(t *testing.T) {
 	t.Parallel()
 
+	// Large stdio will trigger GC which will trigger events.
```
codefromthecrypt, line 53: we had an existing heavier test (which flaked in go 1.17 until a recent workaround), so re-purposed it to also prove that the poll logging scope works.

PR #1074, opened 2023-01-29
logging: adds exit scope and fixes mtim bug

codefromthecrypt:
This allows you to specify the exit scope amongst existing logging scopes, both in API and the CLI.

e.g for the CLI.
```bash
$ wazero run --hostlogging=exit,filesystem --mount=.:/:ro cat.wasm
```

e.g. for Go
```go
	logging.NewHostLoggingListenerFactory(&log, logging.LogScopeExit|logging.LogScopeFilesystem))
```

This is helpful to know if the wasm called exit or if it exited implicitly. This is one of the few host functions that exists in three places: assemblyscript, gojs and wasi.

### DIFF
--- a/internal/wasi_snapshot_preview1/logging/logging.go
+++ b/internal/wasi_snapshot_preview1/logging/logging.go
@@ -18,6 +18,10 @@ func isClockFunction(fnd api.FunctionDefinition) bool {
 	return strings.HasPrefix(fnd.Name(), "clock_")
 }
 
+func isExitFunction(fnd api.FunctionDefinition) bool {
+	return fnd.Name() == ProcExitName
+}
+
 func isFilesystemFunction(fnd api.FunctionDefinition) bool {
 	switch {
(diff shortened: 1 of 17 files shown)


internal/wasi_snapshot_preview1/logging/logging.go
```diff
@@ -210,7 +220,7 @@ func (i logFilestat) Log(_ context.Context, mod api.Module, w logging.Writer, pa
 		w.WriteString(",size=")              //nolint
 		writeI64(w, le.Uint64(buf[32:]))
 		w.WriteString(",mtim=") //nolint
-		writeI64(w, le.Uint64(buf[40:]))
+		writeI64(w, le.Uint64(buf[48:]))
```
codefromthecrypt, line 223: this was a bug as it was returning atim not mtim

codefromthecrypt:
next and final for this month's version is the "memory" scope.

PR #1084, opened 2023-01-31
wasi: supports rewinding a directory and fixes seek error

codefromthecrypt:
This allows rewinding a directory by specifying the cookie 0, after
already scrolling a directory. This allows zig and clang functionality
to work which expect this.

This also fixes the case where a directory was allowed to be seeked.
This is prohibited by the current version of wasi-testsuite.


codefromthecrypt:
After this, `directory_seek` passes on the rust wasi-testsuite, leaving 5 failures left.

PR #1086, opened 2023-01-31
fs: handles windows edge cases on rename

codefromthecrypt:
This handles edge cases on rename which were previously skipped in testing. Notably, there are delete flags not possible to propagate in current versions of go, which imply some copy/paste meanwhile.

### DIFF
--- a/internal/platform/open_file_windows.go
+++ b/internal/platform/open_file_windows.go
@@ -1,0 +1,70 @@
+package platform
+
+import (
+	"io/fs"
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+func OpenFile(name string, flag int, perm fs.FileMode) (*os.File, error) {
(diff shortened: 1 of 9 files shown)


internal/platform/open_file_windows.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"io/fs"
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+func OpenFile(name string, flag int, perm fs.FileMode) (*os.File, error) {
+	fd, err := windowsOpenFile(name, flag|syscall.O_CLOEXEC, uint32(perm))
```
codefromthecrypt, line 11: I ran into this recently. So, here, we are opening the file, then if not, opening it again. I want to add the comment above `windowsOpenFile` which is why `syscall.O_CLOEXEC` and what case we expect this to err differently than `os.OpenFile` regardless of what the flags are (as the only change is O_CLOEXEC).

PR #1089, opened 2023-02-01
fix: should not panic when a module is instantiated with a closed runtime

### LINKED ISSUE #1088: Runtime.InstantiateModuleFromBinary() should not panic if Runtime has been closed
**Describe the bug**
While upgrading https://github.com/wapc/wapc-go to v1.0.0-pre.8 we found `TestEngineWithRuntime` to fail. Runtime panics when it has been closed.

**To Reproduce**
```
r := NewRuntime(ctx)
r.Close(cox)
r.InstantiateModuleFromBinary(ctx, someBinary)
```

**Expected behavior**
err should be returned, panics instead.

- Go version: go1.19.5 
- wazero Version: 1.0.0-pre.8
- Host architecture: darwin/arm64

evacchi:
fixes #1088


codefromthecrypt (commented):
while this works, I prefer a less cryptic error. how about if we add a closed field to runtime similar to what's in CallContext? then a similar guard to FailIfClosed on the exported methods, except an error message like "runtime closed" instead of an exit error?

internal/wasm/store_module_list.go
```diff
@@ -86,6 +86,9 @@ func (s *Store) requireModules(moduleNames map[string]struct{}) (map[string]*Mod
 func (s *Store) requireModuleName(moduleName string) error {
 	s.mux.Lock()
 	defer s.mux.Unlock()
+	if s.nameToNode == nil {
+		return fmt.Errorf("module[%s] cannot be found: no instantiated modules in this store", moduleName)
+	}
```
evacchi, line 91: keeping this check anyway

runtime_test.go
```diff
@@ -665,6 +665,30 @@ func TestRuntime_Close_ClosesCompiledModules(t *testing.T) {
 	}
 }
 
+func TestRuntime_Close_Instantiate_ShouldErr(t *testing.T) {
+	t.Run("TestRuntime_Close_Instantiate_ShouldErr", func(t *testing.T) {
```
mathetake, line 669: I think this is unnecessary > t.Run

runtime.go
```diff
@@ -203,8 +219,19 @@ func buildListeners(ctx context.Context, internal *wasm.Module) ([]experimentala
 	return listeners, nil
 }
 
+// FailIfClosed returns an error if CloseWithExitCode was called implicitly (by Close) or explicitly.
+func (r *runtime) FailIfClosed() error {
```
codefromthecrypt, line 223: no need to export

runtime_test.go
```diff
@@ -665,6 +665,30 @@ func TestRuntime_Close_ClosesCompiledModules(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, uint32(1), engine.CompiledModuleCount())
+
+		err = r.Close(testCtx)
+		require.NoError(t, err)
+
+		// Closing the runtime should remove the compiler cache if cache is not configured.
+		require.True(t, engine.closed)
+
+		_, err = r.InstantiateModuleFromBinary(testCtx, binaryNamedZero)
```
codefromthecrypt, line 687: add check for CompileModule too

codefromthecrypt (commented):
getting there!

runtime_test.go
evacchi: of course 🤦

runtime_test.go
evacchi: maybe I should just move `failIfClosed()` to `CompileModule()`, since `InstantiateModuleFromBinary()` calls that

runtime_test.go
```diff
@@ -665,6 +665,49 @@ func TestRuntime_Close_ClosesCompiledModules(t *testing.T) {
...
+				_, err := r.InstantiateModuleFromBinary(testCtx, binaryNamedZero)
+				return err
+			},
+		},
+		{
+			name: "CompileModule",
+			instantiateFunc: func(r Runtime) error {
+				_, err := r.CompileModule(testCtx, binaryNamedZero)
+				return err
+			},
```
evacchi, line 685: redundant? @codefromthecrypt

codefromthecrypt: sorry about late feedback. I don't really think it is redundant except that the implementation of `InstantiateModuleFromBinary` happens to call this. I think it will feel less redundant when `InstantiateModule` is guarded and tested. So, if you want to swap one out I would swap out `InstantiateModuleFromBinary` for `InstantiateModule`.

runtime_test.go
```diff
@@ -665,6 +665,49 @@ func TestRuntime_Close_ClosesCompiledModules(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, uint32(1), engine.CompiledModuleCount())
+
+			err = r.Close(testCtx)
+			require.NoError(t, err)
+
+			// Closing the runtime should remove the compiler cache if cache is not configured.
+			require.True(t, engine.closed)
+
+			require.Error(t, tc.instantiateFunc(r))
```
codefromthecrypt, line 706: I think we should use EqualError here, as the Error could otherwise be cryptic in a way we'd not notice

runtime.go
```diff
@@ -271,7 +299,16 @@ func (r *runtime) Close(ctx context.Context) error {
 }
 
 // CloseWithExitCode implements Runtime.CloseWithExitCode
+//
+// Note: it also marks the internal `closed` field
 func (r *runtime) CloseWithExitCode(ctx context.Context, exitCode uint32) error {
+	r.mux.Lock()
```
codefromthecrypt, line 305: I don't think this lock is doing anything so we should remove it. the atomic compareAndSwap below will prevent multiple goroutines from passing.

codefromthecrypt (commented):
sorry about the late feedback I think the main thing here is let's switch to guard `InstantiateModule` instead of `InstantiateModuleFromBinary` as it prevents more misuse and also reduces confusion about the latter

runtime.go
```diff
@@ -219,6 +242,10 @@ func (r *runtime) InstantiateModule(
 	compiled CompiledModule,
 	mConfig ModuleConfig,
 ) (mod api.Module, err error) {
+	if e := r.failIfClosed(); e != nil {
+		return nil, e
```
mathetake, line 246: nit, but we gophers prefer `err`  over errors whenever possible.

```suggestion
	if err = r.failIfClosed(); err != nil {
		return
```

runtime_test.go
```diff
@@ -665,6 +665,49 @@ func TestRuntime_Close_ClosesCompiledModules(t *testing.T) {
 	}
 }
 
+func TestRuntime_Close_Instantiate_ShouldErr(t *testing.T) {
```
mathetake, line 668: we don't have this long top leve test name
```suggestion
// TestRuntime_Closed ensure invocation of closed Runtime's methods is safe.
func TestRuntime_Closed(t *testing.T) {
```

runtime_test.go
```diff
@@ -665,6 +665,49 @@ func TestRuntime_Close_ClosesCompiledModules(t *testing.T) {
...
+	for _, tc := range []struct {
+		name            string
+		instantiateFunc func(r Runtime, mod CompiledModule) error
```
mathetake, line 671: `instantiateFunc` is not representative anymore (CompileModule is nothing to do with instantiation)

runtime_test.go
```diff
@@ -665,6 +665,49 @@ func TestRuntime_Close_ClosesCompiledModules(t *testing.T) {
...
+				_, err := r.InstantiateModule(testCtx, mod, NewModuleConfig())
+				return err
+			},
+		},
+		{
+			name: "CompileModule",
+			instantiateFunc: func(r Runtime, mod CompiledModule) error {
+				_, err := r.CompileModule(testCtx, binaryNamedZero)
+				return err
+			},
```
mathetake, line 685: shall we test other methods as well so that we never reach panics?

mathetake (commented):
almost there!

runtime_test.go
codefromthecrypt: InstantiateModule should be tested as that can be called without calling InstantiateModuleFromBinary. imho that's the last one

runtime.go
mathetake: is this fixed?

mathetake (approved):

runtime.go
evacchi: I think I only fixed one 😬

runtime_test.go
evacchi: it's now InstantiateModule + InstantiateModuleFromBinary + CompileModule so we should be good?

codefromthecrypt (approved):
Thanks!

PR #1057, opened 2023-01-23
wasi: implements link related system calls

mathetake:
This fixes #1029

Issue #1029, opened 2023-01-12
Pass automated tests for zig stdlib


codefromthecrypt:
**Is your feature request related to a problem? Please describe.**
One of our goals of 1.0 is to reduce the amount of glitches using syscall ABIs, notably WASI. While there is an emerging [wasi-testsuite](https://github.com/WebAssembly/wasi-testsuite), and we have a number of ad-hoc tests, we should also pass zig tests as an integration suite.

**Describe the solution you'd like**

#1029 runs automated tests, but we don't yet pass them

from zig clone
```bash
$ zig test --test-cmd wazero --test-cmd run --test-cmd -mount=.:/ --test-cmd-bin -target wasm32-wasi --zig-lib-dir ./lib ./lib/std/std.zig
```

The actual mounts may need to vary once above actually passes. e.g. mounting tmp for writes

**Describe alternatives you've considered**

We currently do this manually and backfill notable gaps as unit tests, but the logic zig uses changes from time to time.

**Additional context**

As of Jan 13, this doesn't proceed as we've not fully implemented wasi syscalls needed, nor have the correct flags to proceed. e.g. /tmp
```
$ zig test --test-cmd wazero --test-cmd run --test-cmd -hostlogging=filesystem --test-cmd -mount=.:/ --test-cmd-bin -target wasm32-wasi --zig-lib-dir ./lib ./lib/std/std.zig
unexpected errno: 21
Unable to dump stack trace: not implemented for Wasm
unexpected errno: 21
Unable to dump stack trace: not implemented for Wasm
expected error.EnvironmentVariableNotFound, found error.OutOfMemory
==> wasi_snapshot_preview1.path_create_directory(fd=3,path=zig-cache)
<== errno=EEXIST
[...]

PR #1090, opened 2023-02-01
ci: runs Zig stdlib tests on mac and windows

mathetake:
part of #1029, extracted from #1057 

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -27,51 +27,87 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  zig:
-    name: Zig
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yaml
```diff
@@ -27,51 +27,87 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  zig:
-    name: Zig
+  # This only ensures the existence of Zig's test binary, and exists in order to work hard to build on
```
codefromthecrypt, line 30: this sentence is awkward. Can this be simplified to "This builds a zig test binary only if the same version tag hasn't been build before. This saves time as we rarely update the zig version."?

codefromthecrypt (approved):
LGTM

PR #1091, opened 2023-02-02
ci: runs TinyGo stdlib tests on mac and windows

### LINKED ISSUE #1039: Pass TinyGo wasi tests

mathetake:
Closes #1039 

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -195,7 +195,6 @@ jobs:
           do
             tinygo test -target wasi -c -o ${{ env.BINARY_OUT }}/${value/\//_}.test $value
           done
-          ls ${{ env.BINARY_OUT }}
 
       - name: Checkout wazero
         if: steps.check-exist.outputs.exist == 'false'
@@ -236,17 +235,6 @@ jobs:
         run: go install ./cmd/wazero
 
       - name: Run standard library tests
-        # On PRs, we only test the minimal std packages to test basic functionality.
-        if: ${{ github.event_name  != 'push' }}
-        run: |
-          for bin in os net math; do
-            echo $bin
-            wazero run -mount=:/ -mount=:/tmp $bin.test -- -test.v
-          done
-
-      - name: Run all standard library tests
-        # Running all tests is pretty slow (~8min), so only do it on the main branch.
-#        if: ${{ github.event_name  == 'push' }}
         run: |
           for bin in *.test; do
             echo $bin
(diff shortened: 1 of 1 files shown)


mathetake:
lol https://github.com/golang/go/blob/40a0986959167e7b82ae16a35a5dd2a347609eff/src/compress/zlib/writer_test.go#L16-L30

mathetake:
cool

.github/workflows/integration.yaml
```diff
@@ -207,6 +192,54 @@ jobs:
...
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Install wazero
+        run: go install ./cmd/wazero
+
+      - name: Run standard library tests
+        run: |
+          for bin in *.test; do
```
mathetake, line 239: it turns out that the majority of the time was spent on compiling wasm binary by tinygo, so it seems all test finishes in like 30 sec or so.

.github/workflows/integration.yaml
```diff
@@ -177,7 +163,6 @@ jobs:
             crypto/sha1 \
             crypto/sha256 \
             crypto/sha512 \
-            debug/macho \
```
codefromthecrypt, line 180: no macho no mas

.github/workflows/integration.yaml
codefromthecrypt: good news cc @dgryski

codefromthecrypt (approved):
cool I'm glad to see you re-using the "build if not distributed" approach to ensure we can test all platforms wazero's compiler engine should work on!

PR #1092, opened 2023-02-02
dirfs: use os.ReadLink instead of syscall variant

mathetake:
This fixes the race found on the current main branch.


mathetake:
verified this works

codefromthecrypt (approved):
great work and awesome comment!

PR #1093, opened 2023-02-02
support O_DIRECTORY and O_NOFOLLOW

achille-roussel:
Follow up to our discussion on Slack, this PR adds support for setting the O_DIRECTORY and O_NOFOLLOW flags when the wasi flags O_DIRECTORY and LOOKUP_SYMLINK_FOLLOW are set.


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -1529,8 +1530,15 @@ func preopenPath(fsc *sys.FSContext, dirFD uint32) (string, Errno) {
...
-	isDir = oflags&O_DIRECTORY != 0
+func openFlags(dirflags, oflags, fdflags uint16) (openFlags int) {
+	if dirflags&LOOKUP_SYMLINK_FOLLOW == 0 {
+		openFlags |= platform.O_NOFOLLOW
+	}
+	if oflags&O_DIRECTORY != 0 {
+		openFlags |= platform.O_DIRECTORY
+	} else if oflags&O_EXCL != 0 {
+		openFlags |= syscall.O_EXCL
+	}
```
achille-roussel, line 1541: I retained the logic that was based on the `isDir` boolean below tho I'm not fully grasping the intent here.

We add flags like O_TRUNC which do not make sense on directories, but explicitly avoid O_EXCL?

codefromthecrypt: I think the early return on directory was too far down, sorry about the confusion. there was no intent to allow nonsense flags on it, though certainly takes too much thinking to know what makes sense or not. one day I hope WASI actually specifies things like this.

TL;DR; move the early return to where it makes sense?

mathetake:
rebased && fixed GOOS=js

mathetake (approved):
@achille-roussel thanks as always!

PR #1096, opened 2023-02-03
ci: supports building with Go 1.20 and raises floor version to 1.18

### LINKED ISSUE #921: update build to go 1.20
We should prepare for Go 1.20, as this is the version we cut 1.0 with early next year.

```diff
-          - "1.19"  # Current Go version
-          - "1.17"  # Floor Go version of wazero (current - 2)
+          - "1.20rc1"  # Next Go version
+          - "1.17"      # Floor Go version of wazero (current - 2)
```

There are some issues to solve with 1.20rc1

* `make lint` fails with typecheck https://github.com/golangci/golangci-lint/issues/3420
  * Note: this doesn't necessarily mean disabling typecheck will fix it. see https://github.com/golangci/golangci-lint/issues/2912
* one unit test fails

codefromthecrypt:
This moves our floor version to the same we'll release 1.0 with: 1.18. This is congruent with our version policy which is current-2.

Fixes #921

### DIFF
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -362,7 +362,7 @@ const (
 
 	// Offsets for Go's interface.
 	// https://research.swtch.com/interfaces
-	// https://github.com/golang/go/blob/release-branch.go1.17/src/runtime/runtime2.go#L207-L210
+	// https://github.com/golang/go/blob/release-branch.go1.20/src/runtime/runtime2.go#L207-L210
 	interfaceDataOffset = 8
 
 	// Consts for wasm.DataInstance.
--- a/internal/engine/compiler/impl_amd64.go
+++ b/internal/engine/compiler/impl_amd64.go
@@ -4957,7 +4957,7 @@ func (c *amd64Compiler) compileModuleContextInitialization() error {
 		// implementation of interface. This case, we extract "data" pointer as *moduleEngine.
 		// See the following references for detail:
 		// * https://research.swtch.com/interfaces
-		// * https://github.com/golang/go/blob/release-branch.go1.17/src/runtime/runtime2.go#L207-L210
+		// * https://github.com/golang/go/blob/release-branch.go1.20/src/runtime/runtime2.go#L207-L210
 		c.assembler.CompileMemoryToRegister(amd64.MOVQ, amd64CallingConventionDestinationFunctionModuleInstanceAddressRegister, moduleInstanceEngineOffset+interfaceDataOffset, tmpRegister)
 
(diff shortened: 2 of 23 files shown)


internal/engine/compiler/engine.go
```diff
@@ -362,7 +362,7 @@ const (
 
 	// Offsets for Go's interface.
 	// https://research.swtch.com/interfaces
-	// https://github.com/golang/go/blob/release-branch.go1.17/src/runtime/runtime2.go#L207-L210
+	// https://github.com/golang/go/blob/release-branch.go1.20/src/runtime/runtime2.go#L207-L210
```
mathetake, line 365: Is this still correct lines?

internal/engine/compiler/impl_amd64.go
```diff
@@ -4957,7 +4957,7 @@ func (c *amd64Compiler) compileModuleContextInitialization() error {
 		// implementation of interface. This case, we extract "data" pointer as *moduleEngine.
 		// See the following references for detail:
 		// * https://research.swtch.com/interfaces
-		// * https://github.com/golang/go/blob/release-branch.go1.17/src/runtime/runtime2.go#L207-L210
+		// * https://github.com/golang/go/blob/release-branch.go1.20/src/runtime/runtime2.go#L207-L210
```
mathetake, line 4960: Same

mathetake (approved):

codefromthecrypt:
oops windows hates us again. I'll look into it later

internal/engine/compiler/engine.go
codefromthecrypt: yep I checked!

codefromthecrypt:
yay green on windows finally! I'll squash

internal/sys/fs.go
```diff
@@ -394,15 +394,18 @@ func (c *FSContext) Renumber(from, to uint32) error {
...
+	// If toFile is already open, we close it to prevent windows lock issues.
+	//
+	// The doc is unclear and other implementations do nothing for already-opened To FDs.
 	// https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-fd_renumberfd-fd-to-fd---errno
 	// https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-common/src/snapshots/preview_1.rs#L531-L546
+	if toFile, ok := c.openedFiles.Lookup(to); ok {
+		if toFile.IsPreopen {
+			return syscall.ENOTSUP
+		}
+		_ = toFile.File.Close()
```
codefromthecrypt, line 406: go 1.20 is more strict on windows, so we have to watch file close everywhere, not just mostly everywhere.

PR #1097, opened 2023-02-03
benchmark: bumps vs tests wasmtime and wasmedge to latest

codefromthecrypt:
I tried to run wasmedge locally as they publish arm64 os/x now, but it fails like below even sourcing their env. I left that alone for now

```
dyld[88003]: Library not loaded: @rpath/libwasmedge.0.dylib
  Referenced from: <B1821F61-C84A-38EE-BC33-F3DFB1837BEB> /private/var/folders/vd/1cf8zdb1721f4z5rjggy8bp40000gn/T/go-build1872793839/b001/wasmedge.test
  Reason: tried: '/System/Volumes/Preboot/Cryptexes/OS@rpath/libwasmedge.0.dylib' (no such file), '/usr/local/lib/libwasmedge.0.dylib' (no such file), '/usr/lib/libwasmedge.0.dylib' (no such file, not in dyld cache)
signal: abort trap
FAIL	github.com/tetratelabs/wazero/internal/integration_test/vs/wasmedge	0.240s
FAIL
```

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -154,7 +154,7 @@ jobs:
         # The version here is coupled to internal/integration_test/go.mod, but it
         # isn't always the same as sometimes the Go layer has a broken release.
         env:
-          WASMEDGE_VERSION: 0.11.1
+          WASMEDGE_VERSION: 0.11.2
 
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
verified CI can still use wasmedge even if I can't locally

```
BenchmarkAllocation/Call-16
wazero-compiler 5933.16   ns/op
wasmedge        167909.74 ns/op
BenchmarkFactorial/Call-16
wazero-compiler 941.73   ns/op
wasmedge        14413.08 ns/op
wazero-compiler 115.78  ns/op
wasmedge        2421.91 ns/op
Benchmark/memory.i32/Call-16
wazero-compiler 372.47  ns/op
wasmedge        9234.57 ns/op
Benchmark/memory.i64/Call-16
wazero-compiler 328.78  ns/op
wasmedge        9515.26 ns/op
```

Issue #1036, opened 2023-01-15
Pass wasi-testsuite


codefromthecrypt:
We should pass [wasi-testsuite](https://github.com/WebAssembly/wasi-testsuite) prior to 1.0. This includes the following:

- [x] add a wazero adapter on a fork of wasi-testsuite https://github.com/tetratelabs/wasi-testsuite/tree/wazero
- [x] change the adapter or wazero until all tests pass, or exceptions understood
- [x] add CI in wazero to run tests using that adapter
  - [x] ensure setup steps happen https://github.com/WebAssembly/wasi-testsuite/issues/43
- [ ] raise a pull request to wasi-testsuite to include wazero by default

codefromthecrypt:
Made a shared fork including @achille-roussel's start and a small glitch fix. The wrapper syntax will change again once I get how mounts are expected to work working.

However, the more important thing is the failing tests:
```bash
$ python3 test-runner/wasi_test_runner.py  -t ./tests/assemblyscript/testsuite/  ./tests/c/testsuite/ -r adapters/wazero.sh 
```

- [ ] proc_exit-failure
- [ ] lseek
- [ ] stat-dev-ino
- [ ] fdopendir-with-access
- [ ] pwrite-with-access

Note: we were earlier told that the "rights" system was deprecated and not to be implemented as it was removed from wasi. This may or may not be related to the failures. I'll look into each of the above and tick off or explain.

codefromthecrypt:
see https://github.com/WebAssembly/wasi-testsuite/issues/43 on proc_exit-failure

codefromthecrypt:
proc_exit-failure wasn't a real failure. it was about drift so I reworded the upstream issue

codefromthecrypt:
To move forward we need to implement at least `fd_tell` and `pd_pwrite` as well add inodes to the stat results

codefromthecrypt:
next step is to implement io.WriterAt for pwrite. This is supported in windows and unix, so we should be ok. The lack of support in embed:fs is moot as embed:fs is read-only anyway. once this is weaved in, we should be able to pass the ` pwrite-with-access` test

codefromthecrypt:
I will next Writer/ReaderAt stuff from WASI so that it is defined at the syscallfs layer (re-used for gojs this way, but also saner place to test it). after that, backfill integration tests for embed:fs, and finally implement pwrite in WASI and equiv in gojs

codefromthecrypt:
[...]

PR #1098, opened 2023-02-05
wasi: skips fd_readdir until fixed upstream

codefromthecrypt:
The current fd_readdir test is invalid as it requires dot and dot-dot entries, which both aren't required by POSIX nor returned by go. Once this is addressed upstream, we can remove the skip.

See https://github.com/WebAssembly/wasi-testsuite/issues/52
See #1036

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -282,7 +282,7 @@ jobs:
       - name: Initialize Python environment
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10' # latest version of python 3
+          python-version: '3.11' # latest version of python 3
           cache: pip
 
       - name: Install dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yaml
```diff
@@ -297,18 +297,13 @@ jobs:
           toolchain: stable
           target: wasm32-wasi
 
-      # TODO: Remove after https://github.com/WebAssembly/wasi-testsuite/issues/49
```
codefromthecrypt, line 300: oops this PR isn't merged, yet.. I'll revert https://github.com/WebAssembly/wasi-testsuite/pull/54

codefromthecrypt:
the testsuite repo is overlapping the wazero mount so not finding the exclude. I'll move it to the testsuite fork..

codefromthecrypt:
ok finally works! technically, we aren't losing any coverage because we have a lot of tests around `fd_readdir` including various cookie positions, due to the work in support of PDFium.

```
Suite: WASI Rust tests
  Total: 11
  Passed:  7
  Failed:  3
  Skipped: 1
```

codefromthecrypt:
specifically, we have not stat and entries tests for zig, clang, rust and go (GOOS=js) in our normal test tree, plus the zig integration tests, and when tinygo implements readdir we'll have those as well.

mathetake (approved):

codefromthecrypt:
so, we are down to 3 legit failures from wasi-testsuite, now. We promised to pass all or have a very good reason for why not, so I think we're on track for this commitment by next version.

PR #1099, opened 2023-02-05
cli: adds an example Dockerfile

codefromthecrypt:
It is important for understanding the recent trends in Docker and WebAssembly to see what a Dockerfile looks like in wazero right now. By seeing what they are now, it is easier to understand what's different.

For example, things that natively support wasm right now can take off the 4.5MB our binary costs, as well hide some of the configuration you can see, such as coercing arguments or ENV.

That said, wasm isn't quite prime time in Docker, I suspect some will be happy to see what things look like. This will get even easier in March when we release 1.0, as we'll have formal packages and such.

### DIFF
--- a/cmd/wazero/Dockerfile
+++ b/cmd/wazero/Dockerfile
@@ -13,4 +13,4 @@ RUN git clone --depth 1 https://github.com/tetratelabs/wazero.git \
 FROM scratch
 COPY --from=build /build/wazero/wazero /bin/wazero
 
-ENTRYPOINT ["/bin/wazero", "run", "-cachedir=/cache", "-hostlogging=filesystem", "-mount=.:/"]
+ENTRYPOINT ["/bin/wazero", "run", "-cachedir=/cache", "-mount=.:/"]
--- a/cmd/wazero/Dockerfile
+++ b/cmd/wazero/Dockerfile
@@ -1,0 +1,16 @@
+# This is an example Dockerfile used to show a way to integrate wasm.
+FROM golang:1.20-alpine AS build
+
+RUN apk add --no-cache git
+
+WORKDIR /build
+
+# wazero doesn't publish a Docker image or binary, yet, so build on on demand.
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
found a [docker wasm example](https://github.com/redhat-et/wasm-demo-app) that isn't pinned to wasmedge (tested with wasmtime). While it doesn't do much, it can help tell the story of what is different normal docker/podman containers vs wasm etc. Thanks @font!

```diff
--- a/cmd/wazero/Dockerfile
+++ b/cmd/wazero/Dockerfile
@@ -9,8 +9,15 @@ WORKDIR /build
 RUN git clone --depth 1 https://github.com/tetratelabs/wazero.git \
   && (cd wazero; go build -o wazero -ldflags "-w -s" ./cmd/wazero)
 
+# we can't FROM this layer as it is wasm32 arch and we aren't!
+FROM quay.io/ifont/wasm-demo-app as wasm
+
 # wazero has no dependencies, so it can run on scratch
 FROM scratch
 COPY --from=build /build/wazero/wazero /bin/wazero
 
 ENTRYPOINT ["/bin/wazero", "run", "-cachedir=/cache", "-mount=.:/"]
+
+# a native-wasm image has only the CMD part
+COPY --from=wasm /wasm-demo-app.wasm /wasm-demo-app.wasm
+CMD ["/wasm-demo-app.wasm"]
```

codefromthecrypt:
A better example would be something that uses ENV and variables. Native docker/podman etc integration has a nice thing as it hides the CLI-specific means to map env variables. For example you can use the normal `ENV` directive which passes directly to the runtime, as opposed to having to coerce them into args to the runtime CLI (in our case `-env` flag). This benefit is less obvious until you have something that needs it, such as python. It will be nice when there's a portable python image that works as it would showcase this (e.g. for PYTHONHOME etc).

mathetake (approved):

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/1117 for automatic env propagation

PR #1100, opened 2023-02-06
wasi: implements filestatset

mathetake:
Part of #1036. With this, only two cases are left to pass 
> Test suites: 1 failed, 0 total
Tests:       2 failed, 8 passed, 1 skipped, 1 total


codefromthecrypt (approved):
cool work, and again great notes.


thoughts but not specific to this change: We'll need focus time maybe next week to see what we can do about portability. Go is really great, and using os.X vs syscall.X gives us a lot of interop and free bug fixes. the current wasi is what's used and there's some curiosity what the next will want to work. it may make sense to add a temporary impl of what's being called snapshot-02 as well backfill tests for GOOS=js (no priority), then see what sort of file abstraction makes sense w/o breaking all three, and without causing us to re-implement go's os-syscall layer for every platform that might be used.

PR #1101, opened 2023-02-06
wasi: makes fd_advise noop, instead of NoSys

mathetake:
part of #1036. `fd_advise.rs` is still failing due to the unimplemented `fd_allocate`. https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/fd_advise.rs#L44-L50


codefromthecrypt (approved):
this is excellent.

note for me: polish later as we aren't consistent with logging flags and they help to print because sometimes ordinals don't match. (I think we found this between some inspections of snapshot01 vs 02)

PR #1107, opened 2023-02-08
example: concurrent instantiation per Goroutine

### LINKED ISSUE #1103: Add example to instantiate  Wasm instances concurrently on single `wazero.Runtime`
Unfortunately I was unable to find any information for thread safety guarantees of the engine. 

Is it safe to use a single engine from multiple go routines under the assumption that all called wasm modules are thread-safe itself? E.g. every go routine has it's own distinct module that it's execting.

Thanks for any information about this topic and for this really awesome project 🙏 .

mathetake:
closes #1103


codefromthecrypt (approved):
Looks good, but I wonder if we shouldn't merge this with the compilation cache? if we put the cache into the foreground here, we can "kill two birds with one stone"

mathetake:
well maybe I won't merge it as I wanted to clarify the purpose of each example! (not strong opinion tough :D )

codefromthecrypt:
works for me, thanks!

Issue #422, opened 2022-03-31
Add option to limit CPU usage


codefromthecrypt:
split from #417 requested by @inkeliz

> Also, would be interesting to limit the CPU usage, but I think it's hard to track the CPU usage, that could be more useful when WASM supports thread directly.

### Prior art
* wasmtime re-implemented a Lucet feature similar to this https://github.com/bytecodealliance/wasmtime/pull/2611
* any credit/balance system ex https://httpwg.org/specs/rfc7540.html#WINDOW_UPDATE

### Related

#421 could be a similar or the same implementation, as if you have a concept of a runtime-wide credit you can deplete it to stop all commands, though what happens when out of credit is likely different (exit instead of pause)

codefromthecrypt:
here's something from ewasm that may be relevant https://ewasm.readthedocs.io/en/mkdocs/determining_wasm_gas_costs/

codefromthecrypt:
@JanFalkin do you mind commenting on which metering/limiting features you need and any links if a standard practice?

JanFalkin:
@codefromthecrypt Our need was for actual billing for our content-fabric.  Things we serve up and the effort to do so as a function of bitcode.  Ultimately I would like to know cpu and memory usage of the bitcode itself.  I was hoping just for any type of metering even if they were pseudo cycles that had a basis in reality.  This was one of the main reasons I was heading for Wasmer, as they have metering built in at an opcode level.  So you can submit a list of "costs" per wasm opcode and Wasmer will track this as a measure of work done.  I did not look that carefully at whether Wasmer was using a Cranelift feature or it had implemented it's own opcode metering.  As for limiting execution, it would be very interesting, but it was not our must have at release features.

pims:
👋 , having used the fuel feature in Wasmtime linked above, that's also something that I would love to be able to use in wazero. The main benefits would:

1. guarantee that a process will eventually stop – when running out of fuel
2. it could provide the necessary hook points to yield to the host to be able to interrupt execution based on external signals.

codefromthecrypt:
thanks @pims I can see some related pull requests on wasmtime, and also they had another impl iirc also. sounds like the status quo "fuel" one is good enough. there are differences between this and wasmer, and also its inspiration. I wonder if anyone has it speced or this will be the only thing they decide not to make into a new wasm/wasi spec? :D

codefromthecrypt:
[...]

PR #1108, opened 2023-02-08
Support for context Cancelation and Timeout in func executions

### LINKED ISSUE #509: Consider integration of go's cancel context
from @sam-at-luther on #421

>From the go docs, it's not immediately clear if canceling the passed ctx will stop execution. From this discussion I gather that it does not (I have not tried it), and that the ctx is more for tracing purposes--is that right? 

Indeed a cancel context only works if you pay attention to it.

Ex. reverseproxy.go in the main Go codebase guards its while loop on canceled, so it should stop abrupt if canceled. 
```go
	for {
		nr, rerr := src.Read(buf)
		if rerr != nil && rerr != io.EOF && rerr != context.Canceled {
			p.logf("httputil: ReverseProxy read error during body copy: %v", rerr)
		}
```

Most code I've seen in go doesn't pay attention to context.Canceled and we don't yet. There are some places it might be handy, ex in bulk memory (if really big) or a large I/O in WASI. It could also be the case in some loop forms we will need something like this to stop it from going forever, or some other mechanism to cancel (signal etc)

mathetake:
Closes #509 and #422. The latter is slightly different, but I believe almost all use cases will be covered by this since
in any way exact CPU usage measurement is impossible. Timeout and Cancelation can effectively be used by
users to limit the execution times approximately. That is exactly the same as the concept of "fuel" where fuel is an abstractive thing and *no one can interpret* how 1 fuel corresponds to exactly how much CPU cycles are used across multiple platforms.

Notably, this adds `WithCloseOnContextDone` option for RuntimeConfig as below:
```go
// WithCloseOnContextDone ensures the executions of functions to be closed under one of the following circumstances:
//
// 	- context.Context passed to the Call method of api.Function is canceled during execution. (i.e. ctx by context.WithCancel)
// 	- context.Context passed to the Call method of api.Function reaches timeout during execution. (i.e. ctx by context.WithTimeout or context.WithDeadline)
// 	- Close or CloseWithExitCode of api.Module is explicitly called during execution.
//
// This is especially useful when one wants to run untrusted Wasm binaries since otherwise, any invocation of
// api.Function can potentially block the corresponding Goroutine forever. Moreover, it might block the
// entire underlying OS thread which runs the api.Function call. See "Why it's safe to execute runtime-generated
// machine codes against async Goroutine preemption" section in internal/engine/compiler/RATIONALE.md for detail.
//
// Note that this comes with a bit of extra cost when enabled. The reason is that internally this forces
// interpreter and compiler runtimes to insert the periodical checks on the conditions above. For that reason,
// this is disabled by default.
//
// See examples in context_done_example_test.go for the end-to-end demonstrations.
//
// When the invocations of api.Function are closed due to this, sys.ExitError is raised to the callers and
// the api.Module from which the functions are derived is made closed.
WithCloseOnContextDone(bool) RuntimeConfig
```

The only use case I think this cannot cover is that relative fuel consumption - users might want to compare how much Wasm instructions are used among different module function invocations. It can be used, for example, to bill Wasm function owners depending on the consumed fuel * pre determined price per fuel. In that case "exact" CPU cycles are not necessarily needed. This use case is out-of-scope until anyone asks since the implementation will be much more complex and different than this context integration. In any case, that can be implemented later.


codefromthecrypt:
cc @pims!

internal/integration_test/engine/adhoc_test.go
```diff
@@ -85,8 +88,56 @@ var (
 	overflowWasm []byte
 	//go:embed testdata/global_extend.wasm
 	globalExtendWasm []byte
+	//go:embed testdata/infinite_loop.wasm
+	infiniteLoopWasm []byte
 )
 
+func testEnsureTerminationOnClose(t *testing.T, r wazero.Runtime) {
```
mathetake, line 95: so this is the beauty of what's possible now

ensure_termination_example_test.go
```diff
@@ -0,0 +1,130 @@
+package wazero_test
```
mathetake, line 1: added the examples

pims:
Gave this a try with the wasi-python example I had and it works well ⛽ !

mathetake:
@evacchi I think it would be great if you could review 👍

internal/wazeroir/compiler.go
```diff
@@ -472,6 +482,16 @@ operatorSwitch:
...
+		// Insert the exit code check on the loop header, which is the only necessary point in the function body
+		// to prevent infinite loop.
+		//
+		// Note that this is a little aggressive: this checks the exit code regardless the loop header is actually
+		// the loop. In other words, this checks even when no br/br_if/br_table instructions jumping to this loop
+		// exist. However, in reality, that shouldn't be an issue since such "noop" loop header will highly likely be
+		// optimized out by almost all guest language compilers which have the control flow optimization passes.
+		if c.ensureTermination {
+			c.emit(OperationBuiltinFunctionCheckExitCode{})
+		}
```
mathetake, line 494: this is the key

evacchi:
Will do!

config.go
```diff
@@ -128,6 +128,24 @@ type RuntimeConfig interface {
 	//	c, err := r.CompileModule(ctx, wasm)
 	//	customSections := c.CustomSections()
 	WithCustomSections(bool) RuntimeConfig
+
+	// WithEnsureTermination ensures the executions of functions to be terminated under one of the following circumstances:
+	//
+	// 	- context.Context passed to the Call method of api.Function is canceled during execution. (i.e. ctx by context.WithCancel)
+	// 	- context.Context passed to the Call method of api.Function reaches timeout during execution. (i.e. ctx by context.WithTimeout)
```
codefromthecrypt, line 135:
```suggestion
	// 	- context.Context passed to the Call method of api.Function reaches timeout during execution. (i.e. ctx by context.WithTimeout or context.WithDeadline)
```

config.go
```diff
@@ -128,6 +128,24 @@ type RuntimeConfig interface {
...
+	// api.Function can potentially block the corresponding Goroutine forever. Moreover, it might block the
+	// entire underlying OS thread which runs the api.Function call. See "Why it's safe to execute runtime-generated
+	// machine codes against async Goroutine preemption" section in internal/engine/compiler/RATIONALE.md for detail.
+	//
+	// Note that this comes with a bit of extra cost when enabled. The reason is that internally this forces
+	// interpreter and compiler runtimes to insert the periodical checks on the conditions above. For that reason,
+	// this is disabled by default.
+	//
+	// See examples in ensure_termination_example_tes.go for the end-to-end demonstrations.
+	WithEnsureTermination(bool) RuntimeConfig
```
codefromthecrypt, line 148: I think the word Context should be in the config name. cc @pims @Pryz for thoughts

```suggestion
	WithCloseOnContextCancel(bool) RuntimeConfig
```

Also we should document that regardless this ends up an ExitError

internal/engine/compiler/engine.go
```diff
@@ -941,6 +953,13 @@ entry:
 				ce.builtinFunctionFunctionListenerBefore(ce.ctx, callCtx.WithMemory(ce.memoryInstance), caller)
 			case builtinFunctionIndexFunctionListenerAfter:
 				ce.builtinFunctionFunctionListenerAfter(ce.ctx, callCtx.WithMemory(ce.memoryInstance), caller)
+			case builtinFunctionIndexCheckExitCode:
+				// Note: this operation must be done in Go, not native code. The reason is that
```
codefromthecrypt, line 957: great comment

sys/error.go
```diff
@@ -2,12 +2,23 @@
...
 
-// ExitError is returned to a caller of api.Function still running when
-// api.Module CloseWithExitCode was invoked. ExitCode zero value means success,
-// while any other value is an error.
+// These two special exit codes are reserved by wazero for context Cancel and Timeout integrations.
+// The assumption here is that well-behaving Wasm programs won't use these two exit codes.
+const (
+	// ExitCodeContextCanceled corresponds to context.Canceled and returned by ExitError.ExitCode in that case.
+	ExitCodeContextCanceled uint32 = 0xffffffff
+	// ExitCodeContextTimeout corresponds to context.DeadlineExceeded and returned by ExitError.ExitCode in that case.
```
codefromthecrypt, line 14: maybe rename
```suggestion
	// ExitCodeDeadlineExceeded corresponds to context.DeadlineExceeded and returned by ExitError.ExitCode in that case.
```

internal/wasm/call_context_test.go
```diff
@@ -288,3 +290,75 @@ func TestCallContext_CallDynamic(t *testing.T) {
 		require.False(t, ok, "expected no opened files")
 	})
 }
+
+func TestCallContext_SetExitCodeOnCanceledOrTimeout(t *testing.T) {
+	t.Run("timeout", func(t *testing.T) {
```
codefromthecrypt, line 295: main thing I'm concerned with is wrapped contexts, e.g. a WithValue over a Timeout or a WithCancel over a WithCancel. Mind adding tests and if there are any usage constraints as a result update the docs?

internal/wasm/call_context_test.go
```diff
@@ -288,3 +290,75 @@ func TestCallContext_CallDynamic(t *testing.T) {
...
+		ctx, cancel := context.WithTimeout(context.Background(), duration)
+		defer cancel()
+		done := cc.SetExitCodeOnCanceledOrTimeout(ctx)
+		time.Sleep(duration * 2)
+		defer done()
+
+		err := cc.FailIfClosed()
+		require.EqualError(t, err, "module \"test\" closed with context deadline exceeded")
+	})
+	t.Run("cancel", func(t *testing.T) {
```
codefromthecrypt, line 307: maybe also a redundant call to cancel() to make sure nothing panics or otherwise gets weird

api/wasm.go
```diff
@@ -318,6 +318,13 @@ type Function interface {
 	//
 	// To safely encode/decode params/results expressed as uint64, users are encouraged to
 	// use api.EncodeXXX or DecodeXXX functions. See the docs on api.ValueType.
+	//
+	// When RuntimeConfig.WithEnsureTermination is toggled, the invocation of this Call method is ensured to terminate
```
codefromthecrypt, line 322: let's not use different words for close, especially as they don't link to godoc. maybe try to rewrite this so it can link to Close
```suggestion
	// When RuntimeConfig.WithCloseOnContextCancel is toggled, the invocation of this Call method will ... 
```

internal/wasm/call_context_test.go
```diff
@@ -288,3 +290,75 @@ func TestCallContext_CallDynamic(t *testing.T) {
...
+		done := cc.SetExitCodeOnCanceledOrTimeout(ctx)
+		cancel()
+		defer done()
+
+		time.Sleep(time.Second)
+		err := cc.FailIfClosed()
+		require.EqualError(t, err, "module \"test\" closed with context canceled")
+	})
+
+	t.Run("cancel works", func(t *testing.T) {
```
codefromthecrypt, line 319: maybe add something for CancelCauseFunc as the cancel error can be arbitrary

codefromthecrypt (approved):
Excellent job! so many have asked about this, it is great to have a chance prior to 1.0!

Main thing I wonder is if this should be called `RuntimeConfig.WithCloseOnContextCancel` because I believe at the end of the day, all these things end up as cancelation causes. In the very least it should be obvious by name that this is about context integration because the other ways (e.g. someone calling module.Close()` already exist w/o this feature.

happy for other ideas, too. Besides the name bike shedding I only had some questions about edge cases, specifically to know if someone wraps a context things will still work.

codefromthecrypt:
or `RuntimeConfig.WithCloseOnCanceledContext` ...

codefromthecrypt:
@ncruces thoughts on this one?

codefromthecrypt:
@daixiang0 @ItalyPaleAle @yaron2 so this means wasm middleware can stop based on incoming context. I'm not sure if it is routine to use context-cancelation in dapr, but we can integrate it once this feature is cut if so.

internal/wasm/call_context_test.go
mathetake: that is the Go 1.20 feature, so won't add it here yet.

mathetake:
I will go with  `WithCloseOnContextDone` since it includes Timeout and Cancel and matches the meaning of Go's context.Context.Done

internal/engine/compiler/arch_amd64.s
```diff
@@ -2,8 +2,8 @@
...
-TEXT ·nativecall(SB),NOSPLIT|NOFRAME,$0-24
-        MOVQ ce+8(FP),R13                     // Load the address of *callEngine. into amd64ReservedRegisterForCallEngine.
-        MOVQ moduleInstanceAddress+16(FP),R12 // Load the address of *wasm.ModuleInstance into amd64CallingConventionModuleInstanceAddressRegister.
-        MOVQ codeSegment+0(FP),AX             // Load the address of native code.
-        JMP AX                                // Jump to native code.
+TEXT ·nativecall(SB), NOSPLIT|NOFRAME, $0-24
+	MOVQ ce+8(FP), R13                     // Load the address of *callEngine. into amd64ReservedRegisterForCallEngine.
+	MOVQ moduleInstanceAddress+16(FP), R12 // Load the address of *wasm.ModuleInstance into amd64CallingConventionModuleInstanceAddressRegister.
+	MOVQ codeSegment+0(FP), AX             // Load the address of native code.
+	JMP  AX                                // Jump to native code.
```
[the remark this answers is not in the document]
mathetake: sorry I recently setup the formatter for Plan 9 assembly https://github.com/klauspost/asmfmt and this is the result, somewhat intentionally :D

internal/wasm/engine.go
```diff
@@ -13,7 +13,7 @@ type Engine interface {
 	Close() (err error)
 
 	// CompileModule implements the same method as documented on wasm.Engine.
-	CompileModule(ctx context.Context, module *Module, listeners []experimental.FunctionListener) error
+	CompileModule(ctx context.Context, module *Module, listeners []experimental.FunctionListener, ensureTermination bool) error
```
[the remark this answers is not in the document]
mathetake: not only it doesn't make sense, but also that would cause cycle imports that is forbidden in Go!

internal/wasm/call_context_test.go
mathetake: in the implementation we are just doing <-context.Done(), and nothing above that. Wrapping context's Done is working is already ensured by Go's standard library test itests. Do you still want me to do that?

mathetake: https://github.com/tetratelabs/wazero/pull/1108/commits/6f360688011d008542d2a3649896fd8d25b666af added in any way!

codefromthecrypt: the main thinking was to prevent a change in impl breaking things. literally I was thinking if later someone decided to optimize things for fields etc. I think that was over-thinking, so clicking resolve. no action needed!

ncruces:
@codefromthecrypt I made a quick pass over the code, but not enough to make detailed comments.
The feature seems useful, well thought out, and implemented.

From my experience, I'm unlikely to use it with SQLite (outside of tests?) because, although calling `exit(1)` randomly (this seems equivalent) is _not_ supposed to lead to data corruption, it is still a pretty heavy handed way of doing it with a trusted library, that might have other, gentler, ways to handle interrupts.

I'm likely to use it with `dcraw`. I'm already using contexts in that environment (web server), and cancelling seems appropriate and not nefarious. It mimics killing an `exec.Cmd` subprocess on cancellation, which is a nice ability to have.

mathetake:
Thank you folks, enjoy! 🍻

pims:
Thank YOU @mathetake 🙇🏻

yaron2:
> @daixiang0 @ItalyPaleAle @yaron2 so this means wasm middleware can stop based on incoming context. I'm not sure if it is routine to use context-cancelation in dapr, but we can integrate it once this feature is cut if so.

We're building context cancellations for most components now so this would certainly become useful.

Issue #580, opened 2022-05-19
Compiler(amd64): return false on SSE4 unsupported CPUs from platform.CompilerSupported


mathetake:
**Is your feature request related to a problem? Please describe.**

We've already used SSE4.x features in floating-point conversions and SIMD implementations. However, some _fairly old_ amd64 CPUs do not have support for these features.

Practically speaking, SSE4.1 features was introduced [15-years ago](https://en.wikipedia.org/wiki/SSE4), and actually, the SIMD proposal has been merged with the baseline being SSE4 on Intel CPUs(https://github.com/WebAssembly/simd/pull/201#issuecomment-600240445). It seems really hard (maybe impossible) to implement without the feature (https://github.com/bytecodealliance/wasmtime/issues/3809, https://github.com/bytecodealliance/wasm-micro-runtime/issues/1022,https://github.com/wasmerio/wasmer/pull/2775, https://github.com/wasmerio/wasmer/pull/2775).

It's nice if we could have support for these CPUs, but given that almost all amd64 CPUs have SSE4, that's not our priority and revisit what we can do.

**Describe the solution you'd like**

Get the available features by [`CPUID`](https://en.wikipedia.org/wiki/CPUID) instruction, and raise a compilation error if the features are necessary while not available in the hardware.

**Describe alternatives you've considered**

N/A

**Additional context**

Go has introduced `GOAMD64` flag as of [1.18](https://tip.golang.org/doc/go1.18), and we could leverage the flag and check if the flag has >=v2.

mathetake:
ps I would like to put this limitation only on SIMD instructions, so I will identity where SSE4.1 features are used in non-SIMD instructions (one place I can think of is [floating conversions](https://github.com/tetratelabs/wazero/blob/b3fc76ed6eb40b689f0383d23eb574c653181894/internal/engine/compiler/impl_amd64.go#L2106-L2109).

mathetake:
assigning @evacchi

PR #1112, opened 2023-02-09
Fix #1111.

### LINKED ISSUE #1111: amd64 compiler bug
**Describe the bug**

While developing [`github.com/ncruces/go-sqlite3`](https://github.com/ncruces/go-sqlite3) I found an issue where a unit test was breaking on my windows machine (but not on GitHub CI).

Basically, running the same code produces a different result on that computer, failing the test. I found that switching to the interpreter fixes the issue on that computer. On other platforms, on GitHub CI, and on a friend's Windows computer, I can't repro the bug. The bug is consistent (always the same output), and survived a computer reboot.

**To Reproduce**

On the affected computer:

```sh
git clone https://github.com/ncruces/wazero-sqlite3-bug.git
cd wazero-sqlite3-bug
go run ./cmd
```

Expected output:
```
[2023/02/08 00:00:00] compiled "2.500000", interpreted "2.500000", want "2.500000"
```

Actual output in affected computer:
```
[2023/02/08 00:00:00] compiled "100000000000000000.000000", interpreted "2.500000", want "2.500000"
```

- go version go1.20 windows/amd64
- go version go1.19.4 windows/amd64
- wazero Version: v1.0.0-pre.8
- Host architecture: amd64
- Runtime mode: compiler
- Windows: 10.0.19045.2546

**Additional context**

The repo [above](https://github.com/ncruces/wazero-sqlite3-bug.git) runs the test in a [matrix of OS versions](https://github.com/ncruces/wazero-sqlite3-bug/blob/main/.github/workflows/go.yml) (all the ones available on GitHub), and can't trigger the issue.

The [command](https://github.com/ncruces/wazero-sqlite3-bug/blob/main/cmd/main.go) that runs is not actually the failing unit test. It's just an attempt at a simpler case. The tests that fail are [these two](https://github.com/ncruces/wazero-sqlite3-bug/blob/main/stmt_test.go#L203-L208). Also float to text conversions gone bad. I've inspected the code that converts float to text in [SQLite](https://github.com/sqlite/sqlite/blob/master/src/printf.c) and nothing in particular comes up.

Not sure what to make of this…

ncruces:
I know this can't go in because of the `cpuid` dependency, but maybe it helps.

I can confirm all tests pass on Windows+WSL on the affected machine, and on Windows+macOS+Linux on other machines.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,3 +11,5 @@ retract (
 	v1.0.0-beta.2
 	v1.0.0-beta.1
 )
+
+require github.com/intel-go/cpuid v0.0.0-20220614022739-219e067757cb
--- a/go.mod
+++ b/go.mod
@@ -12,4 +12,4 @@ retract (
 	v1.0.0-beta.1
 )
 
-require github.com/intel-go/cpuid v0.0.0-20220614022739-219e067757cb
+require github.com/zcalusic/sysinfo v0.9.5
--- a/go.sum
+++ b/go.sum
@@ -1,0 +1,2 @@
+github.com/intel-go/cpuid v0.0.0-20220614022739-219e067757cb h1:Fg0Y/RDZ6UPwl3o7/IzPbneDq8g9+gH6DPs42KFUsy8=
+github.com/intel-go/cpuid v0.0.0-20220614022739-219e067757cb/go.mod h1:RmeVYf9XrPRbRc3XIx0gLYA8qOFvNoPOfaEZduRlEp4=
--- a/go.sum
+++ b/go.sum
@@ -1,2 +1,2 @@
-github.com/intel-go/cpuid v0.0.0-20220614022739-219e067757cb h1:Fg0Y/RDZ6UPwl3o7/IzPbneDq8g9+gH6DPs42KFUsy8=
-github.com/intel-go/cpuid v0.0.0-20220614022739-219e067757cb/go.mod h1:RmeVYf9XrPRbRc3XIx0gLYA8qOFvNoPOfaEZduRlEp4=
+github.com/zcalusic/sysinfo v0.9.5 h1:ivoHyj9aIAYkwzo1+8QgJ5s4oeE6Etx9FmZtqa4wJjQ=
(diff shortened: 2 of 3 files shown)


codefromthecrypt:
thanks, @ncruces I think considering we wrote our own assembler, I think we can find a way to get similar code needed here (with unit tests). e.g. in our assembly we have compileXXX functions that could add instructions needed like https://github.com/aregm/cpuid/blob/master/cpuidlow_amd64.s

Thanks for making the change. To proceed I think ideally we can make the build fail first, then we can adapt in logic we need until it passes.

codefromthecrypt:
@ncruces thanks again for the help.

Basically you are saying that since we don't compile for other archs (only the current one), detecting the host cpuid is going to be correct for all module compilation.

So, the idea is that we could have detection implemented as a _amd64 file, for example leveraging a `.s` file as the dep is doing. We could store that as a constant in `amd64Compiler`, though for unit tests maybe a field is fine also.

As you also mentioned, we could just generally avoid the problematic instructions, but I don't personally know the impact of that. It is possible that impact could be evaluated with benchmarks, also, especially if we allow toggle of this for testing, the test even where `LZCNTL` and `TZCNTL` are fine can try wasm that would compile down to using them, without using them.

I'm going to hand this off to @evacchi to noodle on. If I mis-restated anything above lemme know!

ncruces:
It needs to be a field, not a `const`. Or, even, you run the check every single time.

Basically by compile-time I mean load-time, when you AOT compile the WASM binary into ASM. Then, at “WASM-runtime” you can use the instruction without checking since it's the same CPU.

This means you can freely use instructions where just making the check negates the benefit.

mathetake:
cool, let's implement `cpuid.HasExtraFeature(cpuid.ABM)` by ourselves. Assigning @evacchi

ncruces:
New implementation, I expect, should be easier to follow.

mathetake:
with the CPUID, https://github.com/tetratelabs/wazero/issues/580 this could also be resolved

PR #1115, opened 2023-02-10
cli: add an option to force interpreter

yamt:
Background: I wanted to evaluate the interpreter like: https://github.com/yamt/toywasm/blob/5b7f8a870a2b8a7600330f0421b14bf01e2249d8/benchmark/ffmpeg.sh#L58-L61
But I couldn't find a convenient way.

Signed-off-by: YAMAMOTO Takashi <imuwoto@gmail.com>

### DIFF
--- a/cmd/wazero/wazero.go
+++ b/cmd/wazero/wazero.go
@@ -114,6 +114,9 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
 	var help bool
 	flags.BoolVar(&help, "h", false, "print usage")
 
+	var interp bool
+	flags.BoolVar(&interp, "interp", false, "force interpreter")
+
 	var envs sliceFlag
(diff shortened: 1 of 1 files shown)


mathetake (approved):
cool indeed this could be helpful in the future trouble shooting as well.

cmd/wazero/wazero.go
```diff
@@ -114,6 +114,9 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
 	var help bool
 	flags.BoolVar(&help, "h", false, "print usage")
 
+	var interp bool
+	flags.BoolVar(&interp, "interp", false, "force interpreter")
```
codefromthecrypt, line 118: interp is rather cryptic. I'll raise a PR to change this to something less!

codefromthecrypt: https://github.com/tetratelabs/wazero/pull/1116

codefromthecrypt:
thanks @yamt, this is a good idea

PR #1118, opened 2023-02-10
add support for querying cpuid

### LINKED ISSUE #1111: amd64 compiler bug

evacchi:
fixes #1111, adds setup for fixing #580 

While I haven't tested this on a proper x86 machine (but I do have a dusty core duo laptop which may really be the thing I need, if I manage to turn it on :D) I have sketched a solution by putting together everyone's feedback (including @ncruces' 🚀). In the meantime I have also discovered that `GOARCH=amd64` has the ABM feature flag off on Rosetta2.

- this is in a new `cupid_amd64.{s,go}` because I felt it would pollute too much `impl` and `arch` didn't feel like the right place (that's for runtime)
- For now I have only added the constants that we use and/or plan to use (e.g. SSE3, 4.1, 4.2), and only the related bitmaps
- you still have to carefully pick the right set of flags because there is no proper type checking in place (all are uint64). Suggestions welcome for better ergonomics/idioms

Values are kept in a struct that is local to the compiler instance. This is re-inited at for each `amd64Compiler` -- TBF it may be created once and then copied over when necessary. I wanted to keep it local to the compiler instance, instead of global, to make sure we can properly test it.

Speaking of testing: the compiler test right now is kind of silly because it checks for the length of the generated byte array. The one reason is that I couldn't figure out how to decode or inspect the list of opcodes the right way. This is the main reason why this is draft.

### DIFF
--- a/internal/engine/compiler/impl_amd64.go
+++ b/internal/engine/compiler/impl_amd64.go
@@ -8,7 +8,6 @@ import (
 	"bytes"
 	"fmt"
 	"math"
-	"runtime"
 
 	"github.com/tetratelabs/wazero/internal/asm"
 	"github.com/tetratelabs/wazero/internal/asm/amd64"
@@ -83,8 +82,9 @@ func (c *amd64Compiler) compileNOP() asm.Node {
 }
 
 type amd64Compiler struct {
-	assembler amd64.Assembler
-	ir        *wazeroir.CompilationResult
+	assembler   amd64.Assembler
+	ir          *wazeroir.CompilationResult
+	cpuFeatures platform.CpuFeatureFlags
 	// locationStack holds the state of wazeroir virtual stack.
 	// and each item is either placed in register or the actual memory stack.
 	locationStack *runtimeValueLocationStack
@@ -103,6 +103,7 @@ func newAmd64Compiler() compiler {
 	c := &amd64Compiler{
 		assembler:     amd64.NewAssembler(),
 		locationStack: newRuntimeValueLocationStack(),
+		cpuFeatures:   platform.CpuFeatures,
 	}
 	return c
 }
@@ -114,6 +115,7 @@ func (c *amd64Compiler) Init(ir *wazeroir.CompilationResult, withListener bool) 
 	*c = amd64Compiler{
 		labels:       map[string]*amd64LabelInfo{},
 		ir:           ir,
+		cpuFeatures:  c.cpuFeatures,
 		withListener: withListener,
 		currentLabel: wazeroir.EntrypointLabel,
 	}
@@ -1170,7 +1172,7 @@ func (c *amd64Compiler) compileClz(o *wazeroir.OperationClz) error {
 		return err
 	}
 
-	if runtime.GOOS != "darwin" && runtime.GOOS != "freebsd" {
+	if c.cpuFeatures.HasExtra(platform.CpuExtraFeatureABM) {
 		if o.Type == wazeroir.UnsignedInt32 {
 			c.assembler.CompileRegisterToRegister(amd64.LZCNTL, target.register, target.register)
 		} else {
@@ -1233,7 +1235,7 @@ func (c *amd64Compiler) compileCtz(o *wazeroir.OperationCtz) error {
 		return err
 	}
 
-	if runtime.GOOS != "darwin" && runtime.GOOS != "freebsd" {
+	if c.cpuFeatures.HasExtra(platform.CpuExtraFeatureABM) {
 		if o.Type == wazeroir.UnsignedInt32 {
 			c.assembler.CompileRegisterToRegister(amd64.TZCNTL, target.register, target.register)
 		} else {
--- a/internal/engine/compiler/impl_amd64_test.go
+++ b/internal/engine/compiler/impl_amd64_test.go
@@ -1,11 +1,13 @@
 package compiler
 
 import (
+	"encoding/hex"
 	"testing"
 	"unsafe"
 
 	"github.com/tetratelabs/wazero/internal/asm"
 	"github.com/tetratelabs/wazero/internal/asm/amd64"
+	"github.com/tetratelabs/wazero/internal/platform"
 	"github.com/tetratelabs/wazero/internal/testing/require"
 	"github.com/tetratelabs/wazero/internal/wasm"
 	"github.com/tetratelabs/wazero/internal/wazeroir"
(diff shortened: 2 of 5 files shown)


ncruces:
I tried this on the affected machine and I can report that it worked:
- [x] `wazero` tests pass
- [x] [`wazero-sqlite3-bug`](https://github.com/ncruces/wazero-sqlite3-bug) tests too

evacchi:
I am not 100% sure I am doing the right thing here 😬 

in your PR you are and-ing 0x0400_0000, in this PR it's 0x20. For some reason on my (dusty old and excruciatingly slow) machine this still prints "100000000..." instead of "2.500..."; and it only works with 0x0400_0000. 

However, according to the spec the 5th bit on ECX should be 0x20 🤪

ncruces:
> I am not 100% sure I am doing the right thing here 😬
> 
> in your PR you are and-ing 0x0400_0000, in this PR it's 0x20. For some reason on my (dusty old and excruciatingly slow) machine this still prints "100000000..." instead of "2.500..."; and it only works with 0x0400_0000.
> 
> However, according to the spec the 5th bit on ECX should be 0x20 🤪

OK, then I need to recheck. Did the tests pass for me because of caching, or because I have both `0x20` and `0x0400_0000`?

 I did notice that. From the docs, I assumed it's the 5th bit from the left: `((1<<31) >> 5) == 0x0400_0000`, which did the right thing on my machine. You assumed the opposite `(1 << 5) == 0x20`.

I thing we need to figure it out from other implementations.

ncruces:
I think, based on the Intel "sponsored" library, the correct implementation is to first get `leaf0x80000000` (which I didn't do, to make sure you _can_ get `leaf0x80000001`) then get `leaf0x80000001`.
https://github.com/aregm/cpuid/blob/219e067757cbc0ddb71908b19826ddca4177987d/cpuid_amd64.go#L218-L229

Then compare `extraFeatureFlags` with `0x20`.
This is `HasExtraFeature`: https://github.com/aregm/cpuid/blob/master/cpuid.go#L101-L104
This is `ABM`: https://github.com/aregm/cpuid/blob/219e067757cbc0ddb71908b19826ddca4177987d/cpuid.go#L462-L467

So it seems you're right, and I messed up, but maybe your excruciatingly slow computer needs to check `leaf0x80000000` before doing `leaf0x80000001`?

internal/engine/compiler/cpuid_amd64.s
```diff
@@ -0,0 +1,14 @@
+#include "textflag.h"
+
+// lifted from github.com/intel-go/cpuid and src/internal/cpu/cpu_x86.s
+// func cpuid(arg1, arg2 uint32) (eax, ebx, ecx, edx uint32)
+TEXT ·cpuid(SB), NOSPLIT, $0-24
+	MOVL arg1+0(FP), AX
+	MOVL arg2+4(FP), CX
+	CPUID
+	MOVL AX, eax+8(FP)
+	MOVL BX, ebx+12(FP)
```
mathetake, line 10: do we need AX and BX register values? I think we could just delete them if you don't use.
```suggestion

TEXT ·cpuid(SB), NOSPLIT, $0-16
	MOVL arg1+0(FP), AX
	MOVL arg2+4(FP), CX
	CPUID
```

internal/engine/compiler/cpuid_amd64.go
```diff
@@ -0,0 +1,43 @@
+package compiler
```
mathetake, line 1: can we move this to under platform/**? Then I could reuse it later on the new backend

internal/engine/compiler/impl_amd64.go
```diff
@@ -103,6 +103,7 @@ func newAmd64Compiler() compiler {
 	c := &amd64Compiler{
 		assembler:     amd64.NewAssembler(),
 		locationStack: newRuntimeValueLocationStack(),
+		cpuFeatures:   loadCpuFeatureFlags(),
```
mathetake, line 106: I think we should get this in global variable (e.g. `platform.CPUFeature`) and set it in `init` function of platform package. That's because the CPU features must be same across the process and we don't need to get multiple times

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -441,6 +441,61 @@ func TestAmd64Compiler_preventCrossedTargetdRegisters(t *testing.T) {
...
+			newCompiler := func() compiler {
+				c := newCompiler().(*amd64Compiler)
+				// override auto-detected CPU features with the test case
+				c.cpuFeatures = tt.cpuFeatures
+				return c
+			}
+
+			compiler := env.requireNewCompiler(t, newCompiler, nil)
+			err := compiler.compilePreamble()
+			require.NoError(t, err)
```
mathetake, line 483: I don' think these two are needed for this test.

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -441,6 +441,61 @@ func TestAmd64Compiler_preventCrossedTargetdRegisters(t *testing.T) {
...
+
+			err = compiler.compileClz(&wazeroir.OperationClz{Type: wazeroir.UnsignedInt64})
+			require.NoError(t, err)
+			//
+			err = compiler.compileReturnFunction()
+			require.NoError(t, err)
```
mathetake, line 489: same

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -441,6 +441,61 @@ func TestAmd64Compiler_preventCrossedTargetdRegisters(t *testing.T) {
...
+// by overriding the field in the corresponding struct.
+func TestAmd64Compiler_ensureClz_ABM(t *testing.T) {
+	tests := []struct {
+		name        string
+		cpuFeatures cpuFeatureFlags
+		expectedLen int
+	}{
+		{
+			name:        "with ABM",
+			expectedLen: 140,
```
mathetake, line 455: rather than length, let's check the actual hex just like we've done in asm/amd64 package tests

internal/engine/compiler/impl_amd64_test.go
```diff
@@ -441,6 +441,61 @@ func TestAmd64Compiler_preventCrossedTargetdRegisters(t *testing.T) {
 	}
 }
 
+// Relates to #1111: older AMD64 CPUs do not support the LZCNT instruction
+// CPUID should be used instead. We simulate presence/absence of the feature
+// by overriding the field in the corresponding struct.
+func TestAmd64Compiler_ensureClz_ABM(t *testing.T) {
```
mathetake, line 447: not only Clz, but also Ctz requires tests

internal/engine/compiler/cpuid_amd64.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+}
+
+// cpuid exposes the CPUID instruction to the Go layer (https://www.amd.com/system/files/TechDocs/25481.pdf)
+// implemented in impl_amd64.s
+func cpuid(arg1, arg2 uint32) (eax, ebx, ecx, edx uint32)
+
+// cpuidAsBitmap combines the result of invoking cpuid to uint64 bitmap
+func cpuidAsBitmap(arg1, arg2 uint32) uint64 {
+	_ /* eax */, _ /* ebx */, ecx, edx := cpuid(1, 0)
```
ncruces, line 31: Shouldn't this be `cpuid(arg1, arg2)`?

internal/engine/compiler/cpuid_amd64.s
ncruces: Other feature queries might use all 4 (e.g. BMI uses BX).

ncruces: Other feature queries might use all 4 (e.g. BMI uses BX).

ncruces: Also, in theory (maybe not in practice) you should call `cpuid(0, 0)` and inspect `EAX` to figure out if you can trust `cpuid(1, 0)`, and similarly call `cpuid(0x8000_0000, 0)` before `cpuid(0x8000_0001, 0)`.

https://www.amd.com/system/files/TechDocs/25481.pdf
> The smallest function number of the standard function range is Fn0000_0000. The largest function number of the standard function range, for a particular implementation, is returned in CPUID
Fn0000_0000_EAX.
> The smallest function number of the extended function range is Fn8000_0000. The largest function number of the extended function range, for a particular implementation, is returned in CPUID
Fn8000_0000_EAX.

internal/engine/compiler/cpuid_amd64.go
ncruces: For what it's worth, tests pass with this change, on all my computers.

internal/engine/compiler/cpuid_amd64.s
mathetake: cool, thanks for the info!

internal/engine/compiler/cpuid_amd64.go
evacchi: 🙈 likely

internal/engine/compiler/impl_amd64_test.go
evacchi: I tried already, but it looks like the state is inconsistent without them.

without the preamble:

```
panic: runtime error: index out of range [18446744073709551615] with length 0 [recovered]
...
github.com/tetratelabs/wazero/internal/engine/compiler.(*runtimeValueLocationStack).pop(...)
```

internal/engine/compiler/impl_amd64_test.go
evacchi: without this (only `without_ABM`)

```
    require.go:312: expected no error, but was jump target must not be nil for relative JMP: JMP {<nil>}
        /Users/evacchi/Devel/github.com/tetratelabs/wazero/internal/engine/compiler/impl_amd64_test.go:492
```

internal/engine/compiler/cpuid_amd64.go
evacchi: confirmed on my old laptop too

internal/engine/compiler/impl_amd64_test.go
evacchi: so this I solve by pushing a constant i32

internal/engine/compiler/impl_amd64_test.go
evacchi: the jump that is generated in "without_ABM" needs a "next" instruction so, I am just adding a NOP

ncruces (approved):

internal/platform/cpuid_amd64.go
```diff
@@ -0,0 +1,73 @@
+package platform
+
+const (
+	CpuFeatureSSE3   = uint64(1)
```
mathetake, line 4: Let’s put the comment as much as possible on exported comets even its under internal. That helps us in the future

internal/platform/cpuid_amd64.go
```diff
@@ -0,0 +1,73 @@
...
+	CpuFeatureSSE4_2 = uint64(1) << 20
+)
+
+const (
+	CpuExtraFeatureABM = uint64(1) << 5 // Advanced Bit Manipulation
+)
+
+var CpuFeatures CpuFeatureFlags = loadCpuFeatureFlags()
+
+type CpuFeatureFlags interface {
```
mathetake, line 15: Same

mathetake (approved):
Good job!👏

codefromthecrypt:
OSS collaboration like this makes my day, thanks @ncruces @evacchi @mathetake!

PR #1119, opened 2023-02-13
Fixes races around {api.Module, wazero.Runtime}.CloseWithExitCode

mathetake:
This is a regression introduced in #1108 and hasn't been well covered in the unit tests with -race.
Thanks to @ckaznocha for reporting this!

### DIFF
--- a/internal/wasm/call_context.go
+++ b/internal/wasm/call_context.go
@@ -67,27 +67,29 @@ func (m *CallContext) FailIfClosed() (err error) {
 	return nil
 }
 
-// SetExitCodeOnCanceledOrTimeout take a context `ctx`, which might be a Cancel or Timeout context,
+// CloseModuleOnCanceledOrTimeout take a context `ctx`, which might be a Cancel or Timeout context,
 // and spawns the Goroutine to check the context is canceled ot deadline exceeded. If it reaches
 // one of the conditions, it sets the appropriate exit code.
 //
 // Callers of this function must invoke the returned context.CancelFunc to release the spawned Goroutine.
-func (m *CallContext) SetExitCodeOnCanceledOrTimeout(ctx context.Context) context.CancelFunc {
+func (m *CallContext) CloseModuleOnCanceledOrTimeout(ctx context.Context) context.CancelFunc {
 	goroutineDone, cancelFn := context.WithCancel(context.Background())
-	go m.setExitCodeOnCanceledOrTimeoutClosure(ctx, goroutineDone)()
+	go m.closeModuleOnCanceledOrTimeoutClosure(ctx, goroutineDone)()
 	return cancelFn
 }
 
-// setExitCodeOnCanceledOrTimeoutClosure is extracted from SetExitCodeOnCanceledOrTimeout for testing.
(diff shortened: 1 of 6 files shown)


internal/wasm/call_context.go
```diff
@@ -122,14 +124,24 @@ func (m *CallContext) Close(ctx context.Context) (err error) {
...
-func (m *CallContext) setExitCode(exitCode uint32) {
+// closeWithExitCode is the same as CloseWithExitCode besides this doesn't delete it from Store.moduleList.
+func (m *CallContext) closeWithExitCode(ctx context.Context, exitCode uint32) (err error) {
+	if !m.setExitCode(exitCode) {
+		return nil
+	}
+	return m.ensureResourcesClosed(ctx)
+}
+
+func (m *CallContext) setExitCode(exitCode uint32) bool {
```
mathetake, line 142: I accidentally deleted the boolean of the `swap` from `CompareAndSwapUint64` and dropped the check. so fixed it!

internal/wasm/call_context.go
```diff
@@ -122,14 +124,24 @@ func (m *CallContext) Close(ctx context.Context) (err error) {
 
 // CloseWithExitCode implements the same method as documented on api.Module.
 func (m *CallContext) CloseWithExitCode(ctx context.Context, exitCode uint32) (err error) {
-	m.setExitCode(exitCode)
+	if !m.setExitCode(exitCode) {
+		return nil
```
codefromthecrypt, line 128:
```suggestion
		return nil // not an error to have already closed
```

internal/wasm/call_context.go
```diff
@@ -122,14 +124,24 @@ func (m *CallContext) Close(ctx context.Context) (err error) {
...
+	}
 	_ = m.s.deleteModule(m.Name())
 	return m.ensureResourcesClosed(ctx)
 }
 
-func (m *CallContext) setExitCode(exitCode uint32) {
+// closeWithExitCode is the same as CloseWithExitCode besides this doesn't delete it from Store.moduleList.
+func (m *CallContext) closeWithExitCode(ctx context.Context, exitCode uint32) (err error) {
+	if !m.setExitCode(exitCode) {
+		return nil
```
codefromthecrypt, line 137:
```suggestion
		return nil // not an error to have already closed
```

internal/wasm/call_context_test.go
```diff
@@ -147,30 +148,46 @@ func TestCallContext_Close(t *testing.T) {
...
-		// We use side effects to determine if Close in fact called Context.Close (without repeating sys_test.go).
-		// One side effect of Context.Close is that it clears the openedFiles map. Verify our base case.
-		_, ok := fsCtx.LookupFile(3)
-		require.True(t, ok, "sysCtx.openedFiles was empty")
-
-		// Closing should not err.
-		require.NoError(t, m.Close(testCtx))
+				// Closing should not err.
+				if concurrent {
+					hammer.NewHammer(t, 100, 10).Run(func(name string) {
```
codefromthecrypt, line 171: nit: indentation is getting intense, but ok. if you see a cheap way to reduce this, by extracting a new top-level test, or a function do it.

codefromthecrypt (approved):
thanks for fixing this!

Issue #581, opened 2022-05-20
Add emulation tests with legacy AMD64 CPUs


mathetake:
Relevant to #580, we would like to make sure that our engines work well even on [the baseline x64 CPU](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels).

By default, SSE4 (meaning x86-64-v2) is available on QEMU, so it would be great I we could  add the job
in [commit.yaml](https://github.com/tetratelabs/wazero/blob/main/.github/workflows/commit.yaml) and [spectest.yaml](https://github.com/tetratelabs/wazero/blob/main/.github/workflows/spectest.yaml) along with other emulation tasks.

PR #1121, opened 2023-02-13
compiler(amd64): return false on SSE4 unsupported CPUs

### LINKED ISSUE #580: Compiler(amd64): return false on SSE4 unsupported CPUs from platform.CompilerSupported

evacchi:
I am not sure whether we should also add some test case to exercise this code path. Maybe enabling emulated tests as suggested in #581? Even my old Window laptop is newer and passes :D

Fixes #580

### DIFF
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -10,6 +10,9 @@ import (
 	"runtime"
 )
 
+// archRequirementsVerified is set by platform-specific init to true if the platform is supported
+var archRequirementsVerified = false
+
 // CompilerSupported is exported for tests and includes constraints here and also the assembler.
 func CompilerSupported() bool {
 	switch runtime.GOOS {
@@ -18,13 +21,7 @@ func CompilerSupported() bool {
 		return false
 	}
 
-	switch runtime.GOARCH {
-	case "amd64", "arm64":
-	default:
-		return false
-	}
-
-	return true
+	return archRequirementsVerified
 }
 
 // MmapCodeSegment copies the code into the executable region and returns the byte slice of the region.
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -11,7 +11,7 @@ import (
 )
 
 // archRequirementsVerified is set by platform-specific init to true if the platform is supported
-var archRequirementsVerified = false
+var archRequirementsVerified bool
 
 // CompilerSupported is exported for tests and includes constraints here and also the assembler.
 func CompilerSupported() bool {
--- a/internal/platform/platform_arm64.go
+++ b/internal/platform/platform_arm64.go
@@ -2,6 +2,6 @@ package platform
 
 // init verifies that the current CPU supports the required ARM64 features
 func init() {
-	// no further checks currently needed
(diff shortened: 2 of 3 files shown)


internal/platform/platform_arm64.go
```diff
@@ -0,0 +1,7 @@
+package platform
+
+// init verifies that the current CPU supports the required ARM64 features
+func init() {
+	// no further checks currently needed
```
mathetake, line 5:
```suggestion
	// No further checks currently needed.
```

internal/platform/platform.go
```diff
@@ -10,6 +10,9 @@ import (
 	"runtime"
 )
 
+// archRequirementsVerified is set by platform-specific init to true if the platform is supported
+var archRequirementsVerified = false
```
mathetake, line 14:
```suggestion
var archRequirementsVerified bool
```

mathetake (approved):

mathetake:
well, at least I would like you to verify it via QEMU, but no need to add it to the CI I think :D

evacchi:
TBF I tried to setup QEMU and I gave up because it was so slow, but I just realized I could just download a preinstalled image instead of starting from scratch 🤦‍♂️ (probably docker would do too...)

mathetake:
If setting toolchains locally is something too slow, you could build a binary `go test -c -o outputbin`  on your main machine, and only send the built test binary to the emulated environment. That's why you could avoid the need to install Go, Git, etc

mathetake:
anyway merging as I _think_ this works!

mathetake:
https://github.com/tetratelabs/wazero/pull/1122

evacchi:
ok I confirm that it works on a qemu machine started with `-cpu qemu64,-sse4_1,-sse4_2`; i.e. `CompilerSupported()` returns false.

mathetake:
Coolio

PR #1124, opened 2023-02-14
wasi: fix nanotime, scale QPC value correctly on Windows

egonelbre:
Performance counter should be scaled by performance frequency to convert to nanoseconds.

Signed-off-by: Egon Elbre <egonelbre@gmail.com>


mathetake (approved):
Awesome, thanks!

PR #1125, opened 2023-02-14
cmd/wazero: fix temp dir handling in TestRun

egonelbre:
TestRun clears the environment, which means the os specific temporary directory is also cleared; which then either creates the directory to the wrong place or it doesn't have any access to create the directory altogether.

Move the temporary directory creation before clearing the environment variables.

Signed-off-by: Egon Elbre <egonelbre@gmail.com>


codefromthecrypt (approved):
wow eagle eyes. thanks!

PR #1128, opened 2023-02-15
Implements fd_datasync in WASI and sync in GOOS=js

codefromthecrypt:
This implements `fd_datasync` in WASI, falling back to normal `File.Sync` when unsupported. This also backfills missing usage of sync in GOOS=js. Finally, this updates the WASI status chart based on what's implemented.

### DIFF
--- a/internal/gojs/fs.go
+++ b/internal/gojs/fs.go
@@ -82,6 +82,14 @@ var (
 	oEXCL = float64(os.O_EXCL)
 )
 
+// The following interfaces are used until we finalize our own FD-scoped file.
+type (
+	// syncer is implemented by os.File in file_posix.go
+	syncer interface{ Sync() error }
+	// truncater is implemented by os.File in file_posix.go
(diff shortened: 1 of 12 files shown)


internal/gojs/fs.go
```diff
@@ -411,7 +419,7 @@ func (processChdir) invoke(ctx context.Context, mod api.Module, args ...interfac
 	path := args[0].(string)
 
 	if s, err := syscallStat(mod, path); err != nil {
-		return nil, ToErrno(err)
+		return nil, err
```
codefromthecrypt, line 422: ToErrno is centralized in GOOS=js

internal/platform/sync.go
```diff
@@ -0,0 +1,7 @@
+package platform
+
+// Fdatasync is like syscall.Fdatasync except that's only defined in linux.
```
codefromthecrypt, line 3: I'll do something like this to complete fd_advise next

Issue #1105, opened 2023-02-07
Allow to pass ModuleConfig to InstantiateModuleFromBinary 


stffabi:
**Is your feature request related to a problem? Please describe.**
When instantiating a module directly from binary it's not possible to pass a `ModuleConfig`, for example to use a custom FS for the module.

**Describe the solution you'd like**
A possibility to use `InstantiateModuleFromBinary` with a custom `ModuleConfig`

**Describe alternatives you've considered**
It's possible to do this with calling first a `CompileModule` and `InstantiateModule`. But since there's a all in one function it would also be awesome if the `ModuleConfig` could be used there.

codefromthecrypt:
We used to have this and we removed it to be conscious of the size of the public API and the number of similar methods in it. That said, ModuleConfig is how you pass things like filesystem config and breaking that into separate steps is likely a first timer pain vs a feature. Happy to add it back.

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/1129

stffabi:
Awesome, thanks so much for this.

PR #1129, opened 2023-02-15
API BREAK: renames InstantiateModuleFromBinary to Instantiate

codefromthecrypt:
This renames `InstantiateModuleFromBinary` to `Instantiate` to both make first time use simpler to write and also de-complicate adding a `WithConfig` variant as requested in #1105

End users in simple case need to change their signature like so.
```diff
-       mod, err := r.InstantiateModuleFromBinary(ctx, addWasm)
+       mod, err := r.Instantiate(ctx, addWasm)
```

In practice, many will not need to change their signature because they had to use the `InstantiateModule` function in order to assign configuration such as the module name, filesystem or use a real clock. Instead, they had to use the more complicated chain of `CompileModule` and `InstantiateModule` even when only assigning config. Users in this situation can opt into the more simplified syntax below:

```go
mod, err := r.InstantiateWithConfig(ctx, addWasm,
	wazero.NewModuleConfig().WithName("adder"))
```


codefromthecrypt:
There are actually a lot of use of the `InstantiateModuleFromBinary` API on GitHub, but I can help migrate them. I think we should do another .pre before the rc at the end of the month.

mathetake (approved):

PR #1130, opened 2023-02-16
platform: adjusts errno on dangling symlink on Windows

mathetake:
part of #1036


mathetake:
OK now the only test fails is inode related https://github.com/tetratelabs/wazero/actions/runs/4189804425/jobs/7262549820

codefromthecrypt (approved):
Thanks for the deep digging, especially on windows!

PR #1132, opened 2023-02-16
wasi/platform: supports inode and dev on Windows

mathetake:


evacchi (approved):
lgtm!!

internal/gojs/testdata/writefs/main.go
```diff
@@ -111,22 +111,20 @@ func Main() {
...
+		log.Panicln("unexpected error", err)
+	}
+	atimeNsec, mtimeNsec, _, _, _, _, err := Stat(df, st)
+	if err != nil {
 		log.Panicln("unexpected error", err)
-	} else {
-		atimeNsec, mtimeNsec, _ := statTimes(st)
-		fmt.Println("times:", atimeNsec, mtimeNsec)
-
-		// statDeviceInode cannot be tested against real device values because
```
codefromthecrypt, line 120: I will add a commit to resurrect what I was aiming for here!

codefromthecrypt: now hoping it still passes :)

codefromthecrypt:
great milestone on windows, @mathetake!

p.s. I think we're nearing or at feature coverage needed to start looking at our own file type, which can really help us pay down the accruing platform-specific tech debt, or at least organize it :p

mathetake:
yep, I think now we've revealed all the requirements that our file type should have now.

PR #1135, opened 2023-02-17
sysfs: adds chmod

codefromthecrypt:
This adds `FS.Chmod` and implements it for `GOOS=js`. This function isn't defined in WASI snapshot01, but it is in [wasi-filesystem](https://github.com/WebAssembly/wasi-filesystem), e.g. [change-file-permissions-at](https://github.com/WebAssembly/wasi-filesystem/blob/main/wasi-filesystem.md#-change-file-permissions-at).


mathetake (approved):
good

PR #1137, opened 2023-02-19
Removes redundant Magic validation in runtime

dmvolod:
If I'm not wrong, with current `Magic` validation in `runtime.CompileModule` never reach this code
https://github.com/tetratelabs/wazero/blob/e2ebce5d230eceff7d89e4d8677a166dd9ebabca/internal/wasm/binary/decoder.go#L29-L31

and will always send `errors.New("invalid binary")` in case of incorrect Magic in wasm .

### DIFF
--- a/runtime.go
+++ b/runtime.go
@@ -1,7 +1,6 @@
 package wazero
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


runtime.go
```diff
@@ -190,7 +189,7 @@ func (r *runtime) CompileModule(ctx context.Context, binary []byte) (CompiledMod
 		return nil, errors.New("binary == nil")
 	}
 
-	if len(binary) < 4 || !bytes.Equal(binary[0:4], binaryformat.Magic) {
+	if len(binary) < 4 {
```
codefromthecrypt, line 192: I think you can take this entire if statement out, it is from when we used to look for the text format.

dmvolod: Thank you for review. Fixed

mathetake (approved):
Thanks!

PR #1138, opened 2023-02-20
wazeroir: less allocations

mathetake:
This stops allocating each `wazeroir.Operation` but instead uses them by values.
In other words, this simply changes `*wazeroir.OperationXYZ` to `wazeroir.OperationXYZ`,
and the following is the bench result on compiling gojs binary:

```
$ benchstat old.txt new.txt
_compile-10     1.12s ± 4%     1.06s ± 1%   -5.54%  (p=0.000 n=20+19)

_compile-10     672MB ± 0%     659MB ± 0%   -1.92%  (p=0.000 n=18+19)

_compile-10     13.5M ± 0%     11.7M ± 0%  -13.14%  (p=0.000 n=19+20)
```


codefromthecrypt (approved):
stack for the win

PR #1139, opened 2023-02-20
Reduces allocations in func validation

mathetake:
Avoids allocating `bytes.Reader` a lot, but instead use the one `bytes.Reader` at the top.

The following is the bench result for gojs binary.

```
_compile-10     1.12s ± 4%     1.10s ± 3%  -1.67%  (p=0.005 n=20+20)

_compile-10     672MB ± 0%     665MB ± 0%  -1.07%  (p=0.000 n=18+19)

_compile-10     13.5M ± 0%     13.3M ± 0%  -1.11%  (p=0.000 n=19+20)

name          old time/op    new time/op    delta
_validate-10    56.9ms ± 2%    52.6ms ± 2%   -7.60%  (p=0.000 n=19+20)

name          old alloc/op   new alloc/op   delta
_validate-10    36.1MB ± 0%    28.9MB ± 0%  -19.94%  (p=0.000 n=16+19)

name          old allocs/op  new allocs/op  delta
_validate-10      747k ± 0%      597k ± 0%  -20.03%  (p=0.000 n=20+20)

```

### DIFF
--- a/internal/wasm/func_validation.go
+++ b/internal/wasm/func_validation.go
@@ -74,6 +74,8 @@ func (m *Module) validateFunctionWithMaxStackValues(
 	// Create the valueTypeStack to track the state of Wasm value stacks at anypoint of execution.
 	valueTypeStack := &valueTypeStack{}
 
+	r := bytes.NewReader(nil)
+
 	// Now start walking through all the instructions in the body while tracking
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
good idea!

PR #1142, opened 2023-02-20
wazeroir: uses u64 as unique keys for labels

mathetake:
what a low-hanging fruit! 

```
_compile-10     1.04s ± 2%     0.93s ± 2%  -10.52%  (p=0.000 n=18+19)

_compile-10     646MB ± 0%     609MB ± 0%   -5.67%  (p=0.000 n=18+20)

_compile-10     10.7M ± 0%      9.4M ± 0%  -12.59%  (p=0.000 n=20+20)
```

### DIFF
--- a/internal/wazeroir/operations.go
+++ b/internal/wazeroir/operations.go
@@ -1,6 +1,8 @@
 package wazeroir
 
-import "fmt"
+import (
+	"fmt"
+)
 
 // UnsignedInt represents unsigned 32-bit or 64-bit integers.
 type UnsignedInt byte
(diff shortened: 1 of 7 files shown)


internal/wazeroir/operations.go
```diff
@@ -730,6 +732,15 @@ type Label struct {
 	Kind    LabelKind
 }
 
+// LabelID is the unique identifier for blocks in a single functions.
+type LabelID = uint64
```
codefromthecrypt, line 736: supernit is this an alias for a reason?

codefromthecrypt (approved):
thanks for turning the cranks!

internal/wazeroir/operations.go
mathetake: no reason :D

mathetake: https://github.com/tetratelabs/wazero/pull/1142/commits/6c345a6ff9f62cc94d486335ccbbdef281fce5d2

PR #1144, opened 2023-02-20
wazeroir: avoids allocation in applyToStack

mathetake:
```
_compile-10     917ms ± 1%     902ms ± 2%   -1.71%  (p=0.000 n=17+18)

_compile-10     609MB ± 0%     607MB ± 0%   -0.43%  (p=0.000 n=20+20)

_compile-10     9.39M ± 0%     7.68M ± 0%  -18.28%  (p=0.000 n=20+20)

```


codefromthecrypt (approved):
a lot of return for a small change!

PR #1146, opened 2023-02-21
wazeroir: reduces allocations with controlFrames

mathetake:
```
_compile-10     886ms ± 0%     876ms ± 1%  -1.11%  (p=0.000 n=20+19)

_compile-10     597MB ± 0%     589MB ± 0%  -1.25%  (p=0.000 n=20+20)

_compile-10     7.40M ± 0%     7.22M ± 0%  -2.46%  (p=0.000 n=20+20)

```

### DIFF
--- a/internal/wazeroir/compiler.go
+++ b/internal/wazeroir/compiler.go
@@ -33,7 +33,7 @@ type (
 		blockType                    *wasm.FunctionType
 		kind                         controlFrameKind
 	}
-	controlFrames struct{ frames []*controlFrame }
+	controlFrames struct{ frames []controlFrame }
 )
 
 func (c *controlFrame) ensureContinuation() {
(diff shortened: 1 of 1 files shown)


internal/wazeroir/compiler.go
```diff
@@ -300,17 +301,11 @@ func CompileFunctions(enabledFeatures api.CoreFeatures, callFrameStackSizeInUint
 				GoFunc:         code.GoFunc,
 				Signature:      sig,
 			})
-
-			if len(sig.Params) > 0 && sig.ParamNumInUint64 == 0 {
-				panic("a")
-			} else if len(sig.Results) > 0 && sig.ResultNumInUint64 == 0 {
-				panic("if len(sig.Results) > 0 && sig.ResultNumInUint64 == 0")
-			}
```
mathetake, line 308: what a garbage 🤕

codefromthecrypt (approved):
thanks for all the polishing work. it will result in a much better 1.0!

gedw99:
Great work @mathetake

PR #1148, opened 2023-02-21
ci: stops building Rust binaries in wasi-testsuite

mathetake:


codefromthecrypt (approved):
thanks for following up

PR #1149, opened 2023-02-21
adds platform.Readdirnames and uses in gojs

codefromthecrypt:
This adds `platform.Readdirnames` which is preparation work before doing something similar for reading the directory. We use this in gojs as it doesn't actually need dirents, rather just names.

### DIFF
--- a/internal/platform/dir.go
+++ b/internal/platform/dir.go
@@ -1,0 +1,38 @@
+package platform
+
+import (
+	"io/fs"
+	"syscall"
+)
+
+// readdirnamesFile is implemented by os.File in dir.go
+// Note: we use this until we finalize our own FD-scoped file.
+type readdirnamesFile interface {
+	Readdirnames(n int) (names []string, err error)
+}
+
+// Readdirnames is like the function on os.File, but for fs.File. This returns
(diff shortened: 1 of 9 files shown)


internal/platform/dir.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// Readdirnames is like the function on os.File, but for fs.File. This returns
+// syscall.ENOTDIR if not a directory or syscall.EBADF if closed.
+func Readdirnames(f fs.File, n int) (names []string, err error) {
+	switch f := f.(type) {
+	case readdirnamesFile:
+		names, err = f.Readdirnames(n)
+	case fs.ReadDirFile:
+		entries, err := f.ReadDir(n)
```
mathetake, line 21: I think err is shadowed in this block, so non-nil error here won't be captured?

mathetake (approved):

internal/platform/dir.go
codefromthecrypt: great eye

codefromthecrypt: great eye. fixed and added a test.

PR #1151, opened 2023-02-22
site: adds community page, seeded with a couple projects

### LINKED ISSUE #577: New webpage: How to add cross-platform plugin support to a Go program using WebAssembly.
Many are interested in wazero as a way to add cross-platform plugin support to their applications. This topic covers too much ground for a one-liner or even a paragraph. We should have a web page on https://wazero.io for this.

This should overview the general opportunities and constraints. For example, what's portable by default (eg can also run in something besides wazero ex WASI), and what you can define and choose to do. How wazero helps in general by being testable in go, etc.

It would mention that the program being extended can communicate with plugins in wasm via sandboxed memory, host callbacks, as well channels allowed in WASI (ex files).

It can overview the common stages done in plugin systems, such as an "init" which can setup reference data in memory for the plugin to use, as well some common vocabulary used in wasm plugin architectures (such as ABI, host, guest).

It would also have some section describing the impact of constrained languages, such as no_std in Rust and TinyGo, and some libraries like [gjson](https://github.com/tidwall/gjson) and [karmem](https://github.com/inkeliz/karmem), pointing to our allocation examples for common issues.

Finally, this would point to a new "baby plugin" example, which must be carefully done to be small, but outline the common things above.

Note: we'd have to be careful this doesn't turn into a book (at least immediately) as completing this is more important than covering all topics. We can link to some existing plugin systems at the end for reference.

We can defer entirely to documentation on other sites, like https://github.com/wapc/wapc-go or an example like https://github.com/tetratelabs/proxy-wasm-go-sdk. However, the basic ideas should be introduced first as plugins particularly in wasm have some nuance to them, which could be explained.

We could also make a presentation, and probably will, but a website page would be helpful anyway, and also could be translated as necessary.

**Additional context**
[suggestion of mosn integration by @tanjianfeng](https://github.com/google/gvisor/issues/5811#issuecomment-1129550352)
[request in hackernews](https://news.ycombinator.com/item?id=31418851)

codefromthecrypt:
This adds infrastructure for community members to add their own projects or products.

I added a couple I help maintain (http-wasm and go-plugin) as well waggy from @syke99 to start with. Once this is merged, others can add their own a lot easier!

fixes #577


codefromthecrypt:
here's the preview link https://deploy-preview-1151--wazero.netlify.app/community/users/

mathetake:
Waggy is nothing to do with wazero, no?

codefromthecrypt:
> Waggy is nothing to do with wazero, no?

indeed! sorry I had assumed it did, and didn't check. I think that one is a better fit for the WAGI community page. thanks for the fresh eyes!

codefromthecrypt:
looks like there is a real WAGI one in progress though, but I'll leave it to @evacchi to decide to add or not ;)

evacchi (approved):

codefromthecrypt:
@ncruces since I usually talk about your [dcraw](https://github.com/ncruces/dcraw) and [gosqlite-3](https://github.com/ncruces/go-sqlite3), I expect you wouldn't be opposed to these being on the wazero community page. wdyt?

ncruces:
Not at all! 😀

PR #1154, opened 2023-02-23
wasi: adds platform.Dirent in preparation of inode fetching

codefromthecrypt:
wasi_snapshot_preview1 recently requires fd_readdir to return actual inode values. On zero, wasi-libc will call fdstat to retrieve them.

This introduces our own `platform.Dirent` type and `Readdir` function which a later change will allow fetching of inodes.

See https://github.com/WebAssembly/wasi-libc/pull/345

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -813,7 +813,7 @@ func fdReaddirFn(_ context.Context, mod api.Module, params []uint64) Errno {
 		if err != nil {
 			return ToErrno(err)
 		}
-		rd, dir = f.File.(fs.ReadDirFile), f.ReadDir
+		rd, dir = f.File, f.ReadDir
 	}
 
 	// First, determine the maximum directory entries that can be encoded as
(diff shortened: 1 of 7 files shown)


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -831,26 +831,26 @@ func fdReaddirFn(_ context.Context, mod api.Module, params []uint64) Errno {
 
 	// The host keeps state for any unread entries from the prior call because
 	// we cannot seek to a previous directory position. Collect these entries.
-	entries, errno := lastDirEntries(dir, cookie)
+	dirents, errno := lastDirents(dir, cookie)
```
evacchi, line 834: real nit, but I guess `Dirent` instead of `DirEnt` and `Readdir` instead of `ReadDir` is for consistency with WASI (libc) defs, right?

internal/platform/dir.go
```diff
@@ -21,11 +22,60 @@ func Readdirnames(f fs.File, n int) (names []string, err error) {
...
+
+// IsDir returns true if the Type is fs.ModeDir.
+func (d *Dirent) IsDir() bool {
+	return d.Type == fs.ModeDir
+}
+
+// Readdir is like the function on os.File, but for fs.File. This returns
+// syscall.ENOTDIR if not a directory or syscall.EIO if closed or read
+// redundantly.
+func Readdir(f fs.File, n int) (dirents []*Dirent, err error) {
```
evacchi, line 64: I would add a note on the semantics of `n`

evacchi (approved):
overall ok, just minor nits about casing and one tiny note about docs

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: indeed posix mostly, but also there's an odd difference in go between ReadDir and Readdir, and this is closer.

codefromthecrypt: added a comment

codefromthecrypt:
thanks for the look @evacchi even though this is internal, we should look more carefully as we near to 1.0!

PR #1155, opened 2023-02-23
sys.FileTable => descriptor.Table

achille-roussel:
@codefromthecrypt following up from our conversation on Slack, this PR modifies the implementation of `sys.FileTable` to be based off of a generic `descriptor.Table` data structure.

I moved the data structure to a new `internal/descriptor` package so other instances do not have to take a dependency on `internal/sys` (it might limit coupling between packages), I also think that `descriptor.Table` communicates well what the type represents.

I changed `sys.FileTable` to be a type alias on `descriptor.Table` to keep the changes to a minimum (no need to change places that used `sys.FileTable).

I don't have much visibility into the other use cases where you imagined using it, so let me know if this proposed change won't fit and I can make the needed adjustments.

Hope this will be helpful!


codefromthecrypt (approved):
fantastic work. I have nothing to add

PR #1156, opened 2023-02-23
Fix issue with a delayed check of the context cancelation

ckaznocha:
If a context is canceled after a guest func call has returned it could cause the module to close in some cases. This change ensures that a delayed check of the context cancelation is ignored.

I've also reduced the cost of context cancelation a bit.

### DIFF
--- a/internal/wasm/call_context.go
+++ b/internal/wasm/call_context.go
@@ -73,17 +73,26 @@ func (m *CallContext) FailIfClosed() (err error) {
 //
 // Callers of this function must invoke the returned context.CancelFunc to release the spawned Goroutine.
 func (m *CallContext) CloseModuleOnCanceledOrTimeout(ctx context.Context) context.CancelFunc {
+	// Creating an empty channel in this case is a bit more efficient than
+	// creating a context.Context and canceling it with the same effect. We
+	// really just need to be notified when to stop listening to the users
+	// context. Closing the channel will unblock the select in the goroutine
+	// causing it to return an stop listening to ctx.Done().
 	cancelChan := make(chan struct{})
-	go m.closeModuleOnCanceledOrTimeoutClosure(ctx, cancelChan)
+	go m.closeModuleOnCanceledOrTimeout(ctx, cancelChan)
 	return func() { close(cancelChan) }
 }
 
-// closeModuleOnCanceledOrTimeoutClosure is extracted from CloseModuleOnCanceledOrTimeout for testing.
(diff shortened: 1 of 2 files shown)


internal/wasm/call_context.go
```diff
@@ -73,29 +73,27 @@ func (m *CallContext) FailIfClosed() (err error) {
 //
 // Callers of this function must invoke the returned context.CancelFunc to release the spawned Goroutine.
 func (m *CallContext) CloseModuleOnCanceledOrTimeout(ctx context.Context) context.CancelFunc {
-	goroutineDone, cancelFn := context.WithCancel(context.Background())
-	go m.closeModuleOnCanceledOrTimeoutClosure(ctx, goroutineDone)()
-	return cancelFn
+	cancelChan := make(chan struct{})
```
codefromthecrypt, line 76: maybe add a comment about why not context.WithCancel or if too long a blurb in RATIONALE.md? Seems took enough work to want to keep that info in case it gets "fixed" later ;)

codefromthecrypt (approved):
Thanks very much for the help on this very subtle bug!

internal/wasm/call_context.go
codefromthecrypt: I guess basically the main thing is to say how this prevents arbitrary choice of the wrong branch in the select, not necessarily about cancelcontext, but in any case whichever part of this is critical to correctness would be a great place to comment/warn to prevent someone from undoing this later.

ckaznocha: For sure, added some comments. To be clear this bit didn't fix the bug, this was just to reduce the resources enabling `WithCloseOnContextDone` uses a bit.

internal/wasm/call_context.go
```diff
@@ -73,29 +73,36 @@ func (m *CallContext) FailIfClosed() (err error) {
...
-				return
-			case <-goroutineDone.Done():
-				return
+// closeModuleOnCanceledOrTimeout is extracted from CloseModuleOnCanceledOrTimeout for testing.
+func (m *CallContext) closeModuleOnCanceledOrTimeout(ctx context.Context, cancelChan <-chan struct{}) {
+	select {
+	case <-ctx.Done():
+		select {
+		case <-cancelChan:
+			// In some cases by the time this goroutine is scheduled, the caller
```
codefromthecrypt, line 92: fantastic!

codefromthecrypt (approved):
great job

mathetake:
awesome, thanks @ckaznocha !

PR #1158, opened 2023-02-24
Close and return immediately if the context is already canceled

ckaznocha:


codefromthecrypt (approved):
As this is guarded with a boolean check, I don't think we need to do the before/after bench.

Thanks for backfilling the tests!

mathetake:
👍

PR #1165, opened 2023-02-24
Set Memory.Max to min(user-memory-limit, sys-memory-limit) on load

### LINKED ISSUE #1153: Let modules with max memory beyond memory limit grow up to memory limit
When compiling a binary with emscripten's `-sALLOW_MEMORY_GROWTH=1`, the resulting memory is configured as `initial=256 max=32768`. When wazero is uses with a memory limit between those bounds using [WithMemoryCapacityFromMax](https://pkg.go.dev/github.com/tetratelabs/wazero#RuntimeConfig.WithMemoryCapacityFromMax), wazero will fail with an error like
```
section memory: max 32768 pages (2 Gi) over limit of 512 pages (32 Mi)
```
at module compilation.

When the wasm module's max memory is beyond wazero memory limit, but the initial setting is below the wazero limit, wazero successfully compiles and instantiate a module and only triggers a failure once the requested memory grows beyond the limit.

- Check if emscripten could compile without a maximum memory configuration. From my look into the project they conditionally compile things differently if >2GB could be addressed, which is presumably why they have a max rather than not specifying a max.
- Understand how this works in other runtimes. I didn't find any cli arguments in wasmer/wasmtime to limit memory size and didn't manage to use their rust bindings to test.

**Additional context**
Based on my reading of https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#memories%E2%91%A0
> The [min](https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-limits) size in the [limits](https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-limits) of the memory type specifies the initial size of that memory, while its [max](https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#syntax-limits), if present, restricts the size to which it can grow later. Both are in units of [page size](https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#page-size).

And section [4.5.3.7. Growing memories](https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#growing-memories%E2%91%A0), it seems like the max limit should apply when modules are growing their memory. While these official specs don't mention anything about implementation defined limits, limiting on growth fit my expectations better than what happens now.

To reproduce:
```
$ cat main.cc
#include <iostream>

int main() {
    std::cout << "Hello World!";
    return 0;
}
$ emcc --version
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.30-git
$ emcc -sSTANDALONE_WASM -sALLOW_MEMORY_GROWTH=1 -o main.wasm main.cc
$ wasm-objdump -x --section='memory' main.wasm

main.wasm:      file format wasm 0x1

Section Details:

Memory[1]:
 - memory[0] pages: initial=256 max=32768
```
produces the message above with at `v1.0.0-pre.9`:
```go
package main

import (
	"context"
	_ "embed"
	"fmt"
	"log"
	"os"

	"github.com/tetratelabs/wazero/imports/emscripten"
)

//go:embed main.wasm
var wasmBytes []byte

func main() {
	ctx := context.Background()

	runtimeConfig := wazero.NewRuntimeConfig().
		WithMemoryLimitPages(512)
	r := wazero.NewRuntimeWithConfig(ctx, runtimeConfig)
	defer r.Close(ctx)

		WithStdout(os.Stdout).
		WithStderr(os.Stderr)

	emscripten.MustInstantiate(ctx, r)

	compiledModule, err := r.CompileModule(ctx, wasmBytes)
	if err != nil {
		log.Panicln(err)
	}
	if _, err := r.InstantiateModule(ctx, compiledModule, config); err != nil {
		} else if !ok {
			log.Panicln(err)
		}
	}
}
```

evacchi:
Updated `newMemorySizer()` to return the updated value, but also
ensure that an invalid `max` still throws an error (invalid module).

- Minor cleanup to use the `memorySizer` type instead of the full
  func signature for clarity
- Added a wat+wasm under `internal/integration_test/vs/testdata/`
  simply because that's where the other cache-related testdata was.

Closes #1153


codefromthecrypt (approved):
great quality work!

PR #1166, opened 2023-02-25
Readdir: handles io.EOF

codefromthecrypt:
This handles EOF even if current and possibly future wasi don't have a way to propagate an EOF signal. This is mainly to match up with go semantics and ensure we don't have any error conditions (by adding tests).


evacchi (approved):

codefromthecrypt:
this worked for ReadDir, but not Readdir. basically the EOF result is not platform consistent, so I moved back to no-EOF

PR #1167, opened 2023-02-25
site: adds sqlite3 as a user

ncruces:


codefromthecrypt (approved):
Thanks for all the work on go-sqlite!

PR #1172, opened 2023-02-27
windows: fixes unlinking symlink to dir, rmdir on opened empty dir.

mathetake:


mathetake:
not found a fix yet.

internal/sysfs/dirfs_test.go
```diff
@@ -375,27 +375,55 @@ func TestDirFS_Rmdir(t *testing.T) {
...
 		require.NoError(t, os.Mkdir(realPath, 0o700))
 
-		err := testFS.Unlink(name)
+		err := testFS.Unlink(dir)
 		require.EqualErrno(t, syscall.EISDIR, err)
 
 		require.NoError(t, os.Remove(realPath))
 	})
 
+	t.Run("target: symlink to dir", func(t *testing.T) {
```
mathetake, line 402: this one is the new one and only windows failing

mathetake:
interesting as I expected this only fails with 1.20.

mathetake:
looks like this has existed for a while but somehow hasn't been caught until Zig test has started failing with 1.20 upgrade

mathetake:
ok original issue has been fixed, but another comes up:

```

test "makePath in a directory that no longer exists" {
    if (builtin.os.tag == .windows) return error.SkipZigTest; // Windows returns FileBusy if attempting to remove an open dir

    var tmp = tmpDir(.{});
    defer tmp.cleanup();
    try tmp.parent_dir.deleteTree(&tmp.sub_path);

    try testing.expectError(error.FileNotFound, tmp.dir.makePath("sub-path"));
}

```

codefromthecrypt:
@mathetake the other one looks like they are skipping on windows, which I guess is annoying because we can't do that without excluding "makePath in a directory that no longer exists" from all operating systems (due to the filter being a compile option). one way could be to exclude it from zig, but re-add the test in our repo so that it runs when != windows. wdyt?

mathetake:
`makePath in a directory that no longer exists"` this is not the only test in zig skipped on window, and there's tens and we previously passing. Plus wasmtime passes so there must be a way to pass on windows. Also, we should be able to provide consistent behavior regardless of the host. So I will figure out this rather than adding skip on windows.

mathetake:
minimize the case

```zig
test "minimized" {
    var tmp = tmpDir(.{});
    tmp.parent_dir.deleteTree(&tmp.sub_path) catch @panic("aaaaaaaa");
}
```

mathetake:
found the repro which only fails on windows https://github.com/tetratelabs/wazero/pull/1172/commits/1e35027971bc7452863da464cb31f05e9a7557e2

mathetake:
maybe time to resolve https://github.com/tetratelabs/wazero/blame/75aa6b2a6efa86b20e9b4eecd2889c83b714b1ca/internal/platform/open_file_windows.go#L40

internal/platform/open_file_windows.go
```diff
@@ -140,6 +140,10 @@ func open(path string, mode int, perm uint32) (fd syscall.Handle, err error) {
 			}
 		}
 	}
+	if createmode == syscall.OPEN_EXISTING && access == syscall.GENERIC_READ {
```
mathetake, line 143: lifted from the change made in Go 1.20 https://github.com/golang/go/blob/go1.20/src/syscall/syscall_windows.go#L308-L379 and I guess that's why we haven't encountered it before.

mathetake:
LOL only windows 1.18 is failing LOL

mathetake:
🤞

internal/platform/open_file_windows.go
```diff
@@ -140,6 +142,16 @@ func open(path string, mode int, perm uint32) (fd syscall.Handle, err error) {
 			}
 		}
 	}
+
+	if strings.Contains(runtime.Version(), "go1.20") {
```
codefromthecrypt, line 146: maybe cache this as a var since it won't change?

mathetake: will do after CI went green

mathetake:
all went green

internal/platform/open_file_windows.go
```diff
@@ -140,6 +142,18 @@ func open(path string, mode int, perm uint32) (fd syscall.Handle, err error) {
...
+			// Necessary for opening directory handles.
+			attrs |= syscall.FILE_FLAG_BACKUP_SEMANTICS
+		}
+	}
+
 	h, e := syscall.CreateFile(pathp, access, sharemode, sa, createmode, attrs, 0)
 	return h, e
 }
+
+var isGo120 = strings.Contains(runtime.Version(), "go1.20")
```
codefromthecrypt, line 159: This will break in 1.21 but I can sort it after merge

codefromthecrypt (approved):
thanks!

PR #1173, opened 2023-02-27
Add timeout support in `wazero run` cli

evacchi:
- ensure the module initialization function is evaluated
  regardless if it's WASI or GoJS
- add a `-timeout duration` flag
- ensure flag gets propagated to the rt config builder
  (also ensure cache flag gets propagated to the rt config builder)
- print a message to stderr when the deadline is exceeded
- configure GitHub Actions to use `-timeout=10m` flag
  instead of GHA's `timeout-minutes: 10`


cmd/wazero/wazero.go
```diff
@@ -195,7 +201,13 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
 		rtc = wazero.NewRuntimeConfig()
 	}
 	if cache := maybeUseCacheDir(cacheDir, stdErr, exit); cache != nil {
-		rtc.WithCompilationCache(cache)
+		rtc = rtc.WithCompilationCache(cache)
```
evacchi, line 204: we forgot to reassign here too

cmd/wazero/wazero.go
```diff
@@ -195,7 +201,13 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
...
+	}
+	if timeout != 0 {
+		newCtx, cancel := context.WithTimeout(ctx, timeout)
+		ctx = newCtx
+		defer cancel()
+		rtc = rtc.WithCloseOnContextDone(true)
```
evacchi, line 210: I assume returning a clone of `rtc` instead of mutating the receiver is a well-known pattern for Gophers, but it was a bit surprising for me :P

cmd/wazero/wazero.go
```diff
@@ -231,11 +243,17 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
...
 		err = gojs.Run(ctx, rt, code, conf)
+	} else {
+		_, err = rt.InstantiateModule(ctx, code, conf)
 	}
 
 	if err != nil {
 		if exitErr, ok := err.(*sys.ExitError); ok {
-			exit(int(exitErr.ExitCode()))
+			exitCode := exitErr.ExitCode()
+			if timeout > 0 && exitCode == sys.ExitCodeDeadlineExceeded {
```
mathetake, line 253: do we need `timeout >0`? I guess `sys.ExitCodeDeadlineExceeded` won't be set if `timeout == 0`

cmd/wazero/wazero.go
```diff
@@ -231,11 +243,17 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
...
+				fmt.Fprintf(stdErr, "canceled: wasm binary '%s' exceeded timeout (%v)\n", wasmExe, timeout)
```
mathetake, line 254: can we just reuse `exitErr.Error()` https://github.com/tetratelabs/wazero/blob/main/sys/error.go#L59-L68 ?

cmd/wazero/wazero.go
```diff
@@ -133,6 +134,11 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
 			"This may be specified multiple times. When <wasm path> is unset, <path> is used. "+
 			"For read-only mounts, append the suffix ':ro'.")
 
+	var timeout time.Duration
+	flags.DurationVar(&timeout, "timeout", 0*time.Second,
+		"if a wasm binary runs longer than the given duration, then panic. "+
```
mathetake, line 139: for non gophers, the duration format for this string might not be common. Maybe add example here??

codefromthecrypt:
also for another PR I think we should test ctrl-c (sigint) on an endless loop. Not sure, but maybe we'll end up needing context integration and a signal handle..

cmd/wazero/wazero.go
evacchi: done, although the error becomes slightly more cryptic:

> error: module "" closed with context deadline exceeded (timeout 1ms)

I have still added the timeout

cmd/wazero/wazero.go
evacchi: the message is tad longer now, but it should be good (basically lifted the one in the Go docs for ParseDuration except the note about it being negative because it makes no sense)

cmd/wazero/wazero.go
```diff
@@ -92,11 +93,11 @@ func doCompile(args []string, stdErr io.Writer, exit func(code int)) {
 
 	c := wazero.NewRuntimeConfig()
 	if cache := maybeUseCacheDir(cacheDir, stdErr, exit); cache != nil {
-		c.WithCompilationCache(cache)
+		c = c.WithCompilationCache(cache)
 	}
 
 	ctx := context.Background()
-	rt := wazero.NewRuntime(ctx)
+	rt := wazero.NewRuntimeWithConfig(ctx, c)
```
evacchi, line 100: we forgot to use c here

codefromthecrypt (approved):
agree the desc is a little wordy now, but not too much and we can blame it on go ;)

good job!

evacchi:
> also for another PR I think we should test ctrl-c (sigint) on an endless loop. Not sure, but maybe we'll end up needing context integration and a signal handle..

Unless I misunderstood, I think it already works? 🤔 

    go install ./cmd/wazero
    wazero run cmd/wazero/testdata/infinite_loop.wasm

c-c works as expected (no change required)

codefromthecrypt:
sorry about ctrl-c I didn't know if it did or not (we should backfill a test. there's a similar one in func-e under the e2e directory). anyway different but similar topic though out of scope here!

PR #1174, opened 2023-02-27
sysfs: drops os.File special casing for fs.FS to pass wasi-testsuite

codefromthecrypt:
This adds a wazero adapter which passes wasi-testsuite 100pct on darwin, linux and windows. While the main change was adding inodes to the wasi `fd_readdir` dirents, there was a lot of incidental work needed.

Most of the work was troubleshooting in nature, around windows specifically, but also wrapping of files. This backfills a lot of tests and reworked how wrapping works, particularly around windows.

To accommodate this, we drop os.File special casing except for sysfs.DirFS.

### Why don't we attempt to pass wasi-testsuite on user-defined `fs.FS`?

While most cases work fine on an `os.File` based implementation, we won't promise wasi-testsuite compatibility on user defined wrappers of `os.DirFS`. The only option for real systems is to use our `sysfs.FS`.

There are a lot of areas where windows behaves differently, despite the `os.File` abstraction. This goes well beyond file locking concerns (e.g. `EBUSY` errors on open files). For example, errors like `ACCESS_DENIED` aren't properly mapped to `EPERM`. There are trickier parts too. `FileInfo.Sys()` doesn't return enough information to build inodes needed for WASI. To rebuild them requires the full path to the underlying file, not just its directory name, and there's no way for us to get that information. At one point we tried, but in practice things became tangled and functionality such as read-only wrappers became untenable. Finally, there are version-specific behaviors which are difficult to maintain even in our own code. For example, go 1.20 opens files in a different way than versions before it.


codefromthecrypt:
will start on windows as soon as I see if the testsuite agrees with me on linux and darwin ;)

codefromthecrypt:
ok I've organized pretty much what works and doesn't work. I'll work on the windows side when my laptop stops windows update ;)

codefromthecrypt:
I'll raise an issue post merge about that composite fs doesn't return inodes properly. this really can't happen until we have our own file type as trying now is really messy

.github/wasi_testsuite_adapter.py
```diff
@@ -0,0 +1,47 @@
+# adapter for wazero until/unless https://github.com/WebAssembly/wasi-testsuite/pull/55
```
codefromthecrypt, line 1: this adapter was the culmination of work by @achille-roussel @evacchi and @mathetake

internal/platform/file.go
```diff
@@ -0,0 +1,50 @@
+package platform
+
+import (
+	"io"
+	"io/fs"
+)
+
+// ReadFile declares all read interfaces defined on os.File used by wazero.
```
codefromthecrypt, line 8: these types were moved from sysfs package, and yeah soon we'll swap these out for our own file type. nearly there.

internal/platform/file.go
```diff
@@ -0,0 +1,50 @@
...
+	io.Writer
+	io.WriterAt // for pwrite
+	chmodFile
+	syncFile
+	truncateFile
+}
+
+// The following interfaces are used until we finalize our own FD-scoped file.
+type (
+	// PathFile is implemented on files that retain the path to their pre-open.
```
codefromthecrypt, line 30: this is needed to keep windows working even if we wrap an FS read-only

internal/platform/stat_windows.go
```diff
@@ -63,6 +60,18 @@ func statFile(f fs.File, st *Stat_t) (err error) {
 	return defaultStatFile(f, st)
 }
 
+// inoFromFileInfo uses stat to get the inode information of the file.
+func inoFromFileInfo(f readdirFile, t fs.FileInfo) (ino uint64, err error) {
+	if pf, ok := f.(PathFile); ok {
+		var st Stat_t
+		inoPath := path.Clean(path.Join(pf.Path(), t.Name()))
+		if err = lstat(inoPath, &st); err == nil {
```
codefromthecrypt, line 68: this is where we use our special interface method in order to build up the path to the directory entry. lstat is the right choice vs stat for dirents, as for example, links aren't followed.

internal/platform/wrap_windows.go
```diff
@@ -0,0 +1,132 @@
...
+
+// windowsWrappedFile deals with errno portability issues in Windows. This code
+// is likely to change as we complete WASI and GOOS=js.
+//
+// If we don't map to syscall.Errno, wasm will crash in odd way attempting the
+// same. This approach is an alternative to making our own fs.File public type.
+// We aren't doing that yet, as mapping problems are generally contained to
+// Windows. Hence, file is intentionally not exported.
+//
+// Note: Don't test for this type as it is wrapped when using sysfs.NewReadFS.
```
codefromthecrypt, line 16: trying to test for this type lost some time, hence `PathFile` workaround until we make our file type.

internal/sysfs/dirfs.go
```diff
@@ -42,11 +42,7 @@ func (d *dirFS) Open(name string) (fs.File, error) {
 
 // OpenFile implements FS.OpenFile
 func (d *dirFS) OpenFile(path string, flag int, perm fs.FileMode) (fs.File, error) {
-	if f, err := platform.OpenFile(d.join(path), flag, perm); err != nil {
```
codefromthecrypt, line 45: wrapping in the syscall abstraction worked but not for all cases. it is now handled where the other OS logic is.

internal/sysfs/sysfs_test.go
```diff
@@ -68,73 +66,132 @@ func testOpen_O_RDWR(t *testing.T, tmpDir string, testFS FS) {
...
-	file1 := "file1"
-	fileInDir := path.Join(dirRealPath, file1)
-	require.NoError(t, os.WriteFile(fileInDir, []byte{2}, 0o600))
-
-	emptydir := "emptydir"
-	emptydirRealPath := path.Join(tmpDir, emptydir)
-	err = os.Mkdir(emptydirRealPath, 0o700)
-	require.NoError(t, err)
-
+func testOpen_Read(t *testing.T, testFS FS, expectIno bool) {
```
codefromthecrypt, line 69: this was rejigged to use the same testdata as everywhere else.

internal/sysfs/readfs_test.go
```diff
@@ -15,10 +16,12 @@ func TestNewReadFS(t *testing.T) {
 	tmpDir := t.TempDir()
 
 	// Doesn't double-wrap file systems that are already read-only
-	adapted := Adapt(os.DirFS(tmpDir))
-	require.Equal(t, adapted, NewReadFS(adapted))
 	require.Equal(t, UnimplementedFS{}, NewReadFS(UnimplementedFS{}))
 
+	// Wraps a fs.FS because it allows access to Write
```
codefromthecrypt, line 21: this was a miss before

internal/sysfs/readfs_test.go
```diff
@@ -124,11 +127,35 @@ func TestReadFS_Utimes(t *testing.T) {
...
-	testFS := NewReadFS(writeable)
+	require.NoError(t, fstest.WriteTestFiles(tmpDir))
+
+	type test struct {
+		name      string
+		fs        FS
+		expectIno bool
+	}
+
+	tests := []test{
```
codefromthecrypt, line 141: this identified a bunch of issues

codefromthecrypt (commented):
notes!

.github/wasi_testsuite_adapter.py
mathetake: Oh I did nothing, and I just copy-pasta 🍝 in the PR :D

mathetake (approved):
huge milestone! (last piece?)

codefromthecrypt:
> huge milestone! (last piece?)


So, yeah it is a great milestone to be maybe the only runtime that passes wasi-testsuite on all platforms (not sure if others do, as they aren't testing windows).

This is the last for this release candidate. There are a couple corner cases we haven't finished for GOOS=js which once complete reveals the entire feature surface needed to make our file type. These will continue after the cut and maybe our internal structure can be stable by 1.0.

PR #1177, opened 2023-02-28
data segment: encode the right prefix if a const expr is missing

evacchi:
If `OffsetExpression` is nil, we assume passive mode and use `0x1`

0x2 is only for the cases where the memory index is >=0,
i.e. even though 0 is a valid value, it only makes sense to use
when it is > 0. However, currently the spec does not support indices > 0.

So, for the time being, we only do the simple thing and correctly encode
passive segments as 0x1, otherwise defaulting to 0x0

This is a port of the patch at https://github.com/tetratelabs/wabin/pull/7

### DIFF
--- a/internal/wasm/binary/data.go
+++ b/internal/wasm/binary/data.go
@@ -79,9 +79,13 @@ func decodeDataSegment(r *bytes.Reader, enabledFeatures api.CoreFeatures) (*wasm
 }
 
 func encodeDataSegment(d *wasm.DataSegment) (ret []byte) {
-	// Currently multiple memories are not supported.
-	ret = append(ret, leb128.EncodeInt32(0)...)
-	ret = append(ret, encodeConstantExpression(d.OffsetExpression)...)
+	if d.OffsetExpression == nil {
+		ret = append(ret, leb128.EncodeInt32(int32(dataSegmentPrefixPassive))...)
+	} else {
+		// Currently multiple memories are not supported.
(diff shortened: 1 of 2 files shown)


mathetake:
Binaryencoding spec version 2 is in general not implemented as opposed to 1 which is previously tested (https://github.com/tetratelabs/wazero/blob/5b2c25cfed9e5ebecc28679fd3557c947b9aa37c/internal/integration_test/spectest/v1/spec_test.go#L24-L26). The reason why encoding has existed in the first place is to add support for wat2wasm in wazero natively (which is no longer the case), and we partially use it in unittests

I think we should decide wether we should implement all the binary encoding here in wazero or just remove it. If we proceed this, we should revive the previous test I mentioned ^ on encoding of v1 plus we should add v2 encoding tests. wdyt? @codefromthecrypt

codefromthecrypt:
honestly I forget where if anywhere we are using binary encoding. Some reason I thought the compiler cache used it.

If it is only in tests, then we should remove all this code before 1.0 or at least move it to the test tree. I think we had plans to attempt to compile wabt or wasm-tools to wasm and avoid anything about encoding here, yet still keep our build with less 3rd party platform-specific deps.

mathetake:
yeah agreed at least we should move all the encoding related codes into internal/testing/binaryencoding/**

codefromthecrypt:
ok good idea. @evacchi mind doing the refactoring?

evacchi:
can do!

PR #1182, opened 2023-03-01
Ensure `nil` is accepted as FS config

evacchi:
In previous releases, passing a `nil` value as an FS config did not cause any error. However, in 1.0.0-rc.1 this leads
to the creation of an invalid `adapter{fs: nil}`, which eventually leads to a panic (nil):

    (f *FileEntry) Stat(st *platform.Stat_t) =>
        (r *lazyDir) file() =>
            r.fs.OpenFile(".", os.O_RDONLY, 0)

with fs == nil

The backwards-compatible fix is to make Adapt() return UnimplementedFS.

However, we may want to consider returning an error somewhere, because configuring a nil FS may be unintended.

### DIFF
--- a/config.go
+++ b/config.go
@@ -667,7 +667,11 @@ func (c *moduleConfig) WithEnv(key, value string) ModuleConfig {
 
 // WithFS implements ModuleConfig.WithFS
 func (c *moduleConfig) WithFS(fs fs.FS) ModuleConfig {
-	return c.WithFSConfig(NewFSConfig().WithFSMount(fs, ""))
+	var config FSConfig
+	if fs != nil {
+		config = NewFSConfig().WithFSMount(fs, "")
+	}
+	return c.WithFSConfig(config)
 }
 
 // WithFSConfig implements ModuleConfig.WithFSConfig
--- a/config_test.go
+++ b/config_test.go
@@ -361,6 +361,24 @@ func TestModuleConfig_toSysContext(t *testing.T) {
 			),
 		},
 		{
+			name:  "WithFS nil",
+			input: base.WithFS(nil),
+			expected: requireSysContext(t,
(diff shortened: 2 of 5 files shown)


codefromthecrypt (approved):
I think it is cleaner to handle this at the call site. The adapter is too late imho, and nil there a bug not an intended behavior. otoh keeping the nil guard in adapt is also fine!

so main change I prefer is to check nil inside here:
```
// WithFS implements ModuleConfig.WithFS
func (c *moduleConfig) WithFS(fs fs.FS) ModuleConfig {
	return c.WithFSConfig(NewFSConfig().WithFSMount(fs, ""))
}
```

to
```
// WithFS implements ModuleConfig.WithFS
func (c *moduleConfig) WithFS(fs fs.FS) ModuleConfig {
	var config FSConfig
	if fs != nil {
		config = NewFSConfig().WithFSMount(fs, "")
	}
	return c.WithFSConfig(config)
}
```

evacchi:
I originally did it that way and figured that it would do the same if I handled at Adapt; but it's ok to be a tad more defensive. I pushed the following, which avoids invoking `withFSConfig()` when fs is nil

```go
// WithFS implements ModuleConfig.WithFS
func (c *moduleConfig) WithFS(fs fs.FS) ModuleConfig {
	if fs == nil {
		return c
	}
	return c.WithFSConfig(NewFSConfig().WithFSMount(fs, ""))
}
```

codefromthecrypt:
sorry to be a nitpicker, but any WithXXX(nil) should clear previous don't you think?

evacchi:
oh right I didn't think about that case!

codefromthecrypt (approved):
Thanks for bearing with me!

PR #1187, opened 2023-03-02
refactor binary encoding to its own package

evacchi:
Binary encoding is only for testing. We move the encoder to its own package `binaryencoding` and fix all the related references.

As a strategy:
- I have kept all constants (except Magic) private (i.e. I duplicated them). To be honest, there were just a few in names.go
- I have exported the encoding methods that are used within decoding tests


internal/wasm/module.go
```diff
@@ -29,7 +29,7 @@ type DecodeModule func(
 ) (result *Module, err error)
 
 // EncodeModule encodes the given module into a byte slice depending on the format of the implementation.
-// See binary.EncodeModule
+// See binaryencoding.EncodeModule
 type EncodeModule func(m *Module) (bytes []byte)
```
evacchi, line 33: This looks like some sort of leftover. I can't really see any other references to this. Remove?

codefromthecrypt (approved):
looks good to me on green. only nit I would suggest is if the package isn't binary, maybe binaryformat as then we can blame the arbitrary on this heading :) https://www.w3.org/TR/wasm-core-1/#binary-format%E2%91%A0

appreciate the refactoring!

internal/wasm/module.go
codefromthecrypt: if there are no references to something internal and the build passes, delete it.

evacchi:
only issue: we are already aliasing `binary` to `binaryformat` in some places, for instance https://github.com/tetratelabs/wazero/blob/92e0a488b6e2134d77383a7772ed09a23792f51e/internal/integration_test/spectest/spectest.go#L19

evacchi:
last forced push fixed `make check`

codefromthecrypt:
agree on duplicating when there are only a few symbols. this is go idiomatic also as in general you'll see this even in go about duplication before generalization.

codefromthecrypt:
PS thanks for the note about binaryformat already in use, and thanks for checking. These are internal so we can play around later as needed, but it isn't terribly important either.

PR #1189, opened 2023-03-02
ci: do not fail the Windows build if the certificate is missing

evacchi:
The CI fails if the secret key for the Windows certificate is not set.

We should fall back to the default value instead -- or, even better, for routine checks, we may _always_ use the default key.

I am submitting this patch, but feel free to close and resolve it in a better way, this is mostly to show where the issue is.

In particular, this would silently fall-back to the default key even when we want it to actually fail!

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -50,15 +50,23 @@ jobs:
         run: sudo apt update -qq && sudo apt install -qq -y wixl osslsigncode
 
       - name: Download Windows code signing certificate
+        env:
+          WINDOWS_CODESIGN_P12_BASE64: ${{ secrets.WINDOWS_CODESIGN_P12_BASE64 }}
         run: |
-          echo ${{ secrets.WINDOWS_CODESIGN_P12_BASE64 }} | base64 --decode > windows-certificate.p12
-          echo "WINDOWS_CODESIGN_P12=windows-certificate.p12" >> $GITHUB_ENV
+          if [ $WINDOWS_CODESIGN_P12_BASE64 ]; then
+            echo $WINDOWS_CODESIGN_P12_BASE64 | base64 --decode > windows-certificate.p12
+            echo "WINDOWS_CODESIGN_P12=windows-certificate.p12" >> $GITHUB_ENV
+          fi
 
       - name: Make artifacts (test)
         if: github.event_name != 'push' || !contains(github.ref, 'refs/tags/')
         env:
+          WINDOWS_CODESIGN_P12_BASE64: ${{ secrets.WINDOWS_CODESIGN_P12_BASE64 }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -50,15 +50,23 @@ jobs:
         run: sudo apt update -qq && sudo apt install -qq -y wixl osslsigncode
 
       - name: Download Windows code signing certificate
+        env:
+          WINDOWS_CODESIGN_P12_BASE64: ${{ secrets.WINDOWS_CODESIGN_P12_BASE64 }}
```
evacchi, line 54: use an env var so we are sure not to leak its contents to the log

.github/workflows/release.yaml
```diff
@@ -50,15 +50,23 @@ jobs:
...
+          fi
 
       - name: Make artifacts (test)
         if: github.event_name != 'push' || !contains(github.ref, 'refs/tags/')
         env:
+          WINDOWS_CODESIGN_P12_BASE64: ${{ secrets.WINDOWS_CODESIGN_P12_BASE64 }}
           WINDOWS_CODESIGN_PASSWORD: ${{ secrets.WINDOWS_CODESIGN_PASSWORD }}
         run: |
+          if [ -n $WINDOWS_CODESIGN_P12_BASE64 ]; then
+            unset WINDOWS_CODESIGN_PASSWORD
```
evacchi, line 68: unset the password so the makefile uses the default value

codefromthecrypt:
I think it is a security thing that secrets aren't available from forked PRs. thanks for raising this and reviewing now!

.github/workflows/release.yaml
```diff
@@ -67,8 +75,12 @@ jobs:
         # Triggers only on tag creation.
         if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
         env:
+          WINDOWS_CODESIGN_P12_BASE64: ${{ secrets.WINDOWS_CODESIGN_P12_BASE64 }}
           WINDOWS_CODESIGN_PASSWORD: ${{ secrets.WINDOWS_CODESIGN_PASSWORD }}
         run: | # Note: MSI_VERSION requires . as a separator, so replace "-" in the tag with ".".
+          if [ -n $WINDOWS_CODESIGN_P12_BASE64 ]; then
+            unset WINDOWS_CODESIGN_PASSWORD
```
mathetake, line 82: this shouldn't happen as this only runs on main repo's tag cut.

codefromthecrypt: feel free to raise a PR to unglitch this, I think you need the commits if you want to stay 🥇 :D

mathetake:
turns out this broke the original repo commit. working on the fix now

evacchi:
yay my first screw-up! 🤣

codefromthecrypt:
> yay my first screw-up! 🤣

I lost count of mine :)

PR #1191, opened 2023-03-02
ci: polishes fork PR protect

mathetake:


codefromthecrypt (approved):
thanks especially for the comments!

PR #1192, opened 2023-03-03
ci: remove -short flag on main branch test

mathetake:
why we have been using `-short` on main..


mathetake:
ah -  `-race` without short could be exhausting

PR #1193, opened 2023-03-03
ci: fixes CLI release flow on main repository

mathetake:


.github/workflows/release.yaml
```diff
@@ -60,12 +60,9 @@ jobs:
...
       - name: Make artifacts (test)
         if: github.event_name != 'push' || !contains(github.ref, 'refs/tags/')
-        env:
-          WINDOWS_CODESIGN_P12_BASE64: ${{ secrets.WINDOWS_CODESIGN_P12_BASE64 }}
-          WINDOWS_CODESIGN_PASSWORD: ${{ secrets.WINDOWS_CODESIGN_PASSWORD }}
         run: | # On the fork PRs, our org secret is not visible. We unset the required env so that `make dist` uses default self-signed cert.
-          if [ -n $WINDOWS_CODESIGN_P12_BASE64 ]; then
-            unset WINDOWS_CODESIGN_PASSWORD
+          if [ $WINDOWS_CODESIGN_P12 ]; then
+            export WINDOWS_CODESIGN_PASSWORD=${{ secrets.WINDOWS_CODESIGN_PASSWORD }}
```
codefromthecrypt, line 65: maybe echo into env conditionally like https://github.com/tetratelabs/func-e/blob/68c7393a0f65334e88006fac9865c01a3866626f/.github/workflows/internal-images.yml#L68-L76?

mathetake: I think this is more readable, so I prefer not, but not strong.

mathetake:
https://github.com/tetratelabs/wazero/pull/1194 verified to work on fork PR as well.

.github/workflows/release.yaml
mathetake: and this is only used in this bash scope, not the subsequent one, so I don't think adding to GITHUB_ENV work heer.

codefromthecrypt: makes sense!

codefromthecrypt (commented):
👍

.github/workflows/release.yaml
evacchi: the reason this was inverted is because I wanted to make sure the log didn't echo the expanded version, since the substitution of `${{ secrets.WINDOWS_CODESIGN_PASSWORD }}` occurs before execution. Using `env:` prevents that. 

However it looks like the log is smart enough to censor that 🤔 

https://github.com/tetratelabs/wazero/actions/runs/4319624179/jobs/7539006689#step:7:2

PR #1199, opened 2023-03-05
wasi: implements platform.UtimesFile for fd_filestat_set_times

codefromthecrypt:
This implements `platform.UtimesFile` which is similar to futimes. Before, we were using path-based functionality even though the call site was for a file descriptor.

Note: right now, there's no obvious code in Go to invoke the `futimens` syscall. This means non-windows isn't implemented at nanos granularity, so ends up falling back to the path based option.

Finally, this removes tests for the seldom supported updates with negative epoch time. There's little impact to this as setting times on files before 1970 isn't a typical use case.

### DIFF
--- a/internal/platform/utimes.go
+++ b/internal/platform/utimes.go
@@ -1,0 +1,35 @@
+package platform
+
+import (
+	"io/fs"
+	"syscall"
+)
+
+// UtimesNano is like syscall.UtimesNano. This returns syscall.ENOENT if the
+// path doesn't exist.
+//
+// Note: This is like the function `utimensat` with `AT_FDCWD` in POSIX.
+// See https://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html
+func UtimesNano(path string, atimeNsec, mtimeNsec int64) error {
+	err := syscall.UtimesNano(path, []syscall.Timespec{
+		syscall.NsecToTimespec(atimeNsec),
+		syscall.NsecToTimespec(mtimeNsec),
+	})
+	return UnwrapOSError(err)
+}
+
+// UtimesNanoFile is like syscall.Futimes, but for nanosecond precision and
+// fs.File instead of a file descriptor. This returns syscall.EBADF if the file
+// or directory was closed, or syscall.EPERM if the file wasn't opened with
+// permission to update its utimes. On syscall.EPERM, or on syscall.ENOSYS, use
+// UtimesNano with the original path.
+//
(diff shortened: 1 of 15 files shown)


internal/sysfs/sysfs.go
```diff
@@ -238,8 +238,8 @@ type FS interface {
 	//   - syscall.EACCES: `path` doesn't have write access.
 	Truncate(path string, size int64) error
 
-	// Utimes is similar to syscall.UtimesNano, except the path is relative to
-	// this file system.
+	// UtimesNano is similar to syscall.UtimesNano, except the path is relative
```
codefromthecrypt, line 241: renamed because I forgot the default utimes was microsecond resolution

internal/platform/utimes_unsupported.go
```diff
@@ -0,0 +1,12 @@
+//go:build !windows
+
+package platform
+
+import "syscall"
+
+func futimens(fd uintptr, atimeNsec, mtimeNsec int64) error {
+	// Go exports syscall.Futimes, which is microsecond granularity, and
```
codefromthecrypt, line 8: @evacchi are you interested in trying to implement calling futimens on darwin, freebsd and linux? It would be cleaner for sure.

evacchi: sure why not 💪

codefromthecrypt:
rebased

internal/platform/utimes.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"io/fs"
+	"syscall"
+)
+
+// UtimesNano is like syscall.UtimesNano. This returns syscall.ENOENT if the
+// path doesn't exist.
+//
+// See https://linux.die.net/man/3/futimens
```
codefromthecrypt, line 11: ps I'm thinking we change to posix for links like this instead of man, since we're often told to look at posix when in doubt. wdyt @evacchi @mathetake https://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html

internal/platform/utimes_unsupported.go
codefromthecrypt: #1205 will resolve this!

codefromthecrypt:
ok I've reworked this as windows is finicky. I want to land this before the improvement, as basically this shows how dire making our own file type is (particularly in windows you need path to do basically anything directory based). After this, I can start working on our file type.

mathetake (approved):

PR #1203, opened 2023-03-06
wazeroir: implements fmt.Stringer for operations

mathetake:
Previously `wazeroir.Format([]wazeroir.Operation)` has only been partially implemented, and not consistent with `OperationKind.String()`. That was one of the tech debts.


internal/wazeroir/format.go
```diff
@@ -14,193 +11,13 @@ func Format(ops []Operation) string {
 
 	_, _ = buf.WriteString(EntrypointLabel + "\n")
 	for _, op := range ops {
-		formatOperation(buf, op)
-	}
-
-	return buf.String()
-}
-
-func formatOperation(w io.StringWriter, b Operation) {
```
mathetake, line 23: this format was inconsistent with OperationKind.String() and clearly was a garbage.

internal/wazeroir/compiler_test.go
```diff
@@ -3216,6 +3216,7 @@ func Test_ensureTermination(t *testing.T) {
 			}
 			res, err := CompileFunctions(api.CoreFeaturesV2, 0, mod, tc.ensureTermination)
 			require.NoError(t, err)
+			fmt.Println(Format(res[0].Operations))
```
codefromthecrypt, line 3219:
```suggestion
```

codefromthecrypt (approved):
thanks. This is a lot better

PR #1205, opened 2023-03-06
implement futimens on Darwin and Linux

evacchi:
I have tested this on Fedora 37 (x86) and macOS Ventura (arm64). 

The `futimens` syscall is not present in versions before macOS High Sierra, so in that case we would have to implement in terms of e.g. fsetattrlist (e.g. see https://github.com/bytecodealliance/rustix/pull/275) but this is still an advancement 

this is a PR to @codefromthecrypt's branch


internal/platform/utimes_linux.go
```diff
@@ -0,0 +1,20 @@
+//go:build !darwin
```
codefromthecrypt, line 1: remove as the file name convention already takes care of this.

internal/platform/utimes_darwin.go
```diff
@@ -0,0 +1,28 @@
+package platform
+
+import (
+	"syscall"
+	"unsafe"
+)
+
+var libc_futimens_trampoline_addr uintptr
+
+//go:cgo_import_dynamic libc_futimens futimens "/usr/lib/libSystem.B.dylib"
```
codefromthecrypt, line 10: if this is a note, maybe make a comment above and tab-indent the go:cgo_import_dynamic, if this is a real cgo import, which I think it is, see the below comment :D

internal/platform/utimes_darwin.go
```diff
@@ -0,0 +1,28 @@
...
+	if e1 != 0 {
+		return e1
+	}
+	return nil
+}
+
+// we need to use this instead of syscall.Syscall6
+func syscall_syscall6(fn, a1, a2, a3, a4, a5, a6 uintptr) (r1, r2 uintptr, err syscall.Errno)
+
+//go:linkname syscall_syscall6 syscall.syscall6
```
codefromthecrypt, line 28: similarly I think this is a cgo pragma which usually requires this and the file being dependent on a build flag or naming convention _cgo. If a note, you would put it into a comment and tab-indent.

Otherwise, format like time_cgo to make all obvious. such as

```go
//go:build cgo && !windows

package platform

import _ "unsafe" // for go:linkname

// nanotime uses runtime.nanotime as it is available on all platforms and
// benchmarks faster than using time.Since.
//
//go:noescape
//go:linkname nanotime runtime.nanotime
func nanotime() int64
```

internal/platform/utimes_unsupported.go
```diff
@@ -1,4 +1,4 @@
-//go:build !windows
+//go:build !windows && !linux && !darwin
```
codefromthecrypt, line 1: I think the last will need to change to `&& !(darwin && cgo)`

internal/platform/utimes_darwin.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"syscall"
+	"unsafe"
+)
+
+var libc_futimens_trampoline_addr uintptr
+
+//go:cgo_import_dynamic libc_futimens futimens "/usr/lib/libSystem.B.dylib"
+
+func futimens(fd uintptr, atimeNsec, mtimeNsec int64) error {
```
codefromthecrypt, line 12: add any comments here about version requirements (high sierra)

codefromthecrypt (commented):
correct me if I'm wrong at assuming cgo is needed for this. maybe you can temporarily set the build to also test macOS 11 to see what the error will be and if ENOSYS is returned in that case or if it all breaks?

codefromthecrypt:
you can rebase and reopen against main, now. hopefully github will allow you to reopen this

internal/platform/utimes_darwin.go
evacchi: yes this is required dynamically link against libc, import `futimens` and populate `var libc_futimens_trampoline_addr` with the address of the trampoline that is defined in the companion `.s` file

PR #1206, opened 2023-03-07
doc: adds wazero architecture

### LINKED ISSUE #931: Document wazero architecture
Right now, there's no document covering the architecture of wazero, and leaving implicit assumptions slows engagement on how to proceed certain topics, and above all an understanding of what's in the box.

We should have maybe on the website, a page for wazero architecture. It should have diagrams, so if we do it on the website we can use the stuff we render in Hugo.

Here are some notes I made when @Xe asked today

compiledmodule has the static stuff, including function bytes and the intermediate representation. compiled module is mmapped and reused for each instance of it

instantiated module has in its scope a memory (or imported) table etc. and the parts needed to satisfy wasi (rng etc assigned with ModuleConfig or defaulted) generally we have a XXXInstance type for things in the instantiated module

there's an even smaller scope for function call graphs called "call engine"

mathetake:
closes #931

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,87 @@
+# The architecture of wazero
+
+This documents describes the internal architecture of wazero for those who are interested in how it works, especially
+who want to contribute to wazero's source code. For technical decisions, and rationales for end users, please refer to [RATIONALE.md][rationale].
+
+## wazero.CompiledModule, api.Module and api.Function
+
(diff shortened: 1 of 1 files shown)


mathetake:
Will do the website on post merge

codefromthecrypt (commented):
starter note: move this into site and change the doc to render with goat like the languages page. put the languages page into the appendix section and make this page top-level "Docs" and architecture being a heading. Make a simple "Overview" heading before it, which I can help flesh out later.

PR #1210, opened 2023-03-07
implement futimens on Darwin and Linux

evacchi:

### DIFF
--- a/internal/platform/utimes_darwin.go
+++ b/internal/platform/utimes_darwin.go
@@ -1,0 +1,34 @@
+package platform
+
+import (
+	"syscall"
+	"unsafe"
+)
+
+func futimens(fd uintptr, atimeNsec, mtimeNsec int64) error {
+	tv := []syscall.Timespec{
+		syscall.NsecToTimespec(atimeNsec),
+		syscall.NsecToTimespec(mtimeNsec),
+	}
+
+	// Warning: futimens only exists since High Sierra (10.13).
+	_, _, e1 := syscall_syscall6(libc_futimens_trampoline_addr, fd, uintptr(unsafe.Pointer(&tv[0])), 0, 0, 0, 0)
+	if e1 != 0 {
+		return e1
+	}
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


internal/platform/utimes_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+	return nil
+}
+
+// libc_futimens_trampoline_addr is the address of the libc_futimens_trampoline symbol, defined in utimes_darwin.s
+// we use this value to invoke the syscall through syscall_syscall6 imported below
+var libc_futimens_trampoline_addr uintptr
+
+// Imports the futimens symbol from libc as libc_futimens
+//go:cgo_import_dynamic libc_futimens futimens "/usr/lib/libSystem.B.dylib"
```
codefromthecrypt, line 26: so go:cgo_import_dynamic doesn't need cgo?

.github/workflows/commit.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
     strategy:
       fail-fast: false  # don't fail fast as sometimes failures are arch/OS specific
       matrix:  # Use versions consistent with wazero's Go support policy.
-        os: [ubuntu-22.04, macos-12, windows-2022]
+        os: [ubuntu-22.04, macos-11, macos-12, windows-2022]
```
evacchi, line 65: the macos-11 run was successful but I am going to remove it from CI, I just wanted to give it a go for double-check

internal/platform/utimes_darwin.go
codefromthecrypt: will do a follow-up doc change, as @mathetake said "regardless of CGO_ENABLED, Go always uses “CGO” mechanism in the runtime layer (of darwin), because that's where Apple provides the compatibility layer." put another way, you can't build a statically linked binary for darwin.

.github/workflows/commit.yaml
codefromthecrypt: thanks!

PR #1213, opened 2023-03-08
Pass correct api.Module to host functions

### LINKED ISSUE #1211: Executing a Function from another Module
Executing a host function from a function called from another module returns a pointer to original module. Memory address referenced from the other module is not accessible, as the pointer is from the original module.

```go
//go:wasm-module host
//export printModuleName()
func printModuleName()


//export function1
func function1() {
	printModuleName()
}
```

```go
//go:wasm-module module1
//export function1
func function1()

func function2() {
	function1()
}
```

```go
func printModuleName(ctx, module) error {
	fmt.Println(module.Name())
}

````

calling function2 from module1
```go
module2.ExportedFunction("function2").Call()
```

prints

```
module2
```

Thus memory addressed in module1 are not accessible from function1 being called by function2. 

Not sure if this is intended or not.

mathetake:
Fixes #1211 

Previously, host functions are getting `api.Module` for the "originating" module, which is the module for api.Function currently invoked, except that the api.Module is modified withMemory for the caller's memory instance, therefore there haven't been no problem. The only issues were the methods except `Memory()` of api.Module, and this commit fixes them.

### DIFF
--- a/internal/engine/compiler/compiler_controlflow_test.go
+++ b/internal/engine/compiler/compiler_controlflow_test.go
@@ -17,6 +17,8 @@ func TestCompiler_compileHostFunction(t *testing.T) {
 	err := compiler.compileGoDefinedHostFunction()
 	require.NoError(t, err)
 
+	// Get the location of caller function's location stored in the stack, which depends on the type.
+	// In this test, the host function has empty sig.
 	_, _, callerFuncLoc := compiler.runtimeValueLocationStack().getCallFrameLocations(&wasm.FunctionType{})
 
 	// Generate the machine code for the test.
--- a/internal/engine/compiler/compiler_controlflow_test.go
+++ b/internal/engine/compiler/compiler_controlflow_test.go
@@ -17,13 +17,21 @@ func TestCompiler_compileHostFunction(t *testing.T) {
 	err := compiler.compileGoDefinedHostFunction()
 	require.NoError(t, err)
 
-	// Generate and run the code under test.
+	_, _, callerFuncLoc := compiler.runtimeValueLocationStack().getCallFrameLocations(&wasm.FunctionType{})
+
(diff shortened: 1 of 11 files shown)


mathetake:
will run 2hr fuzzing. I will make it ready after it finishes.

mathetake:
both on amr64 and amd64 of course!

mathetake:
arm64

internal/engine/compiler/compiler_controlflow_test.go
```diff
@@ -17,13 +17,21 @@ func TestCompiler_compileHostFunction(t *testing.T) {
 	err := compiler.compileGoDefinedHostFunction()
 	require.NoError(t, err)
 
-	// Generate and run the code under test.
+	_, _, callerFuncLoc := compiler.runtimeValueLocationStack().getCallFrameLocations(&wasm.FunctionType{})
```
codefromthecrypt, line 20: maybe comment on what this line is doing or if the empty ft is relevant

internal/engine/compiler/engine.go
```diff
@@ -925,7 +929,7 @@ entry:
 			fn := calleeHostFunction.parent.goFunc
 			switch fn := fn.(type) {
 			case api.GoModuleFunction:
-				fn.Call(ce.ctx, callCtx.WithMemory(ce.memoryInstance), stack)
+				fn.Call(ce.ctx, ce.callerFunctionInstance.Module.CallCtx.WithMemory(ce.memoryInstance), stack)
```
codefromthecrypt, line 932: due to the chain here, feels like a comment is needed, maybe extracting as a variable.

codefromthecrypt (approved):
Looks good and fast turnaround. maybe a bit of a summary in the PR desc

PR #1214, opened 2023-03-08
Deletes callContext.withMemory

mathetake:
Follow up on #1213. We no longer need withMemory.


codefromthecrypt (approved):
awesome. I like code in red. It is a nice color

PR #1215, opened 2023-03-08
platform: Allows sysfs to implement utimens natively

codefromthecrypt:
This moves away from `syscall.UtimesNano` as it has intentionally avoided common features in POSIX, such as handling UTIME_NOW and UTIME_OMIT. When we eventually expose this API, users will be free to override `UTIME_NOW` with a fake clock, possibly the same that was supplied to wazero's `ModuleConfig`.

Notes:
* Once merged, I will implement `path_filestat_set_times` in wasi, which is literally the only fs function left stubbed in WASI.
* Like elsewhere, windows support is dodgy with at least Go 1.18, so we skip tests on that. It may be that Go 1.19 is fine, but for now only supporting 1.20.
* Even on Go 1.20, there is a limitation on windows as we can't update timestamps on an opened directory as it requires O_RDWR which isn't valid for them, also it might not work anyway as other ops require path-based access. This isn't a big deal as we'll shortly after make a file type which caches the path for a natural alternative.


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -435,70 +435,38 @@ func fdFilestatSetTimesFn(_ context.Context, mod api.Module, params []uint64) Er
 		return ErrnoBadf
 	}
 
-	// Unchanging a part of time spec while executing utimes is extremely complex to add support for all platforms,
```
codefromthecrypt, line 438: so the key change here is moving this part outside of wasi preview1 to a place we can re-use it everywhere including wasi-filesystem

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -435,70 +435,38 @@ func fdFilestatSetTimesFn(_ context.Context, mod api.Module, params []uint64) Er
...
-	// and actually there's an outstanding issue on Go
-	// - https://github.com/golang/go/issues/32558.
-	// - https://go-review.googlesource.com/c/go/+/219638 (unmerged)
-	//
-	// Here, we emulate the behavior for empty flag (meaning "do not change") by get the current time stamp
-	// by explicitly executing File.Stat() prior to UtimesNano.
-	var atime, mtime int64
-	var nowAtime, statAtime, nowMtime, statMtime bool
+	// times[0] == atim, times[1] == mtim
+	times := [2]syscall.Timespec{}
```
codefromthecrypt, line 439: luckily this seems to be defined everywhere

internal/platform/utimes.go
```diff
@@ -3,33 +3,140 @@ package platform
...
+//   - syscall.ENOTDIR: `path` exists and is a file.
+//
+// # Notes
+//
+//   - This is similar to syscall.UtimesNano, except that doesn't have flags to
+//     control expansion of symbolic links. It also doesn't support special
+//     values UTIME_NOW or UTIME_NOW.
+//   - This is like `utimensat` with `AT_FDCWD` in POSIX. See
+//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html
+func Utimesns(path string, times *[2]syscall.Timespec, symlinkFollow bool) error {
```
codefromthecrypt, line 48: pointer signature allows us to tell nil from not. symlinkFollow is lifted from WASI in recognition there's only one real flag to handle

internal/platform/utimes_linux.go
```diff
@@ -2,17 +2,27 @@ package platform
...
+	if !symlinkFollow {
+		flags = 0
 	}
-	return nil
+	return utimensat(_AT_FDCWD, path, times, flags)
+}
+
+// On linux, implement futimens via utimensat with the empty path.
+func futimens(fd uintptr, times *[2]syscall.Timespec) error {
+	return utimensat(int(fd), "", times, 0)
```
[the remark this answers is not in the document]
evacchi: the change above fixes the tests in utimes_test.go on Linux, but does not fix the issues with futimens (fs_test.go), and that is probably due to some shenanigans with the special values `UTIME_NOW` and `UTIME_OMIT`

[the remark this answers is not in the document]
codefromthecrypt: thanks for the progress, and ps typos in apis are not nits.. so hard to fix later if ever :)

codefromthecrypt:
ok only windows left

mathetake (approved):

codefromthecrypt:
Thanks @evacchi for the help overnight my time as that's the only reason we are able to land this your morning!

PR #1220, opened 2023-03-12
wasi: implements path_filestat_set_times

codefromthecrypt:
This implements `path_filestat_set_times` which is the last remaining filesystem function in wasi we formerly stubbed.

Other minor changes:
* this removes all places we import alias path as pathutil, introducing a utility function `joinPath` where that was used to reduce name conflicts.
* this fixes places where we used different variable names for the same parameter between main and test source.

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -1508,11 +1508,11 @@ func pathLinkFn(_ context.Context, mod api.Module, params []uint64) Errno {
 		return errno
 	}
 
-	newFd := uint32(params[4])
+	newFD := uint32(params[4])
 	newPath := uint32(params[5])
 	newPathLen := uint32(params[6])
 
-	newFS, newName, errno := atPath(fsc, mem, newFd, newPath, newPathLen)
+	newFS, newName, errno := atPath(fsc, mem, newFD, newPath, newPathLen)
 	if errno != ErrnoSuccess {
 		return errno
 	}
@@ -1675,7 +1675,8 @@ func atPath(fsc *sys.FSContext, mem api.Memory, fd, path, pathLen uint32) (sysfs
 	} else if f.IsPreopen { // don't append the pre-open name
 		return f.FS, pathName, ErrnoSuccess
 	} else {
-		return f.FS, joinPath(f.Name, pathName), ErrnoSuccess
+		// Join via concat to avoid name conflict on path.Join
+		return f.FS, f.Name + "/" + pathName, ErrnoSuccess
 	}
 }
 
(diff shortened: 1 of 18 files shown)


imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -3227,13 +3230,212 @@ func Test_pathFilestatGet(t *testing.T) {
 	}
 }
 
-// Test_pathFilestatSetTimes only tests it is stubbed for GrainLang per #271
 func Test_pathFilestatSetTimes(t *testing.T) {
```
codefromthecrypt, line 3233: these are the new tests

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -787,42 +788,44 @@ func Test_fdFilestatSetTimes(t *testing.T) {
 			expectedErrno: ErrnoSuccess,
 			mtime:         1234, // Must be ignored.
 			atime:         55555500,
-			flags:         FileStatAdjustFlagsAtim | FileStatAdjustFlagsMtimNow,
+			flags:         FstflagsAtim | FstflagsMtimNow,
 			expectedLog: `
-==> wasi_snapshot_preview1.fd_filestat_set_times(fd=4,atim=55555500,mtim=1234,fst_flags=9)
+==> wasi_snapshot_preview1.fd_filestat_set_times(fd=4,atim=55555500,mtim=1234,fst_flags=ATIM|MTIM_NOW)
```
codefromthecrypt, line 793: other minor thing was pretty-printing the flags when logging

site/content/specs.md
```diff
@@ -113,7 +113,7 @@ Notes:
 | fd_write                |   ✅    | Rust,TinyGo,Zig |
 | path_create_directory   |   ✅    | Rust,TinyGo,Zig |
 | path_filestat_get       |   ✅    | Rust,TinyGo,Zig |
-| path_filestat_set_times |   ❌    |                 |
+| path_filestat_set_times |   ✅    |            libc |
```
mathetake, line 116: what's the difference with libc vs Rust? I see libc is only here

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -8,8 +8,9 @@ import (
 	"io/fs"
 	"math"
 	"os"
-	"path"
+	pathutil "path"
```
mathetake, line 11: I prefer not to use alias..

codefromthecrypt: we do this in many files, and we do this because we have fields literally named path in specs. do you want me to revert all of them or just here?

codefromthecrypt: the other impact is it forces us to choose dumb names like pathName etc.

mathetake: ok fine.

site/content/specs.md
codefromthecrypt: actually you are right, rust uses this. https://github.com/rust-lang/rust/blob/master/library/std/src/sys/wasi/fd.rs#L196

imports/wasi_snapshot_preview1/fs_test.go
codefromthecrypt: I think on a whole things are worse, but doing it.

codefromthecrypt: done

site/content/specs.md
codefromthecrypt: cited

codefromthecrypt:
@mathetake PTAL I added a utility function `joinPath` to reduce the main problem around path.Join, so nothing aliases it anymore.

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -1349,11 +1360,11 @@ var pathCreateDirectory = newHostFunc(
 func pathCreateDirectoryFn(_ context.Context, mod api.Module, params []uint64) Errno {
 	fsc := mod.(*wasm.CallContext).Sys.FS()
 
-	dirFD := uint32(params[0])
```
codefromthecrypt, line 1352: some places we accidentally used dirFD which was a test property, in the main tree, where the actual param is called `fd`.

mathetake (approved):
nice👍

Issue #1222, opened 2023-03-13
gojs: resolve or explain remaining os package failures


codefromthecrypt:
Before extracting a file system interface, we should know why any remaining failures exist on the standard go compiler.

There are currently 23 tests that don't pass still, some might be legitimate as they aren't covered in tests such as wasi-testsuite.
```bash
$ GOARCH=wasm GOOS=js go test -c -o os.wasm os
$ wazero run -mount=/usr/local/go/src/os:/:ro -mount=/tmp:/tmp -mount=/etc:/etc:ro -mount=/usr:/usr:ro -mount=/dev:/dev:ro os.wasm | grep '^--- FAIL'

--- FAIL: TestStatError (0.00s)
--- FAIL: TestStatSymlinkLoop (0.00s)
--- FAIL: TestHardLink (0.00s)
--- FAIL: TestSymlink (0.00s)
--- FAIL: TestLongSymlink (0.00s)
--- FAIL: TestRename (0.00s)
--- FAIL: TestRenameOverwriteDest (0.00s)
--- FAIL: TestRenameNotExisting (0.00s)
--- FAIL: TestRenameCaseDifference (0.00s)
--- FAIL: TestFileChdir (0.00s)
--- FAIL: TestOpenError (0.00s)
--- FAIL: TestWriteAtInAppendMode (0.00s)
--- FAIL: TestAppend (0.00s)
--- FAIL: TestSameFile (0.00s)
--- FAIL: TestDevNullFile (0.00s)
--- FAIL: TestStatRelativeSymlink (0.00s)
--- FAIL: TestDirFSRootDir (0.00s)
--- FAIL: TestMkdirStickyUmask (0.00s)
--- FAIL: TestMkdirAllWithSymlink (0.00s)
--- FAIL: TestMkdirAllAtSlash (0.00s)
--- FAIL: TestDirAndSymlinkStats (0.01s)
--- FAIL: TestFileAndSymlinkStats (0.01s)
--- FAIL: TestSymlinkWithTrailingSlash (0.00s)
```

codefromthecrypt:
currently down to 12. There are some nuance in the args to pass. For example, we shouldn't mask any root mounts like /etc as read-only as that affects errors and causes some to fail for a different reason.

```bash
[...]

PR #1226, opened 2023-03-13
gojs: introduces `--experimental-workdir` CLI arg

codefromthecrypt:
When compiled to `GOOS=js`, wasm does not maintain the working directory: it is defined by the host. While not explicitly documented, [os.TestDirFSRootDir](https://github.com/golang/go/blob/master/src/os/os_test.go#L2841) in Go suggests the working directory must be valid to pass (literally the directory holding the file).

This adds an experimental CLI flag that gives the initial working directory. This is experimental because while GOOS=js uses this, current WASI compilers will not, as they maintain working directory in code managed by wasi-libc, or as a convention (e.g. in Zig).

It is not yet known if [wasi-cli](https://github.com/WebAssembly/wasi-cli) will maintain working directory externally or not.

After this change, our GOOS=js implementation is down to less than ten test failures.
```bash
$ wazero run -mount=/:/ --experimental-workdir=$(go env GOROOT)/src/os os.wasm|grep '^--- FAIL'|wc -l
       9
```

See #1222

### DIFF
--- a/cmd/wazero/wazero.go
+++ b/cmd/wazero/wazero.go
@@ -123,10 +123,15 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
 	flags.Var(&envs, "env", "key=value pair of environment variable to expose to the binary. "+
 		"Can be specified multiple times.")
 
-	var envExport bool
-	flags.BoolVar(&envExport, "env-inherit", false,
+	var envInherit bool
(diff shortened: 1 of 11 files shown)


mathetake (approved):

codefromthecrypt:
I don't have my windows laptop with me, but I suspect the path needs to be posix and isn't... will shotgun a try

PR #1228, opened 2023-03-13
dirfs: corrects path joining bug when mounting root

codefromthecrypt:
Before, we created invalid directory paths when the host directory was "/", and now we don't!

This brings tests left failing on GOOS=js to 7
```bash
$ wazero run -mount=/:/ --experimental-workdir=$(go env GOROOT)/src/os os.wasm| grep '^--- FAIL'|wc -l
       7
```

See #1222


mathetake (approved):

codefromthecrypt:
windows again.. heh ok fine :D

PR #1231, opened 2023-03-13
optimize: requireModules return ModuleInstance maybe empty

1046102779:


mathetake:
Could you tell us what’s the specific problem you encountered and how this PR fixes that? Also we definitely need unittest.

mathetake:
Also, we you say “optimize”, we would like to see the benchmark result along with the concrete use case.

mathetake:
@1046102779 feel free to reopen if you're still up for this PR.

PR #1233, opened 2023-03-13
fix: --experimental-workdir test case on Windows

evacchi:
Interesting issue due to a lot of weird interactions between
path handling and system defaults.

The test will pass outside of CI (e.g. on my local machine),
because usually you will run this on C:\, which is the root
of your filesystem.

However, GitHub Actions mounts the test filesystem on D:,
causing the path handling routines to miss it.

You can simulate this even if you don't have a second drive:
assuming you have `wazero` under `%USERPROFILE%\Devel\wazero`

    subst X: %USERPROFILE%\Devel
    cd X:\wazero
    make test

However, we are ignoring path information because:

```go
func WithWorkdir(ctx context.Context, workdir string) context.Context {
	// Ensure if used on windows, the input path is translated to a POSIX one.
	workdir = workdir[len(filepath.VolumeName(workdir)):]
```

strips the Volume

The other weird bit happens in the test itself:

```
--mount=/:/ --experimental-workdir=$bearDir
```

What is happening here is that we are mount the root of our FS
to / (thereby exposing the entire FS to the test) and then
telling the guest to set the CWD to `$bearDir`.

However, on Windows we don't have a unique `/` (well technically
we do; see [UNC Paths][0] but that's another story) but each `/` has a volume attached.

So, on Windows we shouldn't `--mount=/:/` but rather one root volume
(in fact, the other tests work because they do `--mount=$bearDir:/`)
i.e. X:\ or, more generally `VolumeName($bearDir)`

Finally, although `WithWorkdir` is stripping the volume name,
we should probably pass a path without the volume name or at least ensuring
that that CWD makes sense together with the host mount directive; i.e.

when we pass `--mount=x:\:/` we should either pass `--workdir=\wazero\...`
or ensure that the volume name matches `--workdir=x:\wazero\...` and
return an error otherwise.

For now I am only fixing the test.

[0]: https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats

### DIFF
--- a/cmd/wazero/wazero_test.go
+++ b/cmd/wazero/wazero_test.go
@@ -345,9 +345,13 @@ func TestRun(t *testing.T) {
 			expectedStdout: "pooh\n",
 		},
 		{
-			name:           "GOARCH=wasm GOOS=js workdir",
-			wasm:           wasmCatGo,
-			wazeroOpts:     []string{"--mount=/:/", fmt.Sprintf("--experimental-workdir=%s", bearDir)},
+			name: "GOARCH=wasm GOOS=js workdir",
+			wasm: wasmCatGo,
+			wazeroOpts: []string{
+				// --mount=X:\:/ on Windows, --mount=/:/ everywhere else
+				"--mount=" + filepath.VolumeName(bearDir) + string(os.PathSeparator) + ":/",
+				fmt.Sprintf("--experimental-workdir=%s", bearDir[len(filepath.VolumeName(bearDir)):]),
+			},
 			wasmArgs:       []string{"bear.txt"},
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
Thanks for getting to the bottom of this @evacchi!

FWIW the reason we have this property at all is to make sure we can be used as a go test runner, instead of node, like this https://github.com/golang/go/blob/master/misc/wasm/wasm_exec_node.js#L24

right now, node passes through the whole fs and the current work dir, though I'm not sure if this test runner works in windows ;) In any case I think the better way out is to special-case inheriting the CWD, but that's for a future PR. This is the correct fix for the code that's in main now!

PR #1234, opened 2023-03-13
gojs: fixes symlink

mathetake:
```
$ wazero run -mount=/:/ --experimental-workdir=$(go env GOROOT)/src/os os.wasm -test.run '^TestSymlink|^TestLongSymlink|^TestStatRelativeSymlink|^TestLongSymlink'
PASS
```

### DIFF
--- a/internal/gojs/fs.go
+++ b/internal/gojs/fs.go
@@ -711,7 +711,7 @@ func (jsfsLink) invoke(ctx context.Context, mod api.Module, args ...interface{})
 type jsfsSymlink struct{}
 
 func (jsfsSymlink) invoke(ctx context.Context, mod api.Module, args ...interface{}) (interface{}, error) {
-	path := resolvePath(ctx, args[0].(string))
+	path := args[0].(string)
 	link := resolvePath(ctx, args[1].(string))
 	callback := args[2].(funcWrapper)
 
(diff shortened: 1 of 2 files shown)


mathetake:
I quickly noticed this as I had exactly this issue when I was working on WASI's symlink :D

mathetake:
hmm
```
>> git diff            
diff --git a/internal/gojs/testdata/fs/main.go b/internal/gojs/testdata/fs/main.go
index 9c07f3dc..dcd45598 100644
--- a/internal/gojs/testdata/fs/main.go
+++ b/internal/gojs/testdata/fs/main.go
@@ -70,4 +70,10 @@ func testAdHoc() {
                log.Panicln(err)
        }
        fmt.Println("empty:" + string(b))
+
+       old := "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
+       err = os.Symlink(old, "/symlink.txt")
+       if err != nil {
+               log.Panicln(err)
+       }
 }

```

gives `panic: symlink 0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef /symlink.txt: not implemented on js`

mathetake:
ah got it

codefromthecrypt (approved):
ah seems familiar indeed!

PR #1235, opened 2023-03-14
cli: changes experimental workdir flag to inherit

codefromthecrypt:
This changes the experimental flag introduced yesterday to `-experimental-workdir-inherit=true`, to reduce edge cases needed to produce the same behavior as go's wasm test runner.

After this, we have the same failure count, though the latter is likely related to my machine:

```bash
$ wasm=$PWD/os.wasm; (cd $(go env GOROOT)/src/os; wazero run -mount=/:/ --experimental-workdir-inherit=true $wasm)
--- FAIL: TestMkdirAllAtSlash (0.00s)
    path_test.go:111: MkdirAll "/_go_os_test/dir": mkdir /_go_os_test: I/O error, I/O error
--- FAIL: TestOpenFileLimit (0.01s)
    rlimit_test.go:31: open rlimit.go: I/O error
FAIL
```

See https://github.com/golang/go/blob/master/misc/wasm/wasm_exec_node.js#L24

### DIFF
--- a/cmd/wazero/wazero.go
+++ b/cmd/wazero/wazero.go
@@ -125,18 +125,20 @@ func doRun(args []string, stdOut io.Writer, stdErr logging.Writer, exit func(cod
 
 	var envInherit bool
 	flags.BoolVar(&envInherit, "env-inherit", false,
-		"inherits any environment variables from the calling process."+
+		"inherits any environment variables from the calling process. "+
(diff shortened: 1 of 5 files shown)


mathetake (approved):

codefromthecrypt:
my workstation is working so I'll look into this fail

PR #1239, opened 2023-03-14
fix: poll_oneoff expected layout EventTypeFdRead, EventTypeFdWrite

evacchi:
While fds are indeed aligned at 4 bytes, the `subscription_u`
union's "`contents_offset`" is 8, so proper content starts
at +8+8 instead of +8+4.


iluminae:
@evacchi I tested your branch here in my test harness and it passes reading from stdin and writing to stdout 👍

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -69,19 +69,22 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) Errno {
...
 		switch eventType {
 		case EventTypeClock: // handle later
-			// +8 past userdata +8 name alignment
+			// +8 past userdata +8 contents_offset
 			errno = processClockEvent(ctx, mod, inBuf[inOffset+8+8:])
 		case EventTypeFdRead, EventTypeFdWrite:
-			// +8 past userdata +4 FD alignment
-			errno = processFDEvent(mod, eventType, inBuf[inOffset+8+4:])
+			// +8 past userdata +8 contents_offset
+			errno = processFDEvent(mod, eventType, inBuf[inOffset+8+8:])
```
codefromthecrypt, line 87: yep. no idea how I got that wrong before, but at least I documented it so it looked very deliberate ;)

codefromthecrypt (approved):
thanks for sorting it out!

PR #1240, opened 2023-03-14
gojs: refactors out gojs.Config type

mathetake:
In order to support more configuration, we should stop using context as
it is getting gnarly.


codefromthecrypt:
@mathetake I repurposed this to sort out config so that it is cleaner to add inheriting the current user (in a different change)

mathetake:
lg

codefromthecrypt (approved):

PR #1241, opened 2023-03-15
Removes usage of host functions with Wasm optocdes

mathetake:
Since #1213, engines no longer can support the Wasm-body-implemented host functions,
which has been only used to stub out the unimplemented functions. 
Also, this reduces the complexity of the future development of engines.


codefromthecrypt (approved):
nice cleanup

mathetake:
🧹

PR #1244, opened 2023-03-15
Forbids empty name module imports

mathetake:
This change disallows importing empty-named modules, and we no longer
support the host modules with the empty name.

This will enable us to optimize the "anonymous" instantiation in the future by
special-casing empty-named module instantiation.


examples/concurrent-instantiation/main.go
```diff
@@ -41,12 +40,8 @@ func main() {
 		go func(i int) {
 			defer wg.Done()
 
-			// Important: each instance needs a unique "name", so we create new wazero.ModuleConfig per instance,
-			// and assigns the iteration counter as the name.
-			config := wazero.NewModuleConfig().WithName(strconv.Itoa(i))
```
mathetake, line 46: I like this

internal/integration_test/bench/bench_test.go
```diff
@@ -123,14 +122,14 @@ func runInitializationConcurrentBench(b *testing.B, r wazero.Runtime) {
 	wg.Add(b.N)
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		go func(name string) {
-			m, err := r.InstantiateModule(testCtx, compiled, config.WithName(name))
+		go func() {
+			m, err := r.InstantiateModule(testCtx, compiled, config)
```
mathetake, line 126: 🚀

internal/wasm/host.go
```diff
@@ -84,7 +85,7 @@ func NewHostModule(
 	if moduleName != "" {
 		m = &Module{NameSection: &NameSection{ModuleName: moduleName}}
 	} else {
-		m = &Module{}
+		return nil, errors.New("empty name is not allowed")
```
codefromthecrypt, line 88: good!

codefromthecrypt (approved):
Looks good. I feel there must be a godoc somewhere where we suggest new runtimes to avoid name conflict.. can you briefly sweep to ensure the new behavior is incorporated into them?

mathetake:
also realized that this makes "Runtime.Close" not close all of modules, and anonymous modules must explicitly called Closed. Will look into the docs tomorrow.

mathetake:
https://github.com/tetratelabs/wazero/pull/1244/commits/bb57f1abaa57971213b426e04bc4952ae9c0fe6a ok added the anonymous module tracking

mathetake:
running bench again :waiting:

mathetake:
hmm sequential perf dropped (though there's this tendency before sharding)

```
benchstat old.txt new.txt
name                                    old time/op    new time/op    delta
Initialization/interpreter-32             36.2µs ± 2%    35.6µs ± 1%   -1.66%  (p=0.000 n=27+26)
Initialization/interpreter-multiple-32    39.2µs ± 9%    18.5µs ± 8%  -52.71%  (p=0.000 n=30+28)
Initialization/compiler-32                28.7µs ± 8%    38.5µs ± 1%  +34.25%  (p=0.000 n=30+28)
Initialization/compiler-multiple-32       24.5µs ± 5%    14.4µs ± 2%  -41.18%  (p=0.000 n=28+29)
```

mathetake:
thinking about adding `WithNoName` option instead, on which we document like don't forget to call `api.Module.Close`

mathetake:
having said that, perf drop on sequential instantiation is not a big deal as such case in anyway people do not want perf, plus the drop is in micro sec scale.

mathetake:
@codefromthecrypt need your input

codefromthecrypt:
my thought summary is to forbid use of empty name modules, like no importing or lookups by name, so that we can optimize this post RC

mathetake:
🤞

internal/wasm/module.go
```diff
@@ -459,6 +459,9 @@ func (m *Module) validateMemory(memory *Memory, globals []GlobalType, _ api.Core
 func (m *Module) validateImports(enabledFeatures api.CoreFeatures) error {
 	for i := range m.ImportSection {
 		imp := &m.ImportSection[i]
+		if imp.Module == "" {
+			return errors.New("importing empty named module is not allowed")
```
codefromthecrypt, line 463:
```suggestion
			return fmt.Errorf("import[%d] has an empty module name", i)
```

internal/wasm/module_test.go
```diff
@@ -572,6 +572,12 @@ func TestModule_validateImports(t *testing.T) {
 		expectedErr     string
 	}{
 		{name: "empty import section"},
+		{
+			name:            "reject empty named module",
+			enabledFeatures: api.CoreFeaturesV1,
+			i:               &Import{Module: "", Name: "n", Type: ExternTypeFunc, DescFunc: 0},
+			expectedErr:     "importing empty named module is not allowed",
```
codefromthecrypt, line 579:
```suggestion
			expectedErr:     "import[0] has an empty module name",
```

internal/integration_test/fuzz/wazerolib/nodiff.go
```diff
@@ -86,6 +86,10 @@ func requireNoDiff(wasmBin []byte, checkMemory bool, requireNoError func(err err
 	defer interpreter.Close(ctx)
 
 	compilerCompiled, err := compiler.CompileModule(ctx, wasmBin)
+	if err != nil && err.Error() == "importing empty named module is not allowed" {
```
codefromthecrypt, line 89: my other suggestion will break this as you'll need to use strings.HasSuffix "empty module name" ...

runtime_test.go
```diff
@@ -455,6 +455,12 @@ func TestRuntime_InstantiateModule_WithName(t *testing.T) {
 
 	require.Nil(t, internal.Module("0"))
 	require.Equal(t, internal.Module("2"), m2)
+
+	// Empty name module shouldn't be returned via Module() for future optimization.
```
codefromthecrypt, line 459: good

codefromthecrypt (approved):
only polish related suggestions

PR #1245, opened 2023-03-15
gojs: adds support for uid and gid

codefromthecrypt:
This adds `gojs.WithOSUser` which passes through current user IDs so
that GOOS=js compiled wasm can read them. This also adds support for
reading back the uid and gid on files. In summary, this passes
`os.TestChown` except on windows where it will not work due to lack of
support.


internal/gojs/builtin.go
```diff
@@ -1,18 +1,24 @@
...
 	"github.com/tetratelabs/wazero/internal/gojs/goos"
 )
 
 // newJsGlobal = js.Global() // js.go init
-func newJsGlobal(rt http.RoundTripper) *jsVal {
+func newJsGlobal(config *config.Config) *jsVal {
 	var fetchProperty interface{} = goos.Undefined
-	if rt != nil {
+	uid, gid, euid := config.Uid, config.Gid, config.Euid
+	// We don't populate groups because our stat can't read them back
```
codefromthecrypt, line 12: even when we don't, TestChown tries to verify the stat has the correct gid on files afterwards. So, it still fails. I'll look into if we can return the group info or not.. possibly backfilling with -1 if that's what windows does.

mathetake (approved):
Verified with this, wazero passes the os tests on ubuntu

PR #1251, opened 2023-03-17
poll_oneoff: do not wait for input if fd is not a tty

evacchi:
WIP: I have something that works on *NIX using `ioctl`. On Windows I am hacking something for the console using `GetNumberOfConsoleInputEvents` (not included here yet) but I need a bit more time

tests will be provided

EDIT: I thought I have found the perfect hack with `isTerminal(fd)` but unfortunately it doesn't work on Windows 😬 

-----

EDIT: 

    This is sort-of a hack/approximation, but it works.

    Rationale: poll_oneoff is expected to return immediately if there
    isn't any input ready.

    When the fd is `stdin`, the input either is there
    or it is not; i.e. if you are expecting interactive input,
    then you probably want to do a blocking read(), otherwise
    poll would immediately return with an answer: there is no data.

    If indeed there is data, then it means you are not doing
    interactive input, but piping it into your stdin.

    Thus, platform.isTerminal(fd) gives you the right answer.
    However, fd=0,1,2 is not a valid value on Windows, so
    we fix the implementation for that platform, and there we go.

    I assume this to be a temporary fix, before we do implement
    poll_oneoff properly, but for now it is good enough.

    As a test, I updated the test for poll so that
    the result type is consistent with the state of os.Stdin


codefromthecrypt:
@achille-roussel can you help tag-team on this with @evacchi? I think getting the approach here correct is a blocking concern of 1.0 (pun absolutely intended)

achille-roussel:
🤔 

Have you considered using `poll(2)` or potentially integrating with the Go I/O facilities (e.g. internal/poll, etc... maybe possible via `syscall.RawConn` for sockets)?

Something that is not well documented for `poll_oneoff` is whether it is supposed to wait for the events to occur; the fact that clock subscriptions are set would indicate that it's expected to wait for at least one event before it returns, in which case just checking the number of bytes available on the input buffer wouldn't be sufficient.

The "input buffer" might mean different things depending on the file descriptor as well; for a file on disk, is it the number of bytes loaded in page cache? or the number of bytes between the seek position and the end of file? etc... Files tend to not work well with polling mechanisms, to my knowledge linux and darwin say that files are always ready for either reading or writing when passed to `poll(2)`, `epoll(2)` or `kevent(2)`.

Unless you plan to support the socket functions of WASI, I would take the same path than linux and if you get a file descriptor for an open file in `poll_oneoff`, just declare it ready for reading/writing and return immediately?

evacchi:
Oh this is not about sockets, TBF it is mostl

evacchi:
(erroneously closed because mobile and fat fingers)

achille-roussel:
OK, stdin is likely to be a pipe, tho it could be anything, assuming it's a pipe will probably cover most use cases.

I double-checked with this program (tho it's probably something you had verified before):
```c
#include <stdio.h>
#include <unistd.h>
#include <sys/ioctl.h>

int main() {
  int fd[2];
  pipe(fd);
  write(fd[1], "hello world\n", 12);

  int nbytes = 0;
  ioctl(fd[0], FIONREAD, &nbytes);

  printf("FIONREAD = %d\n", nbytes);
  return 0;
}
```
On both darwin and linux the program gives this output:
```
FIONREAD = 12
```

Pipes work well with polling capabilities of the OS, which would give you the ability to not busy loop (e.g. if the program is waiting for data to be available on stdin), but you'd have to modify the clock subscriptions to not be managed by Go timers, tho considering the timeline the current approach sounds like a good tradeoff 👍

evacchi:
yep FIONREAD [works for Linux and Darwin,](https://github.com/tetratelabs/wazero/pull/1251/files#diff-51224c78577d47e094125195dd23bfa00833eb12572a1a37e64f34ded21edb79R140) but I can't find nothing as slick for Windows :D 

something that sort-of-worked was 

	t, err := syscall.GetFileType(handle)
	if err != nil {
 		return false, err
 	}
 	if t == syscall.FILE_TYPE_CHAR { // it is a ~file
	if t == syscall.FILE_TYPE_PIPE { // it is.. well you get it

but this should be sort-of-equivalent to `platform.isTerminal` but then I stopped trying. I'll resume some experiment today.

The culprit may have been simply that fd=0,1,2 is not really a valid handle on Windows and I have to match them to syscall.Stdin,Stdout,Stderr instead

achille-roussel:
Do you care about windows for the particular use case of container2wasm?

Docker and friends will all be running on linux, so maybe you can delay portability and on windows you assume that file descriptors are always ready?

evacchi:
well I only care about the semantics being the same across platforms. If you run that example on Windows it will hang waiting for stdin whereas it should return (which it does on other Wasm runtimes) :P

evacchi:
maybe it is also worth adding the test case here https://github.com/ktock/container2wasm/pull/23#discussion_r1139886299 too? but I will do this at a later time, I would prefer to simplify it and clean it (eg. I think that example is not trimming the buffer correctly)

codefromthecrypt:
Thanks for the work here! I'm making a comment not a review as I'm not sure if this is meant to be non-draft or not ;)

I think a test for this specific issue should be in wasi's testdata and invoked in wasi_stdlib_test.go. I don't think we should merge without any integration test because this is unlike other areas.. it isn't tested in wasi-testsuite nor any of our language stdlib tests. Even if only starting with C that's fine, even if it skips on windows, but our CI should know what it works on.

I think even if not on this change, what @achille-roussel suggests is probably the right course, which is having the primary interface made in a way that go isn't always managing the polling. For example, it can be go-managed when the input filesystem is not raw, or in a fallback situation like windows. We can manage doing this afterwards because this isn't the only area the abstraction leaks to where wasi is handling too much logic.

codefromthecrypt (approved):
Thanks so much @evacchi and @achille-roussel. As I'm traveling monday, I don't want to block anything, so made a pre-emptive "review" in case this is intentionally in review state. IMHO, the only thing left are tests:

* in our `platform` package try to add a test to ensure isTerminal passes now and wouldn't have before. If it isn't possible, comment why in isTerminal at the line of terminal_windows.go
* in our `wasi_snapshot_preview1/testdata/zig-cc/wasi.c` add `main_poll` with logic that led to this issue and update `wasi_stdlib_test.go` to invoke it where supported and have a comment where not. This will make sure when we refactor, we don't accidentaly break this again. If it is easy to do the other languages, (zig, cargo-wasi) go for it, but that's not necessary imho at least one is fine.

codefromthecrypt:
Rambling:

Before we expose sysfs, and plenty of time for that, one test that an API is designed correctly, is that it can be implemented natively with syscalls and adapted without. Both of these are the real test. For example, if we can't adapt it, that's a problem worth exploring, as much as blocking native would be. We have more work than other runtimes, as it seems many don't support virtual filesystems. However, we do need to recognize it is more common to want native to work right (via real syscalls or the jumping around thing for windows ;))

The real test of this is implementing with two ABI: that's how we can see what's accidentally implemented at the wrong abstraction. This is one reason it has been good to work on gojs: it has shown some areas where we accidentally implemented too much in wasi. I think post 1.0 some experiments around how wasi-filesystem handles or doesn't handle polling will be useful before we cement our api as public

evacchi:
I have added some tests, the main issue however is that I cannot really test when `fd` _is_ a tty; or, at least, I couldn't come up with a way that was not running it manually:

```
wazero on  stdin-ensure-bytes [$] via 🐹 v1.20.1
❯ wazero run imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm poll
NOINPUT

wazero on  stdin-ensure-bytes [$] via 🐹 v1.20.1
❯ echo test | wazero run imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm poll
STDIN
```

in the test environment, this always returns STDIN because (I guess) STDIN is always being connected to a pipe; this in turn makes it pretty hard to mock because even passing a pipe makes it -- duh -- connected to a pipe.

This is not really testing the bug we had, because the bug we had was that STDIN was always returned. The only thing I can think of is to add a script somewhere...

internal/platform/terminal_windows.go
```diff
@@ -8,7 +8,20 @@ import (
...
 
 func isTerminal(fd uintptr) bool {
+	// If a bare fd is given, we translate std{in,out,err} to a valid Windows handle.
+	var handle uintptr
+	switch fd {
+	case 0:
+		handle = uintptr(syscall.Stdin)
+	case 1:
+		handle = uintptr(syscall.Stdout)
+	case 3:
```
achille-roussel, line 18: Is this supposed to be 2?

evacchi: 🤦‍♂️...yes

codefromthecrypt (approved):
thanks for putting in the tests and comments!

PR #1252, opened 2023-03-17
path_filestat_get: interpret LOOKUP_SYMLINK_FOLLOW

achille-roussel:
This PR adds support for passing lookup flags to `path_filestat_get`. The motivation came from the work on https://github.com/golang/go/issues/58141, especially implementing `syscall.Lstat`.

Without this change, an alternative implementation is possible using `path_open` + `fd_filestat_get` + `fd_close` because wazero allows passing stats, however, this fails on OSX because `LOOKUP_SYMLINK_FOLLOW` is translated to `O_NOFOLLOW` which causes `open` to fail with `ELOOP` because wazero doesn't set `O_SYMLINK` (I can follow up with a fix for that in a different PR).

I updates the tests to use the `LOOKUP_SYMLINK_FOLLOW` flag since that was the historical behavior being tested; however, I haven't found a paved path for testing with symbolic links. The tests use `fstest.FS` which ends up creating an internal adapter between `fs.FS` and `sysfs.FS` and always returns `ENOSYS` on calls to `Lstat`.

It seems that we would want tests that actually work on a local file system (e.g. something in `testdata`) so we can exercise an actual call to `lstat`, I wanted to get feedback before making too many changes to the test suite. Let me know what you would recommend!

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -1419,11 +1419,7 @@ func pathFilestatGetFn(_ context.Context, mod api.Module, params []uint64) Errno
 	fsc := mod.(*wasm.CallContext).Sys.FS()
 
 	fd := uint32(params[0])
-
-	// TODO: flags is a lookupflags and it only has one bit: symlink_follow
-	// https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#lookupflags
-	_ /* flags */ = uint32(params[1])
-
+	flags := uint16(params[1])
 	path := uint32(params[2])
 	pathLen := uint32(params[3])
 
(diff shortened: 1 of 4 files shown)


imports/wasi_snapshot_preview1/fs.go
```diff
@@ -1432,9 +1428,23 @@ func pathFilestatGetFn(_ context.Context, mod api.Module, params []uint64) Errno
...
 
-	// Stat the file without allocating a file descriptor
+	// Stat the file without allocating a file descriptor.
+	//
+	// Note: `preopen` is a `sysfs.FS` interface, so passing the address of `st`
+	// causes the value to escape to the heap because the compiler doesn't know
+	// whether the pointer will be retained by the method.
+	//
+	// This could be optimized by modifying Stat/Lstat to return the `Stat_t`
+	// value instead of passing a pointer as output parameter.
```
achille-roussel, line 1438: This is unrelated to the change but I wanted to record this observation somewhere, let me know if you'd rather have it somewhere else.

achille-roussel:
The failing tests seem to be related to the use of `fs.FS` as well in `wasi_stdlib_test.go`, calls to `path_filestat_get` with `LOOKUP_SYMLINK_FOLLOW` now fail with `ENOSYS` when backed by a Go file system.

achille-roussel:
I maybe found a compromise by implementing `Lstat` in terms of `Stat` in the `fs.FS` adapter; I left comments to explain the reasoning and added tests to exercise `path_filestat_get` with and without `LOOKUP_SYMLINK_FOLLOW`.

Let me know if you would like to see anything changed.

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: interesting point, let's leave it for now. I guess we can't solve this with `//go:noescape`? In any case, it was only conventional to pass like syscall.Stat. Feel free to change in this PR or next.

internal/sysfs/adapter.go
```diff
@@ -61,6 +61,19 @@ func (a *adapter) Stat(path string, stat *platform.Stat_t) error {
 	return platform.StatFile(f, stat)
 }
 
+// Lstat implements FS.Lstat
+func (a *adapter) Lstat(path string, stat *platform.Stat_t) error {
+	// At this time, we make the assumption that fs.FS instances do not support
```
codefromthecrypt, line 66: I love this

codefromthecrypt (approved):
Great work!

achille-roussel:
> interesting point, let's leave it for now. I guess we can't solve this with //go:noescape? In any case, it was only conventional to pass like syscall.Stat. Feel free to change in this PR or next.

Unfortunately `//go:noescape` only applies to functions that do not have a body (e.g. implemented in assembly) so it won't help in this context.

Thanks for the quick review!

PR #1253, opened 2023-03-17
fix file type of directory entries

achille-roussel:
Prior to this change, when reading directory entries, wazero would assume that all files are regular files except directories. This causes issues in client applications that work with symbolic links for example.

The issue was detected while working on passing the `testing/fstest` tests as part of the implementation of https://github.com/golang/go/issues/58141 because `fstest.TestFS` validates that the file type seen by `fs.DirEntry.Type()` and `fs.FileInfo.Mode().Type()` are the same.

Regarding testing this change, to reuse the existing test suite would likely require updating `internal/fstest.FS` to contain a symbolic link, but this fixture object is used in so many places that it may be difficult to land the fix. An alternative could be to make a local copy of `fstest.FS` and add the symbolic link there, but it defeats the documented intent of `fstest` to reduce drift between systems. Let me know how you would prefer to proceed!

### DIFF
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -1857,7 +1857,24 @@ var (
 		'a', 'b', '-', // name
 	}
 
-	dirents = append(append(append(append(direntDot, direntDotDot...), dirent1...), dirent2...), dirent3...)
+	// TODO: this entry is intended to test reading of a symbolic link entry,
+	// tho it requires modifying fstest.FS to contain this file.
+	dirent4 = []byte{
+		6, 0, 0, 0, 0, 0, 0, 0, // d_next = 6
+		0, 0, 0, 0, 0, 0, 0, 0, // d_ino = 0
+		2, 0, 0, 0, // d_namlen = 2 characters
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
@achille-roussel let's defer the test thing for reasons you noted. the priority has been external tests passing and there's still a lot of tech debt to deal with for the interior. Once #1251 is done, I think we can start on the file type work and after that, consolidating all the various tests as you proposed ages ago. After it is all shiny, expose for a 1.x release

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -1857,7 +1857,24 @@ var (
 		'a', 'b', '-', // name
 	}
 
-	dirents = append(append(append(append(direntDot, direntDotDot...), dirent1...), dirent2...), dirent3...)
+	// TODO: this entry is intended to test reading of a symbolic link entry,
```
codefromthecrypt, line 1860: yeah or using a real file for this special case, until fstest.FS has symlink support. If you want to follow-up with a real file that's probably easiest/best alternative prior to some wider change

codefromthecrypt (approved):

PR #1255, opened 2023-03-19
sys/fs: removes redundant cachedStat allocation

mathetake:


internal/sys/fs.go
```diff
@@ -194,7 +194,6 @@ func (f *FileEntry) CachedStat() (ino uint64, fileType fs.FileMode, err error) {
 		if err = f.Stat(&st); err != nil {
```
mathetake, line 194: this call to `Stat` is enough to set `f.cachedStat`

codefromthecrypt (approved):
thx for the fix

Issue #1258, opened 2023-03-20
optimize module instantiation when there's no name


codefromthecrypt:
Currently, our locks around module instantiation are name based. This protects the runtime from conflicts where one module can mask the exports of another. In #1244, we forbade importing extern types, such as functions, from modules without a name. We called that concept "anonymous modules".

What's left to do is safely optimize instantiation performance for anonymous modules. For example, a module without a name can use a cheaper lock than a string based one. It can use a numeric key such as from its pointer, or a lookup key, such as how we handle open file descriptors. This key would be used to ensure a runtime closes it, and also that if the module is closed explicitly, it is deregistered from the runtime. The last part is important to ensure that the "open modules" doesn't grow unbounded in scenarios where many modules are created.

### don't decouple anonymous modules from their runtime.

It is important that implementing this is done safely. For example, the defer `rt.Close(ctx)` convention works exactly as today, even if optimized underneath. Most compilers do not set a module name, and people are trained to trust runtime.Close. The runtime should remain the owner of its modules, regardless of their name.

A future feature of "forget" or "unsafe" module could be made to deregister a module so that closing a runtime never closes it. This requires careful use to avoid safety problems. In any case, this should be done *after* optimization of anonymous modules, so that the precise performance gains can be put in context of its cons.

### Notes

Here are some notes from @mathetake on the last attempt, in #1244

> one thing is that even anonymous modules should be captured by wasm.Store, so something like this https://github.com/tetratelabs/wazero/pull/1244/commits/bb57f1abaa57971213b426e04bc4952ae9c0fe6a should be required to create a set of anonymous modules

ref: https://gophers.slack.com/archives/C04CG4A2NKX/p1678974561121089?thread_ts=1678864254.795459&cid=C04CG4A2NKX

cc @achille-roussel  @Pryz @ckaznocha

achille-roussel:
I've been doing some more thinking on the topic.

First, I think we have a good base to _enable_ multiple instantiation of anonymous modules in https://github.com/tetratelabs/wazero/pull/1259, we can potentially move forward with the PR tho it has limitations in terms of usability.

The main limitation is the way where wazero automatically inherits the module name from the wasm binary if none were specified by the application, see https://github.com/tetratelabs/wazero/blob/main/runtime.go#L285-L288 If the compiler injected a name into a wasm module, there is no mechanism by which we can remove that name and instantiate it multiple times.

[...]

PR #1259, opened 2023-03-20
Support registering multiple instances of anonymous modules

achille-roussel:
This change is a follow up from our conversation on Slack regarding potentially not registering anonymous modules to the runtime instance that originally created them.

One key consideration to take into account is that, with the exception of host modules from the imports such as wasi_snapshot_preview1, modules would by default be anonymous since the default module config has no name.

To minimize the impact of this change, we can maintain the historical behavior of the helper `Instantiate` methods, by inheriting the name from the module binary, which means the only way to create a anonymous modules is using `InstantiateModule`.

I haven't yet spent time working on documentation as I think the priority would be to agree that this is the change we want to bring to wazero, there is probably a need for a test or two as well. This PR is a starting point to bring material to the conversation.

PS: as a side note, I was pleased to see how little had to change to make it work, wazero really is a high quality code base 🙌

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -344,6 +344,6 @@ func (b *hostModuleBuilder) Instantiate(ctx context.Context) (api.Module, error)
 		return nil, err
 	} else {
 		compiled.(*compiledModule).closeWithModule = true
-		return b.r.InstantiateModule(ctx, compiled, NewModuleConfig())
+		return b.r.InstantiateModule(ctx, compiled, NewModuleConfig().WithName(b.moduleName))
 	}
 }
--- a/builder.go
+++ b/builder.go
@@ -344,6 +344,6 @@ func (b *hostModuleBuilder) Instantiate(ctx context.Context) (api.Module, error)
 		return nil, err
 	} else {
 		compiled.(*compiledModule).closeWithModule = true
-		return b.r.InstantiateModule(ctx, compiled, NewModuleConfig().WithName(b.moduleName))
+		return b.r.InstantiateModule(ctx, compiled, NewModuleConfig())
 	}
 }
--- a/imports/assemblyscript/assemblyscript_test.go
+++ b/imports/assemblyscript/assemblyscript_test.go
@@ -437,7 +437,7 @@ func requireProxyModule(t *testing.T, fns FunctionExporter, config wazero.Module
 	envCompiled, err := builder.Compile(ctx)
 	require.NoError(t, err)
 
-	_, err = r.InstantiateModule(ctx, envCompiled, config)
+	_, err = r.InstantiateModule(ctx, envCompiled, config.WithName("env"))
 	require.NoError(t, err)
 
 	proxyBin := proxy.NewModuleBinary("env", envCompiled)
--- a/imports/assemblyscript/assemblyscript_test.go
+++ b/imports/assemblyscript/assemblyscript_test.go
@@ -437,7 +437,7 @@ func requireProxyModule(t *testing.T, fns FunctionExporter, config wazero.Module
 	envCompiled, err := builder.Compile(ctx)
 	require.NoError(t, err)
 
-	_, err = r.InstantiateModule(ctx, envCompiled, config.WithName("env"))
+	_, err = r.InstantiateModule(ctx, envCompiled, config)
 	require.NoError(t, err)
 
 	proxyBin := proxy.NewModuleBinary("env", envCompiled)
--- a/imports/wasi_snapshot_preview1/wasi_test.go
+++ b/imports/wasi_snapshot_preview1/wasi_test.go
@@ -95,7 +95,8 @@ func requireProxyModule(t *testing.T, config wazero.ModuleConfig) (api.Module, a
 	wasiModuleCompiled, err := wasi_snapshot_preview1.NewBuilder(r).Compile(ctx)
 	require.NoError(t, err)
 
-	_, err = r.InstantiateModule(ctx, wasiModuleCompiled, config)
+	_, err = r.InstantiateModule(ctx, wasiModuleCompiled, config.
+		WithName(wasi_snapshot_preview1.ModuleName))
 	require.NoError(t, err)
 
 	proxyBin := proxy.NewModuleBinary(wasi_snapshot_preview1.ModuleName, wasiModuleCompiled)
@@ -126,7 +127,8 @@ func requireErrnoNosys(t *testing.T, funcName string, params ...uint64) string {
 	wasiModuleCompiled, err := wasi_snapshot_preview1.NewBuilder(r).Compile(ctx)
 	require.NoError(t, err)
 
-	_, err = r.InstantiateModule(ctx, wasiModuleCompiled, wazero.NewModuleConfig())
+	_, err = r.InstantiateModule(ctx, wasiModuleCompiled, wazero.NewModuleConfig().
+		WithName(wasi_snapshot_preview1.ModuleName))
 	require.NoError(t, err)
 
 	proxyBin := proxy.NewModuleBinary(wasi_snapshot_preview1.ModuleName, wasiModuleCompiled)
--- a/imports/wasi_snapshot_preview1/wasi_test.go
+++ b/imports/wasi_snapshot_preview1/wasi_test.go
@@ -95,8 +95,7 @@ func requireProxyModule(t *testing.T, config wazero.ModuleConfig) (api.Module, a
 	wasiModuleCompiled, err := wasi_snapshot_preview1.NewBuilder(r).Compile(ctx)
 	require.NoError(t, err)
 
-	_, err = r.InstantiateModule(ctx, wasiModuleCompiled, config.
-		WithName(wasi_snapshot_preview1.ModuleName))
+	_, err = r.InstantiateModule(ctx, wasiModuleCompiled, config)
 	require.NoError(t, err)
 
 	proxyBin := proxy.NewModuleBinary(wasi_snapshot_preview1.ModuleName, wasiModuleCompiled)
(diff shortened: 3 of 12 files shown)


codefromthecrypt:
"the default module config has no name" is only in effect if the name section doesn't define one. please see #1258 for why I really don't want special naming conventions resulting in safety control differences. It is similar to the problem in java of using exceptions for control flow. Let's find another way so we can have both.

codefromthecrypt:
Remember right now, we are 4 days from 1.0 and all of our examples, say this:

```go
	// Create a new WebAssembly Runtime.
	r := wazero.NewRuntime(ctx)
	defer r.Close(ctx) // This closes everything this Runtime created.
```

We have code all over GitHub that would have to change as well code no one can see. And if that doesn't change, people will have resource leaks. I can't get behind this as something that happens without an explicit config opt in. If you and @mathetake do this anyway, it is ok. please help bear the burden of contacting all known users.. similar to what I did for far safer assumption breaks.

achille-roussel:
> If you and @mathetake do this anyway, it is ok. please help bear the burden of contacting all known users.. similar to what I did for far safer assumption breaks.

I agree, I don't think this is the right model, we need to find something that plays better by the rules that wazero has established already 👍 it must be additive, not modify the way existing code behaves.

I needed to see it to get convinced 😄

mathetake:
can we see the perf diff with `go test -run='^$' -benchmem -bench 'BenchmarkCompilation|BenchmarkInitialization' -count=5 ./internal/integration_test/bench/...`?

achille-roussel:
Currently the benchmarks show no differences, but I think that's expected since we're _allowing_ the instantiation of anonymous modules, so there were no benchmarks to exercise it before?

codefromthecrypt:
maybe runInitializationConcurrentBench needs to change to not specify a module name?

ckaznocha:
> maybe runInitializationConcurrentBench needs to change to not specify a module name?

I would still expect very little difference in benchmarks. The only savings in this implementation is the cost of inserting an element into a map.

mathetake:
```diff
diff --git a/internal/integration_test/bench/bench_test.go b/internal/integration_test/bench/bench_test.go
index 74c9c017..73f76ffb 100644
--- a/internal/integration_test/bench/bench_test.go
+++ b/internal/integration_test/bench/bench_test.go
@@ -6,7 +6,6 @@ import (
        _ "embed"
        "fmt"
        "runtime"
-       "strconv"
        "sync"
        "testing"
 
@@ -123,14 +122,14 @@ func runInitializationConcurrentBench(b *testing.B, r wazero.Runtime) {
        wg.Add(b.N)
        b.ResetTimer()
        for i := 0; i < b.N; i++ {
-               go func(name string) {
-                       m, err := r.InstantiateModule(testCtx, compiled, config.WithName(name))
+               go func() {
+                       m, err := r.InstantiateModule(testCtx, compiled, config)
                        if err != nil {
                                b.Error(err)
                        }
                        m.Close(testCtx)
                        wg.Done()
-               }(strconv.Itoa(i))
+               }()
        }
        wg.Wait()
        b.StopTimer()

```

try this in the bench... without this, bench result won't be different...

mathetake:
little bit unfair, but in any situation, we have to pay the similar price in reality right?

ckaznocha:
for sure!

mathetake:
plus without the change, the bench doesn't measure the anonymous instantiation at all.

mathetake:
https://github.com/tetratelabs/wazero/commit/bb57f1abaa57971213b426e04bc4952ae9c0fe6a this kind of sharding might improve the perf, but I guess the change here is a good starting point.

achille-roussel:
Here are the numbers with the modified benchmarks:

```
goos: darwin
goarch: arm64
                                    │ /tmp/wazero.bench.main │        /tmp/wazero.bench.2         │
                                    │         sec/op         │   sec/op     vs base               │
Initialization/interpreter                      12.27µ ± 12%   12.67µ ± 7%        ~ (p=0.394 n=6)
Initialization/interpreter-multiple             36.55µ ±  7%   18.35µ ± 3%  -49.79% (p=0.002 n=6)
Initialization/compiler                         8.881µ ±  4%   9.167µ ± 3%   +3.22% (p=0.015 n=6)
Initialization/compiler-multiple               16.117µ ± 27%   9.756µ ± 2%  -39.47% (p=0.002 n=6)
Compilation/with_extern_cache                  108.52µ ±  6%   94.55µ ± 1%  -12.88% (p=0.002 n=6)
Compilation/without_extern_cache                2.254m ±  5%   2.136m ± 3%   -5.24% (p=0.026 n=6)
geomean                                         50.04µ         40.17µ       -19.74%

                                    │ /tmp/wazero.bench.main │         /tmp/wazero.bench.2          │
                                    │          B/op          │     B/op      vs base                │
Initialization/interpreter                      132.4Ki ± 0%   132.4Ki ± 0%       ~ (p=1.000 n=6) ¹
Initialization/interpreter-multiple             133.3Ki ± 0%   132.9Ki ± 0%  -0.29% (p=0.002 n=6)
Initialization/compiler                         133.1Ki ± 0%   133.1Ki ± 0%       ~ (p=1.000 n=6) ¹
Initialization/compiler-multiple                137.8Ki ± 0%   137.4Ki ± 0%  -0.28% (p=0.002 n=6)
Compilation/with_extern_cache                   53.25Ki ± 0%   53.23Ki ± 0%  -0.05% (p=0.002 n=6)
Compilation/without_extern_cache                1.070Mi ± 0%   1.070Mi ± 0%  +0.01% (p=0.011 n=6)
geomean                                         163.2Ki        163.0Ki       -0.10%
¹ all samples are equal

                                    │ /tmp/wazero.bench.main │         /tmp/wazero.bench.2          │
                                    │       allocs/op        │  allocs/op   vs base                 │
Initialization/interpreter                        29.00 ± 0%    29.00 ± 0%        ~ (p=1.000 n=6) ¹
Initialization/interpreter-multiple               46.50 ± 1%    42.00 ± 0%   -9.68% (p=0.002 n=6)
Initialization/compiler                           29.00 ± 0%    29.00 ± 0%        ~ (p=1.000 n=6) ¹
Initialization/compiler-multiple                  36.00 ± 3%    32.00 ± 0%  -11.11% (p=0.002 n=6)
Compilation/with_extern_cache                     979.0 ± 0%    979.0 ± 0%        ~ (p=1.000 n=6) ¹
Compilation/without_extern_cache                 11.33k ± 0%   11.33k ± 0%        ~ (p=0.076 n=6)
geomean                                           158.1         152.4        -3.59%
¹ all samples are equal
```

achille-roussel:
With `RunParallel`, we can compare the performance with different GOMAXPROCS versions, here's the different between GOMAXPROCS=1 and GOMAXPROCS=4

```
goos: darwin
goarch: arm64
                                    │ /tmp/wazero.bench.branch │     /tmp/wazero.bench.parallel     │
                                    │          sec/op          │   sec/op     vs base               │
Initialization/interpreter                        12.30µ ± 13%   11.57µ ± 2%        ~ (p=0.394 n=6)
Initialization/interpreter-multiple               35.65µ ±  1%   16.68µ ± 2%  -53.23% (p=0.002 n=6)
Initialization/compiler                           8.874µ ±  4%   9.274µ ± 4%   +4.51% (p=0.002 n=6)
Initialization/compiler-multiple                  16.97µ ± 10%   10.43µ ± 2%  -38.56% (p=0.002 n=6)
Compilation/with_extern_cache                    106.07µ ±  1%   90.39µ ± 2%  -14.78% (p=0.002 n=6)
Compilation/without_extern_cache                  2.141m ±  1%   1.641m ± 2%  -23.39% (p=0.002 n=6)
geomean                                           49.66µ         37.47µ       -24.54%

                                    │ /tmp/wazero.bench.branch │      /tmp/wazero.bench.parallel      │
                                    │           B/op           │     B/op      vs base                │
Initialization/interpreter                        132.4Ki ± 0%   132.4Ki ± 0%  +0.00% (p=0.002 n=6)
Initialization/interpreter-multiple               132.9Ki ± 0%   132.9Ki ± 0%  -0.05% (p=0.002 n=6)
Initialization/compiler                           133.1Ki ± 0%   133.1Ki ± 0%       ~ (p=1.000 n=6) ¹
Initialization/compiler-multiple                  137.4Ki ± 0%   137.4Ki ± 0%  -0.05% (p=0.002 n=6)
Compilation/with_extern_cache                     53.25Ki ± 0%   53.31Ki ± 0%  +0.12% (p=0.002 n=6)
Compilation/without_extern_cache                  1.070Mi ± 0%   1.073Mi ± 0%  +0.22% (p=0.002 n=6)
geomean                                           163.1Ki        163.1Ki       +0.04%
¹ all samples are equal

                                    │ /tmp/wazero.bench.branch │     /tmp/wazero.bench.parallel      │
                                    │        allocs/op         │  allocs/op   vs base                │
Initialization/interpreter                          29.00 ± 0%    29.00 ± 0%       ~ (p=1.000 n=6) ¹
Initialization/interpreter-multiple                 43.00 ± 0%    42.00 ± 0%  -2.33% (p=0.002 n=6)
Initialization/compiler                             29.00 ± 0%    29.00 ± 0%       ~ (p=1.000 n=6) ¹
Initialization/compiler-multiple                    33.00 ± 0%    32.00 ± 0%  -3.03% (p=0.002 n=6)
Compilation/with_extern_cache                       979.0 ± 0%    979.0 ± 0%       ~ (p=1.000 n=6) ¹
Compilation/without_extern_cache                   11.33k ± 0%   11.33k ± 0%  +0.01% (p=0.022 n=6)
geomean                                             153.8         152.4       -0.90%
¹ all samples are equal
```

codefromthecrypt (approved):
Looks good!

I think the results say that the sequential overhead is <1us for compiler backend, yet significantly unlocks parallel performance. This will notably help dapr who recently have event-driven endpoints, as well proxies or gateways who do parallel wasm instantiations (either because they need to because they need to invoke _start per request, or they just don't trust their clients to re-use memory, so re-instantiate each time)

P.S. I would suggest to redo the name of this PR and rewrite the desc to what it is currently doing ;)

achille-roussel:
Yep, the title was outdated, I updated it.

codefromthecrypt:
Thanks @achille-roussel and @ckaznocha!

PR #1261, opened 2023-03-20
sysfs: return st instead of accepting it

codefromthecrypt:
Hopefully, this can help reduce references that escape to heap, but needs to be checked.

@achille-roussel can you quickly check if this is on the right track?


codefromthecrypt:
I found that returning a value was a lot better overall, so if this is cool, let's go with it!

```bash
$ go test -run='^$' -bench '^Benchmark_pathFilestat/sysfs.DirFS.*$' ./imports/wasi_snapshot_preview1 -count=6 > old.txt
$ benchstat old.txt return-pointer.txt return-val.txt 
goos: darwin
goarch: arm64
pkg: github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1
                                          │   old.txt   │        return-pointer.txt         │          return-val.txt           │
                                          │   sec/op    │   sec/op     vs base              │   sec/op     vs base              │
_pathFilestat/sysfs.DirFS_fd=root-12        4.235µ ± 2%   4.146µ ± 1%  -2.10% (p=0.009 n=6)   4.139µ ± 1%  -2.27% (p=0.009 n=6)
_pathFilestat/sysfs.DirFS_fd=directory-12   4.499µ ± 1%   4.460µ ± 1%  -0.86% (p=0.009 n=6)   4.434µ ± 1%  -1.44% (p=0.004 n=6)
geomean                                     4.365µ        4.300µ       -1.48%                 4.284µ       -1.86%

                                          │  old.txt   │         return-pointer.txt         │          return-val.txt           │
                                          │    B/op    │    B/op     vs base                │    B/op     vs base               │
_pathFilestat/sysfs.DirFS_fd=root-12        384.0 ± 0%   384.0 ± 0%       ~ (p=1.000 n=6) ¹   304.0 ± 0%  -20.83% (p=0.002 n=6)
_pathFilestat/sysfs.DirFS_fd=directory-12   416.0 ± 0%   416.0 ± 0%       ~ (p=1.000 n=6) ¹   336.0 ± 0%  -19.23% (p=0.002 n=6)
geomean                                     399.7        399.7       +0.00%                   319.6       -20.04%
¹ all samples are equal

                                          │  old.txt   │         return-pointer.txt         │          return-val.txt           │
                                          │ allocs/op  │ allocs/op   vs base                │ allocs/op   vs base               │
_pathFilestat/sysfs.DirFS_fd=root-12        7.000 ± 0%   7.000 ± 0%       ~ (p=1.000 n=6) ¹   6.000 ± 0%  -14.29% (p=0.002 n=6)
_pathFilestat/sysfs.DirFS_fd=directory-12   8.000 ± 0%   8.000 ± 0%       ~ (p=1.000 n=6) ¹   7.000 ± 0%  -12.50% (p=0.002 n=6)
geomean                                     7.483        7.483       +0.00%                   6.481       -13.40%
¹ all samples are equal
```

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -382,7 +382,7 @@ func getWasiFiletype(fm fs.FileMode) uint8 {
 	}
 }
 
-func writeFilestat(buf []byte, st *platform.Stat_t) (err error) {
+func writeFilestat(buf []byte, st platform.Stat_t) (err error) {
```
achille-roussel, line 385: I think in this context it would be OK to keep it a pointer because the compiler sees through the function and knows it isn't being retained.

It's also good to keep usage consistent and pass it by value everywhere 👍

achille-roussel (approved):
Looking nice, thanks for following up on this!

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: cool I'll change this one. thanks!

codefromthecrypt:
smidge faster changing to pointer receiver for writeFilestat. nice eagle eyes @achille-roussel

mathetake (approved):

codefromthecrypt:
Thanks for idea @achille-roussel!

PR #1262, opened 2023-03-21
fix typo in wazero.HostFunctionBuilder.WithFunc documentation

achille-roussel:
It seems `int32` was repeated twice, based on the pattern I assumed the second occurrence was meant to be `int64`.


codefromthecrypt (approved):
thanks!

PR #1264, opened 2023-03-21
sysfs: requires all methods to return syscall.Errno

codefromthecrypt:
This forces all syscall functions, notably filesystem, to return numeric codes as opposed to mapping in two different areas. The result of this change is better consolidation in call sites of `sysfs.FS`, while further refactoring is needed to address consolidation of file errors.

### DIFF
--- a/imports/wasi_snapshot_preview1/wasi.go
+++ b/imports/wasi_snapshot_preview1/wasi.go
@@ -19,6 +19,7 @@ package wasi_snapshot_preview1
 import (
 	"context"
 	"encoding/binary"
+	"syscall"
 
 	"github.com/tetratelabs/wazero"
 	"github.com/tetratelabs/wazero/api"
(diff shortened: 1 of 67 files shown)


imports/wasi_snapshot_preview1/wasi.go
```diff
@@ -274,12 +275,12 @@ func newHostFunc(
...
 // wasiFunc special cases that all WASI functions return a single Errno
 // result. The returned value will be written back to the stack at index zero.
-type wasiFunc func(ctx context.Context, mod api.Module, params []uint64) Errno
+type wasiFunc func(ctx context.Context, mod api.Module, params []uint64) syscall.Errno
 
 // Call implements the same method as documented on api.GoModuleFunction.
 func (f wasiFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 	// Write the result back onto the stack
-	stack[0] = uint64(f(ctx, mod, stack))
+	stack[0] = uint64(ToErrno(f(ctx, mod, stack)))
```
mathetake, line 283: I would special case 0 here to avoid ToErrono as this is the all the WASI function's path and I guess most of the time  return zero

mathetake (approved):

imports/wasi_snapshot_preview1/wasi.go
codefromthecrypt: great idea

PR #1265, opened 2023-03-21
wasi: removes constraint around closing a pre-open

codefromthecrypt:
wasi-testsuite changed its mind.

See https://github.com/WebAssembly/wasi-testsuite/pull/66


codefromthecrypt:
`interesting_paths` is new I guess

codefromthecrypt:
merging because the build is broken until we do it.

PR #1271, opened 2023-03-22
Removes unnecessary uint64 allocation on callContext

mathetake:


internal/wasm/call_context.go
```diff
@@ -15,8 +15,7 @@ import (
 var _ api.Module = &CallContext{}
 
 func NewCallContext(s *Store, instance *ModuleInstance, sys *internalsys.Context) *CallContext {
-	zero := uint64(0)
```
mathetake, line 18: this caused unnecessary allocation as this escaped into moduleContext.Closed.

codefromthecrypt (approved):

PR #1195, opened 2023-03-03
build: fix issue with GNU make vs BSD make

evacchi:
It looks like $(subst from,to,text) works differently on BSD make
vs GNU make. The behavior we expected for

    $(subst,$(space),$(comma),$(var))

is instead the behavior of `$(patsubst ...)` on GNU Make.

The alternative syntax `$(var: =,)` seems portable across the two
and shows the same behavior.

(tested on Fedora 37, closes #1171)

Signed-off-by: Edoardo Vacchi <evacchi@users.noreply.github.com>

PR #1272, opened 2023-03-22
build: correctly replace space with comma regardless GNU/BSD

mathetake:
The fix #1195 was incorrect and resulted in insufficient coverage results (you can check by comparing [before 1995](https://github.com/tetratelabs/wazero/actions/runs/4319671295/jobs/7539103651) and [after](https://github.com/tetratelabs/wazero/actions/runs/4488542958/jobs/7893279544)). Notably, the current main only produces the coverage result of the top `wazero` package.


codefromthecrypt (approved):
thanks!

PR #1281, opened 2023-03-24
Import function type check at validation phrase

mathetake:


internal/wasm/store.go
```diff
@@ -407,19 +407,12 @@ func (m *ModuleInstance) resolveImports(module *Module, importedModules map[stri
 
 		switch i.Type {
 		case ExternTypeFunc:
-			typeIndex := i.DescFunc
-			// TODO: this shouldn't be possible as invalid should fail validate
-			if int(typeIndex) >= len(module.TypeSection) {
```
mathetake, line 412: this now is done during validation phrase

internal/wasm/store.go
```diff
@@ -430,6 +423,7 @@ func (m *ModuleInstance) resolveImports(module *Module, importedModules map[stri
 			if expected.Type != importedTable.Type {
 				err = errorInvalidImport(i, idx, fmt.Errorf("table type mismatch: %s != %s",
 					RefTypeName(expected.Type), RefTypeName(importedTable.Type)))
+				return
```
mathetake, line 426: while there, I noticed this missing `return` and we haven't added the unit test to cover this, so I backfilled it.

internal/wasm/table_test.go
```diff
@@ -61,6 +61,20 @@ func Test_resolveImports_table(t *testing.T) {
 		err := m.resolveImports(&Module{ImportSection: []Import{{Module: moduleName, Name: name, Type: ExternTypeTable, DescTable: importTableType}}}, importedModules)
 		require.EqualError(t, err, "import[0] table[test.target]: maximum size mismatch: 10, but actual has no max")
 	})
+	t.Run("type mismatch", func(t *testing.T) {
```
mathetake, line 64: this is the backfilled test.

internal/wasm/store.go
codefromthecrypt: oops :)

PR #1284, opened 2023-03-24
Avoids returning ExitError on exit code zero, and optimizes for no allocations

### LINKED ISSUE #1283: An error occurred running zig wasi
An error occurred when build `wasi` using `zig build-exe src/main.zig -target wasm32-wasi`:
```sh
failed to instantiate module: module "" closed with exit_code(0)
```
It runs successfully when remove the `_start` function, as follows: 
```go
config := wazero.NewModuleConfig().WithStartFunctions()
mod, err := r.InstantiateWithConfig(ctx, addWasm, config)
...
```

codefromthecrypt:
This removes the need to handle ExitError for success (exit code zero). It also removes the module name from the struct which allows us to re-use the same error object for the common case of success.

Fixes #1283

### DIFF
--- a/imports/assemblyscript/assemblyscript.go
+++ b/imports/assemblyscript/assemblyscript.go
@@ -177,7 +177,7 @@ func abort(ctx context.Context, mod api.Module, _ []uint64) {
 	_ = mod.CloseWithExitCode(ctx, exitCode)
 
 	// Prevent any code from executing after this function.
-	panic(sys.NewExitError(mod.Name(), exitCode))
+	panic(sys.NewExitError(exitCode))
 }
 
 // traceDisabled ignores the input.
--- a/runtime.go
+++ b/runtime.go
@@ -34,11 +34,14 @@ type Runtime interface {
 	//
 	//	mod, _ := r.Instantiate(ctx, wasm)
 	//
-	// See InstantiateWithConfig for configuration overrides.
+	// # Notes
+	//
+	//   - See notes on InstantiateModule for error scenarios.
+	//   - See InstantiateWithConfig for configuration overrides.
 	Instantiate(ctx context.Context, source []byte) (api.Module, error)
 
 	// InstantiateWithConfig instantiates a module from the WebAssembly binary
-	// (%.wasm) or errs if invalid.
+	// (%.wasm) or errs for reasons including exit or validation.
 	//
 	// Here's an example:
 	//	ctx := context.Background()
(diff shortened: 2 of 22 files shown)


runtime.go
```diff
@@ -87,15 +91,25 @@ type Runtime interface {
...
 	//
-	// While CompiledModule is pre-validated, there are a few situations which can cause an error:
+	// While CompiledModule is pre-validated, there are a few situations which
+	// can cause an error:
 	//   - The module name is already in use.
-	//   - The module has a table element initializer that resolves to an index outside the Table minimum size.
+	//   - The module has a table element initializer that resolves to an index
+	//     outside the Table minimum size.
 	//   - The module has a start function, and it failed to execute.
+	//   - The module was compiled to WASI and exited. Even on success, exit
```
achille-roussel, line 109: Do you think it would be useful to not error in that case? 

It seems like the error would always be non-nil for WASI programs, or depend on compiler specific behavior (e.g calling proc_exit(0) after main returns).

It reminds me of how the net/http.Server always returns a non nil error when calling ListenAndServe... it's such a pain to deal with.

achille-roussel (approved):

runtime.go
codefromthecrypt: I was looking through the history and didn't see a reason why. I think at first it was trying to be like a CancelationError. SGTM to remove that

codefromthecrypt: I think it is less impact to do this as a special case, like for start functions invoked inside Instantiate, not for all functions.

achille-roussel: Agreed

mathetake (approved):

codefromthecrypt:
I took an advantage to make a small api break, removing the module name from the error message. If someone wants the module name, they can take it from the call site. Removing the module name from the struct allows us to use a constant for success, which makes invoking a normal module a bit lighter.

codefromthecrypt:
only a few outside wazero are allocating their own exit errors. I will help migrate if re-approved https://github.com/search?q=wazero+NewExitError+-org%3Atetratelabs&type=code

cc @ncruces @lthibault @mmathys

imports/assemblyscript/assemblyscript.go
```diff
@@ -177,7 +177,7 @@ func abort(ctx context.Context, mod api.Module, _ []uint64) {
 	_ = mod.CloseWithExitCode(ctx, exitCode)
 
 	// Prevent any code from executing after this function.
-	panic(sys.NewExitError(mod.Name(), exitCode))
+	panic(sys.NewExitError(exitCode))
```
codefromthecrypt, line 180: so this is the break for those defining closing host functions.. don't put the module name in

evacchi (approved):

lthibault:
All good on my end.  Thanks for the heads-up!

Also, +1 to the decision to remove the module name from the error message.  Much cleaner this way 👍

ncruces:
Figured out that by compiling as a "reactor" (`-mexec-model=reactor`) I won't have the dependency on `proc_exit`.

codefromthecrypt:
@ncruces cool.. thanks for sharing! (I feel I often say this to you, but I always mean it :D)

PR #1285, opened 2023-03-24
Shrink the store's nameToModule map

ckaznocha:
Putting this out there. Go does not shrink maps when elements are deleted so a long lived map with lots of elements added and deleted can consume a lot of memory.

This change would track the largest size the nameToNode map has grown to and if that size is more than twice the size it needs to be it gets replaced with a new map. Since we wouldn't want to be constantly resizing small maps I've set a floor and set the initial capacity of the map to that floor. I some what arbitrarily chose 10,000 as the floor.

The scenario this helps with is if a long running application has bursty periods of time when it instantiates a lot of named modules but most of the time has a very low volume of named modules.

I think it might be nice to make an option to disable or adjust the floor for applications running in memory constrained environments who know they will never need to instantiate a lot of named modules.

If there is interest in merging this I'll update the tests and add that option.

### DIFF
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -12,6 +12,8 @@ import (
 	"github.com/tetratelabs/wazero/sys"
 )
 
+const initialNameToNodeListSize = 10_000
+
 type (
 	// Store is the runtime representation of "instantiated" Wasm module and objects.
 	// Multiple modules can be instantiated within a single store, and each instance,
@@ -31,6 +33,10 @@ type (
 		// nameToModule holds the instantiated Wasm modules by module name from Instantiate.
 		// It ensures no race conditions instantiating two modules of the same name.
 		nameToModule map[string]*ModuleInstance // guarded by mux
+
+		// nameToNodeCap tracks the growth of the nameToNode map in order to
+		// track when to shrink it.
+		nameToNodeCap int // guarded by mux
 
 		// EnabledFeatures are read-only to allow optimizations.
 		EnabledFeatures api.CoreFeatures
@@ -258,7 +264,7 @@ func NewStore(enabledFeatures api.CoreFeatures, engine Engine) *Store {
 		typeIDs[k] = v
 	}
 	return &Store{
-		nameToModule:     map[string]*ModuleInstance{},
+		nameToModule:     make(map[string]*ModuleInstance, initialNameToNodeListSize),
 		EnabledFeatures:  enabledFeatures,
 		Engine:           engine,
 		typeIDs:          typeIDs,
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -12,9 +12,10 @@ import (
 	"github.com/tetratelabs/wazero/sys"
 )
 
-// initialNameToModuleListSize is the initial capacity of the nameToModule map.
-// The capacity will never be smaller than this.
-const initialNameToModuleListSize = 100
+// nameToModuleShrinkThreshold is the size the nameToModule map can grow to
+// before it starts to be monitored for shrinking.
+// The capacity will never be smaller than this once the threshold is met.
+const nameToModuleShrinkThreshold = 100
 
 type (
 	// Store is the runtime representation of "instantiated" Wasm module and objects.
@@ -266,8 +267,8 @@ func NewStore(enabledFeatures api.CoreFeatures, engine Engine) *Store {
 		typeIDs[k] = v
 	}
 	return &Store{
-		nameToModule:     make(map[string]*ModuleInstance, initialNameToModuleListSize),
-		nameToModuleCap:  initialNameToModuleListSize,
+		nameToModule:     map[string]*ModuleInstance{},
+		nameToModuleCap:  nameToModuleShrinkThreshold,
 		EnabledFeatures:  enabledFeatures,
 		Engine:           engine,
 		typeIDs:          typeIDs,
--- a/internal/wasm/store.go
+++ b/internal/wasm/store.go
@@ -12,7 +12,9 @@ import (
 	"github.com/tetratelabs/wazero/sys"
 )
 
-const initialNameToNodeListSize = 10_000
+// initialNameToModuleListSize is the initial capacity of the nameToModule map.
+// The capacity will never be smaller than this.
+const initialNameToModuleListSize = 100
 
 type (
 	// Store is the runtime representation of "instantiated" Wasm module and objects.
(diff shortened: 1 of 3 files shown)


internal/wasm/store.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/tetratelabs/wazero/sys"
 )
 
+const initialNameToNodeListSize = 10_000
```
codefromthecrypt, line 15: I don't know the right number here either, I guess this is the size of the total in-flight modules? Whatever this controls we should add as a comment.

ckaznocha: Yeah ideally it would be some kind of average of the expected in-flight modules. Really some number that is low enough not to consume too much memory but high enough to fit a typical set of named modules. Since it can still grow I think undershooting is better than overshooting

codefromthecrypt: gotcha, and I think it is cute that you feel 10k is a low number of in-flight modules. seems you may be in a competition with @ShivanshVij on wasm throughput? ;)

ckaznocha: lol, in truth it's more that I thought 10k string headers was not much memory which is still telling…

ShivanshVij: The following only matters if the variable is tied to performance/throughput:

Could we make this number editable? Like even if it was an exported variable as long as it was a knob we could tweak I’d be happy.

ShivanshVij:
It might also make sense to add a knob disabling this - we want the most throughput possible on the system and a massive preallocated map works for that. 

Not the most efficient use of memory, sure, but you can always buy more ram.

internal/wasm/store.go
ckaznocha: Yes, I think it can be made configurable!

ShivanshVij: Would we have to move the variable out of the internal package then?

ShivanshVij: I’d be just as happy if we just made it possible to disable (and keep it enabled by default, that doesn’t matter to us)

codefromthecrypt:
FWIW I prefer a better, affordable default value vs a million config switches for something that is unlikely to be set by most people. Config is hugely distracting and I recall some places I worked at you'd have a bunch of weedy config and it was hard to feel the project was relevant as they didn't config important things and fuzzed over minute knobs.

my 2p

ckaznocha:
Updated to shrink the `nameToModule` to adjust to the changes in #1359. Added tests and reduced the initial cap to 100. No settings added for now. Easer to add later than it is remove premature settings!

codefromthecrypt (approved):
Looks good to me, can you find a relevant benchmark and make sure the runtime cost isn't higher?

ncruces:
Using `initialNameToModuleListSize` seems absolutely necessary to use as a minimum when reclaiming space (to avoid degenerate allocation patterns). But I wonder if it really is important to set the initial capacity.

Since anonymous modules many runtimes will have… less than a handful names in the map? The fixed cost of 100 is probably low, and if it's beneficial to other use cases, I guess it's fine. But I just wonder if it makes sense.

ckaznocha:
@ncruces fair point. Removed the initial map size so shrinking shouldn't have much impact until the map grows organically.

@codefromthecrypt benchmarks
```
goos: darwin
goarch: arm64
                                       │    old.txt    │               new.txt                │
                                       │    sec/op     │    sec/op     vs base                │
Initialization/interpreter-10             11.31µ ± ∞ ¹   12.33µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Initialization/interpreter-multiple-10   12.064µ ± ∞ ¹   9.467µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Initialization/compiler-10                10.76µ ± ∞ ¹   10.18µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Initialization/compiler-multiple-10       10.48µ ± ∞ ¹   10.50µ ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                   11.14µ         10.57µ        -5.07%

                                       │    old.txt    │                new.txt                │
                                       │     B/op      │     B/op       vs base                │
Initialization/interpreter-10            132.3Ki ± ∞ ¹   132.3Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Initialization/interpreter-multiple-10   132.3Ki ± ∞ ¹   132.3Ki ± ∞ ¹       ~ (p=1.000 n=1) ³
Initialization/compiler-10               132.3Ki ± ∞ ¹   132.3Ki ± ∞ ¹       ~ (p=1.000 n=1) ³
Initialization/compiler-multiple-10      132.3Ki ± ∞ ¹   132.3Ki ± ∞ ¹       ~ (p=1.000 n=1) ³
geomean                                  132.3Ki         132.3Ki        -0.00%
² all samples are equal
³ need >= 4 samples to detect a difference at alpha level 0.05

Initialization/interpreter-10            23.00 ± ∞ ¹   23.00 ± ∞ ¹       ~ (p=1.000 n=1) ²
Initialization/interpreter-multiple-10   23.00 ± ∞ ¹   23.00 ± ∞ ¹       ~ (p=1.000 n=1) ²
Initialization/compiler-10               23.00 ± ∞ ¹   23.00 ± ∞ ¹       ~ (p=1.000 n=1) ²
Initialization/compiler-multiple-10      23.00 ± ∞ ¹   23.00 ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                  23.00         23.00        +0.00%
² all samples are equal
```

internal/wasm/store_module_list.go
```diff
@@ -29,6 +29,22 @@ func (s *Store) deleteModule(m *ModuleInstance) error {
 
 	if m.ModuleName != "" {
 		delete(s.nameToModule, m.ModuleName)
+
+		// shrink the map if it's allocated more than twice the size of the list
+		nameToModuleLen := len(s.nameToModule)
+		if s.nameToModuleCap > nameToModuleShrinkThreshold && s.nameToModuleCap >= nameToModuleLen*2 {
```
ncruces, line 35: Say `s.nameToModuleCap == 101` and `nameToModuleLen == 50`. The test passes, and we enter the if to resize the map from `101`, to `100` (`nameToModuleShrinkThreshold`). We'll have 50 slots of spare capacity, but it hardly seems worth it to resize.

One fix is to do:
```go
if s.nameToModuleCap > nameToModuleShrinkThreshold*2 && s.nameToModuleCap >= nameToModuleLen*2 {
```

Another, which I personally find clearer is:
```go
newCap := len(s.nameToModule)
if newCap < nameToModuleShrinkThreshold {
	newCap = nameToModuleShrinkThreshold
}
if newCap*2 <= s.nameToModuleCap {
	// do the resize
}
```

internal/wasm/store_module_list.go
```diff
@@ -29,6 +29,22 @@ func (s *Store) deleteModule(m *ModuleInstance) error {
 
 	if m.ModuleName != "" {
 		delete(s.nameToModule, m.ModuleName)
+
+		// shrink the map if it's allocated more than twice the size of the list
```
mathetake, line 33: nit
```suggestion
		// Shrink the map if it's allocated more than twice the size of the list.
```

ncruces (approved):

mathetake (approved):

PR #1286, opened 2023-03-24
fixes nil panic on close

codefromthecrypt:
The last PR over-solved the WASI close problem as it special cased both "_start", used by WASI, and the start section (wasm level) which isn't used by WASI. In the latter case, we ended up returning nil for both the module *and* the error result. We should never do that.. If we coerce exit error zero to nil, we have to return a non-nil module, even if it is unusable as otherwise code like `defer mod.Close(ctx)` will panic.

Note: it would be entirely rare that someone would define a start section and call close in it. This is just something I noticed while looking carefully at the logic.

### DIFF
--- a/runtime.go
+++ b/runtime.go
@@ -308,11 +308,6 @@ func (r *runtime) InstantiateModule(
 		if code.closeWithModule {
 			_ = code.Close(ctx) // don't overwrite the error
 		}
-		if se, ok := err.(*sys.ExitError); ok {
-			if se.ExitCode() == 0 { // Don't err on success.
-				err = nil
-			}
-		}
 		return
 	}
 
--- a/runtime_test.go
+++ b/runtime_test.go
@@ -478,8 +478,9 @@ func TestRuntime_InstantiateModule_ExitError(t *testing.T) {
 		expectedErr error
 	}{
 		{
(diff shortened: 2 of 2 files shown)


runtime.go
```diff
@@ -308,11 +308,6 @@ func (r *runtime) InstantiateModule(
 		if code.closeWithModule {
 			_ = code.Close(ctx) // don't overwrite the error
 		}
-		if se, ok := err.(*sys.ExitError); ok {
-			if se.ExitCode() == 0 { // Don't err on success.
-				err = nil
```
codefromthecrypt, line 313: if we clear the err here, we end up returning (nil, nil) which would be bad ;)

runtime_test.go
```diff
@@ -478,8 +478,9 @@ func TestRuntime_InstantiateModule_ExitError(t *testing.T) {
 		expectedErr error
 	}{
 		{
-			name:     "start: exit code 0",
-			exitCode: 0,
+			name:        "start: exit code 0",
+			exitCode:    0,
+			expectedErr: sys.NewExitError(0),
```
codefromthecrypt, line 483: so people adding a start function which calls close will see an exit error zero. this is likely no-one.

PR #1289, opened 2023-03-24
Add go-pdfium to users

jerbob92:
@codefromthecrypt do you want me to also add [pdfium-cli](https://github.com/klippa-app/pdfium-cli)? It's just using go-pdfium but it does bring Wazero to end users.


codefromthecrypt (approved):

codefromthecrypt:
you can add the suffix as another commit, you deserve them :D

PR #1290, opened 2023-03-25
sysfs: stubs in approach to defining a new file type

codefromthecrypt:
After the last couple months of thinking, I eventually got to the idea that defining a new file type has two problems:

* What sort of type do we define?
* How in the world do we migrate to a new type?

What this PR does is spike an approach to the latter without solidifying the former. My thinking is that the best way is to incrementally migrate functions one-by-one to a new type. Meanwhile, temporarily expose a `.File()` function to access the `fs.File` we want to rid eventually. This makes it easy to identify what's left as you just look for any calls to `.File()`.

If this approach helps, lemme know and I can polish for merge. I mainly had this branch locally so want to decide to delete it or not!

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -97,9 +97,9 @@ func fdAllocateFn(_ context.Context, mod api.Module, params []uint64) syscall.Er
 		return syscall.EINVAL
 	}
 
-	st, err := f.Stat()
-	if err != nil {
-		return platform.UnwrapOSError(err)
+	st, errno := f.Stat()
+	if errno != 0 {
+		return errno
 	}
 
 	if st.Size >= tail {
@@ -107,7 +107,7 @@ func fdAllocateFn(_ context.Context, mod api.Module, params []uint64) syscall.Er
 		return 0
 	}
 
-	osf, ok := f.File.(truncateFile)
+	osf, ok := f.File.File().(truncateFile)
 	if !ok {
 		return syscall.EBADF
 	}
@@ -154,7 +154,7 @@ func fdDatasyncFn(_ context.Context, mod api.Module, params []uint64) syscall.Er
 	if f, ok := fsc.LookupFile(fd); !ok {
 		return syscall.EBADF
 	} else {
-		return sysfs.FileDatasync(f.File)
+		return sysfs.FileDatasync(f.File.File())
 	}
 }
 
@@ -211,18 +211,18 @@ func fdFdstatGetFn(_ context.Context, mod api.Module, params []uint64) syscall.E
 	}
 
 	var fdflags uint16
-	var st fs.FileInfo
-	var err error
+	var st platform.Stat_t
+	var errno syscall.Errno
 	if f, ok := fsc.LookupFile(fd); !ok {
 		return syscall.EBADF
-	} else if st, err = f.File.Stat(); err != nil {
-		return platform.UnwrapOSError(err)
-	} else if _, ok := f.File.(io.Writer); ok {
+	} else if st, errno = f.Stat(); errno != 0 {
+		return errno
+	} else if _, ok := f.File.File().(io.Writer); ok {
 		// TODO: maybe cache flags to open instead
 		fdflags = wasip1.FD_APPEND
 	}
 
-	filetype := getWasiFiletype(st.Mode())
+	filetype := getWasiFiletype(st.Mode)
 	writeFdstat(buf, filetype, fdflags)
 
 	return 0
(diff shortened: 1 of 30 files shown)


internal/platform/file.go
```diff
@@ -3,8 +3,83 @@ package platform
...
+	//
+	// The following errors are expected:
+	//   - syscall.EBADF if the file or directory was closed.
+	//
+	// # Notes
+	//
+	//   - An fs.FileInfo backed implementation sets atim, mtim and ctim to the
+	//     same value.
+	//   - Windows allows you to stat a closed directory.
+	Stat() (Stat_t, syscall.Errno)
```
codefromthecrypt, line 42: so basically this stubs in by only defining stat and close. Main point is to carve in a way to migrate (to anything). then if it works, multiple people can work on whatever the direction is easier.

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -97,17 +97,17 @@ func fdAllocateFn(_ context.Context, mod api.Module, params []uint64) syscall.Er
...
+		return errno
 	}
 
 	if st.Size >= tail {
 		// We already have enough space.
 		return 0
 	}
 
-	osf, ok := f.File.(truncateFile)
+	osf, ok := f.File.File().(truncateFile)
```
codefromthecrypt, line 110: so the idea of this iterative approach shows up places like this. Any time you see `File.File()`, you know there's a function not yet implemented on the base type that we need. When all of these disappear, the file type is feature complete and the function `File.File()` can be removed.

achille-roussel:
The migration path makes a lot of sense.

Have you recorded where your mind's at regarding what the final form could look like (e.g. a sketch of what would be exposed)?

codefromthecrypt:
> Have you recorded where your mind's at regarding what the final form could look like (e.g. a sketch of what would be exposed)?

## no big design up front on all functions.

I think in our change culture, we don't need to do a big design up front, deliberating on 10 or more function signatures before implementing them. The reason is that it would end up bad and it is hard to reason with. The signatures would end up different in the PRs anyway. So this PR chooses exactly to not try to do that, which keeps review burden down by breaking it up into incremental steps. This way, people can see the exact signatures, as they are used vs try to imagine them in their head. It also avoids a massive change to try to figure them out up front. Imagine multiplying out the CallWithStack thing into one google doc or a huge PR. I don't have time for it and I work full time. I don't think we have time for ceremony and I don't think it will be beneficial vs iterative approach.

## How not what

So, in general I thought about specific on how to expose, no the exact form of each of the maybe 10 functions. Basically, the functions would be exposed in syscall form (so return errno which collects error mapping inside the File object which helps with for example the windows mapping sprawl etc).

Concretely, I called attention to how we know we are done, vs trying to solve each in this PR or some spec doc. We know things are done when we can remove `File.File()`. If you like, you can search for all places `File.File()` and unique them.

Others, like `Write` may have some stylistic design depending on which way we go, but again, I can't imagine doing that before we can see how they are used. Practice helps us from wandering on what ifs that aren't relevant to the call sites. Remember, the only purpose of this API is for implementing things like WASI, this isn't for general use.

Here's an example, of what I mean for Chmod. So we already went through work for this on sysfs.FS, so no need to redo that work. We simply copy/paste and make it coherent. 

```go
```

```go
```

## Yes it is messy for a while, but we've been stuck

There will eventually be some repackaging needed, but I don't think that should happen now. For example, I think sysfs.FS is great and having default stuff there is also great, but probably the OS file impl of that, its code should be in sysfs/internal somewhere. platform.File would move to the sysfs package once things are done as it isn't coherent where it is, just there's also no harm leaving the file code there until we're done.

## Summary

We need to start the process and we've been stalled. I think this process actually unlocks us so we can get out of the muck possibly in a month.

After this PR we'd inline each function call type in separate PRs, and review them in. They don't need to be in perfect form, as with practice we'll maybe decide some slight adjustments. However, each PR should be as good as we can do then, and obviously with coherence and godocs.

Then, once everything is in the right place and still internal, we'd do a coherence sweep changing any of the function signatures based on something that feels off or maybe after testing doesn't work as right as we like.

So, step-wise.
1. land base type (this PR)
2. convert functions one-by-one (some easy like chmod, some less easy)
3. remove File.File()
4. move platform.File into sysfs and its os supporting code to a package under the sysfs directory
5. do a coherence sweep that the forms of sysfs.FS and File feel like we do, knowing that changing either of this has impact as they are already in use!
6. refactor the tests to remove duplication and allow for external implementation (now that we aren't internally drifting)
7. top-level the sysfs package along with any built-in impls such as the fs.FS wrapper, the composite FS and anything else, such as a mem one that may come around.

I think this can finish in a month or two and work with part-time attention, with only black swan being how wasi-filesystem wants to make things work in the future. We can possibly choose to not expose until we try using that.

codefromthecrypt:
PS I'll bump out the functions to convert to an issue.. I can make a list!

codefromthecrypt:
ok looks like after this change, about 13 functions, maybe I missed one or two.

mathetake (approved):

PR #1291, opened 2023-03-27
Refactors NewModuleEngine and NewCallEngine

mathetake:
This drops the unnecessary parameters from `NewModuleEngine` and `NewCallEngine` 
interface methods.


codefromthecrypt (commented):
excellent! you are on a roll

PR #1292, opened 2023-03-27
Removes validatedActiveElementSegment

mathetake:
We allocated `[]validatedActiveElementSegment` during compilation, but 
no need to do that at the end of the day.


codefromthecrypt (commented):
👍

PR #1295, opened 2023-03-27
fs: stat and cache mode for stdio devices

evacchi:
fs: stat and cache mode for stdio devices

Ensure that stdio device modes are consistent with the given
file descriptors by stat'ing, instead of returning mocks.

Use `Stat()` on `poll_oneoff()` too, instead of `IsTerminal()`,
thus avoiding a useless syscall.

Delete leftover type decl `fileModeStat`.

Delete IsPlatform() and related, as they are superseded by the stat() check.

---
I originally noticed this because I was [trying to run this Python interpreter but the REPL hung](https://github.com/vmware-labs/webassembly-language-runtimes/releases/tag/python%2F3.11.1%2B20230217-15dfbed).

The main culprit is that we are stubbing stdin/out/err and `fs.ModeCharDevice` must imply `fs.ModeDevice`; this in turn causes 

https://github.com/tetratelabs/wazero/blob/a4226906cf30ba5261d2fab2ff4df62a8109d6ea/imports/wasi_snapshot_preview1/fs.go#L360-L365

to return UNKNOWN instead (value 0)

I have dug further, and it seems that we are defaulting stdin/stdout/stderr to be block devices instead of CharDevices

https://github.com/tetratelabs/wazero/blob/372e0222ddb7dbc64bc7829fc112dd0b88a15e9e/internal/sys/fs.go#L83-L87

this in turn causes my new `TestStdioStat` to fail, but if I flip those defaults I cause _other_ tests to fail.

Hence, it follows that we are not really being consistent with the operating system, but we are testing our own mocks (at least in some cases) 🤔 

~Draft for now.~


imports/wasi_snapshot_preview1/poll.go
```diff
@@ -142,9 +142,10 @@ func processFDEvent(mod api.Module, eventType byte, inBuf []byte) syscall.Errno
 	// files.
 	errno := syscall.ENOTSUP
 	if eventType == wasip1.EventTypeFdRead {
-		if _, ok := fsc.LookupFile(fd); ok {
-			// If fd is a pipe, IsTerminal is false.
-			if platform.IsTerminal(uintptr(fd)) {
+		if f, ok := fsc.LookupFile(fd); ok {
+			st, _ := f.Stat()
+			// if fd is a pipe, then it is not a char device (a tty)
+			if st.Mode&fs.ModeCharDevice != 0 {
```
evacchi, line 148: here I am also proposing to Stat() instead of calling IsTerminal() , but I am not sure which one is cheaper

evacchi:
In practice, my doubts amount to: "shouldn't we stat stdin/stderr/stdout at boot time instead of mocking?"

internal/sys/fs.go
```diff
@@ -34,7 +34,7 @@ const (
 
 const (
 	modeDevice     = uint32(fs.ModeDevice | 0o640)
-	modeCharDevice = uint32(fs.ModeCharDevice | 0o640)
+	modeCharDevice = uint32(fs.ModeDevice | fs.ModeCharDevice | 0o640)
```
achille-roussel, line 37: This subtle construct of requiring both ModeDevice and ModeCharDevice has bitten me so many times.

imports/wasi_snapshot_preview1/poll.go
achille-roussel: The mode of a file would never change so you could potentially cache it to amortize then stat call. 

I'd leave this to a later update when you have data to demonstrate that the optimization is needed.

achille-roussel (approved):
Nice fix!

imports/wasi_snapshot_preview1/poll.go
codefromthecrypt: after verifying the original issue that led to platform.IsTerminal works this way, go back and delete platform.IsTerminal! https://github.com/tetratelabs/wazero/pull/935 was an issue raised both in python and RethinkRaw, so maybe manually check both (especially as there's now a working python build!)

internal/sys/fs.go
```diff
@@ -301,25 +301,16 @@ func stdioWriter(w io.Writer, defaultStat stdioFileInfo) *FileEntry {
 }
 
 func stdioStat(f interface{}, defaultStat stdioFileInfo) fs.FileInfo {
-	if f, ok := f.(*os.File); ok && platform.IsTerminal(f.Fd()) {
-		return stdioFileInfo{defaultStat[0], modeCharDevice}
+	if f, ok := f.(*os.File); ok {
+		mode := modeCharDevice
```
codefromthecrypt, line 305: so as far as I understand, this will fallback to a char device if there's an error getting stat. I would prefer to change the signature to propagate the error. It might take changing multiple functions, but I think it doesn't have end user API impact.

evacchi: ok will try

imports/wasi_snapshot_preview1/poll.go
evacchi: indeed, this works with Python (which I checked manually, will check RethinkRaw too)

evacchi: checked both and they seem to work, but I'll ping @ncruces to make sure :) would you like to check? 🙏

ncruces: Oh, hey… users! I'll check for sure, thanks for the heads up.

I really need to spin off `dcraw` into a package with unit tests.

ncruces: @evacchi all good, it seems.

Issue #1202, opened 2023-03-06
wazeroir: avoid allocations per Operation


mathetake:
Currently, we result in allocating an [wazeroir.Operation](https://github.com/tetratelabs/wazero/blob/25493fe271d39a55dfd3581e5c8d8c06041e5822/internal/wazeroir/operations.go#L132-L133) object each time without caching or whatever. This is one of the bottlenecks during compilation (!= execution or instantiation) besides #1060:

<img width="616" alt="image" src="https://user-images.githubusercontent.com/13513977/223015267-b719876f-80c3-4925-bd99-69ea4c631c3c.png">

_memory profiling result on the compilation of GOOS=js binary_

The root cause of this is that we emit a concrete type for `Operation`  interface which is a translation for Wasm instruction.
For example, each `global.get 1` allocates `wazeroir.OperationGlobalGet{index: 1}`:
https://github.com/tetratelabs/wazero/blob/25493fe271d39a55dfd3581e5c8d8c06041e5822/internal/wazeroir/compiler.go#L922-L925

And we store them in the slice of Operation interfaces 

https://github.com/tetratelabs/wazero/blob/25493fe271d39a55dfd3581e5c8d8c06041e5822/internal/wazeroir/compiler.go#L221-L222

One possible solution to this is to make it streaming given that our current compiler/interpreter does stream compilation. Instead of translating the Wasm instructions all at once, just make wazeroir compiler stateful and use the fixed-length bytes array (the length could be the maximum size of concrete operation struct type which is statically known) to store the IR info from compilers/interpreters. 

There might be an easier and better way to optimize this, so I'm open to discussion!

mathetake:
another idea could be to serializeOperation as bytes and store it in `Operations []byte`, and decodes them via unsafe.Pointer and reflect.SliceHeader by compiler/interpreter.

achille-roussel:
Have you considered changing `Operation` from an interface to a type with a structure like this?
```go
type Operation struct {
  Kind OperationKind
  // all possible field types
  // ...
}
```
Go doesn't have union so you can either use a representation that is very memory greedy and declare each parameters as struct fields (this is the approach used in https://pkg.go.dev/go.jpap.org/zydis#DecodedInstruction), for example:
```go
type Operation struct {
  Kind OperationKind
  Label struct { // fields from OperationLabel
    Label Label
  }
  Br struct { // fields from OperationBr
    Target Label
  }
}
```
[...]

PR #1296, opened 2023-03-27
(wip) refactor wazeroir to use quasi-union

evacchi:
a PoC of the change mentioned in #1202 the way I have understood it. Commented out code is intended as this is still wip


internal/wazeroir/operations.go
```diff
@@ -1171,22 +1204,26 @@ func (OperationSet) Kind() OperationKind {
...
+//// Kind implements Operation.Kind
+//func (OperationGlobalGet) Kind() OperationKind {
+//	return OperationKindGlobalGet
+//}
 
-// Kind implements Operation.Kind
-func (OperationGlobalGet) Kind() OperationKind {
-	return OperationKindGlobalGet
+func NewOperationGlobalGet(index uint32) OperationUnion {
+	return OperationUnion{OpKind: OperationKindGlobalGet, Us: []uint64{uint64(index)}}
```
evacchi, line 1226: although we are obviously giving up type-safety, it may be still worth to introduce a named constructor for readability and to avoid mistakes. Consider commit https://github.com/tetratelabs/wazero/pull/1296/commits/34861a200eed3bf1f5b196093f3f21a7057f5680 -- at the beginning I forgot to explicitly set `OpKind` and everything defaulted to `Unknown` :D

mathetake, line 1226: please do not use `Us []uint64` in general as much as possible, to avoid the allocation here. Instead, you can add `uint32` or `uint64` field in the union.

evacchi: sure, I didn't add new fields as I wanted it to be identical to `interfaceOp` so that we can unify easily at some point

https://github.com/tetratelabs/wazero/blob/main/internal/engine/interpreter/interpreter.go#L998-L1004

if we add a uint32 field I feel like _maybe_ we should add it to `interfaceOp` too? 🤔

mathetake: You won’t need interpreterOp any more with this union.

mathetake: You can delete the type

mathetake: what you are supposed to do is essentially lift entire lowerIr function (https://github.com/tetratelabs/wazero/blob/main/internal/engine/interpreter/interpreter.go#L290-L732) into wazeroir/compiler.go, and delete  `interpreterOp` and reuse the union everywhere in interpreter.

evacchi: right, I was trying to do it instruction-by-instruction because it is very disruptive for the compiler part, so my plan was to duplicate `interpreterOp` into a new struct, and then at some point delete the original one.

evacchi: it looks like we can use 2 uint64s in most places except

		case wazeroir.OperationV128Shuffle:
			op.us = make([]uint64, 16)
			for i, l := range o.Lanes {
				op.us[i] = uint64(l)
			}

mathetake: 👍

evacchi:
(more wip on https://github.com/evacchi/wazero/tree/wip/refactor-wazeroir-quasiunion)

mathetake:
I guess we can close this and separate PRs are coming to us 🏁

PR #1297, opened 2023-03-28
Implements api.Function directly over engine specific callEngine

mathetake:
Consequently, we can not only avoid the unnecessary allocation in `api.Module.ExportFunction`,
but also nuke the `wasm.CallEngine` interface.


codefromthecrypt (approved):
Looks great! did any of this drift the docs on how calls work?

PR #1298, opened 2023-03-28
interpreter: add u1, u2 int64 to `interpreterOp`

evacchi:
In preparation to further refactoring on the compiler side,
add `u1, u2 uint64` fields to `interpreterOp` avoid a useless
allocation of a slice when only 2 int64s are needed

### DIFF
--- a/internal/engine/interpreter/interpreter.go
+++ b/internal/engine/interpreter/interpreter.go
@@ -212,6 +212,7 @@ type interpreterOp struct {
 	kind     wazeroir.OperationKind
 	b1, b2   byte
 	b3       bool
+	u1, u2   uint64
 	us       []uint64
 	rs       []*wazeroir.InclusiveRange
 	sourcePC uint64
@@ -313,10 +314,9 @@ func (e *engine) lowerIR(ir *wazeroir.CompilationResult) (*code, error) {
 			// as we translate branch operations to the direct address jmp.
 			continue
 		case wazeroir.OperationBr:
-			op.us = make([]uint64, 1)
 			if o.Target.IsReturnTarget() {
 				// Jmp to the end of the possible binary.
-				op.us[0] = math.MaxUint64
+				op.u1 = math.MaxUint64
 			} else {
 				labelID := o.Target.ID()
 				addr, ok := labelAddress[labelID]
@@ -325,11 +325,11 @@ func (e *engine) lowerIR(ir *wazeroir.CompilationResult) (*code, error) {
 					// the target is not emitted yet, so resolve the address later.
 					onLabelAddressResolved[labelID] = append(onLabelAddressResolved[labelID],
 						func(addr uint64) {
-							op.us[0] = addr
+							op.u1 = addr
 						},
 					)
 				} else {
-					op.us[0] = addr
+					op.u1 = addr
 				}
 			}
 		case wazeroir.OperationBrIf:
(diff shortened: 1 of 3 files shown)


internal/engine/interpreter/interpreter.go
```diff
@@ -212,6 +212,7 @@ type interpreterOp struct {
 	kind     wazeroir.OperationKind
 	b1, b2   byte
 	b3       bool
+	u1, u2   uint64
 	us       []uint64
```
achille-roussel, line 216: Do you know if the number of items in this slice is unbounded?

evacchi: it's likely to be bound to the max num of targets here:

			targets := append([]*wazeroir.BranchTargetDrop{o.Default}, o.Targets...)
			op.Rs = make([]*wazeroir.InclusiveRange, len(targets))
			op.Us = make([]uint64, len(targets))


this is emitted by the compiler at `case wasm.OpcodeBrTable:` and `numTargets` is

 		numTargets, n, err := leb128.DecodeUint32(r)

achille-roussel: Answering my own question, I found your comment here https://github.com/tetratelabs/wazero/pull/1296#discussion_r1150550133

It may be interesting to change us to *[16]uint64 to save space on the union type since the field is almost never used (slices take 1 pointer and 2 ints).

achille-roussel: Sorry I wasn't very clear. 

I think having the u1 and u2 fields is the right change. 

I was suggesting we could also improve further by changing the slice type to *[16]uint64 so we waste less space in most cases.

evacchi:
oh no Windows *shakes fist*

internal/engine/interpreter/interpreter.go
mathetake: yeah, the br_table instruction needs unbounded list of indexes

mathetake (approved):

internal/engine/interpreter/interpreter.go
```diff
@@ -212,6 +212,7 @@ type interpreterOp struct {
 	kind     wazeroir.OperationKind
 	b1, b2   byte
 	b3       bool
```
mathetake, line 214: while you're at it, could you remove `b3 bool` and reuse the `b1 byte == 1` as a bool in the next PR?

internal/engine/interpreter/interpreter.go
achille-roussel: Oh interesting OK 👍 

Kind of crazy but kind of cool at the same time 😁

internal/engine/interpreter/interpreter.go
evacchi: sure!

evacchi: (probably better `b1 != 0`, I think the compiled code would be better)

evacchi: there are a few instances where this would not be possible

```
		case wazeroir.OperationITruncFromF:
			op.b1 = byte(o.InputType)
			op.b2 = byte(o.OutputType)
			op.b3 = o.NonTrapping

[...]

		case wazeroir.OperationV128ExtractLane:
			op.b1 = o.Shape
			op.b2 = o.LaneIndex
			op.b3 = o.Signed
```

PR #1299, opened 2023-03-28
wasm: type ModuleInstance.Closed as atomic.Uint64

twilly:
atomic.CompareAndSwapUint64 requires 64 bit alignment on the uint64 pointer passed to it. This is not easy to guarantee across platforms and may cause invalid pointer faults. However, atomic.Unit64 is guaranteed to have correct alignment.

See https://pkg.go.dev/sync/atomic#pkg-note-BUG for the quirky behaviors in atomic.

This change does alter the expected access behavior to the Closed field, but it appears that it's not (yet) consumed.

### DIFF
--- a/internal/wasm/module_instance.go
+++ b/internal/wasm/module_instance.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync/atomic"
 
 	"github.com/tetratelabs/wazero/api"
 	"github.com/tetratelabs/wazero/sys"
@@ -12,7 +11,7 @@ import (
 
 // FailIfClosed returns a sys.ExitError if CloseWithExitCode was called.
 func (m *ModuleInstance) FailIfClosed() (err error) {
-	if closed := atomic.LoadUint64(&m.Closed); closed != 0 {
+	if closed := m.Closed.Load(); closed != 0 {
 		return sys.NewExitError(uint32(closed >> 32)) // Unpack the high order bits as the exit code.
 	}
 	return nil
@@ -100,7 +99,7 @@ func (m *ModuleInstance) closeWithExitCode(ctx context.Context, exitCode uint32)
 
 func (m *ModuleInstance) setExitCode(exitCode uint32) bool {
 	closed := uint64(1) + uint64(exitCode)<<32 // Store exitCode as high-order bits.
-	return atomic.CompareAndSwapUint64(&m.Closed, 0, closed)
+	return m.Closed.CompareAndSwap(0, closed)
 }
 
 // ensureResourcesClosed ensures that resources assigned to CallContext is released.
--- a/internal/wasm/module_instance_test.go
+++ b/internal/wasm/module_instance_test.go
@@ -96,7 +96,7 @@ func TestCallContext_Close(t *testing.T) {
 				// Closing should not err.
 				require.NoError(t, tc.closer(ctx, m))
 
-				require.Equal(t, tc.expectedClosed, m.Closed)
+				require.Equal(t, tc.expectedClosed, m.Closed.Load())
 
 				// Verify our intended side-effect
 				require.Nil(t, s.Module(moduleName))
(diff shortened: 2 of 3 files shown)


twilly:
Ah, I forgot that these atomic types are new in 1.19 and wazero still supports 1.18.

It may be best to hold on this PR. :)

codefromthecrypt:
@twilly you read my mind, but this is exactly the change we should do when a version drops off, and anything that we are doing internally that's better off in 1.18 (vs 1.17) welcome.

Since we can't move up to 1.19 for several months, maybe re-open then? Usually, we try to keep the PR queue clean and I've already nagged folks as it is accumulating.

twilly:
> @twilly you read my mind, but this is exactly the change we should do when a version drops off, and anything that we are doing internally that's better off in 1.18 (vs 1.17) welcome.
> 
> Since we can't move up to 1.19 for several months, maybe re-open then? Usually, we try to keep the PR queue clean and I've already nagged folks as it is accumulating.

No problem at all. Pardon the dust. :)

ncruces:
But then, if `Closed` is not 64-bit aligned on 32-bit platforms, this is a bug.
The [suggested fix](https://pkg.go.dev/sync/atomic#pkg-note-BUG) would be to move `Closed uint64` to the top of the structure until then.

I understand that might be ugly (readability concerns), but maybe a TODO to do the update once 1.19 can be used helps.

@codefromthecrypt?

codefromthecrypt:
a good point. I focused on the code used for the fix, not the bug. PR welcome for the fix which comments `TODO: use atomic.Unit64 once the floor go version is 1.19`

twilly:
> But then, if `Closed` is not 64-bit aligned on 32-bit platforms, this is a bug. The [suggested fix](https://pkg.go.dev/sync/atomic#pkg-note-BUG) would be to move `Closed uint64` to the top of the structure until then.

The suggested fix works because structures are 64 bit aligned. You can also assert alignment in a test with unsafe.Alignof on the Closed field. I unfortunately don't have a 32 bit setup to really run these tests (this PR was a thought experiment for me while reading wazero's code).

The way I'd solve this problem is doing what [Go](https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/sync/atomic/type.go;l=138go.uber.org/atomic) and [go.uber.org/atomic](https://github.com/uber-go/atomic/blob/v1.10.0/uint64.go#L32) does: make uint64 atomic type. That way moving the field around isn't dangerous.

codefromthecrypt:
@twilly sounds ok to me. We have a lot of internal packages anyway, just any PR take care to not copy/paste other people's code/tests :D

ncruces:
Keep in mind that Uber isn't doing the same as Go.

Structures _are not_ guaranteed to be 64-bit aligned. The base of heap allocations (i.e. “malloc”) is. Uber relies on this by always returning heap allocated `*Uint64`. I'm pretty sure that using `*uint64` initialized with `new uint64` works as well.

Go is doing something different, that only works from 1.19, and only in the standard library: a [compiler intrinsic that forces alignment](https://github.com/golang/go/blob/f80e270babcddd28e29a52c1cb834ffed427bd8a/src/sync/atomic/type.go#L197).

codefromthecrypt:
I'm glad to have squinting eyes available!

ncruces:
But we're getting ahead of ourselves. This field is used as a bill, basically. Isn't it simpler to just switch to `uint32` which has none of these issues?

ncruces:
Nevermind, this stores the exit code as well. Disregard my previous comment.

codefromthecrypt:
@ncruces so the background is we use a uint64 to pack both the exit code and also whether it is closed or not, into the same value. Another way is to forbid one exit code, as otherwise we don't know the difference between exit 0 and never exited, or end up in a position of needing two fields for an atomic op. Thoughts?

ncruces:
You're right. I'd either move the field to the top of the structure or, if that's a readability issue, copy the Uber implementation. My only remark is that if you copy Uber, it needs to be `*Uint64` you can't embed it in the structure.

codefromthecrypt:
moving to the top with a TODO around 1.19 seems the best way out. PR welcome!

note this uint64 thing is used in multiple places. It may also drift offsets in the compiler engine, but unit tests will catch that

twilly:
I'm loving this lively discussion. :)

@ncruces Sorry about my wording; I clearly wasn't specific enough. You're absolutely right that it's not an intrinsic behavior of structs but rather it's in the technique of aligning the top of the struct.

I agree, perhaps just moving to the top of the struct is the simplest way out. It's not like we have to wait long for Go 1.21. I'm sure the odd field placement and comment means it'll get cleaned up, even if I forget about it in the coming months.

I'll toss up a different PR up soon because it's so simple. 👍

twilly:
New PR: #1321

PR #1301, opened 2023-03-29
Stop using pointer of function pointers in sys.Context

mathetake:
This reduces the allocations.


codefromthecrypt (approved):
agree. this was overengineering to be able to perform equality which is only needed in tests.

PR #1310, opened 2023-03-29
wazeroir: rm nullary Operations, move interpreterOp to UnionOperation

evacchi:
- refactor: remove nullary Operation objects, use unified object
- refactor: migrate nullary operations to the new OperationNullary
- refactor: remove commented-out code

While working on #1296 I realized there is a quick win / iterative change that could be merged to `main` already. Instead of creating a number of different instances of empty structs, we instead create a single `OperationNullary` with a `kind` field.

This can be migrated to the unified `interpreterOps` object later. Notice that since these are nullary, we may actually do away even with the constructor and define instead a bunch of vars.

### DIFF
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -1094,8 +1094,6 @@ func compileWasmFunction(cmp compiler, ir *wazeroir.CompilationResult) (*code, e
 		switch o := op.(type) {
 		case wazeroir.OperationLabel:
 			// Label op is already handled ^^.
-		//case wazeroir.OperationUnreachable:
-		//	err = cmp.compileUnreachable()
 		case wazeroir.OperationBr:
 			err = cmp.compileBr(o)
 		case wazeroir.OperationBrIf:
@@ -1134,10 +1132,6 @@ func compileWasmFunction(cmp compiler, ir *wazeroir.CompilationResult) (*code, e
 			err = cmp.compileStore16(o)
 		case wazeroir.OperationStore32:
 			err = cmp.compileStore32(o)
-		//case wazeroir.OperationMemorySize:
-		//	err = cmp.compileMemorySize()
-		//case wazeroir.OperationMemoryGrow:
-		//	err = cmp.compileMemoryGrow()
 		case wazeroir.OperationConstI32:
 			err = cmp.compileConstI32(o)
 		case wazeroir.OperationConstI64:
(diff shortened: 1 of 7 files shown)


internal/engine/compiler/engine.go
```diff
@@ -1368,8 +1334,54 @@ func compileWasmFunction(cmp compiler, ir *wazeroir.CompilationResult) (*code, e
...
+			case wazeroir.OperationKindUnreachable:
+				err = cmp.compileUnreachable()
+			case wazeroir.OperationKindMemorySize:
+				err = cmp.compileMemorySize()
+			case wazeroir.OperationKindMemoryGrow:
+				err = cmp.compileMemoryGrow()
+
+			case wazeroir.OperationKindI32WrapFromI64:
+				err = cmp.compileI32WrapFromI64()
+
```
evacchi, line 1348: spacing in this function is intended, to recall that there are "other ops" that are not handled here

internal/wazeroir/operations.go
```diff
@@ -1393,36 +1379,24 @@ func (OperationStore32) Kind() OperationKind {
...
-
-// Kind implements Operation.Kind.
-func (OperationMemorySize) Kind() OperationKind {
-	return OperationKindMemorySize
+func NewOperationMemorySize() OperationNullary {
+	return OperationNullary{kind: OperationKindMemorySize}
 }
 
-// OperationMemoryGrow implements Operation.
-type OperationMemoryGrow struct{ Alignment uint64 }
```
evacchi, line 1412: MemoryGrow is not really nullary but we discard the `Alignment` field everywhere

evacchi:
cancelled test runs because they are timing out 🤦

internal/engine/interpreter/interpreter.go
```diff
@@ -300,8 +300,19 @@ func (e *engine) lowerIR(ir *wazeroir.CompilationResult) (*code, error) {
 			op.sourcePC = ir.IROperationSourceOffsetsInWasmBinary[i]
 		}
 		switch o := original.(type) {
-		case wazeroir.OperationBuiltinFunctionCheckExitCode:
-		case wazeroir.OperationUnreachable:
+		case wazeroir.OperationNullary:
```
mathetake, line 303: do you want to proceed how this can be done in the union struct? In the end, you have to delete `wazeroir.OperationNullary` and replace it with the union struct, no?

mathetake (changes requested):
I would suggest to introduce the union struct in this PR, instead of soon-to-be-deleted `OperationNullary`.

internal/engine/compiler/engine.go
```diff
@@ -1368,8 +1334,54 @@ func compileWasmFunction(cmp compiler, ir *wazeroir.CompilationResult) (*code, e
 			err = cmp.compileV128Narrow(o)
 		case wazeroir.OperationV128ITruncSatFromF:
 			err = cmp.compileV128ITruncSatFromF(o)
-		case wazeroir.OperationBuiltinFunctionCheckExitCode:
-			err = cmp.compileBuiltinFunctionCheckExitCode()
+		case wazeroir.OperationNullary:
+			switch op.Kind() {
```
mathetake, line 1338: what I meant is like this

```suggestion
		case wazeroir.UnionOperation:
			switch op.Kind {
```

instead of define "new" implementation of old Operation interface.

internal/engine/interpreter/interpreter.go
evacchi: sure that is correct. The only reason I introduced `OperationNullary` is only to make this PR fully self-contained

mathetake: no need to make it self-contained, but I appreciate what you intended to do! let's just make a shortcut to the goal of nuking the old operation

internal/wazeroir/operations.go
```diff
@@ -948,19 +924,39 @@ func (b BranchTargetDrop) String() (ret string) {
...
+// only relevant when in context of its kind.
+type UnionOperation struct {
+	// OpKind determines how to interpret the other fields in this struct.
+	OpKind   OperationKind
+	B1, B2   byte
+	B3       bool
+	U1, U2   uint64
+	Us       []uint64
+	Rs       []*InclusiveRange
+	SourcePC uint64
```
mathetake, line 942: let's add this later as you suggested incrementally so that we can detect unnecessary portions! 
```suggestion
```

mathetake (approved):
good

internal/wazeroir/operations.go
mathetake: oh you've deleted interpreterOp?

mathetake:
we can change it later, so let's merge it

internal/wazeroir/operations.go
evacchi: yes! will do as I migrate more instructions

mathetake: 👯

internal/wazeroir/operations.go
```diff
@@ -948,19 +924,39 @@ func (b BranchTargetDrop) String() (ret string) {
...
+// Not all operations result in a UnionOperation, e.g. wazeroir.OperationI32ReinterpretFromF32, and some operations are
+// more complex than others, e.g. wazeroir.OperationBrTable.
 //
-// The engines are expected to exit the execution with wasmruntime.ErrRuntimeUnreachable error.
-type OperationUnreachable struct{}
+// Note: This is a form of union type as it can store fields needed for any operation. Hence, most fields are opaque and
+// only relevant when in context of its kind.
+type UnionOperation struct {
+	// OpKind determines how to interpret the other fields in this struct.
+	OpKind   OperationKind
```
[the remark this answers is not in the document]
evacchi: we can't, because there is already `Kind()`. In order to make migration smooth I am letting `UnionOperation` implement `Operation`. When no other `Operation` will be left, we can remove the interface and use the field directly, losing the need for the `Op` prefix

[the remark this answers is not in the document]
mathetake: oh ic make sense!

PR #1312, opened 2023-03-30
doc: note on using compilation cache in non-main package tests

mathetake:


config.go
```diff
@@ -117,6 +117,13 @@ type RuntimeConfig interface {
 	//	// Creates two runtimes while sharing compilation caches.
 	//	foo := wazero.NewRuntimeWithConfig(context.Background(), config)
 	// 	bar := wazero.NewRuntimeWithConfig(context.Background(), config)
+	//
+	// Note: the cached files contain the version of wazero which is obtained from go.mod of your application,
```
codefromthecrypt, line 121:
```suggestion
	// # Cache Key
	//
	// Cached files are keyed on the version of wazero. This is obtained from go.mod of your application,
```

codefromthecrypt (approved):
maybe cc the poor soul that discovered this issue ;)

PR #1315, opened 2023-03-30
asm/amd64: avoids allocations on forward jump resolution

mathetake:


mathetake:
need more code comments and unittests

internal/asm/amd64/impl.go
```diff
@@ -407,6 +411,36 @@ func (a *AssemblerImpl) InitializeNodesForEncoding() {
...
+					// Assign this as a jump destination.
+					nop.forwardJumpTarget = true
+					n.jumpTarget = nop
+
+					target = nop
+				} else {
+					target.forwardJumpTarget = true
+				}
+
+				tail := target
```
codefromthecrypt, line 433: comment the for please

codefromthecrypt (approved):
a lot more than I was expecting!

PR #1318, opened 2023-03-30
wazeroir: migrate unary operations to UnionOperation

evacchi:
- refactor: OperationCall, OperationGlobalGet, OperationGlobalSet
- refactor: Constant Operations

Refactor some unary operations to use `UnionOperation`. Again, we do this in waves to avoid creating a huge big-bang PR.

Unfortunately I have to do this mostly by hand to ensure that I am not breaking something, and to preserve comments and docs. At least I am double-checking it :D

Continues implementation of #1202

### DIFF
--- a/internal/engine/compiler/compiler_global_test.go
+++ b/internal/engine/compiler/compiler_global_test.go
@@ -140,7 +140,7 @@ func TestCompiler_compileGlobalSet(t *testing.T) {
 			}
 			env.stack()[loc.stackPointer] = valueToSet
 
-			var index uint32 = 1
+			const index = 1
 			op := wazeroir.NewOperationGlobalSet(index)
 			err = compiler.compileGlobalSet(op)
 			requireRuntimeLocationStackPointerEqual(t, 0, compiler)
@@ -188,7 +188,7 @@ func TestCompiler_compileGlobalSet_v128(t *testing.T) {
 	hi.valueType = runtimeValueTypeV128Hi
 	env.stack()[hi.stackPointer] = valueToSetHi
 
-	var index uint32 = 1
+	const index = 1
 	op := wazeroir.NewOperationGlobalSet(index)
 	err = compiler.compileGlobalSet(op)
 	requireRuntimeLocationStackPointerEqual(t, 0, compiler)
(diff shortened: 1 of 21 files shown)


internal/engine/compiler/compiler_global_test.go
```diff
@@ -140,7 +140,8 @@ func TestCompiler_compileGlobalSet(t *testing.T) {
 			}
 			env.stack()[loc.stackPointer] = valueToSet
 
-			op := wazeroir.OperationGlobalSet{Index: 1}
+			var index uint32 = 1
```
mathetake, line 143:
```suggestion
			const index = 1
```

internal/wazeroir/operations.go
```diff
@@ -943,7 +937,25 @@ type UnionOperation struct {
...
+	case OperationKindConstI32,
+		OperationKindConstI64:
+		return fmt.Sprintf("%s %#x", o.Kind(), o.U1)
+
+	case OperationKindConstF32:
+		return fmt.Sprintf("%s %f", o.Kind(), math.Float32frombits(uint32(o.U1)))
+	case OperationKindConstF64:
+		return fmt.Sprintf("%s %f", o.Kind(), math.Float64frombits(o.U1))
+	default:
+		return o.Kind().String()
```
mathetake, line 956: can we put this `panic("TODO")` so that we won't miss the implementation of String on all kinds?

Let's add this test later similar to https://github.com/tetratelabs/wazero/blob/main/internal/wazeroir/operations_test.go#L9-L14

internal/engine/interpreter/interpreter.go
```diff
@@ -277,14 +277,28 @@ func (e *engine) lowerIR(ir *wazeroir.CompilationResult) (*code, error) {
 	labelAddress := map[wazeroir.LabelID]uint64{}
 	onLabelAddressResolved := map[wazeroir.LabelID][]func(addr uint64){}
 	for i, original := range ops {
-		op := &wazeroir.UnionOperation{OpKind: original.Kind()}
+		var op *wazeroir.UnionOperation
```
mathetake, line 280: oh why do you have to use pointer?

mathetake:
could you make sure always link the issue #1202 in the description? Also, add the package name or more descriptive prefix in the PR titles instead of `refactor`.

internal/engine/interpreter/interpreter.go
mathetake: ah ok nvm

evacchi:
weird @mathetake I only saw some comments _after_ I hard pushed and merged 🤦

PR #1321, opened 2023-03-30
wasm: move ModuleInstance.Closed to the top

twilly:
ModuleInstance.Closed is an atomic variable meant to be loaded and swapped with sync/atomic. Closed, being a 64 bit integer, requires 64 bit alignment. The simplest way we can get alignment is to place these atomic fields at the top of their struct.

Closed can be moved to a more logical place once support for Go 1.18 is dropped and its type changed to atomic.Uint64.

### DIFF
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -348,13 +348,13 @@ const (
 	functionSize                     = 56
 
 	// Offsets for wasm.ModuleInstance.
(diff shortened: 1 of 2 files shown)


mathetake (approved):
Awesome. Thank you!

codefromthecrypt:
nice thanks! funny I thought we had more of these, but I think @mathetake's million cleanups must have buffed them out!

PR #1332, opened 2023-04-01
api: adds CallTo to avoid allocations

inkeliz:
Before that change each function call creates a new slice, which copies the result from wasm function. That causes a allocation on each call, since golang-compiler are not able to optimize such case.

That patch introduce a new function, CallTo, which makes possible to re-use an already existent slice. In some cases that makes possible to call multiple functions without any allocation.

Currently, such optimization only affects compiler/engine.go, and was not fully ported to interpreter/interpreter.go.

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -329,6 +329,13 @@ type Function interface {
 	// WithCloseOnContextDone on wazero.RuntimeConfig for detail. See examples in context_done_example_test.go for
 	// the end-to-end demonstrations of how these terminations can be performed.
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
+
+	// CallTo invokes the function with the given parameters and writes the
+	// results to the given slice. The slice must be large enough to hold all
+	// results.
+	//
+	// See Call for more details.
+	CallTo(ctx context.Context, out []uint64, params ...uint64) error
 }
 
 // GoModuleFunction is a Function implemented in Go instead of a wasm binary.
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -647,8 +647,8 @@ func (ce *callEngine) Definition() api.FunctionDefinition {
 	return ce.initialFn.def
 }
 
-// Call implements the same method as documented on wasm.ModuleEngine.
-func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uint64, err error) {
+// CallTo implements the same method as documented on wasm.ModuleEngine.
+func (ce *callEngine) CallTo(ctx context.Context, out []uint64, params ...uint64) (err error) {
 	m := ce.initialFn.moduleInstance
 	if ce.fn.parent.withEnsureTermination {
 		select {
@@ -656,7 +656,7 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 			// If the provided context is already done, close the call context
 			// and return the error.
 			m.CloseWithCtxErr(ctx)
-			return nil, m.FailIfClosed()
+			return m.FailIfClosed()
 		default:
 		}
 	}
@@ -665,7 +665,11 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 
 	paramCount := len(params)
 	if tp.ParamNumInUint64 != paramCount {
-		return nil, fmt.Errorf("expected %d params, but passed %d", ce.initialFn.funcType.ParamNumInUint64, paramCount)
+		return fmt.Errorf("expected %d params, but passed %d", ce.initialFn.funcType.ParamNumInUint64, paramCount)
+	}
+
+	if len(out) < tp.ResultNumInUint64 {
+		return fmt.Errorf("expected %d return values, but passed %d", tp.ResultNumInUint64, len(out))
 	}
 
 	// We ensure that this Call method never panics as
@@ -693,10 +697,18 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 	// returned a re-slice, the caller could accidentally or purposefully
 	// corrupt the stack of subsequent calls
 	if resultCount := tp.ResultNumInUint64; resultCount > 0 {
-		results = make([]uint64, resultCount)
-		copy(results, ce.stack[:resultCount])
+		copy(out, ce.stack[:resultCount])
 	}
 	return
+}
+
+func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uint64, err error) {
+	resultCount := ce.initialFn.funcType.ResultNumInUint64
+	if resultCount == 0 {
+		return nil, ce.CallTo(ctx, nil, params...)
+	}
+	ret := make([]uint64, resultCount)
+	return ret, ce.CallTo(ctx, ret, params...)
 }
 
 // initializeStack initializes callEngine.stack before entering native code.
--- a/internal/engine/interpreter/interpreter.go
+++ b/internal/engine/interpreter/interpreter.go
@@ -663,6 +663,19 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 	return ce.call(ctx, ce.compiled, params)
 }
 
+// CallTo implements the same method as documented on api.Function.
+func (ce *callEngine) CallTo(ctx context.Context, out []uint64, params ...uint64) (err error) {
+	ret, err := ce.call(ctx, ce.compiled, params)
+	if err != nil {
+		return err
+	}
+	if len(out) < len(ret) {
+		return fmt.Errorf("expected %d results, but got %d", len(out), len(ret))
+	}
+	copy(out, ret)
+	return nil
+}
+
 func (ce *callEngine) call(ctx context.Context, tf *function, params []uint64) (results []uint64, err error) {
 	m := ce.compiled.moduleInstance
 	if ce.compiled.parent.ensureTermination {
--- a/internal/wasm/store_test.go
+++ b/internal/wasm/store_test.go
@@ -480,6 +480,15 @@ func (ce *mockCallEngine) Call(_ context.Context, _ ...uint64) (results []uint64
 	return
 }
 
+// CallTo implements the same method as documented on api.Function.
+func (ce *mockCallEngine) CallTo(_ context.Context, out []uint64, _ ...uint64) (err error) {
+	if ce.callFailIndex >= 0 && ce.index == Index(ce.callFailIndex) {
+		err = errors.New("call failed")
+		return
+	}
+	return
+}
+
 func TestStore_getFunctionTypeID(t *testing.T) {
 	t.Run("too many functions", func(t *testing.T) {
 		s := newStore()


mathetake:
Thanks, I understand the motivation, but we have to be really careful adding a new API.
One quick question is I wonder if this is really the bottleneck of your application. I would appreciate it if you could give us the benchmark and profiling result. Thanks!

inkeliz:
Currently, I don't have any benchmarks to give. But in the simple benchmarks I've ran, the performance difference between the original implementation and my suggested improvements is hardly noticeable. But, I am worried that the additional allocation would increase the GC work, which may impact the performance.

Maybe CallTo method isn't a good idea. The "With_Something_" methods is already used by Wazero, so maybe we can create one "WithOutputBuffer([]byte)" method that would enable all results to be transferred to that buffer.

Also, the "input" have issues. My patch doesn't try to address that. Currently, the input is `params ...uint64` that is one "implicit slice". Calling as `Call(ctx, input)` that `x` will get allocated too. However, if I do something like: `paramBuf := []int{input}` and then `Call(ctx, paramBuf...)` it will not allocate. Maybe we could have one function that uses `param []uint64` and re-use the buffer.

ncruces:
If reducing these allocations is considered relevant, one other way to achieve it is to replace these lines:
https://github.com/tetratelabs/wazero/blob/84ea9efc143f20f5e2dc5feaf37401952c63ad75/internal/engine/compiler/engine.go#L703-L706

With (reusing the `params` slice for the result):
```go
	if resultCount := tp.ResultNumInUint64; resultCount > 0 {
		results = append(params[:0], ce.stack[:resultCount]...)
	}
```

Callers that don't care, are _probably_ using varargs, the slice gets created for them; this avoids them some (50%?) allocations "for free."
Callers that do care, can take care to keep a `[]uint64` to avoid these allocations entirely.

OTOH, this causes issues for callers that reuse the argument slice expecting it not to be modified (which is a reasonable expectation).

I couldn't really measure any appreciable (CPU) difference in benchmarks. But allocs/op do drop 60% on some benchmarks (with these changes).

inkeliz:
I'm not sure if `results = append(params[:0], ce.stack[:resultCount]...)` would be great. That mitigates the allocation, but not remove them. Also, variadic also causes allocations in some cases, and that might force to be one of them.

I'm using the following code:

```go
        wasmfn := make([]api.Function, 100) 
        for i := 0; i < 100; i++ {
		wasm,_, _ := initWebAssemblyModule(data, 256, _guestZIG)
		// ...
		wasmfn[i] = wasm.ExportedFunction("work")
	}

        res := make([][]uint64, 100)
	for i := range res {
		res[i] = make([]uint64, 1)
	}
	in := make([]uint64, 1)
        in[0] = 515249275
	
	for i := 0; i < runtime.NumCPU(); i++ {
		go func() {
			for j := range jobs {
				if err := wasmfn[j].CallTo(nil, res[j], in...); err != nil {
					return
				}
				// ...
```

I have 100 runtimes/modules and I'm calling them with the same input and storing each result to later process. In that case, `res` is re-used for each call to the same module, and I can them loop each result and do what I need to do. Also, I'm always calling with the same input (from ` in[0] = 515249275`). The ` wasmfn[j].CallTo(nil, res[j], in...)`  is `0 allocs/op`, instead of `2 allocs/op`.

So, doing some benchmarks, such as:

```go
    for i := 0; i < b.N; i++ {
		wait.Add(len(wasmfn))
		for i := range wasmfn {
			i := i
			jobs <- i
		}
		wait.Wait()
    }
```

That is calling 100 wasm-function per benhmark-loop.

Without any optimization that was `201 allocs/op`:

```go
	res, err :=wasmfn[j].Call(nil, 515249275)
	if err != nil {
		b.Error(err)
		return
	}
```

One allocation due to `515249275` instead of re-using the input (`in...`); Another because each `Call` instead of  `CallTo`.

Replacing the `Call` to `CallTo` and also `515249275` to `in...` results on 1 alloc/op. Previously, it was 201 allocs/op, due to wazero allocations.

I guess GoCompiler can optimize `in...`, so it drops o `1 allocs/op`. Then, using `CallTo` drops to 0. But, I'm almost sure that modifying `Call` to `results = append(params[:0], ce.stack[:resultCount]...)` will prevent such optimization. 🧐

---------

The performance still almost the same, but I guess the performance hit will be caused by GC. So, `CallTo` might put less pressure on GC.

inkeliz:
I think the issue is when the called function have more returns than input argument, which seems like to cause allocations. I didn't test it. That might be more noticeable when calling functions with zero arguments and one return value.

codefromthecrypt:
If optimizing, it would seem less a cognitive punch to use something like GoFunc/GoFunction so at least the style of invocation is known. Ack it doesn't have an error return but even if it had similarity it would feel more familiar even if also weird.

```golang
// GoFunction is an optimized form of GoModuleFunction which doesn't require
// the Module parameter. See GoFunc for an example.
//
// For example, this function does not need to use the importing module's
// memory or exported functions.
type GoFunction interface {
	Call(ctx context.Context, stack []uint64)
}
```

inkeliz:
> No that's not an issue, not really. You can easily pass in a slice that has length 0 but capacity 10; append can reuse that capacity.

You are right. I assumed I would call `Call(ctx)`, without `params`. Since the type is ...uint64 and not []uint64, I initially overlooked the possibility of passing in a 'empty' slice, but with capacity.  😓

ncruces:
I understand the carefulness of adding new API, but `callEngine.Call` is indeed responsible for a significant number of allocations.

This is with me reusing the params slice, so just a bunch of `[1]uint64` allocated for the (single) return value (the 120MB simply disappear with my change above):
```
(pprof) top 10
Showing nodes accounting for 1370.98MB, 91.26% of 1502.30MB total
Dropped 147 nodes (cum <= 7.51MB)
Showing top 10 nodes out of 97
  486.16MB 32.36% 32.36%   486.16MB 32.36%  github.com/tetratelabs/wazero/internal/wasm.NewMemoryInstance (inline)
  276.04MB 18.37% 50.74%   276.04MB 18.37%  github.com/tetratelabs/wazero/internal/engine/compiler.(*moduleEngine).newCallEngine
  170.17MB 11.33% 62.06%   170.17MB 11.33%  github.com/tetratelabs/wazero/internal/engine/compiler.(*engine).NewModuleEngine
  127.50MB  8.49% 70.55%   188.13MB 12.52%  github.com/tetratelabs/wazero/internal/engine/compiler.(*callEngine).Call
  110.64MB  7.36% 77.92%  1324.19MB 88.14%  github.com/maragudk/sqlite-benchmark.(*DB).ReadPost
   83.64MB  5.57% 83.48%   132.64MB  8.83%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnText (inline)
      41MB  2.73% 86.21%       41MB  2.73%  github.com/ncruces/go-sqlite3/driver.maybeTime
   27.12MB  1.81% 88.02%    27.12MB  1.81%  github.com/tetratelabs/wazero/internal/wasm.(*MemoryInstance).Grow
   24.89MB  1.66% 89.67%    24.89MB  1.66%  github.com/tetratelabs/wazero/internal/asm/amd64.(*nodePool).allocNode (inline)
   23.81MB  1.59% 91.26%    68.86MB  4.58%  github.com/tetratelabs/wazero/internal/engine/compiler.compileWasmFunction
```

If I we can coalesce around an acceptable API, I don't mind doing some/all of the legwork.

achille-roussel:
Maybe one way to take a first step is to add an interface in the experimental package that adds the CallTo method?

The Go program using wazero can then type assert api.Module on that interface to access the CallTo method to enable the optimization. 

This lets us experiment without committing to an exported interface. After getting experience with it we can consider graduating the API in a 1.x version of wazero so the type assert isn't needed anymore. 

If you think this can work feel free to go ahead and send a PR, I'll be happy to review 👍

codefromthecrypt:
@ncruces thanks for the help on this. A couple questions (and we are likely to cut a release tues fyi)

> it should return the appended-to stack, and for ergonomics, even if the return is empty. Otherwise simple usages of the API eventually “leak” the slice (it gets GCed) and have to create a new one.

IIUC the godoc in `Notes` for this would be something like "this returns the input slice if it was large enough to hold any results. Otherwise, this returns a newly allocated slice"

> If I we can coalesce around an acceptable API, I don't mind doing some/all of the legwork.

Since `api.Function` is already a interface and we don't support custom implementations of it, we can add a method and argue it is a compatible change.

Here's a stab, please wrap manually at 80 columns when polished. :)
```go
	// CallWithStack is an optimized variation of Call that saves memory allocations when the
	// stack slice is large enough to hold the maximum length of functions parameters or results.
	//
	// The caller adds parameters in order to the stack, and reads any results in order from the
	// result stack, except in error case. This means the caller must be aware of the count of
	// parameters and results.
	//
	// # Example
	// Add one
	// # Notes
	//
	//   - When the stack parameter is large enough to hold the max count of parameters or results, it is returned. Otherwise, a new slice is allocated and returned. This means the caller should always prefer using the result slice for subsequent calls.
where parameters are added in order
	//   - This is similar to GoFunc, except for using calling functions instead of implementing them. Moreover, this is used regardless of whether the callee is a host or wasm defined function.
	CallWithStack(ctx context.Context, stack []uint64) ([]uint64, error)
```

codefromthecrypt:
PS ideally Call can be reverse implemented with CallWithStack, though it isn't required to do this if it doesn't work out.

codefromthecrypt:
wires crossed on @achille-roussel's comment. I think that approach is compatible with my guidance too. the internal types implementing will be the same, and the signature discussion is the same.

Having CallWithStack in an interface `experimental.StackFunction` first only changes where it is first. Either way works for me: tentatively adding as experimental (expecting some sort of alternate signature to be more ideal than what we thought of) or adding directly (if we expect we've analyzed this enough).

codefromthecrypt:
one slight advantage of making this experimental is we can sneak it in without a minor bump. Main cost is that consumers using it can expect their code to break when we move it out which might not be very long from now. OTOH those who are asking about it have dealt with plenty of breaks so far ;) EOF

codefromthecrypt:
last point is that https://github.com/tetratelabs/wazero/issues/1364 is likely to land before our next release Tuesday which would make next version 1.1.0 not 1.0.2, that could negate the point I just made.

achille-roussel:
Not requiring a minor version bump was in the back of my mind too. 

If we can afford it I'd prefer if we made the API experimental at least for one minor version release (so targeting 1.2 since 1.1 is around the corner) to give us the time to learn from using the API before we commit to the 1.x compatibility promise.

Looking for your feedback @ncruces on feasibility, let us know what you think!

codefromthecrypt:
Let's keep Tuesday a patch (means this change -> experimental). thanks for thinking this through!

ncruces:
I'll respond to the rest later, but first of all: no urgency.
You guys do the release, and let's take time with API design.
Experimental is fine.

My post with numbers was just to show that yes, this can be a significant source of memory pressure, even if it's just a bunch of `[1]uint64` (my case), so people may want to optimize it, and we should give them a way too, even if it isn't the default.

ncruces:
OK, some further notes that might inform the way forward.

The benchmark I'm running is [`github.com/maragudk/sqlite-benchmark`](https://github.com/maragudk/sqlite-benchmark) adapted to use my driver rather than mattn's. Nothing special about this, but that's pretty much the point: it's not designed to stress this, rather it tries to simulate a realistic workload. I'm running my benchmark with `-benchtime=1000x` to get (more) consistent numbers, but there's non-determinism (goroutines and connection pools).  Onto numbers.

These changes do very little for CPU time, because the benchmark spends most of its time on syscalls. I'll investigate, but I suspect I/O:
```
Showing nodes accounting for 12.92s, 92.95% of 13.90s total
Dropped 168 nodes (cum <= 0.07s)
Showing top 10 nodes out of 212
     8.08s 58.13% 58.13%      8.08s 58.13%  syscall.syscall
     2.54s 18.27% 76.40%      2.54s 18.27%  syscall.syscall6
     0.78s  5.61% 82.01%      0.78s  5.61%  runtime.madvise
     0.42s  3.02% 85.04%      0.42s  3.02%  runtime.pthread_cond_wait
     0.28s  2.01% 87.05%      0.28s  2.01%  runtime._ExternalCode
     0.27s  1.94% 88.99%      0.27s  1.94%  syscall.syscall6X
     0.26s  1.87% 90.86%      0.26s  1.87%  runtime.pthread_cond_signal
     0.11s  0.79% 91.65%      0.11s  0.79%  runtime.memclrNoHeapPointers
     0.09s  0.65% 92.30%      0.09s  0.65%  runtime.kevent
     0.09s  0.65% 92.95%      0.09s  0.65%  runtime.usleep
```

There are two sides to a `api.Function.Call`: the input `params ...uint64` and the output `[]uint64`.
The both input/output seem to always escape to heap, I assume because `api.Function` is an interface.
We can try to optimize neither, either or both. My SQLite wrapper makes it "easy" to test because all calls go through a single wrapper function.

Neither:
```
Showing nodes accounting for 1474.46MB, 87.07% of 1693.40MB total
Dropped 149 nodes (cum <= 8.47MB)
Showing top 10 nodes out of 99
  455.23MB 26.88% 26.88%   455.23MB 26.88%  github.com/tetratelabs/wazero/internal/wasm.NewMemoryInstance (inline)
  289.08MB 17.07% 43.95%   289.08MB 17.07%  github.com/tetratelabs/wazero/internal/engine/compiler.(*moduleEngine).newCallEngine
  162.43MB  9.59% 53.55%   162.43MB  9.59%  github.com/tetratelabs/wazero/internal/engine/compiler.(*engine).NewModuleEngine
     120MB  7.09% 60.63%   169.50MB 10.01%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnRawText
  114.16MB  6.74% 67.37%  1505.74MB 88.92%  github.com/maragudk/sqlite-benchmark.(*DB).ReadPost
  104.50MB  6.17% 73.54%   166.44MB  9.83%  github.com/tetratelabs/wazero/internal/engine/compiler.(*callEngine).Call
   73.12MB  4.32% 77.86%   242.63MB 14.33%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnText (inline)
      72MB  4.25% 82.11%   101.50MB  5.99%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnType
      52MB  3.07% 85.19%       52MB  3.07%  github.com/ncruces/go-sqlite3/driver.maybeTime
   31.93MB  1.89% 87.07%    31.93MB  1.89%  github.com/tetratelabs/wazero/internal/wasm.(*MemoryInstance).Grow
```

So the interesting thing about this is that varargs slices escape on the caller, not the callee. I have a [single function](https://github.com/ncruces/go-sqlite3/blob/v0.5.3/module.go#L199-L207) that performs _all_ `api.Function.Call`s but that is itself varargs so the allocations don't pop up there.

The more popular (for this benchmark) callers of that function are `ColumnRawText` and `ColumnType`, so they show up quite high on the list.

If optimize input params allocations, which I did with this [2-line change](https://github.com/ncruces/go-sqlite3/commit/005b15610a2dea884660db59bd864325328a2cbe#diff-c95a9381cb3d755657cf984eba630f599695529c52a3b407d902d38bcea81f8aR203-R204), we get:
```
Showing nodes accounting for 1456.97MB, 92.02% of 1583.26MB total
Dropped 131 nodes (cum <= 7.92MB)
Showing top 10 nodes out of 98
  529.20MB 33.42% 33.42%   529.20MB 33.42%  github.com/tetratelabs/wazero/internal/wasm.NewMemoryInstance (inline)
  324.69MB 20.51% 53.93%   324.69MB 20.51%  github.com/tetratelabs/wazero/internal/engine/compiler.(*moduleEngine).newCallEngine
  174.59MB 11.03% 64.96%   174.59MB 11.03%  github.com/tetratelabs/wazero/internal/engine/compiler.(*engine).NewModuleEngine
  128.50MB  8.12% 73.08%   182.50MB 11.53%  github.com/tetratelabs/wazero/internal/engine/compiler.(*callEngine).Call
  110.63MB  6.99% 80.06%  1412.18MB 89.19%  github.com/maragudk/sqlite-benchmark.(*DB).ReadPost
   66.61MB  4.21% 84.27%   114.62MB  7.24%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnText (inline)
      48MB  3.03% 87.30%       48MB  3.03%  github.com/ncruces/go-sqlite3/driver.maybeTime
   26.99MB  1.70% 89.01%    26.99MB  1.70%  github.com/tetratelabs/wazero/internal/wasm.(*MemoryInstance).Grow
   25.50MB  1.61% 90.62%    25.50MB  1.61%  github.com/tetratelabs/wazero/internal/asm/amd64.(*nodePool).allocNode (inline)
   22.25MB  1.41% 92.02%    70.74MB  4.47%  github.com/tetratelabs/wazero/internal/engine/compiler.compileWasmFunction
```

Those two instantly drop out of the list, `callEngine.Call` stays because it's allocating the return slice. This can be achieved _now_ with no API change. My case was easy, for others it may be harder.

Next, if I use the [above trick](https://github.com/tetratelabs/wazero/pull/1332#issuecomment-1502124325) (reuse the input slice for the output slice), we get:
```
(pprof) top
Showing nodes accounting for 1264.24MB, 90.87% of 1391.21MB total
Dropped 145 nodes (cum <= 6.96MB)
Showing top 10 nodes out of 102
  494.90MB 35.57% 35.57%   494.90MB 35.57%  github.com/tetratelabs/wazero/internal/wasm.NewMemoryInstance (inline)
  297.60MB 21.39% 56.96%   297.60MB 21.39%  github.com/tetratelabs/wazero/internal/engine/compiler.(*moduleEngine).newCallEngine
  159.67MB 11.48% 68.44%   159.67MB 11.48%  github.com/tetratelabs/wazero/internal/engine/compiler.(*engine).NewModuleEngine
  111.63MB  8.02% 76.47%  1214.02MB 87.26%  github.com/maragudk/sqlite-benchmark.(*DB).ReadPost
   64.62MB  4.65% 81.11%    64.62MB  4.65%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnText (inline)
   48.50MB  3.49% 84.60%    48.50MB  3.49%  github.com/ncruces/go-sqlite3/driver.maybeTime
   27.79MB  2.00% 86.59%    27.79MB  2.00%  github.com/tetratelabs/wazero/internal/wasm.(*MemoryInstance).Grow
   24.89MB  1.79% 88.38%    24.89MB  1.79%  github.com/tetratelabs/wazero/internal/asm/amd64.(*nodePool).allocNode (inline)
   18.72MB  1.35% 89.73%    18.72MB  1.35%  github.com/tetratelabs/wazero/internal/wasm.(*ModuleInstance).applyData
   15.91MB  1.14% 90.87%    62.01MB  4.46%  github.com/tetratelabs/wazero/internal/engine/compiler.compileWasmFunction
```

Now `callEngine.Call` drops out as well. Total heap allocations went from `1693.40MB` to `1391.21MB` which is a significant change.

Finally, there's the possibility of optimizing the return value but not the parameters. This means [above trick](https://github.com/tetratelabs/wazero/pull/1332#issuecomment-1502124325), but calling with varargs and no cacheing. Not a huge fan of this (because changing an input slice is surprising) but maybe it's worth considering since it likely won't break _anyone_ and is a "free" optimization for users.

```
Showing nodes accounting for 1442.61MB, 87.66% of 1645.68MB total
Dropped 146 nodes (cum <= 8.23MB)
Showing top 10 nodes out of 94
  492.88MB 29.95% 29.95%   492.88MB 29.95%  github.com/tetratelabs/wazero/internal/wasm.NewMemoryInstance (inline)
  312.14MB 18.97% 48.92%   312.14MB 18.97%  github.com/tetratelabs/wazero/internal/engine/compiler.(*moduleEngine).newCallEngine
  156.36MB  9.50% 58.42%   156.36MB  9.50%  github.com/tetratelabs/wazero/internal/engine/compiler.(*engine).NewModuleEngine
  122.65MB  7.45% 65.87%  1445.59MB 87.84%  github.com/maragudk/sqlite-benchmark.(*DB).ReadPost
   99.50MB  6.05% 71.92%    99.50MB  6.05%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnRawText
      78MB  4.74% 76.66%       78MB  4.74%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnType
   75.63MB  4.60% 81.25%   175.13MB 10.64%  github.com/ncruces/go-sqlite3.(*Stmt).ColumnText (inline)
      52MB  3.16% 84.41%       52MB  3.16%  github.com/ncruces/go-sqlite3/driver.maybeTime
   27.25MB  1.66% 86.07%    27.25MB  1.66%  github.com/tetratelabs/wazero/internal/wasm.(*MemoryInstance).Grow
   26.19MB  1.59% 87.66%    76.56MB  4.65%  github.com/tetratelabs/wazero/internal/engine/compiler.compileWasmFunction
```

As expected `callEngine.Call` is out as well. The reduction is not as significant as optimizing params, because functions tend to have more params than returns (this C API as a single return value for 99% of functions, only `free` returns void). Still somewhat significant.

codefromthecrypt:
@ncruces thanks for the excellent analysis!

Here are some summary takeaways, important for use cases that call into the same module many times. 

These changes might not do a lot for wasm that only call in via entry points, e.g. WASI's `_start` or http-wasm's `handle_request`. However, FFI interactions which call back into the same module a lot, like sqlite or maybe proxy-wasm or @jerbob92's PDF utilities may end up benefiting.

I'll rule out changing Function from an interface to a struct as I think that ship has sailed. It would be hard to do something like this without adding circular deps as well.

## Changes with no API impact?

Users can: copy in parameters to a reused slice to prevent varargs from escaping.

For example:
```diff
-	r, err := fn.Call(m.ctx, params...)
+	m.arg = append(m.arg[:0], params...)
+	r, err := fn.Call(m.ctx, m.arg...)
```

wazero can: re-use the params slice (from varargs) for return values (internal to callEngine)

https://github.com/tetratelabs/wazero/pull/1332#issuecomment-1502124325

## How to we go this route?

If we went this route, this is mostly documentation revision. For example, in a new heading on api.Function, describe slice nuance, including the trick to avoid allocations on input and also to beware that the input slice will be re-used for results (usually only 1 value).

## Should we go this route?

I agree that it is unlikely anyone will expect to re-use values inside a varargs slice, except it could already be optimized. For example, if someone was passing constant input parameters via a pre-allocated slice, this would corrupt their calls. For example, `var fnFooParams = []uint64{1, 2, 3}`. I'm not saying this is normal, but it would make an uberbug I think.

Personally, I feel this is super tempting, but I also don't think a new API `CallWithStack` method similar to our `GoFunc` is a bad thing. It would be more a discussion if we didn't already have a `GoFunc`. OTOH the trick to prevent escapes (return the stack to the caller), is going to need to be documented. However, documenting that is no more tricky than documenting how to prevent varargs from escaping (e.g. re-using a slice and copy-ing into that). Finally, the nice thing about adding a new method is `CallWithStack` is already advanced, so it contains the advanced stuff in on godoc.

## What do you think?

I'm giving a perspective not porting anything like sqlite, pdfium, pdfx etc. Mainly, I'm thinking about api shape in general and maintenance both of code and docs, as well potential Uber-bombs. Take my opinion as just an opinion.

ncruces:
My opinion matches this sentence exactly: _“Personally, I feel this is super tempting”_ … but deep down, I know it's wrong.

As long as it's possible, I'm relatively agnostic on the mechanism, and can totally get behind https://github.com/tetratelabs/wazero/pull/1332#issuecomment-1507793809.

The only thing I'd add, is that you may want codify the notion that some wazero APIs are not meant for external implementation and are fair game to extend in this way. Given the way wazero is structured, I think it makes sense.

One way to do that is in documentation. The other way is to add an unexpected method to those interfaces (breaking compatibility right now for all of them, but not carrying the problem forward).

codefromthecrypt:
A different PR that helps us solidify the notion that most types are not for external implementation would be a fantastic add imho. I don't know state of the art here.. I suspect we aren't 100pct covered with docs, and some might be surprised by cast errors meanwhile.

codefromthecrypt:
obviated by #1407

PR #1334, opened 2023-04-02
wazeroir: migrate vector, table, branch and all other remaining ops to compact repr

evacchi:
Follow up to #1202. 

Implements all the missing Operations. It doesn't try to be clever because we can always improve at the next iteration.

The one place where I _had_ to be a little more inventive is branches, because we did not have a `BranchTargetDrop` equivalent; instead I am putting `LabelID`s in the `Us` slice and `*InclusiveRange`s in `Rs` (how it is meant to be).

We no longer need a Label{} struct (only used in test for now, but can be removed later)

~Implements mostly v128 ops + a bunch of others so far.~

~This is a draft to allow @mathetake to take an early look so that we can catch any issues earlier. There are only 11 more ops to go :)~

~Left some comments inline for doubts I had.~

~I am leaving branches/labels for last because migrating those is going to be less mechanical.~


internal/engine/interpreter/interpreter.go
```diff
@@ -459,143 +511,7 @@ func (e *engine) lowerIR(ir *wazeroir.CompilationResult) (*code, error) {
...
-			op.B1 = o.Shape
-		case wazeroir.OperationV128Div:
-			op.B1 = o.Shape
-		case wazeroir.OperationV128Neg:
-			op.B1 = o.Shape
-		case wazeroir.OperationV128Sqrt:
-			op.B1 = o.Shape
-		case wazeroir.OperationV128Abs:
-			op.B1 = o.Shape
-		case wazeroir.OperationV128Popcnt:
```
evacchi, line 550: apparently Shape is actually being ignored everywhere for this Op. Is this correct? (I assume so, just double-checking)

internal/engine/compiler/impl_vec_amd64.go
```diff
@@ -490,7 +510,9 @@ func (c *amd64Compiler) compileV128Shuffle(o wazeroir.OperationV128Shuffle) erro
 	}
 
 	consts := [32]byte{}
-	for i, lane := range o.Lanes {
+	lanes := o.Us
+	for i, unsignedLane := range lanes {
+		lane := byte(unsignedLane)
```
evacchi, line 515: is there a better way to do this?

internal/wazeroir/compiler.go
```diff
@@ -2019,889 +2019,890 @@ operatorSwitch:
...
-				OperationV128Swizzle{},
+				NewOperationV128Swizzle(),
 			)
 		case wasm.OpcodeVecV128i8x16Shuffle:
 			c.pc++
-			op := OperationV128Shuffle{}
-			copy(op.Lanes[:], c.body[c.pc:c.pc+16])
+			lanes := [16]byte{}
+			copy(lanes[:], c.body[c.pc:c.pc+16])
+			op := NewOperationV128Shuffle(lanes)
```
evacchi, line 2140: is this initialization done right?

internal/wazeroir/operations.go
```diff
@@ -2140,336 +2068,205 @@ const (
...
-
-// Kind implements Operation.Kind.
-//
-// This corresponds to wasm.OpcodeVecV128i8x16ShuffleName.
-func (OperationV128Shuffle) Kind() OperationKind {
-	return OperationKindV128Shuffle
+// NewOperationV128Shuffle is a constructor for UnionOperation with Kind OperationKindV128Shuffle.
+func NewOperationV128Shuffle(lanes [16]byte) UnionOperation {
+	op := UnionOperation{OpKind: OperationKindV128Shuffle}
+	op.Us = make([]uint64, 16)
```
mathetake, line 2165: Unnecessary copy and allocation here. Let's just make this function accept []uint64 in the first place.
```suggestion
func NewOperationV128Shuffle(lanes []uint64) UnionOperation {
	op := UnionOperation{OpKind: OperationKindV128Shuffle, Us: lanes}
```

internal/wazeroir/compiler.go
mathetake: left a comment in the impl of NewOperationV128Shuffle.

internal/wazeroir/operations.go
```diff
@@ -1008,6 +958,70 @@ func (o UnionOperation) String() string {
...
+		OperationKindV128FloatPromote,
+		OperationKindV128FloatDemote,
+
+		OperationKindV128FConvertFromI,
+		OperationKindV128Dot,
+		OperationKindV128Narrow:
+		return o.Kind().String()
+
+	case OperationKindV128ITruncSatFromF:
+		// String implements fmt.Stringer.
```
mathetake, line 1018: unneeded comment.

internal/wazeroir/compiler.go
```diff
@@ -2019,889 +2019,890 @@ operatorSwitch:
...
 			)
 		case wasm.OpcodeVecI8x16Abs:
 			c.emit(
-				OperationV128Abs{Shape: ShapeI8x16},
+				NewOperationV128Abs(ShapeI8x16),
 			)
 		case wasm.OpcodeVecI8x16Popcnt:
 			c.emit(
-				OperationV128Popcnt{},
+				NewOperationV128Popcnt(0),
```
mathetake, line 2589:
```suggestion
				NewOperationV128Popcnt(ShapeI8x16),
```

internal/engine/compiler/impl_vec_amd64.go
mathetake: I think it's ok for now

internal/wazeroir/compiler.go
evacchi: if we don't do a defensive copy here we're going to share the subrange in the slice (?). Is this ok? (I assume so because we I don't think we make changes in-place, but...)

mathetake: no in any way you end up copying. please try my suggestion in the comment.

mathetake: basically, you are copying twice in your current code.

evacchi:
ran the Zig test cases on Windows, all passed.
> 1980 passed; 143 skipped; 0 failed.

evacchi:
(another minor update, Zig+Windows still works locally)

internal/engine/compiler/compiler_controlflow_test.go
```diff
@@ -310,120 +310,128 @@ func TestCompiler_compileBrTable(t *testing.T) {
...
 
-	getBranchTargetDropFromFrameID := func(frameid uint32) *wazeroir.BranchTargetDrop {
-		return &wazeroir.BranchTargetDrop{
-			Target: wazeroir.Label{FrameID: frameid, Kind: wazeroir.LabelKindHeader},
-		}
+	getBranchLabelFromFrameID := func(frameid uint32) uint64 {
+		return uint64(wazeroir.Label{FrameID: frameid, Kind: wazeroir.LabelKindHeader}.ID())
 	}
 
+	var defaultedRange []*wazeroir.InclusiveRange = nil //[]*wazeroir.InclusiveRange{{ /* at pos 0 there is always the empty range */ }}
```
mathetake, line 317:
```suggestion
	var defaultedRange []*wazeroir.InclusiveRange
```

internal/wazeroir/operations.go
```diff
@@ -819,27 +746,32 @@ func (l LabelID) FrameID() int {
 
 // ID returns the LabelID for this Label.
 func (l Label) ID() (id LabelID) {
-	id = LabelID(l.Kind) | LabelID(l.FrameID)<<32
+	id = NewLabelID(l.Kind, l.FrameID)
 	return
 }
 
+func NewLabelID(kind LabelKind, frameID uint32) LabelID {
```
mathetake, line 753: Always keep in mind to check if all the exported symbols really do need to be exported, and if so, make godoc for all of them.

internal/wazeroir/operations.go
```diff
@@ -819,27 +746,32 @@ func (l LabelID) FrameID() int {
...
+	return LabelID(kind) | LabelID(frameID)<<32
+}
+
 // String implements fmt.Stringer.
-func (l Label) String() (ret string) {
-	switch l.Kind {
+func (lid LabelID) String() (ret string) {
```
mathetake, line 758: image

```suggestion
func (l LabelID) String() (ret string) {
```

internal/wazeroir/operations.go
mathetake: this is very import for any Go project, we take this seriously.

internal/wazeroir/operations.go
```diff
@@ -798,7 +725,7 @@ func NewOperationBuiltinFunctionCheckExitCode() UnionOperation {
 }
 
 // Label is the label of each block in wazeroir where "block" consists of multiple operations,
-// and must end with branching operations (e.g. OperationBr or OperationBrIf).
+// and must end with branching operations (e.g. NewOperationBr or NewOperationBrIf).
 type Label struct {
```
mathetake, line 729: I believe this `Label` is no longer needed to be exported

internal/wazeroir/compiler.go
```diff
@@ -794,29 +790,33 @@ operatorSwitch:
...
+			NewOperationBrTable(
+				append([]uint64{uint64(defaultTargetID)}, targetLabels...),
+				append([]*InclusiveRange{defaultTargetDrop}, targetDrops...),
+			),
+			//NewOperationBrTable{
+			//	Targets: targets,
+			//	Default: &BranchTargetDrop{
+			//		ToDrop: defaultTargetDrop, Target: defaultTargetID,
+			//	},
+			//},
```
mathetake, line 806: leftover.

internal/wazeroir/operations.go
```diff
@@ -1027,82 +1084,55 @@ func NewOperationUnreachable() UnionOperation {
...
-}
-
-// String implements fmt.Stringer.
-func (o OperationBrTable) String() string {
-	targets := make([]string, len(o.Targets))
-	for i, t := range o.Targets {
-		targets[i] = t.String()
+//
+// if targetRanges is nil, it is padded to the length of targetLabels
+func NewOperationBrTable(targetLabels []uint64, targetRanges []*InclusiveRange) UnionOperation {
```
mathetake, line 1127:
```suggestion
func NewOperationBrTable(defalultLabel Label, targetLabels []Labels, defaultRange *InclusiveRange, targetRanges []*InclusiveRange) UnionOperation {
```

Try this change, and avoid the allocation with `append` at the call site of NewOperationBrTable.

internal/wazeroir/compiler.go
```diff
@@ -794,29 +790,33 @@ operatorSwitch:
...
 		c.emit(
-			OperationBrTable{
-				Targets: targets,
-				Default: &BranchTargetDrop{
-					ToDrop: defaultTargetDrop, Target: defaultTarget,
-				},
-			},
+			NewOperationBrTable(
+				append([]uint64{uint64(defaultTargetID)}, targetLabels...),
+				append([]*InclusiveRange{defaultTargetDrop}, targetDrops...),
```
mathetake, line 799: see the comment on NewOperationBrTable. you should be able to avoid allocation by not using append here.

internal/wazeroir/compiler.go
```diff
@@ -731,20 +727,18 @@ operatorSwitch:
...
+		continuationLabel := NewLabelID(LabelKindHeader, c.nextID())
+		c.result.LabelCallers[continuationLabel]++
 		c.emit(
-			OperationBrIf{
-				Then: BranchTargetDrop{ToDrop: drop, Target: target},
-				Else: continuationLabel.asBranchTargetDrop(),
-			},
+			NewOperationBrIf(
+				/* then */ BranchTargetDrop{ToDrop: drop, Target: targetID},
+				/* else */ continuationLabel.asBranchTargetDrop(),
```
mathetake, line 738: no need to name params.
```suggestion
				BranchTargetDrop{ToDrop: drop, Target: targetID},
				continuationLabel.asBranchTargetDrop(),
```

internal/wazeroir/compiler.go
```diff
@@ -526,20 +524,18 @@ operatorSwitch:
...
-			OperationBrIf{
-				Then: thenLabel.asBranchTargetDrop(),
-				Else: elseLabel.asBranchTargetDrop(),
-			},
-			OperationLabel{
-				Label: thenLabel,
-			},
+			NewOperationBrIf(
+				/* then */ thenLabel.asBranchTargetDrop(),
+				/* else */ elseLabel.asBranchTargetDrop(),
```
mathetake, line 536: same.

mathetake (changes requested):

internal/engine/compiler/impl_amd64.go
```diff
@@ -501,7 +501,10 @@ func (c *amd64Compiler) compileBrIf(o wazeroir.OperationBrIf) error {
 	}
 
 	// Make sure that the next coming label is the else jump target.
-	thenTarget, elseTarget := o.Then, o.Else
+	thenTarget := wazeroir.LabelID(o.Us[0])
+	thenToDrop := o.Rs[0]
+	elseTarget := wazeroir.LabelID(o.Us[1])
+	// elseToDrop := o.Rs[1]
```
mathetake, line 507:
```suggestion
	// elseToDrop := o.Rs[1]
```

internal/engine/compiler/engine.go
```diff
@@ -1375,12 +1248,134 @@ func compileWasmFunction(cmp compiler, ir *wazeroir.CompilationResult) (*code, e
 
 			// Drop..Init
 			//
```
mathetake, line 1250: can we delete this kind of comments?

internal/wazeroir/compiler.go
```diff
@@ -3015,13 +3018,14 @@ func (c *compiler) stackPush(ts UnsignedType) {
 func (c *compiler) emit(ops ...Operation) {
```
mathetake, line 3018: I guess this can accept UnionOperation now
```suggestion
func (c *compiler) emit(ops ...UnionOperation) {
```

internal/wazeroir/compiler.go
```diff
@@ -3015,13 +3018,14 @@ func (c *compiler) stackPush(ts UnsignedType) {
...
-			switch o := op.(type) {
-			case OperationDrop:
+			switch op.Kind() {
+			case OperationKindDrop:
 				// If the drop range is nil,
 				// we could remove such operations.
 				// That happens when drop operation is unnecessary.
 				// i.e. when there's no need to adjust stack before jmp.
-				if o.Depth == nil {
+				o := op.(UnionOperation)
```
mathetake, line 3027: then this conversion won't be necessary.

internal/wazeroir/operations.go
mathetake, line 1: now you should be able to delete `Operation` interface

mathetake (commented):
Left some comments to be addressed, but over all this is looking good. Thank you for your hard work 👍

mathetake:
> ran the Zig test cases on Windows, all passed.
> 
> > 1980 passed; 143 skipped; 0 failed.

thank you for checking!

internal/wazeroir/operations.go
```diff
@@ -1027,82 +1084,55 @@ func NewOperationUnreachable() UnionOperation {
...
+	if targetRanges == nil {
+		targetRanges = make([]*InclusiveRange, len(targetLabels))
+	}
```
mathetake, line 1130: Don't allocate unnecessary stuff.

internal/wazeroir/operations.go
evacchi: good point. I think eventually we should add a lint check if possible 🤔

internal/wazeroir/operations.go
evacchi: -> `label`, and updated a few tests that used that to use `NewLabelID()` instead. There are still a few usages of `label` but they should all be tests now.

internal/wazeroir/operations.go
```diff
@@ -770,8 +766,8 @@ func (lid LabelID) String() (ret string) {
 	return
 }
 
-func (l LabelID) IsReturnTarget() bool {
-	return l.Kind() == LabelKindReturn
+func (lid LabelID) IsReturnTarget() bool {
```
mathetake, line 769: we prefer single character for the receiver so shall we keep `l`?
```suggestion
func (lid LabelID) IsReturnTarget() bool {
```

internal/wazeroir/operations.go
```diff
@@ -1027,82 +1079,52 @@ func NewOperationUnreachable() UnionOperation {
...
-	Then, Else BranchTargetDrop
-}
-
-// String implements fmt.Stringer.
-func (o OperationBrIf) String() string { return fmt.Sprintf("%s %s, %s", o.Kind(), o.Then, o.Else) }
-
-// Kind implements Operation.Kind
-func (OperationBrIf) Kind() OperationKind {
-	return OperationKindBrIf
+// The engines are expected to pop a value and branch into OperationBrIf.Then (Us[0]) label if the value equals 1.
```
mathetake, line 1098: can we unwrap `()`?
```suggestion
// The engines are expected to pop a value and branch into OperationBrIf.Then Us[0] label if the value equals 1.
```

internal/wazeroir/operations.go
```diff
@@ -1027,82 +1079,52 @@ func NewOperationUnreachable() UnionOperation {
...
+// Otherwise, the code branches into OperationBrIf.Else (Us[1]) label.
```
mathetake, line 1099: same

mathetake:
```suggestion
// Otherwise, the code branches into Us[1] label.
```

internal/wazeroir/format.go
```diff
@@ -6,7 +6,7 @@ import (
 
 const EntrypointLabel = ".entrypoint"
 
-func Format(ops []Operation) string {
+func Format(ops []UnionOperation) string {
 	buf := bytes.NewBuffer(nil)
 
 	_, _ = buf.WriteString(EntrypointLabel + "\n")
```
mathetake, line 12: the for loop below results in copy now as `ops` is the slice of values, not pointers. could you make it not copy? (refer somewhere else in the code base where we do the same thing)

internal/engine/interpreter/interpreter.go
```diff
@@ -277,85 +277,15 @@ func (e *engine) lowerIR(ir *wazeroir.CompilationResult) (*code, error) {
...
 	onLabelAddressResolved := map[wazeroir.LabelID][]func(addr uint64){}
 	for i, original := range ops {
-		var op *wazeroir.UnionOperation
-		if o, ok := original.(wazeroir.UnionOperation); ok {
-			op = &o
-		} else {
-			op = &wazeroir.UnionOperation{OpKind: original.Kind()}
-		}
+		// copy to op to avoid mutation in-place
+		op := original
```
mathetake, line 281: do we really need this?

internal/engine/compiler/engine.go
mathetake, line 1: same comment on `for i, op := range ir.Operations {` to avoid unnecessary copy.

internal/engine/compiler/compiler_controlflow_test.go
```diff
@@ -310,120 +310,109 @@ func TestCompiler_compileBrTable(t *testing.T) {
...
-	getBranchTargetDropFromFrameID := func(frameid uint32) *wazeroir.BranchTargetDrop {
-		return &wazeroir.BranchTargetDrop{
-			Target: wazeroir.Label{FrameID: frameid, Kind: wazeroir.LabelKindHeader},
-		}
+	getBranchLabelFromFrameID := func(frameid uint32) uint64 {
+		return uint64(wazeroir.NewLabelID(wazeroir.LabelKindHeader, frameid))
+	}
+
+	newOperationBrTableWithDefaultRanges := func(us ...uint64) wazeroir.UnionOperation {
+		targetRanges := make([]*wazeroir.InclusiveRange, len(us))
```
mathetake, line 318: wonder if we still need this?

internal/wazeroir/operations_test.go
```diff
@@ -16,68 +16,15 @@ func TestOperationKind_String(t *testing.T) {
...
-		OperationKindFloor,
-		OperationKindTrunc,
-		OperationKindNearest,
-		OperationKindSqrt,
-		OperationKindMin,
-		OperationKindMax,
-		OperationKindCopysign,
-		OperationKindBuiltinFunctionCheckExitCode,
-	} {
+	for k := OperationKind(0); k < operationKindEnd; k++ {
```
mathetake, line 19: 🙌

internal/wazeroir/operations.go
```diff
@@ -431,19 +424,19 @@ func (o OperationKind) String() (ret string) {
...
-	// OperationKindBr is the OpKind for OperationBr.
+	// OperationKindBr is the OpKind for NewOperationBr.
 	OperationKindBr
-	// OperationKindBrIf is the OpKind for OperationBrIf.
+	// OperationKindBrIf is the OpKind for NewOperationBrIf.
 	OperationKindBrIf
-	// OperationKindBrTable is the OpKind for OperationBrTable.
+	// OperationKindBrTable is the OpKind for NewOperationBrTable.
 	OperationKindBrTable
 	// OperationKindCall is the OpKind for OperationCall.
```
mathetake, line 435: `OperationCall` doesn't exist anymore? (same comment on similar ops)

internal/wazeroir/operations.go
```diff
@@ -797,49 +717,57 @@ func NewOperationBuiltinFunctionCheckExitCode() UnionOperation {
 	return UnionOperation{OpKind: OperationKindBuiltinFunctionCheckExitCode}
 }
 
-// Label is the label of each block in wazeroir where "block" consists of multiple operations,
-// and must end with branching operations (e.g. OperationBr or OperationBrIf).
-type Label struct {
+// label is the label of each block in wazeroir where "block" consists of multiple operations,
+// and must end with branching operations (e.g. NewOperationBr or NewOperationBrIf).
+type label struct {
```
mathetake, line 722: curious if we need this type in the first place. Looks like we could remove this and name `LabelID` to `Label`.

internal/wazeroir/operations.go
```diff
@@ -1663,52 +1676,39 @@ func NewOperationI32WrapFromI64() UnionOperation {
...
+	return UnionOperation{
+		OpKind: OperationKindITruncFromF,
+		B1:     byte(inputType),
+		B2:     byte(outputType),
+		B3:     nonTrapping,
+	}
 }
 
-// OperationFConvertFromI implements Operation.
+// NewOperationFConvertFromI is a constructor for UnionOperation with Kind OperationKindFConvertFromI.
```
mathetake, line 1698: `Kind` is redundant I guess? (same goes for all the comments.

```suggestion
// NewOperationFConvertFromI is a constructor for UnionOperation with OperationKindFConvertFromI.
```

mathetake:
almost there and this is exciting. I will review tomorrow morning here in japan, and will merge it!

internal/engine/interpreter/interpreter.go
evacchi: If I remove that, several tests start to fail (e.g. `TestInterpreter_ModuleEngine_Call_HostFn`). Maybe there is a better way?

internal/wazeroir/format.go
evacchi: I think you mean `for i := range ops { op:=&ops[i] }` right?

internal/wazeroir/operations.go
evacchi: oh maybe it reads that way, the meaning was "the `Kind` field is `OperationKind<blah>`. I can remove it if it feels redundant.

internal/engine/compiler/engine.go
evacchi: I tried to change that but that requires that all `compileX(o)` accept a `*UnionOperation`. Are we ok with that change?

internal/wazeroir/operations.go
evacchi: removed :)

internal/engine/compiler/compiler_controlflow_test.go
evacchi: ps: I wouldn't worry too much about this because it's a test; point is, either we allow to pass a nil slice here or we'll have to fix where this is handled (probably at the interpreter site and the compilation site)

evacchi: heh I think I managed to fix it :)

internal/engine/interpreter/interpreter.go
evacchi: instead of doing a copy I am now doing `op := &ops[i]`; this works but I wonder if the mutation at `op.SourcePC =...` is ok...

evacchi:
today I am away from my Windows machine, I'll check tomorrow

achille-roussel:
Does the use of pointers not force the compiler to put these values in the heap?

evacchi:
fair question, I'll leave it to @mathetake's judgement

mathetake:
That's bad and comes with heap. Will revert the commit.

internal/engine/compiler/engine.go
mathetake: yes.

mathetake: see https://github.com/tetratelabs/wazero/pull/1334/commits/01644577656aab385a5927ad998ecc8c5668ccff if you are curious about how I avoided it.

internal/engine/compiler/impl_amd64.go
```diff
@@ -410,20 +410,20 @@ func (c *amd64Compiler) compileGlobalSet(o wazeroir.UnionOperation) error {
...
 }
 
 // branchInto adds instruction necessary to jump into the given branch target.
 func (c *amd64Compiler) branchInto(target wazeroir.Label) error {
 	if target.IsReturnTarget() {
 		return c.compileReturnFunction()
 	} else {
-		labelID := target.ID()
-		if c.ir.LabelCallers[labelID] > 1 {
+		label := target
```
mathetake, line 425: seems unnecessary

mathetake (approved):
Thank you so much @evacchi for the hard work here and your patience. I really appreciate it! 💯 👍 🚀

PR #1063, opened 2023-01-25
gojs/test: resolve race under scheduleTimeoutEvent

mathetake:
fixes #992

PR #1335, opened 2023-04-02
Allows WithMemoryLimitPages to override the default pages

### LINKED ISSUE #1330: Initialize memory from limit (with WithMemoryLimitPages + WithMemoryCapacityFromMax)
I would like to prevent re-allocation, so I tried to use `WithMemoryLimitPages(256).WithMemoryCapacityFromMax(true)`, in that
case it would allocate 256 pages and never change.

However, that is not what happens. Since the WASM-module is not defining `max memory`, the `WithMemoryCapacityFromMax(true)` is useless.

Maybe exposing the memory, in general, so we can provide any custom Memory implementation, or provide more option to avoid re-allocation.

I tried `WithMemoryLimitPages(255).WithMemoryCapacityFromMax(true)`, but it doesn't allocates 255 ahead-of-time.

mathetake:
Fixes #1330. This is more of a fix rather than behavioral change left in #1063.

### DIFF
--- a/internal/wasm/binary/decoder.go
+++ b/internal/wasm/binary/decoder.go
@@ -185,6 +185,9 @@ func newMemorySizer(memoryLimitPages uint32, memoryCapacityFromMax bool) memoryS
 			}
 			return minPages, minPages, *maxPages
 		}
+		if memoryCapacityFromMax {
+			return minPages, memoryLimitPages, memoryLimitPages
+		}
 		return minPages, minPages, memoryLimitPages
 	}
 }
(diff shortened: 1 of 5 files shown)


internal/wasm/binary/decoder.go
```diff
@@ -185,6 +185,9 @@ func newMemorySizer(memoryLimitPages uint32, memoryCapacityFromMax bool) memoryS
 			}
 			return minPages, minPages, *maxPages
 		}
+		if memoryCapacityFromMax {
+			return minPages, memoryLimitPages, memoryLimitPages
+		}
```
mathetake, line 190: this is the fix

PR #1336, opened 2023-04-03
Defer resource closure on context cancelation to m.FailIfClosed

### LINKED ISSUE #1314: Potential data race on sys.FS when context cancelation enabled.
Hitting a race condition that I don't quite understand.  Made a reproducible test for it here:

https://github.com/abraithwaite/wazero/commit/ca10894d

- Go version: 1.20
- wazero Version: tested v1.0.1 and 8a9d3f28
- Host architecture: arm64 (darwin) and amd64 (github runners)
- Runtime mode: Compiler

Not sure if the way we're using wazero is crazy, so please feel free to comment on that if you think this is improper usage of the library. 🙇‍♂️

mathetake:
Fixes #1314. Previously, when context cancelation is enabled, the resource is 
also closed in the `closeModuleOnCanceledOrTimeout` function which is 
asynchronously called concurrently to the Wasm function call. On the other hand,
host functions, especially wasi functions, do not check if the module is closed or not
therefore, if we reached the context cancelation while executing wasi functions, we
had race conditions.

This makes change to `closeModuleOnCanceledOrTimeout` so that it won't close resources,
and defer closing them to in `FailIfClosed` which is called synchronously during `api.Function.Call`.

### DIFF
--- a/internal/wasm/module_instance.go
+++ b/internal/wasm/module_instance.go
@@ -13,7 +13,7 @@ import (
 // FailIfClosed returns a sys.ExitError if CloseWithExitCode was called.
 func (m *ModuleInstance) FailIfClosed() (err error) {
 	if closed := atomic.LoadUint64(&m.Closed); closed != 0 {
-		switch closed & 0xff {
+		switch closed & exitCodeFlagMask {
 		case exitCodeFlagResourceClosed:
 		case exitCodeFlagResourceNotClosed:
 			// This happens when this module is closed asynchronously in CloseModuleOnCanceledOrTimeout,
@@ -52,6 +52,8 @@ func (m *ModuleInstance) closeModuleOnCanceledOrTimeout(ctx context.Context, can
 			// case go will randomize which branch of the outer select to enter
 			// and we don't want to close the module.
 		default:
+			// This is the same logic as CloseWithCtxErr except this calls closeWithExitCodeWithoutClosingResource
+			// so that we can defer the resource closure in FailIfClosed.
 			switch {
 			case errors.Is(ctx.Err(), context.Canceled):
 				// TODO: figure out how to report error here.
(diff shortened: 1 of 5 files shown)


mathetake:
@achille-roussel  @ckaznocha could you take a look when you have time?

internal/wasm/module_instance.go
```diff
@@ -13,6 +13,13 @@ import (
 // FailIfClosed returns a sys.ExitError if CloseWithExitCode was called.
 func (m *ModuleInstance) FailIfClosed() (err error) {
 	if closed := atomic.LoadUint64(&m.Closed); closed != 0 {
+		switch closed & 0xff {
```
achille-roussel, line 16: Is it worth declaring a const for the magic 0xff here? (e.g. `exitCodeFlagMask`)

internal/wasm/module_instance.go
```diff
@@ -45,7 +52,14 @@ func (m *ModuleInstance) closeModuleOnCanceledOrTimeout(ctx context.Context, can
...
 		default:
-			m.CloseWithCtxErr(ctx)
+			switch {
+			case errors.Is(ctx.Err(), context.Canceled):
+				// TODO: figure out how to report error here.
+				_ = m.closeWithExitCodeWithoutClosingResource(sys.ExitCodeContextCanceled)
+			case errors.Is(ctx.Err(), context.DeadlineExceeded):
+				// TODO: figure out how to report error here.
+				_ = m.closeWithExitCodeWithoutClosingResource(sys.ExitCodeDeadlineExceeded)
+			}
```
achille-roussel, line 62: Maybe comment that this code is an altered and inlined version of CloseWithError? It might be useful to understand why it's OK to not handle the default case.

achille-roussel (approved):
Looking good, and the fix makes sense 👍

mathetake:
@achille-roussel  Thanks for the review! 🙌

PR #1346, opened 2023-04-07
wasi: introduce platform.Select and use it for poll_oneoff

### LINKED ISSUE #1317: Support epoll on stdin
I thought we fixed this, but must have missed a spot. If you remove [nonBlockReader](https://github.com/ktock/container2wasm/blob/main/tests/wazero/main.go#L59) from container2wasm's wazero runner. Tests fail. Let's see what implementation is needed to get things together.

See https://github.com/ktock/container2wasm/pull/28#discussion_r1152863786

cc @evacchi @ncruces @ktock

evacchi:
The PR introduces the `platform.Select()` API, wrapping `select(2)` on POSIX and emulated in some cases on Windows. RATIONALE.md contains a full explanation of the approach followed in `poll_oneoff` to handle Stdin and the other types of file descriptors, and the clock subscriptions.

It also introduces an abstraction (`StdioFilePoller`) to allow the simulation of different scenarios (waiting for input, input ready, timeout expired, etc.) when unit-testing interactive input.

This closes #1317.

<details><summary>Old text</summary>

this is now using syscall.Select on *NIX

I have tested this manually with [imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.c](imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.c) and the reproducer at #1317 (modified similarly to support an additional arg: the timeout in seconds)

Testing manually, the expected behavior is the following.

1. in these 3 cases there should be (almost) no delay between the start of the program and the output to the console

```
❯ echo a | wazero run imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm poll 0
STDIN

❯ echo a | wazero run imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm poll 3
STDIN

❯ wazero run imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm poll 0
NOINPUT
```

2. in this case the console should wait 3 seconds if no input occurs

```
❯ wazero run imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm poll 3
NOINPUT
```

however, if input occurs on the stdin followed by ENTER (within the 3 seconds delay), then it should return immediately with STDIN:

```
❯ wazero run imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm poll 3
garbage here
STDIN
```

## Testing Strategy

I have now added customized the `StdioFileReader` so that it is easy to instantiate with a custom `Mode` and plug it into a module. This allows for simulating situations such as "`ModeCharDevice` with unready buffer"; e.g. the new `blockingReader` (`wasi_test.go`) is an `io.Reader` that blocks on `Read()` (unless the embedded context is cancelled)

A tty reader waiting on input can be simulated with:

```go
			moduleConfig := wazero.NewModuleConfig().WithArgs(tc.args...).
				WithStdin(internalsys.NewStdioFileReader(
				newBlockingReader(t), // simulate waiting for input
				stdinFileInfo(fs.ModeDevice|fs.ModeCharDevice|0o640),
				internalsys.PollerNeverReady),
))
```

A tty reader with a full buffer can be simulated with:

```go
			moduleConfig := wazero.NewModuleConfig().WithArgs(tc.args...).
				WithStdin(internalsys.NewStdioFileReader(
				strings.NewReader("test"), // input ready
				stdinFileInfo(fs.ModeDevice|fs.ModeCharDevice|0o640),
				internalsys.PollerAlwaysReady),
```

Although this is obviously not 100% equivalent to reading the real Stdin, it should give us a way to be a bit more confident about breaking it in the future :)

## Caveat: Windows Support

Currently, I have mocked `Select` on Windows to support only 2 cases 

- timeout, no FDs: always waits for timeout to expire (uses sleep)
- FD 0: always returns it has data ready without waiting for the timeout

the first case probably is not even necessary (poll_oneoff sleeps directly)
the second case is not "correct" in that is not equivalent to the real select, but it's good enough (it was the behavior we were having originally before this PR)

## Caveat: Zig cc

~the example under `zig-cc` is actually compiled using the `wasi-sdk` because `0.11.0-dev.2401+348751462` is generating a binary that panics for `wazero run wasi.wasm ls .` (unreachable code)~

EDIT it turns out it was not `zig cc` after all -- see #1371 

As usual it is kind of hard to write tests for this, especially for STDIN. I am opening this draft and then will backfill the tests.

~Roughly, this PR uses goroutines and a cancellable context to simulate the behavior of poll/select.~

### Special casing Stdin

Although it may look kind of dirty, apparently most major implementations are handling stdin in a special way, so it looks like this is the most common way to do it.

### Consuming Stdin

 As far as I can tell, there is no way **in the Go stdlib** to **peek** at stdin without _consuming_ the input unless you resort to the `syscall` package. In that case, we have `syscall.Select()` but only on *nix. This solution should work on Windows too (I can't try right now, will do later).

However, this solution _does_ consume the input, because the trick I am using is to wrap the stdioReader into a buffered reader and invoke Peek(1), returning the buffered Reader to the unsuspecting end user 😈 

For most intents and purposes this should make no difference. However, observe:

```
❯ ./a.out poll 3
blah
STDIN

❯ blah
zsh: command not found: blah
```

vs.

```
❯ wazero run imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm poll 3
blah
STDIN
```

because wazero really _is_ consuming the input, the input is not returned to the console, and it is instead swallowed by wazero; a native executable "leaks" it to the prompt which I guess is the "correct" behavior of a "real" `select()`. 

I also believe this is nitpicking, since the VM is a hosted execution environment and I don't think we are supposed to give any guarantees about the behavior of the virtualized syscalls (in practice, one may even go as a far as arguing that swallowing the text and not leaking it may be even "better"). 

In any case, this is worth noting, because it is not something that we are doing intentionally, but it is a consequence of using a buffered reader for lack of a better API.

</details>


imports/wasi_snapshot_preview1/poll_test.go
```diff
@@ -137,7 +124,7 @@ func Test_pollOneoff_Errors(t *testing.T) {
 			resultNevents: 512, // past out
 			expectedMem: []byte{
 				0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, // userdata
-				byte(expectedFdReadErr), 0x0, // errno is 16 bit
+				byte(wasip1.ErrnoSuccess), 0x0, // errno is 16 bit
```
evacchi, line 127: technically this is no longer an error and it is no longer returning unsupported or EBADF so maybe it makes sense to give the test case a different name.

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -72,88 +81,148 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
...
+	// Write the event corresponding to the processed subscription.
+	// https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-event-struct
+	copy(outBuf, value.userData)            // userdata
+	outBuf[value.outOffset+8] = value.errno // uint16, but safe as < 255
+	outBuf[value.outOffset+9] = 0
+	le.PutUint32(outBuf[value.outOffset+10:], uint32(value.eventType))
+	// TODO: When FD events are supported, write outOffset+16
+}
+
+// validateFDEvent returns a validation error or syscall.ENOTSUP as file or socket
```
evacchi, line 207: in practice we are now returning ENOTSUP in fewer cases than we originally did -- we really are supporting (some) FD events in some way...

evacchi:
zig:

> 525/2123 test.timestamp... FAIL (TestUnexpectedResult)

🤔

evacchi:
urm, "Test / amd64, windows-2022, Go-1.18 (pull_request)" failed, and I suspect a timeout. Restarting, but that smells like a flaky test -- not really surprising considering we are indeed testing timeouts.

mathetake:
https://github.com/tetratelabs/wazero/actions/runs/4659148348/jobs/8245692603 
this error seems something to do with your change in this PR. I am against increasing the timeout, but instead I suggest you find the root cause in the test and fix it.

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -65,95 +77,148 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
...
-			outBuf[outOffset+8] = byte(wasip1.ToErrno(errno)) // uint16, but safe as < 255
-		} else { // special case ass ErrnoSuccess is zero
-			outBuf[outOffset+8] = 0
+	// process timeout and interactive inputs (if any)
+	if timeout > 0 && readySubs == 0 {
+		timeoutCtx, cancelFunc := context.WithTimeout(ctx, timeout)
+		defer cancelFunc()
+
+		for _, s := range stdinSubs {
+			go processDelayedStdinReader(s, outBuf, cancelFunc)
```
ktock, line 143: Goroutines should be canceled otherwise they can leak maybe.

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -65,95 +77,148 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
...
 
-		// Write the event corresponding to the processed subscription.
-		// https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-event-struct
-		copy(outBuf, inBuf[inOffset:inOffset+8]) // userdata
-		if errno != 0 {
-			outBuf[outOffset+8] = byte(wasip1.ToErrno(errno)) // uint16, but safe as < 255
-		} else { // special case ass ErrnoSuccess is zero
-			outBuf[outOffset+8] = 0
+	// process timeout and interactive inputs (if any)
+	if timeout > 0 && readySubs == 0 {
```
ktock, line 138: It would be great if it checks stdin even when timeout=0.

c.f. https://linux.die.net/man/3/poll

> If none of the defined events have occurred on any selected file descriptor, poll() shall wait at least timeout milliseconds for an event to occur on any of the selected file descriptors. If the value of timeout is 0, poll() shall return immediately. If the value of timeout is -1, poll() shall block until a requested event occurs or until the call is interrupted.

mathetake:
could you do the first pass if you have time? @achille-roussel

ktock (commented):
Looks very cool, thanks for your work!

imports/wasi_snapshot_preview1/poll.go
ktock: Thanks for the comment.

> if stdin is interactive, then poll 0 would mean that the end user has not the chance to write anything interactively, so we may give up anyway

I think there are some cases where there are chances for user to write strings

e.g.

```c
#include <stdio.h>
#include <unistd.h>
#include <sys/select.h>

int main(int argc, char **argv)
{
  char buf[5];
  int len = 5, ret;
  fd_set rfds;
  struct timeval tv;

  for (int i = 0; i < 10; i++) {
    printf("checking inputs...\n");
    fflush(stdout);
    FD_ZERO(&rfds);
    FD_SET(0, &rfds);
    tv.tv_sec = 0;
    tv.tv_usec = 0;
    ret = select(1, &rfds, NULL, NULL, &tv);
    if ((ret > 0) && FD_ISSET(0, &rfds)) {
      ret = read(0, buf, len);
      if (ret > 0) {
        printf("got %.*s (first 5 chars)\n", ret, buf);
      }
    } else {
      printf("no input: %d\n", ret);
    }
    sleep(1); // do something
  }
}
```

host:

```console
$ ./a.out
checking inputs...
no input: 0
checking inputs...
no input: 0
hi
checking inputs...
got hi
 (first 5 chars)
```

But the current implementation of poll_oneoff can't receive the input so the above code doesn't work (i.e. input isn't detected).

evacchi: ah, great point. in fact, if I remove the `timeout > 0` (and TBF even if it's more expensive, the entire if check can be removed) the whole thing hangs as soon as there is one valid input 🤦 

```
❯ wazero run poll.wasm
checking inputs...
no input: 0
checking inputs...
no input: 0
x
checking inputs...
got x
 (first 5 chars)
checking inputs...
<hangs here>
```

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -65,95 +77,148 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
...
-			// +8 past userdata +8 contents_offset
-			errno = processFDEvent(mod, eventType, inBuf[inOffset+8+8:])
+			fsc := mod.(*wasm.ModuleInstance).Sys.FS()
+			fd := le.Uint32(argBuf)
+
+			stdinReader := processFDEvent(fsc, fd, evt)
+			if stdinReader == nil {
+				// if stdinReader is not an interactive session, then write back immediately
+				readySubs++
+				writeEvent(outBuf, evt)
```
ktock, line 127: It looks like `outBuf` isn't guaranteed to be zero-initialized? I think this should be initialized with zero at the beginning of this function.

imports/wasi_snapshot_preview1/poll.go
ktock, line 143: We should make sure that `outBuf` is fully written by `processDelayedStdinReader` before `pollOneoffFn` returns.

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -65,95 +77,148 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
...
 		if f, ok := fsc.LookupFile(fd); ok {
-			st, _ := f.Stat()
-			// if fd is a pipe, then it is not a char device (a tty)
-			if st.Mode&fs.ModeCharDevice != 0 {
-				errno = syscall.EBADF
+			// if fd corresponds to an interactive StdioFileReader then return it
+			if reader, ok := f.File.(*internalsys.StdioFileReader); ok && reader.IsInteractive() {
+				return reader
+			} else {
+				e.errno = wasip1.ErrnoSuccess
```
ktock, line 188: Stdin should be checked here as well otherwise `select` can report stdin even when there is no input happens. 

```c
#include <stdio.h>
#include <unistd.h>
#include <sys/select.h>

int main(int argc, char **argv)
{
  char buf[5];
  int len = 5, ret;
  fd_set rfds;
  struct timeval tv;

  FD_ZERO(&rfds);
  FD_SET(0, &rfds);

  tv.tv_sec = 3;
  tv.tv_usec = 0;
  printf("waiting for inputs...\n");
  ret = select(1, &rfds, NULL, NULL, &tv);
  if ((ret > 0) && FD_ISSET(0, &rfds)) {
    printf("input detected (ret=%d), reading...\n", ret);
    ret = read(0, buf, len);
    if (ret > 0) {
      printf("got %.*s (first 5 chars)\n", ret, buf);
    }
  } else {
    printf("no input\n");
  }
}
```

host:

```
$ cat | ./a.out
waiting for inputs...
no input
```

On wazero, it reports inputs even when no input happens and the following `read` blocks.

```
$ cat | go run main.go ./out/out.wasm
waiting for inputs...
input detected (ret=1), reading...
```

imports/wasi_snapshot_preview1/poll.go
ktock: Maybe https://github.com/tetratelabs/wazero/pull/1346#discussion_r1162492497  is related?

imports/wasi_snapshot_preview1/poll.go
evacchi: thanks!

codefromthecrypt:
gonna ask @achille-roussel to pinch-hit on review for me. Reasons include him being a better developer ;)

codefromthecrypt:
per slack offline I suggest we only handle poll for *nix and make a RATIONALE.md comment about windows (why not, which I guess there's no known heuristic). If helpful, we can special case stdio (including stdin as mentioned here) instead of os.File's in general. Further notes about how to emulate non-blocking could be extracted to an issue so they aren't lost.

Hope this helps!

codefromthecrypt (commented):
I think this approach is not using the underlying syscalls, which is ok provided we pass the tests. It feels a little gnarly as the normal path, not because of you, but because of the nature of trying to make blocking i/o feel as if it isn't.

More importantly, I noticed some new C code added as comments and unsure they have been integrated into wasi.c to ensure edge cases are covered.

In general, I prefer to go down the `syscall.Select` path, and only support darwin/freebsd/linux for now with ENOSYS otherwise. We could then make a portability try using readers for windows if syscall doesn't work. 

It would be similar to functions like `platform.UtimensFile` acknowledgin Select is probably the most complicated syscall we would try this way.

evacchi:
@codefromthecrypt I haven't added only this test (yet):

```
  for (int i = 0; i < 10; i++) {
    printf("checking inputs...\n");
    fflush(stdout);
    FD_ZERO(&rfds);
    FD_SET(0, &rfds);
    tv.tv_sec = 0;
    tv.tv_usec = 0;
    ret = select(1, &rfds, NULL, NULL, &tv);
    if ((ret > 0) && FD_ISSET(0, &rfds)) {
      ret = read(0, buf, len);
      if (ret > 0) {
        printf("got %.*s (first 5 chars)\n", ret, buf);
      }
    } else {
      printf("no input: %d\n", ret);
    }
    sleep(1); // do something
  }
```

 because it is kind of hard to try it non-interactively. Basically it verifies that even when the timeout to `select()` is 0, if you intersperse writes to stdin then the following `select()` will return > 0;

i.e. it is equivalent to:

```
    ret = select(1, &rfds, NULL, NULL, &tv);
    assert(ret == 0);
    sleep(1); // <-- write to stdin during this sleep
    ret = select(1, &rfds, NULL, NULL, &tv);
    assert(ret != 0);
```

evacchi:
I like how the STDIN vs NOINPUT test fails on only _some_ Linuxes :D

codefromthecrypt (commented):
Thanks for the work on this. It seems to be narrowing down to the things needed precisely. I'm happy to see the progress, and I think there are some dents to knock out so that the project cohesion is in tact afterwards.

First, I think the PR title and description need revision. This is more about adding platform.Select in order to implement normal posix select with timeout.

Please keep the impl portable, and not abstracted anywhere except where necessary to avoid copy/pasta within the platform package.

The platform package can test similar to x/sys but please in general let's not copy/pasta anything directly from other code, especially main. If we are copying tests, prefer Golang projects and cite exactly where they are coming from. e.g.

https://github.com/golang/sys/blob/master/unix/syscall_unix_test.go#L569-L647

I made some comments and really I think this must be like the other platform functions. A well defined single exported function that matches closest to posix, go syscall or x/sys.

Please don't here or anywhere in the future try to emulate a 3rd party library as re-use isn't worth the distraction of being different than normal go, and also makes IP unnecessarily complicated at the same time.

internal/platform/select_darwin.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+)
+
+func SelectTimeout(timeout time.Duration) error {
+	return Select(0, nil, nil, nil, timeout)
+}
+
+func SelectStdin(timeout time.Duration) (bool, error) {
+	fdSet := &FDSet{}
+	fdSet.Set(0)
```
codefromthecrypt, line 16: please use a constant for stdin's fd. When this is done at the call site for wasi, it will be easier as we have FdStdin.

internal/platform/select_darwin.go
```diff
@@ -0,0 +1,56 @@
...
+func Select(n int, r, w, e *FDSet, timeout time.Duration) error {
+	t := syscall.NsecToTimeval(timeout.Nanoseconds())
+	return syscall.Select(n, (*syscall.FdSet)(r), (*syscall.FdSet)(w), (*syscall.FdSet)(e), &t)
+}
+
+// Code lifted from https://github.com/creack/goselect (MIT licensed)
+// Ports the macros FD_SET, FD_CLEAR, etc. to methods
+
+// FDSet wraps syscall.FdSet with convenience methods
+type FDSet syscall.FdSet
```
codefromthecrypt, line 33: sorry I forgot to add the comment of an alternate suggestion.

It seems to me that the platform-specific stuff we are ending up in copy/paste between impls are around two things: the type syscall.FdSet and also the mask value of NFDBITS.

NFDBITS isn't something library specific to goselect, it was likely copied/derived from x/sys and in any case, let's model x/sys not other codebases e.g. https://github.com/golang/sys/blob/master/unix/fdset.go

one trick you can see they did nicely is to define a separate file for fdset (and fdset_test). This allows you to centralize the copy/paste stuff. you can also put a dummy value in for _windows (as windows isn't supported, but this lets you compile).

codefromthecrypt (commented):
Sorry I left out a comment for an alternate suggestion. Basically, I think if we model x/sys instead, you can see how they have extracted FdSet into its own file. This is a good approach as we can re-use that file for unsupported platforms, and keep them compiling.

The `platform.Select` we export will be driven by a `select` function in `select_something.go`. This will hold the value of NFDBITS which can either be cited fully (e.g. looking up the OS value or making a comment etc), or stubbed out (e.g. in unsupported os)

so for example, select_unsupported.go (used for anything not implemented otherwise, such as windows) would have

```
int NFDBITS = 8*8 // simulate (in fact I would actually not export this, e.g nfdbits)

// select returns syscall.ENOSYS as it isn't supported.
func select(nfd int, r *FdSet, w *FdSet, e *FdSet, timeout *Timeval) (n int, err error) {
   return 0, syscall.ENOSYS
}
```

This approach should be codeshape more similar to how we've done other things, and my last two reviews more or less the missing "how to" on adding new syscall fakes in general. It boils down to..

* priority order: be like syscall package, x/sys and either way POSIX. Don't emulate further in the platform package
* avoid copy/paste between platforms as much as possible, as this prevents drift annoyance. sometimes this means making new files.
* only export things needed outside the package, and those exports usually dispatch to platform-specific stuff
* always document similar to the "last best" e.g. now we are citing the real POSIX docs with notes etc.

Ack the last things only matter when working, but anyway hope these tips help as I'll be a bit offline while this is finishing up!

internal/platform/select_64.go
```diff
@@ -0,0 +1,10 @@
+//go:build !darwin && !netbsd && !openbsd && (amd64 || arm64 || ppc64le || mips64 || mips64le || s390x || riscv64 || loong64)
+
+package platform
+
+// darwin, netbsd and openbsd uses uint32 on both amd64 and 386
```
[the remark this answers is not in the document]
evacchi: upon further inspection of x/sys I think this was misguided or a bad copypaste

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -72,88 +80,160 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
...
+			// stdin has data ready to for reading, write back all the events
+			for i := range stdinSubs {
+				evt := stdinSubs[i]
+				evt.errno = 0
+				writeEvent(outBuf, evt)
+			}
+		}
+	} else {
+		// No subscribers, just wait for the given timeout.
+		time.Sleep(timeout)
```
evacchi, line 167: I think this should be back to `sysCtx.Nanosleep()` but if I just revert this line naively a bunch of tests will break, so I will investigate this further

imports/wasi_snapshot_preview1/poll_test.go
```diff
@@ -61,17 +62,6 @@ func Test_pollOneoff_Errors(t *testing.T) {
...
-	// `go test` forks processes. Instead, we test if this is consistent. For
-	// example, when run in a debugger, this could end up true.
-	// See also `terminal_test.go`.
-	expectedFdReadErr := wasip1.ErrnoNotsup
-	if stat, err := os.Stdin.Stat(); err != nil {
-		if stat.Mode()&fs.ModeCharDevice != 0 {
-			expectedFdReadErr = wasip1.ErrnoBadf
-		}
-	}
-
```
evacchi, line 74: this test alone should make less sense because now we can simulate the different states of the device

imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.c
```diff
@@ -37,16 +39,16 @@ void main_stat() {
   printf("/ isatty: %s\n", formatBool(isatty(3)));
 }
 
-void main_poll() {
+void main_poll(int timeout, int millis) {
```
evacchi, line 42: main_poll accepts durations now

imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.c
```diff
@@ -55,13 +57,41 @@ void main_poll() {
   }
 }
 
+void main_sleepmillis(int millis) {
```
evacchi, line 60: new test to check the clock (it wasn't tested originally)

imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm
evacchi, line 1: as mentioned, this was built using wasi-sdk because for some reason the last `zig cc` was producing a binary that panic'ed (`unreachable`)

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
```diff
@@ -217,18 +220,82 @@ func compileAndRun(t *testing.T, config wazero.ModuleConfig, bin []byte) (consol
...
-	// The "real" expected behavior is to return "NOINPUT",
-	// however the poll API is currently relying on stat'ing the file
-	// descriptor for stdin which makes the behavior platform-specific
-	// **during tests** and unfortunately hard to mock.
-	// For now, we just make sure the result is consistent.
-	if stat, err := os.Stdin.Stat(); err != nil {
-		if stat.Mode()&fs.ModeCharDevice != 0 {
-			require.Equal(t, "NOINPUT\n", console)
-			return
-		}
```
evacchi, line 231: same as above, this test should make less sense because now we can simulate more precisely different scenarios, without relying on the system STDIN

internal/platform/futimens_darwin.go
```diff
@@ -47,10 +47,3 @@ var libc_futimens_trampoline_addr uintptr
 // Note: CGO mechanisms are used in darwin regardless of the CGO_ENABLED value
 // or the "cgo" build flag. See /RATIONALE.md for why.
 //go:cgo_import_dynamic libc_futimens futimens "/usr/lib/libSystem.B.dylib"
-
-// syscall_syscall6 is a private symbol that we link below. We need to use this
-// instead of syscall.Syscall6 because the public syscall.Syscall6 won't work
-// when fn is an address.
-//
-//go:linkname syscall_syscall6 syscall.syscall6
-func syscall_syscall6(fn, a1, a2, a3, a4, a5, a6 uintptr) (r1, r2 uintptr, err syscall.Errno)
```
evacchi, line 56: moved to its own file, shared with Select

internal/sys/sys_test.go
```diff
@@ -22,6 +22,8 @@ func TestContext_WalltimeNanos(t *testing.T) {
 }
 
 func TestDefaultSysContext(t *testing.T) {
+	t.Skip("fixme: temporarily skip")
```
evacchi, line 25: I can't really figure out why the two lists of files are no longer equal. I suspect the changes to `stdioFileReader` have to do with this, but I can't tell what's wrong. Maybe something related to func references?

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
```diff
@@ -217,18 +220,83 @@ func compileAndRun(t *testing.T, config wazero.ModuleConfig, bin []byte) (consol
...
+			args: []string{"wasi", "poll"},
+			stdin: internalsys.NewStdioFileReader(
+				bufio.NewReader(strings.NewReader("test")), // input ready
+				stdinFileInfo(fs.ModeDevice|fs.ModeCharDevice|0o640),
+				internalsys.PollerAlwaysReady),
+			expectedOutput:  "STDIN",
+			expectedTimeout: 0 * time.Millisecond,
+		},
+		{
+			name: "custom reader, data ready, not tty, .5sec",
```
evacchi, line 244: TBH I am not very happy about these time-sensitive tests as they obviously tend to become flaky, as I am noticing right now. Suggestions welcome

imports/wasi_snapshot_preview1/poll.go
achille-roussel: We should also probably select on context cancellation in case the program wants to interrupt the blocking call?
```go
timer := time.NewTimer(timeout)
defer timer.Stop()
select {
case <-timer.C:
case <-ctx.Done():
}
```
It might be tricky to make context cancellation work with `platform.Select` as well.

internal/platform/select_darwin.s
```diff
@@ -0,0 +1,8 @@
+// lifted from golang.org/x/sys unix
+#include "textflag.h"
+
+TEXT libc_select_trampoline<>(SB), NOSPLIT, $0-0
+	JMP libc_select(SB)
+
+GLOBL ·libc_select_trampoline_addr(SB), RODATA, $8
+DATA ·libc_select_trampoline_addr(SB)/8, $libc_select_trampoline<>(SB)
```
achille-roussel, line 8: What is the reason we need this and cannot use `syscall.Select` for GOOS=darwin? https://pkg.go.dev/syscall?GOOS=darwin#Select

imports/wasi_snapshot_preview1/poll.go
codefromthecrypt: yeah it will be tricky to avoid leaking the underlying clock if/when a fake impl can work.

imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm
codefromthecrypt: "last zig cc" you mean latest from here https://ziglang.org/download/

after things are buttoned up, can you do a diff on the wasm2wat on the before and after and raise to zig? Along with the version of wasi-sdk as it might not be the same as what zig is currently using.

using wasi-sdk is a lot of work dev/x wise. Once the issue is raised, you can add a comment to the Makefile and remove this wasm saying add it back when zig cc works again (citing the issue). Otherwise, using make to recompile examples will overwrite and might surprise someone who forgot.

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
codefromthecrypt: I think you need a larger timeout. for example, we have a comment like this in Test_Nanosleep

```
	// In CI, Nanosleep(50ms) returned after 197ms.
	// As we can't control the platform clock, we have to be lenient
	ns := int64(50 * time.Millisecond)
```

codefromthecrypt:
> It might be tricky to make context cancellation work with platform.Select as well.

I think context cancelation won't affect things inside blocking io like reader.Read either. I think the main thing is to ensure that if we are looping we don't loop forever without checking the context isn't done. Even nicer if we can directly integrate, but I doubt that's mechanically possible. My 2p

imports/wasi_snapshot_preview1/poll.go
achille-roussel: If we want to support context cancellation, a common approach is to add a _signal_ file descriptor to the set, e.g. the read-end of a pipe or an eventfd on linux. When the module is closed, maybe asynchronously by the goroutine watching for context cancellation, we unblock the `platform.Select` call by writing to the fd.

codefromthecrypt: works for me! cool thing is because WASI doesn't support signals we don't need to worry about conflict interpreting them ;)

That said, this will need to be a part of the platform api sig not wasi, as it is an implementation detail of how we are supporting the context param.

So we'll have to special case the function signature or its impl, to support that. You might want to give some more concrete hints in the go side (code snippet) if you have some in mind.

Thanks for the idea!

internal/platform/select_darwin.s
evacchi: one reason is that `syscall.Select` on darwin insists on _not_ returning an `(int, error)` pair and only giving the error https://github.com/golang/go/blob/release-branch.go1.20/src/syscall/syscall_darwin.go#L188; the other reason is that it wraps the `syscall.Errno` into an `error` (although that's less of an issue because we have `UnwrapOSError`)

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
evacchi: yeah I increased them to 500 millis, hope that's good enough

internal/sys/sys_test.go
evacchi: my spider sense was tingling the right way, [`func` types are not comparable](https://go.dev/ref/spec#Comparison_operators)

evacchi: strategy pattern it is... don't hit me I am a Java dev recovering 🤣 

```
type StdioFilePoller interface {
	Poll(duration time.Duration) (bool, error)
}

type pollerDefaultStdin struct{}
...

// Poll implements StdioFilePoller for pollerDefaultStdin
func (*pollerDefaultStdin) Poll(duration time.Duration) (bool, error) { ... }
var PollerDefaultStdin = &pollerDefaultStdin{}
```

evacchi:
hey @ktock would you like to give this another go? :)

internal/sys/fs.go
```diff
@@ -61,21 +61,72 @@ func (w *stdioFileWriter) Close() error {
 	return nil
 }
 
-type stdioFileReader struct {
-	r io.Reader
-	s fs.FileInfo
+// StdioFilePoller is a strategy for polling a StdioFileReader for a given duration.
```
evacchi, line 64: The changes to this file have expanded significantly, maybe it's worth moving `Poller*` to a separate file?

internal/platform/select_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"os"
+	"runtime"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/tetratelabs/wazero/internal/testing/require"
+)
+
+func TestSelect(t *testing.T) {
```
evacchi, line 13: this test closely follows the tests in x/sys

internal/platform/select.go
```diff
@@ -0,0 +1,7 @@
+package platform
+
+import "time"
+
+func Select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error) {
```
evacchi, line 5: I haven't discussed this change, the one adaptation I did for the platform API is to use a Duration instead of a Timeval. There is a caveat though: the `select` syscall actually mutates the pointed timeval value (e.g. if the `select(2)` call terminates early; e.g. NetBSD, cf.: https://github.com/golang/sys/blob/master/unix/syscall_unix_test.go#L606-L617

internal/platform/fdset_test.go
```diff
@@ -0,0 +1,109 @@
+package platform
+
+import (
+	"runtime"
+	"testing"
+
+	"github.com/tetratelabs/wazero/internal/testing/require"
+)
+
+func TestFdSet(t *testing.T) {
```
evacchi, line 10: adapted from x/sys

imports/wasi_snapshot_preview1/poll.go
achille-roussel: I think we're going to quickly hit the case where someone runs into the limitations of the current implementation, but it's OK to keep it for a follow up 👍

internal/platform/select_darwin.s
achille-roussel: Even if `syscall.Select` returns a the number of events that completed, I believe the `select(2)` API still requires you to scan the whole fd set looking for file descriptors that are ready, so knowing how many there are does not really change the implementation isn't it?

internal/sys/fs.go
evacchi: I am also starting to wonder whether this change really is beneficial or if all I am doing is testing my own mocks... :/ 

The basic issue is that I wanted to be able to simulate various stdin scenarios and I could not find any other way other than mocking (I thought `os.Pipe()` would return no data with `Select()` but even those immediately return "data available" even when empty)

if I revert all these "poller" changes and restore `fs` then we lose the ability to simulate all these scenarios in tests (e.g. "a stdin that's not written for N millis"), but on the other hand, we are not really testing `select` but the behavior of the Poller (?)

internal/platform/select_windows.go
```diff
@@ -0,0 +1,29 @@
+package platform
+
+import (
+	"syscall"
+	"time"
+)
+
+const WasiFdStdin = 0
```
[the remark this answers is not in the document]
codefromthecrypt: that os.Stdin.Fd() is not 0 on windows is scary af. thanks for telling me :D

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -72,88 +80,161 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
...
+			writeEvent(outBuf, evt)
+		case wasip1.EventTypeFdRead:
+			fd := le.Uint32(argBuf)
+			if fd == internalsys.FdStdin {
+				// if the fd is Stdin, do not ack yet,
+				// append to a slice for delayed evaluation.
+				stdinSubs = append(stdinSubs, evt)
+			} else {
+				evt.errno = processFDEventRead(fsc, fd)
+				writeEvent(outBuf, evt)
```
ktock, line 130: Related to https://github.com/tetratelabs/wazero/pull/1346#discussion_r1162492497

When calling `poll_oneoff` multiple times it seems to return the result of the previous call.

sample c code:

```c
#include <stdio.h>
#include <unistd.h>
#include <sys/select.h>

int main(int argc, char **argv)
{
  char buf[5];
  int len = 5, ret;
  fd_set rfds;
  struct timeval tv;

  for (int i = 0; i < 10; i++) {
    printf("[%d] checking inputs ------------ \n", i);
    fflush(stdout);
    FD_ZERO(&rfds);
    FD_SET(0, &rfds);
    tv.tv_sec = 1;
    tv.tv_usec = 0;
    ret = select(1, &rfds, NULL, NULL, &tv);
    if ((ret > 0) && FD_ISSET(0, &rfds)) {
      printf("input detected %d\n", ret);
      ret = read(0, buf, len);
      if (ret > 0) {
        printf("got %.*s (first 5 chars)\n", ret, buf);
      }
    } else {
      printf("no input: %d\n", ret);
    }
  }
}
```

For debugging, we can inspect the result contents of `poll_oneoff` by the following patch:

```diff
diff --git a/imports/wasi_snapshot_preview1/poll.go b/imports/wasi_snapshot_preview1/poll.go
index 80ed4873..4e29fa24 100644
--- a/imports/wasi_snapshot_preview1/poll.go
+++ b/imports/wasi_snapshot_preview1/poll.go
@@ -4,6 +4,7 @@ import (
        "context"
        "syscall"
        "time"
+       "fmt"
 
        "github.com/tetratelabs/wazero/api"
        "github.com/tetratelabs/wazero/internal/platform"
@@ -71,6 +72,10 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
        if !ok {
                return syscall.EFAULT
        }
+       fmt.Println("outBuf value (initial): ", outBuf)
+       defer func() {
+               fmt.Println("outBuf value (result): ", outBuf)
+       }()
 
        // Eagerly write the number of events which will equal subscriptions unless
        // there's a fault in parsing (not processing).
```

As shown in the following, when an input happens on a call of `select`, next invocation of `select` also falsely detects the input. `poll_oneoff` seems to returns the the previous call's result which is contained in the uninitialized buffer passed as the argument.

```console
$  go run main.go ./out/out.wasm
[0] checking inputs ------------ 
outBuf value (initial):  [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 202 154 59 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
outBuf value (result):  [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 202 154 59 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
no input: 0
[1] checking inputs ------------ 
outBuf value (initial):  [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 202 154 59 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
a
outBuf value (result):  [0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 202 154 59 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
input detected 1
got a
 (first 5 chars)
[2] checking inputs ------------ 
outBuf value (initial):  [0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 202 154 59 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
outBuf value (result):  [0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 202 154 59 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
input detected 1
```

This can be fixed by explicitly initializing the buffer passed by the argument.

```diff
diff --git a/imports/wasi_snapshot_preview1/poll.go b/imports/wasi_snapshot_preview1/poll.go
index 80ed4873..826b6888 100644
--- a/imports/wasi_snapshot_preview1/poll.go
+++ b/imports/wasi_snapshot_preview1/poll.go
@@ -71,6 +71,9 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
        if !ok {
                return syscall.EFAULT
        }
+       for i := range outBuf {
+               outBuf[i] = 0
+       }
 
        // Eagerly write the number of events which will equal subscriptions unless
        // there's a fault in parsing (not processing).
```

ktock (commented):
Thanks!

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -60,6 +68,11 @@ func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) syscall.
 		return syscall.EFAULT
 	}
 	outBuf, ok := mem.Read(out, nsubscriptions*32)
+	// zero-out all buffer before writing
+	for i := range outBuf {
```
codefromthecrypt, line 72: I checked and seems this is idiomatic for filling with zeros https://groups.google.com/g/golang-nuts/c/8WSkwEs4YkU

internal/platform/select.go
codefromthecrypt: maybe move your comments here to godoc?

internal/platform/select_darwin.s
codefromthecrypt: can you summarize these notes as comments into the .s file? it is hard to see these after the PR is merged.

internal/platform/select_test.go
```diff
@@ -0,0 +1,90 @@
...
+	if runtime.GOOS != "linux" && runtime.GOOS != "darwin" {
```
codefromthecrypt, line 14: I see you have windows implementation. do you want to make a select_test_windows.go or maybe change this expression to allow windows (then conditionally skip some parts that don't work on it?)

internal/platform/select_windows.go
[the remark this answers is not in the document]
codefromthecrypt: unexport with a comment that we are using wasi fd for stdin because windows stdin is not fd 0

codefromthecrypt (approved):
I'm happy with this change, especially your notes and tenacity through the process. It is all internal, so polishing later can happen afterwards, that said I have a few polish requested 😎

imports/wasi_snapshot_preview1/poll.go
evacchi: you mean `copy(arr, make([]int, len(arr)))` right?

ktock (approved):
Thanks! LGTM

imports/wasi_snapshot_preview1/poll.go
codefromthecrypt: I mean what you are doing seems better than that, yeah (as it avoids alloc and I don't see anyone saying to do anything otherwise really)

evacchi: oh ok 👍

internal/platform/select_darwin.s
evacchi: @achille-roussel AFAIU, it is common to check the result to be `>0`; for instance, compare @ktock's example in #1317. 

@codefromthecrypt I'll update the godoc on `syscall_select`

evacchi: actually I think the godoc says enough

achille-roussel: I think the point I'm trying to make is it's unfortunate to take on the maintenance of platform dependent assembly code just to get an integer returned that only saves us from scanning the fd set when there is a timeout. 

An alternative could be to implement platform.Select on Darwin by calling syscall.Select and scanning the fd sets to generate the count of ready fds.

I'm also wondering why Go chose not to return a count of ready fds on Darwin, I would be surprised if it was an error, so I'd assume there's a reason which is probably related portability (maybe in some versions of OSX or iOS the return value was broken?).

Now that you know everything I know, I'll get behind the choice you make whether it's keeping the code as-is or changing 😊

evacchi:
Alright, thanks for the suggestions and the final ACKs. Merging this now.

internal/platform/select_darwin.s
evacchi: well, we already have a precedent on Darwin for `utimens` where we needed to do it for a similar reason. Anyway, we can always dump and update this code, after all the goal of this PR was to get the approach right, I would be ok if we reverted to using `syscall.Select()` at any point.

as for why Go did not return the result in the first place, I don't know if it was a mistake, but I do know that on x/sys/unix it returns `(int, error)`

https://github.com/golang/sys/blob/master/unix/syscall_darwin.go#L615

PR #1349, opened 2023-04-10
Allow perf to resolve guest symbols with perf maps

Pryz:
Related to #1350, this change adds the ability for Wazero to create perfmap files during the module compilation.
The "demangling" of the function names is let as the client tool discretion. `perf` and other `go tool pprof` already support demangle of c++ and rust function names.

I'm not certain about the way I'm propagating the enable flag for the perfmap feature. Happy to adapt to anyway we like.

### DIFF
--- a/config.go
+++ b/config.go
@@ -12,7 +12,6 @@ import (
 	"github.com/tetratelabs/wazero/api"
 	"github.com/tetratelabs/wazero/internal/engine/compiler"
 	"github.com/tetratelabs/wazero/internal/engine/interpreter"
-	"github.com/tetratelabs/wazero/internal/filecache"
 	"github.com/tetratelabs/wazero/internal/platform"
 	internalsys "github.com/tetratelabs/wazero/internal/sys"
 	"github.com/tetratelabs/wazero/internal/sysfs"
@@ -174,7 +173,7 @@ func NewRuntimeConfig() RuntimeConfig {
 	return newRuntimeConfig()
 }
 
-type newEngine func(context.Context, api.CoreFeatures, filecache.Cache, bool) wasm.Engine
+type newEngine func(context.Context, wasm.EngineConfig) wasm.Engine
 
 type runtimeConfig struct {
 	enabledFeatures       api.CoreFeatures
--- a/config.go
+++ b/config.go
@@ -161,6 +161,11 @@ type RuntimeConfig interface {
 	// When the invocations of api.Function are closed due to this, sys.ExitError is raised to the callers and
 	// the api.Module from which the functions are derived is made closed.
 	WithCloseOnContextDone(bool) RuntimeConfig
+
+	// WtihPerfmap enables perfmap files generation at compile time. Defaults to false.
+	//
+	// Perfmap files are used as source for debug symbols by external profiler such as `perf`.
+	WithPerfmap(bool) RuntimeConfig
 }
 
 // NewRuntimeConfig returns a RuntimeConfig using the compiler if it is supported in this environment,
@@ -169,7 +174,7 @@ func NewRuntimeConfig() RuntimeConfig {
 	return newRuntimeConfig()
 }
 
-type newEngine func(context.Context, api.CoreFeatures, filecache.Cache) wasm.Engine
+type newEngine func(context.Context, api.CoreFeatures, filecache.Cache, bool) wasm.Engine
 
 type runtimeConfig struct {
 	enabledFeatures       api.CoreFeatures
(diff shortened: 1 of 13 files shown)


internal/engine/compiler/engine.go
```diff
@@ -829,17 +843,18 @@ func (f *function) getSourceOffsetInWasmBinary(pc uint64) uint64 {
 	}
 }
 
-func NewEngine(_ context.Context, enabledFeatures api.CoreFeatures, fileCache filecache.Cache) wasm.Engine {
-	return newEngine(enabledFeatures, fileCache)
+func NewEngine(_ context.Context, enabledFeatures api.CoreFeatures, fileCache filecache.Cache, enabledPerfmap bool) wasm.Engine {
```
achille-roussel, line 846: Instead of passing a boolean parameter here, would it be possible to add a method like `EnablePerfMap(bool)`?

achille-roussel: Alternatively we could use a config struct for parameters, I'm concerned about seeing the number of arguments grow, we're setting a precedent that will encourage future contributions to keep adding parameters which will become unmanageable eventually.

codefromthecrypt: summary comment is that NewEngine can have a config struct like the other stuff. Seems getting to that spot and this is internal anyway, so worth a before/after (via spiking in a diff!)

Pryz: yes, I was considering the config struct but wanted to open the PR and get feedback before jumping on it. Will work on it!

internal/wasm/engine.go
```diff
@@ -52,3 +53,13 @@ type ModuleEngine interface {
...
 }
+
+// EngineConfig caries extra configuration for an Engine.
+type EngineConfig struct {
+	// EnabledPerfmap enable the generation of perfmap files used for profiling.
+	EnabledPerfmap bool
+
+	EnabledFeatures api.CoreFeatures
+
+	FileCache filecache.Cache
```
mathetake, line 64: these need comments!
```suggestion
	// EnabledFeatures
	EnabledFeatures api.CoreFeatures
	// FileCache
	FileCache filecache.Cache
```

config.go
```diff
@@ -161,6 +160,11 @@ type RuntimeConfig interface {
 	// When the invocations of api.Function are closed due to this, sys.ExitError is raised to the callers and
 	// the api.Module from which the functions are derived is made closed.
 	WithCloseOnContextDone(bool) RuntimeConfig
+
+	// WtihPerfmap enables perfmap files generation at compile time. Defaults to false.
+	//
+	// Perfmap files are used as source for debug symbols by external profiler such as `perf`.
+	WithPerfmap(bool) RuntimeConfig
```
mathetake, line 167: so how about having this as the experimental rather than 1st class API right away which we must promise API stability on? You can define and insert the param in context just like listeners: https://github.com/tetratelabs/wazero/blob/main/experimental/listener.go#L12

internal/wasmdebug/perfmap.go
mathetake, line 1: this file needs tests

internal/engine/compiler/engine.go
evacchi: I wonder if it's worth spinning off the config struct in its own PR so that we get that in earlier, and with its own commit 🤔  wdyt?

codefromthecrypt: I'm personally fine with tools being introduced as they are used. While it makes the commit title harder to write, it is better overall I think.

longer..

Some projects create a lot of infra in PRs and sometimes forget to use that infra, and regardless it can be hard to figure out what was motivating something when they are disconnected. Basically, the slight mushing of work challenges folks to use what they write and also can reduce the amount of polish PRs if what they thought they needed wasn't exactly it. Besides.. it is internal anyway.

codefromthecrypt: @evacchi if this is mainly about time until merge "get that in earlier"

if you need it and this PR is still pending, just add one and I'm sure @Pryz won't mind rebasing. wazero has an implicit no hostage policy ;)

evacchi: oh no, I don't need it. I just wanted to let @Pryz move forward in smaller steps if there were any issues, but I have no objection in keeping this in the same PR

codefromthecrypt:
old PRs are not good health signals. Can someone help pick this up, or is there another path to accomplish the same?

Pryz:
We went on the path of implementing another profiler as described in options 3 from https://github.com/tetratelabs/wazero/issues/1350. The profiler will be released on Thursday so I will close this change for now.
If anyone finds the need to use perf feel free to reopen a start fresh!

PR #1353, opened 2023-04-11
gojs: stop file caching in tests

mathetake:
No need to cache as the compilation perf improved drastically,
plus this prevents crashes when making changes to compiler backends,
which previously forced us to delete the .build dir manually.

```
2023/04/11 13:36:03 gojs.MustInstantiate took 0ms
2023/04/11 13:36:03 CompileModule took 538ms
2023/04/11 13:36:03 gojs.Run took 8ms
```

### DIFF
--- a/experimental/gojs/example/.gitignore
+++ b/experimental/gojs/example/.gitignore
@@ -1,2 +1,1 @@
 main.wasm
-.build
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
now this is a milestone indeed!

PR #1356, opened 2023-04-11
Fixes memory capacity with max larger than limit

mathetake:
thanks for reporting @ncruces


internal/wasm/binary/decoder.go
```diff
@@ -181,7 +181,7 @@ func newMemorySizer(memoryLimitPages uint32, memoryCapacityFromMax bool) memoryS
 			}
 			// This is a valid value, but it goes over the run-time limit: return the limit.
 			if *maxPages > memoryLimitPages {
-				return minPages, memoryLimitPages, memoryLimitPages
+				return minPages, minPages, memoryLimitPages
```
mathetake, line 184: at this point `memoryCapacityFromMax=false`, so this is correct

Issue #1350, opened 2023-04-10
Proposal: add guest CPU profiling


Pryz:
Profiling has been a very popular tool for many engineers to understand their code, work on performance issues and [much more](https://blog.felixge.de/8-unexpected-profiling-use-cases-beyond-performance-optimization/). CPU and Memory profiling are two critical features which are today required to run things in production. For the current proposal, I’m going to focus on CPU profiling but some interesting work can be done on the Memory side as well.

# Implementation

We have multiple options to implement CPU profiling in Wazero. Depending on if we want to provide a profiling mechanism from within Wazero or if we are ok just using an external system. In the second case, it’s all about providing the debugging symbols.

## Option 1: perf map

When using `perf` to profile a program in Linux, [perf maps](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt) are the easiest way to provide symbols for the resolution. The format is quite simple `START SIZE symbolname` and `perf` will automagically discover and use them if the file is written to `/tmp/perf-$pid.map`.

I’ve created a quick draft https://github.com/tetratelabs/wazero/pull/1349 with a potential implementation. This change is hooked to the engine to record the different symbols at compile time. Two caveats with this change: I’m using a dependency to demangle the symbols and it will need to be adapted once we use a single `mmap` to store the compiled code in memory instead of one per function.

Here is what the resolution of Rust symbols looks like:
<img width="807" alt="Screenshot 2023-04-09 at 9 35 14 PM" src="https://user-images.githubusercontent.com/971835/230820531-c44cbdc9-b905-426c-a291-0079010497d2.png">

## Option 2: jitdump+DWARF

Another approach is to create a jitdump file which is a similar approach than perfmap but contains more information and requires more steps to be used (see [wasmtime profiling](https://docs.wasmtime.dev/examples-profiling-perf.html#profiling-with-jitdump)). jitdump are also a bit more annoying to use. During the `perf inject` phase, `perf` will generate one `*.so` file per function which might be a lot depending on the type of program you run in the guest.

Note that some profilers out there, such as [Parca, already support reading symbols from jitdump files](https://github.com/parca-dev/parca-agent/pull/1051).

It is also possible to enhance the jitdump data with DWARF (if present in the WASM binary) which we already parse in Wazero to create stacktraces during panic events.

## Option 3: Profiler within Wazero

[...]

PR #1363, opened 2023-04-13
Provide new StackIterator to Before Listener hook

pelletier:
As mentioned in #1350, this pull request provides the Before Listener hook with a new `StackIterator` that allows the listener to walk the callstack.

The goal is to give the host more introspection capabilities into function calls, to enable use-cases such as CPU and memory profiling without having to implement them directly in Wazero. For example, a memory allocation profile can be generated by hooking into `malloc`-style function calls and recording the stack trace that lead to this invocation as well as the amount of allocated bytes.

First time interacting with this great codebase -- let me know if there is anything I can improve in this change. Hopefully this change does not break the backward compatibility promise of Wazero, since it the Listener API is in `experimental/`! 

I will add comments inline to call out specific parts of the change, and provide context to some questions I have.


experimental/listener.go
```diff
@@ -37,7 +37,7 @@ type FunctionListener interface {
 	//   - paramValues:  api.ValueType encoded parameters.
 	//
 	// Note: api.Memory is meant for inspection, not modification.
-	Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, paramValues []uint64) context.Context
+	Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, paramValues []uint64, stackIterator api.StackIterator) context.Context
```
pelletier, line 40: At that point, `paramValues` is always equivalent to `stackIterator.Next().Args()`. I thought about removing it off the bat, but maybe most people won't care about the whole stack and would prefer an easy access to parameters?

internal/engine/compiler/engine.go
```diff
@@ -1034,19 +1039,72 @@ func (ce *callEngine) builtinFunctionTableGrow(tables []*wasm.TableInstance) {
...
+		return true
+	}
+
+	if si.fn == nil || si.base == 0 {
+		return false
+	}
+
+	frame := si.base + callFrameOffset(si.fn.funcType)
+	si.base = int(si.stack[frame+1] >> 3)
+	si.fn = (*function)(unsafe.Pointer(uintptr(si.stack[frame+2]))) // TODO safer
```
pelletier, line 1075: Would love to know a better way to grab the `*function` pointer from the stackframe!

experimental/listener.go
achille-roussel: I agree with you, let's keep `paramValues` to avoid disrupting existing use cases too much. It would be good to document this tho so nobody is surprised.

internal/engine/interpreter/interpreter.go
```diff
@@ -166,6 +169,8 @@ type callFrame struct {
 	pc uint64
 	// f is the compiled function used in this function frame.
 	f *function
+	// base index in the frame of this function.
+	base int
```
pelletier, line 173: Couldn't figure out a way to detect how many values may be on the stack at the time of a call, so I added a sort of base pointer for this, but maybe I missed an invariant of correct WASM code that makes it unnecessary?

experimental/listener.go
achille-roussel: Would love to get @mathetake opinion on this tho ☝️

experimental/listener.go
```diff
@@ -37,7 +37,7 @@ type FunctionListener interface {
...
 
 	// After is invoked after a function is called.
```
pelletier, line 42: I only provided the `StackIterator` to the `Before` function because we only need it there at the moment, and it feels like it can be easily stored ahead of time in `Before`, only to be retrieved in the subsequent `After` call. It would maybe make sense to add there as well before the need arises. Let me know.

mathetake:
I'm offline until monday in japan, so excuse me for the delays in reviews from me :)

api/wasm.go
```diff
@@ -646,3 +646,21 @@ func EncodeF64(input float64) uint64 {
...
+
+// StackIterator allows iterating on each function of the call stack, starting
+// from the top. At least one call to Next() is required to start the iteration.
+//
+// Example:
+//
+//	for it.Next() {
+//		fmt.Printf("function: %s, args: %v", it.FnType(), it.Args())
+//	}
+type StackIterator interface {
```
codefromthecrypt, line 658: I doubt we can stabilize this type yet. Please move it to an internal package (maybe internal/sys? mainly somewhere to avoid cyclic deps). Then in experimental, copy or alias it to re-export the same methods.

Main thing is not having this in api, yet.

codefromthecrypt (commented):
I think most of this will be @mathetake to review. I'm mainly interested to make sure the exported non-experimental api surface isn't changed yet, except for experimental.

api/wasm.go
codefromthecrypt: ps I didn't mention why experimental fully on this one.

Basically, the only use of this StackIterator is an experimental API, which is why it should be experimental. Otherwise, it is "cart before the horse". When we lift the listener public, this would lift also.

pelletier: Makes sense! Thought I had declared it in experimental already, sorry I missed it.

internal/engine/compiler/engine.go
```diff
@@ -1034,12 +1039,61 @@ func (ce *callEngine) builtinFunctionTableGrow(tables []*wasm.TableInstance) {
...
```
codefromthecrypt, line 1075: I think this comment can live with the line.

```
	// TODO: find a better way to grab the *function pointer from the stackframe
	si.fn = (*function)(unsafe.Pointer(uintptr(si.stack[frame+2])))
```

experimental/listener.go
```diff
@@ -6,6 +6,24 @@ import (
 	"github.com/tetratelabs/wazero/api"
 )
 
+// StackIterator allows iterating on each function of the call stack, starting
```
codefromthecrypt, line 9: Mention somewhere that this is re-used for all calls with this listener?

experimental/listener.go
```diff
@@ -6,6 +6,24 @@ import (
...
+// from the top. At least one call to Next() is required to start the iteration.
+//
+// Example:
```
codefromthecrypt, line 12: maybe add another example in the example test ( or make stackiterator_example_test.go)  that does something like this, then remove this comment.

experimental/listener.go
```diff
@@ -6,6 +6,24 @@ import (
...
+//
+//	for it.Next() {
+//		fmt.Printf("function: %s, args: %v", it.FnType(), it.Args())
+//	}
+type StackIterator interface {
+	// Next moves the iterator to the next function in the stack. Returns false
+	// if it reached the bottom of the stack.
+	Next() bool
+	// FnType returns the function type of the current function.
```
codefromthecrypt, line 21: rename to FunctionDefinition and then you can comment like this. We typically don't abbreviate Function in public symbols.

```suggestion
	// FunctionDefinition returns the same value as the same method on api.Function
```

experimental/listener.go
codefromthecrypt: we can wait until then, but agree.

internal/engine/interpreter/interpreter.go
```diff
@@ -166,6 +169,8 @@ type callFrame struct {
 	pc uint64
 	// f is the compiled function used in this function frame.
 	f *function
+	// base index in the frame of this function.
```
codefromthecrypt, line 172:
```suggestion
	// base index in the frame of this function, used to detect the count of values on the stack.
```

codefromthecrypt (approved):
Some comments for polish.

Based on my scan, it doesn't seem like this should affect performance in any way unless listeners are in use. I don't have any blocking concerns, good work!

internal/engine/compiler/engine.go
```diff
@@ -878,6 +882,7 @@ func (e *moduleEngine) newCallEngine(stackSize uint64, fn *function) *callEngine
 		archContext:   newArchContext(),
 		initialFn:     fn,
 		moduleContext: moduleContext{fn: fn},
+		stackIterator: &stackIterator{},
```
mathetake, line 885: super nit: can we avoid the allocation here by having stackIterator as a value, not ptr?

internal/engine/compiler/engine.go
```diff
@@ -1034,12 +1039,61 @@ func (ce *callEngine) builtinFunctionTableGrow(tables []*wasm.TableInstance) {
 	ce.pushValue(uint64(res))
 }
 
+type stackIterator struct {
```
mathetake, line 1042:
```suggestion
// stackIterator implements experimental.StackIterator.
type stackIterator struct {
```

internal/engine/compiler/engine.go
```diff
@@ -1034,12 +1039,61 @@ func (ce *callEngine) builtinFunctionTableGrow(tables []*wasm.TableInstance) {
...
+	stack   []uint64
+	fn      *function
+	base    int
+	started bool
+}
+
+func (si *stackIterator) Reset(stack []uint64, fn *function, base int) {
```
mathetake, line 1049: q: this is exported for enginetest? if so, let's at least comment that here.

internal/engine/compiler/engine.go
```diff
@@ -1034,12 +1039,61 @@ func (ce *callEngine) builtinFunctionTableGrow(tables []*wasm.TableInstance) {
...
+	si.stack = stack
+	si.fn = fn
+	si.base = base
+	si.started = false
+}
+
+func (si *stackIterator) Clear() {
```
mathetake, line 1056: same q

internal/engine/compiler/engine.go
```diff
@@ -1034,12 +1039,61 @@ func (ce *callEngine) builtinFunctionTableGrow(tables []*wasm.TableInstance) {
...
+	si.stack = nil
+	si.fn = nil
+	si.base = 0
+	si.started = false
+}
+
+func (si *stackIterator) Next() bool {
```
mathetake, line 1063:
```suggestion
// Next implements experimental.StackIterator.
func (si *stackIterator) Next() bool {
```

internal/engine/compiler/engine.go
```diff
@@ -1034,12 +1039,61 @@ func (ce *callEngine) builtinFunctionTableGrow(tables []*wasm.TableInstance) {
...
+		return false
+	}
+
+	frame := si.base + callFrameOffset(si.fn.funcType)
+	si.base = int(si.stack[frame+1] >> 3)
+	si.fn = (*function)(unsafe.Pointer(uintptr(si.stack[frame+2]))) // TODO safer
+	return si.fn != nil
+}
+
+func (si *stackIterator) FnType() api.FunctionDefinition {
```
mathetake, line 1079:
```suggestion
// FnType implements experimental.StackIterator.
func (si *stackIterator) FnType() api.FunctionDefinition {
```

internal/engine/compiler/engine.go
```diff
@@ -1034,12 +1039,61 @@ func (ce *callEngine) builtinFunctionTableGrow(tables []*wasm.TableInstance) {
...
+	return si.fn.def
+}
+
+func (si *stackIterator) Args() []uint64 {
```
mathetake, line 1083:
```suggestion
// Args implements experimental.StackIterator.
func (si *stackIterator) Args() []uint64 {
```

internal/engine/compiler/engine.go
mathetake: unfortunately, this is the only way as we cannot have a typed stack. let's just comment on what's we are doing here.
```suggestion
	// *function lives in the third field of callFrame struct. This must be aligned with the definition of callFrame struct.
	si.fn = (*function)(unsafe.Pointer(uintptr(si.stack[frame+2])))
```

internal/engine/interpreter/interpreter.go
```diff
@@ -198,6 +203,54 @@ func functionFromUintptr(ptr uintptr) *function {
...
 }
 
+type stackIterator struct {
+	stack   []uint64
+	frames  []*callFrame
+	started bool
+	fn      *function
+}
+
+func (si *stackIterator) Reset(stack []uint64, frames []*callFrame, f *function) {
```
mathetake, line 213: same q

internal/engine/interpreter/interpreter.go
```diff
@@ -198,6 +203,54 @@ func functionFromUintptr(ptr uintptr) *function {
...
+	si.fn = f
+	si.stack = stack
+	si.frames = frames
+	si.started = false
+}
+
+func (si *stackIterator) Clear() {
```
mathetake, line 220: same q

internal/engine/interpreter/interpreter.go
```diff
@@ -198,6 +203,54 @@ func functionFromUintptr(ptr uintptr) *function {
...
+	si.stack = nil
+	si.frames = nil
+	si.started = false
+	si.fn = nil
+}
+
+func (si *stackIterator) Next() bool {
```
mathetake, line 227:
```suggestion
// Next implements experimental.StackIterator.
func (si *stackIterator) Next() bool {
```

internal/engine/interpreter/interpreter.go
```diff
@@ -198,6 +203,54 @@ func functionFromUintptr(ptr uintptr) *function {
...
+	}
+
+	frame := si.frames[len(si.frames)-1]
+	si.stack = si.stack[:frame.base]
+	si.fn = frame.f
+	si.frames = si.frames[:len(si.frames)-1]
+	return true
+}
+
+func (si *stackIterator) FnType() api.FunctionDefinition {
```
mathetake, line 244:
```suggestion
// FnType implements experimental.StackIterator.
func (si *stackIterator) FnType() api.FunctionDefinition {
```

internal/engine/interpreter/interpreter.go
```diff
@@ -198,6 +203,54 @@ func functionFromUintptr(ptr uintptr) *function {
...
+	return si.fn.def
+}
+
+func (si *stackIterator) Args() []uint64 {
```
mathetake, line 248:
```suggestion
// Args implements experimental.StackIterator.
func (si *stackIterator) Args() []uint64 {
```

internal/engine/interpreter/interpreter.go
mathetake: I guess this is the only way to get base so I think it's ok to define `base int` here 👍

mathetake (commented):
great work!

experimental/listener.go
```diff
@@ -6,6 +6,24 @@ import (
...
+	FnType() api.FunctionDefinition
+	// Args returns the api.ValueType encoded arguments of the current function.
```
mathetake, line 23: one thing to note: Wasm's args are _mutable_ -- the arguments are part of locals in Wasm functions and can be modified by `local.set` instruction.

experimental/listener.go
mathetake: yeah keeping params sounds good to me

internal/engine/compiler/engine.go
```diff
@@ -133,6 +133,10 @@ type (
 		// contextStack is a stack of contexts which is pushed and popped by function listeners.
 		// This is used and modified when there are function listeners.
 		contextStack *contextStack
+
+		// stackIterator provides a way to iterate over the stack for Listeners.
+		// It is setup and valid only during a call to a Listener hook.
+		stackIterator *stackIterator
```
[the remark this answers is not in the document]
mathetake: oh didn't see this comment. shall we just define this as value?
```suggestion
		stackIterator stackIterator
```

experimental/listener.go
pelletier: In the current implementation of the compiler `Args()` is a slice of the actual stack. Do you think it would be worth making a copy or providing a read-only proxy to avoid a potential mutation?

experimental/listener.go
pelletier: Should this be an implementation detail though? or do you mean to warn the user to not rely on keeping a pointer around?

internal/engine/compiler/engine.go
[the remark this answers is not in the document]
pelletier: Changed.

experimental/listener.go
pelletier: Added a comment on `Before()`.

experimental/listener.go
pelletier: Done

experimental/listener.go
pelletier: 👍🏻 Renamed

pelletier:
Thanks a lot everybody for the reviews! 🙏🏻  I have addressed the comments in the past 3 commits.

experimental/listener.go
codefromthecrypt: maybe all we need is a note to say copy/out immediately and/or don't retain the slice? I notice unsafe slice access in other libs like fasthttp...

pelletier:
@codefromthecrypt sorry I missed some in my first pass, should be all good now.

experimental/listener.go
codefromthecrypt: Actually I see what @mathetake means. I think we should rename this to Params and clarify that the values are current, not necessarily the same as what the function was invoked with (as a note).

codefromthecrypt (approved):
I think there's one dangling comment to address, then done!

experimental/listener.go
pelletier: Let me just rephrase to make sure I understand:

In Wasm, function parameters can be modified with a `local.set` instruction, so `Args()` may return values that are different than what the function was initially called with, so we should add a comment warning about this.

Also seems like the actual name in the Wasm spec is "parameters", not "arguments".

Does that sound right? Happy to make the changes!

pelletier: Took a stab at clarifying here https://github.com/tetratelabs/wazero/pull/1363/commits/70ec64a17b4d258b6c138b6ac2c0c89822b38fe8

mathetake (approved):
awesome!

experimental/listener.go
mathetake: that is exactly what I wanted to see! Thanks @pelletier

mathetake:
@pelletier thanks for the great job!

Issue #1060, opened 2023-01-24
compiler: mmap per module, instead of per function


mathetake:
currently, the machine code is compiled and memory-mapped as executables per-function basis and that is the current bottleneck on compilation performance; The below is the CPU profiling of [without extern cache benchmark](https://github.com/tetratelabs/wazero/blob/v1.0.0-pre.7/internal/integration_test/bench/bench_test.go#L84-L90) and it indicates that 63.7/73.6% = ~80% of `engine.CompileModule` is spent on mmap (for a TinyGo binary).

<img width="677" alt="image" src="https://user-images.githubusercontent.com/13513977/214443782-6e34ed16-6d68-4b06-9394-11fe8cf4ab26.png">

It should be possible to memory-map per module, which should improve the compilation performance better (Actually that is what I am actually doing in the new backend).

mathetake:
unassigning as this doesn't have priority

mathetake:
hmm, I've done some POC, but maybe this won't help that much. seems like in any way this results in the same amount of allocations

achille-roussel:
Would it help reduce the compute footprint still? syscalls tend to be expensive operations, if we can do a single mmap per module it amortizes the cost of context switch back and forth between kernel user space?

PR #1366, opened 2023-04-17
Avoids unnecessary allocations during mmap executables

mathetake:
Previously, we pass io.Reader to platform.MMap, and used io.Copy internally to 
generalize the code between amd64 and arm64 (the latter requires Mprotect after mmap).
That was causing a lot of allocations.

Before #1060, we should do this to see the fair comparison before/after.

```
goos: darwin
goarch: arm64
                                   │   sec/op    │   sec/op     vs base              │
Compilation_sqlite3/compiler-10      196.0m ± 1%   192.3m ± 0%  -1.89% (p=0.001 n=7)
Compilation_sqlite3/interpreter-10   61.87m ± 0%   62.03m ± 0%       ~ (p=0.073 n=7)
geomean                              110.1m        109.2m       -0.82%

Compilation_sqlite3/compiler-10      51.90Mi ± 0%   42.94Mi ± 0%  -17.26% (p=0.001 n=7)
Compilation_sqlite3/interpreter-10   51.75Mi ± 0%   51.75Mi ± 0%        ~ (p=0.073 n=7)
geomean                              51.83Mi        47.14Mi        -9.04%

                                   │  allocs/op  │  allocs/op   vs base               │
Compilation_sqlite3/compiler-10      32.20k ± 0%   27.66k ± 0%  -14.11% (p=0.001 n=7)
Compilation_sqlite3/interpreter-10   13.90k ± 0%   13.90k ± 0%        ~ (p=0.956 n=7)
geomean                              21.16k        19.61k        -7.33%

```


codefromthecrypt (commented):
good job. I like the code a lot better, too!

PR #1368, opened 2023-04-17
examples(allocation): free memory after unmarshalling a result from the guest

lburgazzoli:
This is a follow up PR to amend tinygo's allocation example according to https://github.com/knqyf263/go-plugin/pull/45


codefromthecrypt (approved):
looks good. can you sweep the docs, the tinygo one in site/ and also this one's README to make sure there are no other notes/code that are similar and need an update?

codefromthecrypt:
cool. for now this can merge anyway. more commits aren't a problem!

ncruces:
Not sure if this has a decent solution. The change was sound advice.

What I'm guessing happens is that the malloc heap is after the TinyGo heap. So allocating 10 bytes from malloc forces the  TinyGo heap to allocate (say) the full 1Gb of its own heap, and a page for the malloc heap.
On an OS this won't be a problem because the pages in the middle are not “commited” to memory. But since we're using a single []byte to hold all linear memory, we must grow it to 1Gb+1page.

Assuming a non-moving TinyGo GC, it's probably possible to directly return TinyGo memory, avoid copies, and ensure it's not GCed, but malloc/free is still the cleanest easiest to understand solution.

Are there any TinyGo knobs that might allow one to tweak heap sizes, layouts, etc?

codefromthecrypt:
Hi, can you try this PR again, except make sure the benchmark host code is updated the same way as the example host code was? Maybe this is why the regression happened https://github.com/tetratelabs/wazero/blob/main/internal/integration_test/vs/bench_allocation.go

lburgazzoli:
@codefromthecrypt https://github.com/tetratelabs/wazero/pull/1390

PR #1369, opened 2023-04-17
logging: fixes bug where unsampled logger is called from a sampled one

codefromthecrypt:
We had a logging bug where an unsampled function (such as fd_write to stdout/stderr) would end up with its logging "after" hook called, if it was called from a sampled function.

For example, if a wasm function called fd_write with stdout, the before hook on fd_write would skip, but the after would not, and accidentally use its caller's parameters. This results in a panic due to incorrect length.

This fixes the bug by ensuring we mute the logging context if there's one in progress. It ensures the bug won't pop up again by adding test data that matches the call pattern (from xpdf-wasm).

Thanks to @jerbob92 for helping isolate this!


codefromthecrypt:
a birdie told me this works

PR #1370, opened 2023-04-17
Makes host function index insertion order

codefromthecrypt:
This makes host functions index consistently on insertion order, rather than lexicographic order. This helps with ABI such as Emscripten, which need an expected order.

This also constrains the internal code around host functions to only one export name. More than one was never used. By restricting this, logic is simpler and smaller.


achille-roussel (approved):

codefromthecrypt:
thanks for the quick look @achille-roussel .. this helps my next not be tangled

PR #1371, opened 2023-04-17
wasi: fix zig-cc/wasi.c repeat arg

evacchi:
While working on #1346 I noticed that `zig cc` built a faulty
`wasi.wasm` that would fail on `ls`.

Only now have I realized that the `ls` subroutine was faulty to
begin with because we did not verify that `argv[3]` existed
before dereferencing. Oh, you, UB!


codefromthecrypt (approved):
bomb diffused

PR #1372, opened 2023-04-17
emscripten: adds experimental InstantiateForModule for invoke exports

### LINKED ISSUE #1364: Add emscripten.NewDynamicFunctionExporter

Currently, we hard-code a number of emscripten dynamic imports used in signatures like this.

```wat
	(import "env" "invoke_iiiii" (func $invoke_iiiii
		(param i32 i32 i32 i32 i32) (result i32))))
```

We cherry-picked signatures needed by PDFium, but what's used can change based on factors outside their control. In other words, functions can and do vary when recompiled. Here are some more

```
invoke_viiiddiiiiii
invoke_viiiii
invoke_vidddddd
invoke_viddddiidi
invoke_viiiiiiiii
invoke_viiiiiidii
invoke_dii
invoke_di
invoke_iiiiii
```

Notably, this now includes 'd', though not sure what value type 'd' implies as there's no document I'm aware of that defines these. We'll have to wasm2wat or otherwise to figure it out.

It is probably the right time to switch off statically generated dynamic imports. Along with `emscripten.NewFunctionExporter()` we can add `emscripten.NewDynamicFunctionExporter(guest)`. Since only PDFium used the current invoke_XX functions, we can probably stop making them by default in `emscripten.NewFunctionExporter()` and save some init time.

Thanks to @jerbob92, this is the critical flow, though parsing of types is required as well. Note the logic below won't compile.. it is a draft.

```go
type invokeFns []something

// NewFunctionExporter returns a FunctionExporter object with trace disabled.
func NewFunctionExporter(guest wazero.CompiledModule) FunctionExporter {
	functions := guest.ImportedFunctions()
	ret := invokeFns{}
	for i := range functions {
		importModule, importName, isImport := functions[i].Import()
		if isImport && importModule == "env" && strings.HasPrefix(importName, "invoke_") {
			invoke := // Generate host function here based on the function name signature.
			ret = append(ret, invoke) // for example if we can model dynamicFunctionExporter as a slice
		}
	}
	return ret
}

// ExportFunctions implements FunctionExporter.ExportFunctions
func (i invokeFns) ExportFunctions(builder wazero.HostModuleBuilder) {
	exporter := builder.(wasm.HostFuncExporter)
	exporter.ExportHostFunc(notifyMemoryGrowth)
	for _, invoke := range invokeFns {
		exporter.ExportHostFunc(invoke)
	}
```

Adding more static function imports!

**Additional context**

https://gophers.slack.com/archives/C04CG4A2NKX/p1681412714939329

codefromthecrypt:
This adds emscripten.InstantiateForModule into the experimental package. This builds dynamic invoke exports in the same order and only matching those needed by the importing modules.

Finally, this removes special casing of function type IDs by deferring
resolution of them only in Emscripten.

Fixes #1364


experimental/emscripten/testdata/invoke.wat
```diff
@@ -0,0 +1,115 @@
+(module
```
codefromthecrypt, line 1: this will be deleted once we can  bump minor (and merge the experimental function into the normal emscripten package)

internal/emscripten/emscripten.go
```diff
@@ -0,0 +1,95 @@
...
+	*wasm.FunctionType
+}
+
+// Call implements api.GoModuleFunction by special casing dynamic calls needed
+// for emscripten `invoke_` functions such as `invoke_ii` or `invoke_v`.
+func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
+	m := mod.(*wasm.ModuleInstance)
+
+	// Lookup the type of the function we are calling indirectly.
+	typeID, err := m.GetFunctionTypeID(v.FunctionType)
```
codefromthecrypt, line 74: trick is here.. by deferring resolution we don't need to do tricks of pre-allocating function type IDs. If any perf help is needed for emscripten later, e.g. if GetFunctionTypeID is slow, we can cache it as a field in InvokeFunc

mathetake (approved):

PR #1376, opened 2023-04-17
compiler: fixes false-positive invalid ptr detection in -race

mathetake:
looks like -race failing on the main after #1363


internal/engine/compiler/engine.go
```diff
@@ -1075,7 +1075,7 @@ func (si *stackIterator) Next() bool {
 	si.base = int(si.stack[frame+1] >> 3)
 	// *function lives in the third field of callFrame struct. This must be
 	// aligned with the definition of callFrame struct.
-	si.fn = (*function)(unsafe.Pointer(uintptr(si.stack[frame+2])))
+	si.fn = *(**function)(unsafe.Pointer(&si.stack[frame+2]))
```
codefromthecrypt, line 1078: wazero loves stars!

codefromthecrypt (commented):
thanks for the fix!

PR #1379, opened 2023-04-18
Add Go test suite against gotip

Pryz:
This change adds a Go test suite using gotip for the target. Once Go 1.21 is released we should be able to remove gotip totally. 

The list of package getting tested is fairly random. I tried to keep things that are "important" without taking hours of compute.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -375,90 +375,52 @@ jobs:
     name: Build Go test binaries
     runs-on: ubuntu-22.04
     env:
-      BINARY_OUT: ~/out
+      BINARY_OUT: ~/go-test-binaries
 
     steps:
-      - name: Check if binary already exists
-        id: check-exist
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: | # https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-artifacts-for-a-repository
-          EXIST=$(
-            (curl -s -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${{ env.GH_TOKEN }}" \
-              -H "X-GitHub-Api-Version: 2022-11-28" \
-              https://api.github.com/repos/tetratelabs/wazero/actions/artifacts | \
-              jq -e '.artifacts[] | select(.name == "go-test-binaries-${{ env.GO_VERSION }}")' &>/dev/null \
-            ) || echo false)
-          echo "exist=${EXIST}" >> $GITHUB_OUTPUT
+      # - name: Clear out cache 
+      #   env:
+      #     GH_TOKEN: ${{ github.token }}
+      #   run: |
+      #     curl -L \
+      #       -X DELETE \
+      #       -H "Accept: application/vnd.github+json" \
+      #       -H "Authorization: Bearer ${{ env.GH_TOKEN }}"\
+      #       -H "X-GitHub-Api-Version: 2022-11-28" \
+      #       https://api.github.com/repos/tetratelabs/wazero/actions/caches?key=go-test-binaries-${{ env.GO_VERSION }}
 
-      - name: Install Go
+      - name: Cache Go test binaries
+        id: cache-go-test-binaries
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: ${{ env.BINARY_OUT }}
+          key: go-test-binaries-${{ env.GO_VERSION }}
+
+      - if: ${{ steps.cache-go-test-binaries.outputs.cache-hit != 'true' }}
+        name: Install Go
         uses: actions/setup-go@v3
-        if: steps.check-exist.outputs.exist == 'false'
         with:
           go-version: ${{ env.GO_VERSION }}
 
-      - name: Install Go tip
+      - if: ${{ steps.cache-go-test-binaries.outputs.cache-hit != 'true' }}
+        name: Install and download Go tip
         run: |
           go install golang.org/dl/gotip@latest
           echo "$GOPATH/bin:$PATH" >> $GITHUB_PATH
-
-      - name: Download tip
-        run: |
           gotip download
 
-      - name: Build Test Binaries
-        if: steps.check-exist.outputs.exist == 'false'
-        # The following list of packages is coming from running : `find ./src -name '*_test.go' -exec dirname {} \; | sort | uniq`
-        # in the Go repository and manually filtering packages which don't make sense to be tested such as the 'cmd' package.
+      - if: ${{ steps.cache-go-test-binaries.outputs.cache-hit != 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yaml
```diff
@@ -370,3 +370,166 @@ jobs:
...
+          echo "$GOPATH/bin:$PATH" >> $GITHUB_PATH
+          gotip download
+
+      - if: ${{ steps.cache-go-test-binaries.outputs.cache-hit != 'true' }}
+        name: Build Test Binaries
+        run: |
+          mkdir -p ${{ env.BINARY_OUT }}/tests
+          # Build all Go std test binaries from gotip
+          pushd $(gotip env GOROOT)
+          for value in src/archive/tar \
```
codefromthecrypt, line 420:
```suggestion
          # Choose important packages to limit execution time.
          for value in src/archive/tar \
```

.github/workflows/integration.yaml
```diff
@@ -370,3 +370,166 @@ jobs:
...
           cd $(go env GOROOT)/src/os; wazero run -mount=/:/ ~/bin/os.wasm -test.v
+
+  build_go_tests:
+    name: Build Go test binaries
+    runs-on: ubuntu-22.04
+    env:
+      BINARY_OUT: ~/go-test-binaries
+
+    steps:
+      # - name: Clear out cache
```
codefromthecrypt, line 381: wonder if we shouldn't add this to the clear_cache workflow and add workflow_dispatch so that we can trigger manually instead of having commented out code?

.github/workflows/integration.yaml
```diff
@@ -370,3 +370,166 @@ jobs:
...
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Install wazero
+        run: go install ./cmd/wazero
+
+      - name: Run standard library tests
+        run: |
+          echo "Running $(find ${{ env.BINARY_OUT }} -name *.test | wc -l) test binaries"
+
+          for bin in ${{ env.BINARY_OUT }}/tests/*.test; do
```
codefromthecrypt, line 530:
```suggestion
          # Go tests often look for files relative to the source. Change to the corresponding directory.
          for bin in ${{ env.BINARY_OUT }}/tests/*.test; do
```

codefromthecrypt (approved):
LGTM on green

Pryz:
Looks like it needs some more files from gotip. I will cache the entire tree instead of just the src directory.

.github/workflows/integration.yaml
Pryz: I can try. I'm new to GH workflow so it's a nice learning opportunity

mathetake:
I imagine this doesn't pass on windows in any way for the same reasons as gojs :D

.github/workflows/integration.yaml
```diff
@@ -370,3 +370,172 @@ jobs:
...
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: ${{ env.BINARY_OUT }}
+          key: go-test-binaries-${{ env.GO_VERSION }}
+
+      - if: ${{ steps.cache-go-test-binaries.outputs.cache-hit != 'true' }}
+        run: |
+          echo "Go test binaries cache is empty"
+          exit 1
```
mathetake, line 514: let's use `fail-on-cache-miss` like other jobs
```suggestion
        with:
          path: ${{ env.BINARY_OUT }}
          key: go-test-binaries-${{ env.GO_VERSION }}
          fail-on-cache-miss: true
```

.github/workflows/integration.yaml
```diff
@@ -370,3 +370,172 @@ jobs:
...
+      #       -H "Accept: application/vnd.github+json" \
+      #       -H "Authorization: Bearer ${{ env.GH_TOKEN }}"\
+      #       -H "X-GitHub-Api-Version: 2022-11-28" \
+      #       https://api.github.com/repos/tetratelabs/wazero/actions/caches?key=go-test-binaries-${{ env.GO_VERSION }}
+
+      - name: Cache Go test binaries
+        id: cache-go-test-binaries
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-node-modules
```
mathetake, line 396: curious about what's this for... comments would be helpful!

.github/workflows/integration.yaml
```diff
@@ -370,3 +370,154 @@ jobs:
...
+    name: Build Go test binaries
+    runs-on: ubuntu-22.04
+    env:
+      BINARY_OUT: ./go-test-binaries
+
+    steps:
+      - name: Cache Go test binaries
+        id: cache-go-test-binaries
+        uses: actions/cache@v3
+        with:
```
mathetake, line 384: you need `enableCrossOsArchive: true` to enable this on windows

.github/workflows/integration.yaml
```diff
@@ -370,3 +370,154 @@ jobs:
...
         run: |
           cd $(go env GOROOT)/src/os; wazero run -mount=/:/ ~/bin/os.wasm -test.v
+
+  build_go_tests:
+    name: Build Go test binaries
+    runs-on: ubuntu-22.04
+    env:
+      BINARY_OUT: ./go-test-binaries
+
+    steps:
```
mathetake, line 380: you might want to checkout wazero _before_ the cache action in order to place the relative cache dir under wazero directory in the following jobs

.github/workflows/integration.yaml
```diff
@@ -370,3 +370,154 @@ jobs:
...
+    name: Go (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false # don't fail fast as sometimes failures are arch/OS specific
+      matrix:
+        os: [ubuntu-22.04, macos-12, windows-2022]
+    env:
+      BINARY_OUT: ./go-test-binaries
+
+    steps:
```
mathetake, line 489: also here let's checkout wazero before cache

.github/workflows/integration.yaml
mathetake: you still need this here as well @Pryz

.github/workflows/integration.yaml
mathetake: also @Pryz

Pryz:
I've only enabled Ubuntu for now. Windows and Macos are failing because they can't find gotip somehow. Not going to have much more time to work on this integration for now so I would propose we merge the first version with Ubuntu and look into macos and windows later. Thoughts?

Looks like we don't have the windows-2022 job working on gojs either.

codefromthecrypt:
> Looks like we don't have the windows-2022 job working on gojs either.

yeah basically go tests have a check for platform, but it is not able to read that the real host platform is windows. I thought GOHOSTOS would have done something but I felt I tried and it didn't matter (e.g. tests couldn't read that as GOOS)

PR #1380, opened 2023-04-18
Fixes syntax errors in examples

kevburnsjr:
I tried to copy/paste the example code but it had some syntax errors so I opened this PR to fix them.

There's a better way to do examples in Go that prevents syntax errors.
https://go.dev/blog/examples

Would you like me to migrate all the examples to Testable Examples after this PR is merged?


codefromthecrypt:
@kevburnsjr the one issue iirc in go examples is that you can't align them to interface methods. Is that still the case?

PR #1381, opened 2023-04-19
cache: fixes consistency with CloseOnContextDone and listeners

mathetake:
This fixes the compilation cache consistency with respect to the 
listeners and CloseOnContextDone configurations by including them
in the wasm.Module.ID.


internal/integration_test/README.md
mathetake, line 1: realized that this file is stale, so I just deleted while here.

internal/integration_test/filecache/filecache_test.go
mathetake, line 1: the tests in this file are the ones we lacked.

internal/wasm/module.go
```diff
@@ -191,9 +192,24 @@ const (
 	MaximumTableIndex    = uint32(1 << 27)
 )
 
-// AssignModuleID calculates a sha256 checksum on `wasm` and set Module.ID to the result.
-func (m *Module) AssignModuleID(wasm []byte) {
-	m.ID = sha256.Sum256(wasm)
+// AssignModuleID calculates a sha256 checksum on `wasm` and other args, and set Module.ID to the result.
+// See the doc on Module.ID on what it's used for.
+func (m *Module) AssignModuleID(wasm []byte, withListener, withEnsureTermination bool) {
```
mathetake, line 197: this is the essential change and others are just misc

codefromthecrypt (approved):
Thanks for the lightning turnaround!

Issue #1374, opened 2023-04-17
Where is binary located on Windows when installing from *.msi or with `winget`?


inliquid:
I just tried `winget install wazero` on Windows 10 and realized it provides zero information regarding where the binary is located on the system.

It's not available nowhere within the `%PATH%`.

When installing with `--verbose` flag it just says that installer was taken from:
https://github.com/tetratelabs/wazero/releases/download/v1.0.1/wazero_1.0.1_windows_amd64.msi

And running it manually also gives zero knowledge of where to locate the binary. I'm not even sure that it is saved after all.

codefromthecrypt:
Hi, sorry for your frustrations. Certainly, that's a non-goal!

This should place the wazero binary into to [program files 64 directory under wazero](https://github.com/tetratelabs/wazero/blob/main/packaging/msi/wazero.wxs#L78-L79). Also, it should be added to the PATH (which is a [part of our tests](https://github.com/tetratelabs/wazero/blob/main/packaging/msi/verify_msi.cmd#L22-L26), to ensure it is available). Note you have to refresh your shell to pickup PATH additions.

Do other winget installers introduce a manual step to control the installation folder or echo it to the screen? I'm wondering how we can do this more differently, and examples help.

inliquid:
Thank you, so effectively `wazero` binary was installed in:
`C:\Program Files\wazero` dir. (Shame that `winget` does not provide that info, there is already an issue: https://github.com/microsoft/winget-cli/issues/2298.)

However it's not in the `%PATH%` for some reason. I double-checked and it's not there.

I would of course install just by compiling it with `go`, but wanted to try that way as well.

codefromthecrypt:
I'll leave this open as I don't have my windows laptop with me. If someone else can take a look at if they also don't get PATH update and report back, much obliged.

evacchi:
yep I can confirm I can't find it in neither in my User %PATH% neither in the system %PATH% :/

codefromthecrypt:
@evacchi maybe have a look at func-e and/or podman to see if there was any drift or differences? (e.g. does winget func-e still work?) Maybe something skewed as sometimes utility upgrades break things (e.g. the thing that builds the installer).

evacchi:
same with func-e @codefromthecrypt TBH I noticed that something wasn't registered to PATH even with `choco` while setting this machine up, so it may not one specific issue of ours 🤔 nevertheless...

evacchi:
[...]

PR #1383, opened 2023-04-19
packaging: use setx /m to register wazero instead

evacchi:
Addresses #1374.

The most important issue with this strategy is that it has
the potential to break the end-user PATH variable if for some
reason the combined length is longer than 1024 chars.

The reason the current strategy is not working are at least two:

1. `%PATH%` is not expanded in:

    <CustomAction Id="ChangePath"
        ExeCommand="PATH &quot;%PATH%;[INSTALLDIR]&quot;"
        Property="setx" ...

    The "correct" syntax seems to be `[%PATH]` instead of `%PATH%`

2. SETX is not working(*) inside MSI (at least on my Win11 machine)
   unless we use the flag `/M` which modifies the env variable
   at the system level.

Furthermore:

- `SETX PATH "C:\Program Files\wazero;%PATH%"` expands `%PATH%`
  AND *truncates* it to 1024 chars. Because of this, *appending* may
  be safer: worst case, it truncates the path you have appended

- However, for some reason I couldn't make it work from within
  the MSI installer, so I still had to prepend...

(*) I assume it's running as a different user than %currentuser% 
and SYSTEM; maybe "Local Service"?

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -141,6 +141,10 @@ jobs:
       - name: Test Windows Installer
         if: runner.os == 'Windows'
         run: |
+          SET GARBAGE="12345678901234567890123456789012345678901234567890123456789012345678901234567890"
+          SET GARBAGE="%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%"
+          SET GARBAGE="%GARBAGE%%GARBAGE%"
+          SETX PATH "%PATH%;%GARBAGE"
           set MSI_FILE="dist\wazero_${{ needs.pre_release.outputs.VERSION }}_windows_amd64.msi"
           call packaging\msi\verify_msi.cmd
         shell: cmd
--- a/packaging/msi/wazero.wxs
+++ b/packaging/msi/wazero.wxs
@@ -94,10 +94,10 @@
             <ComponentRef Id="WazeroBin"/>
         </Feature>
 
-        <!-- Add wazero to the default PATH once installation succeeds.
+        <!-- Add wazero to the SYSTEM PATH once installation succeeds.
              Replace with <Component><Environment> after https://gitlab.gnome.org/GNOME/msitools/-/merge_requests/42 -->
         <Property Id="setx" Value="setx.exe"/>
-        <CustomAction Id="ChangePath" ExeCommand="PATH &quot;%PATH%;[INSTALLDIR]&quot;" Property="setx"
+        <CustomAction Id="ChangePath" ExeCommand="/M PATH &quot;[INSTALLDIR];[%PATH]&quot;" Property="setx"
                       Execute="deferred"/>
         <InstallExecuteSequence>
             <RemoveExistingProducts Before="InstallInitialize"/>


mathetake (commented):
can this be testable in [verify_msi.cmd](https://github.com/tetratelabs/wazero/blob/main/packaging/msi/verify_msi.cmd) which is executed in the [pre_release_test](https://github.com/tetratelabs/wazero/blob/main/.github/workflows/release.yaml#L145) CI?

evacchi:
Will do. I am also further investigating a different approach (using the actual WIX tool chain instead)

evacchi:
I have updated the wix file to v4 (it may not be strictly required TBF) and actually used the `Environment` section which seems to be working as expected (it shouldn't be a surprise, since I think this is the official tool). If we do go this route it'll require proper setup of the CI (but I think there are a bunch of GHA that may be used) see WIP https://github.com/evacchi/wazero/commit/464a9e37ca6905e3ab960481870d2350a5f45581#diff-217b40bac6395c4cd76179d71ebc5cda14719c5fec343444884d6faff2e61b2fR98

codefromthecrypt (approved):
I'm fine either way (this or switching to v4, or in steps both)

for verification, somehow it looked like we were verifying before, so maybe that's a different task to fix the false positive, but ideally done alongside the fix.

evacchi:
yes, I was just checking it and apparently we already verifying this. The issue may be that GHA is running as a privileged user so the problem doesn't present itself.

wix v3 is already present in the windows-2022 runner, it could be configured under Ubuntu but it requires to install wine and some fiddling... 🤔 so either we break the build into 2 steps or we figure out how to configure wix v3 under the Ubuntu runner... 🤔

codefromthecrypt:
just a thought.. if CI was working because the result path <1024 chars, maybe the CI step could be configured with a really long PATH (even if invalid entries) to force the issue..

evacchi:
in fact, in the meantime I have tested this smaller patch under the native wix3 toolchain and it seems to work https://github.com/evacchi/wazero/commit/wix3-native-build

.github/workflows/release.yaml
```diff
@@ -141,6 +141,10 @@ jobs:
       - name: Test Windows Installer
         if: runner.os == 'Windows'
         run: |
+          SET GARBAGE="12345678901234567890123456789012345678901234567890123456789012345678901234567890"
+          SET GARBAGE="%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%%GARBAGE%"
+          SET GARBAGE="%GARBAGE%%GARBAGE%"
+          SETX PATH "%PATH%;%GARBAGE"
```
evacchi, line 147: the cmd script is not failing with non-zero (I'll try and see how to make it actually fail), but this is indeed breaking PATH:

```
WARNING: The data being saved is truncated to 1024 characters.

SUCCESS: Specified value was saved.
installing "dist\wazero_8df8b975d660774e980afe9381df559e8e871167_windows_amd64.msi"
Refreshing environment variables from registry for cmd.exe. Please wait...Finished..
ensuring wazero was installed
8df8b975d660774e980afe9381df559e8e871[16](https://github.com/tetratelabs/wazero/actions/runs/4755047545/jobs/8448708184#step:6:17)7
uninstalling wazero
ensuring wazero was uninstalled
'wazero' is not recognized as an internal or external command,
operable program or batch file.
```

codefromthecrypt: thanks for your tenacity again. you seem to be a magnet for subtlety broken things :)

evacchi:
I have created https://github.com/tetratelabs/wazero/pull/1385 because I wanted to show both approaches. If I may explain visually:



In this PR:
- I could not manage to make the CI fail reliably when the %PATH% is corrupted. For that matter, I am not sure I was able to corrupt %PATH%
- The only approach that WorksForMe™️ is to _prepend_ the install path to the *system* path using `SETX /M`, which for long PATH env may push one of the valid elements of the list off the 1024 boundary, effectively corrupting the system PATH. Notice that I am not an expert, so I am just assuming.
- Another approach may be editing the registry directly; I didn't try this.

in PR #1385:
- msitools do not support the Environment directive, so we can try the "official" tools
- I was not able to run native WiX under macOS/Linux. They are not supported because they use native platform libraries. In theory there are ways to run it under wine: it didn't work for our project, in my tests (tried that too!)
- I am running the entire `make dist` step under on the `windows-2022` runner so that we can use native WiX Tools v3 
- alternative: run only `make dist "msi"` under a Windows runner

codefromthecrypt:
Thanks for the writeup. I think it is ok for us to run the release on windows until someone can compile the official tools to wasm, so we can run them on linux.

I was hoping msitools would progress, but the environment thing there went no-where.

As long as @mathetake is ok running on windows, I think we should go with PR https://github.com/tetratelabs/wazero/pull/1385. Maybe you can build an MSI on that and make sure the original person who had the problem is able to use it (e.g. attach the MSI to the PR or the issue leading to it, zipping if necessary)

evacchi:
closing in favor of https://github.com/tetratelabs/wazero/pull/1385

PR #1385, opened 2023-04-20
packaging: switch to native WiX toolchain and Windows runner

evacchi:
Cross-platform msitools do not support the `Environment` directive,
which should be the proper way to set up the PATH environment variable, 
so we adapt the Makefile to assume a Windows environment when building.
We address #1374.

`make dist` now runs on all platforms, but on Windows, assuming
`candle.exe`, `light.exe` and GNU `tar` are installed, it builds
all tarballs, zip and msi packages.

On other platforms it will only build the tarballs.

We also address some other issues with msitools
and change the GUID for UpgradeCode which caused
the wazero package to be mistaken for func-e
when installing/uninstalling both on the same
machine.


.github/workflows/release.yaml
```diff
@@ -44,10 +44,12 @@ jobs:
...
-      # ubuntu is missing wixl https://github.com/actions/virtual-environments/issues/3857
-      # ubuntu is also missing osslsigncode (no issue, yet)
-      - name: "Install GNOME msitools (wixl) and osslsigncode"
-        run: sudo apt update -qq && sudo apt install -qq -y wixl osslsigncode
+      # windows-2022 is missing osslsigncode (no issue, yet)
+      - name: "Install osslsigncode, 7zip; setup wix"
+        run: |
+          choco install osslsigncode -y
+          choco install 7zip -y
+          echo "C:\Program Files (x86)\WiX Toolset v3.11\bin" >> $GITHUB_PATH
```
evacchi, line 52: wix tools are installed on windows-2022 but not on the path

packaging/msi/wazero.wxs
```diff
@@ -49,9 +49,9 @@
              Language="1033" Codepage="1252">
         <!-- TODO: set <Property Id="ALLUSERS"... after https://gitlab.gnome.org/GNOME/msitools/-/issues/40 -->
 
-             InstallScope="perUser" is default (Id="ALLUSERS" doesn't work on msitools 0.101) -->
+            <!-- InstallScope="perUser" is default (Id="ALLUSERS" doesn't work on msitools 0.101) -->
```
evacchi, line 52: this "nested" comment was actually invalid syntax

packaging/msi/wazero.wxs
```diff
@@ -92,17 +97,11 @@
...
         </Feature>
 
-        <!-- Add wazero to the default PATH once installation succeeds.
-             Replace with <Component><Environment> after https://gitlab.gnome.org/GNOME/msitools/-/merge_requests/42 -->
-        <Property Id="setx" Value="setx.exe"/>
-        <CustomAction Id="ChangePath" ExeCommand="PATH &quot;%PATH%;[INSTALLDIR]&quot;" Property="setx"
-                      Execute="deferred"/>
         <InstallExecuteSequence>
             <RemoveExistingProducts Before="InstallInitialize"/>
-            <!-- Add PATH remove after after https://gitlab.gnome.org/GNOME/msitools/-/merge_requests/42 -->
```
evacchi, line 104: not necessary: `<Environment ... Permanent="no"...>` should do the trick

codefromthecrypt:
great notes. I think this is the path forward, but I can also see a reluctance to depend on a windows phase. That said, basically every maintainer needs to be able to run windows occasionally at this point, so I feel personally not too critical about it!

mathetake:
yeah let's go with this rather than #1383

Makefile
```diff
@@ -288,7 +288,7 @@ build/wazero_%/wazero.exe:
 dist/wazero_$(VERSION)_%.tar.gz: build/wazero_%/wazero
 	@echo tar.gz "tarring $@"
 	@mkdir -p $(@D)
-	@tar -C $(<D) -cpzf $@ $(<F)
+	@tar -C $(<D) -cpzf $@ --mode='+rx' $(<F)
```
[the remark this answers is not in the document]
evacchi: now setting this conditionally

.github/workflows/release.yaml
```diff
@@ -44,10 +44,12 @@ jobs:
...
 
-      # ubuntu is missing wixl https://github.com/actions/virtual-environments/issues/3857
-      # ubuntu is also missing osslsigncode (no issue, yet)
-      - name: "Install GNOME msitools (wixl) and osslsigncode"
-        run: sudo apt update -qq && sudo apt install -qq -y wixl osslsigncode
+      # windows-2022 is missing osslsigncode (no issue, yet)
+      - name: "Install osslsigncode, infozip; setup wix"
+        run: |
+          choco install osslsigncode -y
+          choco install zip -y
```
evacchi, line 51: installing zip instead of 7-zip so this part of the Makefile is left unchanged. This also makes it possible to run on non-Windows although it is currently skipped.

Makefile
```diff
@@ -288,7 +288,9 @@ build/wazero_%/wazero.exe:
 dist/wazero_$(VERSION)_%.tar.gz: build/wazero_%/wazero
 	@echo tar.gz "tarring $@"
 	@mkdir -p $(@D)
-	@tar -C $(<D) -cpzf $@ $(<F)
+# On Windows, we pass the special flag `--mode='+rx' to ensure that we set the executable flag.
+# This is only supported by GNU Tar, so we set it conditionally.
+	@tar -C $(<D) -cpzf $@ $(if $(findstring Windows_NT,$(OS)),--mode='+rx',) $(<F)
```
codefromthecrypt, line 293: 👍

Makefile
```diff
@@ -343,4 +351,4 @@ sha256sum := $(if $(findstring darwin,$(shell go env GOOS)),shasum -a 256,sha256
 $(checksum_txt):
 	@cd $(@D); touch $(@F); $(sha256sum) * >> $(@F)
 
-dist: $(non_windows_archives) $(windows_archives) $(checksum_txt)
+dist: $(non_windows_archives)  $(if $(findstring Windows_NT,$(OS)),$(windows_archives),) $(checksum_txt)
```
codefromthecrypt, line 354:
```suggestion
dist: $(non_windows_archives) $(if $(findstring Windows_NT,$(OS)),$(windows_archives),) $(checksum_txt)
```

codefromthecrypt (approved):
Thanks for turning this into a small change!

Makefile
```diff
@@ -261,12 +261,11 @@ fuzz:
...
 VERSION ?= dev
-# Default to a dummy version 0.0.1.rc1, which is always lower than a real release.
-# This must be in the form of [0-255].[0-255].[0-65535] plus optional fourth element which will be ignored.
-# We use the fourth field to represent the rc portion of release tag (e.g. rc1 of 1.0.0-rc2).
+# Default to a dummy version 0.0.1.1, which is always lower than a real release.
+# Legal version values should look like 'x.x.x.x' where x is an integer from 0 to 65534.
 # https://learn.microsoft.com/en-us/windows/win32/msi/productversion?redirectedfrom=MSDN
 # https://stackoverflow.com/questions/9312221/msi-version-numbers
-MSI_VERSION ?= 0.0.1.rc1
+MSI_VERSION ?= 0.0.1.1
```
evacchi, line 268: @codefromthecrypt notice that candle.exe does not support 0.0.1.x with x being text. It *must* be an integer or it will cause an error.

codefromthecrypt (approved):
Thanks. If you don't mind doing func-e once done as even though no-one has reported that, yet, we should have it in before people notice.

evacchi:
@codefromthecrypt speaking of func-e I noticed that while trying `func-e` _and_ `wazero` on my Windows machine, _and_ uninstalling I think it tried to uninstall wazero instead of func-e (or the other way round, don't remember right now). I suspect the reason is that they share the same GUID here:

    
    <?ifndef var.UpgradeCode?>
    <?define UpgradeCode = "FC6C326B-1DB7-4D48-8F05-E13E79DF4D0E" ?>
    <?endif?>

https://github.com/tetratelabs/func-e/blob/master/packaging/msi/func-e.wxs#L36-L40

shall we change that on `wazero` maybe? I can do it in this PR.

codefromthecrypt:
> shall we change that on wazero maybe? I can do it in this PR.

OMG please do

evacchi:
I removed the warning because it was being printed unconditionally always for any `make` target.

mathetake (approved):
💯

PR #1390, opened 2023-04-21
examples(allocation): free memory after unmarshalling a result from the guest

lburgazzoli:
Signed-off-by: Luca Burgazzoli <lburgazzoli@gmail.com>


lburgazzoli:
This is the benchamrks result I get by comparing main vs this PR:

```
➜ benchstat old.txt new.txt 
goos: linux
goarch: amd64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/compiler
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
                          │   old.txt    │                new.txt                 │
                          │    sec/op    │    sec/op      vs base                 │
Allocation/Compile-12       12.00m ± 15%    12.16m ±  2%          ~ (p=0.699 n=6)
Allocation/Instantiate-12   598.9µ ±  1%    622.0µ ±  3%     +3.85% (p=0.009 n=6)
Allocation/Call-12          2.031µ ±  6%   28.617µ ± 24%  +1308.99% (p=0.002 n=6)
geomean                     244.4µ          600.5µ         +145.71%

                          │   old.txt    │                new.txt                 │
                          │     B/op     │     B/op       vs base                 │
Allocation/Compile-12       2.553Mi ± 0%   2.553Mi ±  0%          ~ (p=0.818 n=6)
Allocation/Instantiate-12   348.0Ki ± 0%   348.0Ki ±  0%          ~ (p=0.626 n=6)
Allocation/Call-12            48.00 ± 0%    586.00 ± 10%  +1120.83% (p=0.002 n=6)
geomean                     34.94Ki        80.45Ki         +130.26%

                          │  allocs/op  │  allocs/op   vs base               │
Allocation/Compile-12       2.023k ± 0%   2.023k ± 0%        ~ (p=1.000 n=6)
Allocation/Instantiate-12    841.5 ± 0%    841.5 ± 0%        ~ (p=1.000 n=6)
Allocation/Call-12           5.000 ± 0%    6.000 ± 0%  +20.00% (p=0.002 n=6)
geomean                      204.2         217.0        +6.27%

pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/interpreter
                          │   old.txt    │               new.txt                │
                          │    sec/op    │    sec/op      vs base               │
Allocation/Compile-12       2.761m ±  8%    3.238m ± 14%  +17.27% (p=0.009 n=6)
Allocation/Instantiate-12   168.9µ ± 30%    152.4µ ± 82%        ~ (p=0.589 n=6)
Allocation/Call-12          53.99µ ± 11%   107.08µ ±  9%  +98.33% (p=0.002 n=6)
geomean                     293.1µ          375.2µ        +28.01%

                          │   old.txt    │               new.txt                │
                          │     B/op     │     B/op      vs base                │
Allocation/Compile-12       1.449Mi ± 0%   1.449Mi ± 0%         ~ (p=0.589 n=6)
Allocation/Instantiate-12   222.8Ki ± 0%   222.8Ki ± 0%         ~ (p=0.186 n=6)
Allocation/Call-12          2.241Ki ± 0%   6.569Ki ± 0%  +193.18% (p=0.002 n=6)
geomean                     90.49Ki        129.5Ki        +43.13%

                          │   old.txt   │                new.txt                │
                          │  allocs/op  │  allocs/op   vs base                  │
Allocation/Compile-12       1.076k ± 0%   1.077k ± 0%         ~ (p=0.061 n=6)
Allocation/Instantiate-12    747.0 ± 0%    747.0 ± 0%         ~ (p=1.000 n=6) ¹
Allocation/Call-12           98.00 ± 0%   264.00 ± 0%  +169.39% (p=0.002 n=6)
geomean                      428.7         596.6        +39.19%
¹ all samples are equal
```

I'm not sure if I did everything correct but it looks like there is still a large difference

internal/integration_test/vs/bench_allocation.go
```diff
@@ -51,13 +51,24 @@ func allocationCall(m Module, _ int) error {
...
 
 	// Now, we can call "greet", which reads the string we wrote to memory!
-	if err = m.CallI32I32_V(testCtx, "greet", namePtr, nameSize); err != nil {
-		return err
+	ptrSize, fnErr := m.CallI32I32_I64(testCtx, "greet", namePtr, nameSize)
+	if fnErr != nil {
+		return fnErr
 	}
 
 	// This pointer was allocated by Rust, but owned by Go, So, we have to
```
ncruces, line 59: Nit: TinyGo not Rust.

lburgazzoli: yeah, that was part already of the code so I didn't change it

ncruces:
The memory increase seems negligible (it's 1000%+, but just 500 _bytes_). My previous analysis is totally invalidated because I read Mi in the first line and assumed 500 _megabytes_ below. So please disregard that.

The CPU increase I don't know how to evaluate. Is a 20µs increase a problem?

If so, I would test a couple of things, to test the assumption that TinyGo using `cgo` might be the issue.

One is "caller allocates" convention: host calls malloc with a (perhaps oversized) buffer, passes the pointer and length in, guest returns the length.

The other is a global variable for the result. In the guest:
```go
// Use this global variable to keep the return value from an external function from being GCed.
var lastResult any

// Call this function to allow GCing the return value from an external function.
func ClearLastResult() { lastResult = nil }
```

codefromthecrypt:
I would chime in on 20us being a problem. I think it is because for example a web handler can return in less than a microsecond and tinygo is used to implement file handlers.

@dmvolod seemed to have a regression fix around tinygo recently and you can tell in these benchmarks that even using things like protobuf, the whole round trip is less than the additional overhead added here. https://github.com/knqyf263/go-plugin/pull/48

TL;DR; something's wrong and we shouldn't advocate (merge) this unless the overhead becomes more a percentage than a factor ;)

internal/integration_test/vs/bench_allocation.go
codefromthecrypt: leaving things better than you found them is a good practice! please fix typos as you see them.

lburgazzoli:
@codefromthecrypt, the second commit here, uses the patter suggested here https://github.com/tinygo-org/tinygo/issues/2787#issuecomment-1103780412

Results are:

```
➜ benchstat old.txt new.txt 
goos: linux
goarch: amd64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/compiler
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
panic: 
                          │   sec/op    │   sec/op     vs base                │
Allocation/Compile-12       11.93m ± 6%   12.01m ± 6%       ~ (p=0.937 n=6)
Allocation/Instantiate-12   608.9µ ± 2%   610.2µ ± 1%       ~ (p=0.937 n=6)
Allocation/Call-12          2.016µ ± 2%
geomean                     244.7µ        2.707m       +0.42%               ¹
¹ benchmark set differs from baseline; geomeans may not be comparable

                          │   old.txt    │               new.txt                │
                          │     B/op     │     B/op      vs base                │
Allocation/Compile-12       2.553Mi ± 0%   2.554Mi ± 0%  +0.02% (p=0.030 n=6)
Allocation/Instantiate-12   348.0Ki ± 0%   348.0Ki ± 0%       ~ (p=0.697 n=6)
Allocation/Call-12            48.00 ± 0%
geomean                     34.94Ki        953.9Ki       +0.01%               ¹
¹ benchmark set differs from baseline; geomeans may not be comparable

Allocation/Compile-12       2.023k ± 0%   2.034k ± 0%  +0.54% (p=0.002 n=6)
Allocation/Instantiate-12    841.0 ± 0%    841.0 ± 0%       ~ (p=1.000 n=6)
Allocation/Call-12           5.000 ± 0%
geomean                      204.1        1.308k       +0.27%               ¹
¹ benchmark set differs from baseline; geomeans may not be comparable

pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/interpreter
                          │   old.txt    │
                          │    sec/op    │
Allocation/Compile-12       3.017m ± 19%
Allocation/Instantiate-12   154.9µ ± 14%
Allocation/Call-12          58.87µ ± 10%
geomean                     301.9µ

                          │   old.txt    │
                          │     B/op     │
Allocation/Compile-12       1.449Mi ± 0%
Allocation/Instantiate-12   222.8Ki ± 0%
Allocation/Call-12          2.239Ki ± 0%
geomean                     90.46Ki

                          │   old.txt   │
                          │  allocs/op  │
Allocation/Compile-12       1.077k ± 0%
Allocation/Instantiate-12    747.0 ± 0%
Allocation/Call-12           98.00 ± 0%
geomean                      428.8

panic: runtime error: invalid memory address or nil pointer dereference
                          │   new.txt    │
                          │    sec/op    │
Allocation/Compile-12       2.902m ± 13%
Allocation/Instantiate-12   150.6µ ±  4%
geomean                     661.2µ

                          │   new.txt    │
                          │     B/op     │
Allocation/Compile-12       1.458Mi ± 0%
Allocation/Instantiate-12   222.8Ki ± 0%
geomean                     576.7Ki

                          │   new.txt   │
                          │  allocs/op  │
Allocation/Compile-12       1.087k ± 0%
Allocation/Instantiate-12    747.0 ± 0%
geomean                      901.1
```

internal/integration_test/vs/bench_allocation.go
lburgazzoli: done

codefromthecrypt:
hmm `panic: runtime error: invalid memory address or nil pointer dereference` so I think something is busted, but luckily doesn't need the benchmark to find it. Can you have a look at `TestAllocation`?.

internal/integration_test/vs/bench_allocation.go
```diff
@@ -37,7 +37,7 @@ func init() {
 
 func allocationCall(m Module, _ int) error {
 	nameSize := uint32(len(allocationParam))
-	// Instead of an arbitrary memory offset, use Rust's allocator. Notice
+	// Instead of an arbitrary memory offset, use guest's allocator. Notice
```
codefromthecrypt, line 40: nit: "guest" isn't a proper name, so please switch to TinyGo or "the guest"

internal/integration_test/vs/runtime.go
```diff
@@ -183,6 +184,18 @@ func (m *wazeroModule) CallI32I32_V(ctx context.Context, funcName string, x, y u
 	return
 }
 
+func (m *wazeroModule) CallI32I32_I64(ctx context.Context, funcName string, x, y uint32) (uint64, error) {
+	results, err := m.funcs[funcName].Call(ctx, uint64(x), uint64(y))
+	if err != nil {
+		return 0, err
+	}
+	if len(results) > 0 {
```
codefromthecrypt, line 192: this should never be true, basically this is a bug if the sig is I32I32_I64, so panic or return an error.

lburgazzoli:
@codefromthecrypt  with the latest commit which implements the suggestion in https://github.com/tinygo-org/tinygo/issues/2787#issuecomment-1103780412, the results are:

```
➜ benchstat old.txt new.txt 
goos: linux
goarch: amd64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/compiler
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
Allocation/Compile-12       11.93m ± 6%   11.96m ± 4%        ~ (p=1.000 n=6)
Allocation/Instantiate-12   608.9µ ± 2%   624.0µ ± 2%   +2.47% (p=0.041 n=6)
Allocation/Call-12          2.016µ ± 2%   2.488µ ± 4%  +23.41% (p=0.002 n=6)
geomean                     244.7µ        264.8µ        +8.23%

Allocation/Compile-12       2.553Mi ± 0%   2.554Mi ± 0%   +0.03% (p=0.041 n=6)
Allocation/Instantiate-12   348.0Ki ± 0%   352.2Ki ± 0%   +1.22% (p=0.002 n=6)
Allocation/Call-12            48.00 ± 0%     56.00 ± 0%  +16.67% (p=0.002 n=6)
geomean                     34.94Ki        36.93Ki        +5.71%

                          │  allocs/op  │  allocs/op   vs base               │
Allocation/Compile-12       2.023k ± 0%   2.033k ± 0%   +0.49% (p=0.002 n=6)
Allocation/Instantiate-12    841.0 ± 0%    843.0 ± 0%   +0.24% (p=0.002 n=6)
Allocation/Call-12           5.000 ± 0%    6.000 ± 0%  +20.00% (p=0.002 n=6)
geomean                      204.1         217.5        +6.52%

pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/interpreter
                          │    sec/op    │    sec/op     vs base               │
Allocation/Compile-12       3.017m ± 19%   2.978m ± 10%        ~ (p=1.000 n=6)
Allocation/Instantiate-12   154.9µ ± 14%   151.1µ ±  1%        ~ (p=0.310 n=6)
Allocation/Call-12          58.87µ ± 10%   65.98µ ±  3%  +12.08% (p=0.004 n=6)
geomean                     301.9µ         309.6µ         +2.55%

                          │   old.txt    │              new.txt               │
Allocation/Compile-12       1.449Mi ± 0%   1.458Mi ± 0%  +0.57% (p=0.002 n=6)
Allocation/Instantiate-12   222.8Ki ± 0%   223.0Ki ± 0%  +0.06% (p=0.002 n=6)
Allocation/Call-12          2.239Ki ± 0%   2.330Ki ± 0%  +4.06% (p=0.002 n=6)
geomean                     90.46Ki        91.86Ki       +1.55%

                          │  allocs/op  │  allocs/op   vs base              │
Allocation/Compile-12       1.077k ± 0%   1.087k ± 0%  +0.93% (p=0.002 n=6)
Allocation/Instantiate-12    747.0 ± 0%    748.0 ± 0%  +0.13% (p=0.002 n=6)
Allocation/Call-12           98.00 ± 0%   103.00 ± 1%  +5.10% (p=0.002 n=6)
geomean                      428.8         437.5       +2.03%
```

internal/integration_test/vs/runtime.go
lburgazzoli: If I add something like

```go
	if len(results) == 0 {
		panic("unexpected")
	}
```

Then it panics, but don't know exactly why

codefromthecrypt: ok lemme pull this and take a look

internal/integration_test/vs/bench_allocation.go
```diff
@@ -50,14 +50,25 @@ func allocationCall(m Module, _ int) error {
 		return err
 	}
 
-	// Now, we can call "greet", which reads the string we wrote to memory!
-	if err = m.CallI32I32_V(testCtx, "greet", namePtr, nameSize); err != nil {
-		return err
+	// Now, we can call "greeting", which reads the string we wrote to memory!
+	ptrSize, fnErr := m.CallI32I32_I64(testCtx, "greeting", namePtr, nameSize)
```
codefromthecrypt, line 54: so the main thing here is that the benchmark is doing something different than before. Due to this, it cannot be compared directly to the prior (now it is calling "greeting" but before it was calling "greet"

internal/integration_test/vs/runtime.go
codefromthecrypt: the problem is that the code was switched to the signature of "greeting" but the function name called was "greet" which didn't return a value.

codefromthecrypt:
in some ways it looks like we are doing the same thing as tinymem now.

However, mixing and matching the built-in "malloc" and "free" with an explicit version (similar to tinygo) is a little confusing I think, especially in an example.

I prefer the example either use the built-in functions like it did before or manually manage, but not both.

So if you think that we shouldn't use malloc/free exported by tinygo then I think we just inline the latest version of the code in tinymem into the example/benchmark, then refer to tinymem as a repo that does this for you.

https://github.com/tetratelabs/tinymem/blob/main/example/greeting.go

codefromthecrypt:
so to the point of @ncruces here https://github.com/tetratelabs/wazero/pull/1390#issuecomment-1517755372


> One is a "caller allocates" convention: host calls malloc with a (perhaps oversized) buffer, passes the pointer and length in, guest returns the length.

this is what was going on before, and kindof is still. Similar to rust, this was using the underlying malloc/free built-in imports.

> The other is a global variable for the result. In the guest:

This is kindof happening now, except using a map instead of a single field. IMHO if doing something more complicated (like using a map), I think the code that is almost like tinymem should be exactly like that, since that code was reviewed etc, no need to inline something almost the same. In other words, not copy/paste from my old comment, rather the latest version of it.

OTOH, I don't know that the special malloc/free exports are going away any time, except possibly not exported by default. Frankly, I have lost track of the real motivation to change anything, but I also don't mind changing as long as it is coherent!

lburgazzoli:
@codefromthecrypt this whole exercise is an attempt to add some clarification about how host/guest memory should be handled for future devs that like me where looking for an example as it looks like this topic is not much clear, yet.

And this PR seems to be a confirmation :)

I don't have any preference about using timymem (which I discovered only today) or the builtin free/malloc and CGO so I would love to have some guidance about which one you'd like to be part of the examples. Also, should add a dedicated example and related benchmark ?

ncruces:
I guess if tinymem's goal is to encode best practices in terms of allocations across the host-guest boundary, it's only natural that an _example_ either uses it or reimplements it.

My suggestions were more with the goal of root causing the benchmark regression. From the data @lburgazzoli collected it seems using CGo from TinyGo might be the culprit?

And if that's so, that probably confirms that tinymem's approach is the best one, for a TinyGo guest.

codefromthecrypt:
Surprisingly or not I think I am on the same page with both! Maybe took a bit for my brain to catch up.

Conflicting advice is basically the opposite of clear, so that's why I think our example should not mix both in the same code. When they are mixed it is really difficult to figure out what advice is and even more unclear is the perf impacts.

So, let's recap our entrypoints: We have the example which shows new users how to do allocation, and we also have a safeguard Allocation benchmark which both serves to compare more realistic interaction performance (vs wasmtime etc), and also acts in some ways as a soak test. For example, it was through the benchmark we found things breaking functionally or performance. That is a surprising helpful thing.

Along the way, though, we changed the example to something it wasn't before. Specifically "greeting" vs "greet", so we can't really compare performance of the two based on the prior commit. Until the exact use case is run, both ways I don't think we can say for sure an outcome.

**Here's an idea, if folks have patience for it. Note: this is temporary to achieve an outcome of knowing if both approaches work or have dramatically different outcomes when compared exactly.**

 Change the tinygo allocation source to use a CLI arg to determine the style to use. Include in the source both approaches to implement greeting. both pure CGO including go imports and also the tinymem approach. For tinymem, literally copy/paste the code from tinymem so it isn't subtly different.

For example export "greeting_cgo" and "greeting_tinymem" then in the host side do both approaches switched on an arg. So, for example the cgo style would use "malloc" and tinymem style "_malloc". Another way is to make another directory, like tinygo-malloc, if it helps.

Then, run the benchmark against each. You can make a manual benchmark or modify BenchmarkAllocation to allow choosing which. The main idea is benchmarking exactly the same use case, both to see the results and also that they complete without error.

codefromthecrypt:
I also want to clarify from my personal POV, I didn't help make tinymem because I thought it was a better idea. This project was made defensively because tinygo were considering removing the CGO imports (those in use by go-plugin, trivy and others). I don't personally want folks to have to do memory things manually, as wasm is hard enough as it is!

lburgazzoli:
> **Here's an idea, if folks have patience for it. Note: this is temporary to achieve an outcome of knowing if both approaches work or have dramatically different outcomes when compared exactly.**
> 
> Change the go allocation source that calls tinygo's wasm to use a CLI arg to determine the style to use. Change the tinygo source to implement both approaches to "greeting": pure CGO including go imports and also the tinymem approach. For tinymem, literally copy/paste the code from tinymem so it isn't subtly different.
> 
> For example export "greeting_cgo" and "greeting_tinymem" then in the host side do both approaches switched on an arg. So, for example the cgo style would use "malloc" and tinymem style "_malloc". Another way is to make another directory, like tinygo-malloc, if it helps.
> 
> Then, run the benchmark against each. You can make a manual benchmark or modify BenchmarkAllocation to allow choosing which. The main idea is benchmarking exactly the same use case, both to see the results and also that they complete without error.

If there is not urgency, I would like to take this task

lburgazzoli:
In my latest commit, I added a POC for what has been described by @codefromthecrypt in the latest comments. 
Some names must probably be changed for something better but here I just want to have some feedback

codefromthecrypt (commented):
Thanks for progressing this! can you paste output of bench results for these two? This will help guide the discussion as people are more likely to engage when they see the difference (vs pulling and running manually)

internal/integration_test/vs/bench_allocation_tinymem.go
```diff
@@ -0,0 +1,89 @@
...
+	testCall(t, runtime, allocationTinyMemConfig, testAllocationTinyMemCall)
+}
+
+func testAllocationTinyMemCall(t *testing.T, m Module, instantiation, iteration int) {
+	err := AllocationTinyMemCall(m, iteration)
+	require.NoError(t, err, "instantiation[%d] iteration[%d] failed: %v", instantiation, iteration, err)
+}
+
+func RunTestBenchmarkAllocationTinyMem_Call_CompilerFastest(t *testing.T, vsRuntime Runtime) {
+	runTestBenchmark_Call_CompilerFastest(t, allocationTinyMemConfig, "Allocation", AllocationTinyMemCall, vsRuntime)
```
codefromthecrypt, line 84: Check the `make bench.check` target, as you may be able to get these to run automatically (e.g. in actions you can see below under 'test'. Possibly by renaming "Allocation" to something else, as I can see Allocation here, but not the new ones you made...

```
BenchmarkAllocation/Call-16
wazero-compiler 3261.21   ns/op
wasmedge        105794.42 ns/op
BenchmarkFactorial/Call-16
wazero-compiler 1117.06  ns/op
wasmedge        12121.32 ns/op
wazero-compiler 113.91  ns/op
wasmedge        2686.57 ns/op
Benchmark/memory.i32/Call-16
wazero-compiler 374.12   ns/op
wasmedge        12607.52 ns/op
Benchmark/memory.i64/Call-16
wazero-compiler 380.94   ns/op
wasmedge        13439.59 ns/op
```

lburgazzoli: done

lburgazzoli:
> Thanks for progressing this! can you paste output of bench results for these two? This will help guide the discussion as people are more likely to engage when they see the difference (vs pulling and running manually)

```
➜ go test -run='^$' -bench '^BenchmarkAllocation.*' ./internal/integration_test/vs/compiler ./internal/integration_test/vs/interpreter -count=6
goos: darwin
goarch: arm64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/compiler
BenchmarkAllocation/Compile-8        291           3816010 ns/op         2522077 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        302           3618673 ns/op         2521508 B/op       1843 allocs/op
BenchmarkAllocation/Compile-8        331           3593862 ns/op         2521452 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        333           3576255 ns/op         2521385 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        331           3607171 ns/op         2521252 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        313           3735278 ns/op         2521359 B/op       1842 allocs/op
BenchmarkAllocation/Instantiate-8                   8986            152531 ns/op          331507 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   8193            156792 ns/op          331300 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   8977            159237 ns/op          331570 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   9030            162087 ns/op          331296 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   8905            152789 ns/op          331316 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   8966            153668 ns/op          331833 B/op        805 allocs/op
BenchmarkAllocation/Call-8                        809857              1483 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        813021              1478 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        813132              1604 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        816072              1478 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        814972              1482 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        808874              1483 ns/op              48 B/op          5 allocs/op

BenchmarkAllocationCGO/Compile-8                     331           3631028 ns/op         2514949 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     319           3671809 ns/op         2514931 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     327           3624131 ns/op         2514931 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     312           3713385 ns/op         2514908 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     333           3740863 ns/op         2514931 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     331           3624208 ns/op         2514945 B/op       1831 allocs/op
BenchmarkAllocationCGO/Instantiate-8                7207            177094 ns/op          326909 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                7368            184852 ns/op          326917 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                5750            174698 ns/op          326932 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                5533            211052 ns/op          326936 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                4027            262184 ns/op          326942 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                4050            278901 ns/op          329254 B/op        803 allocs/op
BenchmarkAllocationCGO/Call-8                     828213              1282 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     824278              1355 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     804750              1296 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     923024              1317 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     921345              1326 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     848708              1382 ns/op              64 B/op          7 allocs/op

BenchmarkAllocationTinyMem/Compile-8                 243           4741950 ns/op         2514947 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 229           5053305 ns/op         2514922 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 235           4934741 ns/op         2514907 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 229           4926298 ns/op         2514976 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 228           5335124 ns/op         2514889 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 225           5060070 ns/op         2514925 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            5203            224622 ns/op          326907 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4581            260680 ns/op          326916 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4768            241596 ns/op          326912 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4498            254726 ns/op          326919 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4207            255406 ns/op          326921 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4459            244556 ns/op          326934 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Call-8                 791721              1302 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 897240              1313 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 789856              1284 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 805621              1292 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 926949              1301 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 804992              1332 ns/op              64 B/op          7 allocs/op
PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/vs/compiler     281.555s
goos: darwin
goarch: arm64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/interpreter
BenchmarkAllocation/Compile-8        930           1313748 ns/op         1529469 B/op       1086 allocs/op
BenchmarkAllocation/Compile-8        894           1332110 ns/op         1529249 B/op       1086 allocs/op
BenchmarkAllocation/Compile-8        908           1300130 ns/op         1529055 B/op       1085 allocs/op
BenchmarkAllocation/Compile-8        873           1330047 ns/op         1529204 B/op       1085 allocs/op
BenchmarkAllocation/Compile-8        891           1299263 ns/op         1529536 B/op       1086 allocs/op
BenchmarkAllocation/Compile-8        927           1291028 ns/op         1529410 B/op       1086 allocs/op
BenchmarkAllocation/Instantiate-8                  10000            102160 ns/op          228271 B/op        748 allocs/op
BenchmarkAllocation/Instantiate-8                  10000            102371 ns/op          228268 B/op        748 allocs/op
BenchmarkAllocation/Instantiate-8                  10000            102485 ns/op          228269 B/op        748 allocs/op
BenchmarkAllocation/Instantiate-8                  10000            101532 ns/op          228262 B/op        748 allocs/op
BenchmarkAllocation/Instantiate-8                  10000            104661 ns/op          228266 B/op        748 allocs/op
BenchmarkAllocation/Instantiate-8                  10000            125214 ns/op          228261 B/op        748 allocs/op
BenchmarkAllocation/Call-8                         37723             32914 ns/op            2313 B/op         99 allocs/op
BenchmarkAllocation/Call-8                         37684             31981 ns/op            2307 B/op         99 allocs/op
BenchmarkAllocation/Call-8                         37664             32971 ns/op            2308 B/op         99 allocs/op
BenchmarkAllocation/Call-8                         37072             32771 ns/op            2313 B/op         99 allocs/op
BenchmarkAllocation/Call-8                         36356             32028 ns/op            2308 B/op         99 allocs/op
BenchmarkAllocation/Call-8                         37735             32050 ns/op            2312 B/op         99 allocs/op

BenchmarkAllocationCGO/Compile-8                     912           1292939 ns/op         1531150 B/op       1082 allocs/op
BenchmarkAllocationCGO/Compile-8                     926           1321487 ns/op         1531123 B/op       1083 allocs/op
BenchmarkAllocationCGO/Compile-8                     904           1353543 ns/op         1531141 B/op       1083 allocs/op
BenchmarkAllocationCGO/Compile-8                     886           1372992 ns/op         1531154 B/op       1082 allocs/op
BenchmarkAllocationCGO/Compile-8                     868           1331904 ns/op         1530869 B/op       1082 allocs/op
BenchmarkAllocationCGO/Compile-8                     934           1280476 ns/op         1531209 B/op       1083 allocs/op
BenchmarkAllocationCGO/Instantiate-8               10000            100332 ns/op          228139 B/op        747 allocs/op
BenchmarkAllocationCGO/Instantiate-8               10000            101137 ns/op          228144 B/op        747 allocs/op
BenchmarkAllocationCGO/Instantiate-8               12020             99502 ns/op          228145 B/op        747 allocs/op
BenchmarkAllocationCGO/Instantiate-8               10000            100263 ns/op          228136 B/op        747 allocs/op
BenchmarkAllocationCGO/Instantiate-8               10000            103855 ns/op          228126 B/op        747 allocs/op
BenchmarkAllocationCGO/Instantiate-8               12069            101501 ns/op          228146 B/op        747 allocs/op
BenchmarkAllocationCGO/Call-8                      38554             30020 ns/op            1767 B/op         77 allocs/op
BenchmarkAllocationCGO/Call-8                      39892             29189 ns/op            1764 B/op         77 allocs/op
BenchmarkAllocationCGO/Call-8                      43006             28127 ns/op            1765 B/op         77 allocs/op
BenchmarkAllocationCGO/Call-8                      39718             28627 ns/op            1766 B/op         77 allocs/op
BenchmarkAllocationCGO/Call-8                      43081             28666 ns/op            1764 B/op         77 allocs/op
BenchmarkAllocationCGO/Call-8                      41674             29142 ns/op            1767 B/op         77 allocs/op

BenchmarkAllocationTinyMem/Compile-8                 896           1298247 ns/op         1531165 B/op       1083 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 919           1298649 ns/op         1531050 B/op       1083 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 907           1300333 ns/op         1531412 B/op       1082 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 926           1296340 ns/op         1531361 B/op       1083 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 912           1305726 ns/op         1531433 B/op       1083 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 884           1351239 ns/op         1531302 B/op       1083 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8           10000            103462 ns/op          228143 B/op        747 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8           10000            102069 ns/op          228144 B/op        747 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8           10000            102167 ns/op          228135 B/op        747 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8           10000            101954 ns/op          228142 B/op        747 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8           10000            107825 ns/op          228134 B/op        747 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8           10000            105625 ns/op          228133 B/op        747 allocs/op
BenchmarkAllocationTinyMem/Call-8                  42105             28323 ns/op            1917 B/op         84 allocs/op
BenchmarkAllocationTinyMem/Call-8                  42680             28304 ns/op            1919 B/op         84 allocs/op
BenchmarkAllocationTinyMem/Call-8                  42364             28272 ns/op            1918 B/op         84 allocs/op
BenchmarkAllocationTinyMem/Call-8                  42566             28209 ns/op            1921 B/op         84 allocs/op
BenchmarkAllocationTinyMem/Call-8                  42592             28195 ns/op            1921 B/op         84 allocs/op
BenchmarkAllocationTinyMem/Call-8                  42763             29074 ns/op            1918 B/op         84 allocs/op
PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/vs/interpreter  73.635s
```

examples/allocation/tinygo/testdata/greet.go
```diff
@@ -53,7 +55,7 @@ func _greet(ptr, size uint32) {
 func _greeting(ptr, size uint32) (ptrSize uint64) {
```
codefromthecrypt, line 55: correct me if I'm wrong. It seems like in your bench of CGO vs tiny, both call into the same function? Did you mean to have a different guest wasm. e.g. one with tinymem approach which uses an alive pointers map and one that doesn't by using malloc free directly?

lburgazzoli: This is the "old" example, which I reverted to be the same as before, then I added a new example: `examples/allocation/tinygo-malloc` which exports `_greeting_cgo` and `_greeting_tinymem` 

Note: names must probably be changed to something more meaningful

examples/allocation/tinygo-malloc/testdata/greet.go
```diff
@@ -0,0 +1,122 @@
...
+
+// #include <stdlib.h>
+import "C"
+
+// main is required for TinyGo to compile to Wasm.
+func main() {}
+
+// greeting gets a greeting for the name.
+func greeting(name string) string {
+	return fmt.Sprint(name, "!")
```
codefromthecrypt, line 16: can you please change this to exactly the same as the other thing as having differences makes the benchmarks hard to assess

e.g.

```
// greeting gets a greeting for the name.
func greeting(name string) string {
	return fmt.Sprint("Hello, ", name, "!")
}
```

examples/allocation/tinygo-malloc/testdata/greet.go
```diff
@@ -0,0 +1,122 @@
...
+// Note:
+//   - This uses an uint64 instead of two result values for compatibility with
+//     WebAssembly 1.0.
+//   - The pointer returned by the function must be freed by the host using builtin free
+//     as it has been allocated using CGO malloc
+//
+//export greeting_cgo
+func _greeting_cgo(ptr, size uint32) (ptrSize uint64) {
+	name := ptrToString(ptr, size)
+	g := greeting("cgo: " + name)
```
codefromthecrypt, line 59: please don't add concatenation here.

examples/allocation/tinygo-malloc/testdata/greet.go
```diff
@@ -0,0 +1,122 @@
...
+// Note:
+//   - This uses an uint64 instead of two result values for compatibility with
+//     WebAssembly 1.0.
+//   - The pointer returned by the function must be freed by the host by using free_tinymem
+//     as it has been kept alive in a local map
+//
+//export greeting_tinymem
+func _greeting_tinymem(ptr, size uint32) (ptrSize uint64) {
+	name := ptrToString(ptr, size)
+	g := greeting("tinymem: " + name)
```
codefromthecrypt, line 95: please don't add concatenation here.

codefromthecrypt (commented):
thanks for pointing me to the right file!

Can you please make this exactly like the other (same length of string and concatenation approach), then re-run the bench and I'll take another look!

P.s. don't include interpreter output as that's not a performant env. Main thing we're looking for is using the same backend (compiler) what is the difference between these.

cheers!

examples/allocation/tinygo-malloc/testdata/greet.go
lburgazzoli: done

examples/allocation/tinygo-malloc/testdata/greet.go
lburgazzoli: done

examples/allocation/tinygo-malloc/testdata/greet.go
lburgazzoli: done

lburgazzoli:
Here the new bench:

```
➜ go test -run='^$' -bench '^BenchmarkAllocation.*' ./internal/integration_test/vs/compiler -count=6
goos: darwin
goarch: arm64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/compiler

BenchmarkAllocation/Compile-8        316           3734238 ns/op         2522058 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        314           3791311 ns/op         2521428 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        328           3681339 ns/op         2521260 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        320           3633035 ns/op         2522348 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        315           3751517 ns/op         2521278 B/op       1842 allocs/op
BenchmarkAllocation/Compile-8        325           3710645 ns/op         2521228 B/op       1842 allocs/op
BenchmarkAllocation/Instantiate-8                   8624            155833 ns/op          331516 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   7946            162025 ns/op          331298 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   8402            157819 ns/op          331590 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   8947            164243 ns/op          331292 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   8954            157033 ns/op          331314 B/op        805 allocs/op
BenchmarkAllocation/Instantiate-8                   7219            176556 ns/op          331969 B/op        805 allocs/op
BenchmarkAllocation/Call-8                        815618              1669 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        804968              1536 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        816934              1539 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        799848              1545 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        815667              1535 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-8                        747400              1532 ns/op              48 B/op          5 allocs/op

BenchmarkAllocationCGO/Compile-8                     298           3931316 ns/op         2514327 B/op       1832 allocs/op
BenchmarkAllocationCGO/Compile-8                     316           3923335 ns/op         2514323 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     306           3751290 ns/op         2514284 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     292           3731128 ns/op         2514301 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     309           3760348 ns/op         2514269 B/op       1831 allocs/op
BenchmarkAllocationCGO/Compile-8                     313           3874987 ns/op         2514300 B/op       1831 allocs/op
BenchmarkAllocationCGO/Instantiate-8                6733            209859 ns/op          326912 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                5056            242084 ns/op          326914 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                4530            260508 ns/op          326921 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                3813            274212 ns/op          326932 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                3998            252612 ns/op          326938 B/op        803 allocs/op
BenchmarkAllocationCGO/Instantiate-8                4941            236459 ns/op          326944 B/op        803 allocs/op
BenchmarkAllocationCGO/Call-8                     911188              1460 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     796794              1349 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     771747              1330 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     818712              1312 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     807490              1299 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-8                     802276              1310 ns/op              64 B/op          7 allocs/op

BenchmarkAllocationTinyMem/Compile-8                 238           5088182 ns/op         2554111 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 212           5165166 ns/op         2514299 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 222           5127355 ns/op         2514317 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 224           5164966 ns/op         2514299 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 228           4899677 ns/op         2514238 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Compile-8                 230           5132746 ns/op         2514305 B/op       1831 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4591            240156 ns/op          326915 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4712            240882 ns/op          326908 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            3678            273542 ns/op          326911 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4569            240208 ns/op          326915 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4210            281070 ns/op          326908 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Instantiate-8            4215            264439 ns/op          326917 B/op        803 allocs/op
BenchmarkAllocationTinyMem/Call-8                 912447              1333 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 800275              1291 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 833378              1292 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 779216              1330 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 732708              1387 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-8                 812904              1296 ns/op              64 B/op          7 allocs/op

PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/vs/compiler     278.537s
```

codefromthecrypt:
ok one last thing I didn't notice before asking others to analyze. we have a another place we've introduced variables.

Please revert examples/allocation/tinygo changes and run again. We can argue that even if you "should" free after use, most code in practice doesn't. Let's separate the topics, so people can know what sort of overhead if any a change here will cost vs *what they do now*.

lburgazzoli:
**Bench**
```
➜ go test -run='^$' -bench '^BenchmarkAllocation.*' ./internal/integration_test/vs/compiler -count=6
goos: linux
goarch: amd64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/compiler
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
```

**Leaking**
```
BenchmarkAllocation/Compile-12                72          26800928 ns/op         2957281 B/op       2022 allocs/op
BenchmarkAllocation/Compile-12                80          19399234 ns/op         2953547 B/op       2018 allocs/op
BenchmarkAllocation/Compile-12                81          17107695 ns/op         2953098 B/op       2018 allocs/op
BenchmarkAllocation/Compile-12                70          15503359 ns/op         2952115 B/op       2019 allocs/op
BenchmarkAllocation/Compile-12                81          15342295 ns/op         2953347 B/op       2019 allocs/op
BenchmarkAllocation/Compile-12                79          15211598 ns/op         2952159 B/op       2018 allocs/op
BenchmarkAllocation/Instantiate-12                  3180            336051 ns/op          361222 B/op        843 allocs/op
BenchmarkAllocation/Instantiate-12                  3792            353417 ns/op          361020 B/op        843 allocs/op
BenchmarkAllocation/Instantiate-12                  3765            356393 ns/op          361355 B/op        843 allocs/op
BenchmarkAllocation/Instantiate-12                  3646            366901 ns/op          361020 B/op        843 allocs/op
BenchmarkAllocation/Instantiate-12                  3788            360638 ns/op          361032 B/op        843 allocs/op
BenchmarkAllocation/Instantiate-12                  3774            359608 ns/op          361044 B/op        843 allocs/op
BenchmarkAllocation/Call-12                       358167              3330 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12                       370910              3275 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12                       362886              3211 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12                       371457              3265 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12                       372108              3218 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12                       365642              3238 ns/op              48 B/op          5 allocs/op
```

**CGO**
```
BenchmarkAllocationCGO/Compile-12                     79          14928299 ns/op         2949799 B/op       2027 allocs/op
BenchmarkAllocationCGO/Compile-12                     82          15034444 ns/op         2949814 B/op       2027 allocs/op
BenchmarkAllocationCGO/Compile-12                     82          15239788 ns/op         2949923 B/op       2027 allocs/op
BenchmarkAllocationCGO/Compile-12                     75          14733278 ns/op         2981526 B/op       2028 allocs/op
BenchmarkAllocationCGO/Compile-12                     82          14734207 ns/op         2949853 B/op       2027 allocs/op
BenchmarkAllocationCGO/Compile-12                     82          14820402 ns/op         2949784 B/op       2027 allocs/op
BenchmarkAllocationCGO/Instantiate-12               3426            412632 ns/op          361006 B/op        843 allocs/op
BenchmarkAllocationCGO/Instantiate-12               3505            379676 ns/op          361017 B/op        843 allocs/op
BenchmarkAllocationCGO/Instantiate-12               3547            371487 ns/op          361017 B/op        843 allocs/op
BenchmarkAllocationCGO/Instantiate-12               3566            360025 ns/op          361020 B/op        843 allocs/op
BenchmarkAllocationCGO/Instantiate-12               3525            368470 ns/op          361040 B/op        843 allocs/op
BenchmarkAllocationCGO/Instantiate-12               3514            359207 ns/op          361034 B/op        843 allocs/op
BenchmarkAllocationCGO/Call-12                    414781              2741 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12                    434466              2775 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12                    429614              2749 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12                    439882              2746 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12                    439764              2730 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12                    431773              2736 ns/op              64 B/op          7 allocs/op
```

**TinyMem**
```
BenchmarkAllocationTinyMem/Compile-12                 81          14882575 ns/op         2949935 B/op       2027 allocs/op
BenchmarkAllocationTinyMem/Compile-12                 81          14708287 ns/op         2949819 B/op       2027 allocs/op
BenchmarkAllocationTinyMem/Compile-12                 81          14384014 ns/op         2949824 B/op       2027 allocs/op
BenchmarkAllocationTinyMem/Compile-12                 81          14227781 ns/op         2949882 B/op       2028 allocs/op
BenchmarkAllocationTinyMem/Compile-12                 81          14247338 ns/op         2949872 B/op       2027 allocs/op
BenchmarkAllocationTinyMem/Compile-12                 81          14245075 ns/op         2949883 B/op       2027 allocs/op
BenchmarkAllocationTinyMem/Instantiate-12           3792            394353 ns/op          362300 B/op        843 allocs/op
BenchmarkAllocationTinyMem/Instantiate-12           3421            375745 ns/op          361017 B/op        843 allocs/op
BenchmarkAllocationTinyMem/Instantiate-12           3312            380304 ns/op          361025 B/op        843 allocs/op
BenchmarkAllocationTinyMem/Instantiate-12           3486            364910 ns/op          361009 B/op        843 allocs/op
BenchmarkAllocationTinyMem/Instantiate-12           3474            375216 ns/op          361020 B/op        843 allocs/op
BenchmarkAllocationTinyMem/Instantiate-12           3448            374584 ns/op          361017 B/op        843 allocs/op
BenchmarkAllocationTinyMem/Call-12                415803              2819 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-12                435906              2759 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-12                433790              2768 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-12                434032              2771 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-12                432223              2759 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationTinyMem/Call-12                436834              2846 ns/op              64 B/op          7 allocs/op
```

codefromthecrypt:
thanks tons @lburgazzoli looks excellent. So we were pinged about our docs on the site which suggest something "out of date" by tinygo folks, whatever the current recommendation is (CGO) can you verify which with them and also update the site?

I'm fine with either approach, and think you did the right diligence to show the leaking way isn't saving anything. Between the two others (use tinygo's CGO or don't (use tinymem)) I'm fine.

I'm re-pinging @ncruces and adding @inkeliz for a final say on this

lburgazzoli:
> thanks tons @lburgazzoli looks excellent. So we were pinged about our docs on the site which suggest something "out of date" by tinygo folks, whatever the current recommendation is (CGO) can you verify which with them and also update the site?
> 

I will

lburgazzoli:
@codefromthecrypt should I squash the commits to remove some noise form the review ?

codefromthecrypt:
so if the recommendation is to use either CGO or tinymem, what I would do is delete the old allocation tinymem and keep your two. If the recommendation is one or the other, I'd suggest replace it. So, you can squash for now, then when a decision is made on that topic, adjust accordingly!

codefromthecrypt:
so I'm guessing from lack of info contrary, that the preferred approach is the CGO one. feel free to remove the other approaches and squash for review!

lburgazzoli:
> so I'm guessing from lack of info contrary, that the preferred approach is the CGO one. feel free to remove the other approaches and squash for review!

to be clear, should the original allocation example/bench be kept ?

codefromthecrypt:
> to be clear, should the original allocation example/bench be kept ?

right any tests or bench there before should stay, just the tinygo impl should change to match CGO approach. Thanks!

lburgazzoli:
@codefromthecrypt I've adapted the allocation example to use CGO for the `greeting` function and squashed the commits. 

I kept the `BenchmarkAllocationCGO` so there is one bench that runs against `greet` and one against `greeting`:
 

```
➜ go test -run='^$' -bench '^BenchmarkAllocation.*' ./internal/integration_test/vs/compiler -count=6
goos: linux
goarch: amd64
pkg: github.com/tetratelabs/wazero/internal/integration_test/vs/compiler
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz

BenchmarkAllocation/Compile-12                73          32071839 ns/op         2958373 B/op       2027 allocs/op
BenchmarkAllocation/Compile-12                73          14590089 ns/op         2953361 B/op       2022 allocs/op
BenchmarkAllocation/Compile-12                81          15130377 ns/op         2953334 B/op       2022 allocs/op
BenchmarkAllocation/Compile-12                80          14616454 ns/op         2953033 B/op       2022 allocs/op
BenchmarkAllocation/Compile-12                84          14875860 ns/op         2952820 B/op       2021 allocs/op
BenchmarkAllocation/Compile-12                79          15151539 ns/op         2952698 B/op       2021 allocs/op
BenchmarkAllocation/Instantiate-12          3987            338833 ns/op          359861 B/op        852 allocs/op
BenchmarkAllocation/Instantiate-12          3866            347517 ns/op          359903 B/op        852 allocs/op
BenchmarkAllocation/Instantiate-12          3709            362471 ns/op          359768 B/op        852 allocs/op
BenchmarkAllocation/Instantiate-12          3744            357544 ns/op          359742 B/op        852 allocs/op
BenchmarkAllocation/Instantiate-12          3633            362594 ns/op          359736 B/op        852 allocs/op
BenchmarkAllocation/Instantiate-12          3699            368921 ns/op          359766 B/op        852 allocs/op
BenchmarkAllocation/Call-12               342232              3376 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12               362686              3294 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12               366472              3247 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12               348795              3282 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12               367167              3263 ns/op              48 B/op          5 allocs/op
BenchmarkAllocation/Call-12               343881              3523 ns/op              48 B/op          5 allocs/op

BenchmarkAllocationCGO/Compile-12             79          14663107 ns/op         2952708 B/op       2022 allocs/op
BenchmarkAllocationCGO/Compile-12             80          15036263 ns/op         2952842 B/op       2021 allocs/op
BenchmarkAllocationCGO/Compile-12             81          14828323 ns/op         2952739 B/op       2022 allocs/op
BenchmarkAllocationCGO/Compile-12             81          14646740 ns/op         2952697 B/op       2021 allocs/op
BenchmarkAllocationCGO/Compile-12             80          14670556 ns/op         2952770 B/op       2022 allocs/op
BenchmarkAllocationCGO/Compile-12             81          14636888 ns/op         2952636 B/op       2021 allocs/op
BenchmarkAllocationCGO/Instantiate-12       3385            417383 ns/op          360447 B/op        852 allocs/op
BenchmarkAllocationCGO/Instantiate-12       3056            377058 ns/op          359731 B/op        852 allocs/op
BenchmarkAllocationCGO/Instantiate-12       3600            368788 ns/op          359733 B/op        852 allocs/op
BenchmarkAllocationCGO/Instantiate-12       3544            376303 ns/op          359736 B/op        852 allocs/op
BenchmarkAllocationCGO/Instantiate-12       3094            362826 ns/op          359752 B/op        852 allocs/op
BenchmarkAllocationCGO/Instantiate-12       3470            362647 ns/op          359761 B/op        852 allocs/op
BenchmarkAllocationCGO/Call-12            413808              2770 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12            436496              2749 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12            427888              2754 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12            426919              2745 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12            435013              2769 ns/op              64 B/op          7 allocs/op
BenchmarkAllocationCGO/Call-12            436152              2771 ns/op              64 B/op          7 allocs/op

PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/vs/compiler     68.776s
```

Let me know if this matches what you had in mind.

internal/integration_test/vs/bench_allocation_cgo.go
```diff
@@ -0,0 +1,89 @@
+package vs
```
codefromthecrypt, line 1: thanks, but no thanks on this, as these benchmarks run on every change. If we are going to spend on this it should be a notably different wasm use case, not a slight variant of an existing one.

examples/allocation/tinygo/testdata/greet.go
codefromthecrypt: please revert this name, but keep the comment. It is still wasm memory, so it isn't native. I think the change is actually more confusing than before.

internal/integration_test/vs/runtime.go
```diff
@@ -37,6 +37,7 @@ type Runtime interface {
 type Module interface {
 	CallI32_I32(ctx context.Context, funcName string, param uint32) (uint32, error)
 	CallI32I32_V(ctx context.Context, funcName string, x, y uint32) error
+	CallI32I32_I64(ctx context.Context, funcName string, x, y uint32) (uint64, error)
```
codefromthecrypt, line 40: revert this and the supporting changes if only used for the CGO benchmark.

internal/integration_test/vs/wasmedge/wasmedge_test.go
```diff
@@ -24,6 +24,30 @@ func TestBenchmarkAllocation_Call_CompilerFastest(t *testing.T) {
 	vs.RunTestBenchmarkAllocation_Call_CompilerFastest(t, runtime())
 }
 
+func TestAllocationCGO(t *testing.T) {
```
codefromthecrypt, line 27: please keep only the existing Test* Benchmark* functions in the vs package. drop the ones used for the investigation

codefromthecrypt (commented):
Thanks for the help again. I made some requests to pare down the change to the minimum. Don't forget to update the guidance in the site directory for tinygo!

examples/allocation/tinygo/testdata/greet.go
codefromthecrypt: the comment ended up on the wrong line somehow. reposting it.

internal/integration_test/vs/wasmedge/wasmedge_test.go
lburgazzoli: Ah ok, so you don't want any of the cgo things to be benchamrked, correct ?

examples/allocation/tinygo/testdata/greet.go
```diff
@@ -53,7 +56,7 @@ func _greet(ptr, size uint32) {
...
 	name := ptrToString(ptr, size)
 	g := greeting(name)
-	ptr, size = stringToPtr(g)
+	ptr, size = stringToNativePtr(g)
```
codefromthecrypt, line 59: so I mistook this as a rename, when it is a second function. Go doesn't have a way to name this, I'm aware of, but I think the main thing is to not say "native".

Maybe `stringToLeakedPtr` instead, as the essential thing here is that the remote caller is taking ownership of the pointer. At least our rust example calls this leak, so it is somewhat consistent.

examples/allocation/tinygo/testdata/greet.go
```diff
@@ -77,3 +80,19 @@ func stringToPtr(s string) (uint32, uint32) {
 	unsafePtr := uintptr(unsafe.Pointer(ptr))
 	return uint32(unsafePtr), uint32(len(buf))
 }
+
+// stringToNativePtr returns a pointer and size pair for the given string in a way
+// compatible with WebAssembly numeric types. The pointer is not automatically
+// managed by TinyGo hence it must be freed by the host.
+func stringToNativePtr(s string) (uint32, uint32) {
+	if len(s) == 0 {
+		return 0, 0
```
codefromthecrypt, line 89: Is this special case valid? if so add a trailing comment why (for example,`// CGO.free ignores a pointer of zero`)
If not, delete the special case.

codefromthecrypt (commented):
more notes as my prior ended up in the wrong place. Thanks again!

internal/integration_test/vs/wasmedge/wasmedge_test.go
codefromthecrypt: we don't need to add new benchmarks (in vs). we did ones in the interim to make sure nothing regressed with the new advice, that's all.

codefromthecrypt:
ps if you feel like benchmarking "greeting" is better than "greet", feel free to switch. What I mean is let's keep only one for the allocation example. I don't care strongly which.

examples/allocation/tinygo/testdata/greet.go
lburgazzoli: As far as I recall, calling `malloc` with a size equals to 0, is implementation dependent hence I opted not to call it in such case. 

I Will add a note if that's fine with you but in this case it is probably never happening.

codefromthecrypt: This is an example, so we can't really add suggestions to special case zero unless it will work. We don't have to consider externally provided impls of malloc/free in the example, though. What tinygo uses with this example and flags given to build is enough.

@anuraaga @dgryski do either of you happen to know what's better:

* not dodging calling malloc dependent on length
* dodging calling malloc with zero length (results in free called with zero)

codefromthecrypt: here's the tinygo code in question. Personally, I prefer to not special case without saying why we are doing it. Most people won't know why they are calling malloc/free in the first place and code like this gets copy/pasted around. As best I can tell, we are assuming the result will be zero and so we are returning that instead of calling things. That would be a perf optimization, which we'd comment why.

If we don't have a good reason tied to tinygo, I prefer we don't do any special casing in this function. Without comments, special casing raises more questions than answers, especially that a zero pointer is something safe to return back to tinygo on free. For example, pointers are often memory offsets in wasm, so zero would be the initial memory position. While free shouldn't have any side effects there and it is a safe guess maybe to assume it will do nothing on free.. it is a lot better to leave the problem on what to do with length zero completely up to tinygo (e.g. don't special case). Cheaper to review, less code, easier example.

https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/src/runtime/arch_tinygowasm_malloc.go#L13-L34

examples/allocation/tinygo/testdata/greet.go
lburgazzoli: done

internal/integration_test/vs/wasmedge/wasmedge_test.go
lburgazzoli: done

examples/allocation/tinygo/testdata/greet.go
lburgazzoli: done

examples/allocation/tinygo/testdata/greet.go
lburgazzoli: remove d the special case

internal/integration_test/vs/runtime.go
lburgazzoli: done

internal/integration_test/vs/bench_allocation_cgo.go
lburgazzoli: done

lburgazzoli:
> ps if you feel like benchmarking "greeting" is better than "greet", feel free to switch. What I mean is let's keep only one for the allocation example. I don't care strongly which.

I don't have any strong opinion, so let's keep the benchmark as it is, it wont be too complex to add a new one when/if it would become a need.

codefromthecrypt (approved):
Thanks, I'll go ahead and merge this and do the site updates on own vs asking again here. Appreciate your help getting the underlying solution together and benchmarking to ensure it didn't cause a regression.

lburgazzoli:
I can try to update the docs before EOW if you want me to.

ncruces:
Epic effort for something that started out with improving example code.
Thanks, and really sorry if I kinda lost track along the way!

codefromthecrypt:
@lburgazzoli I raised https://github.com/tetratelabs/wazero/pull/1429 as I think we should get this sorted to avoid a confusing comment chain over a week again. This sort of stuff is cognitively deep and best way I can find is to contain it by time to shortest as possible, as this avoids things like having to cross when people are out for holidays etc.

In general, personally I try to complete everything PR of any kind within a couple days for the same reason, even if not cognitively high. Even small things can get difficult to load back into the brain if drug out over several days, much less weeks, with several people.

ncruces:
@lburgazzoli sorry for just spotting this, I guess I only grasped it when looking at committed code in whole.
But I might be missing something obvious and I'd like to get your understanding before filing an issue/PR.

What keeps `message` alive (i.e. not available for GC) after `stringToPtr` returns?
https://github.com/tetratelabs/wazero/blob/9dd8b1b9b3bd9267502440fea1fb2bacc577c58c/examples/allocation/tinygo/testdata/greet.go#L20-L24

I'm wondering if we need to add `runtime.KeepAlive(message)` after `_log`.

codefromthecrypt:
I'll update tinymem once above is resolved. We do tend to see things like this in syscall code, after the call like https://github.com/golang/go/blob/0d347544cbca0f42b160424f6bc2458ebcc7b3fc/src/syscall/fs_wasip1.go#L824-L829 cc @Pryz @achille-roussel

lburgazzoli:
> @lburgazzoli sorry for just spotting this, I guess I only grasped it when looking at committed code in whole. But I might be missing something obvious and I'd like to get your understanding before filing an issue/PR.
> 
> What keeps `message` alive (i.e. not available for GC) after `stringToPtr` returns?
> 

I completely missed that usage to be honest.

> https://github.com/tetratelabs/wazero/blob/9dd8b1b9b3bd9267502440fea1fb2bacc577c58c/examples/allocation/tinygo/testdata/greet.go#L20-L24
> 
> I'm wondering if we need to add `runtime.KeepAlive(message)` after `_log`.

I think the usage of `message` is safe, however the implementation of `stringToPtr` is not as it performs a [copy](https://github.com/tetratelabs/wazero/blob/9dd8b1b9b3bd9267502440fea1fb2bacc577c58c/examples/allocation/tinygo/testdata/greet.go#L78) of the string's data but that slice goes out of scope hence the address returned by the function may become invalid if a GC is triggered. 

I think the only safest option would be to use `stringToLeakedPtr` ad free up memory with `defer`, similar to what it is done in the host code.

lburgazzoli:
> `KeepAlive`

A more unsafe way would be to `runtime.KeepAlive` the `message` and return the actual data that backs the string with `unsafe.StringData`, but there will be dragons :)

ncruces:
Yes, that too. The only additional unsafety to that it's supposed to be read-only, and the host can change it, but the host can always change it, so…

codefromthecrypt:
> The only additional unsafety to that it's supposed to be read-only, and the host can change it, but the host can always change it, so…

agree the host can always change anything. Doing anything in the guest that makes it seem like it cannot is probably not a good idea. Like adding racing stripes to a Honda, doesn't make it faster. We should make the guest code concise, cheap as possible and most importantly relevant. If the current code isn't really leaking it, or it is more complicated than an alternative to do the same. PR welcome, especially before we communicate widely a change.

PR #1391, opened 2023-04-21
fs: returns EBADF on negative file descriptor

codefromthecrypt:
This changes file descriptors from uint32 to int32 and the corresponding file table to reject negative values. This ensures invalid values aren't mistaken for very large descriptor entries, which can use a lot of memory as the table implementation isn't designed to be sparse.

Thanks to @jerbob92 for noticing this!

See https://pubs.opengroup.org/onlinepubs/9699919799/functions/dirfd.html#tag_16_90

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -39,7 +39,7 @@ var fdAdvise = newHostFunc(
 )
 
 func fdAdviseFn(_ context.Context, mod api.Module, params []uint64) syscall.Errno {
-	fd := uint32(params[0])
+	fd := int32(params[0])
 	_ = params[1]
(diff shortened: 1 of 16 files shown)


achille-roussel (approved):
Looking good 👍 

Regarding the attack vector of using fd_renumber with a very large value, we could add a "max open file" limit which then defines the maximum value that a file descriptor can take?

codefromthecrypt:
> Regarding the attack vector of using fd_renumber with a very large value, we could add a "max open file" limit which then defines the maximum value that a file descriptor can take?

sounds like a nice follow-up when we are ready to do a minor.. maybe `FSConfig.WithMaxFiles(int32)`? Then change the table to guard on that (EBADF when over or something else..)

PR #1395, opened 2023-04-24
asm/amd64: removes enablePadding flag

mathetake:
this is a vestige from the golang-asm era


codefromthecrypt (commented):
thanks for tech debt cleanup!

PR #1396, opened 2023-04-24
Proposal: "close" certain api interfaces for external implementation

ncruces:
As discussed in https://github.com/tetratelabs/wazero/pull/1332#issuecomment-1510010359, many of the interfaces in [`github.com/tetratelabs/wazero`](https://pkg.go.dev/github.com/tetratelabs/wazero) and [`github.com/tetratelabs/wazero/api`](https://pkg.go.dev/github.com/tetratelabs/wazero/api) are not meant to be externally implemented.

In strict terms, an interface can't be extended after it has been made public. Once made public, alternative implementations may be created, and adding a method to such an interface, means those implementations no longer compile, which breaks semver.

wazero 1.0 shipped with many such interfaces, which we may want to extend (e.g. #1332).

These are the paths available to us:
1. **that ship has sailed**: wazero 1.0 shipped, we simply can't extend these interfaces;
2. **documentation**: document *now* that certain interfaces are not meant to be implemented outside wazero;
3. **this PR**: "enforce" this, avoiding users implementing them by accident.

Option **1** is basically doing nothing. In the context of #1332 I can propose a different alternative, which will basically be a new `api.Function2` (name not important) interface that wraps the [`CallWithStack`](https://github.com/tetratelabs/wazero/pull/1332#issuecomment-1507793809) method. Then users either type-assert this interface, or we create new builders returning this interface.

Options **2** and **3** require deciding which interfaces are only for internal implementation.

Option **2** is a documentation change to [`github.com/tetratelabs/wazero`](https://pkg.go.dev/github.com/tetratelabs/wazero) and [`github.com/tetratelabs/wazero/api`](https://pkg.go.dev/github.com/tetratelabs/wazero/api), and to each such interface.

Option **3** is this PR, repeated for every such interface.

In the documentation of each of these interfaces, you'll [see](https://pkg.go.dev/go/types#Object):
```go
type Interface interface {
	PublicMethod() // public method
	// contains filtered or unexported methods
}
```

When you try to implement such an interface (create your own type and use it as if it implemented it), you get this error message:
```
cannot use &myCallEngine{…} (value of type *myCallEngine) as api.Function value in return statement: *myCallEngine does not implement api.Function (missing method wazeroOnly)
```

If you're doing it for mocking, which is the only way you'd do it for these wazero types, you can still deliberately sidestep the check by embedding the interface:
```go
type mockCallEngine struct {
	index         Index
	callFailIndex int
	api.Function
}
```

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"fmt"
 	"math"
+
+	"github.com/tetratelabs/wazero/internal/internalapi"
 )
 
 // ExternType classifies imports and exports with their respective types.
@@ -188,7 +190,7 @@ type Module interface {
 	// Closer closes this module by delegating to CloseWithExitCode with an exit code of zero.
 	Closer
 
-	wazeroOnly()
+	internalapi.WazeroOnly
 }
 
 // Closer closes a resource.
@@ -239,7 +241,7 @@ type ExportDefinition interface {
 	// so "" is possible.
 	ExportNames() []string
 
-	wazeroOnly()
+	internalapi.WazeroOnly
 }
 
 // MemoryDefinition is a WebAssembly memory exported in a module
@@ -261,7 +263,7 @@ type MemoryDefinition interface {
 	// unbounded.
 	Max() (uint32, bool)
 
-	wazeroOnly()
+	internalapi.WazeroOnly
 }
 
 // FunctionDefinition is a WebAssembly function exported in a module
@@ -325,7 +327,7 @@ type FunctionDefinition interface {
 	// available for one or more results.
 	ResultNames() []string
 
-	wazeroOnly()
+	internalapi.WazeroOnly
 }
 
 // Function is a WebAssembly function exported from an instantiated module
@@ -364,7 +366,7 @@ type Function interface {
 	// the end-to-end demonstrations of how these terminations can be performed.
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
 
-	wazeroOnly()
+	internalapi.WazeroOnly
 }
 
 // GoModuleFunction is a Function implemented in Go instead of a wasm binary.
@@ -493,7 +495,7 @@ type MutableGlobal interface {
 	// See Global.Type for how to encode this value from a Go type.
 	Set(v uint64)
 
-	wazeroOnly()
+	internalapi.WazeroOnly
 }
 
 // Memory allows restricted access to a module's memory. Notably, this does not allow growing.
@@ -620,7 +622,7 @@ type Memory interface {
 	// WriteString writes the string to the underlying buffer at the offset or returns false if out of range.
 	WriteString(offset uint32, v string) bool
 
-	wazeroOnly()
+	internalapi.WazeroOnly
 }
 
 // CustomSection contains the name and raw data of a custom section.
@@ -635,7 +637,7 @@ type CustomSection interface {
 	// Data is the raw data of the custom section
 	Data() []byte
 
-	wazeroOnly()
+	internalapi.WazeroOnly
 }
 
 // EncodeExternref encodes the input as a ValueTypeExternref.
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -355,6 +355,8 @@ type Function interface {
 	// WithCloseOnContextDone on wazero.RuntimeConfig for detail. See examples in context_done_example_test.go for
 	// the end-to-end demonstrations of how these terminations can be performed.
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
+
+	private()
 }
 
 // GoModuleFunction is a Function implemented in Go instead of a wasm binary.
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -187,6 +187,8 @@ type Module interface {
 
 	// Closer closes this module by delegating to CloseWithExitCode with an exit code of zero.
 	Closer
+
+	wazeroOnly()
 }
 
 // Closer closes a resource.
@@ -236,6 +238,8 @@ type ExportDefinition interface {
 	// Note: The empty name is allowed in the WebAssembly Core Specification,
 	// so "" is possible.
 	ExportNames() []string
+
+	wazeroOnly()
 }
 
 // MemoryDefinition is a WebAssembly memory exported in a module
@@ -256,6 +260,8 @@ type MemoryDefinition interface {
 	// Max returns the possibly zero max count of 64KB pages, or false if
 	// unbounded.
 	Max() (uint32, bool)
+
+	wazeroOnly()
 }
 
 // FunctionDefinition is a WebAssembly function exported in a module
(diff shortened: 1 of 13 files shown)


codefromthecrypt:
Below should have been added to all interfaces except those that require the users to implement them. What happened was we did this to the top-level ones and didn't go recursively down when we should have.

```
// # Notes
//
//   - This is an interface for decoupling, not third-party implementations. All implementations are in wazero.
```

We should not escalate this doc bug as a perpetual promise. The only way someone could even use api.Function would be to wrap, and that would be a hard job as they'd also have to wrap what it returns. I suspect there are max one or two people who could have done this and we can apologize to them individually for the doc bug. If this includes you, please accept my apology!


Knowing this, here's my 2p

*  I mentioned to someone I think about the missing docs, but I should have done them myself to make sure they landed in the last release. I'll do that now, so that it remain any longer.
* I'm fine also adding the private thingy, but not as an alternative to docs. This should be in addition to docs.
* Unlike api.GoModuleFunction, api.Function was not documented for implementation. Feel free to add the function you need, I think I mentioned this offline a while but if I hadn't apologies!

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/1397

Sorry, I should have done this immediately when it was brought up last, as it would have been in 1.0.3, then. I often leave work that is easy to do for others, as a way for folks to up stats etc. This is one of those things I should have done right away.

api/wasm.go
```diff
@@ -329,6 +329,8 @@ type Function interface {
 	// WithCloseOnContextDone on wazero.RuntimeConfig for detail. See examples in context_done_example_test.go for
 	// the end-to-end demonstrations of how these terminations can be performed.
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
+
+	private()
```
codefromthecrypt, line 333: so I think maybe I understand this. Basically, this doesn't stop external implementation, rather it slows it. So people need to do embed as we do in other parts of the PR.

If others aren't against this approach, I'm fine with it. We'd boilerplate docs. We can even be more explicit in the name if we want!

```suggestion
	// wazeroOnly means this should only be implemented by wazero.
	wazeroOnly()
```

ncruces: Exactly, agree, and `wazeroOnly` is a good name.

achille-roussel:
I thought it might be worth chiming in with some examples of how the interfaces can be useful as extension mechanisms. I'm guilty of abusing the documentation stating that they should not be implemented outside of wazero, but there's an interesting use case in wazergo I think https://github.com/stealthrocket/wazergo/blob/main/module.go#L211-L222

The `wazergo.ModuleInstance[T]` type implements `api.Module` so when accessing exported functions, the context needed to invoke methods of the host module can be injected automatically.

I'm calling it out for two reasons:
- if the documented rules are not enforced, they will likely continue to be abused (by me probably, but it's unlikely that I will remain the only offender)
- there may be useful extension opportunities using the interfaces defined in wazero, maybe a follow up could be to start a discussion around how/where it makes sense for wazero to allow applications to leverage interfaces for customization

When this PR gets merged it will break wazergo so I'll follow up with a fix in that repository to address.

ncruces:
You're for sure not the only one doing it, @achille-roussel. I [have](https://github.com/ncruces/go-sqlite3/blob/main/internal/util/mock.go) `mockModule` and `mockMemory` for testing.

The major difference is that (my) mocks don't ship downstream, so it's not an issue if my clients update `wazero` to a version newer than the one I used

OTOH, `wazergo` might get broken if users update `wazero` beyond the one declared in `wazergo` `go.mod`, which is an issue I don't really know to fix, except by saying `wazero` **can't** ever update these public interfaces.

achille-roussel:
wazergo is not in 1.0 and the README calls out that some breaking changes may be needed for things like "design flaws" so I think it's OK at this stage.

codefromthecrypt:
FWIW if there's a secret way to "use at your own risk" I think it is fair to try not to break vs intentionally break.

in any case, the break would be adding new methods, not removing or changing existing ones. So, this means we would only break those who are sneakily wrapping. OTOH this means those wrapping can never be non-experimental or we can decide a version that can't add either.

For example, we can say each even version (e.g. 2.0) is super LTS meaning no functions added either. I don't think we should do that on 1.0 though.

internal/wasm/memory.go
```diff
@@ -23,9 +23,6 @@ const (
 	MemoryPageSizeInBits = 16
 )
 
-// compile-time check to ensure MemoryInstance implements api.Memory
-var _ api.Memory = &MemoryInstance{}
```
ncruces, line 27: These checks become useless. 😔

internal/wasm/global.go
```diff
@@ -8,10 +8,9 @@ import (
 
 type mutableGlobal struct {
 	g *GlobalInstance
-}
 
-// compile-time check to ensure mutableGlobal is a api.Global.
-var _ api.Global = &mutableGlobal{}
+	api.MutableGlobal
```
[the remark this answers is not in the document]
achille-roussel: Maybe a different approach could be used:

In an internal package
```go
package internalpackage 

type WazeroOnly interface {
  wazeroOnly()
}

type WazeroOnlyType struct{}

func (WazeroOnlyType) wazeroOnly() {}
```

Then use the concrete type and interface:
```go
package api

type Global interface {
  ...
  internalpackage.WazeroOnly // embed the interface
}
```
```go
type mutableGlobal struct {
  internalpackage.WazeroOnlyType // empty struct as first field so it uses zero bytes
  ...
}
```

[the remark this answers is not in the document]
ncruces: You're right. I'll try your approach. But I agree, if the runtime cost is unavoidable, this is a bad idea. Documentation is enough.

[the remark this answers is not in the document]
ncruces: This is a much better approach, thanks @achille-roussel!
For the structures I tested (with `unsafe.SizeOf`) it's indifferent whether the empty struct is the first or last field.

We can now bikeshed the name. 😄

codefromthecrypt:
Funny or not, I'm fine with the name. Looking good!

Since there was some chatter on this, I think when ready a RATIONALE.md update is warranted. you can see some other notes about things we do that may be a bit strange in there. One note to put into this is that we are only adding these to "api" (as opposed to all non-internal interfaces) because "api" is most likely what users would expect to implement vs just consume.

internal/wasm/memory.go
ncruces: Not anymore, thanks to @achille-roussel.

ncruces:
About only doing it for `api`, I started out that way out of laziness, TBH.
It wasn't meant to be a decision, though it can be.

I _also_ skipped `api.Global` (again out of laziness) because that one has non-struct implementations:
https://github.com/tetratelabs/wazero/blob/c82ad896f6019708f22ddd7826ff47319b7a1e54/internal/wasm/global.go#L45-L48

codefromthecrypt:
FWI I'm fine with this not being 100pct coverage. The main types will convey the intent well enough ™️

codefromthecrypt (approved):

ncruces:
OK. If folks think I should do another round for types in the main `wazero` package, please say so.

codefromthecrypt:
Well done, and I've already re-used your approach in another project ;) https://github.com/tetratelabs/car/blob/master/internal/internal.go#L19

codefromthecrypt:
ps I re-summarized the outcome in the commit message https://github.com/tetratelabs/wazero/commit/197facf7a93c4b3cae1af896660b506368201a3d

PR #1398, opened 2023-04-24
wasi: improve select(2) impl on Windows

evacchi:
An improved version of the `select(2)` syscall emulation on Windows for Stdin. We use `PeekNamedPipe` when `Stdin` is a FILE_TYPE_CHAR, and we invoke it every 100ms if duration > 0. Follow up to #1346 and #1317.

I checked the examples manually for interactive input and they seem to work for me.

### DIFF
--- a/internal/platform/select_windows.go
+++ b/internal/platform/select_windows.go
@@ -3,16 +3,31 @@ package platform
 import (
 	"syscall"
 	"time"
+	"unsafe"
 )
 
 // wasiFdStdin is the constant value for stdin on Wasi.
 // We need this constant because on Windows os.Stdin.Fd() != 0.
 const wasiFdStdin = 0
 
+// pollInterval is the interval between each calls to peekNamedPipe in pollNamedPipe
+const pollInterval = 100 * time.Millisecond
+
+// procPeekNamedPipe is the syscall.LazyProc in kernel32 for PeekNamedPipe
+var procPeekNamedPipe = kernel32.NewProc("PeekNamedPipe")
+
 // syscall_select emulates the select syscall on Windows for two, well-known cases, returns syscall.ENOSYS for all others.
-// If r contains fd 0, then it immediately returns 1 (data ready on stdin) and r will have the fd 0 bit set.
+// If r contains fd 0, and it is a regular file, then it immediately returns 1 (data ready on stdin)
+// and r will have the fd 0 bit set.
+// If r contains fd 0, and it is a FILE_TYPE_CHAR, then it invokes PeekNamedPipe to check the buffer for input;
+// if there is data ready, then it returns 1 and r will have fd 0 bit set.
 // If n==0 it will wait for the given timeout duration, but it will return syscall.ENOSYS if timeout is nil,
 // i.e. it won't block indefinitely.
+//
+// Note: idea taken from https://stackoverflow.com/questions/6839508/test-if-stdin-has-input-for-c-windows-and-or-linux
+// PeekNamedPipe: https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe
+// "GetFileType can assist in determining what device type the handle refers to. A console handle presents as FILE_TYPE_CHAR."
+// https://learn.microsoft.com/en-us/windows/console/console-handles
 func syscall_select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error) {
 	if n == 0 {
 		// don't block indefinitely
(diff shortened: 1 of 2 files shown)


evacchi:
Marking back as draft because the tickers are not being cleaned up

evacchi:
@achille-roussel I did learn about PeekNamedPipe online (will dig up a reference), but the poll loop I came up with myself. Open to suggestions because I know it's not great 😬 but I also thought it's very specific to stdin, so good enough for this one thing.

codefromthecrypt:
[1] Vacchi, Edoardo. "Not really great, but works", 2023

internal/platform/select_windows.go
```diff
@@ -23,9 +33,84 @@ func syscall_select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error)
...
+// pollNamedPipe polls the given named pipe handle for the given duration.
+//
+// The implementation actually polls at
+func pollNamedPipe(pipeHandle syscall.Handle, duration *time.Duration) (bool, error) {
+	// Short circuit when the duration is nil.
+	if duration != nil && *duration == time.Duration(0) {
+		return peekNamedPipe(pipeHandle)
+	}
+	// Ticker that emits at every pollInterval.
+	tick := time.NewTicker(pollInterval)
```
ncruces, line 66: `defer tick.Stop()` to release resources?

internal/platform/select_windows.go
```diff
@@ -23,9 +33,84 @@ func syscall_select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error)
...
+				return false, err
+			}
+			if res {
+				println("ret")
+				return res, nil
+			}
+		}
+	} else {
+		// Otherwise, leave after the given duration, and check every pollInterval.
+		after := time.After(*duration)
```
ncruces, line 81: `time.NewTimer(…)` and `defer timer.Stop()` to release resources?

evacchi:
> [1] Vacchi, Edoardo. "Not really great, but works", 2023

the title of my memoir

internal/platform/select_windows.go
```diff
@@ -23,9 +38,86 @@ func syscall_select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error)
...
+	defer tick.Stop()
+	// If the duration is nil, then poll forever.
+	if duration == nil {
+		for range tick.C {
+			res, err := peekNamedPipe(pipeHandle)
+			if err != nil && err != syscall.Errno(0) {
+				return false, err
+			}
+			if res {
+				println("ret")
```
mathetake, line 81: garbage

evacchi: whooopss!

evacchi:
@ktock fyi

evacchi:
@mathetake nope! the main issue is that it is hard to simulate a Stdin with FILE_TYPE_CHAR in the test runner. However, I think I may be able to test `peekNamedPipe` with a custom handle. Will look into this.

internal/platform/select_windows.go
```diff
@@ -23,9 +39,97 @@ func syscall_select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error)
...
 	}
 	return -1, syscall.ENOSYS
 }
+
+// pollNamedPipe polls the given named pipe handle for the given duration.
+//
+// The implementation actually polls every 100 milliseconds until it reaches the given duration.
+// The duration may be nil, in which case it will wait undefinely. The given ctx is
+// used to allow for cancellation. Currently used only in tests.
+func pollNamedPipe(pipeHandle syscall.Handle, duration *time.Duration, ctx context.Context) (bool, error) {
```
ktock, line 68: `ctx` should be the first parameter.

https://pkg.go.dev/context#pkg-overview

> The Context should be the first parameter, typically named ctx

internal/platform/select_windows.go
```diff
@@ -23,9 +39,97 @@ func syscall_select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error)
...
+	// Short circuit when the duration is nil.
```
ktock, line 69:
```suggestion
	// Short circuit when the duration is zero.
```

evacchi: that's actually supposed to be nil

internal/platform/select_windows.go
```diff
@@ -23,9 +39,97 @@ func syscall_select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error)
...
+				if res {
+					return res, nil
+				}
+			}
+		}
+	}
+	return false, nil
+}
+
+// [in]            HANDLE  hNamedPipe,
```
codefromthecrypt, line 116: can you move these comments to trailing on the lines they are set. e.g. `uintptr(handle), // [in]            HANDLE  hNamedPipe`

internal/platform/select_windows.go
```diff
@@ -23,9 +39,97 @@ func syscall_select(n int, r, w, e *FdSet, timeout *time.Duration) (int, error)
...
+					return false, err
+				}
+				if res {
+					return res, nil
+				}
+			}
+		}
+	} else {
+		// Otherwise, leave after the given duration, and check every pollInterval.
+		after := time.NewTimer(*duration)
```
codefromthecrypt, line 94: I wonder if you can make this an empty timer and avoid the if/else copy/pasta? So when there is no duration a timer that never stops?

Not sure if this would work.
```
var neverEnding =  &time.Ticker{C: make(chan time.Time)}
```

codefromthecrypt (approved):
Looks good, one suggestion for code tightening but not sure it would work or not.

internal/platform/select_windows.go
ncruces: Receiving from the `nil` channel blocks forever (and thus never satisfies `select`). Your `var` can be the channel, and you initialize it inside an `if` otherwise it's `nil`.

internal/platform/select_windows.go
```diff
@@ -1,31 +1,123 @@
...
+			return false, nil
+		case <-afterCh:
+			return false, nil
+		case <-tichCh:
+			res, err := peekNamedPipe(pipeHandle)
+			if err != nil {
+				return false, err
+			}
+			if res {
+				return res, nil
```
ncruces, line 101: super nit: I find `return true, nil` more explicit.

ncruces (approved):

internal/platform/select_windows.go
evacchi: Yeah fair enough. That smelled a little bit for me too :D

ncruces (approved):

PR #1400, opened 2023-04-24
logging: correctly omits result.newoffset in Before

mathetake:


internal/wasip1/logging/logging.go
```diff
@@ -157,6 +157,7 @@ func Config(fnd api.FunctionDefinition) (pSampler logging.ParamSampler, pLoggers
 			name = resultParamName(name)
 			logger = logMemI64(idx).Log
 			rLoggers = append(rLoggers, resultParamLogger(name, logger))
+			continue
```
mathetake, line 160: this was missing

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -2519,7 +2519,7 @@ func Test_fdSeek(t *testing.T) {
 				'?',
 			},
 			expectedLog: `
-==> wasi_snapshot_preview1.fd_seek(fd=4,offset=4,whence=0,4557430888798830399)
```
mathetake, line 2522: just for anyone curious: basically this weird number is the u64 value on mem[newoffset]  _before_ we execute fd_seek.

mathetake: which is useless and meaningless in any way

codefromthecrypt:
thanks!

PR #1403, opened 2023-04-25
Attach DWARF names to function definitions and use in stacktraces

pelletier:
Ultimately the goal of this change is to leverage DWARF information to provide more human-readable names for functions of languages that emit "mangled" function names (rust, c++, among others).

Practically speaking, I would like to use those function names in `experimental.Listeners`. However, I felt like it would benefit Wazero users to also see them in stack traces (see Example below).

The high level changes are:

* Leverage the `linkage_name` DWARF attribute attached to subprograms to match their mangled name to their human-friendly name at the time of loading the module.
* Use those names when printing stack traces.
* Provide a new `experimental.FunctionDefinition` interface that extends `api.FunctionDefinition` with a new `HumanName()` method to retrieve the more human-friendly function names, without impacting the public API.
* Fallback to the function's debug name when this name is not present.

In this patch, the linkage_name lookup is only performed for Rust. Providing per-language support is a [strategy similar to wasmtime's](https://github.com/bytecodealliance/wasmtime/blob/d2bb4aa13b7f018854147ed41b272958a3106791/crates/jit/src/demangling.rs#L18). It can be extended to add support for C++ and other languages. Ultimately, interpreting DWARF data is a task very context sensitive, so we may want to eventually expose the raw DWARF sections through the API to let the user decide what's best for their context.

Note: I don't love the name `HumanName`. I'm eager for better suggestions!

Related #wazero-dev discussion: https://gophers.slack.com/archives/C04CG4A2NKX/p1682361288022999

## Example

For example, the following rust program (included in this patch):

```rust
fn main() {
    a();
}

#[inline(always)]
fn a() {
    b(42);
}

fn b<A: std::fmt::Debug>(x: A) {
    println!("{:?}", x);
    panic!("unreachable");
}
```

currently results in the following stacktraces when run by Wazero:

```
.__rust_start_panic(i32) i32
	0x60bf: /lib.rs:83:17 (inlined)
	        /lib.rs:37:5
.rust_panic(i32,i32)
	0x5dad: /panicking.rs:740:9
._ZN3std9panicking20rust_panic_with_hook17h91fa5cbfc957c96cE(i32,i32,i32,i32,i32)
	0x5d74: /panicking.rs:710:5
._ZN3std9panicking19begin_panic_handler28_$u7b$$u7b$closure$u7d$$u7d$17hea8cd6e90707c2a1E(i32)
	0x4e2e: /panicking.rs:577:13
._ZN3std10sys_common9backtrace26__rust_end_short_backtrace17h8766955a633f184aE(i32)
	0x4d90: /backtrace.rs:137:18
.rust_begin_unwind(i32)
	0x53e2: /panicking.rs:575:5
._ZN4core9panicking9panic_fmt17hea96ca76090d462fE(i32,i32)
	0xaab8: /panicking.rs:64:14
._ZN4main1b17h47e5ee1ee8873d72E(i32)
	0x305: /main.rs:12:5
._ZN4main4main17hb80dac2ea6f04fdaE()
	0x5d6: /main.rs:7:5 (inlined)
	       /main.rs:2:5
._ZN4core3ops8function6FnOnce9call_once17hb25884e69c21780fE(i32)
	0x142: /function.rs:250:5
._ZN3std10sys_common9backtrace28__rust_begin_short_backtrace17he6b53729c962b36cE(i32)
	0x5b1: /backtrace.rs:121:18
._ZN3std2rt10lang_start28_$u7b$$u7b$closure$u7d$$u7d$17hdfaa3c19d02e0d53E(i32) i32
	0x3d4: /rt.rs:166:18
._ZN3std2rt19lang_start_internal17h38aaea5d7881ae71E(i32,i32,i32,i32,i32) i32
	0x25d6: /function.rs:287:13 (inlined)
	        /panicking.rs:483:40 (inlined)
	        /panicking.rs:447:19 (inlined)
	        /panic.rs:140:14 (inlined)
	        /rt.rs:148:48 (inlined)
	        /panicking.rs:483:40 (inlined)
	        /panicking.rs:447:19 (inlined)
	        /panic.rs:140:14 (inlined)
	        /rt.rs:148:20
._ZN3std2rt10lang_start17hb9ddebcd8dc7db51E(i32,i32,i32,i32) i32
	0x371: /rt.rs:165:17
.__main_void() i32
._start()
```

in contrast with this change:

```
.__rust_start_panic(i32) i32
	0x60bf: /lib.rs:83:17 (inlined)
	        /lib.rs:37:5
.rust_panic(i32,i32)
	0x5dad: /panicking.rs:740:9
rust_panic_with_hook(i32,i32,i32,i32,i32)
	0x5d74: /panicking.rs:710:5
{closure#0}(i32)
	0x4e2e: /panicking.rs:577:13
__rust_end_short_backtrace<std::panicking::begin_panic_handler::{closure_env#0}, !>(i32)
	0x4d90: /backtrace.rs:137:18
begin_panic_handler(i32)
	0x53e2: /panicking.rs:575:5
panic_fmt(i32,i32)
	0xaab8: /panicking.rs:64:14
b<i32>(i32)
	0x305: /main.rs:12:5
main()
	0x5d6: /main.rs:7:5 (inlined)
	       /main.rs:2:5
call_once<fn(), ()>(i32)
	0x142: /function.rs:250:5
__rust_begin_short_backtrace<fn(), ()>(i32)
	0x5b1: /backtrace.rs:121:18
{closure#0}<()>(i32) i32
	0x3d4: /rt.rs:166:18
lang_start_internal(i32,i32,i32,i32,i32) i32
	0x25d6: /function.rs:287:13 (inlined)
	        /panicking.rs:483:40 (inlined)
	        /panicking.rs:447:19 (inlined)
	        /panic.rs:140:14 (inlined)
	        /rt.rs:148:48 (inlined)
	        /panicking.rs:483:40 (inlined)
	        /panicking.rs:447:19 (inlined)
	        /panic.rs:140:14 (inlined)
	        /rt.rs:148:20
lang_start<()>(i32,i32,i32,i32) i32
	0x371: /rt.rs:165:17
.__main_void() i32
._start()
```


internal/wasm/function_definition.go
```diff
@@ -102,13 +102,31 @@ func (m *Module) BuildFunctionDefinitions() {
...
 		}
 	}
+
+	if len(m.FunctionDefinitionSection) == 0 || m.DWARFLines == nil {
+		return
+	}
+	switch m.DWARFLines.Language() {
+	case wasmdebug.DwarfLangRust:
+		m.DWARFLines.LinkageNamesDo(func(linkageName, name string) {
+			for i := 0; i < len(m.FunctionDefinitionSection); i++ {
```
pelletier, line 112: There may be opportunities for improvement by keeping the index from one call to another, which would lower the number of iterations if the DIEs are roughly in the same order as how functions are laid out in the WASM module.

internal/wasmdebug/dwarf.go
```diff
@@ -33,6 +35,130 @@ func NewDWARFLines(d *dwarf.Data) *DWARFLines {
...
+		// Assumes same language for all compilation units.
+		*d.lang = DwarfLang(langAttr.(int64))
+		break
+	}
+	return *d.lang
+}
+
+// LinkageNamesDo parses the DWARF data and execute f for all subprograms that
+// have a linkage and a name attributes.
+func (d *DWARFLines) LinkageNamesDo(f func(linkageName, name string)) {
```
pelletier, line 130: Went for a callback style iterator instead of:

* generating a huge map: would lead to even more allocations.
* an iterator style, which may be tricky with the lock.

internal/wasmdebug/dwarf.go
```diff
@@ -33,6 +35,130 @@ func NewDWARFLines(d *dwarf.Data) *DWARFLines {
...
+		return
+	}
+
+	d.mux.Lock()
+	defer d.mux.Unlock()
+
+	r := d.d.Reader()
+
+	for {
+		ent, err := r.Next()
```
pelletier, line 141: There may be opportunities for speedup by leveraging `r.SkipChildren()`.

pelletier:
For reference, this comes from https://github.com/tetratelabs/wazero/issues/1350#issuecomment-1501380158.

internal/engine/compiler/engine.go
```diff
@@ -785,7 +785,11 @@ func (ce *callEngine) deferredOnCall(recovered interface{}) (err error) {
 					sources = p.sourceModule.DWARFLines.Line(offset)
 				}
 			}
-			builder.AddFrame(def.DebugName(), def.ParamTypes(), def.ResultTypes(), sources)
+			name := def.DebugName()
+			if expf, ok := def.(experimental.FunctionDefinition); ok {
+				name = expf.HumanName()
+			}
```
achille-roussel, line 791: Is it worth putting a function to do this in the `experimental` package so we don't repeat it in the interpreter and compiler?

```go
func DebugName(def api.FunctionDefinition) string {
   ...
}
```

experimental/experimental.go
```diff
@@ -2,3 +2,14 @@
...
+
+import "github.com/tetratelabs/wazero/api"
+
+// FunctionDefinition is an extend interface of api.FunctionDefinition with
+// additional experimental methods.
+type FunctionDefinition interface {
+	api.FunctionDefinition
+	// HumanName is a human-readable name for this function, which may not be
+	// the same as Name. It will fallback to DebugName() if not available.
+	HumanName() string
```
achille-roussel, line 14: What do you think about `SourceName` instead?

achille-roussel: Could you also speak of why not implement this in `api.FunctionDefinition.DebugName` directly?

internal/wasmdebug/dwarf.go
```diff
@@ -17,7 +17,9 @@ type DWARFLines struct {
 	// linesPerEntry maps dwarf.Offset for dwarf.Entry to the list of lines contained by the entry.
 	// The value is sorted in the increasing order by the address.
 	linesPerEntry map[dwarf.Offset][]line
-	mux           sync.Mutex
+	// Cache for Language(). nil when Language() has never been called.
+	lang *DwarfLang
```
achille-roussel, line 21: Any reason not to use `DwarfLangUnkonwn` instead of a pointer to indicate that the value is uninitialized?

pelletier: Ah sorry I meant to add a comment about this.

It is to distinguish between "Language() has not been called" and "DWARF info doesn't say anything about language". Could have also used a bit flag or an extra boolean variable, but that seemed the most straightforward that way.

internal/engine/compiler/engine.go
pelletier: Would it be confusing for the users of `experimental` that they have access to that information both using a function and through the `experimental.FunctionDefinition` API?

experimental/experimental.go
pelletier: Happy with `SourceName`!

I didn't want to touch `DebugName` because it seemed like it has special semantic:

> DebugName identifies this function based on its Index or Name in the module.

For example, `HumanName` don't have any guarantees (doesn't necessarily include information about the module). Definitely agree that having 3 names on one function definition is a little much though.

internal/engine/compiler/engine.go
achille-roussel: Maybe this is the good place for `FunctionHumanReadableName` or `BestDebugName` (doesn't have to be called the same as the method name).

experimental/experimental.go
codefromthecrypt: Having been the one to add DebugName, I think this is absolutely fine to use Dwarf when we can. It isn't stable anyway, so maybe rewrite the godoc and use it?

> DebugName returns the best non-empty name available for this function.
>
> For example, If DWARF is available, this will match the original source code name.
> Otherwise, this returns Name when non-empty or a generated name based on the
> Index.

pelletier: Sound great to me! I'll update.

internal/engine/compiler/engine.go
pelletier: This seems unnecessary now as `DebugName()` is repurposed.

pelletier:
Updated to use `DebugName` instead of a dedicated interface according to https://github.com/tetratelabs/wazero/pull/1403#discussion_r1177148782.

experimental/experimental.go
pelletier: Done in https://github.com/tetratelabs/wazero/pull/1403/commits/5ab31ab98611fb6a2df3f62566f163fec501076a.

internal/wasm/function_definition.go
mathetake: I wonder how slow could this be in reality - the sqlite3 binary has roughly 1.5k functions, and I guess this could be really slow https://github.com/ncruces/go-sqlite3/blob/main/embed/sqlite3.wasm

codefromthecrypt: we have a benchmark right? just @pelletier needs to manually download the sqlite wasm iirc to replace internal/integration_test/bench/testdata/case.wasm then run Benchmark in internal/integration_test/bench/bench_test.go right?

mathetake: we have a bench but dwarf is not available in these checked-in binaries as it is huge.

codefromthecrypt: https://github.com/ncruces/go-sqlite3/blob/main/embed/sqlite3.wasm has dwarf tho right? for running bench offline I mean (I didn't mean to commit the sqlite to this PR, rather run it to check)

mathetake: plus the case.wasm is in reality not that big compared to sqlite3 - the number of functions 10x smaller and this for-loop can be O(N^2) in worst case maybe

mathetake: > which would lower the number of iterations if the DIEs are roughly in the same order as how functions are laid out in the WASM module.

on this point, I remember Zig produces the "weird" dwarf as I added in the comment in this file. Maybe this assumption could be wrong

```
 		// Note that the dwarf.LineEntry.SeekPC API shouldn't be used because the Go's dwarf package assumes 
		// all the line entries in an Entry are sorted in increasing order which *might not* be true
		// for some languages. Such order requirement is not a part of DWARF specification,
		// and in fact Zig language tends to emit interleaved line information.
```

mathetake:
Regarding https://github.com/tetratelabs/wazero/pull/1403#discussion_r1178525749, I definitely would love to see if this works well with Zig too

internal/wasm/function_definition.go
codefromthecrypt: for reducing likelihood of worst case, we can do the normal map lookup thing right?  We can also move the map somewhere it is re-used. We have lookups elsewhere during compilation though agree maybe this would be a larger one.

style: I personally would not use a callback rather imperative and a map to prevent worst case (knowing it will allocate to the size of max functions). I think we rarely, if ever, make do/callback style in our code regardless of locking. It would be cheaper to stay imperative as well, so that helps with the cost.

mathetake:
one idea worth thinking is to do lazy initialization of `FunctionDefinition.DebugName()`

codefromthecrypt:
> one idea worth thinking is to do lazy initialization of FunctionDefinition.DebugName(), then we don't need to care about the perf

This is a great idea as I don't think we eagerly resolve it anywhere except error (deferredOnCall)!

mathetake:
>  I suggest we close this change and re-open a PR when we have matured more in that space, as the final solution may look different. What do you think?

ack and agree with your point on this change not necessarily being helpful to users considering the limitation. In anyway, thank you for your work here and I think we could reuse the code here in the future!

PR #1404, opened 2023-04-25
Give Listener a read-only view of globals

pelletier:
As a follow up to https://github.com/tetratelabs/wazero/pull/1363.

This change is motivated by the fact that Go maintains [its own stack in the linear memory](https://docs.google.com/document/d/131vjr4DH6JFnb-blm_uRdaC0_Nv3OUwjEY5qVCxCup4/edit#heading=h.58es2esmswsk), instead of using the regular wasm stack. The Go runtime stores the stack pointer in global 0. In order to retrieve arguments from the stack, the listener has to read the value of global 0, then the memory.

I didn't include this as part of `Listener.Args()` because it is Go-specific. This logic can be implemented in a Listener without making Wazero aware of its specificities, as well as keeping StackIterator an iterator of the wasm stack.

One of the feedback on #1363 was that a user could modify the data presented to them in the Listener, while it was expected to be read-only. Here I took the approach of doing a proxy to achieve this – instead of just passing a slice to the globals, or making a copy of them. Let me know what you think!


achille-roussel:
Personally I would prefer a slice because it's less code to maintain. The `api.Global` interface does not allow modifying the value of the global, which achieves the desirable read-only behavior I believe.

If the user mutates the content of the `[]api.Global` argument, it can be made a non-issue; are we protecting against users mutating the wasm stack in function listeners?

pelletier:
> The api.Global interface does not allow modifying the value of the global, which achieves the desirable read-only behavior I believe.

I initially thought the same, but `wasm.GlobalInstance` doesn't implement `api.Global`. I don't see why it couldn't, but that would add to the public API. Do you think it's worth creating another type that implements `api.Global` with the same shape as `wasm.GlobalInstance`, and copy the slice module instance slice of globals every time? Or is there something that I missed?

> are we protecting against users mutating the wasm stack in function listeners?

We're not, but it was something we flagged to do, so wanted to do better this time :)

experimental/listener.go
```diff
@@ -24,6 +24,16 @@ type StackIterator interface {
 	Parameters() []uint64
 }
 
+// Globals is a proxy to the set of globals available to the context of a
+// function call.  This is useful for inspecting globals, but not modifying
+// them.
+type Globals interface {
+	// Count returns the number of globals registered at this moment.
+	Count() int
```
codefromthecrypt, line 32: the count of globals is static for the life of the module isn't it? If so, "registered at this moment" may be misleading.

Are you adding this here because the data isn't available on CompiledModule?

codefromthecrypt (commented):
I think I understand why you are going for this, but I think the cleaner way is to make an unexported view of api.Module instead. This is secretly why the functions on api.Module are prefixed Exported

so maybe make an interface experimental.Module (or a clever name) which has the extra module access you need, and cast into it?

experimental/listener.go
pelletier: > the count of globals is static for the life of the module isn't it?

Didn't realize that, thank you!

> Are you adding this here because the data isn't available on CompiledModule?

Added it because `api.Module` doesn't provide that information (only for exported globals).

pelletier:
@codefromthecrypt sounds good to me!

How about `experimental.FullModule`?

What do you think about replacing the `api.Module` in the `Before`/`After` function signatures with a `experimental.FullModule`, instead of having the implementation of those functions make a cast?

codefromthecrypt:
> How about experimental.FullModule?

I prefer experimental.InternalModule because I don't really think we ever want to add this normally.

> What do you think about replacing the api.Module in the Before/After function signatures with a experimental.FullModule, instead of having the implementation of those functions make a cast?

The reason we don't expose the full module is because you can monkey inside of it (as you are trying to do) and invalidate state. The listener wasn't originally trying for a use case to modify internal state, or even listen on it. So I don't really want to encourage use of this by changing the signature.

codefromthecrypt:
More thoughts.. others may have different opinions and I might come around, too.

I think wasm generally knows things they export will be messed with, and the contrary false. So, I think we should be really careful about accidentally exporting the whole thing just because of a current thing with Go. This is exactly the kind of thing that requires a lot of opinions before it becomes normal. Extern types that are missing etc, are a no brainer, making the black box grey.. little more concerning.

hope the ramble helps!

codefromthecrypt:
discussion here is fine!

pelletier:
Removed the proxy stuff to provide `experimental.InternalModule` instead. Let me know what you think!

internal/wasm/store.go
```diff
@@ -126,6 +126,9 @@ type (
 		aliases []string
 		// Definitions is derived from *Module, and is constructed during compilation phrase.
 		Definitions []FunctionDefinition
+
+		// globalsView holds a read-only view of all globals for experimental.InteralModule.
+		globalsView []api.Global
```
mathetake, line 131: I would like to avoid allocating this slice regardless of the usage of the experimental feature

experimental/experimental.go
```diff
@@ -2,3 +2,15 @@
...
+import (
+	"github.com/tetratelabs/wazero/api"
+)
+
+// InternalModule exposes additional module information not available through
+// api.Module.
+type InternalModule interface {
+	// ViewGlobals return a copy of all the globals defined in the module. It
+	// includes non-exported globals.
+	ViewGlobals() []api.Global
```
mathetake, line 15: If we limit this to by-index-access only, then we should be able to avoid allocating `globalsView []api.Global`?
```suggestion
	Global(index int) []api.Global
	GlobalsCount() int
```

pelletier:
@mathetake changed the API to use a Count + index style instead. I implemented `api.Global` on the value of a new internal struct `constantGlobal`. That way, there shouldn't be any allocation regardless of using that experimental feature.

After chatting with @achille-roussel , I took advantage of that change to factor in type-specific versions of globals (globalI32 and friends) into `constantGlobal` and the bulk of `mutableGlobal`.

experimental/experimental.go
```diff
@@ -2,3 +2,17 @@
...
+
+// InternalModule exposes additional module information not available through
+// api.Module.
+type InternalModule interface {
+	// GlobalsCount returns the count of all globals in the module.
+	GlobalsCount() int
+
+	// ViewGlobal provides a read-only view for a given global index. Panics if idx > GlobalsCount().
+	ViewGlobal(idx int) api.Global
+}
```
achille-roussel, line 18: I'd make a few changes here, tho some might be opinionated, so others may have different perspectives.

- Let's have `InternalModule` implement `api.Module`, since it's intended to be an extension of it
- Unless we've used the *Count name elsewhere in the wazero APIs, how about naming the method `NumGlobal` (e.g. `reflect.Type` uses `NumField`, `NumMethod`, etc...)
- Rename `ViewGlobal` to just `Global`

```go
type InternalModule interface {
    api.Module
    
    NumGlobal() int
    
    Global(i int) api.Global
}
```

internal/wasm/global.go
```diff
@@ -44,82 +36,16 @@ func (g *mutableGlobal) String() string {
...
-	return uint64(g)
-}
+// compile-time check to ensure mutableGlobal is a api.Global.
+var _ api.Global = mutableGlobal{}
 
-// String implements fmt.Stringer
-func (g globalF64) String() string {
-	return fmt.Sprintf("global(%f)", api.DecodeF64(g.Get()))
+// Set implements the same method as documented on api.MutableGlobal.
+func (g *mutableGlobal) Set(v uint64) {
```
achille-roussel, line 49: Let's pass the `mutableGlobal` receiver by value? Then change the allocation from `&mutableGlobal{...}` to `mutableGlobal{...}` (the pointer size struct value can then be inlined into the interface and avoid heap allocations).

internal/wasm/global.go
```diff
@@ -44,82 +36,16 @@ func (g *mutableGlobal) String() string {
...
-// compile-time check to ensure globalF64 is a api.Global
-var _ api.Global = globalF64(0)
-
-// Type implements the same method as documented on api.Global.
-func (g globalF64) Type() api.ValueType {
-	return ValueTypeF64
+// mutableGlobal extends constantGlobal to allow updates.
+type mutableGlobal struct {
+	constantGlobal
+	internalapi.WazeroOnlyType
```
achille-roussel, line 42: This needs to be the first parameter so it can be optimized as a zero-size field by the compiler.

internal/wasm/global.go
```diff
@@ -44,82 +36,16 @@ func (g *mutableGlobal) String() string {
...
-
-// compile-time check to ensure globalF64 is a api.Global
-var _ api.Global = globalF64(0)
-
-// Type implements the same method as documented on api.Global.
-func (g globalF64) Type() api.ValueType {
-	return ValueTypeF64
+// mutableGlobal extends constantGlobal to allow updates.
+type mutableGlobal struct {
+	constantGlobal
```
achille-roussel, line 41: Expressing `mutableGlobal` in terms of `constantGlobal` is tempting to remove a bit of code but it's also a bit misleading.

The `String` function is main piece of logic that is really worth reusing, how about factoring this out in a function and keeping the implementations of `mutableGlobal` and `constantGlobal` separated?
```go
func (g *GlobalInstance) String() string {
  ...
}
```
```go
func (g constantGlobal) String() string {
    return g.g.String()
}
```
```go
func (g mutableGlobal) String() string {
    return g.g.String()
}
```

internal/wasm/global.go
pelletier: TIL thank you!

internal/wasm/global.go
pelletier: Meant to do that but forgot to remove that pointer 🤦🏻 thank you!

internal/wasm/global.go
pelletier: Fair enough. It did feel a little weird replacing it everywhere.

internal/testing/enginetest/enginetest.go
```diff
@@ -632,7 +758,7 @@ func (f *fnListener) NewListener(api.FunctionDefinition) experimental.FunctionLi
 	return f
 }
 
-func (f *fnListener) Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, paramValues []uint64, stackIterator experimental.StackIterator) context.Context {
+func (f fnListener) Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, paramValues []uint64, stackIterator experimental.StackIterator) context.Context {
```
mathetake, line 761: could you stick to the pointer or value in receiver?

```suggestion
func (f *fnListener) Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, paramValues []uint64, stackIterator experimental.StackIterator) context.Context {
```

internal/testing/enginetest/enginetest.go
```diff
@@ -623,6 +624,131 @@ func RunTestModuleEngineBeforeListenerStackIterator(t *testing.T, et EngineTeste
 	require.Equal(t, 0, len(expectedCallstacks))
 }
 
+// This tests that the Globals provided by the Engine to the Before hook of the
+// listener is properly able to read the values of the globals.
+func RunTestModuleEngine_BeforeListenerGlobals(t *testing.T, et EngineTester) {
```
mathetake, line 629:
```suggestion
func RunTestModuleEngineBeforeListenerGlobals(t *testing.T, et EngineTester) {
```

experimental/experimental.go
pelletier: > Unless we've used the *Count name elsewhere in the wazero APIs, how about naming the method NumGlobal (e.g. reflect.Type uses NumField, NumMethod, etc...)

Good idea. On a quick search, I don't see any precedent of either.

pelletier: > Rename ViewGlobal to just Global

Happy to. I wrote it that way to signal that it's going to be read-only (if you know that a given global should be mutable, you're still not going to be able to edit it through this API).

internal/testing/enginetest/enginetest.go
pelletier: Sorry for the mixup.

mathetake (approved):
Looks good to me now. good to go once @achille-roussel or @codefromthecrypt feels ready

pelletier:
Thanks a lot for the feedback! I believe I've addressed @achille-roussel's concerns in the last couple commits.

internal/wasm/global.go
```diff
@@ -1,125 +1,55 @@
...
-	"fmt"
-
 	"github.com/tetratelabs/wazero/api"
 	"github.com/tetratelabs/wazero/internal/internalapi"
 )
 
-type mutableGlobal struct {
-	internalapi.WazeroOnlyType
+// constantGlobal wraps GlobalInstance to implement api.Global.
+type constantGlobal struct {
```
achille-roussel, line 9: This type should have `internalapi.WazeroOnlyType` as well!

internal/wasm/module_instance.go
```diff
@@ -225,17 +225,15 @@ func (m *ModuleInstance) ExportedGlobal(name string) api.Global {
 	if g.Type.Mutable {
 		return &mutableGlobal{internalapi.WazeroOnlyType{}, g}
```
achille-roussel, line 226:
```suggestion
		return mutableGlobal{g: g}
```

achille-roussel (approved):
Looking good, thanks for the quick follow ups!

I left a few more comments but overall this looks good to go 👍

internal/wasm/global.go
pelletier: added. neat!

PR #1405, opened 2023-04-26
ci: cleanup GOOS=wasip1

mathetake:
This enables macos-12 plus removes the need to install gotip when the cache exists.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -435,7 +435,6 @@ jobs:
             src/os/signal \
             src/os/user \
             src/path \
-            src/path/filepath \
             src/reflect \
             src/regexp \
             src/regexp/syntax \
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -355,14 +355,14 @@ jobs:
           cd $(go env GOROOT)/src/os; wazero run -mount=/:/ ~/bin/os.wasm -test.v
 
   go_tests:
-    # Due to the embedding of the GOOS of the building env(https://github.com/golang/go/blob/3c59639b902fada0a2e5a6a35bafd10fc9183b89/src/os/os_test.go#L112),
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yaml
```diff
@@ -354,27 +354,30 @@ jobs:
         run: |
           cd $(go env GOROOT)/src/os; wazero run -mount=/:/ ~/bin/os.wasm -test.v
 
-  build_go_tests:
-    name: Build Go test binaries
-    runs-on: ubuntu-22.04
-    env:
-      BINARY_OUT: ~/go-test-binaries
+  go_tests:
+    # Due to the embedding of the GOOS of the building env(https://github.com/golang/go/blob/3c59639b902fada0a2e5a6a35bafd10fc9183b89/src/os/os_test.go#L112),
```
codefromthecrypt, line 358: nice comment

mathetake:
```
--- FAIL: TestWalkSymlinkRoot (0.01s)
    --- FAIL: TestWalkSymlinkRoot/double_link_with_slash (0.00s)
```

https://github.com/tetratelabs/wazero/actions/runs/4804126555/jobs/8549308990?pr=1405

seems like a legit failure on macos. @achille-roussel any idea?

mathetake:
```
		{
			desc:      "double link with slash",
			root:      linklink + string(filepath.Separator),
			want:      []string{linklink, filepath.Join(linklink, "foo")},
			buggyGOOS: []string{"darwin", "ios"}, // https://go.dev/issue/59586
		},
```

hah :D

.github/workflows/integration.yaml
```diff
@@ -434,7 +435,6 @@ jobs:
             src/os/signal \
             src/os/user \
             src/path \
-            src/path/filepath \
```
mathetake, line 437: for now, disabled path/filepath as I suppose it would fail on windows as well

PR #1406, opened 2023-04-26
ci: try enabling Windows on wasip1 test

mathetake:
started running, but failing randomly :D 
https://github.com/tetratelabs/wazero/actions/runs/4804810021/jobs/8550649684

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -484,6 +484,6 @@ jobs:
             fi
 
             echo ${MOUNT} ${IN_WASM_PWD} ~/sdk/gotip/$dir
-            wazero run -mount=${MOUNT} -env PWD=${IN_WASM_PWD} $bin -- -test.short -test.v
+            wazero run -mount=${MOUNT} -hostlogging=filesystem -env PWD=${IN_WASM_PWD} $bin -- -test.short -test.v
             popd
           done
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -466,8 +466,26 @@ jobs:
         # TODO: delete - temporarily enabled the cache working regardless.
         continue-on-error: true
         run: |
+
           GOOS=$(go env GOOS)
           echo "Running $(find ~/sdk/tests -name *.test | wc -l) test binaries"
+
+          GOROOT=$(gotip env GOROOT)
+        
+          if [[ "$GOOS" != "windows" ]]; then
+            MOUNT=/:/
+            SCRIPT="$HOME/sdk/tests.sh"
+            RUNNER="sh ${SCRIPT}"
+            POSTFIX=""
+          else
+            MOUNT=c:\\:/
+            SCRIPT="$HOME/sdk/tests.cmd"
+            # Trim `/c` from the in-Wasm GOROOT.
+            GOROOT=$(cygpath -u $GOROOT | cut -c 3-)
+            # Append early exit on cmd.
+            POSTFIX="if %errorlevel% neq 0 exit /b %errorlevel%"
+            RUNNER="cmd //c %USERPROFILE%\sdk\tests.cmd"
+          fi
 
           # Go tests often look for files relative to the source. Change to the corresponding directory.
           for bin in ~/sdk/tests/*.test; do
@@ -475,15 +493,23 @@ jobs:
             pushd ~/sdk/gotip/$dir
 
             if [[ "$GOOS" != "windows" ]]; then
-              MOUNT=/:/
               IN_WASM_PWD=$(pwd)
             else
-              MOUNT=c:\\:/
               # Trim `/c` from the in-Wasm pwd.
               IN_WASM_PWD=$(pwd | cut -c 3-)
+              # Convert to a Windows path.
+              bin=`cygpath -w $bin`
             fi
 
+            # Create a script with all the tests (do not run yet).
             echo ${MOUNT} ${IN_WASM_PWD} ~/sdk/gotip/$dir
-            wazero run -mount=${MOUNT} -hostlogging=filesystem -env PWD=${IN_WASM_PWD} $bin -- -test.short -test.v
+            COMMAND="wazero run -mount=${MOUNT} -hostlogging=filesystem -env PWD=${IN_WASM_PWD} -env GOROOT=${GOROOT} -env GOOS=wasip1 $bin -- -test.short -test.v"
+            echo $COMMAND >> $SCRIPT
+            # Uncomment the following line for early exit on error on Windows. 
+            # Otherwise the tests will report are successful evne on failure.
+            # echo $POSTFIX >> $SCRIPT
             popd
           done
+
+          # Run all the tests in their own shell.  
+          $RUNNER
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -462,46 +462,47 @@ jobs:
       - name: Install wazero
         run: go install ./cmd/wazero
 
-      - name: Run standard library tests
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yaml
```diff
@@ -465,13 +463,21 @@ jobs:
...
 
       - name: Run standard library tests
+        # TODO: delete - temporarily enabled the cache working regardless.
+        continue-on-error: true
         run: |
+          GOOS=$(go env GOOS)
+          if [[ "$GOOS" != "windows" ]]; then
+            MOUNT=/:/
+          else
+            MOUNT=c:\\:/c
```
mathetake, line 473: I guess this shouldn't contain `c`

.github/workflows/integration.yaml
```diff
@@ -465,13 +463,21 @@ jobs:
...
+          fi
           echo "Running $(find ~/sdk/tests -name *.test | wc -l) test binaries"
-
           # Go tests often look for files relative to the source. Change to the corresponding directory.
           for bin in ~/sdk/tests/*.test; do
             dir=$(basename $bin); dir=${dir%.test}; dir=${dir//_/\/}
             pushd ~/sdk/gotip/$dir
-            wazero run -mount=/:/ -env PWD=$PWD $bin -- -test.short -test.v
+            echo $MOUNT $(pwd) ~/sdk/gotip/$dir
+            wazero run -mount=${MOUNT} -env PWD=$(pwd) $bin -- -test.short -test.v
```
mathetake, line 481: and instead we should trim `/c` from `$(pwd)` on windows

mathetake:
```
=== RUN   TestReader
=== RUN   TestReader/gnu.tar
==> wasi_snapshot_preview1.path_filestat_get(fd=3,flags=SYMLINK_FOLLOW,path=C:/Program Files/Git/Users/runneradmin/sdk/gotip/src/archive/tar/testdata/gnu.tar)
<== (filestat=,errno=EINVAL)
    reader_test.go:628: unexpected error: open testdata/gnu.tar: Invalid argument
```

where's this `C:/Program Files/Git` come from..

evacchi:
confirming I can reproduce this locally:

```
~/sdk/gotip/src/archive/tar ~/sdk
/Users/ADMIN/sdk/gotip/src/archive/tar
c:\:/ /Users/ADMIN/sdk/gotip/src/archive/tar /c/Users/ADMIN/sdk/gotip/src/archive/tar
==> wasi_snapshot_preview1.fd_prestat_get(fd=3)
<== (prestat={pr_name_len=1},errno=ESUCCESS)
==> wasi_snapshot_preview1.fd_prestat_dir_name(fd=3)
<== (path=/,errno=ESUCCESS)
==> wasi_snapshot_preview1.fd_prestat_get(fd=4)
<== (prestat=,errno=EBADF)
=== RUN   TestReader
=== RUN   TestReader/gnu.tar
==> wasi_snapshot_preview1.path_filestat_get(fd=3,flags=SYMLINK_FOLLOW,path=C:/Program Files/Git/Users/ADMIN/sdk/gotip/src/archive/tar/testdata/gnu.tar)
<== (filestat=,errno=EINVAL)
    reader_test.go:628: unexpected error: open testdata/gnu.tar: Invalid argument
```

evacchi:
whew. that was a journey. So, this is not done, there are a few elements at play.

1. MSYS2 bash ("Git Bash") is doing some fancy shell expansion on things that look like paths (see https://stackoverflow.com/a/21907301 and https://www.msys2.org/docs/filesystem-paths/); so instead I am collecting all the tests that we need to run into a script and start that script at the end of the GHA. We run in `cmd` on Windows, `sh` otherwise. Basically whenever it encountered things like `/Users` it expanded `/` to `C:\Program Files\Git` (lol)

2. GOROOT on Windows is being set to `%USERPROFILE%\sdk". If I understand correctly... tests are setting this at build-time (?); in any case, when the test is running under wazero, it points to some phantom "C:\Users\$USER\sdk\gotip" which doesn't exist on the vis. We fix this by supplying `-env GOROOT=/Users/$USER/sdk/gotip`; this caused some tests (e.g. tzdata-related) to fail

3. Some path manipulation tests are still failing. Some I thought may have to do with `GOOS` not being set correctly; eg.:
   ```
   func TestReaddirnamesOneAtATime(t *testing.T) {
	t.Parallel()

	// big directory that doesn't change often.
	dir := "/usr/bin"
	switch runtime.GOOS {
	case "android":
		dir = "/system/bin"
	case "iOS":
		wd, err := Getwd()
		if err != nil {
			t.Fatal(err)
		}
		dir = wd
	case "plan9":
		dir = "/bin"
	case "windows":
		dir = Getenv("SystemRoot") + "\\system32"
	}
   ``` 
   but this may be completely misguided. In fact, I think this test is still failing. It is also a good question how it should be fixed :) In any case GOOS == "wasip1" is not handled at all!

4. Some other fs-related tests are failing for other reasons; patches are probably needed. E.g. 
     -  os_test.go:2123: Open(/dev/null): open /dev/null: No such file or directory
     -  os_test.go:1752: Open("/Users/ADMIN/sdk/gotip/go.env/no-such-file", 1) = _, "No such file or directory"; want "Not a directory"
     - os_test.go:1172: unexpected error when renaming from "renameFROM" to "RENAMEfrom": rename renameFROM RENAMEfrom: No such file or directory <--- this probably won't work when the underlying system is case insensitive

I am committing the changes that fix 1-2, but won't for 3-4 for today :)

mathetake:
I think after adding skips on the known failing tests, we could land this

mathetake:
let me send patch on TestReaddirnamesOneAtATime to golang/go

mathetake:
https://go-review.googlesource.com/c/go/+/489575 cc @achille-roussel

mathetake:
omg I didn't know that -run cannot accept negative matching. that's a shame..

mathetake:
ok -test.list cannot be used 🤷

mathetake:
ok I gave up - I will just keep continue-on-error only on windows and raise a tracking issue on the failing tests

mathetake:
raised https://github.com/tetratelabs/wazero/issues/1410 for follow-ups!

mathetake:
I feel like I would split the step for clarity

.github/workflows/integration.yaml
```diff
@@ -475,3 +474,45 @@ jobs:
...
+            pushd ~/sdk/gotip/$dir
+
+            # Trim `/c` from the in-Wasm pwd.
+            IN_WASM_PWD=$(pwd | cut -c 3-)
+            # Convert to a Windows path.
+            bin=`cygpath -w $bin`
+
+            # Create a script with all the tests (do not run yet).
+            echo ${MOUNT} ${IN_WASM_PWD} ~/sdk/gotip/$dir
+            COMMAND="wazero run -mount=${MOUNT} ${EXTRAPARAMS} -hostlogging=filesystem -env PWD=${IN_WASM_PWD} -env GOROOT=${GOROOT} -env GOOS=wasip1 $bin -- -test.short -test.v"
```
mathetake, line 509: I _think_ GOROOT and GOOS won't be needed as envs here, but leave them for now

mathetake:
@evacchi thank you so much for your time and effort!

.github/workflows/integration.yaml
evacchi: GOOS probably not, but I am quite sure GOROOT is. Timezone tests fallback to look into GOROOT in some cases and on windows that pointed to c:\users\...\sdk\gotip which on the VFS of WASI would never make sense (I guess GOROOT defaults to the compile-time value?)

evacchi: (should have left it in a code comment...)

PR #1407, opened 2023-04-26
api: adds CallWithStack to avoid allocations

ncruces:
This implements the API discussed in #1332.

It still needs unit tests, and redoing the benchmarks to confirm the improvements.

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -382,7 +382,7 @@ type Function interface {
 	//   - This is similar to GoFunc, except for using calling functions
 	//     instead of implementing them. Moreover, this is used regardless of
 	//     whether the callee is a host or wasm defined function.
-	CallWithStack(ctx context.Context, stack []uint64) ([]uint64, error)
+	CallWithStack(ctx context.Context, stack []uint64) error
 
 	internalapi.WazeroOnly
 }
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -367,19 +367,15 @@ type Function interface {
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
 
 	// CallWithStack is an optimized variation of Call that saves memory
-	// allocations when the stack slice is large enough to hold the maximum
-	// length of function parameters and results.
+	// allocations when the stack slice is reused across calls.
 	//
+	// Stack length must be at least the max of parameter or result length.
 	// The caller adds parameters in order to the stack, and reads any results
-	// in order from the result stack, except in the error case.
+	// in order from the stack, except in the error case.
 	//
 	// # Notes
 	//
-	//   - When the stack parameter is large enough to hold the max count of
-	//     parameters and results, it is returned. Otherwise, a new slice is
-	//     allocated and returned. This means the caller should always prefer
-	//     using the result slice for subsequent calls.
-	//   - This is similar to GoFunc, except for using calling functions
+	//   - This is similar to GoModuleFunction, except for using calling functions
 	//     instead of implementing them. Moreover, this is used regardless of
 	//     whether the callee is a host or wasm defined function.
 	CallWithStack(ctx context.Context, stack []uint64) error
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -366,6 +366,24 @@ type Function interface {
 	// the end-to-end demonstrations of how these terminations can be performed.
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
 
+	// CallWithStack is an optimized variation of Call that saves memory
+	// allocations when the stack slice is large enough to hold the maximum
+	// length of function parameters and results.
+	//
+	// The caller adds parameters in order to the stack, and reads any results
+	// in order from the result stack, except in the error case.
+	//
+	// # Notes
+	//
+	//   - When the stack parameter is large enough to hold the max count of
+	//     parameters and results, it is returned. Otherwise, a new slice is
+	//     allocated and returned. This means the caller should always prefer
+	//     using the result slice for subsequent calls.
+	//   - This is similar to GoFunc, except for using calling functions
+	//     instead of implementing them. Moreover, this is used regardless of
+	//     whether the callee is a host or wasm defined function.
+	CallWithStack(ctx context.Context, stack []uint64) ([]uint64, error)
+
 	internalapi.WazeroOnly
 }
 
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -373,6 +373,20 @@ type Function interface {
 	// The caller adds parameters in order to the stack, and reads any results
 	// in order from the stack, except in the error case.
 	//
+	// For example, the following reuses the same stack slice to call searchFn
+	// repeatedly saving one allocation per iteration:
+	//
+	//	stack := make([]uint64, 4)
+	//	for i, search := range searchParams {
+	//		// copy the next params to the stack
+	//		copy(stack, search)
+	//		if err := searchFn.CallWithStack(ctx, stack); err != nil {
+	//			return err
+	//		} else if stack[0] == 1 { // found
+	//			return i // searchParams[i] matched!
+	//		}
+	//	}
+	//
 	// # Notes
 	//
 	//   - This is similar to GoModuleFunction, except for using calling functions
--- a/internal/emscripten/emscripten.go
+++ b/internal/emscripten/emscripten.go
@@ -76,8 +76,11 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 		panic(err)
 	}
 
+	// This needs copy (not reslice) because the stack is reused for results.
+	// Consider invoke_i (zero arguments, one result): index zero (tableOffset)
+	// is needed to store the result.
 	tableOffset := wasm.Index(stack[0]) // position in the module's only table.
-	params := stack[1:]                 // parameters to the dynamic function being called
+	copy(stack, stack[1:])              // pop the tableOffset.
 
 	// Lookup the table index we will call.
 	t := m.Tables[0] // Note: Emscripten doesn't use multiple tables
@@ -86,10 +89,8 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 		panic(err)
 	}
 
-	ret, err := m.Engine.NewFunction(idx).Call(ctx, params...)
+	err = m.Engine.NewFunction(idx).CallWithStack(ctx, stack)
 	if err != nil {
 		panic(err)
 	}
-	// if there are any results, copy them back to the stack
-	copy(stack, ret)
 }
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -714,21 +714,11 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 
 // CallWithStack implements the same method as documented on wasm.ModuleEngine.
 func (ce *callEngine) CallWithStack(ctx context.Context, stack []uint64) error {
-	var params, results []uint64
-
-	ft := ce.initialFn.funcType
-	if n := ft.ParamNumInUint64; n > len(stack) {
-		return fmt.Errorf("need %d params, but stack size is %d", n, len(stack))
-	} else {
-		params = stack[:n]
+	params, results, err := wasm.SplitCallStack(ce.initialFn.funcType, stack)
+	if err != nil {
+		return err
 	}
-	if n := ft.ResultNumInUint64; n > len(stack) {
-		return fmt.Errorf("need %d results, but stack size is %d", n, len(stack))
-	} else {
-		results = stack[:n]
-	}
-
-	_, err := ce.call(ctx, params, results)
+	_, err = ce.call(ctx, params, results)
 	return err
 }
 
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -705,15 +705,34 @@ func (ce *callEngine) Definition() api.FunctionDefinition {
 
 // Call implements the same method as documented on wasm.ModuleEngine.
 func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uint64, err error) {
+	ft := ce.initialFn.funcType
+	if n := ft.ParamNumInUint64; n != len(params) {
+		return nil, fmt.Errorf("expected %d params, but passed %d", n, len(params))
+	}
 	return ce.call(ctx, params, nil)
 }
 
 // CallWithStack implements the same method as documented on wasm.ModuleEngine.
-func (ce *callEngine) CallWithStack(ctx context.Context, stack []uint64) (res []uint64, err error) {
-	return ce.call(ctx, stack, stack[:0])
+func (ce *callEngine) CallWithStack(ctx context.Context, stack []uint64) error {
+	var params, results []uint64
+
+	ft := ce.initialFn.funcType
+	if n := ft.ParamNumInUint64; n < len(stack) {
+		return fmt.Errorf("need %d params, but stack size is %d", n, len(stack))
+	} else {
+		params = params[:n]
+	}
+	if n := ft.ResultNumInUint64; n < len(stack) {
+		return fmt.Errorf("need %d results, but stack size is %d", n, len(stack))
+	} else {
+		results = results[:n]
+	}
+
+	_, err := ce.call(ctx, params, results)
+	return err
 }
 
-func (ce *callEngine) call(ctx context.Context, params []uint64, results []uint64) (_ []uint64, err error) {
+func (ce *callEngine) call(ctx context.Context, params, results []uint64) (_ []uint64, err error) {
 	m := ce.initialFn.moduleInstance
 	if ce.ensureTermination {
 		select {
@@ -721,16 +740,9 @@ func (ce *callEngine) call(ctx context.Context, params []uint64, results []uint6
 			// If the provided context is already done, close the call context
 			// and return the error.
 			m.CloseWithCtxErr(ctx)
-			return results, m.FailIfClosed()
+			return nil, m.FailIfClosed()
 		default:
 		}
-	}
-
-	tp := ce.initialFn.funcType
-
-	paramCount := len(params)
-	if tp.ParamNumInUint64 != paramCount {
-		return results, fmt.Errorf("expected %d params, but passed %d", ce.initialFn.funcType.ParamNumInUint64, paramCount)
 	}
 
 	// We ensure that this Call method never panics as
@@ -745,7 +757,8 @@ func (ce *callEngine) call(ctx context.Context, params []uint64, results []uint6
 		}
 	}()
 
-	ce.initializeStack(tp, params)
+	ft := ce.initialFn.funcType
+	ce.initializeStack(ft, params)
 
 	if ce.ensureTermination {
 		done := m.CloseModuleOnCanceledOrTimeout(ctx)
@@ -756,13 +769,11 @@ func (ce *callEngine) call(ctx context.Context, params []uint64, results []uint6
 
 	// This returns a safe copy of the results, instead of a slice view. If we
 	// returned a re-slice, the caller could accidentally or purposefully
-	// corrupt the stack of subsequent calls
-	if tp.ResultNumInUint64 > cap(results) {
-		results = make([]uint64, tp.ResultNumInUint64)
-	} else {
-		results = results[:tp.ResultNumInUint64]
+	// corrupt the stack of subsequent calls.
+	if results == nil && ft.ResultNumInUint64 > 0 {
+		results = make([]uint64, ft.ResultNumInUint64)
 	}
-	copy(results, ce.stack[:tp.ResultNumInUint64])
+	copy(results, ce.stack)
 	return results, nil
 }
 
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -717,15 +717,15 @@ func (ce *callEngine) CallWithStack(ctx context.Context, stack []uint64) error {
 	var params, results []uint64
 
 	ft := ce.initialFn.funcType
-	if n := ft.ParamNumInUint64; n < len(stack) {
+	if n := ft.ParamNumInUint64; n > len(stack) {
 		return fmt.Errorf("need %d params, but stack size is %d", n, len(stack))
 	} else {
-		params = params[:n]
+		params = stack[:n]
 	}
-	if n := ft.ResultNumInUint64; n < len(stack) {
+	if n := ft.ResultNumInUint64; n > len(stack) {
 		return fmt.Errorf("need %d results, but stack size is %d", n, len(stack))
 	} else {
-		results = results[:n]
+		results = stack[:n]
 	}
 
 	_, err := ce.call(ctx, params, results)
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -705,6 +705,15 @@ func (ce *callEngine) Definition() api.FunctionDefinition {
 
 // Call implements the same method as documented on wasm.ModuleEngine.
 func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uint64, err error) {
+	return ce.call(ctx, params, nil)
+}
+
+// CallWithStack implements the same method as documented on wasm.ModuleEngine.
+func (ce *callEngine) CallWithStack(ctx context.Context, stack []uint64) (res []uint64, err error) {
+	return ce.call(ctx, stack, stack[:0])
+}
+
+func (ce *callEngine) call(ctx context.Context, params []uint64, results []uint64) (_ []uint64, err error) {
 	m := ce.initialFn.moduleInstance
 	if ce.ensureTermination {
 		select {
@@ -712,7 +721,7 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 			// If the provided context is already done, close the call context
 			// and return the error.
 			m.CloseWithCtxErr(ctx)
-			return nil, m.FailIfClosed()
+			return results, m.FailIfClosed()
 		default:
 		}
 	}
@@ -721,7 +730,7 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 
 	paramCount := len(params)
 	if tp.ParamNumInUint64 != paramCount {
-		return nil, fmt.Errorf("expected %d params, but passed %d", ce.initialFn.funcType.ParamNumInUint64, paramCount)
+		return results, fmt.Errorf("expected %d params, but passed %d", ce.initialFn.funcType.ParamNumInUint64, paramCount)
 	}
 
 	// We ensure that this Call method never panics as
@@ -748,11 +757,13 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 	// This returns a safe copy of the results, instead of a slice view. If we
 	// returned a re-slice, the caller could accidentally or purposefully
 	// corrupt the stack of subsequent calls
-	if resultCount := tp.ResultNumInUint64; resultCount > 0 {
-		results = make([]uint64, resultCount)
-		copy(results, ce.stack[:resultCount])
+	if tp.ResultNumInUint64 > cap(results) {
+		results = make([]uint64, tp.ResultNumInUint64)
+	} else {
+		results = results[:tp.ResultNumInUint64]
 	}
-	return
+	copy(results, ce.stack[:tp.ResultNumInUint64])
+	return results, nil
 }
 
 // initializeStack initializes callEngine.stack before entering native code.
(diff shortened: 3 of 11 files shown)


internal/engine/compiler/engine.go
```diff
@@ -702,14 +702,23 @@ func (ce *callEngine) Definition() api.FunctionDefinition {
 
 // Call implements the same method as documented on wasm.ModuleEngine.
 func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uint64, err error) {
+	return ce.call(ctx, params, nil)
+}
+
+// CallWithStack implements the same method as documented on wasm.ModuleEngine.
+func (ce *callEngine) CallWithStack(ctx context.Context, stack []uint64) (res []uint64, err error) {
```
achille-roussel, line 709: Supposedly the caller needs to know how many return values there are, would it make sense to accept the params and results as inputs? Changing the signature to:
```go
CallWithStack(ctx context.Context, params, results []uint64) error { ... }
```

ncruces:
Re-ran the benchmarks from #1332 ([`github.com/maragudk/sqlite-benchmark`](https://github.com/maragudk/sqlite-benchmark) adapted to use my [driver](https://pkg.go.dev/github.com/ncruces/go-sqlite3/driver) rather than [mattn](https://pkg.go.dev/github.com/mattn/go-sqlite3)).

Again this benchmark is I/O bound, so improvements to CPU time are negligible, and there's plenty of non-determinism. OTOH, it's not a synthetic made to make this look great.

The reduction in `allocs/op` is stark (this is a rough proxy for the amount of GC work, not necessarily peak memory usage).

Before:
```
goos: darwin
goarch: amd64
pkg: github.com/maragudk/sqlite-benchmark
BenchmarkSelect1-12                                         2000              8643 ns/op           28301 B/op         13 allocs/op
BenchmarkDB_ReadPost/mutex_false-12                         2000             30699 ns/op           26302 B/op         76 allocs/op
BenchmarkDB_ReadPost/mutex_true-12                          2000             29405 ns/op           28953 B/op         77 allocs/op
BenchmarkDB_WritePost/mutex_false-12                        2000            982031 ns/op           20309 B/op         37 allocs/op
BenchmarkDB_WritePost/mutex_true-12                         2000            767473 ns/op            7634 B/op         36 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_false/comment_rate_0.01-12                   2000             43738 ns/op           24638 B/op        218 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_false/comment_rate_0.1-12                    2000            195674 ns/op           31664 B/op       1476 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_false/comment_rate_1-12                      2000          11684738 ns/op          230575 B/op      15605 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_true/comment_rate_0.01-12                    2000             62366 ns/op           96816 B/op        220 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_true/comment_rate_0.1-12                     2000            301926 ns/op          627289 B/op       1633 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_true/comment_rate_1-12                       2000           2598808 ns/op          225410 B/op      15603 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteCommentWithPool/comment_rate_0.01-12                       2000             42212 ns/op           36322 B/op        183 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteCommentWithPool/comment_rate_0.1-12                        2000            202609 ns/op          102164 B/op       1727 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteCommentWithPool/comment_rate_1-12                          2000           2937426 ns/op          225967 B/op      15606 allocs/op
PASS
ok      github.com/maragudk/sqlite-benchmark    40.663s
```

After:
```
goos: darwin
goarch: amd64
pkg: github.com/maragudk/sqlite-benchmark
BenchmarkSelect1-12                                         2000             10291 ns/op           23294 B/op          8 allocs/op
BenchmarkDB_ReadPost/mutex_false-12                         2000             29984 ns/op           26373 B/op         47 allocs/op
BenchmarkDB_ReadPost/mutex_true-12                          2000             33873 ns/op           32017 B/op         49 allocs/op
BenchmarkDB_WritePost/mutex_false-12                        2000            973562 ns/op           24745 B/op         24 allocs/op
BenchmarkDB_WritePost/mutex_true-12                         2000            794107 ns/op            7523 B/op         23 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_false/comment_rate_0.01-12                   2000             38152 ns/op           29614 B/op         74 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_false/comment_rate_0.1-12                    2000            320714 ns/op           22022 B/op        509 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_false/comment_rate_1-12                      2000          10405274 ns/op          159338 B/op       6665 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_true/comment_rate_0.01-12                    2000             68912 ns/op          100010 B/op        117 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_true/comment_rate_0.1-12                     2000            315064 ns/op          617469 B/op        758 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteComment/mutex_true/comment_rate_1-12                       2000           2373769 ns/op          154330 B/op       6664 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteCommentWithPool/comment_rate_0.01-12                       2000             53201 ns/op           46634 B/op        123 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteCommentWithPool/comment_rate_0.1-12                        2000            192710 ns/op           96767 B/op        662 allocs/op
BenchmarkDB_ReadPostAndMaybeWriteCommentWithPool/comment_rate_1-12                          2000           2552369 ns/op          155078 B/op       6666 allocs/op
PASS
ok      github.com/maragudk/sqlite-benchmark    37.593s
```

internal/engine/compiler/engine.go
ncruces: The caller doesn't _really_ need to know how many return values there are.
We could demand that they do.

This how I'm planing to use `CallWithStack`, vs. what I [currently do](https://github.com/ncruces/go-sqlite3/blob/a6226c3b31fe51636a28a7b188df95be76c68914/module.go#L207-L216):
```go
func (m *module) call(fn api.Function, params ...uint64) []uint64 {
	m.arg = append(m.arg[:0], params...)
	r, err := fn.CallWithStack(m.ctx, m.arg)
	if err != nil {
		// The module closed or panicked; release resources.
		m.vfs.Close()
		panic(err)
	}
	return r
}
```

This `call` method wraps all my `api.Function` uses; so `m.arg` is a `[]uint64` reused across all calls into the same `api.Module`. It eventually grows to hold enough space for the maximum number of args/results, and no more temporary allocations happen.

With your API, I'd have to do:
```go
func (m *module) call(fn api.Function, params ...uint64) []uint64 {
	resCount := len(fn.Definition().ResultTypes())
	if resCount > cap(m.arg) {
		m.arg = make([]uint64, resCount)
	}
	m.arg = append(m.arg[:0], params...)
	err := fn.CallWithStack(m.ctx, m.arg, m.arg[:resCount])
	if err != nil {
		// The module closed or panicked; release resources.
		m.vfs.Close()
		panic(err)
	}
	return m.arg[:resCount]
}
```

Slightly more convoluted, but maybe I pigeonholed into my own API.

ncruces:
@inkeliz, since you wanted this, I'd really appreciate your feedback on whether this solves your issue, and what's the more ergonomic API.

I don't want to get too much caught up into my own solution.

Thanks!

internal/engine/compiler/engine.go
achille-roussel: I agree the caller doesn't need to know, but if it doesn't care it can pass a buffer "large enough" (in the common case there will rarely be more than a few results, usually zero or one). If it needs a larger buffer it will be visible on a memory profile.

If we want to keep the returned `[]uint64` I think that's fine, but not mixing the input and output in a single parameter seems like a safer API. If the program really wants to use a single buffer we still can support it (it's the current implementation), but not forcing it would likely be less error prone and cover more use cases.

internal/engine/compiler/engine.go
```diff
@@ -745,11 +754,13 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
...
-		results = make([]uint64, resultCount)
-		copy(results, ce.stack[:resultCount])
+	if tp.ResultNumInUint64 > cap(results) {
+		results = make([]uint64, tp.ResultNumInUint64)
+	} else {
+		results = results[:tp.ResultNumInUint64]
 	}
-	return
+	copy(results, ce.stack[:tp.ResultNumInUint64])
+	return results, nil
```
inkeliz, line 763: There's any specific reason to overwrite the params instead of "concatenate" it.

```go
	if tp.ParamNumInUint64 + tp.ResultNumInUint64 > cap(results) {
		results = make([]uint64, tp.ResultNumInUint64)
	} else {
		results = results[tp.ParamNumInUint64:tp.ParamNumInUint64 + tp.ResultNumInUint64]
	}
	copy(results, ce.stack[:tp.ResultNumInUint64])
	return results, nil
```


I think that will allow to have some function like:

```
param := make([]uint64, 2, 3)
param[0] = 42
param[1] = 7


for {
   res, err := fn.Call(ctx, param)
   // ...
}
```

In that case ` []uint64{42, 7}` will not be overwrite, and can be re-used across multiple calls. The response will be stored into the param[2], which is already allocated.

internal/engine/compiler/engine.go
ncruces: I went with @codefromthecrypt's suggested API. TBH I don't have a strong preference.

The advantage of returning the result is that the `len` of the returned slice is the number of returned values and a middleman doesn't need to know.

How about we expose the private `call` method from this PR? Would that be a decent API? You can pass independent input/output slices, or the same slice (I'm doing both here). If the output slice does not have enough capacity for the result, a new one is returned. A little like `append` and `Append` methods of the standard lib.

I'd honestly like to see more people trying this in other codebases to be sure.

internal/engine/compiler/engine.go
ncruces: No specific reason.

Basically, that's what a single argument stack API implies IMO: pop the arguments push the results. At least that's my interpretation of @codefromthecrypt's suggested API.

You and @achille-roussel seem to be converging on your `CallTo` suggested API: arguments are independent from results.

I honestly don't know which is better.

I'd like to see more usage scenarios to understand the advantages/disadvantages of either design.

Do you have examples you can share of how you'd use either API?

inkeliz: I can't share the actual code, yet. But, I currently use an "Update" function that is called multiple times per second per instance. This function takes in two parameters, `id` and `offset`, and returns a uint32 value. The id is unique per instance, while the offset is rarely updated. So, I could re-use `params` between multiple calls. 

In that case, I would need to do something like: params[0] = id and params[1] = offset for each call.

I think `CallWithStack(ctx context.Context, params, results []uint64)` allows both worlds to exists:

If I would like to reduce some memory consumption and overwrite I use: `CallWithStack(ctx, params, params[:0])`
If I would like to re-use the params for future calls then I can use:  `CallWithStack(ctx, params, results)`

_In both cases, assuming that params/results have enough capacity._

ncruces:
I think in #1332 and here we've established the usefulness of this, but can't settle for an API design (which in this case, I find, is understandably hard).

I'm happy to keep this updated (tests passing) until more people are ready to share instances where they'd use this, so we can improve ergonomics and settle on a design.

Or someone makes the call, because the perfect is the enemy of the good.

I'd use this in exactly one call site, so I'm probably not the right one to make the call, any design works for me.

internal/engine/compiler/engine.go
codefromthecrypt: I really don't like being different from the other API especially as this is an advanced one anyway. It seems we have more opportunities to optimize internally as well, when the shape matches the host side. Having something both unlike api.Function an also unlike api.GoXFunction seems far from ideal iotw.

codefromthecrypt:
My summary comment is that the api surface of a codebase should have coherence with itself as a feature, even if it wasn't also optimizable due to this.

It would almost be self-deprecating humor to have to tell people the secret way to make our API whose goal is to reduce allocations, do less via things like this: `CallWithStack(ctx, params, params[:0])` when we already have an API design the same people will have to understand to reduce allocations host side.

When designing something explicitly for advanced users, we shouldn't fear them being able to do min(len(params), len(results)) if trying to size things. The exact concept is required for the host side, even if they aren't the ones sizing the slice.

All this said, If others feel like they really want the third API style and that's better than two, go for it. I've said my piece and feel folks here can grok where I'm coming from.

codefromthecrypt:
If we do go for api style 3, please don't link it to GoModuleFunc, as it no longer has any relationship to it.

anuraaga:
I happened to find this and #1332 - as a user with go-re2, something like this would definitely be useful. For example, with WAF, hundreds of regex are executed per request, which is hundreds of wasm calls. Reducing allocations would be helpful if possible, though given everything else most logic that uses big regex does, it wouldn't be a greatly sad event if nothing was added as it's still within margin of error in the big scheme of things I guess.

I see mention of "api style 3" but wasn't actually able to find 3 options so not sure which it is, but personally I find adapting `GoModuleFunc` to be very natural, I assume that option looks like `func (f Module) CallWithStack(ctx context.Context, stack []uint64) error`. If so, I would use it in go-re2 like this

https://github.com/wasilibs/go-re2/blob/main/internal/re2_wazero.go#L212

becomes

```go
func match(re *Regexp, s cString, matchesPtr uintptr, nMatches uint32) bool {
	ctx := context.Background()
        stack := [8]uint64{uint64(re.ptr), uint64(s.ptr), uint64(s.length), 0, uint64(s.length), 0, uint64(matchesPtr), uint64(nMatches)}
	if err := re.abi.cre2Match.CallWithStack(ctx, stack[:]); err != nil {
		panic(err)
	}

	return stack[0] == 1
}
```

For a function that takes no parameters and returns a result, `var stack [1]uint64`.

I really enjoy the fact that the result slice isn't returned as it makes error handling simpler - for me, `max(len(params), len(results))` to size things would not be a problem at all especially if it gives me this code benefit. It's a very minor point though.

I'm sure any of the other options would work fine too, to quote @ncruces "any design works for me". Thought it might be helpful to have another code example of where this API would end up being used.

codefromthecrypt:
@anuraaga thanks for the feedback! I should have linked api style 3 is from this thread, and worth a look in case changes your opinion https://github.com/tetratelabs/wazero/pull/1407#discussion_r1178069476

anuraaga:
@codefromthecrypt Ah thanks for the link, so that's style 3. I'd still use it, but it would be a bit annoying to have to add `var out [1]uint64` on every invocation. I suppose lines-of-code could be returned by inlining the array declaration `[8]uint64{a,b,c,d,e...}[:]` but I'd probably avoid it since declaring an array and slicing it in the same code looks confusing to me. So for go-re2, it's probably not optimal, but I do see mention of some use cases like invoking in a loop where it may be so understand the motivation. go-re2 is just one example so meant as a data point more than anything else.

inkeliz:
> but it would be a bit annoying to have to add var out [1]uint64 on every invocation

I think that is the "pain-point" of the 3rd design. You don't _need_ `var out [1]uint64`, but that is not clear, but could be documented. I guess "Style 3" can be used in three (or more) ways:

Option A:

```go
var out [1]uint64
var param [8]uint64{a,b,c,d,e...}

CallWithStack(ctx, param[:], out[:])
```

Option B:

```go
var param [9]uint64{a,b,c,d,e...}
CallWithStack(ctx, param[:8], out[8:])

out := param[8] // output is the last item, not overwrites
```

Option C:

```go
var param [8]uint64{a,b,c,d,e...}

CallWithStack(ctx, param[:], param[:0])

out := param[0] // output is the first, and overwrite the first param
```

In all cases, we are potentially reducing allocations. Currently, this PR forces you to use the "Option C" (which overwrites), while the old CallTo (or the "Style 3") makes it more flexible. Is that flexibility worth it? I'm not sure. _The only specific case is when you do multiple calls with the same arguments_.

The `[:0]` is not so weird, that is used in some std-lib packages, such as `Seal` on AEAD (https://pkg.go.dev/crypto/cipher#AEAD).

anuraaga:
Ah got it - yeah I agree option B there would have similar writability, it didn't come to mind from the API definition but with examples it would make sense. Thanks for clarifying it 🙏

ncruces:
Thanks, @inkeliz, @anuraaga, this discussion is super helpful. As I expected I'm pigeonholed into my own usage, where (for other reasons) I concentrated all my `api.Function.Call`s into a wrapper.

I agree that an API where each call site can use stack allocated arrays (this requires constant size) to make the call is probably significantly more generally useful, _if these don't escape to heap_, in other words, _if we can make it work_.

Unfortunately, I don't think it's that easy: `CallWithStack`[^1] is an interface call, so its arguments will escape to heap, regardless. Right?

[^1]: whatever the name, with this API `CallTo` is more appropriate

anuraaga:
I'm not particularly authoritative but my understanding is while an interfere implies any struct implementing it that is used as an interface has it's contents on the heap, function calls themselves are not related. Definitely would expect being able to keep the arguments on the stack for this API so hope it's possible!

achille-roussel:
@ncruces is right that since it's an interface method, Go cannot currently determine that the values do not escape and has to put them on the heap.

I believe the goal of this PR is to offer a way to reuse the slices, not to allow stack-allocation; for this, we would need to use a concrete type allowing the compiler to track the lifetime of method arguments.

achille-roussel:
@ncruces @codefromthecrypt I buy the argument of API consistency, I didn't see the relation with `api.GoModuleFunction` at first but keeping it coherent makes most sense 👍 Thanks for spending the time to share your perspectives!

Let's go with this?
```go
CallWithStack(ctx context.Context, stack []uint64) error
```

ncruces:
@anuraaga please see [this](https://go.dev/play/p/sQEJqww59ug).

Building it with go 1.20.3:
```sh
go build -gcflags=-m buh.go   
# command-line-arguments
./main.go:9:6: can inline (*impl).Call
./main.go:13:6: can inline directCall
./main.go:16:8: inlining call to (*impl).Call
./main.go:19:6: can inline ifaceCall
./main.go:27:12: inlining call to directCall
./main.go:28:11: inlining call to ifaceCall
./main.go:27:12: inlining call to (*impl).Call
./main.go:28:11: devirtualizing v.Call to *impl
./main.go:9:19: in does not escape
./main.go:9:32: out does not escape
./main.go:13:17: v does not escape
./main.go:19:16: leaking param: v
./main.go:20:6: moved to heap: in
./main.go:21:6: moved to heap: out
```

On an interface call all arguments escape to heap. I'd really like to support this, but can't see how.

The idea of this PR is not to prevent them from escaping, they always do. Rather it is to allow you to reuse an heap allocated slice you have lying around, somewhere. If concurrent calls into a module are not OK, you can keep a single slice per module for all calls, otherwise, probably a `sync.Pool` is better.

ncruces:
@achille-roussel, thanks! Slices are subtle, though.

At entry of `CallWithStack` in https://github.com/tetratelabs/wazero/pull/1407#issuecomment-1526003485, do we assume:
1. `len(stack) == len(args) && cap(stack) >= len(res)`
2. `len(stack) == max(len(args), len(res))`
3. `len(stack) >= max(len(args), len(res))`
4. …?

I get the parallel with `GoModuleFunction`, and agree consistency reduces cognitive load, so thought experiment…

Would it be fair to consider a `GoModuleFunc` that uses `CallWithStack` to forward a "host call" to a "guest call" to either the same or another module?

If so, if that's the "consistency" we want, according to the [docs](https://pkg.go.dev/github.com/tetratelabs/wazero/api#GoModuleFunction), we want option **2** (or maybe **3**) above.

If everyone approves of this I'll make the change.
cc: @codefromthecrypt

codefromthecrypt:
I think we can change the comment on GoModuleFunc so that it says "at least" not "equals" max blah blah https://github.com/tetratelabs/wazero/pull/1407#issuecomment-1526124384. Then we can be consistent on this point between the two sides of the API.

api/wasm.go
```diff
@@ -355,6 +355,24 @@ type Function interface {
...
+	// The caller adds parameters in order to the stack, and reads any results
+	// in order from the result stack, except in the error case.
+	//
+	// # Notes
+	//
+	//   - When the stack parameter is large enough to hold the max count of
+	//     parameters and results, it is returned. Otherwise, a new slice is
+	//     allocated and returned. This means the caller should always prefer
+	//     using the result slice for subsequent calls.
+	//   - This is similar to GoFunc, except for using calling functions
```
codefromthecrypt, line 372: Let's use GoModuleFunction as it has more docs on it.

```suggestion
	//   - This is similar to GoModuleFunction, except for using calling functions
```

api/wasm.go
```diff
@@ -355,6 +355,24 @@ type Function interface {
 	// WithCloseOnContextDone on wazero.RuntimeConfig for detail. See examples in context_done_example_test.go for
 	// the end-to-end demonstrations of how these terminations can be performed.
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
+
+	// CallWithStack is an optimized variation of Call that saves memory
+	// allocations when the stack slice is large enough to hold the maximum
+	// length of function parameters and results.
+	//
+	// The caller adds parameters in order to the stack, and reads any results
+	// in order from the result stack, except in the error case.
```
codefromthecrypt, line 364: Since this is an interface type, we can't use _Example tests to embed an example. Please add an example block under here, especially showing usage of the result value in a subsequent call

internal/engine/compiler/engine.go
```diff
@@ -702,14 +702,23 @@ func (ce *callEngine) Definition() api.FunctionDefinition {
...
 func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uint64, err error) {
+	return ce.call(ctx, params, nil)
+}
+
+// CallWithStack implements the same method as documented on wasm.ModuleEngine.
+func (ce *callEngine) CallWithStack(ctx context.Context, stack []uint64) (res []uint64, err error) {
+	return ce.call(ctx, stack, stack[:0])
+}
+
+func (ce *callEngine) call(ctx context.Context, params []uint64, results []uint64) (_ []uint64, err error) {
```
codefromthecrypt, line 713: in the future, who knows.. it might pass the stack further down. For now, this is nice as it keeps the patch smaller. One nit:

```suggestion
func (ce *callEngine) call(ctx context.Context, params, results []uint64) (_ []uint64, err error) {
```

anuraaga:
> On an interface call all arguments escape to heap. I'd really like to support this, but can't see how.

Thanks for sharing this! Learned something new. It's too bad that stack allocation can't work but it is what it is.

Luckily, currently wasm calls are guarded by a lock so it'll be easy to reuse a heap slice. Eventually when atomics are supported I'm looking forward to removing that lock and per-regex module instantiation and then I'd probably have to see if sync.Pool performs better than the small allocation. But that'll be a long ways away.

codefromthecrypt (approved):
Great, and I offer to write the missing "Advanced API" section of [docs](https://wazero.io/docs/) afterwards. The api parity makes this easier as we can say. 


"WebAssembly is a stack based virtual machine. wazero includes advanced, stack based APIs for defining or calling functions..."


Some small requests:

Can you parameterize BenchmarkHostFunctionCall so we can see any overall diff before/after?

Also, maybe we should use this at least one non-test place. For example, in runtime.go

[_, err = start.Call(ctx)](https://github.com/tetratelabs/wazero/blob/main/runtime.go#L324)

Basically, I would assume any v_v type should use this with an empty slice, right? Possibly, we should automatically dispatch like this, since we know the type of the function before it is called, but anyway..

internal/engine/compiler/engine.go
```diff
@@ -717,13 +745,6 @@ func (ce *callEngine) Call(ctx context.Context, params ...uint64) (results []uin
 		}
 	}
 
-	tp := ce.initialFn.funcType
-
-	paramCount := len(params)
-	if tp.ParamNumInUint64 != paramCount {
```
ncruces, line 723: This checked is moved to the wrappers, since it's different for each one.
This is a behavior change: arguments are checked before module closure.
Is that a problem?

codefromthecrypt:
> Not necessarily the best example. A v_v function can use Call without allocations: the params slice is nil, when varargs are empty; we take care never to allocate for empty results (and this is actually tested).

good point and this is why you should think about my suggestions before implementing them ;)

internal/engine/compiler/engine.go
```diff
@@ -705,6 +705,34 @@ func (ce *callEngine) Definition() api.FunctionDefinition {
...
+	}
+	return ce.call(ctx, params, nil)
+}
+
+// CallWithStack implements the same method as documented on wasm.ModuleEngine.
+func (ce *callEngine) CallWithStack(ctx context.Context, stack []uint64) error {
+	var params, results []uint64
+
+	ft := ce.initialFn.funcType
+	if n := ft.ParamNumInUint64; n > len(stack) {
```
codefromthecrypt, line 720: I extracted the logic here and in interpreter into a different PR to test it https://github.com/tetratelabs/wazero/pull/1414

api/wasm.go
```diff
@@ -366,6 +366,20 @@ type Function interface {
 	// the end-to-end demonstrations of how these terminations can be performed.
 	Call(ctx context.Context, params ...uint64) ([]uint64, error)
 
+	// CallWithStack is an optimized variation of Call that saves memory
+	// allocations when the stack slice is reused across calls.
+	//
+	// Stack length must be at least the max of parameter or result length.
```
codefromthecrypt, line 372: I'd love to see an example prior to merge.

```go
// # Example
// ...
```

codefromthecrypt (approved):
Nice and concise!

internal/engine/compiler/engine.go
ncruces: Thanks for the refactor, saves me writing a bunch of test code.

api/wasm.go
ncruces: Not forgotten. Just thinking about a good example that passes the message with only "local" _context_ (to reduce cognitive load).

Thinking about what @anuraaga said, maybe the simplest way to transmit the idea without resorting to global variables, is this: a host calling a guest function 100 times in a loop, and reusing the stack for (just) those 100 calls. Like matching text against 100 regexes.

Wdyt?

internal/emscripten/emscripten.go
```diff
@@ -86,10 +89,8 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 		panic(err)
 	}
 
-	ret, err := m.Engine.NewFunction(idx).Call(ctx, params...)
+	err = m.Engine.NewFunction(idx).CallWithStack(ctx, stack)
```
ncruces, line 92: Found the non-test/non-example usage. Also neatly validates the design: we're using `CallWithStack` to implement a `GoModuleFunction`.

This should have a real world impact on @jerbob92's emscripten uses.

codefromthecrypt: hehe except I'm not sure he uses this part anymore. can you also hit this one? https://github.com/tetratelabs/wazero/blob/main/internal/emscripten/emscripten.go#L89

api/wasm.go
codefromthecrypt: sounds good, sounds something like a index correct version of this.

```
stack = ...
for i, search := range searchParams {

// copy the next params to the stack
copy(stack, search)
if err := searchFn.CallWithStack(ctx, stack) {
return err
} else if stack[0] == 1 { // found
return i // searchParams[i] matched!
}
}
```

internal/emscripten/emscripten.go
codefromthecrypt: disregard I hadn't coffee yet ;)

internal/emscripten/emscripten.go
```diff
@@ -76,8 +76,11 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 		panic(err)
 	}
 
+	// This needs copy (not reslice) because the stack is reused for results.
```
codefromthecrypt, line 79: great comment!

codefromthecrypt (approved):
Mark ready for review when you are!

ncruces:
> Can you parameterize BenchmarkHostFunctionCall so we can see any overall diff before/after?

```sh
% go test -bench=BenchmarkHostFunctionCall -benchmem
goos: darwin
goarch: amd64
BenchmarkHostFunctionCall/call_go_reflect_host-12                         515949              2197 ns/op             160 B/op          8 allocs/op
BenchmarkHostFunctionCall/call_go_reflect_host_with_stack-12              529288              2200 ns/op             144 B/op          6 allocs/op
BenchmarkHostFunctionCall/call_go_host-12                                9936541               120.9 ns/op            16 B/op          2 allocs/op
BenchmarkHostFunctionCall/call_go_host_with_stack-12                    13865826                85.96 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/tetratelabs/wazero/internal/integration_test/bench   5.955s
```

codefromthecrypt (approved):
:shipit:

mathetake (approved):
💯 🏅 🚀

PR #1408, opened 2023-04-27
ci: run -race on PRs as well

mathetake:
we recently missed two race bugs introduced via PRs but detected on the main branch.
The reason is that previously we had run tests with -race only on main branch.

This enables the race detection on PRs as well.


mathetake:
only taks 5m I guess is acceptable given that we now have more slow integration tests

codefromthecrypt (approved):
Thanks, I noticed this at some point, just thought I was confused!

Issue #1409, opened 2023-04-27
compiler: perform BuiltinFunctionCheckExitCode natively


mathetake:
wazeroir's [BuiltinFunctionCheckExitCode operation](https://github.com/tetratelabs/wazero/blob/main/internal/wazeroir/operations.go#L714-L715) has been introduced in the context cancelation integration in #1108. Basically, this operation is supposed to perform the check on `ModuleInstance.Closed` value, and exit the execution if the value is not zero.

Currently, the compiler's implementation is pretty naive like simply exiting the native world and call `ModuleInstance.FailIfClosed`.
https://github.com/tetratelabs/wazero/blob/86444c67a37dbf9e693ae5b365901f64968d9025/internal/engine/compiler/impl_amd64.go#L271-L272

https://github.com/tetratelabs/wazero/blob/86444c67a37dbf9e693ae5b365901f64968d9025/internal/engine/compiler/engine.go#L1012-L1016

but this comes with the cost - `BuiltinFunctionCheckExitCode` is [inserted at all of the `loop` instructions](https://github.com/tetratelabs/wazero/blob/86444c67a37dbf9e693ae5b365901f64968d9025/internal/wazeroir/compiler.go#L475) and each time we call a Go function. `FailIfClosed` function checks the exit code atomically, but it is not necessary as in anyway the implementation is not synchronous (e.g. we don't check the code until the Wasm VM reaches loop instruction even when the context reaches timeout 10 years ago in extreme case).

Given that assumption we do not need to check it synchronously, we could optimize the translation of `BuiltinFunctionCheckExitCode` by simply checking the exit code natively. 

The proposed change should come with 
- Native check of `ExitCode` in `compiler.compileBuiltinFunctionCheckExitCode` function for both arm64 and amd64
    - native code has access to it via `callEngine.moduleInstance`
- The benchmark on the cost of context cancelation integration

mathetake:
@evacchi could you take on after #1406

evacchi:
Nice!👍

evacchi:
If I understand correctly this boils down to `c.compileExitFromNativeCode(exitCode)` when `exitCode` is nonzero; where `exitCode` is extracted from `(*moduleInstance).Closed` (`closed >> 32`); `Closed` is at the address `callEngineModuleContextModuleInstanceOffset` because it is at offset `0` of the struct

mathetake:
[...]

PR #1413, opened 2023-04-28
compiler: check exit code in native code

evacchi:
addresses #1409 

- compiler: check exit code in native code
- wip
- amd64


internal/engine/compiler/engine.go
```diff
@@ -1009,13 +1009,13 @@ entry:
...
-				// native code cannot be preempted and that means it can block forever if there are not
-				// enough OS threads (which we don't have control over).
-				if err := m.FailIfClosed(); err != nil {
-					panic(err)
-				}
+			case builtinFunctionIndexExitUnconditionally:
+				// Leave with the given exit code: the check has been performed in native code.
+				//if err := m.FailIfClosed(); err != nil {
+				//	panic(err)
+				//}
```
evacchi, line 1016: garbage

internal/engine/compiler/impl_arm64.go
```diff
@@ -490,13 +490,27 @@ func (c *arm64Compiler) setLocationStack(newStack *runtimeValueLocationStack) {
...
+	if err := c.compileCallGoFunction(nativeCallStatusCodeCallBuiltInFunction, builtinFunctionIndexExitUnconditionally); err != nil {
 		return err
 	}
 
-	// After return, we re-initialize reserved registers just like preamble of functions.
-	c.compileReservedStackBasePointerRegisterInitialization()
-	c.compileReservedMemoryRegisterInitialization()
+	//// After return, we re-initialize reserved registers just like preamble of functions.
+	//c.compileReservedStackBasePointerRegisterInitialization()
+	//c.compileReservedMemoryRegisterInitialization()
```
evacchi, line 510: I think this is not necessary since we are exiting, thus we don't need to clean after ourselves

internal/integration_test/engine/adhoc_test.go
```diff
@@ -65,7 +65,7 @@ func runAllTests(t *testing.T, tests map[string]func(t *testing.T, r wazero.Runt
 		name := name   // pin
 		testf := testf // pin
 		t.Run(name, func(t *testing.T) {
-			t.Parallel()
+			//t.Parallel()
```
evacchi, line 68: with this enabled, the context-related tests hang on arm64. It doesn't fail on amd64 but it may be due to rosetta emulation adding a little bit of delay (I haven't tested on native amd64 yet)

evacchi:
so there _is_ a race after all 🤔

internal/engine/compiler/engine.go
```diff
@@ -1009,13 +1010,11 @@ entry:
...
-			case builtinFunctionIndexCheckExitCode:
-				// Note: this operation must be done in Go, not native code. The reason is that
-				// native code cannot be preempted and that means it can block forever if there are not
-				// enough OS threads (which we don't have control over).
-				if err := m.FailIfClosed(); err != nil {
-					panic(err)
-				}
+			case builtinFunctionIndexExitUnconditionally:
+				// Leave with the given exit code: the check has been performed in native code.
+				closed := atomic.LoadUint64(&m.Closed)
```
evacchi, line 1015: this is likely just masking the underlying issue: it still fails randomly with `-race` (not surprising, since it's failing with `t.Parallel()`)

evacchi: and the root cause seems to be exactly read/writes to this field

(prev run: https://github.com/tetratelabs/wazero/actions/runs/4831243660/jobs/8608429404)

evacchi: actually, I may be wrong, because inspecting that further, it kind of looks like the failure is at `wasm.(*ModuleInstance).closeModuleOnCanceledOrTimeout()`; in that case that would mean that the race is on the `select/case` ? is that even possible, or is the error misleading?

internal/engine/compiler/impl_amd64.go
```diff
@@ -269,13 +269,30 @@ func (c *amd64Compiler) label(label wazeroir.Label) *amd64LabelInfo {
...
+
+	jump := c.assembler.CompileJump(amd64.JEQ)
+
+	if err := c.compileCallBuiltinFunction(builtinFunctionIndexExitUnconditionally); err != nil {
 		return err
 	}
 
 	// After the function call, we have to initialize the stack base pointer and memory reserved registers.
 	c.compileReservedStackBasePointerInitialization()
 	c.compileReservedMemoryPointerInitialization()
```
mathetake, line 292: these lines can be deleted (same goes for arm64 as well) since we don't exit the machine code world anymore in this branch

internal/engine/compiler/impl_amd64.go
```diff
@@ -269,13 +269,30 @@ func (c *amd64Compiler) label(label wazeroir.Label) *amd64LabelInfo {
...
 func (c *amd64Compiler) compileBuiltinFunctionCheckExitCode() error {
-	if err := c.compileCallBuiltinFunction(builtinFunctionIndexCheckExitCode); err != nil {
+	dstReg, err := c.allocateRegister(registerTypeGeneralPurpose)
+	if err != nil {
+		panic("BUG: cannot find free register")
+	}
+
+	c.assembler.CompileMemoryToRegister(amd64.MOVQ,
+		amd64ReservedRegisterForCallEngine, callEngineModuleContextModuleInstanceOffset, dstReg)
+	c.assembler.CompileMemoryToRegister(amd64.MOVQ, dstReg, 0, dstReg)
```
mathetake, line 279: this zero should be defined as const like `moduleInstanceGlobalsOffset` and must be tested in compiler_test.go like others

internal/engine/compiler/impl_amd64.go
```diff
@@ -269,13 +269,30 @@ func (c *amd64Compiler) label(label wazeroir.Label) *amd64LabelInfo {
 
 // compileBuiltinFunctionCheckExitCode implements compiler.compileBuiltinFunctionCheckExitCode for the amd64 architecture.
 func (c *amd64Compiler) compileBuiltinFunctionCheckExitCode() error {
-	if err := c.compileCallBuiltinFunction(builtinFunctionIndexCheckExitCode); err != nil {
+	dstReg, err := c.allocateRegister(registerTypeGeneralPurpose)
+	if err != nil {
+		panic("BUG: cannot find free register")
```
mathetake, line 274:
```suggestion
		return err
```

internal/engine/compiler/impl_arm64.go
mathetake: yes can be deleted

internal/engine/compiler/impl_arm64.go
```diff
@@ -490,13 +490,27 @@ func (c *arm64Compiler) setLocationStack(newStack *runtimeValueLocationStack) {
...
 func (c *arm64Compiler) compileBuiltinFunctionCheckExitCode() error {
-	if err := c.compileCallGoFunction(nativeCallStatusCodeCallBuiltInFunction, builtinFunctionIndexCheckExitCode); err != nil {
+	dstReg := arm64ReservedRegisterForTemporary
+
+	// dstReg := callEngine.moduleContext.moduleInstance
+	c.assembler.CompileMemoryToRegister(arm64.LDRD,
+		arm64ReservedRegisterForCallEngine, callEngineModuleContextModuleInstanceOffset,
+		dstReg)
+	// dstReg = *dstReg; i.e. (*(callEngine.moduleContext.moduleInstance)).Closed
+	c.assembler.CompileMemoryToRegister(arm64.LDRD, dstReg, 0, dstReg)
```
mathetake, line 500: same on `0`

internal/engine/compiler/impl_amd64.go
mathetake: no that should be wrong. Could you just delete this panic again?

mathetake (commented):
also I would suggest adding `TestCompiler_compileBuiltinFunctionCheckExitCode`  in compiler_controlfow_test.go

mathetake:
Closing as I noted in https://github.com/tetratelabs/wazero/issues/1409#issuecomment-1530536903

PR #1414, opened 2023-04-29
Adds SplitCallStack for use in CallWithStack

codefromthecrypt:
This adds an internal function `wasm.SplitCallStack` for use in #1407. This is separate because the diff is a lot larger than the destination change ;)

### DIFF
--- a/internal/wasm/func_validation.go
+++ b/internal/wasm/func_validation.go
@@ -1935,3 +1935,19 @@ var (
 	blockType_v_funcref   = &FunctionType{Results: []ValueType{ValueTypeFuncref}, ResultNumInUint64: 1}
 	blockType_v_externref = &FunctionType{Results: []ValueType{ValueTypeExternref}, ResultNumInUint64: 1}
 )
+
+// SplitCallStack returns the input stack resliced to the count of params and
+// results, or errors if it isn't long enough for either.
+func SplitCallStack(ft *FunctionType, stack []uint64) (params []uint64, results []uint64, err error) {
+	if n := ft.ParamNumInUint64; n > len(stack) {
+		return nil, nil, fmt.Errorf("need %d params, but stack size is %d", n, len(stack))
+	} else if n > 0 {
+		params = stack[:n]
+	}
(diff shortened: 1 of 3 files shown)


internal/wasm/func_validation.go
```diff
@@ -1935,3 +1935,19 @@ var (
...
+
+// SplitCallStack returns the input stack resliced to the count of params and
+// results, or errors if it isn't long enough for either.
+func SplitCallStack(ft *FunctionType, stack []uint64) (params []uint64, results []uint64, err error) {
+	if n := ft.ParamNumInUint64; n > len(stack) {
+		return nil, nil, fmt.Errorf("need %d params, but stack size is %d", n, len(stack))
+	} else if n > 0 {
+		params = stack[:n]
+	}
+	if n := ft.ResultNumInUint64; n > len(stack) {
```
mathetake, line 1947: should be able to reuse `len(stack)`  above

codefromthecrypt: good idea

internal/wasm/func_validation.go
```diff
@@ -1939,13 +1939,13 @@ var (
...
-	if n := ft.ParamNumInUint64; n > len(stack) {
-		return nil, nil, fmt.Errorf("need %d params, but stack size is %d", n, len(stack))
+	if n, len := ft.ParamNumInUint64, len(stack); n > len {
+		return nil, nil, fmt.Errorf("need %d params, but stack size is %d", n, len)
 	} else if n > 0 {
 		params = stack[:n]
 	}
-	if n := ft.ResultNumInUint64; n > len(stack) {
-		return nil, nil, fmt.Errorf("need %d results, but stack size is %d", n, len(stack))
+	if n, len := ft.ResultNumInUint64, len(stack); n > len {
```
mathetake, line 1947: no I meant you should be able to calculate len(stack) once

mathetake: in entire this function

codefromthecrypt: sorry you are right!

mathetake (approved):

PR #1416, opened 2023-04-30
CONTRIBUTING.md: add bullet point about the merge process

achille-roussel:


CONTRIBUTING.md
```diff
@@ -57,8 +57,14 @@ You can add the sign off when creating the git commit via `git commit -s`.
...
-* During the review, address the comments and commit the changes _without_ squashing the commits.
-This facilitates incremental reviews since the reviewer does not go through all the code again to
-find out what has changed since the last review.
+* A single approval is sufficient to merge. If a reviewer asks for
+changes in a PR they should be addressed before the PR is merged,
+even if another reviewer has already approved the PR.
+* During the review, address the comments and commit the changes
+_without_ squashing the commits. This facilitates incremental reviews
+since the reviewer does not go through all the code again to find out
+what has changed since the last review.
```
mathetake, line 66: 👍

ncruces:
I'm not well versed in the GitHub flow, so maybe guilty of this myself.

When synching with main, is it better to rebase (what I usually do) or merge changes? Do we want to make that clear?

codefromthecrypt:
> When synching with main, is it better to rebase (what I usually do) or merge changes? Do we want to make that clear?

I think some projects ask contributors to tidy up into a single commit, rebased onto latest main. So, I think we can ask for that, but as some struggle with git, not make it required. Main thing is that contributors ensure any commit is cleaned up, ideally from the submitter.

ncruces:
If I understood @achille-roussel, this is asking for the opposite: if you change things, add commits, don't squash+force. Which makes sense, merging the PR does that, and you can always see the entire diff. This helps partial reviews.

My question was, when updating the PR requires synching main (because of a merge conflict, or my latest PR @codefromthecrypt factored parts of it out, and I was supposed to use them) is rebase+force preferred, or merge?

Sorry for restating the question, maybe I didn't understand your answer.

codefromthecrypt:
Sorry, I'll put line comments as agree it is confusing without context!

CONTRIBUTING.md
```diff
@@ -57,8 +57,14 @@ You can add the sign off when creating the git commit via `git commit -s`.
 
 ## Code Reviews
 
-* A single approval is sufficient to merge. If a reviewer asks for changes in a PR they should be
-addressed before the PR is merged, even if another reviewer has already approved the PR.
-* During the review, address the comments and commit the changes _without_ squashing the commits.
-This facilitates incremental reviews since the reviewer does not go through all the code again to
-find out what has changed since the last review.
+* A single approval is sufficient to merge. If a reviewer asks for
```
codefromthecrypt, line 60:
```suggestion
* The pull request title should describe what the change does and not embed issue numbers.
The pull request should only be blank when the change is minor. Any feature should include
a description of the change and what motivated it. If the change or design changes through
review, please keep the title and description updated accordingly.
* A single approval is sufficient to merge. If a reviewer asks for
```

CONTRIBUTING.md
```diff
@@ -57,8 +57,14 @@ You can add the sign off when creating the git commit via `git commit -s`.
...
+changes in a PR they should be addressed before the PR is merged,
+even if another reviewer has already approved the PR.
+* During the review, address the comments and commit the changes
+_without_ squashing the commits. This facilitates incremental reviews
+since the reviewer does not go through all the code again to find out
+what has changed since the last review.
+* Commits are squashed prior to merging a pull request, using the title
+as commit message by default. Maintainers may request contributors to
+edit the pull request tite to ensure that it remains descriptive as a
+commit message.
```
codefromthecrypt, line 70:
```suggestion
commit message. Alternatively, maintainers may change the commit message directly.
```

CONTRIBUTING.md
codefromthecrypt, line 66:
```suggestion
what has changed since the last review. When a change goes out of sync with main,
please rebase and force push, keeping the original commits where practical.
```

codefromthecrypt (approved):
Good point by @ncruces I don't particularly like having to describe a one way especially as I think there's zero value in retaining commits like "fuzz" "fix" etc. however, if we are specifying things it is harder to enumerate nuance like this and easier to focus on the worst case (big change where the diffs matter). Meanwhile, let's as a team try not to be cops. Writing things down is good, but I don't want us to become pedantic about whether commits are allowed to be squashed by the submitter. I think in many cases, they should, except large diffs or change that lasts a longer time.

codefromthecrypt:
Added a comment. summary is I'm fine with this.. I think being disallowed to squash things would actually irritate me as not many projects require this, and some actually ask you to. As long as we don't get too harsh on people, and we won't, the guidance is fine.

achille-roussel:
Uh... looks like committing the suggestions from the UI didn't sign the commits 😕

codefromthecrypt:
yeah just click on the DCO when this happens

CONTRIBUTING.md
```diff
@@ -57,8 +57,19 @@ You can add the sign off when creating the git commit via `git commit -s`.
...
-This facilitates incremental reviews since the reviewer does not go through all the code again to
-find out what has changed since the last review.
+* The pull request title should describe what the change does and not embed issue numbers.
+The pull request should only be blank when the change is minor. Any feature should include
+a description of the change and what motivated it. If the change or design changes through
+review, please keep the title and description updated accordingly.
+* A single approval is sufficient to merge. If a reviewer asks for
+changes in a PR they should be addressed before the PR is merged,
+even if another reviewer has already approved the PR.
+* During the review, address the comments and commit the changes
```
codefromthecrypt, line 67: I think I would be a lot more happy with the "no squash" policy if we say, "once you have a review" please don't squash

because it is annoying especially in the case where no one is around and they haven't reviewed yet either. The whole point it to prevent re-review.

So, how about we clarify as that matches the intent?

codefromthecrypt: this would implicitly cover "draft" PRs as typically folks don't review them in draft state.

codefromthecrypt: or maybe 'During the review" says that anyway, if we think it does nevermind!

mathetake: I think `During the review` says that :D

ncruces (approved):

PR #1418, opened 2023-05-01
lifts emscripten.InstantiateForModule out of experimental

codefromthecrypt:
We added emscripten.InstantiateForModule into an experimental package so that we could make a patch release and include it. Now that we've changed api.Function, our next release needs to be a minor. This lifts InstantiateForModule into the normal emscripten package as we know it works and there's no known better signature to use for it.


mathetake (approved):

codefromthecrypt:
cc @jerbob92

PR #1419, opened 2023-05-01
Reduces FunctionDefinition usages

mathetake:
This refactors interpreter/compiler around the FunctionDefinition usage,
and reduces the usage and memory footprint during installation.

```
goos: darwin
goarch: arm64
pkg: github.com/tetratelabs/wazero/internal/integration_test/bench
                                       │   old.txt    │              new.txt              │
                                       │    sec/op    │   sec/op     vs base              │
Initialization/interpreter-10            14.08µ ±  1%   14.09µ ± 1%       ~ (p=0.832 n=7)
Initialization/interpreter-multiple-10   17.28µ ±  3%   16.58µ ± 3%  -4.03% (p=0.004 n=7)
Initialization/compiler-10               15.95µ ± 11%   15.48µ ± 6%       ~ (p=0.073 n=7)
Initialization/compiler-multiple-10      15.21µ ±  6%   14.67µ ± 6%       ~ (p=0.053 n=7)
Compilation/with_extern_cache-10         114.5µ ±  1%   114.4µ ± 1%       ~ (p=0.902 n=7)
Compilation/without_extern_cache-10      1.274m ±  1%   1.275m ± 1%       ~ (p=0.318 n=7)
Compilation/interpreter-10               432.0µ ±  1%   433.8µ ± 1%       ~ (p=0.318 n=7)
Compilation_sqlite3/compiler-10          177.2m ±  1%   177.3m ± 1%       ~ (p=0.902 n=7)
Compilation_sqlite3/interpreter-10       60.77m ±  0%   60.55m ± 1%       ~ (p=0.456 n=7)
geomean                                  324.6µ         320.9µ       -1.17%

                                       │   old.txt    │               new.txt                │
                                       │     B/op     │     B/op      vs base                │
Initialization/interpreter-10            132.3Ki ± 0%   132.3Ki ± 0%       ~ (p=1.000 n=7) ¹
Initialization/interpreter-multiple-10   132.4Ki ± 0%   132.4Ki ± 0%       ~ (p=1.000 n=7)
Initialization/compiler-10               132.3Ki ± 0%   131.5Ki ± 0%  -0.66% (p=0.001 n=7)
Initialization/compiler-multiple-10      132.3Ki ± 0%   131.5Ki ± 0%  -0.66% (p=0.001 n=7)
Compilation/with_extern_cache-10         36.92Ki ± 0%   36.92Ki ± 0%       ~ (p=0.596 n=7)
[...]

PR #1420, opened 2023-05-01
interpreter: refactors around FunctionDefinition

mathetake:
I forgot committing the change of interpreter in #1419


codefromthecrypt (approved):
great. thanks

PR #1421, opened 2023-05-01
wasip1: fix file open modes used by wasi-libc

achille-roussel:
This PR extends wazero's interpretation of rights in `wasi_snapshot_preview1.path_open` to allow programs to open files in either read-only, write-only, or read-write mode.


codefromthecrypt (approved):
Looks like how you described in slack. I'm glad this is 100pct contained to wasip1!

achille-roussel:
Seems like I introduced a regression on wasi-testsuite, looking into it.

achille-roussel:
So it seems we were _accidentally_ passing one of the wasi-testsuite tests https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/directory_seek.rs#L40-L44

The test validates that a directory cannot get the FD_SEEK right, since we were zeroing all bits it tricked the test into believing that we were denying seek permission on directories.

achille-roussel:
I tuned the approach a bit to address the wasi-testsuite issue: instead of setting all bits of `fs_rights_base` and `fs_rights_inheriting` in the result of `fd_fdstat_get`, wazero now sets different rights depending on the file type: for directories, the rights related to file operations (e.g. read/write/seek) are removed.

imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm
mathetake, line 1: 99% reduction 🙄

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -222,24 +222,66 @@ func fdFdstatGetFn(_ context.Context, mod api.Module, params []uint64) syscall.E
...
-	0, 0, 0, 0, 0, 0, 0, 0, // fs_rights_base
-	0, 0, 0, 0, 0, 0, 0, 0, // fs_rights_inheriting
-}
-
-func writeFdstat(buf []byte, filetype uint8, fdflags uint16) {
-	// memory is re-used, so ensure the result is defaulted.
-	copy(buf, blockFdstat)
-	buf[0] = filetype
-	buf[2] = byte(fdflags)
+const fileRights = wasip1.RIGHT_FD_DATASYNC |
```
mathetake, line 243: q: so this is basically exactly the same as the one defined the CL?

mathetake: I meant the CL to the golang/go

internal/platform/open_file_windows.go
```diff
@@ -106,9 +106,6 @@ func open(path string, mode int, perm uint32) (fd syscall.Handle, err error) {
 	case syscall.O_RDWR:
 		access = syscall.GENERIC_READ | syscall.GENERIC_WRITE
 	}
-	if mode&syscall.O_CREAT != 0 {
-		access |= syscall.GENERIC_WRITE
-	}
```
achille-roussel, line 111: @codefromthecrypt I had to make this change to have the test pass on windows. I'm not sure I understand why this was needed, since you introduced this code initially you may have more context to share?

codefromthecrypt: so the godoc on this code says it was originally from go, but actually it has drifted since, which I guess is a good thing. We should probably make comments under the link about exceptions.

```
// # Differences from syscall.Open
//
// This code is based on syscall.Open from the below link with some differences
// https://github.com/golang/go/blame/go1.20/src/syscall/syscall_windows.go#L308-L379
//
//   - syscall.O_CREAT doesn't imply syscall.GENERIC_WRITE as that breaks flag expectations
//      in wasi.
```

This was the change in go over a decade ago which is now gone (fine with me) https://github.com/golang/go/commit/035265975fbf93fa67aa6a6e7611ba640691d693

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -222,24 +222,66 @@ func fdFdstatGetFn(_ context.Context, mod api.Module, params []uint64) syscall.E
...
-	filetype := getWasiFiletype(st.Mode)
-	writeFdstat(buf, filetype, fdflags)
+	var fsRightsBase uint32
+	var fsRightsInheriting uint32
+	fileType := getWasiFiletype(st.Mode)
+
+	switch fileType {
+	case wasip1.FILETYPE_DIRECTORY:
+		// To satisfy wasi-testsuite, we must advertise that directories cannot
+		// be given seek permission (RIGTH_FD_SEEK).
```
codefromthecrypt, line 232:
```suggestion
		// be given seek permission (RIGHT_FD_SEEK).
```

codefromthecrypt (approved):
Probably worth splatting the comment recommendation, but either way thanks a lot for getting to the bottom of this and finding a pragmatic way out!

mathetake (approved):

Issue #1417, opened 2023-05-01
complete internal platform.File


codefromthecrypt:
This issue tracks completion of the plaform.File type, started in #1290

Generally, this process includes migration off every place we use fs.FS or something cast from `File.File()` to functions on `platform.File`. We'll know we are feature complete when the helper function `platform.File.File()` can be removed. Each function should prefer the same signature (removing path) on `sysfs.FS` and when absent try to balance the signature between an existing "fd" syscall function with practicalities of the call site.

Here are the functions to convert. The function signature on `platform.File` will be different at least errno result.
- [x] `fs.File.Stat() (FileInfo, error)` e.g. for  `fs.fstat` in gojs
  * `Stat() (Stat_t, syscall.Errno)` in #1290
- [x] `fs.File.Close() error` e.g. for  `fs.close` in gojs
  * `Close() syscall.Errno` in #1290
- [x] `Sync() error` e.g. for `fs.fsync` in gojs
  * `Sync() syscall.Errno` in #1426
- [x] `Chmod(fs.FileMode) error` e.g. for `fs.fchmod` in gojs
  * `Chmod(fs.FileMode) syscall.Errno` #1423
- [x] `platform.ChownFile(f fs.File, uid, gid int) syscall.Errno` e.g. for `fs.fchown` in gojs
  * `Chown(uid, gid int) syscall.Errno` #1422
- [x] `Truncate(size int64) error` e.g. for `fs.ftruncate` in gojs
  * `Truncate(int) syscall.Errno` #1417
- [x] `FileDatasync(f fs.File) (err syscall.Errno)` e.g. for `wasi_snapshot_preview1.fd_datasync`
  * `Datasync() syscall.Errno` in #1427
- [x] `io.Reader` e.g. for `fs.read` in gojs; `wasi_snapshot_preview1.fd_read`
  * `read(p []byte) (n int, errno syscall.Errno)` in #1439
- [x] `io.ReaderAt` e.g. for `fs.read` and `pread` in gojs; `wasi_snapshot_preview1.fd_read` and `fd_pread`
- [x] `.(io.Writer)` e.g. to check if a file is writeable without writing
  * `AccessMode() int` in #1438
- [x] `io.Writer` e.g. for `fs.write` in gojs; `wasi_snapshot_preview1.fd_write`
  * `Write(p []byte) (n int, errno syscall.Errno)` in #1438
[...]

PR #1422, opened 2023-05-01
Adds platform.File.Chown

codefromthecrypt:
This migrates the function `platform.ChownFile` to the work-in-progress `platform.File` type without changing its implementation.

See #1417

### DIFF
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -169,6 +169,15 @@ func (r *lazyDir) Stat() (platform.Stat_t, syscall.Errno) {
 	}
 }
 
+// Chown implements the same method as documented on platform.File
+func (r *lazyDir) Chown(uid, gid int) syscall.Errno {
+	if f, err := r.file(); err != 0 {
+		return err
+	} else {
+		return f.Chown(uid, gid)
+	}
+}
+
 // File implements the same method as documented on platform.File
 func (r *lazyDir) File() fs.File {
 	if f, err := r.file(); err != 0 {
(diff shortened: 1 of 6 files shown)


internal/sys/fs.go
```diff
@@ -169,6 +169,15 @@ func (r *lazyDir) Stat() (platform.Stat_t, syscall.Errno) {
 	}
 }
 
+// Chown implements the same method as documented on platform.File
+func (r *lazyDir) Chown(uid, gid int) syscall.Errno {
+	if f, err := r.file(); err != 0 {
+		return err
```
mathetake, line 175: the logic seems not in line with `File()`:

```go
	if f, err := r.file(); err != 0 {
		panic(err) // Bad, but temporary
	}
```

codefromthecrypt: so `lazyDir` is a `platform.File`, so it can return a `syscall.Errno` (the second result of `r.file()`).

I guess you mean there's a generally problematic thing about error mapping because the implementation of `r.file()` opens the file lazily. If it is later missing, I think we should return EBADF not panic, right? as that's normal for things like this. So maybe panic if something besides ENOENT?

mathetake: ah ok that makes sense and I'm good with as-is. Thank you for clarifying!

mathetake (approved):

internal/sys/fs.go
codefromthecrypt: well it is half ok.. we are leaking more errnos than ENOENT, so I need to fix it. Good catch

codefromthecrypt: I made the panic more precise. Good eye!

PR #1427, opened 2023-05-02
Adds Datasync to platform.File

codefromthecrypt:
See #1417


internal/sysfs/sysfs_test.go
```diff
@@ -686,47 +685,6 @@ func TestWriterAtOffset_Unsupported(t *testing.T) {
 	require.Equal(t, syscall.ENOSYS, err)
 }
 
-// Test_FileSync doesn't guarantee sync works because the operating system may
```
codefromthecrypt, line 689: we don't need to test this again because adapterFS only job is to return a platform.File, and the impl is already tested.

mathetake (approved):

PR #1428, opened 2023-05-02
Adds Truncate to platform.File

codefromthecrypt:
See #1417

### DIFF
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -210,6 +210,11 @@ func (r *lazyDir) Datasync() syscall.Errno {
 	}
 }
 
+// Truncate implements the same method as documented on platform.File
+func (r *lazyDir) Truncate(int64) syscall.Errno {
+	return syscall.EISDIR
+}
+
 // File implements the same method as documented on platform.File
 func (r *lazyDir) File() fs.File {
 	if f, ok := r.file(); !ok {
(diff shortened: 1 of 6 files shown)


internal/sys/fs.go
```diff
@@ -273,7 +283,6 @@ type FileEntry struct {
 	// was called.
 	ReadDir *ReadDir
 
-	openPath string
```
codefromthecrypt, line 276: this is no longer needed to be stored externally

codefromthecrypt:
looks like there's disagreement in the syscall.Errno returned when ftruncate acts on a directory not a file. I'll sort it out and mark draft meanwhile

codefromthecrypt:
meh fix one another breaks (windows). I'll mark draft again

evacchi:
apparently the code on `main` can't acquire this lock, which trickles down into the original error. Your recent change is now acquiring the lock, so it returns with "use of closed file"; code on this branch succeeds in acquiring the lock

```
// fd_windows.go:
func (fd *FD) Write(buf []byte) (int, error) {
	if err := fd.writeLock(); err != nil {
		return 0, err
	}

// the callee returns the actual error
func (fd *FD) writeLock() error {
	if !fd.fdmu.rwlock(false) {
		return errClosing(fd.isFile)
	}
	return nil
}

```

now the question is whether the behavior on main is actually correct

codefromthecrypt:
will rework this after #1430

evacchi (approved):

PR #1429, opened 2023-05-02
Revises instructions to latest guidance from TinyGo

codefromthecrypt:
This follows up the change to our example by @lburgazzoli in #1390, which basically switched our guidance to have the host call "free" on strings it receives as results from the guest code.

The notably change was `stringToLeakedPtr` which should be added to [tinymem](https://github.com/tetratelabs/tinymem) soon.
https://github.com/tetratelabs/wazero/blob/main/examples/allocation/tinygo/testdata/greet.go#L87-L94

tinymem should also remove custom memory exports, as they are no longer needed.

One thing missing from this doc are tinygo build flags to explicitly export "malloc" and "free", so that tinygo can eventually stop exporting these by default. If someone knows the syntax, let me know and I will update this.

### DIFF
--- a/site/content/languages/tinygo.md
+++ b/site/content/languages/tinygo.md
@@ -146,7 +146,7 @@ calls a free function, with the same `ptr`, afterwards and unconditionally.
 Note: wazero includes an [example project][8] that shows this.
 
 The general call patterns are the following. Host is the process embedding the
-WebAssembly runtime, such as wazero. Guest is the Tinygo source compiled to
+WebAssembly runtime, such as wazero. Guest is the TinyGo source compiled to
 target wasi.
 
 * Host allocates a string to call an exported Guest function
--- a/site/content/languages/tinygo.md
+++ b/site/content/languages/tinygo.md
@@ -145,77 +145,38 @@ calls a free function, with the same `ptr`, afterwards and unconditionally.
 
 Note: wazero includes an [example project][8] that shows this.
 
-There are two ways to implement this pattern, and they affect how to implement
-the `ptrToString` function above:
-* Built-in `malloc` and `free` functions
-* Custom `malloc` and `free` functions
+The general call patterns are the following. Host is the process embedding the
+WebAssembly runtime, such as wazero. Guest is the Tinygo source compiled to
+target wasi.
 
-While both patterns are used in practice, TinyGo maintainers only support the
-custom approach. See the following issues for clarifications:
-* [WebAssembly exports for allocation][9]
-* [Memory ownership of TinyGo allocated pointers][10]
+* Host allocates a string to call an exported Guest function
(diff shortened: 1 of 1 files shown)


site/content/languages/tinygo.md
```diff
@@ -145,77 +145,38 @@ calls a free function, with the same `ptr`, afterwards and unconditionally.
...
-custom approach. See the following issues for clarifications:
-* [WebAssembly exports for allocation][9]
-* [Memory ownership of TinyGo allocated pointers][10]
-
-#### Built-in `malloc` and `free` functions
-
-The least code way to allow the host to allocate memory is to call the built-in
-`malloc` and `free` functions exported by TinyGo:
+The general call patterns are the following. Host is the process embedding the
+WebAssembly runtime, such as wazero. Guest is the Tinygo source compiled to
```
achille-roussel, line 149: Assuming the README title is the expected spelling https://github.com/tinygo-org/tinygo#tinygo---go-compiler-for-small-places
```suggestion
WebAssembly runtime, such as wazero. Guest is the TinyGo source compiled to
```

achille-roussel (approved):

mathetake (approved):

ncruces (approved):
Sorry for getting a bit lost in #1390.
Documentation change here seems fine.

codefromthecrypt:
Thanks for the help, folks. Happy to revise after the fact, also. I'll go ahead and update/release tinymem accordingly so we consistently provide coherent advice for TinyGo users!

aykevl:
> One thing missing from this doc are tinygo build flags to explicitly export "malloc" and "free", so that tinygo can eventually stop exporting these by default. If someone knows the syntax, let me know and I will update this.

Don't know right away, but I think you can do something like this:

```go
// #cgo LDFLAGS: -Wl,--export=malloc -Wl,--export=free
import "C"
```

You could add these so that it will continue to work in the future if we ever decide to not export `malloc` and `free` by default.

(See https://github.com/tinygo-org/tinygo/pull/3142#issuecomment-1241425440)

codefromthecrypt:
Thanks @aykevl 

Right now, we have 

```go
// #include <stdlib.h>
import "C"
```

If I build the below with or without the include, I get `invalid flag: -Wl,--export=malloc`
```go
// #cgo LDFLAGS: -Wl,--export=malloc -Wl,--export=free
import "C"
```

Maybe something subtle?

aykevl:
I think this can be fixed by adding `--export=` to [cgo/security.go](https://github.com/tinygo-org/tinygo/blob/release/cgo/security.go) and using `//#cgo LDFLAGS: --export=malloc --export=free`. So this needs a small patch in TinyGo to work.

aykevl:
The patch was merged, so `--export` should work starting with the next release.

PR #1431, opened 2023-05-03
Adds Path to platform.File and refactors tests

codefromthecrypt:
This adds Path which is non-empty when the file is a normal file or directory. This immediately supports the re-open use case, and later may make some things in windows easier.

This also adds a platform.NewFsFile function which makes some call sites better.

Finally, this fixes a bug hidden by how our tests were written, where the wrong errno was returned when a file was used after close in windows.

### DIFF
--- a/internal/platform/stat_test.go
+++ b/internal/platform/stat_test.go
@@ -158,14 +158,11 @@ func TestStat(t *testing.T) {
 func TestStatFile(t *testing.T) {
 	tmpDir := t.TempDir()
 
-	var st Stat_t
-
-	tmpDirF, errno := OpenFile(tmpDir, syscall.O_RDONLY, 0)
-	require.EqualErrno(t, 0, errno)
+	tmpDirF := openFsFile(t, tmpDir, syscall.O_RDONLY, 0)
 	defer tmpDirF.Close()
 
 	t.Run("dir", func(t *testing.T) {
-		st, errno = tmpDirF.Stat()
+		st, errno := tmpDirF.Stat()
 		require.EqualErrno(t, 0, errno)
 
 		require.True(t, st.Mode.IsDir())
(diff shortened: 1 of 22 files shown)


codefromthecrypt:
@evacchi I'm stumped for tonight on how I screwed up TestOpenFile_Errors/writing_to_a_read-only_file_is_EBADF on windows. If you have a minute to check it out or fix, much oblidged. otherwise, I'll swing the axe tomorrow. I tried to split code out of the other PRs and this is the one that's ready except this fail.

evacchi:
apparently the code on `main` can't acquire this lock, which trickles down into the original error. In fact the field of `poll.FD.fdmu.state` is 1 for `main` and 0 on this branch.

So, your recent change is now acquiring the lock, and it no longer returns with "use of closed file"

```
// poll/fd_windows.go:
func (fd *FD) Write(buf []byte) (int, error) {
	if err := fd.writeLock(); err != nil {
		return 0, err
	}

// the callee returns the actual error
func (fd *FD) writeLock() error {
	if !fd.fdmu.rwlock(false) {
		return errClosing(fd.isFile)
	}
	return nil
}

```

now I've investigated this further and it looks like the issue is that the test was originally written as:

```
		f := openFsFile(t, path, os.O_RDONLY, 0)
		defer require.Zero(t, f.Close())
```

and apparently the semantics of `defer` are to eagerly evaluate all the arguments to `require.Zero()` and _then_ delay `require.Zero(...)`; in other words `f.Close()` was not really deferred 😨 

in other words, this is also broken on main, we just never noticed 😰 🤣

evacchi:
...and apparently I have pushed it to the other PR #1430 🤦

evacchi:
feel free to revert the commit from either (or both!) PRs, hope this helps

internal/platform/stat_test.go
```diff
@@ -177,39 +174,37 @@ func TestStatFile(t *testing.T) {
...
-		st, errno = fileF.Stat()
+		st, errno := fileF.Stat()
 		require.EqualErrno(t, 0, errno)
 
 		require.False(t, st.Mode.IsDir())
 		require.NotEqual(t, uint64(0), st.Ino)
 	})
 
-	t.Run("closed file", func(t *testing.T) {
+	t.Run("closed fsFile", func(t *testing.T) {
```
codefromthecrypt, line 195:
```suggestion
	t.Run("closed file", func(t *testing.T) {
```

internal/platform/wrap_windows.go
```diff
@@ -61,7 +62,11 @@ func (w *windowsWrappedFile) Write(p []byte) (n int, err error) {
 		return
 	}
 
-	return w.WriteFile.Write(p)
+	n, err = w.WriteFile.Write(p)
+	if errors.Is(err, ERROR_ACCESS_DENIED) {
```
codefromthecrypt, line 66: ok so this is the change huh
```suggestion
	// ERROR_ACCESS_DENIED is often returned instead of EBADF
	// when a file is used after close.
	if errors.Is(err, ERROR_ACCESS_DENIED) {
```

codefromthecrypt:
Thanks for the help @evacchi, super subtle! I didn't mean for you to fix the other PR, just mentioned it (it wasn't it good condition for someone to help on :p).

Anyway, thanks for the help as I was so stumped by I suppose what amounted to assuming the main tests did what they appear. (missing the eager close in the defer that was there before)

great work!

mathetake (approved):

PR #1432, opened 2023-05-03
compiler: update RATIONALE.md for native check

evacchi:
Relates to #1409, #1413. Explains why the native check won't work.


mathetake (approved):

codefromthecrypt (approved):
thanks for the follow-up

PR #1433, opened 2023-05-03
experimental: expose source offset to Listener

pelletier:
This change provides a user of `experimental.StackIterator` the ability to retrieve the offset in the Code section of each function call in the frame of the stack. It enables capabilities such as source-mapping to be implemented by the user outside of wazero (removing the need for #1403).

It works by retrieving the return address of each stack frame and using the `irOperationSourceOffsetsInWasmBinary` attached to the function associated with that stack frame.

Though it works in my tests and on the few programs I used, I feel quite out of my depth here; I'd love some guidance and feedback! :) 

Question 1 – This source offsets mapping is only produced by the compiler when DWARF sections are present in the module. In practice, it's a reasonable limitation for us since we currently are only interested in using this information to correlate with DWARF data anyway. However, we may want to look at other applications in the future. It also meant that I had to provide a minimal DWARF info section for `dwarf/debug` not to trip up and the compiler to emit the offsets map. Would you be interested in adding either a flag to enable the mapping generation regardless of the presence of DWARF sections?

Question 2 – because the stack iterator is used in the`Before` listener hook, the topmost frame of the call stack is always the called function that triggered the listener. As such, I think the source offset should be the offset of the first byte of that function in the code section. In that case, I have observed that the pc is just lower than the first entry of `irOperationSourceOffsetsInWasmBinary` (hence the change in `getSourceOffsetInWasmBinary`). I am struggling to understand why? Returning the first entry of that collection seems to give the expected results, but I don't rationalize it.

I appreciate your consideration!

### DIFF
--- a/experimental/listener.go
+++ b/experimental/listener.go
@@ -18,6 +18,11 @@ type StackIterator interface {
 	Next() bool
 	// FunctionDefinition returns the function type of the current function.
 	FunctionDefinition() api.FunctionDefinition
+	// SourceOffset returns the offset in the Code section where
+	// the function call occured (it is translated for native
+	// functions). Returns 0 if the source offset cannot be
+	// calculated.
+	SourceOffset() uint64
 	// Parameters returns api.ValueType-encoded parameters of the current
 	// function. Do not modify the content of the slice, and copy out any value
 	// you need.
--- a/experimental/listener.go
+++ b/experimental/listener.go
@@ -13,19 +13,22 @@ import (
 // iteration. As a result, parameter values may be different than the ones their
 // function was called with.
 type StackIterator interface {
-	// Next moves the iterator to the next function in the stack. Returns false
-	// if it reached the bottom of the stack.
+	// Next moves the iterator to the next function in the stack. Returns
+	// false if it reached the bottom of the stack.
 	Next() bool
 	// FunctionDefinition returns the function type of the current function.
 	FunctionDefinition() api.FunctionDefinition
-	// SourceOffset returns the offset in the Code section where
-	// the function call occured (it is translated for native
-	// functions). Returns 0 if the source offset cannot be
-	// calculated.
+	// SourceOffset computes the offset in the module Code section where the
+	// call occured (translated for native functions), or the beginning of
+	// function for the top of the stack. Returns 0 if the source offset
+	// cannot be calculated.
+	//
+	// The source offset is meant to help map the function calls to their
+	// location in the original source files.
 	SourceOffset() uint64
 	// Parameters returns api.ValueType-encoded parameters of the current
-	// function. Do not modify the content of the slice, and copy out any value
-	// you need.
+	// function. Do not modify the content of the slice, and copy out any
+	// value you need.
 	Parameters() []uint64
 }
 
--- a/experimental/listener_example_test.go
+++ b/experimental/listener_example_test.go
@@ -89,19 +89,28 @@ func Example_stackIterator() {
 	it := &fakeStackIterator{}
 
 	for it.Next() {
-		fmt.Println("function:", it.FunctionDefinition().DebugName(), "args", it.Args())
+		fmt.Println("function:", it.FunctionDefinition().DebugName())
+		fmt.Println("\tparameters:", it.Parameters())
+		fmt.Println("\tsource offset:", it.SourceOffset())
 	}
 
 	// Output:
-	// function: fn0 args [1 2 3]
-	// function: fn1 args []
-	// function: fn2 args [4]
+	// function: fn0
+	// 	parameters: [1 2 3]
+	// 	source offset: 1234
+	// function: fn1
+	// 	parameters: []
+	// 	source offset: 7286
+	// function: fn2
+	// 	parameters: [4]
+	// 	source offset: 935891
 }
 
 type fakeStackIterator struct {
 	iteration int
 	def       api.FunctionDefinition
 	args      []uint64
+	offset    uint64
 }
 
 func (s *fakeStackIterator) Next() bool {
@@ -109,12 +118,15 @@ func (s *fakeStackIterator) Next() bool {
 	case 0:
 		s.def = &mockFunctionDefinition{debugName: "fn0"}
 		s.args = []uint64{1, 2, 3}
+		s.offset = 1234
 	case 1:
 		s.def = &mockFunctionDefinition{debugName: "fn1"}
 		s.args = []uint64{}
+		s.offset = 7286
 	case 2:
 		s.def = &mockFunctionDefinition{debugName: "fn2"}
 		s.args = []uint64{4}
+		s.offset = 935891
 	case 3:
 		return false
 	}
@@ -126,9 +138,15 @@ func (s *fakeStackIterator) FunctionDefinition() api.FunctionDefinition {
 	return s.def
 }
 
-func (s *fakeStackIterator) Args() []uint64 {
+func (s *fakeStackIterator) Parameters() []uint64 {
 	return s.args
 }
+
+func (s *fakeStackIterator) SourceOffset() uint64 {
+	return s.offset
+}
+
+var _ experimental.StackIterator = &fakeStackIterator{}
 
 type mockFunctionDefinition struct {
 	debugName string
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -880,6 +880,9 @@ func (f *function) getSourceOffsetInWasmBinary(pc uint64) uint64 {
 		}
 		return srcMap.irOperationOffsetsInNativeBinary[i] >= pcOffsetInNativeBinary
 	})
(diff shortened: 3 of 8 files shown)


codefromthecrypt (commented):
summary comments as this morning is a meeting morning.

This sort of feature should have a test in the experimental package to show its use, even minimally.

Also some bikeshedding about returning -1 instead of zero, for example for host functions..

and the length of the code section I think will not be 64bits :) sounds like something to quote and be wrong 20 years later ;)

The comment on the exposed thing should continue to say how it is used. `e.g. to allow debuggers to map this function to its source`

I'll leave other comments to others, then later chime in more if needed.

Meanwhile, cool stuff!

pelletier:
Thanks for the review! 

Expanded the documentation for `SourceOffset` and added it to the existing `StackIterator` example (and fixed it along the way...)

> Also some bikeshedding about returning -1 instead of zero, for example for host functions.
> and the length of the code section I think will not be 64bits :)

That's fair. `getSourceOffsetInWasmBinary` (and equivalent in the interpreter) uses `uint64`. It suffers from the same issue of returning 0 if not found (which I think would be a valid offset if called with the PC at the beginning very first function of the Code section?) and using a rather large type. I opted to use `uint64` for SourceOffset for consistency, as it seems to be the type used for offsets in the Code section in Wazero.

To signal a "not found" case, I thought about either using another sentinel value (like `MaxUint64`) or returning an additional `error` value in that case but eventually decided to lean into zero-values for less ceremony. Aside from that, I don't have much argument either way.

I'm happy to change the signature of `SourceOffset` to any of the above and do the appropriate casts, and/or change `getSourceOffsetInWasmBinary` & friends to have them not use `uint64` either. Let me know your preference!

codefromthecrypt:
I personally prefer -1 to zero for saying not found (e.g. strings.Index) as it feels more idiomatic in Go, I'm not aware of a lot of use of Max or zero for the same thing. It is a bike shed, and to get through them, usually blaming decision on another precedent is the easy way out. I would blame other code that look for an index of something in go returning -1 as the reason.

internal/engine/compiler/engine.go
```diff
@@ -880,6 +880,9 @@ func (f *function) getSourceOffsetInWasmBinary(pc uint64) uint64 {
 		}
 		return srcMap.irOperationOffsetsInNativeBinary[i] >= pcOffsetInNativeBinary
 	})
+	if index == 0 && len(srcMap.irOperationSourceOffsetsInWasmBinary) > 0 {
```
mathetake, line 883: I think I need comment on this for the future

mathetake (commented):
looks good to me. Requesting review from @evacchi as well and delegating to @codefromthecrypt and edo for the final approval. Good work!

experimental/listener.go
```diff
@@ -13,14 +13,22 @@ import (
...
-	// Next moves the iterator to the next function in the stack. Returns false
-	// if it reached the bottom of the stack.
+	// Next moves the iterator to the next function in the stack. Returns
+	// false if it reached the bottom of the stack.
 	Next() bool
 	// FunctionDefinition returns the function type of the current function.
 	FunctionDefinition() api.FunctionDefinition
+	// SourceOffset computes the offset in the module Code section where the
+	// call occured (translated for native functions), or the beginning of
+	// function for the top of the stack. Returns 0 if the source offset
```
codefromthecrypt, line 23: @ncruces have any 👨‍🎨 on the -1 vs zero thing for source offset not available. I know we recently chatted on tinygo (and otherwise) using zero as nil though technically wasm can start from zero.

I think basically nothing will start at line zero in the source file, is why we can use zero here. (it isn't about the binary rather the %.c %.rs etc "the source"

internal/engine/interpreter/interpreter.go
```diff
@@ -261,6 +264,16 @@ func (si *stackIterator) FunctionDefinition() api.FunctionDefinition {
 	return si.fn.definition()
 }
 
+// SourceOffset implements experimental.StackIterator.
```
codefromthecrypt, line 267: nit: to avoid introducing style 3 or 4 of this boilerplate ;)

```suggestion
// SourceOffset implements the same method as documented on experimental.StackIterator.
```

internal/engine/compiler/engine.go
```diff
@@ -1128,13 +1133,25 @@ func (si *stackIterator) Next() bool {
...
 	frame := si.base + callFrameOffset(si.fn.funcType)
+	si.pc = si.stack[frame+0]
 	si.base = int(si.stack[frame+1] >> 3)
 	// *function lives in the third field of callFrame struct. This must be
 	// aligned with the definition of callFrame struct.
 	si.fn = *(**function)(unsafe.Pointer(&si.stack[frame+2]))
 	return si.fn != nil
 }
 
+// SourceOffset implements experimental.StackIterator.
```
codefromthecrypt, line 1144:
```suggestion
// SourceOffset implements the same method as documented on experimental.StackIterator.
``````

File: internal/engine/compiler/engine.go (hunk 1 above)
Comment (codefromthecrypt) [reply internal/engine/compiler/engine.go:883]: personally I would ask "why is this done (only) on index zero" and want the comment to answer that. Up to others what the good comment would be

File: internal/engine/compiler/engine.go [hunk 5]
```diff
@@ -1128,13 +1133,25 @@ func (si *stackIterator) Next() bool {
...
+func (si *stackIterator) SourceOffset() uint64 {
+	p := si.fn.parent
+
+	if len(p.sourceOffsetMap.irOperationSourceOffsetsInWasmBinary) == 0 {
+		return 0
```
codefromthecrypt, line 1149:
```suggestion
		return 0 // source isn't available
```

codefromthecrypt (approved):
looks to me there's zero perf impact unless stackiterator is already in use. cool

nits only, so ready on green after the comment request is in!

internal/engine/compiler/engine.go
evacchi: I wonder if it's worth adding a `func hasSourceOffsetsInWasmBinary() bool { return len(irOperationSourceOffsetsInWasmBinary) != 0 }` (maybe not in this PR)

evacchi (approved):
I don't have much to add besides what @mathetake and @codefromthecrypt already indicated :)

evacchi:
ps: obviously there are still a few failures on CI that should be addressed

internal/engine/compiler/engine.go
pelletier: Added a comment to clarify.

pelletier:
Thank you everyone for the reviews! I have addressed all the comments (unless I misunderstood the -1 vs 0 decision).

experimental/listener.go
ncruces: Yes, I don't think what we chatted about applies here.

TLDR it was: linear memory address zero in C (and probably TinyGo) can't be dereferenced, so I choose to panic when the host gets a zero pointer it's expected to dereference (although the host _can_ read/write address zero).

ncruces: As long as 0 is not a valid offset and given this is unsigned, zero seams reasonable.

ncruces (approved):

experimental/listener.go
codefromthecrypt: apples and oranges, but what's common (fruit?) is we are banishing a zero offset which isn't usually valid.

So, the simile is like memory offset zero is almost always invalid, source offset zero would be pretty invalid.

nerdy side bar:

As far as I can think of, the only languages which could use either are %.wat and an edge case of that, also. For example, text format can use memory offset zero, and also strangely, you can omit `module` and have a function as a top-level element in the file.. otoh I doubt if source-mapped you'd end up with index zero as it would not be in the body part ;) https://www.w3.org/TR/wasm-core-1/#abbreviations%E2%91%A0%E2%91%A5

experimental/listener_example_test.go
```diff
@@ -89,32 +89,44 @@ func Example_stackIterator() {
...
+		fmt.Println("\tsource offset:", it.SourceOffset())
 	}
 
 	// Output:
-	// function: fn0 args [1 2 3]
-	// function: fn1 args []
-	// function: fn2 args [4]
+	// function: fn0
+	// 	parameters: [1 2 3]
+	// 	source offset: 1234
```
codefromthecrypt, line 100: 👍

internal/engine/compiler/engine.go
```diff
@@ -1116,7 +1124,7 @@ func (si *stackIterator) clear() {
 	si.started = false
 }
 
-// Next implements experimental.StackIterator.
+// Next implements the same method as documtend on experimental.StackIterator.
```
codefromthecrypt, line 1127:
```suggestion
// Next implements the same method as documented on experimental.StackIterator.
```

codefromthecrypt (approved):

codefromthecrypt:
nicely done! I'll polish the typo post merge in another change

PR #1434, opened 2023-05-04
examples(allocation): ensures data used in host functions isn't freed before invocation

lburgazzoli:


codefromthecrypt:
Thanks! mind doing this and sharing (before/after bench) in the PR description?

```bash
# from main
go test -run='^$' -bench '^BenchmarkAllocation.*' ./internal/integration_test/vs/compiler -count=6 > old.txt

# from this commit
go test -run='^$' -bench '^BenchmarkAllocation.*' ./internal/integration_test/vs/compiler -count=6 > new.txt

benchstat old.txt new.txt 
```

examples/allocation/tinygo/testdata/greet.go
```diff
@@ -72,13 +74,13 @@ func ptrToString(ptr uint32, size uint32) string {
...
 
-// stringToPtr returns a pointer and size pair for the given string in a way
-// compatible with WebAssembly numeric types.
+// stringToPtr returns a pointer and size pair to the underlying bytes of the given
+// string in a way compatible with WebAssembly numeric types. Since Go strings are
+// immutable, the pointer returned by stringToPtr must not be modified.
 func stringToPtr(s string) (uint32, uint32) {
-	buf := []byte(s)
-	ptr := &buf[0]
+	ptr := unsafe.StringData(s)
```
codefromthecrypt, line 81: Only thing that seems off is that by spec unsafe.StringData can return nil on empty string, and the later unsafe.Pointer is supposed to crash on it.
```go
//	// INVALID: conversion of nil pointer
//	u := unsafe.Pointer(nil)
```

Can you check tinygo source to see if unsafe.StringData on empty string is not nil, then comment above how this is ok because tinygo won't return nil?

codefromthecrypt (commented):
code wise I like it.

While personally, I rarely understand when I'm supposed to do `runtime.KeepAlive`, this seems the pattern to do it. What I like also is that the stringToPtr code is more concise. 

One homework before merge is checking that unsafe.StringData won't return nil when compiled with TinyGo's wasi target.

examples/allocation/tinygo/testdata/greet.go
lburgazzoli: I created a quick example to test the actual behavior:

```go
package main

import (
	"fmt"
	"unsafe"
)

func stringToPtr(s string) (uint32, uint32) {
	ptr := unsafe.StringData(s)
	unsafePtr := uintptr(unsafe.Pointer(ptr))
	return uint32(unsafePtr), uint32(len(s))
}

func main() {
	ptr, size := stringToPtr("foo")
	fmt.Println(ptr, size)
	
	ptr, size = stringToPtr("")
	fmt.Println(ptr, size)
}
```

Which results in something like:

```
➜ tinygo run main.go
2124687 3
0 0
```

I guess here the question is what is the best behavior when the ptr is zero as then it is the beginning of the linear memory I guess. We could add a check against that but in a previous discussion about not invoking `C.malloc` for empty strings or not, the decision was not to cover such cases in the example so maybe, we could leave it as it is here and do a proper handling in tinymem ?

codefromthecrypt: thanks for checking.

The main thing I wanted to know is if tinymem will return nil `ptr := unsafe.StringData("")` (because unsafe.StringData godoc says the impl can return nil). Just to make sure a crash doesn't happen.

By your tests, it seems it isn't. While better to understand why it doesn't (to know this is intentional vs accidental), we can put a comment meanwhile.

```suggestion
	ptr := unsafe.StringData(s) // Safe because TinyGo 0.27 doesn't return nil on empty
```

The only way I would be ok with special casing is if we cite the TinyGo implementation which proves zero is handled on free. I really didn't mean in any way to suggest special casing, just check to see if it crashes, ideally by looking at the source, and good enough the behavior (as you did above)

codefromthecrypt: So if you would like to special case this, which seems the case, you can do the special casing but have to cite why somehow.

If we want to use this behavior to cite why instead of looking at the impl of tinygo free, the special case comment would be something like

```go
if len(s) == 0 {
	// As of TinyGo 0.27, the below code returns 0,0 on empty,
	// so reduce overhead by returning zero.
	return 0, 0
}
```

I really think if we are at this point we should be citing why returning zero is fine (by adding a link to the code that handles zero in tinygo)

In any case, I will pass to @evacchi to complete review as I don't think I have more to offer on this.

lburgazzoli: In fact, `unsafe.StringData~ returns nil, but it seems safe

```go
package main

import (
	"fmt"
	"unsafe"
)

func main() {
	ptr := unsafe.StringData("")
	unp := unsafe.Pointer(ptr)
	uptr := uintptr(unp)

	fmt.Println(ptr, unp, uptr)
}
```

Produces:

```
➜ tinygo run main.go
<nil> <nil> 0
```

evacchi: so even though `unsafe.Pointer(nil)` is illegal, it does compile and run.

```
        u := unsafe.Pointer(nil)
        p := unsafe.Pointer(uintptr(u) + 10)
        println(u,p)
```

will print 

```
0x0 0xa
```

lburgazzoli: According to [tinygo code](https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/compiler/compiler.go#L1699-L1700), `unsafe.StringData` can actually return nil, however I've testes it and it seems that `stringToPtr` is still safe to use.

I've added a comment to explain the rationale, let me know if that's ok of it needs further clarification

ncruces: Yes, it's TinyGo's implementation of `unsafe.Pointer(nil)` that's off spec.
I think us returning zero is correct behaviour, so we should we test the empty string and `return 0, 0`?
Although it complicates the example, I find it highly likely that this will be copy pasted by people.

examples/allocation/tinygo/testdata/greet.go
```diff
@@ -72,13 +74,19 @@ func ptrToString(ptr uint32, size uint32) string {
 	}))
 }
 
-// stringToPtr returns a pointer and size pair for the given string in a way
-// compatible with WebAssembly numeric types.
+// stringToPtr returns a pointer and size pair to the underlying bytes of the given
+// string in a way compatible with WebAssembly numeric types. Since Go strings are
+// immutable, the pointer returned by stringToPtr must not be modified.
```
ncruces, line 79: I'd add another sentence here that ptr doesn't keep the string alive, just because I find it likely that this function (these 3 in fact) get copied around a lot.

ncruces (approved):

ncruces:
While we're at it, is there a reason not to do the below change:
https://github.com/tetratelabs/wazero/blob/3c01f86b7bf6e8400ad72ec52cf6d728e2e531f7/examples/allocation/tinygo/testdata/greet.go#L66-L68

Seems to match guidance from https://github.com/tinygo-org/tinygo/issues/1284

examples/allocation/tinygo/testdata/greet.go
lburgazzoli: No objection on my side however I think the whole example assumes the string is never empty (i.e. stringToPtr would panic at runtime if the given string would be empty within the current impl)

evacchi: @ncruces actually I just tried again with go1.20.4 (this morning go1.20.2) and `unsafe.Pointer(nil)` still yields 0x0

ncruces: Sure. I don't think it's a matter of "it doesn't work". It's more a matter that docs spec it's "invalid" so it may break in the future. I can raise an issue in the golang repo as to what "invalid" means in this context, but I don't think we'll get an answer soon.

evacchi: yeah I figured that "invalid" doesn't mean necessarily "broken", but it's a bit surprising that it doesn't blow up spectacularly :D

codefromthecrypt:
Sorry I wasn't sure which line we were proceeding with, so I commented on https://github.com/tetratelabs/wazero/pull/1436#pullrequestreview-1414067886 instead of this one. Feel free to proceed forward with whatever, as I've already bookmarked 1436 for next time

examples/allocation/tinygo/testdata/greet.go
codefromthecrypt: ps on the whole why does it work casting a nil unsafe.Pointer to uintptr, advice I was given was the scary thing about casting in the unsafe.Pointer godoc was about pointer arithmetic on against a nil coerced uintptr. Incidentally, I tried and that also didn't crash. So maybe the godoc is a relic.. anyway it isn't an issue.

codefromthecrypt: in other words the compiler doesn't seem to do anything and there's no code to crash, possibly it is just a warning that you might end up accidentally accessing some arbitrary address from 0 which could crash your process 🤷  in any case, it isn't relevant to what's going on here.

ncruces: Even then all I take from the sentence is that the nil `unsafe.Pointer` is invalid. I think we could say the same about `NULL` in C and this would still be correct code. We're never dereferencing this pointer anyway.

ncruces: I'd go as far as saying that WASM is the outlier allowing you to dereference the address at zero.

I wrap many of my memory accesses to panic in that case.

codefromthecrypt: interesting about the zero == panic defense certainly makes sense in anything except %.wat and maybe even in %.wat ;)

examples/allocation/tinygo/testdata/greet.go
```diff
@@ -72,13 +78,22 @@ func ptrToString(ptr uint32, size uint32) string {
...
 func stringToPtr(s string) (uint32, uint32) {
-	buf := []byte(s)
-	ptr := &buf[0]
+	// for an empty string the return value by TinyGo 0.27 is nil, however the
```
codefromthecrypt, line 85: As we aren't special casing anything I think it is better we say nothing at all here, or tighten it up to one succinct line. At the moment, this comment dominates the file and has little to do with wasm.

If anyone is interested, they could add details about this into the site doc here or a RATIONALE.md in tinymem, including footnotes here https://github.com/tetratelabs/wazero/pull/1436 or leave it.

lburgazzoli: removed

examples/allocation/tinygo/testdata/greet.go
```diff
@@ -72,13 +78,22 @@ func ptrToString(ptr uint32, size uint32) string {
...
+	// subsequent are still safe and result in the method returning pointer = 0
+	// and size = 0
+	//
+	// Note that:
+	// - we assume that the given string is never empty so there is no special
+	//   handling for such case
+	// - this method does not keep the given string alive, hence it is the
```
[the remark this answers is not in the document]
lburgazzoli: moved to godoc

evacchi:
merged #1436 thanks @lburgazzoli

PR #1437, opened 2023-05-04
experimental: add wazerotest package

achille-roussel:
This PR is a proposal to add `experimental/wazerotest` package containing stub implementations of types from the `api` package which are useful to construct tests.

Since https://github.com/tetratelabs/wazero/pull/1396, we added barriers for external packages to provide mocks of the wazero interfaces in contexts such as tests and benchmarks; see for example https://github.com/stealthrocket/wazergo/blob/main/wasm/memory.go

While programs can work around the limitation imposed by `internalapi.WazeroOnly`, doing so requires embedding the `api` interfaces which loosens type safety since the type _inherits_ all the methods and therefore automatically satisfies the interface even if it intended to provide different implementations of the methods. This causes issues that are difficult to catch such as typos in the method names or mismatching method signatures.

With this change, I have also included an example of how the types could be used by adding a `wazerotest.BenchmarkFunctionListener` function implementing a generic benchmark for function listeners, and used in in the `experimental/logging` package as a demonstration.

As the amount of software depending on wazero grows, providing models for how to test code integrating with the wazero APIs is going to become increasingly relevant, and will likely result in higher software quality in the wazero ecosystem. I believe that we are already seeing the demand for it: first within wazero where we instantiate types from the `internal/wasm` package to construct test conditions, but also in external packages like wazergo.

Please take a look and let me know if you have any questions or feedback!

### DIFF
--- a/experimental/wazerotest/wazerotest.go
+++ b/experimental/wazerotest/wazerotest.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"reflect"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -46,6 +47,11 @@ type Module struct {
 	exportedFunctionDefinitions map[string]api.FunctionDefinition
 	exportedGlobals             map[string]api.Global
 	exportedMemoryDefinitions   map[string]api.MemoryDefinition
+}
+
+// NewModule constructs a Module object with the given memory and function list.
+func NewModule(memory *Memory, functions ...*Function) *Module {
+	return &Module{Functions: functions, ExportMemory: memory}
 }
 
 func (m *Module) String() string {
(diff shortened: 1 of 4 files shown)


evacchi (approved):
I don't have specific comments to add, this looks fine to me! Maybe the other reviewers have more feedback :) (e.g. regarding naming etc)

codefromthecrypt:
interesting and makes sense like other code that have stub/mock implementations built-in. I will give more feedback in a while.

codefromthecrypt (commented):
I think I understand what's going on here. One thing I'm wondering is why this starts as experimental instead of internal first?

I feel we have enough code internally that could benefit from this, that it could be developed to test our own code for the same reasons as your PR. We already have other internal test helpers just not this one.

After it satisfies our own needs (including listener), seems better to expose as experimental then.

In other words, is the skipping step necessary because you have an outside use case now that needs this? If so, we can expose this but feels like it would change like tomorrow. Another way is to develop it as internal/... then once covers everything expose as experimental with a more proven feature set.

experimental/wazerotest/wazerotest.go
```diff
@@ -0,0 +1,605 @@
...
+// Memory is an implementation of the api.Memory interface, representing the
+// memory of a WebAssembly module.
+type Memory struct {
+	internalapi.WazeroOnlyType
+
+	// Byte slices holding the memory pages.
+	//
+	// It is the user's repsonsibility to ensure that the length of this byte
+	// slice is a multiple of the page size.
+	Pages []byte
```
codefromthecrypt, line 349: Pages and also "Byte slices" makes me think it should be `[][65536]byte`. Maybe `Raw` or `Buffer` is better name? Whatever we do I think it shouldn't be named differently than our default impl

experimental/wazerotest/wazerotest.go
```diff
@@ -0,0 +1,605 @@
...
+
+	// Min and max number of memory pages which may be held in this memory.
+	//
+	// Leaving Max to zero means no upper bound.
```
codefromthecrypt, line 353: fair enough. wasm lets you do min=0 max=0  different than min=0 max=unbounded but I think the former is semantically confusing and likely made just to satisfy syntax (else why bother adding memory at all)

experimental/wazerotest/wazerotest.go
```diff
@@ -0,0 +1,605 @@
...
+	Min uint32
+	Max uint32
+
+	// Lazily initialized when accessed through the module.
+	module *Module
+}
+
+const PageSize = 65536
```
codefromthecrypt, line 361: nit: if defining this use it in godoc like where we advise multiple

experimental/wazerotest/wazerotest.go
```diff
@@ -0,0 +1,605 @@
...
+	internalapi.WazeroOnlyType
+
+	// GoModuleFunction may be set to a non-nil value to allow calling of the
+	// function via Call or CallWithStack.
+	//
+	// It is the user's responsibility to ensure that the signature of this
+	// implementation matches the ParamTypes and ResultTypes fields.
+	GoModuleFunction api.GoModuleFunction
+
+	// Type lists representing the function signature. Those fields should be
```
codefromthecrypt, line 223: is it common to doc above multiple fields? If so, teach me. If just a shortcut for now, also ok

experimental/wazerotest/wazerotest.go
```diff
@@ -0,0 +1,605 @@
...
+// Global is an implementation of the api.Global interface, it represents a
+// global in a WebAssembly module.
+type Global struct {
+	internalapi.WazeroOnlyType
+
+	// Type of the global value, used to interpret bits of the Value field.
+	ValueType api.ValueType
+
+	// Value of the global packed in a 64 bits field.
+	Value uint64
```
codefromthecrypt, line 163: works for most things except v128 iirc

experimental/wazerotest/wazerotest.go
```diff
@@ -0,0 +1,605 @@
...
+type Module struct {
+	internalapi.WazeroOnlyType
+	exitStatus uint64
+
+	// The module name that will be returned by calling the Name method.
+	ModuleName string
+
+	// The list of functions of the module. Functions with a non-empty export
+	// names will be exported by the module.
+	Functions []*Function
```
codefromthecrypt, line 35: in our other code we seem to be using `[]Function` (specifically `[]Code`) instead of `[]*Function` and access by index. should we not also do that here for consistency within our codebase?

codefromthecrypt (approved):
I know this was a spike, but to unblock did a code review. I think to keep things easier to read, splitting the file up as we usually do (memory.go etc) will help, also as eventually we'd want tests.

Generally this is in good shape and I'm ok with merging this even before trying it with things except listener internally, though I think we really should scout around for internal opportunities before or after exposing it.

ncruces:
> I think I understand what's going on here. One thing I'm wondering is why this starts as experimental instead of internal first?

Ack. And I don't disagree. But part of my review will be based on how useful this is for me to delete [this](https://github.com/ncruces/go-sqlite3/blob/main/internal/util/mock.go), even if we start with internal, I do think it's helpful if we eventually export it (without committing to a stable API, forever experimental is fine with me).

experimental/wazerotest/wazerotest.go
ncruces: Can't say it is. But Go is pretty informal about this things. You can easily find a mix and match style of styles even in the same struct.
https://pkg.go.dev/net/http#Cookie
https://pkg.go.dev/database/sql#DBStats

ncruces (approved):
All I'm left wanting is a module constructor with an exported memory.

I ended up with this in my tests:
```go
func newMockModule() api.Module {
	return &wazerotest.Module{
		ExportMemory: &wazerotest.Memory{
			Pages: make([]byte, wazerotest.PageSize),
		},
	}
}
```

But maybe that's the price to pay for flexibility, so that's fine.

experimental/wazerotest/wazerotest.go
achille-roussel: I have seen it done across the stdlib, godoc shows the source code of struct declarations so we have a bit of liberty as to how we layout the documentation.

@codefromthecrypt I can change it if you think the layout hurts readability.

experimental/wazerotest/wazerotest.go
achille-roussel: We don't expose v128 in `api.ValueType` yet, I would suggest we keep it as-is for now and if we eventually expose a constant for the 128 bit vector we can revisit this API as well.

experimental/wazerotest/wazerotest.go
achille-roussel: I had a small argument with myself about this; I landed on pointer because the function can be referenced in the `StackFrame` type, so we can write code like this:
```go
fn := &wazerotest.Function{
  ...
}

mod := &wazerotest.Module{
  Functions: []*wazerotest.Function{
    fn,
  },
}

st := []wazerotest.StackFrame{
  {Function: fn},
}
```
That being said, all the code I wrote using this package does something like this:
```go
mod := &wazerotest.Module{
  Functions: []*wazerotest.Function{
    {
      ...
    },
  },
}

st := []wazerotest.StackFrame{
  {Function: mod.Function(0)},
}
```

I also felt like API consistency within the `wazerotest` package was a bit stronger with by-pointer, so I opted for making everything pointers.

Don't hesitate to let me know if I'm missing something, I may keep it this way otherwise :)

achille-roussel:
> One thing I'm wondering is why this starts as experimental instead of internal first?

The reason I would like to export this type is because I have use cases outsides of wazero for it. It has to exist in wazero is because it depends on `internalapi.WazeroOnlyType` to implement the wazero interfaces.

ncruces (approved):

achille-roussel:
Thanks @ncruces and @codefromthecrypt for the reviews!

I made the requested changes or followed up on the comments. Let me know if you have any more feedback, otherwise I will go ahead and merge 👍

ncruces (approved):

codefromthecrypt:
I'm happy to see this merged, also @ncruces can use it also! Great job deliberating and moving forward folks!

PR #1438, opened 2023-05-05
Adds AccessMode Write and Pwrite to platform.File

codefromthecrypt:
This adds `file.AccessMode` to substitute for type checking that formerly was used to see if the file might be writeable (e.g. casting to `io.Writer` so that you can check before actually writing).

This allows us to implement `File.Write` and `File.Pwrite` which might return `syscall.ENOSYS` on read-only implementations.

See #1417

### DIFF
--- a/internal/gojs/fs.go
+++ b/internal/gojs/fs.go
@@ -281,35 +281,28 @@ func (jsfsWrite) invoke(ctx context.Context, mod api.Module, args ...interface{}
 	callback := args[5].(funcWrapper)
 
 	if byteCount > 0 { // empty is possible on EOF
-		n, err := syscallWrite(mod, fd, fOffset, buf.Unwrap()[offset:offset+byteCount])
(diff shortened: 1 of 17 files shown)


internal/gojs/fs.go
```diff
@@ -281,35 +281,28 @@ func (jsfsWrite) invoke(ctx context.Context, mod api.Module, args ...interface{}
...
-	} else if writer, ok = f.File.File().(io.Writer); !ok {
-		err = syscall.EBADF
-	}
-
-	if err != nil {
-		return
-	}
-
-	if nWritten, e := writer.Write(p); e == nil || e == io.EOF {
-		// fs_js.go cannot parse io.EOF so coerce it to nil.
```
codefromthecrypt, line 308: EOF is irrelevant for write. It was accidentally copied from read.

mathetake (approved):

codefromthecrypt:
reader and seeker are next, then the more difficult or at least complicated ones.

PR #1441, opened 2023-05-08
Adds IsDir and Seek to platform.File

codefromthecrypt:
This adds `IsDir` and `Seek` to `platform.File`. `IsDir` is used in several places to error early if the file is the wrong type in file or directory. `Seek` replaces where we originally cast to `io.Seeker`, notably for `fd_seek`.

See #1417

### DIFF
--- a/internal/platform/file.go
+++ b/internal/platform/file.go
@@ -488,8 +488,8 @@ func (f *fsFile) Pread(p []byte, off int64) (n int, errno syscall.Errno) {
 func (f *fsFile) Seek(offset int64, whence int) (int64, syscall.Errno) {
 	if errno := f.isDirErrno(); errno != 0 {
 		return 0, errno
-	} else if whence > io.SeekEnd /* exceeds the largest valid whence */ {
-		return 0, syscall.EINVAL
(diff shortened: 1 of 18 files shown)


codefromthecrypt:
this also cleans up tests to remove most casting from `platform.File` to `fs.File` as it makes it easier to hunt down what's remaining.

internal/platform/file.go
```diff
@@ -395,14 +484,31 @@ func (f *fsFile) Pread(p []byte, off int64) (n int, errno syscall.Errno) {
 	return 0, syscall.ENOSYS // unsupported
 }
 
+// Seek implements File.Seek
+func (f *fsFile) Seek(offset int64, whence int) (int64, syscall.Errno) {
+	if errno := f.isDirErrno(); errno != 0 {
+		return 0, errno
+	} else if whence > io.SeekEnd /* exceeds the largest valid whence */ {
```
ncruces, line 491: Should `whence < 0` be tested as well?

codefromthecrypt: I like it. thanks!

PR #1443, opened 2023-05-08
wasip1: support non-blocking mode on stdio

achille-roussel:
This PR adds the ability to configure non-blocking mode on standard input and output when the underlying files are of type `*os.File`.

This is going to be needed to enable testing the latest Go changes with wazero, see https://go-review.googlesource.com/c/go/+/493356


codefromthecrypt (approved):
Great job and good timing!

PR #1444, opened 2023-05-09
experimental: add FunctionListener and FunctionListenerFactory implementations

achille-roussel:
This PR adds implementations of the `experimental.FunctionListenerFactory` which have been useful to implement listeners in external projects depending on wazero. Bringing those building blocks to wazero will likely benefit other users.

### DIFF
--- a/experimental/listener.go
+++ b/experimental/listener.go
@@ -40,9 +40,9 @@ type FunctionListenerFactoryKey struct{}
 // FunctionListenerFactory returns FunctionListeners to be notified when a
 // function is called.
 type FunctionListenerFactory interface {
-	// NewListener returns a FunctionListener for a defined function. If nil is
-	// returned, no listener will be notified.
-	NewListener(api.FunctionDefinition) FunctionListener
+	// NewFunctionListener returns a FunctionListener for a defined function.
+	// If nil is returned, no listener will be notified.
+	NewFunctionListener(api.FunctionDefinition) FunctionListener
 	// ^^ A single instance can be returned to avoid instantiating a listener
 	// per function, especially as they may be thousands of functions. Shared
 	// listeners use their FunctionDefinition parameter to clarify.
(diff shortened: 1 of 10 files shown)


experimental/listener.go
```diff
@@ -83,6 +83,155 @@ type FunctionListener interface {
...
+//
+// This function is useful when multiple listeners need to be hooked to a module
+// because the propagation mechanism based on installing a listener factory in
+// the context.Context used when instantiating modules allows for a single
+// listener to be installed.
+//
+// The stack iterator passed to the Before method is reset so that each listener
+// can iterate the call stack independently without impacting the ability of
+// other listeners to do so.
+func MultiFunctionListenerFactory(factories ...FunctionListenerFactory) FunctionListenerFactory {
```
codefromthecrypt, line 128: bikeshed: if multi is conventional, works for me. Often, I use the word "composite"

achille-roussel: I used the naming scheme of the `io` package which does a similar composition of multiple interface values:
- https://pkg.go.dev/io#MultiReader
- https://pkg.go.dev/io#MultiWriter

experimental/wazerotest/wazerotest.go
```diff
@@ -653,38 +653,53 @@ type StackFrame struct {
...
+func (si *StackIterator) Reset() {
+	si.index = -1
 }
 
-func newStackIterator(stack []StackFrame) *stackIterator {
-	si := &stackIterator{
-		stack: stack,
+// NewStackIterator constructs a stack iterator from a list of stack frames.
+// The top most frame is the last one.
+func NewStackIterator(stack ...StackFrame) *StackIterator {
```
codefromthecrypt, line 694: this made me realize we have exposed functions that are experimental in wazerotest. We have to make sure that when wazerotest moves out of experimental we scrub for things like this. I don't think anyone can depend on wazerotest meanwhile because it will drift even if a lot of it is stable (based on api package types)

achille-roussel: That's a good point.

The only experimental thing currently exposed in wazerotest is the benchmark function: https://pkg.go.dev/github.com/tetratelabs/wazero@v1.1.1-0.20230509001827-ed17e523ba21/experimental/wazerotest#BenchmarkFunctionListener

Let's move this function up to `experimental`, as well as the `StackIterator` type here?

achille-roussel: I went ahead and moved all dependencies on `experimental` from `wazerotest` in https://github.com/tetratelabs/wazero/pull/1444/commits/979f17735d1a04e4931950fd1eb8f4e8be69fc19

codefromthecrypt (approved):
LGTM thanks again!

PR #1447, opened 2023-05-09
Adds IsNonblock SetNonblock PollRead to platform.File

codefromthecrypt:
This adds `IsNonblock`, `SetNonblock` and `PollRead` to `platform.File`
allowing us to implement stdio files with the `platform.File` type
instead of alternate implementations.

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -201,21 +201,12 @@ func fdFdstatGetFn(_ context.Context, mod api.Module, params []uint64) syscall.E
 		return syscall.EBADF
 	} else if st, errno = f.Stat(); errno != 0 {
 		return errno
-	} else {
-		var nonblock bool
-		switch file := f.File.File().(type) {
-		case *sys.StdioFileReader:
-			nonblock = file.IsNonblock()
-		case *sys.StdioFileWriter:
-			nonblock = file.IsNonblock()
-		default:
-			if f.File.AccessMode() != syscall.O_RDONLY {
-				fdflags |= wasip1.FD_APPEND
-			}
-		}
-		if nonblock {
+	} else if fd <= sys.FdStderr {
+		if f.File.IsNonblock() {
 			fdflags |= wasip1.FD_NONBLOCK
 		}
+	} else if f.File.AccessMode() != syscall.O_RDONLY {
+		fdflags |= wasip1.FD_APPEND
 	}
 
 	var fsRightsBase uint32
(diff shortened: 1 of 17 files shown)


codefromthecrypt:
`TestShorthash` hangs locally not sure why

codefromthecrypt:
looks like I need to return fake stat for stdio, even when a real file, to pass this test https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/fd_filestat_get.rs#L1-L19

codefromthecrypt:
looks like stdout is causing the hang on TestShortHash (trial-error commenting things out and running with cmd/wazero)

codefromthecrypt:
TODO: scaffold tests for the new functions in platform/file_test.go as they are currently tested really later (wasi)

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -296,18 +287,21 @@ func fdFdstatSetFlagsFn(_ context.Context, mod api.Module, params []uint64) sysc
...
-	f, ok := fsc.LookupFile(fd)
-	if ok {
-		nonblock := wasip1.FD_NONBLOCK&wasiFlag != 0
-		switch file := f.File.File().(type) {
-		case *sys.StdioFileReader:
-			return platform.UnwrapOSError(file.SetNonblock(nonblock))
-		case *sys.StdioFileWriter:
-			return platform.UnwrapOSError(file.SetNonblock(nonblock))
+	// yet have support for sockets, so we make a special case.
+	if fd <= sys.FdStderr {
```
codefromthecrypt, line 291: special casing is a little weird, and isn't easy to contain (a couple docs), but keeping the end logic the same helps keep this PR from becoming too gnarly.

imports/wasi_snapshot_preview1/poll_test.go
```diff
@@ -539,14 +528,44 @@ func fdReadSubFd(fd byte) []byte {
...
 var fdReadSub = fdReadSubFd(byte(sys.FdStdin))
 
-type poller struct {
+// ttyStat returns fs.ModeCharDevice
+type ttyStat struct{}
+
+// Stat implements the same method as documented on platform.File
+func (ttyStat) Stat() (platform.Stat_t, syscall.Errno) {
+	return platform.Stat_t{
+		Mode:  fs.ModeDevice | fs.ModeCharDevice,
```
codefromthecrypt, line 537: as far as I know, we basically say it is a tty depending on if it is a char device not really anything else.

internal/platform/file.go
```diff
@@ -41,6 +42,30 @@ type File interface {
 	//   - syscall.O_RDWR: read-write, e.g. os.CreateTemp
 	AccessMode() int
 
+	// IsNonblock returns true if SetNonblock was successfully enabled on this
```
codefromthecrypt, line 45: here is the start of the new method definitions

codefromthecrypt (commented):
notes

imports/wasi_snapshot_preview1/fs.go
achille-roussel: Does it break if stdin/stdout/stderr are closed and the file descriptor resused?

It was one of the reasons I checked for the internal `sys.StdioFile*` types (my bad for not leaving a comment), tho I am also not 100% sure it was solid.

imports/wasi_snapshot_preview1/poll_test.go
achille-roussel: Probably a fine approximation for most use cases, if we wanted to be more specific we could check with https://github.com/mattn/go-isatty/blob/v0.0.18/isatty_tcgets.go#LL11C1-L12C1

imports/wasi_snapshot_preview1/poll_test.go
```diff
@@ -539,14 +528,44 @@ func fdReadSubFd(fd byte) []byte {
...
-func (p *poller) Poll(d time.Duration) (bool, error) {
+// PollRead implements the same method as documented on platform.File
+func (p *pollStdinFile) PollRead(*time.Duration) (ready bool, errno syscall.Errno) {
 	if p.ready {
-		return true, nil
+		return true, 0
 	} else {
-		return false, nil
+		return false, 0
 	}
```
achille-roussel, line 570: This looks like `return p.ready, 0` ?

achille-roussel (approved):

imports/wasi_snapshot_preview1/poll_test.go
codefromthecrypt: I'll put a note for now. thanks for the link to add!

codefromthecrypt:
> Does it break if stdin/stdout/stderr are closed and the file descriptor resused?

I'll switch to make the stdio files explicitly pre-opens (they are, just not in a way we can check).

imports/wasi_snapshot_preview1/poll_test.go
codefromthecrypt: lol yeah

imports/wasi_snapshot_preview1/fs.go
codefromthecrypt: rejigged. thanks!

codefromthecrypt:
cool. This should unblock some of the the other polling work 🥁

PR #1448, opened 2023-05-09
experimental: expose PC

pelletier:
Quick follow-up to https://github.com/tetratelabs/wazero/pull/1433. When benchmarking our use of `SourceOffset()`, we have noticed that it spends a significant amount of time in the binary search of the IR > offsets mapping:

image

Guest took ~60% of CPU time, so that's roughly 40% of the host time spent in the search. This makes it difficult for us to use it in a hot code path.

SourceOffset works in two steps: figure out the PC (usually cheap, as it's mostly stored on the stack), and resolve the PC into a source offset (the expensive part using `sort.Search`).

This change proposes to dissociate those two operations so that collecting the PC can be done on a hot code path, and computing the source offset can be done later, in a slower operation. I tried to make it clear in the comment that PC should be considered an opaque value, and should really only be used to identify a specific point in the program, and to be used with `SourceOffsetForPC`.

~It likely conflicts with https://github.com/tetratelabs/wazero/pull/1444, but we will rebase whichever gets merged second.~

Also, fixes that comment typo I missed in https://github.com/tetratelabs/wazero/pull/1433#discussion_r1187983379, sorry @codefromthecrypt 😬

### DIFF
--- a/experimental/experimental.go
+++ b/experimental/experimental.go
@@ -21,6 +21,11 @@ type InternalModule interface {
 	Global(i int) api.Global
 }
 
+// ProgramCounter is an opaque value representing a specific execution point in
+// a module. It is meant to be used with Function.SourceOffsetForPC and
+// StackIterator.
+type ProgramCounter uint64
+
 // InternalFunction exposes some information about a function instance.
 type InternalFunction interface {
 	// Definition provides introspection into the function's names and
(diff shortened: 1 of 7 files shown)


pelletier:
Rebased against `main` after https://github.com/tetratelabs/wazero/commit/daad2fdb660d50020c3906bc26cdcbf7c17054f3.

evacchi (approved):
looks good to me! But since you say:


>I tried to make it clear in the comment that PC should be considered an *opaque value*, and should really only be used to identify a specific point in the program, and to be used with `SourceOffsetForPC`.

would it make sense to also declare `pc` as a `type ProgramCounter uint64` so that it is actually "more opaque"?

pelletier:
@evacchi sounds good to me. Added!

achille-roussel (approved):
I confirmed that we perf is much better when we can offload the source offset resolution outside of the critical path 👍

PR #1449, opened 2023-05-09
Follow unsafe rules.

ncruces:
See: https://pkg.go.dev/unsafe
And detailed commit messages.


ncruces:
Seems the detailed commit messages don't go through, so here's the summary:
- https://github.com/tetratelabs/wazero/pull/1449/commits/1385324aa64221eb9f6110da1496fd8d5f410bec is about unsafe rule 4: _"Conversion of a Pointer to a uintptr when calling syscall.Syscall"_. [^1]
- https://github.com/tetratelabs/wazero/pull/1449/commits/7ffea1dae2e8e79a12acaf76feaa9d1d1bb89b61 is about unsafe rule 6: _"Conversion of a reflect.SliceHeader or reflect.StringHeader Data field to or from Pointer"_. [^2]

[^1]: These conversions (`unsafe.Pointer` to `uintptr`) should happen as late as possible and typically directly on the argument list of the call. If so, they're safe for calling ASM, `Syscall` and funcs tagged `go:uintptrescapes`.

[^2]: `reflect.SliceHeader` and `reflect.StringHeader` should only be used to point to an existing slice or string variable, they should not be used to create temporaries. See: https://go.dev/issue/40701

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -1976,18 +1976,18 @@ func pathSymlinkFn(_ context.Context, mod api.Module, params []uint64) syscall.E
...
-	return *(*string)(unsafe.Pointer(&reflect.SliceHeader{ //nolint
-		Data: uintptr(unsafe.Pointer(&buf[0])),
-		Len:  l,
-		Cap:  l,
-	}))
+func bufToStr(buf []byte) (s string) {
+	// TODO: use unsafe.String after flooring Go 1.20.
+	hdr := (*reflect.StringHeader)(unsafe.Pointer(&s))
+	hdr.Data = uintptr(unsafe.Pointer(&buf[0]))
+	hdr.Len = len(buf)
```
achille-roussel, line 1989: Often this is simplified as
```suggestion
	s = *(*string)(unsafe.Pointer(&buf))
```
It has the advantage of not panicking if the input byte slice is empty.

achille-roussel (approved):
Thanks for making wazero safer!

I'm curious if you know of a static checker we could run in CI to ensure that we don't add regressions in the future?

ncruces:
> I'm curious if you know of a static checker we could run in CI to ensure that we don't add regressions in the future?

Some of these had `nolint` I assume because of this.

I also had VSCode complain about some of these, because `go vet` was flagging them.

imports/wasi_snapshot_preview1/fs.go
ncruces: You're right, of course:
https://github.com/golang/go/blob/go1.19/src/strings/builder.go#L48

ncruces:
Sorry for the force push, I accidentally committed other stuff I think should be in a separate PR.

ncruces:
> +1 to consider configuring some static checker in the future thinking

The problem is that the same rule that would've caught these triggers these lines as well:
https://github.com/tetratelabs/wazero/blob/daad2fdb660d50020c3906bc26cdcbf7c17054f3/internal/engine/compiler/compiler_post1_0_test.go#L817-L823
https://github.com/tetratelabs/wazero/blob/daad2fdb660d50020c3906bc26cdcbf7c17054f3/internal/engine/compiler/compiler_post1_0_test.go#L825-L831

Which do violate unsafe rules, in that they're recreating GC traced objects from `uintptr` (they're safe as long as the objects are kept alive by something else - I assume they are - and as long we assume a [no-moving GC](https://github.com/go4org/unsafe-assume-no-moving-gc)).

And again, `go vet` does not allow ignoring this line, unless we wrap it with something that figures that out.

ncruces:
Merging, but lets continue to discuss CI options.

codefromthecrypt:
great work, on my bookmark list now!

wrt static checkers, seems a lot of the `go vet` alternatives aren't maintained often, which could be ok if we are ok removing them later. https://github.com/search?q=go%20vet&type=repositories

PR #1452, opened 2023-05-10
experimental: remove error parameter from FunctionListener.After

achille-roussel:
Following up on our discussion in #wazero-dev, I would like to suggest we remove the `error` parameter from `FunctionListener.After` since it is currently unused: wazero always passes the value `nil`.

I believe this change is worth it so users of `FunctionListener` do not have to concern themselves with understanding how to deal with the parameter nor write tests for code paths that will never be taken. It also means that we do not need to write documentation explaining why the parameter exists, and why it's currently unused.

Since this is still an experimental API, the breaking change should be acceptable, as well as considering bringing back a form of error handling in the future when we do have the ability to propagate errors to function listeners.

---

As part of this PR, I also made a few aesthetic changes:
- I renamed `paramValues` and `resultValues` to `params` and `results` since those were terms that were much more common within wazero to represent parameters and result values of function calls
- I modified the internal implementation of the logging listener's `logIndented` method to accept a function parameter instead of a boolean in order to configure which of the `logParams` or `logResults` methods gets called

---

### DIFF
--- a/experimental/listener.go
+++ b/experimental/listener.go
@@ -55,14 +55,14 @@ type FunctionListener interface {
 	//	   instance or parent of the result.
 	//   - mod: the calling module.
 	//   - def: the function definition.
-	//   - paramValues:  api.ValueType encoded parameters.
+	//   - params:  api.ValueType encoded parameters.
 	//   - stackIterator: iterator on the call stack. At least one entry is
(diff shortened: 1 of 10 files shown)


codefromthecrypt (approved):
makes sense

PR #1453, opened 2023-05-10
experimental: remove return value of FunctionListener.Before

achille-roussel:
Another follow up from our discussion in #wazero-dev, this PR modifies the signature of `FunctionListener.Before` to remove the returned `context.Context`.

This change aims to remove the internal context stack tracking return values of the `Before` methods. For context, I am copying a quote from the Slack discussion:

> While I find the API somewhat elegant, it has proven to be unusable in most cases where we've used FunctionListener because embedding values in a context requires at least one heap allocation, but sometimes two or more. The listeners often being on hot code paths of the guest program, and this heap allocation ends up being extremely costly.
>
> We worked around the issue by implementing internal stacks within the listeners, in the same way, that the wazero engine keeps track of a stack of the context.Context returned by Before. Implementing those stacks as slices allows reusing the underlying memory, which ends up being an extremely cheap mechanism for correlation between the Before and After methods: the allocation of context values added 50 to 100ns of latency per function, while the internal stack ends up adding 5 to 10ns (~10x cheaper).
>
> However, we still end up paying for the engine's internal context stack, despite not needing it. On performance-critical code paths, it almost doubles the overall latency of invoking function listeners.

## Using stacks to correlate Before and After

One of the issues was the fact that `After` was not always being called. For example, if a module hit a trap, or a host function panicked, the call to the `After` method would be skipped, which made it challenging to implement stack-based context tracking internally in the listener. This limitation actually impacted the host logging listeners and caused multiple tests within wazero to fail after making the initial change.

I am suggesting that we address it by adding a third `Abort` method on `FunctionListener`, which is invoked when a module execution is aborted, in which case there are no result values to pass to `After` and instead of communicate the reason for aborting to the call to `Abort`.

The guarantee that either `After` or `Abort` will be called allows listeners to use an internal stack to carry state from `Before` to one of these methods.

## State propagation through the call stack

One of the details that @codefromthecrypt shared in #wazero-dev was that the second reason for returning a `context.Context` was to pass state not just between `Before` and `After`, but also to other listeners and host function calls through the call stack. Use cases for that include telemetry like tracing, where a listener creates a span and injects it in the context that eventually gets passed as input to another callee.

While this use case isn't covered in the change I'm submitting here, I believe that it can be served by injecting a _mutable context value_ (as @codefromthecrypt described it). Here would be an example:

```go
// Completely hypothetical API, just to demonstrate the idea

rootSpan := &tracing.Span{
  ParentSpan: ...,
}

ctx := context.WithValue(ctx, SpanKey{}, &rootSpan)
...
// eventually instantiate module or call exported function with ctx
```
```go
func (f *functionTracer) Before(ctx context.Context, ...) {
  spanPtr := ctx.Value(SpanKey{}).(**tracing.Span)
  span := &tracing.Span{
    ParentSpan: *spanPtr,
    ...
  }
  *spanPtr = span // push
  span.Start()
}

func (f *functionTracer) After(ctx context.Context, ...) {
  spanPtr := ctx.Value(SpanKey{}).(**tracing.Span)
  span := *spanPtr
  span.Finish()
  *spanPtr = spanPtr.Parent // pop
}

// + Abort
```
In the future, injecting and extracting the context value may also be abstracted into a generic mechanism to bridge between the `FunctionListener` interface and higher level abstractions.

Some of the benefits I'm seeing to this approach is that it reduces the baseline cost of function listeners, the higher cost of maintaining stacks is only paid when necessary, and can be optimized individually by each listener to suite the use cases.

The trade off is that we offer something that is a bit less _packaged_, in some ways we are lowering the abstraction level of function listeners, leaving more of the implementation decisions to the host.

### DIFF
--- a/experimental/listener.go
+++ b/experimental/listener.go
@@ -95,7 +95,6 @@ type FunctionListener interface {
 	// # Notes
 	//
 	//   - api.Memory is meant for inspection, not modification.
-	//   - Only functions called with api.Function are guaranteed to see this callback.
 	Abort(ctx context.Context, mod api.Module, def api.FunctionDefinition, err error)
 }
 
--- a/experimental/listener.go
+++ b/experimental/listener.go
@@ -46,8 +46,11 @@ type FunctionListenerFactory interface {
 // FunctionListener can be registered for any function via
 // FunctionListenerFactory to be notified when the function is called.
 type FunctionListener interface {
-	// Before is invoked before a function is called. The returned context will
-	// be used as the context of this function call.
+	// Before is invoked before a function is called.
+	//
+	// There is always one corresponding call to After or Abort for each call to
+	// Before. This guarantee allows the listener to maintain an internal stack
+	// to perform correlations between the entry and exit of functions.
 	//
 	// # Params
 	//
@@ -62,19 +65,31 @@ type FunctionListener interface {
 	//
 	// Note: api.Memory is meant for inspection, not modification.
 	// mod can be cast to InternalModule to read non-exported globals.
-	Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, params []uint64, stackIterator StackIterator) context.Context
+	Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, params []uint64, stackIterator StackIterator)
 
 	// After is invoked after a function is called.
 	//
 	// # Params
 	//
-	//   - ctx: the context returned by Before.
+	//   - ctx: the context of the caller function.
 	//   - mod: the calling module.
 	//   - def: the function definition.
 	//   - results: api.ValueType encoded results.
 	//
 	// Note: api.Memory is meant for inspection, not modification.
 	After(ctx context.Context, mod api.Module, def api.FunctionDefinition, results []uint64)
+
+	// Abort is invoked when a function does not return due to a trap or panic.
+	//
+	// # Params
+	//
+	//   - ctx: the context of the caller function.
+	//   - mod: the calling module.
+	//   - def: the function definition.
+	//   - err: the error value representing the reason why the function aborted.
+	//
+	// Note: api.Memory is meant for inspection, not modification.
+	Abort(ctx context.Context, mod api.Module, def api.FunctionDefinition, err error)
 }
 
 // FunctionListenerFunc is a function type implementing the FunctionListener
@@ -88,14 +103,18 @@ type FunctionListener interface {
 type FunctionListenerFunc func(context.Context, api.Module, api.FunctionDefinition, []uint64, StackIterator)
 
 // Before satisfies the FunctionListener interface, calls f.
-func (f FunctionListenerFunc) Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, params []uint64, stackIterator StackIterator) context.Context {
+func (f FunctionListenerFunc) Before(ctx context.Context, mod api.Module, def api.FunctionDefinition, params []uint64, stackIterator StackIterator) {
 	f(ctx, mod, def, params, stackIterator)
-	return ctx
 }
 
 // After is declared to satisfy the FunctionListener interface, but it does
 // nothing.
 func (f FunctionListenerFunc) After(context.Context, api.Module, api.FunctionDefinition, []uint64) {
+}
+
+// Abort is declared to satisfy the FunctionListener interface, but it does
+// nothing.
+func (f FunctionListenerFunc) Abort(context.Context, api.Module, api.FunctionDefinition, error) {
 }
 
 // FunctionListenerFactoryFunc is a function type implementing the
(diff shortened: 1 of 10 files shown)


codefromthecrypt (commented):
I can see the sense in not having the error param, and also the return removal, just not adding Abort when it isn't called (unless I am missing that). I think if we are removing things because they aren't implemented, we shouldn't add new unimplemented things as it makes it confusing. Can you separate these steps?

one idea is to capture your notes about a potential future on abort to the original issue about listeners for telemetry. Personally, I don't really want to noodle on pros and cons of abort when it isn't implementable anyway :D

If I'm missing where abort is used, disregard above and we can start the design discussion, just yeah hopefully we can skip it!

achille-roussel:
> I can see the sense in not having the error param, and also the return removal, just not adding Abort when it isn't called (unless I am missing that). I think if we are removing things because they aren't implemented, we shouldn't add new unimplemented things as it makes it confusing. Can you separate these steps?

This was probably a miss on my part in the description, I can try to clarify.

The host logging code was depending on context propagation to calculate things like indentation levels. Removing the context means we needed a different way. Because the intention was to move state tracking into the function listener instead of having it in the wazero engine, I modified the loggers to track their state without the context, which highlighted the fact that sometimes `After` is not called.

For example, `Test_procExit` makes two calls to `proc_exit` which never calls `After` because the function never returns, and invalidated the state, causing the output to change and the tests to fail. This is just one example of the many places where we use the loggers to validate beahvior, wasi_snapshot_preview1 had many such tests.

In order to fix the issue and reconcile the state of the loggers, I added the `Abort` logic so everything that gets pushed to the inner stack gets popped. Now I couldn't just ensure that `After` was going to always be called because that would cause the loggers to print return values in places where the tests validated that there would be no returns (e.g. `proc_exit`).

I agree that adding `Abort` is expanding the scope, here is the place where it's needed in wazero https://github.com/tetratelabs/wazero/pull/1453/files#diff-d60dc448f4c159bc82ccabf4dca5fb109524074e4fc7810d766bcecb7a649af4R229-R232

Something I have to admit I did not explore was to take the approach I described in **State propagation through the call stack** with the loggers. I can probably get around not adding `Abort` with this model.

After exploring this code further, I feel like the presence of `Abort` is meaningful, tho I understand the desire not to couple changes:
- I can rework the loggers to not need `Abort`, and submit that as a separate PR if you think it's important to separate the change and discussion
- we can continue the discussion here until we feel comfortable with the change or choose not to, there is no rush!

internal/engine/compiler/engine.go
```diff
@@ -844,7 +857,13 @@ func (ce *callEngine) deferredOnCall(recovered interface{}) (err error) {
 				break
 			}
 		}
+
 		err = builder.FromRecovered(recovered)
+		// Call listeners in revers order since they were captured while
+		// rewinding the call stack.
+		for i := len(functionListeners) - 1; i >= 0; i-- {
+			functionListeners[i].Abort(ctx, m, functionListeners[i].def, err)
```
codefromthecrypt, line 865: this was Abort being called, which I missed. I'll re-do review as this changes things!

experimental/listener.go
```diff
@@ -62,19 +65,31 @@ type FunctionListener interface {
...
 	//
 	// # Params
 	//
-	//   - ctx: the context returned by Before.
+	//   - ctx: the context of the caller function.
 	//   - mod: the calling module.
 	//   - def: the function definition.
 	//   - results: api.ValueType encoded results.
 	//
 	// Note: api.Memory is meant for inspection, not modification.
```
codefromthecrypt, line 79:
```suggestion
	// # Notes
	//
	//   - api.Memory is meant for inspection, not modification.
	//   - This is not called when a host function panics, or a guest function traps.
	//      See Abort for more details.
```

experimental/listener.go
```diff
@@ -62,19 +65,31 @@ type FunctionListener interface {
...
+	// Abort is invoked when a function does not return due to a trap or panic.
+	//
+	// # Params
+	//
+	//   - ctx: the context of the caller function.
+	//   - mod: the calling module.
+	//   - def: the function definition.
+	//   - err: the error value representing the reason why the function aborted.
+	//
+	// Note: api.Memory is meant for inspection, not modification.
```
codefromthecrypt, line 91:
```suggestion
	// # Notes
	//
	//   - api.Memory is meant for inspection, not modification.
	//   - Only functions called with api.Function are guaranteed to see this callback.
```

codefromthecrypt (commented):
If I understand correctly, the main thing people need to know is that Abort may not be called at all. If I understand what's going on, we would only see this callback on the entry point calls (e.g. api.Function Call).

A test may show what I mean, for example, if there's a call stack of 3 I would expect not all three listeners to get the Abort(). If I'm wrong, please correct me! In any case, a test maybe can explain this working or not working.

achille-roussel:
> A test may show what I mean, for example, if there's a call stack of 3 I would expect not all three listeners to get the Abort(). If I'm wrong, please correct me! In any case, a test maybe can explain this working or not working.

My intention is for each call to `Before` on a listener to be matched with either one call to `After` or one call to `Abort`. I believe that's what I implemented because some of the existing tests wouldn't pass otherwise, but I can add an extra test to explicitly validate that.

To illustrate, say there is a call stack like this, each function have listeners installed:
```
f1
. f2
  . f3
```
When `f1`, then `f2`, and `f3` are called, each of their listeners' `Before` method is invoked.

- If each of these functions returns normally, their listeners `After` methods are called, but `Abort` is not
- If `f3` traps, then `Abort` is called on the listener of `f3`, `f2`, and `f1`, none of their `After` methods are called
- If `f3` returns, but after that `f2` traps, the `After` method of `f3`'s listener is called, but the `Abort` method of `f2`, then `f1` are called

It's the guarantee that for each call to `Before` exactly one of `After` or `Abort` will be called which allows listeners to track state associated with the call stack internally, and it doesn't get out of sync.

experimental/listener.go
```diff
@@ -62,19 +65,38 @@ type FunctionListener interface {
...
+	//
+	//   - ctx: the context of the caller function.
+	//   - mod: the calling module.
+	//   - def: the function definition.
+	//   - err: the error value representing the reason why the function aborted.
+	//
+	// # Notes
+	//
+	//   - api.Memory is meant for inspection, not modification.
+	//   - Only functions called with api.Function are guaranteed to see this callback.
```
codefromthecrypt, line 98: ok so then maybe this comment isn't valid, as unless I missed something we are guaranteed, regardless of if the stack is host or guest functions, these callbacks apply.
```suggestion
```

codefromthecrypt (approved):
Ok I suppose the status quo is better than I thought it was, or at least this change makes it better, either way. go for it!

experimental/listener.go
achille-roussel: Bad copy/paste, thanks for flagging!

PR #1454, opened 2023-05-10
Adds Readdir to platform.File and removes internal use of fs.FS

### LINKED ISSUE #1417: complete internal platform.File

codefromthecrypt:
This moves the existing semantics of `Readdir` to `platform.File` which allows us to remove internal use of `fs.File` completely.

Notes:
* While this completes feature parity on platform.File, there is still work to do. Notably, we need to packaging more sensible (not split between platform and sysfs), and separate the os.File impls from things generably usable when backed by a `fs.File`. Finally, the windows implementation should be retrofit to our file type. All these work will proceed next and once done we can polish the shape for possibly experimental export. (ETA weeks probably)
* To retrofit the current impl to `fs.File` is a lot of work, so choosing not to do that, especially not in this PR. We can backfill coverage that isn't already redundant to our testslater, when things settle.
* `Readdir` is unlike the system call, and a TODO is there to possibly refactor this. However, the current implementation works for wasip1 and gojs. We should look at wasip not 1 before doing more engineering on it.

Closes #1417

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -844,7 +844,7 @@ func fdReaddirFn(_ context.Context, mod api.Module, params []uint64) syscall.Err
 		if errno != 0 {
 			return errno
 		}
-		rd, dir = f.File.File(), f.ReadDir
+		rd, dir = f.File, f.ReadDir
 	}
 
 	// First, determine the maximum directory entries that can be encoded as
@@ -882,7 +882,7 @@ func fdReaddirFn(_ context.Context, mod api.Module, params []uint64) syscall.Err
 	if entryCount := len(dirents); entryCount < maxDirEntries {
 		// Note: platform.Readdir does not return io.EOF as it is
 		// inconsistently returned (e.g. darwin does, but linux doesn't).
-		l, errno := platform.Readdir(rd, maxDirEntries-entryCount)
+		l, errno := rd.Readdir(maxDirEntries - entryCount)
 		if errno != 0 {
 			return errno
 		}
@@ -925,7 +925,7 @@ func fdReaddirFn(_ context.Context, mod api.Module, params []uint64) syscall.Err
 
 // dotDirents returns "." and "..", where "." because wasi-testsuite does inode
 // validation.
-func dotDirents(f *sys.FileEntry) ([]*platform.Dirent, syscall.Errno) {
+func dotDirents(f *sys.FileEntry) ([]platform.Dirent, syscall.Errno) {
 	if isDir, errno := f.File.IsDir(); errno != 0 {
 		return nil, errno
 	} else if !isDir {
@@ -943,7 +943,7 @@ func dotDirents(f *sys.FileEntry) ([]*platform.Dirent, syscall.Errno) {
 			dotDotIno = st.Ino
 		}
 	}
-	return []*platform.Dirent{
+	return []platform.Dirent{
 		{Name: ".", Ino: dotIno, Type: fs.ModeDir},
 		{Name: "..", Ino: dotDotIno, Type: fs.ModeDir},
 	}, 0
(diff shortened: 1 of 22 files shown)


codefromthecrypt:
I'll try to detangle os.File impls next

PR #1455, opened 2023-05-11
Separates fs.File from os.File implementations of platform.File

### LINKED ISSUE #1417: complete internal platform.File

codefromthecrypt:
Before, we had shared code which constantly checked if a file had a
descriptor or not. This was because the same type was used for files
backed by a real file (`os.File`) vs a potentially fake one (`fs.File`).
This untangles that by sifting logic into two types `fsFile` and
`osFile`. In doing so, we can eliminate more complexity around
windows-specific handling as it is a variation on `osFile`.

This also completes moving file-specific logic from `sys.FileEntry` into
`platform.File`, such as the responsibilty to cache file properties.
* push how to implement `rewinddir` into the `platform.File` type as
  `Seek(0,SeekStart)`. This removes the need to expose `File.Path` which
  is often incorrect. `fs.File` implementations receive a `name` on
  `fs.Open` which isn't usually an absolute path.
* push how to set append mode into the `platform.File` type as
  `IsAppend` and `SetAppend`. This removes the other case where
  re-opening a file happened externally.
* push cached inode into the file as `Ino` on `plaform.File`.

Closes #1417


codefromthecrypt:
this needs more work because opening a directory in Windows is complicated. I think I need to store two versions of the path.. the view the FS wants of it (sub rooted) and the real path, so that when we need to internally re-open we use the correct one. These notes are really for myself as I doubt anyone wants to understand the labyrinth of getting portable outcomes even in windows :D

codefromthecrypt:
I'm on the final lap with this, and excited to be able to move a couple features into the file implementation

`FSContext.ReOpenDir(fd int32)` -> `File.Seek(0, io.SeekCurrent)`
`FSContext.ChangeOpenFlag(fd int32, flag int)` -> `File.SetAppend(bool)`

In trying to streamline implementation, I moved some entry guards out of the "windows wrapper", and looks like I need to put some of them back. I'll buff out these issues and implement the tests tomorrow and ideally move out of draft.

codefromthecrypt:
finally everything passes even on windows. I'll clean up the PR

codefromthecrypt (commented):
notes

internal/sys/fs.go
```diff
@@ -227,7 +245,11 @@ type FileEntry struct {
 	// Name is the name of the directory up to its pre-open, or the pre-open
 	// name itself when IsPreopen.
 	//
-	// Note: This can drift on rename.
+	// # Notes
+	//
+	//   - This can drift on rename.
+	//   - This relates to the guest path, which is not the real file path
```
codefromthecrypt, line 251: I made notes here because the result here (used by wasi) is not the file's real path, rather it is about the what the guest sees as the path.

internal/sysfs/dirfs_test.go
```diff
@@ -758,7 +758,7 @@ func TestDirFS_Truncate(t *testing.T) {
 
 			name := "truncate"
 			realPath := path.Join(tmpDir, name)
-			require.NoError(t, os.WriteFile(realPath, content, 0o0600))
+			require.NoError(t, os.WriteFile(realPath, content, 0o0666))
```
codefromthecrypt, line 761: in windows these always turn to world write, we in other places use 0o0666 because of this. I noticed a couple places we didn't.

PR #1456, opened 2023-05-11
compiler: avoid some unsafe.Pointers

ncruces:
These seem unneeded, so best to avoid.

### DIFF
--- a/internal/engine/compiler/arch.go
+++ b/internal/engine/compiler/arch.go
@@ -11,10 +11,9 @@ var newArchContext func() archContext
 
 // nativecall is used by callEngine.execWasmFunction and the entrypoint to enter the compiled native code.
 // codeSegment is the pointer to the initial instruction of the compiled native code.
-// ce is "*callEngine" as uintptr.
 //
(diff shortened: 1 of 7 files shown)


achille-roussel (approved):
🔥

ncruces:
Do we know the recent source of test failures?

achille-roussel:
Looks like it probably was https://github.com/tetratelabs/wazero/pull/1453, weird that the tests passed in that PR 🤔 

It seems like a linter issue, one of the fields on the compiler engine isn't used anymore, do you want to address as part of this change?

ncruces:
Fixed one of the CI failures, the other one is pretty out of scope.
Trying to address it elsewhere.

mathetake (approved):

codefromthecrypt:
I nuked the cache with the following

```bash
$ gh extension install actions/gh-actions-cache
$ gh actions-cache list
# look for stuff
$ gh actions-cache delete --confirm gojs-test-binary-1.20
```

then had to re-kick all the jobs in because the failure had a dependency, and that was easier.

PR #1457, opened 2023-05-11
Fix base rights for character devices

Pryz:
While testing the latest `python.wasm` from https://github.com/vmware-labs/webassembly-language-runtimes, we were not able to get the python shell working. Turns out this was a rights issue with `fd=0`.

Per https://github.com/WebAssembly/wasi-libc/blob/a6f871343313220b76009827ed0153586361c0d5/libc-bottom-half/sources/isatty.c#L13-L18, "A tty is a character device that we can't seek or tell on.".

This change re-use `fileRightsBase` and remove FD_SEEK and FD_TELL rights for all character devices.

### DIFF
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"syscall"
 	"testing"
+	gofstest "testing/fstest"
 	"time"
 
 	"github.com/tetratelabs/wazero"
@@ -239,6 +240,16 @@ func Test_fdFdstatGet(t *testing.T) {
 	fsc := mod.(*wasm.ModuleInstance).Sys.FS()
 	preopen := fsc.RootFS()
 
+	// replace stdin with a fake TTY file.
+	// TODO: Make this easier once we have in-memory platform.File
+	stdin, _ := fsc.LookupFile(sys.FdStdin)
+	stdinFile, errno := sysfs.Adapt(&gofstest.MapFS{"stdin": &gofstest.MapFile{
+		Mode: fs.ModeDevice | fs.ModeCharDevice | 0o600,
+	}}).OpenFile("stdin", 0, 0)
+	require.EqualErrno(t, 0, errno)
+
+	stdin.File = stdinFile
+
 	fileFD, errno := fsc.OpenFile(preopen, file, os.O_RDONLY, 0)
 	require.EqualErrno(t, 0, errno)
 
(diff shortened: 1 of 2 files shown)


achille-roussel (commented):
Let's add a test in https://github.com/tetratelabs/wazero/blob/main/imports/wasi_snapshot_preview1/fs_test.go#L232 to guarantee that we won't introduce a regression.

codefromthecrypt:
I can help add a test to this

codefromthecrypt (approved):
Thanks for getting to the bottom of this. A broke build is a bad day for everyone, so thanks for fixing it!

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -239,6 +240,16 @@ func Test_fdFdstatGet(t *testing.T) {
 	fsc := mod.(*wasm.ModuleInstance).Sys.FS()
 	preopen := fsc.RootFS()
 
+	// replace stdin with a fake TTY file.
+	// TODO: Make this easier once we have in-memory platform.File
```
codefromthecrypt, line 244: This maybe by next week depending on how much zombification happens during intercontinental back to back travel

achille-roussel (approved):
🚢

codefromthecrypt:
@achille-roussel can you take a look to see if this is legit in the go tests?

> 2023/05/12 00:26:03 write /tmp/go-example-stdout-Example_minimal-0.txt: Bad file number

achille-roussel:
v1.1.0
```
$ PATH=$PATH:$PWD/misc/wasm GOROOT=$PWD GOARCH=wasm GOOS=wasip1 ./bin/go test ./src/archive/tar
ok  	archive/tar	1.132s
```

this PR
```
$ PATH=$PWD/../../github.com/stealthrocket/wazero:$PATH:$PWD/misc/wasm GOROOT=$PWD GOARCH=wasm GOOS=wasip1 ./bin/go test ./src/archive/tar
2023/05/12 00:46:00 write /var/folders/qc/zv1mt0j11sscgjygfx47nxlm0000gn/T//go-example-stdout-Example_minimal-0.txt: Bad file number
FAIL	archive/tar	1.151s
FAIL
```

main
```
$ PATH=$PWD/../../github.com/stealthrocket/wazero:$PATH:$PWD/misc/wasm GOROOT=$PWD GOARCH=wasm GOOS=wasip1 ./bin/go test ./src/archive/tar
2023/05/12 00:46:20 write /var/folders/qc/zv1mt0j11sscgjygfx47nxlm0000gn/T//go-example-stdout-Example_minimal-1.txt: Bad file number
FAIL	archive/tar	0.861s
FAIL
```

So we have a regression but it seems like it's not due to this change.

codefromthecrypt:
thanks for the analysis @achille-roussel. I'll look into the broke main next!

PR #1458, opened 2023-05-12
Clears O_CREATE when re-opening a file to set flags

codefromthecrypt:
Specifically O_CREATE|O_EXCL triggers EEXIST errors when re-opening a file. This ensures we clear the O_CREATE flag, as the goal is to re-create the file descriptor, not the file.

This also fixes a bug where we unconditionally write O_APPEND to any writable file in fdstat results.

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -205,7 +205,7 @@ func fdFdstatGetFn(_ context.Context, mod api.Module, params []uint64) syscall.E
 		if f.File.IsNonblock() {
 			fdflags |= wasip1.FD_NONBLOCK
 		}
-	} else if f.File.AccessMode() != syscall.O_RDONLY {
+	} else if f.IsAppend() {
 		fdflags |= wasip1.FD_APPEND
 	}
 
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -255,6 +255,11 @@ type cachedStat struct {
 	Ino uint64
 }
 
+// IsAppend returns true if the file is open with syscall.O_APPEND.
+func (f *FileEntry) IsAppend() bool {
+	return f.openFlag&syscall.O_APPEND != 0
+}
+
 // Inode returns the cached inode from of platform.Stat_t or an error if it
 // couldn't be retrieved.
 func (f *FileEntry) Inode() (ino uint64, errno syscall.Errno) {
@@ -462,6 +467,8 @@ func (c *FSContext) ChangeOpenFlag(fd int32, flag int) syscall.Errno {
 		return errno
 	} else if isDir {
 		return syscall.EISDIR
+	} else if flag&syscall.O_APPEND == f.openFlag&syscall.O_APPEND {
+		return 0 // don't re-open
 	}
 
 	if flag&syscall.O_APPEND != 0 {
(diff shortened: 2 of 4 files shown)


codefromthecrypt:
whack-a-mole.. different test is failing on gotip

```
=== RUN   TestOffsetWriter_Write/Write
    io_test.go:648: Write failed, n: 0, err: write /tmp/TestOffsetWriter_Write1874476346/001/TestOffsetWriter_Write_Write230495249: I/O error
=== RUN   TestOffsetWriter_Write/Write_Of_Copy_WriteTo
=== NAME  TestOffsetWriter_Write
    io_test.go:633: Write_Of_Copy_WriteTo error. 
        got n: 0, content:  
        expected n: 16, content: 0123456789ABCDEF
=== NAME  TestOffsetWriter_Write/Write_Of_Copy_WriteTo
    testing.go:1490: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

codefromthecrypt:
The other error is about O_APPEND

```
// WriteAt writes len(b) bytes to the File starting at byte offset off.
// It returns the number of bytes written and an error, if any.
// WriteAt returns a non-nil error when n != len(b).
//
// If file was opened with the O_APPEND flag, WriteAt returns an error.
func (f *File) WriteAt(b []byte, off int64) (n int, err error) {
```

codefromthecrypt:
got it, and switched the unit test to fail (so we don't have to wait for stdlib tests)

codefromthecrypt:
next up:

```
=== CONT  TestDoubleCloseError/dir
    os_test.go:2774: second close returned expected error "close /home/runner/sdk/gotip: file already closed"
```

codefromthecrypt:
ps my process from gotip:

```bash
$ GOOS=wasip1 GOARCH=wasm gotip test -v -c -o ./src/os/os.wasm ./src/os
$ cd ./src/os
$ wazero run -hostlogging=filesystem -mount=/:/ -env PWD=$PWD os.wasm -- -test.v -test.run='^TestDoubleCloseError.*$'
```

then I run in my IDE the same args so I can debug it, if the logging wasn't obvious to figure out what happened.

codefromthecrypt:
oops double-close is a red herring. real error was:

```
=== RUN   TestAppend
    os_test.go:2142: writeFile: have "|append" want "new|append"
--- FAIL: TestAppend (0.00s)
```

codefromthecrypt:
finally got it :D

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -205,7 +205,7 @@ func fdFdstatGetFn(_ context.Context, mod api.Module, params []uint64) syscall.E
 		if f.File.IsNonblock() {
 			fdflags |= wasip1.FD_NONBLOCK
 		}
-	} else if f.File.AccessMode() != syscall.O_RDONLY {
```
codefromthecrypt, line 208: so this conflated write with append

internal/sys/fs.go
```diff
@@ -255,6 +255,11 @@ type cachedStat struct {
 	Ino uint64
 }
 
+// IsAppend returns true if the file is open with syscall.O_APPEND.
+func (f *FileEntry) IsAppend() bool {
+	return f.openFlag&syscall.O_APPEND != 0
```
codefromthecrypt, line 260: this is temporary until the file type is sorted out. Ideally we don't need much of a FileEntry thing later

codefromthecrypt (commented):
notes

PR #1459, opened 2023-05-12
memory: avoid integer overflow

ncruces:


internal/wasm/memory_test.go
```diff
@@ -733,6 +733,13 @@ func TestMemoryInstance_Write(t *testing.T) {
 
 	ok = mem.Write(9, buf)
 	require.False(t, ok)
+
+	// Test overflow
+	huge := uint64(math.MaxUint32 + 1 + 4)
```
codefromthecrypt, line 738: I love I didn't have to ask for this test

codefromthecrypt (approved):
thanks a lot!

internal/wasm/memory_test.go
ncruces: Let me see if it goes through in 32-bit.
Does CI check this?
I can pull it on Linux and trying running there.

codefromthecrypt: we don't run tests on 32-bit platforms at the moment, only compilation tests.

internal/wasm/memory_test.go
```diff
@@ -733,6 +733,13 @@ func TestMemoryInstance_Write(t *testing.T) {
...
+	if huge == uint64(int(huge)) { // Only run test on 64-bit platforms
+		ok := mem.Write(4, make([]byte, huge))
```
codefromthecrypt, line 740: actually I'm now wondering if we should make a byte array so big in unit tests :P

internal/wasm/memory_test.go
ncruces: Pass on `GOARCH=386`, fail with `race`.
I'll isolate into its own test, catch the panic and skip the test.

internal/wasm/memory_test.go
codefromthecrypt: should we add another dimension to TestMemoryInstance_HasSize instead?

ncruces: On 64-bit shouldn't be a problem (at least on Linux?): memory isn't committed, just reserved.

But let me try to avoid the issue and create a pretend slice instead.

codefromthecrypt: cool, I think we would end up running this on win, darwin, and linux. One easy enough way is to look at before/after time if any difference, then add a comment like `// don't worry: memory isn't committed, just reserved.`

codefromthecrypt:
ps you can ignore the go wasip1 fail, but they are fixed on master now.

internal/wasm/memory_test.go
```diff
@@ -735,6 +737,23 @@ func TestMemoryInstance_Write(t *testing.T) {
...
+	mem := &MemoryInstance{Buffer: []byte{0, 0, 0, 0, 16, 0, 0, 0}, Min: 1}
+
+	// Test overflow
+	huge := uint64(math.MaxUint32 + 1 + 4)
+	if huge != uint64(int(huge)) {
+		t.Skip("Skipping on 32-bit")
+	}
+
+	buf := []byte{16, 0, 0, 4}
+	header := (*reflect.SliceHeader)(unsafe.Pointer(&buf))
```
codefromthecrypt, line 750: slick!

codefromthecrypt (approved):

PR #1460, opened 2023-05-12
Implement WebAssembly threads proposal with interpreter

anuraaga:
This implements the threads proposal for interpreter

https://github.com/WebAssembly/threads

Unique to interpreter mode is that Go does not provide APIs corresponding to all of the instructions in the proposal - ~~where possible, atomic functions are used, but where not possible, the memory's mutex is used instead~~ so we use a memory-wide mutex to atomicize all operations. The instructions seem to map to common hardware instructions so compiler mode will probably be simpler in that respect (didn't do a deep dive, just cursory check, for example CMPXCHG does return the old value, presumably why the Wasm one also does despite it being uncommon in any standard library SDKs I've seen).

Spectest pass, but spectests only check inputs outputs and are not actually multi-threaded. I also added some hammer tests which I could come up with, I couldn't come up with them for all instructions though, maybe someone will have some cool algorithm idea though.

It should be noted that this does not implement the wasi-threads proposal, which actually spawns threads, but only the WebAssembly threads proposal, which is just atomic operations and block/resume. Unfortunately, LLVM compiles atomics with assumptions from the wasi-threads proposal such as there being one memory instance shared by multiple instantiated modules - this means that in practice, LLVM-compiled programs won't actually be able to take advantage of this until that is addressed.


anuraaga:
Sorry had just ran `make check`, will look at the `make spectest.v2` failure

codefromthecrypt:
Thanks, @anuraaga. One thing that would be nice is to have an integrated code that uses the primitives.. wondering about C code that would compile correctly, then we can drive exports on the go side, which should work atomically?

anuraaga:
zig cc doesn't support threads yet, will look around, perhaps rust does. In the meantime also will switch back to draft until data races are fixed - obviously multithreading is not trivial :D

anuraaga:
Alright should be OK to take a look now, I'll continue researching on the rust or similar integrated test scenario.

https://github.com/tetratelabs/wazero/pull/1460/commits/b38ebccf7051a62ea3b3327225a0e73bfdc88681 was a significant redesign, it's been a while since doing multithreaded stuff so had forgotten can't mix atomic-via-mutex and normal atomic since they would still mix. Luckily the Go race detector reminded me of it. Because many operations can't be implemented with Go, using mutex for all of them seems like the only way to go if I'm not missing any approach.

It means atomic operations with interpreter are dirt slow - but I guess interpreter generally is so hopefully it's ok.

codefromthecrypt:
Adrian's ramble, not review: I don't have cognitive space to understand the Wasm-threads proposal to check the scoping and use of atomics here, but then again there are others who can also. What I would eventually ask for is to have similar tests for func validation wazeroir, etc which at least scaffold in what ops would work. this part is re-used for the compiler and any eventually complete impl will have them in place. Just mentioning in case you end up in a review void a couple days as Edo and Takeshi are out and I'll be out flying to craft-conf early next week. I have another PR that I don't want to leave unfinished by diverting here and sadly have to choose since less time left.

anuraaga:
Thanks, will look through the opcode tests to add more.

For rust, presumably because it uses LLVM, it seems to have the same limitation as wasi-threads, can't call modules concurrently because table/globals access isn't thread-safe.

`expected no error, but was wasm error: indirect call type mismatch`

ncruces:
Presumably, all of these can be implemented with atomics by spinning on CAS? Except CAS itself, which should be a fun challenge. OTOH, mutexing makes the interpreter simpler.

The thing that bothers me with mutexes vs. lock free is that these may be used for "host ⇔ guest" communication, and if we use atomics for the compiler and mutexes for the interpreter, it becomes harder to create a cross platform host.

But I still 👍 for an initial implementation. Any such shenanigans should be in a dedicated, internal `atomics` package.

internal/wasm/memory.go
```diff
@@ -284,3 +292,49 @@ func (m *MemoryInstance) writeUint64Le(offset uint32, v uint64) bool {
 	binary.LittleEndian.PutUint64(m.Buffer[offset:], v)
 	return true
 }
+
+// Wait suspends the caller until the offset is notified by a different agent.
+func (m *MemoryInstance) Wait(offset uint32, timeout int64) (bool, bool) {
```
ncruces, line 297: `(bool, bool)` self documenting names, please?

internal/wasm/memory.go
```diff
@@ -284,3 +292,49 @@ func (m *MemoryInstance) writeUint64Le(offset uint32, v uint64) bool {
...
+
+// Notify wakes up at most count waiters at the given offset.
+func (m *MemoryInstance) Notify(offset uint32, count uint32) uint32 {
+	m.Mux.Lock()
+	defer m.Mux.Unlock()
+
+	res := uint32(0)
+	for num := len(m.waiters[offset]); num > 0 && res < count; num = len(m.waiters[offset]) {
+		w := m.waiters[offset][num-1]
+		m.waiters[offset] = m.waiters[offset][:num-1]
```
ncruces, line 334: This seems to notify in LIFO order.
I'm left wondering about liveness issues?

internal/wasm/memory.go
```diff
@@ -284,3 +292,49 @@ func (m *MemoryInstance) writeUint64Le(offset uint32, v uint64) bool {
...
+
+// Wait suspends the caller until the offset is notified by a different agent.
+func (m *MemoryInstance) Wait(offset uint32, timeout int64) (bool, bool) {
+	m.Mux.Lock()
+
+	if m.waiters == nil {
+		m.waiters = make(map[uint32][]chan struct{})
+	}
+
+	if uint(len(m.waiters[offset])) == math.MaxUint32 {
```
ncruces, line 304: Why `math.MaxUint32` as the limit? I guess I need a comment.
Append will panic way before this on 32-bit, and 4 billion listeners seems way too much…?

internal/engine/interpreter/interpreter.go
```diff
@@ -3896,6 +3896,368 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, m *wasm.ModuleInstance
...
+		case wazeroir.OperationKindAtomicRMW8:
+			val := ce.popValue()
+			offset := ce.popMemoryOffset(op)
+			memoryInst.Mux.Lock()
+			old, ok := memoryInst.ReadByte(offset)
+			if !ok {
+				memoryInst.Mux.Unlock()
+				panic(wasmruntime.ErrRuntimeOutOfBoundsMemoryAccess)
+			}
+			arg := byte(val & 0xFF)
```
ncruces, line 4137: nit: `… & 0xFF` seems unneeded?

ncruces (commented):
This is "above my pay grade" to approve, but I made a pass if this gives other's more confidence.

internal/wasm/memory.go
anuraaga: Thanks for the reminder! Had intended to revisit but forgot, indeed LIFO seems like a bad idea for it (Wasm spec doesn't specify any order though). I switched to `list.List`, which is the same as the Go sempahore implementation

anuraaga (commented):
Thanks @ncruces! Believe I addressed all the comments.

I have also backfilled opcode and other tests, I think I got them all but let me know if I missed any tests and will add them.

anuraaga:
> Presumably, all of these can be implemented with atomics by spinning on CAS? Except CAS itself, which should be a fun challenge. OTOH, mutexing makes the interpreter simpler.

Yeah, unfortunately I think the only pattern (that I can come up with) is something like

```go
for {
  old := atomic.Load
  new := operation(old, arg)
  if CAS(off, old, new) { break }
}
```

An additional load on every single iteration of spinning may be prohibitive, or maybe not. It could be worth an experiment someday in the future indeed.

> if we use atomics for the compiler and mutexes for the interpreter, it becomes harder to create a cross platform host.

@ncruces I don't quite follow this, can you explain more? The mutexes are used to make blocks of code atomic, so they should be semantically equivalent to atomic instructions that correspond to those blocks of code. I'm not sure how this makes a cross platform host harder to achieve so let me know what I'm missing.

internal/wasm/binary/limits.go
```diff
@@ -10,20 +10,20 @@ import (
 // decodeLimitsType returns the `limitsType` (min, max) decoded with the WebAssembly 1.0 (20191205) Binary Format.
 //
 // See https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#limits%E2%91%A6
-func decodeLimitsType(r *bytes.Reader) (min uint32, max *uint32, err error) {
+func decodeLimitsType(r *bytes.Reader) (min uint32, max *uint32, shared bool, err error) {
```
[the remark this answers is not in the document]
ncruces: Thanks! From reading the link it seems only 3 is valid for shared?

_Shared storage requires a maximum size to be specified. In future versions of WebAssembly, shared storage without a maximum size may be allowed._

Should we error out (here or elsewhere, 2 is a valid encoding after all).

ncruces (commented):
Thanks for the updates! This is looking great.

I want to chew on the semaphores a bit longer, but it's a great call to reuse that code.

anuraaga:
> So if guests can use either, depending on compiler vs. interpreter, the host needs to use both (or decide at runtime which is correct).

The guest is webassembly, it is using just atomic instructions like `atomic.memory.load`, nothing else. It is the host, in this case wazero, that compiles it into either mutex or atomic based on compiler or not, which is global to the runtime so affects all modules loaded by it in the same way. So I guess there wouldn't be any inconsistency? 🤔

ncruces:
> The guest is webassembly, it is using just atomic instructions like `atomic.memory.load`, nothing else. It is the host, in this case wazero, that compiles it into either mutex or atomic based on compiler or not, which is global to the runtime so affects all modules loaded by it in the same way. So I guess there wouldn't be any inconsistency? 🤔

In my SQLite wrapper, I change a guest memory location asynchronously from the host to interrupt long running queries. I do it using atomics in the host, the guest uses nothing (because it's lacking these). The race detector doesn't detect this using the compiler (because it can't see into compiled code), and flags it with the interpreter.

With this implementation, if the interpreter uses mutexes, I should change my host to mutex to fix the race. If the compiler uses atomics, my host should continue to use atomics to fix the race. Since I don't know which my users will use, I need to either use both atomics and mutexes, or ask at runtime "is this the interpreter or compiler?" and pick at runtime.

But this is just to say that I feel we should (eventually, this is fine for now!) do these with atomics, even in the interpreter.

experimental/features.go
```diff
@@ -0,0 +1,9 @@
+package experimental
+
+import "github.com/tetratelabs/wazero/api"
+
+// CoreFeaturesThreads enables threads instructions ("threads").
+//
+// Note: The instruction list is too long to enumerate in godoc.
```
codefromthecrypt, line 7: 👍

codefromthecrypt:
```suggestion
// # Notes
//
//   - This is not yet implemented by default, so you will need to use
//      wazero.NewRuntimeConfigInterpreter
//   - The instruction list is too long to enumerate in godoc.
//      See https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md
```

experimental/features.go
```diff
@@ -0,0 +1,9 @@
...
+// See https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md
```
codefromthecrypt, line 8:
```suggestion
```

experimental/features.go
```diff
@@ -0,0 +1,9 @@
...
+const CoreFeaturesThreads = api.CoreFeatureSIMD << 1
```
codefromthecrypt, line 9:
```suggestion
const CoreFeaturesThreads = api.CoreFeatureSIMD << 1
// TODO: Implement the compiler engine
```

codefromthecrypt (commented):
I would still like to see something in experimental (e.g. testdata) that shows this integrated. like features_example.go which has some code that actually can use it. If we can't find or compile anything to use this, I feel it isn't ready to be merged. Maybe someone else can help on this part if you are out of time.

experimental/features_example_test.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		log.Panicln(err)
+	}
+
+	// Channel to synchronize start of goroutines before running.
+	startCh := make(chan struct{})
+	// Channel to synchronize end of goroutines.
+	endCh := make(chan struct{})
+
+	// We start up 8 goroutines and run for 6000 iterations each. The count should reach
```
codefromthecrypt, line 50: is it possible to use hammer.NewHammer instead? It would reduce the sorts of code that launch test goroutines.

anuraaga:
> I would still like to see something in experimental (e.g. testdata) that shows this integrated. like features_example.go which has some code that actually can use it.

I was able to compile a C file with wasi-sdk, unlike Rust just Hello Mutex! doesn't result in a lot of weirdness like call_indirect. Still, things like `__stack_pointer` and `__tls_base` are in globals - the test was passing locally but the race detector is correctly calling out the raciness of updating these wasm globals among the goroutines, so this test doesn't work either.

If someone else can find a compiler that can be accessed from multiple threads with a single module, then that'd be great, otherwise I guess we'll have to close this PR (tried checking in Slack precisely to avoid that but probably didn't communicate it well enough, oh well 😅 )

codefromthecrypt:
@anuraaga what you did is fine on the source. mainly as specs usually drift, an outside test is another way to know if the spec drifted. Thanks for adding the test!

anuraaga:
> what you did is fine on the source. mainly as specs usually drift, an outside test is another way to know if the spec drifted. Thanks for adding the test!

Got it - the test doesn't really work though due to the raciness :( So I wonder if I should make it a main function that can be run manually but not automatically just for manual comparison vs wasi-sdk output?

experimental/features_example_test.go
anuraaga: I thought it'd be better to avoid an internal method in an example, but if it makes sense then I'll switch to it

codefromthecrypt:
> Got it - the test doesn't really work though due to the raciness :( So I wonder if I should make it a main function that can be run manually but not automatically just for manual comparison vs wasi-sdk output?

If things don't work due to raciness, I would just exercise the code without concurrent launching for now and add a TODO: about revisiting the example once it reaches finished state. WDYT?

anuraaga:
Makes sense - I went ahead and kept concurrent launching but added a global lock for them. That allows just deleting lines instead of reworking when things start to work. Can just use a normal invocation instead though if that makes more sense.

anuraaga:
> In my SQLite wrapper, I change a guest memory location asynchronously from the host to interrupt long running queries. I do it using atomics in the host, the guest uses nothing (because it's lacking these). 

@ncruces Thanks for all the explanation, sorry it took so long to finally get it. Indeed, with this PR, atomic operations within the guest are effectively guest-only - even if the host wanted to use mutexes to access, the mutex isn't exposed to even allow it in any way. I wanted to avoid adding too much API in this PR, but I think there will need to be an experimental memory API, likely looking like `AtomicReadUint*Le`, or maybe something else. Then, in the short, or even long term if required, these methods themselves could be implemented differently for compiler vs interpreter, making sure actual callers don't have to worry about the distinction. I suspect this API is best for another PR, but let me know if it's better to get all this together up front to confirm the approach.

codefromthecrypt (approved):
Thanks for the good work and reasoning. I think this is in good shape for someone to attempt to pick up the host-side stuff and/or the engine compiler.

experimental/testdata/pthread.c
```diff
@@ -0,0 +1,18 @@
+#include <pthread.h>
```
codefromthecrypt, line 1: one last request.. if this isn't generated with `make build.examples.zig-cc`, can you add a source comment on when it can (e.g. link to zig or otherwise) and how to compile this meanwhile?

anuraaga: Sure - I had the command documented in features_example_test.go following the pattern in listener_example_test.go, so I went ahead and added the link to zig there. If it's better to have in the source file itself though, I can either move it all here, or copy-paste it all here, let me know if that's better

codefromthecrypt: nope I missed that. all good.

codefromthecrypt:
thanks so much @anuraaga and @ncruces!

PR #1462, opened 2023-05-13
wazero: allow cpu and memory profiling when compiling or running modules

achille-roussel:
This PR adds two new command line options to the `compile` and `run` subcommands: `-cpuprofile` and `-memprofile`.

The intent is to be able to generate CPU and memory profiles of wazero when compiling or running modules. Note that here we are generating profiles for the host (wazero) not the the guest module.

The motivation came from noticing that the compilation of the python version distributed at https://github.com/vmware-labs/webassembly-language-runtimes/releases/tag/python%2F3.11.3%2B20230428-7d1b259 was taking a noticeable amount of time. We can now use the `compile` command to investigate where time is being spent in wazero:
```
$ wazero compile -cpuprofile profile.out python-3.11.3.wasm
```

Performance profiles need to be written when the program exits, so I revisited the implementation of `cmd/wazero` to be able to use `defer` statements: I removed the `exit` callback, which was a reference to `os.Exit` and doesn't run defers; functions of the `main` package now return an exit code instead.

Please take a look and let me know if anything should be changed!

### DIFF
--- a/cmd/wazero/wazero.go
+++ b/cmd/wazero/wazero.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime/pprof"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ncruces (approved):

codefromthecrypt (approved):
Looks useful!

I suppose the only "unit tests" we could make is that files are written and the params don't clobber each-other. If these settings don't work for some of the platforms in our CI, we could skip accordingly.

achille-roussel:
Thanks for the callout for tests @codefromthecrypt, I added tests to validate that the profiles are written where instructed by the command line arguments.

I don't think it's worth going through the pains of making sure that the CPU and memory profiles aren't written to the same files, because even if we get really smart we won't be able to do add fool-proof protections, there is just so many ways to mess things up (e.g. hard or symbolic links or questionable use of overlayfs). `go test` uses the same `cpuprofile` and `memprofile` options and does not do any validation, so I don't think we're diminishing the user experience of developers that would already be familiar with Go.

I'll go ahead and merge but if you feel strongly that we should add more validation I can do this in a follow up 👍

codefromthecrypt:
> I'll go ahead and merge but if you feel strongly that we should add more validation I can do this in a follow up

The way I usually treat these topics is if the maintainer feels strongly about it, they can follow-up. So, it would be on me ;)

PR #1466, opened 2023-05-14
compiler: remove intermediary copy of compiled functions code

achille-roussel:
This PR modifies the compilation of modules to remove the intermediary buffers where the code of each function was copied before being written to the executable memory mapping.

The optimization reduces memory usage since we now only keep a single copy of the module code in memory.

Tested on the compilation of Python:
```
Type: alloc_space
Time: May 14, 2023 at 3:51pm (PDT)
Showing nodes accounting for -51.12MB, 15.15% of 337.43MB total
Dropped 11 nodes (cum <= 1.69MB)
  -55.28MB 16.38% 16.38%   -51.15MB 15.16%  github.com/tetratelabs/wazero.(*runtime).CompileModule
    3.10MB  0.92% 15.46%     3.10MB  0.92%  github.com/tetratelabs/wazero/internal/wazeroir.(*Compiler).emit
    2.60MB  0.77% 14.69%     2.48MB  0.73%  github.com/tetratelabs/wazero/internal/engine/compiler.compileWasmFunction
    2.01MB   0.6% 14.10%     2.01MB   0.6%  github.com/tetratelabs/wazero/internal/engine/compiler.(*runtimeValueLocationStack).push (inline)
   -1.51MB  0.45% 14.55%    -1.51MB  0.45%  debug/dwarf.(*Data).parseAbbrev
   -1.49MB  0.44% 14.99%    -1.49MB  0.44%  github.com/tetratelabs/wazero/internal/engine/compiler.(*runtimeValueLocationStack).cloneFrom (inline)
   -0.64MB  0.19% 15.18%    -0.64MB  0.19%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeCustomSection (inline)
    0.64MB  0.19% 14.99%     6.71MB  1.99%  github.com/tetratelabs/wazero/internal/engine/compiler.(*engine).CompileModule
   -0.54MB  0.16% 15.15%    -0.54MB  0.16%  github.com/tetratelabs/wazero/internal/asm/arm64.(*AssemblerImpl).encodeRelativeBranch
   -0.50MB  0.15% 15.30%    -0.50MB  0.15%  github.com/tetratelabs/wazero/internal/wasm.(*Module).validateFunctionWithMaxStackValues
   -0.50MB  0.15% 15.45%    -2.01MB  0.59%  debug/dwarf.New
    0.50MB  0.15% 15.30%     0.50MB  0.15%  github.com/tetratelabs/wazero/internal/wasm.(*Store).GetFunctionTypeID
    0.50MB  0.15% 15.15%     3.60MB  1.07%  github.com/tetratelabs/wazero/internal/wazeroir.(*Compiler).handleInstruction
```


internal/engine/compiler/engine.go
```diff
@@ -557,39 +575,38 @@ func (e *engine) CompileModule(_ context.Context, module *wasm.Module, listeners
...
-	executable, err := platform.MmapCodeSegment(executableOffset)
-	if err != nil {
-		return err
-	}
 
-	for i, b := range bodies {
-		offset := cm.functions[i].executableOffset
-		copy(executable[offset:], b)
+		copy(executable[executableOffset:], body)
+		executableOffset = (executableOffset + len(body) + 15) &^ 15
```
codefromthecrypt, line 600:
```suggestion
		// Align 16-bytes boundary.
		executableOffset = (executableOffset + len(body) + 15) &^ 15
```

internal/platform/mmap_linux.go
```diff
@@ -0,0 +1,43 @@
...
+func remapCodeSegment(code []byte, size int) ([]byte, error) {
+	p, err := mremap(*(*unsafe.Pointer)(unsafe.Pointer(&code)), len(code), size, __MREMAP_MAYMOVE)
+	if err != nil {
+		return nil, err
+	}
+	return unsafe.Slice((*byte)(p), size), nil
+}
+
+//go:nosplit
+func mremap(oldAddr unsafe.Pointer, oldSize, newSize, flags int) (unsafe.Pointer, error) {
```
codefromthecrypt, line 29: is linux the only one you can use remap in of our supported platforms for compiler (freebsd, darwin, windows)?

internal/engine/compiler/engine.go
```diff
@@ -520,12 +520,29 @@ func (e *engine) CompileModule(_ context.Context, module *wasm.Module, listeners
...
 	}
 
-	bodies := make([][]byte, localFuncs)
 	// As this uses mmap, we need to munmap on the compiled machine code when it's GCed.
 	e.setFinalizer(cm, releaseCompiledModule)
 	ln := len(listeners)
 	cmp := newCompiler()
 	asmNodes := new(asmNodes)
+
+	var executableOffset int
```
codefromthecrypt, line 529: add a summary comment here that the executable is "xxx" and we are doing "xxx"

codefromthecrypt (approved):
I have some suggestions and requests, but looks very promising. Can you run fuzz like in commit.yaml (if you have a machine available to do the full set). Or if that won't catch anything here, lemme know.

internal/platform/mmap_linux.go
achille-roussel: I was only familiar with Linux and assumed BSDs would all look alike but it seems FreeBSD has it https://man.freebsd.org/cgi/man.cgi?query=mremap&manpath=NetBSD+5.0 and obviously the signature is different than Linux 😅

internal/engine/compiler/engine.go
```diff
@@ -557,39 +575,38 @@ func (e *engine) CompileModule(_ context.Context, module *wasm.Module, listeners
 			}
 		}
 
+		functionEndOffset := executableOffset + len(body)
+		if executableLength < functionEndOffset {
+			if executableLength == 0 {
+				executableLength = 65536
+			}
+			for executableLength < functionEndOffset {
+				executableLength *= 2
```
ncruces, line 584: This does not ensure `executableLength >= functionEndOffset`, does it?

Also, we may want to consider if `2` is the correct const always, or if we use another number if the mapping gets really big.

achille-roussel: It's a loop so it will increase the size until it's large enough, isn't it? Or am I missing something?

mathetake:
hah this is looking fantastic

internal/engine/compiler/engine.go
achille-roussel: > Also, we may want to consider if 2 is the correct const always, or if we use another number if the mapping gets really big.

An alternative option could be to align `functionEndOffset` to the next multiple of N KiB (maybe N=64KiB).

ncruces: You're right, I mistook the shorthand `for` for an `if`. 🤦

internal/engine/compiler/engine.go
```diff
@@ -557,39 +575,38 @@ func (e *engine) CompileModule(_ context.Context, module *wasm.Module, listeners
...
+			}
+			b, err := platform.RemapCodeSegment(executable, executableLength)
+			if err != nil {
+				return err
+			}
+			executable = b
+		}
+
 		// The `body` here is the view owned by assembler and will be overridden by the next iteration, so copy the body here.
```
mathetake, line 593: this comment became not applicable anymore, so could you fix?

ncruces (approved):

internal/engine/compiler/engine.go
mathetake, line 584: curious if Go compiler is wise enough to compile this as left shift rather than multiplication as the left shift is much faster in general
```suggestion
				executableLength <<= 1
```

achille-roussel: I believe Go already performs those type of optimizations.

mathetake (approved):

achille-roussel:
I think I addressed all the comments, thank you all for taking the time to review and submit feedback!

@codefromthecrypt it turns out `syscall.SYS_MREMAP` is not defined for FreeBSD; since remapping is an optimization and is emulated via mmap+copy+munmap, I'd suggest we leave as-is until someone cares enough to implement this optimization on FreeBSD.

codefromthecrypt:
> @codefromthecrypt it turns out syscall.SYS_MREMAP is not defined for FreeBSD; since remapping is an optimization and is emulated via mmap+copy+munmap, I'd suggest we leave as-is until someone cares enough to implement this optimization on FreeBSD.

Works for me. These notes are here for the next person if it comes up!

evacchi (approved):
awesome!

achille-roussel:
I completed a fuzz run on linux/amd64, no issues were detected:

```
$ make fuzz fuzz_timeout_seconds=180
...
Done 685736 runs in 181 second(s)
```

PR #1467, opened 2023-05-15
compiler: compress code offsets

achille-roussel:
A major source of memory utilization came from the two `[]uint64` used to map source offsets to the native and wasm code https://github.com/tetratelabs/wazero/blob/main/internal/engine/compiler/engine.go#L1566-L1572

I modified the data structure to use a version of frame-of-reference and delta encoding described in https://lemire.me/blog/2012/02/08/effective-compression-using-frame-of-reference-and-delta-coding/

The data structure works well in this case because the values of code offsets are often close by, and since we don't reorder instructions, the offsets are naturally sorted. Ordering is important here, if the offsets were reordered the deltas between consecutive values would overflow, preventing values from being stored in small integers. If we end up reordering instructions (or if we are already doing it and I missed it, please let me know!) we could consider implementing more advanced delta encodings such as the one used in [Parquet](https://github.com/apache/parquet-format/blob/master/Encodings.md#delta-encoding-delta_binary_packed--5) which handles negative deltas by subtracting the min value to the array, etc... this is work for another day!

The result seems effective, removing another ~45MB of memory allocated when compiling Python:

```
Type: alloc_space
Time: May 14, 2023 at 11:26pm (PDT)
Showing nodes accounting for -47.51MB, 17.69% of 268.61MB total
Dropped 6 nodes (cum <= 1.34MB)
  -43.05MB 16.03% 16.03%   -44.17MB 16.44%  github.com/tetratelabs/wazero/internal/engine/compiler.compileWasmFunction
   -5.55MB  2.07% 18.09%    -5.55MB  2.07%  github.com/tetratelabs/wazero/internal/asm/arm64.(*nodePool).allocNode (inline)
    5.06MB  1.88% 16.21%     5.06MB  1.88%  github.com/tetratelabs/wazero/internal/bitpack.newDeltaArray[...] (inline)
      -3MB  1.12% 17.33%       -3MB  1.12%  debug/dwarf.(*Data).parseAbbrev
      -2MB  0.75% 18.07%       -2MB  0.75%  github.com/tetratelabs/wazero/internal/engine/compiler.(*runtimeValueLocationStack).push (inline)
   -1.27MB  0.47% 18.55%   -43.96MB 16.37%  github.com/tetratelabs/wazero/internal/engine/compiler.(*engine).CompileModule
      -1MB  0.37% 18.92%    -1.50MB  0.56%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeDataSegment
    0.53MB   0.2% 18.72%     0.95MB  0.36%  github.com/tetratelabs/wazero/internal/wazeroir.(*Compiler).handleInstruction
    0.52MB   0.2% 18.53%     0.52MB   0.2%  github.com/tetratelabs/wazero/internal/wasm.(*Module).AllDeclarations
    0.50MB  0.19% 18.34%     0.50MB  0.19%  github.com/tetratelabs/wazero/internal/wasm.(*Module).declaredFunctionIndexes
   -0.50MB  0.19% 18.53%    -0.50MB  0.19%  github.com/tetratelabs/wazero/internal/wazeroir.(*controlFrames).push (inline)
   -0.50MB  0.19% 18.71%    -0.50MB  0.19%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeConstantExpression
    0.50MB  0.19% 18.53%     0.50MB  0.19%  github.com/tetratelabs/wazero/internal/wazeroir.(*funcTypeToIRSignatures).get
    0.50MB  0.19% 18.34%     0.50MB  0.19%  github.com/tetratelabs/wazero/internal/asm/arm64.(*AssemblerImpl).encodeRelativeBranch
    0.49MB  0.18% 18.16%     0.49MB  0.18%  github.com/tetratelabs/wazero/internal/asm/arm64.(*AssemblerImpl).encodeADR
    0.48MB  0.18% 17.98%     0.48MB  0.18%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeCode
    0.42MB  0.16% 17.82%     0.42MB  0.16%  github.com/tetratelabs/wazero/internal/wazeroir.(*Compiler).emit
    0.37MB  0.14% 17.69%     0.37MB  0.14%  bytes.growSlice
```

Please take a look and let me know if you have any questions or feedback!

### DIFF
--- a/internal/bitpack/offset_array.go
+++ b/internal/bitpack/offset_array.go
@@ -1,0 +1,163 @@
+package bitpack
+
+import (
+	"math"
+)
+
+// OffsetArray is an interface representing read-only views of arrays of 64 bits
+// offsets.
+type OffsetArray interface {
+	// Returns the value at index i.
+	//
+	// The method complexity may be anywhere between O(1) and O(N).
+	Index(i int) uint64
+	// Returns the number of offsets in the array.
+	//
+	// The method complexity must be O(1).
+	Len() int
+}
+
+// OffsetArrayLen is a helper function to access the length of an offset array.
+// It is similar to calling Len on the array but handles the special case where
+// the array is nil, in which case it returns zero.
+func OffsetArrayLen(array OffsetArray) int {
+	if array != nil {
+		return array.Len()
+	}
+	return 0
+}
+
+// NewOffsetArray constructs a new array of offsets from the slice of values
+// passed as argument. The slice is not retained, the returned array always
+// holds a copy of the values.
+//
+// The underlying implementation of the offset array applies a compression
+// mechanism derived from Frame-of-Reference and Delta Encoding to minimize
+// the memory footprint of the array. This compression model works best when
+// the input is made of ordered values, otherwise the deltas between values
+// are likely to be too large to benefit from delta encoding.
+//
+// See https://lemire.me/blog/2012/02/08/effective-compression-using-frame-of-reference-and-delta-coding/
+func NewOffsetArray(values []uint64) OffsetArray {
+	if len(values) == 0 {
+		return emptyOffsetArray{}
+	}
+	if len(values) <= smallOffsetArrayCapacity {
+		return newSmallOffsetArray(values)
+	}
+
+	maxDelta := uint64(0)
+	lastValue := values[0]
+	// TODO: the pre-processing we perform here can be optimized using SIMD
+	// instructions.
+	for _, value := range values[1:] {
+		if delta := value - lastValue; delta > maxDelta {
+			maxDelta = delta
+		}
+		lastValue = value
+	}
+
+	switch {
+	case maxDelta > math.MaxUint32:
+		return newOffsetArray(values)
+	case maxDelta > math.MaxUint16:
+		return newDeltaArray[uint32](values)
+	case maxDelta > math.MaxUint8:
+		return newDeltaArray[uint16](values)
+	default:
+		return newDeltaArray[uint8](values)
+	}
+}
+
+type offsetArray struct {
+	values []uint64
+}
+
+func newOffsetArray(values []uint64) *offsetArray {
+	a := &offsetArray{
+		values: make([]uint64, len(values)),
+	}
+	copy(a.values, values)
+	return a
+}
+
+func (a *offsetArray) Index(i int) uint64 {
+	return a.values[i]
+}
+
+func (a *offsetArray) Len() int {
+	return len(a.values)
+}
+
+type emptyOffsetArray struct{}
+
+func (emptyOffsetArray) Index(int) uint64 {
+	panic("index out of bounds")
+}
+
+func (emptyOffsetArray) Len() int {
+	return 0
+}
+
+const smallOffsetArrayCapacity = 15
+
+type smallOffsetArray struct {
+	length int
+	values [smallOffsetArrayCapacity]uint64
+}
+
+func newSmallOffsetArray(values []uint64) *smallOffsetArray {
+	a := &smallOffsetArray{length: len(values)}
+	copy(a.values[:], values)
+	return a
+}
+
+func (a *smallOffsetArray) Index(i int) uint64 {
+	if i < 0 || i >= a.length {
+		panic("index out of bounds")
+	}
+	return a.values[i]
+}
+
+func (a *smallOffsetArray) Len() int {
+	return a.length
+}
+
+type uintType interface {
+	uint8 | uint16 | uint32 | uint64
+}
+
+type deltaArray[T uintType] struct {
+	deltas     []T
+	firstValue uint64
+}
+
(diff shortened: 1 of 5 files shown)


mathetake:
drive-by: how does this compare to packing via leb128?

achille-roussel:
That's a good question.

In the tests I ran, all offsets end up being packed into arrays of 8 bits integers, so in this case, leb128 would do a bit worse because we wouldn't be able to use the full 8 bits for small deltas.

I think leb128 could be interesting if the offsets were out of order because we could encode each delta with a variable number of bytes. This would likely cause the prefix sum to become a bit more expensive because we would have to decode the values; it would also possibly prevent SIMD acceleration.

That being said, it's likely that in some cases leb128 encoding could be the right trade-off, it will be interesting to investigate more!

evacchi (commented):
I love this!!

I wonder if it's worth adding an `Iterate()` method returning a generator that scans linearly each item for cases when you need to do an equivalent to a "`for range`", so that you don't have to do several linear scans for each `i++` -- I can see it only used _maybe_ in `sort.Search()` (and even in that case, maybe you wouldn't use the built-in); but maybe there are other cases?

codefromthecrypt (approved):
I'm cool with this, it is neat and I assume the IP is clean (not copy/pasta from something). I prefer to re-use another encoding we have if the output is the same in benchmarks, also think we should show any runtime difference along with memory. However, no strong preferences, and there are enough folks weighing in!

ps If folks decide the end design is worthwhile a section in one of the RATIONALE.md, nag @achille-roussel, but I won't as the code comments work

ncruces (approved):

achille-roussel:
@codefromthecrypt Besides learning the data structures from the articles I linked, the code itself is authentic and written specifically for wazero, there should be no IP violation 👍 

I added a section in `compiler/RATIONALE.md` about this change, as well as a `TODO` related to optimizing `sort.Search`.

I also did more analysis of the distribution of deltas, and it turned out most were <=15 which creates the opportunity for another specialization of the `OfferArray` type using 4 bits integers to shrink the memory footprint further.

achille-roussel:
@codefromthecrypt 

> I prefer to re-use another encoding we have if the output is the same in benchmarks

There is a measurable difference to switch to leb128, because we lose 1 bit on 1 byte encodings, deltas between 128 and 255 will need two bytes to encode, and we also cannot encode values on less than a byte. The 4 bit integer packing reduces the memory footprint of source offset maps by an extra 25% for Python.

> also think we should show any runtime difference along with memory

Runtime timings are hard to compare because the CPU profiler uses a 100Hz sampling rate, and compilation takes north of 500ms, so there aren't many data points. Using `-count 10` from https://github.com/tetratelabs/wazero/pull/1470 gives a bit more material:
```
$ go tool pprof -text /tmp/profile.base | head
Type: cpu
Time: May 15, 2023 at 11:39am (PDT)
Duration: 5.45s, Total samples = 5.20s (95.36%)
Showing nodes accounting for 4.88s, 93.85% of 5.20s total
Dropped 98 nodes (cum <= 0.03s)
     1.89s 36.35% 36.35%      1.89s 36.35%  runtime.memmove
     1.25s 24.04% 60.38%      1.25s 24.04%  runtime.madvise
     0.17s  3.27% 63.65%      0.33s  6.35%  github.com/tetratelabs/wazero/internal/asm/arm64.(*AssemblerImpl).encodeLoadOrStoreWithConstOffset
     0.17s  3.27% 66.92%      0.30s  5.77%  github.com/tetratelabs/wazero/internal/wasm.(*Module).validateFunctionWithMaxStackValues
```

```
$ go tool pprof -text /tmp/profile.out | head
Type: cpu
Time: May 15, 2023 at 11:40am (PDT)
Duration: 5.52s, Total samples = 5.20s (94.13%)
Showing nodes accounting for 4.79s, 92.12% of 5.20s total
Dropped 105 nodes (cum <= 0.03s)
     3.16s 60.77% 60.77%      3.16s 60.77%  runtime.memmove
     0.40s  7.69% 68.46%      0.40s  7.69%  runtime.madvise
     0.13s  2.50% 70.96%      0.20s  3.85%  github.com/tetratelabs/wazero/internal/asm/arm64.(*AssemblerImpl).encodeLoadOrStoreWithConstOffset
     0.11s  2.12% 73.08%      0.64s 12.31%  github.com/tetratelabs/wazero/internal/asm/arm64.(*AssemblerImpl).Assemble
```

Overall a bit faster than `main`, it looks like there is a transfer cost from `runtime.madvise` to `runtime.memmove`, most of the time is spent in the system. `runtime.madvise` is used internally by the Go memory allocator, while `runtime.memmove` is the implementation of `copy`, and shows up when copying the assembled native code to the execution memory sections. I'm interpreting this as being due to page faults on memory mappings when the kernel needs to associate virtual addresses to physical pages. What is interesting is this seems to become the single point of contention during compilation, I'm going to look into what we can do.

PR #1469, opened 2023-05-15
compiler: optimize struct layout to reduce memory footprint

achille-roussel:
This PR implements a few optimizations detected by https://pkg.go.dev/golang.org/x/tools@v0.9.1/go/analysis/passes/fieldalignment/cmd/fieldalignment

The one I was most interested in was reducing the size of `arm64.nodeImpl` because it represents a significant source of memory allocations since we create so many of those objects.

```
File: wazero
Type: alloc_space
Time: May 15, 2023 at 1:24am (PDT)
Showing nodes accounting for -8.79MB, 2.99% of 294.14MB total
Dropped 7 nodes (cum <= 1.47MB)
   -5.57MB  1.89%  1.89%    -5.57MB  1.89%  github.com/tetratelabs/wazero/internal/asm/arm64.(*nodePool).allocNode (inline)
   -1.97MB  0.67%  2.56%    -5.52MB  1.88%  github.com/tetratelabs/wazero/internal/engine/compiler.compileWasmFunction
   -1.50MB  0.51%  3.07%    -1.50MB  0.51%  debug/dwarf.(*Data).parseAbbrev
    1.50MB  0.51%  2.56%     1.50MB  0.51%  github.com/tetratelabs/wazero/internal/engine/compiler.(*runtimeValueLocationStack).cloneFrom (inline)
    1.29MB  0.44%  2.13%     1.29MB  0.44%  github.com/tetratelabs/wazero/internal/engine/compiler.(*arm64Compiler).label
   -1.09MB  0.37%  2.49%    -1.09MB  0.37%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeCode
   -1.01MB  0.34%  2.84%    -1.01MB  0.34%  github.com/tetratelabs/wazero/internal/engine/compiler.(*runtimeValueLocationStack).push (inline)
   -0.79MB  0.27%  3.11%    -0.79MB  0.27%  bytes.growSlice
    0.78MB  0.26%  2.84%     0.78MB  0.26%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeDataSegment
    0.55MB  0.19%  2.66%     0.55MB  0.19%  github.com/tetratelabs/wazero/internal/asm/arm64.(*AssemblerImpl).encodeRelativeBranch
   -0.51MB  0.17%  2.83%    -0.51MB  0.17%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeElementInitValueVector
   -0.47MB  0.16%  2.99%    -0.47MB  0.16%  github.com/tetratelabs/wazero/internal/wazeroir.(*Compiler).emit
```

Other optimizations in the assembler were automated by `fieldalignment -fix`.

### DIFF
--- a/internal/asm/amd64/impl.go
+++ b/internal/asm/amd64/impl.go
@@ -12,17 +12,17 @@ import (
 
 // nodeImpl implements asm.Node for amd64.
 type nodeImpl struct {
-	staticConst *asm.StaticConst
-
 	// jumpTarget holds the target node in the linked for the jump-kind instruction.
 	jumpTarget *nodeImpl
+
+	// prev and next hold the prev/next node from this node in the assembled linked list.
+	prev, next *nodeImpl
 
 	// forwardJumpOrigins hold all the nodes trying to jump into this node as a
 	// singly linked list. In other words, all the nodes with .jumpTarget == this.
 	forwardJumpOrigins *nodeImpl
 
-	// prev and next hold the prev/next node from this node in the assembled linked list.
-	prev, next *nodeImpl
+	staticConst *asm.StaticConst
 
 	dstConst       asm.ConstantValue
 	offsetInBinary asm.NodeOffsetInBinary
@@ -32,21 +32,17 @@ type nodeImpl struct {
 	// readInstructionAddressBeforeTargetInstruction holds the instruction right before the target of
 	// read instruction address instruction. See asm.assemblerBase.CompileReadInstructionAddress.
 	readInstructionAddressBeforeTargetInstruction asm.Instruction
-	dstMemIndex                                   asm.Register
-	srcMemIndex                                   asm.Register
-	types                                         operandTypes
-	srcMemScale                                   byte
-	dstMemScale                                   byte
-	arg                                           byte
-	dstReg                                        asm.Register
 	flag                                          nodeFlag
+	types                                         operandTypes
+	srcReg, dstReg                                asm.Register
+	srcMemIndex, dstMemIndex                      asm.Register
+	srcMemScale, dstMemScale                      byte
+	arg                                           byte
 
 	// staticConstReferrersAdded true if this node is already added into AssemblerImpl.staticConstReferrers.
 	// Only used when staticConst is not nil. Through re-assembly, we might end up adding multiple times which causes unnecessary
 	// allocations, so we use this flag to do it once.
 	staticConstReferrersAdded bool
-
-	srcReg asm.Register
 }
 
 type nodeFlag byte
--- a/internal/asm/amd64/impl.go
+++ b/internal/asm/amd64/impl.go
@@ -12,35 +12,26 @@ import (
 
 // nodeImpl implements asm.Node for amd64.
 type nodeImpl struct {
-	instruction asm.Instruction
-
-	offsetInBinary asm.NodeOffsetInBinary
-
-	// jumpTarget holds the target node in the linked for the jump-kind instruction.
-	jumpTarget *nodeImpl
-	flag       nodeFlag
-	// prev and next hold the prev/next node from this node in the assembled linked list.
-	prev, next *nodeImpl
-
-	types                    operandTypes
-	srcReg, dstReg           asm.Register
-	srcConst, dstConst       asm.ConstantValue
-	srcMemIndex, dstMemIndex asm.Register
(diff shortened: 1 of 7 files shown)


evacchi (commented):
looks super-promising, make sure you don't lose godoc in the refactoring!

achille-roussel:
Thanks for pointing it out, apparently `fieldalignment -fix` removed the comments when it reodered the fields 🤦

internal/asm/arm64/impl_1_test.go
```diff
@@ -1065,13 +1067,15 @@ func TestAssemblerImpl_EncodeMemoryToVectorRegister(t *testing.T) {
...
 func TestAssemblerImpl_EncodeVectorRegisterToVectorRegister(t *testing.T) {
 	tests := []struct {
-		name               string
-		x1, x2             asm.Register
-		inst               asm.Instruction
-		c                  asm.ConstantValue
-		arr                VectorArrangement
-		srcIndex, dstIndex VectorIndex
-		exp                []byte
+		name     string
```
codefromthecrypt, line 1070: now our tests are lightning fast

codefromthecrypt (approved):
great idea leveraging more analysis tools!

internal/asm/amd64/impl.go
```diff
@@ -12,35 +12,41 @@ import (
...
 
-	// forwardJumpOrigins hold all the nodes trying to jump into this node as a singly linked list. In other words, all the nodes with .jumpTarget == this.
-	forwardJumpOrigins *nodeImpl
 	// staticConstReferrersAdded true if this node is already added into AssemblerImpl.staticConstReferrers.
 	// Only used when staticConst is not nil. Through re-assembly, we might end up adding multiple times which causes unnecessary
 	// allocations, so we use this flag to do it once.
 	staticConstReferrersAdded bool
-	staticConst               *asm.StaticConst
+
+	srcReg asm.Register
```
ncruces, line 49: Odd that it separates `srcReg` from the other `asm.Register`, I guess because it doesn't matter.
It sorts pointers first, then types from larger to smaller.

I suggest (which keeps previous order as much as possible, on the idea that previous order was semantically valid):
```go
type nodeImpl struct {
	// jumpTarget holds the target node in the linked for the jump-kind instruction.
	jumpTarget *nodeImpl

	// prev and next hold the prev/next node from this node in the assembled linked list.
	prev, next *nodeImpl

	// forwardJumpOrigins hold all the nodes trying to jump into this node as a
	// singly linked list. In other words, all the nodes with .jumpTarget == this.
	forwardJumpOrigins *nodeImpl

	staticConst *asm.StaticConst

	offsetInBinary asm.NodeOffsetInBinary
	srcConst       asm.ConstantValue
	dstConst       asm.ConstantValue
	instruction    asm.Instruction

	// readInstructionAddressBeforeTargetInstruction holds the instruction right before the target of
	// read instruction address instruction. See asm.assemblerBase.CompileReadInstructionAddress.
	readInstructionAddressBeforeTargetInstruction asm.Instruction
	flag                                          nodeFlag
	types                                         operandTypes
	srcReg, dstReg                                asm.Register
	srcMemIndex, dstMemIndex                      asm.Register
	srcMemScale, dstMemScale                      byte
	arg                                           byte

	// staticConstReferrersAdded true if this node is already added into AssemblerImpl.staticConstReferrers.
	// Only used when staticConst is not nil. Through re-assembly, we might end up adding multiple times which causes unnecessary
	// allocations, so we use this flag to do it once.
	staticConstReferrersAdded bool
}
```

ncruces (approved):
Great! I meant to do this, guess I'll have to find something else to keep my commit count high. 😆

internal/asm/amd64/impl.go
achille-roussel: Yes, the balance between readability and optimal layout can be difficult to strike sometimes, and `fieldalignment -fix` doesn't do a great job of grouping things that are related to one another.

I made the change you suggested, it's much more readable and likely better for data locality 👍

internal/asm/arm64/impl_1_test.go
achille-roussel: Ahah! That was me learning how to use the tool, I then discovered there is a `-test=false` option to skip test files.

PR #1470, opened 2023-05-15
wazero: add -count option to compile command

achille-roussel:
This option has been useful to run the compilation multiple times in order to get more samples when recording CPU or memory profiles.

I don't have a lot of context on what the intent was for the `compile` subcommand, so maybe this option is out of place, so feel free to let me know if this should be done differently!

### DIFF
--- a/cmd/wazero/wazero.go
+++ b/cmd/wazero/wazero.go
@@ -72,6 +72,10 @@ func doCompile(args []string, stdErr io.Writer) int {
 	var help bool
 	flags.BoolVar(&help, "h", false, "Prints usage.")
 
+	var count int
+	flags.IntVar(&count, "count", 1,
+		"Number of times to perform the compilation. This is useful to benchmark performance of the wazero compiler.")
+
 	var cpuProfile string
 	flags.StringVar(&cpuProfile, "cpuprofile", "",
 		"Enables cpu profiling and writes the profile at the given path.")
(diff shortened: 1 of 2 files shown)


cmd/wazero/wazero.go
```diff
@@ -72,6 +72,10 @@ func doCompile(args []string, stdErr io.Writer) int {
 	var help bool
 	flags.BoolVar(&help, "h", false, "Prints usage.")
 
+	var count int
+	flags.IntVar(&count, "count", 1,
```
codefromthecrypt, line 76: Is there a way to have a flag undocumented? I feel this isn't worth distracting users on -h with. wdyt?

codefromthecrypt (commented):
compile has only 2 use cases
* without cache-dir, make sure it can compile before trying to run it
* with cache-dir, stage the cache so later things run better (think images or GH cache)

cmd/wazero/wazero.go
ncruces: I don't think so. 😞

achille-roussel: We can customize the whole usage message via [`flag.FlatSet.Usage`](https://pkg.go.dev/flag#FlagSet), but I don't know of a way to hide a specific flag.

I think what we could do is expose the option only if the `wazero` version is `"dev"`? So it exists when compiling locally from sources, but not on the release builds.

codefromthecrypt: let's do dev. In fact I think you can also consider if the earlier profiler based stuff would make sense under that, too. I think it probably is, as typical CLI users won't use these.

PR #1471, opened 2023-05-15
platform: fix minor linting issue

evacchi:
return value is not checked in internal/platform/mremap_other.go
for munmapCodeSegment(code)

Probably we forgot to run `make lint` on this one.

### DIFF
--- a/internal/platform/mremap_other.go
+++ b/internal/platform/mremap_other.go
@@ -8,7 +8,10 @@ func remapCodeSegmentAMD64(code []byte, size int) ([]byte, error) {
 		return nil, err
 	}
 	copy(b, code)
-	munmapCodeSegment(code)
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
thanks for the catch!

achille-roussel (commented):
So ignoring the error was actually somewhat intentional, tho maybe not the right approach (and I should have left a comment).

The reason is it's less disruptive to the application to leak the previous block if it could be unmapped than to leak the new block and return an error. 

Realistically, either scenarios sound pretty hard to debug so maybe we should panic instead?

codefromthecrypt:
I'll push a change as it stops the build workflow until we merge something.

codefromthecrypt:
Thanks for the help, folks! `make lint` zero over

PR #1473, opened 2023-05-16
Require max size for shared memory

anuraaga:
So sorry, I completely forgot to address this issue from @ncruces 

https://github.com/tetratelabs/wazero/pull/1460#discussion_r1193557372

### DIFF
--- a/internal/integration_test/engine/threads_test.go
+++ b/internal/integration_test/engine/threads_test.go
@@ -57,6 +57,10 @@ func TestThreadsInterpreter(t *testing.T) {
 }
 
 func incrementGuardedByMutex(t *testing.T, r wazero.Runtime) {
+	P := 8               // max count of goroutines
+	if testing.Short() { // Adjust down if `-test.short`
+		P = 4
+	}
 	tests := []struct {
 		fn string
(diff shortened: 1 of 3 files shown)


codefromthecrypt (approved):

internal/integration_test/engine/threads_test.go
```diff
@@ -57,6 +57,10 @@ func TestThreadsInterpreter(t *testing.T) {
 }
 
 func incrementGuardedByMutex(t *testing.T, r wazero.Runtime) {
+	P := 8               // max count of goroutines
+	if testing.Short() { // Adjust down if `-test.short`
```
anuraaga, line 61: Sorry I should have done this originally but forgot, it is the same pattern as the other hammer tests. I notice CI is run with the `-short` flag

anuraaga:
I've been seeing the windows tests fail with OOM sometimes but not often - just checking if that's been seen before. If not, it must be because of my PR and I'll look into it

codefromthecrypt:
opened https://github.com/tetratelabs/wazero/issues/1476 on the flake. thanks for reporting

PR #1475, opened 2023-05-16
linux: support using huge pages

achille-roussel:
I would like to add support for using huge pages on linux to back memory mappings of code segments. Huge pages are useful to reduce the number of TLB misses when managing large number of memory pages.

Because testing this feature is complex and the benefits may largely depend on workloads, but also because linux distributions may have huge pages disabled by default, I would like to put it behind a feature flag so it can be enabled and disabled at runtime without requiring to recompile wazero. For this, I added a new package in `internal/features` which implements the loading of a comma-separated list feature flags from `WAZEROFEATURES` environment variable. I chose this model because a similar mechanism is employed throughout the Go standard library with the `GODEBUG`. However, I am open to feedback and happy to change it if it doesn't fit.

I considered allowing wazero to more cleverly detect whether it was able to use huge pages but decided against it and in favor of letting the user decide because it seems like there are too many risks of making the wrong choice. Since huge pages are an optimization, I would prefer to have the default behavior favor portability, and give users the ability to extensively test the feature prior to enabling it on production workloads.

For the full documentation on the reasons why using huge pages may be useful, see https://www.kernel.org/doc/Documentation/admin-guide/mm/hugetlbpage.rst


internal/platform/mmap_linux.go
```diff
@@ -0,0 +1,82 @@
...
+			flag: int(bits.TrailingZeros64(n)<<__MAP_HUGE_SHIFT) | __MAP_HUGETLB,
+		})
+	}
+
+	sort.Slice(hugePageConfigs, func(i, j int) bool {
+		return hugePageConfigs[i].size > hugePageConfigs[j].size
+	})
+}
+
+func mmapCodeSegment(size, prot int) ([]byte, error) {
```
evacchi, line 67: could it be worth defining a `var mmapCodeSegmentFunc func(size, prot int) ([]byte, error)` and set it during `init()` so that the entire `for` is skipped when `hugePageConfigs` is empty?

i.e. you'd have (pseudo code):

```
var mmapCodeSegmentFn func(size, prot int) ([]byte, error)
func init() {
  if hugePageEnabled() {
      mmapCodeSegmentFn = mmapCodeSegmentHugePage
  } else {
    mmapCodeSegmentFn = mmapCodeSegmentDefault
  }
}
func mmapCodeSegment(size, prot int) ([]byte, error) {
    return mmapCodeSegmentFn(size, prot)
}
```

obviously there's an hidden pointer indirection there, I don't know if it's worth it all, just an idea...

evacchi: Sure that's fair :)

codefromthecrypt (approved):
I have to be honest, I don't really like mixing config at different abstractions, but your description clearly says this is an alternative to re-compiling (using experimental types).

As long as this isn't something intended to remain ENV once the experiments are over, then all good. I think to make it clearly the case, I might rename "internal/features" to "internal/experimental" so that it is obvious, this is the same concept as experimental, just not exported.

Does that work?

internal/features/features_test.go
```diff
@@ -0,0 +1,43 @@
...
+func TestEnabled(t *testing.T) {
+	require.True(t, features.Enabled("f0"))
+	require.True(t, features.Enabled("f1"))
+	require.True(t, features.Enabled("f2"))
+	require.False(t, features.Enabled("nope"))
+}
+
+func TestAllocsEnabled(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("accessing features allocates memory on windows")
```
codefromthecrypt, line 29: ps I don't fully understand the issue here based on the comment. Why is allocating memory during a test a problem? Is it not de-allocated?

evacchi (approved):

internal/features/features_test.go
achille-roussel: I'll rephrase 👍

achille-roussel:
My intention is to keep this an option in the future. While it is an experiment today, we won't be able to work around the fact that using huge pages may not be desirable in some cases and users should have a way to opt-in. 

I'll send an alternative suggestion for how to achieve this without interpreting an environnement variable.

achille-roussel:
I revisited the implementation of the feature flagging mechanism in the latest update, notably:

- the environment variable is now only used in the wazero CLI and tests, it does not get inherited by programs using wazero as a library
- I opted to keep the feature list as an environment variable in the wazero CLI because it is an interface with a more flexible dependency model than a command line argument; we can drop support of the variable in the future without breaking usage of the CLI nor having to retain a `-features` option which does nothing; but please let me know if you feel strongly otherwise
- I added an `experimental.EnableFeatures` function so programs using wazero can enable features while remaining in charge of defining the user interface for this


If you don't mind, please take another look and let me know if you have more comments!

internal/platform/sync.go
```diff
@@ -7,6 +7,6 @@ import (
 	"syscall"
 )
 
-func sync(f *os.File) syscall.Errno {
+func fsync(f *os.File) syscall.Errno {
```
achille-roussel, line 10: I had to rename this function because it conflicted with `import "sync"`.

internal/platform/mremap_other.go
```diff
@@ -12,21 +12,6 @@ func remapCodeSegmentAMD64(code []byte, size int) ([]byte, error) {
...
-// It is less disruptive to the application to leak the previous block if it
-// could be unmapped than to leak the new block and return an error.
-// Realistically, either scenarios are pretty hard to debug, so we panic. There
-// is less as the dominant runtime.GOOS (linux) doesn't use this anyway.
-func mustMunmapCodeSegment(code []byte) {
-	if err := munmapCodeSegment(code); err != nil {
-		panic(err)
-	}
-}
-
```
achille-roussel, line 29: Moved to `internal/platform/platform.go` because it is used in `mremap_linux.go` now when huge pages are enabled. I also edited out the last line of the godoc.

ncruces (approved):

codefromthecrypt:
> My intention is to keep this an option in the future. While it is an experiment today, we won't be able to work around the fact that using huge pages may not be desirable in some cases and users should have a way to opt-in.

I think something like this can be done, but should be an option. For example, someone can compile once to have wazero read settings from ENV variables, where by default there's no weird global config store.

I don't like starting down this path as it quickly becomes spring, but anyway, as long as env is opt-in, it can be contained. Meanwhile, we're still experimental so don't have to debate here.

RATIONALE.md
```diff
@@ -1105,3 +1105,21 @@ In lieu of formal documentation, we infer this pattern works from other sources
 See https://github.com/golang/go/blob/go1.20/src/sync/waitgroup.go#L64
 See https://github.com/golang/go/issues/5045#issuecomment-252730563
 See https://www.youtube.com/watch?v=VmrEG-3bWyM
+
+## Advanced Wazero Features
+
+### Why enable "hugepages" globally instead of per wazero.Runtime instance?
```
codefromthecrypt, line 1111: I don't really like this here because it is experimental. I feel like this is tail wagging the dog into us normalizing ENV tricks as a normal case. Can you please move this note to experimental because the feature is still experimental?

codefromthecrypt (commented):
I'm not going to click approve because I don't like the env secret setting way this is going. wazero took a lot of work to avoid global state, for example one reason we re-did the assembler was to make it safe and unsurprising to have multiple differently configured runtimes acting at the same time. I'll leave to others to click approve.

Sorry, but my years in tech really have bruised me over libraries starting to poke in ENV escapes and whatnot routinely, it is no longer controllable by an independent config layer. I can't ever really get behind this personally, but I don't have to as this is a team project.

achille-roussel:
Thanks everyone for the review and feedback!

@codefromthecrypt let's try to get this into a shape where we all feel good about the long-term maintenance of the feature! I definitely value your experience and opinion a lot, so if you think this is the wrong move it's a good signal that we should try something different.

The way I see it is there is a fact that we can't work around, using huge pages has an effect on contention of the TLB, which cannot be isolated to a single `wazero.Runtime` instance since it is going to be managed by the kernel, the process is the only unit of isolation that we have. Allowing the feature to be configured per runtime would be misleading. The reason why we make it configurable is not because users can benefit from having it enabled on some runtime instances but not others, but because it may not be desirable to default to using huge pages. Now finding a way to express this form of per-process configuration can be tricky since wazero has so far successfully implemented isolation at the runtime or module scope.

Here are a few suggestions I can make:

- We could remove the ability to enable huge pages on the wazero CLI, making it possible to enable it only for programs using wazero as a library via `experimental.EnableFeatures`. While this would address the use case I have that originally motivated this change, it feels like the wrong move not to have a way to leverage the feature in vanilla wazero.

- We could enable the feature by default; the current implementation falls back to `mmap(2)` without huge pages on error, so it is somewhat flexible and should not break applications running on systems where huge pages are not enabled. The worst risk is seeing degraded performance.

- We could configure the feature via a command line argument, calling `experimental.EnableFeatures` from the wazero CLI. This would effectively remove the configuration via an environment variable. Since the intent is to use this in conjecture with `-cpuprofile`, enabling the command only on `dev` version could be a way to not leak into the user intreface.

- We could decide that the trade-off of introducing global configuration is not worth it, and come up with a way to enable huge pages per wazero runtime. We should likely document the limitations we are aware of in this case, so users don't assume that they will be able to fine-tune each runtime instance.

Each approach makes different trade-offs, I don't think I can come up with the right model on my own, and happy to consider other suggestions as well.

I hope this context is useful, I'll sleep on it to reflect but please chime in if you have questions or comments!

codefromthecrypt:
If I understand correctly, you feel the only valid way to enable this setting is by ENV variable, because of its side effects. I don't think we should change the way we configure wazero based on the side effects of each switch. While true certain things have side effects (such as mounting root writeable!), my suggestion is to separate policy and advice from mechanics. We expose configuration and it is up to the user to decide to use env, flags, whatever means they decide to enable or disable things, no sneaky backdoors they can't undo or prevent.

For example, someone using a CLI tool like cobra etc, they should be able to document with those tools how ENV variables configure things. I wouldn't say.. well mounting root writeable is dangerous, so I think you should do that with ENV ;) I hope the analogy helps explain why I'm troubled about starting to dictate how properties are configured at the process level.

codefromthecrypt:
on the choices btw, here's my 2p.

I don't think these choices are exclusive and that they talk about the CLI makes me think they are really about the CLI and how it is configured, not the wazero library in general!

> We could enable the feature by default

I prefer this until an end user says we shouldn't. I'm not an end user.

More settings are more problems, as this discussion has proven ;)

> We could decide that the trade-off of introducing global configuration is not worth it, and come up with a way to enable huge pages per wazero runtime

We have RuntimeConfig exactly for this purpose, it is what it was made for, and how I would suggest enabling the below.


> We could remove the ability to enable huge pages on the wazero CLI
> We could configure the feature via a command line argument

These two are very specific, about the CLI. What troubles me is making a general abstraction about env (e.g. in another package) instead of addressing things like experiments or otherwise in the CLI. I would react very differently, it this env logic was just a function in the CLI's package.

What we do scoped and only scoped to the CLI is extremely flexible. What is triggering to me is making a generic system to subvert RuntimeConfig via ENV as opposed to allow people (including our CLI) to set properties.

It is maybe this point that I've created a misunderstanding. I think of our CLI as basically a consumer of the library. We can and have hacked it to use experimental stuff. However, its configuration approach should live in its package, and not bleed into the main codebase. the CLI is only in our repo for convenience, and anybody should be able to configure anything it can however they like.

hope this ramble helps!

codefromthecrypt:
ps here's how I would have expected this to be configured, if saying out loud helps.

Right now (experimental)

```go
ctx = experimental.WithHugePages(ctx)
// or
ctx = experimental.WithRuntimeFeatures(ctx, experimental.HUGE_PAGES)
r := wazero.NewRuntimeWithConfig(ctx, config)
defer r.Close(ctx)

// blah blah
```

Later (no longer experimental)

```go
config := wazero.NewRuntimeConfig().
                            WithHugePages(true)
// or
config := wazero.NewRuntimeConfig().
                            WithFeatures(RuntimeFeatures.HUGE_PAGES)
r := wazero.NewRuntimeWithConfig(ctx, config)
defer r.Close(ctx)

// blah blah
```

And the CLI would decide whether to use flags or ad-hoc env depending on if something is experimental or not, just like we enable other stuff such as experimental logging.

achille-roussel:
I learned a bunch in the discussions and implementation of this change, thanks a lot for the discussions and reviews!

internal/platform/mremap_linux.go
```diff
@@ -1,43 +0,0 @@
...
-		uintptr(newSize),
-		uintptr(flags),
-		uintptr(0),
-		uintptr(0),
-	)
-	if err != 0 {
-		return nil, err
-	}
-	return unsafe.Pointer(p), nil
-}
```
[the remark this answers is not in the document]
codefromthecrypt: Thanks for all the comments here, @achille-roussel. It helps to be able to start with these thoughts should we ever need to revisit.

PR #1477, opened 2023-05-16
extracts FS interfaces into fsapi package and consolidates impls

codefromthecrypt:
Before, we had FS impl in the sysfs package and most of File in the platform package. This moves the base types into a new fsapi package, to help with dependency cycles.

This also consolidates file logic from the platform package into the sysfs package. Further refactorings will be simpler after this phase.


codefromthecrypt:
@evacchi can you do me a favor and take a look and try to resolve the `make check` thing? I am not able to reproduce it and need to go board ;)

evacchi:
it seems reproducible on a Linux host

internal/sysfs/select_windows.go
```diff
@@ -14,6 +16,8 @@ const wasiFdStdin = 0
 // pollInterval is the interval between each calls to peekNamedPipe in pollNamedPipe
 const pollInterval = 100 * time.Millisecond
 
+var kernel32 = syscall.NewLazyDLL("kernel32.dll")
```
evacchi, line 19: no big deal (better for clarity), but this may mean that `loadlibrary()` and the proc lookup may be invoked more than once for `kernel32.dll` (but on a low-level it may be cached anyway?)

codefromthecrypt: not sure, but I like your train of thought!

codefromthecrypt:
thanks for the help!

PR #1481, opened 2023-05-18
compiler: zero-copy code assembly

achille-roussel:
This PR modifies the assembler and compiler to remove the copy from the assembler's internal buffer to the memory-mapped code segments.

To do this, I introduced two types in `internal/asm`: `CodeSegment` and `Buffer`. The former represents the code segments generated by the compiler, and the latter is similar to `bytes.Buffer` but is backed by a `CodeSegment` to support assembling instructions directly into the memory-mapped regions.

The intent is to further reduce the memory footprint of the compiler by dropping the need for the intermediary buffer but also to improve the compute time by avoiding a copy and creating opportunities to write instructions more efficiently.

Regarding the second point, I added two methods on the `Buffer` type to improve the assembler throughput.

On ARM64, all instructions are 4 bytes long, so instead of using a `Write([]byte)` method (which always makes a call to `runtime.memmove`), I added a `Write4Bytes` methods which combines all four bytes into a 32 bits integer and writes it to the underlying buffer. 

On AMD64, I revisited the implementation to support writing directly to the output buffer via calls to Go's `append` builtin. This allows the compiler to generate much better code, almost always inlining the writes.

## Benchmarks

The first round of benchmarks I am recording here comes from the internal integration tests. In those benchmarks, we see observe a measurable drop in memory utilization (5-14%) and a 10% lower compute footprint on amd64. 
```
$ go test -v -run '^$' -bench BenchmarkCompilation -benchmem -count 6 ./internal/integration_test/bench/
```
```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.60GHz
                                 │ /tmp/bench.0 │            /tmp/bench.1            │
                                 │    sec/op    │   sec/op     vs base               │
Compilation/with_extern_cache       317.0µ ± 3%   316.7µ ± 2%        ~ (p=0.937 n=6)
Compilation/without_extern_cache    5.659m ± 2%   5.056m ± 2%  -10.65% (p=0.002 n=6)
Compilation/interpreter             1.257m ± 0%   1.247m ± 1%   -0.80% (p=0.002 n=6)
geomean                             1.311m        1.259m        -3.97%

                                 │ /tmp/bench.0 │            /tmp/bench.1            │
Compilation/with_extern_cache      48.96Ki ± 0%   48.94Ki ± 0%  -0.04% (p=0.002 n=6)
Compilation/without_extern_cache   629.9Ki ± 0%   597.9Ki ± 0%  -5.08% (p=0.002 n=6)
Compilation/interpreter            759.0Ki ± 0%   759.0Ki ± 0%       ~ (p=0.327 n=6)
geomean                            286.0Ki        281.1Ki       -1.74%

                                 │ /tmp/bench.0 │            /tmp/bench.1             │
                                 │  allocs/op   │  allocs/op   vs base                │
Compilation/with_extern_cache        420.0 ± 0%    420.0 ± 0%       ~ (p=1.000 n=6) ¹
Compilation/without_extern_cache    1.083k ± 0%   1.073k ± 0%  -0.88% (p=0.002 n=6)
Compilation/interpreter              627.0 ± 0%    626.5 ± 0%       ~ (p=0.545 n=6)
geomean                              658.2         656.1       -0.32%
¹ all samples are equal
```
```
goos: darwin
goarch: arm64
                                 │ /tmp/bench.0 │           /tmp/bench.1            │
                                 │    sec/op    │   sec/op     vs base              │
Compilation/with_extern_cache       148.4µ ± 2%   149.6µ ± 3%       ~ (p=0.818 n=6)
Compilation/without_extern_cache    1.902m ± 6%   1.851m ± 5%       ~ (p=0.093 n=6)
Compilation/interpreter             663.7µ ± 1%   658.0µ ± 1%  -0.86% (p=0.015 n=6)
geomean                             572.3µ        566.9µ       -0.93%

                                 │ /tmp/bench.0 │            /tmp/bench.1             │
Compilation/with_extern_cache      48.81Ki ± 0%   48.80Ki ± 0%        ~ (p=0.374 n=6)
Compilation/without_extern_cache   655.5Ki ± 0%   565.7Ki ± 0%  -13.71% (p=0.002 n=6)
Compilation/interpreter            758.9Ki ± 0%   758.9Ki ± 0%        ~ (p=0.548 n=6)
geomean                            289.6Ki        275.7Ki        -4.80%

                                 │ /tmp/bench.0 │            /tmp/bench.1            │
                                 │  allocs/op   │ allocs/op   vs base                │
Compilation/with_extern_cache        420.0 ± 0%   420.0 ± 0%       ~ (p=1.000 n=6) ¹
Compilation/without_extern_cache     938.0 ± 0%   934.0 ± 0%  -0.43% (p=0.002 n=6)
Compilation/interpreter              626.0 ± 0%   626.0 ± 0%       ~ (p=1.000 n=6) ¹
geomean                              627.1        626.2       -0.14%
¹ all samples are equal
```

The second set of benchmarks here show the differences on the compilation of Python, which I usually run this way:
```sh
# run a few passes of compilation to measure both the baseline and the change
$ time ./wazero compile -count 5 -cpuprofile /tmp/profile.out ~/wasm/bin/python-3.11.3.wasm
...
# analyse with pprof
$ go tool pprof -text -base /tmp/profile.base /tmp/profile.out
```

Overall, the most interesting gains here are a 7% drop in compilation time on amd64, and a few MiB shaved off by removing the intermediary buffer. We're not seeing similar improvements on CPU usage on arm64 tho I'm wondering if this is due to Darwin not dealing with page faults on memory mappings as well as Linux, it seems a lot more time is spent writing to those memory segments (so it might be the bottleneck).

**linux/amd64**
```
File: wazero
Type: cpu
Time: May 18, 2023 at 12:14am (PDT)
Duration: 15.95s, Total samples = 15.72s (98.58%)
Showing nodes accounting for -1.10s, 7.00% of 15.72s total
Dropped 10 nodes (cum <= 0.08s)
     0.79s  5.03%  5.03%     -0.64s  4.07%  github.com/tetratelabs/wazero/internal/asm/amd64.(*AssemblerImpl).encode
    -0.58s  3.69%  1.34%     -1.17s  7.44%  github.com/tetratelabs/wazero/internal/asm/amd64.(*AssemblerImpl).maybeNOPPadding
    -0.56s  3.56%  2.23%     -0.85s  5.41%  bytes.(*Buffer).WriteByte
```
```
File: wazero
Type: alloc_space
Time: May 18, 2023 at 12:13am (PDT)
Showing nodes accounting for -15.31MB, 1.67% of 917.83MB total
Dropped 3 nodes (cum <= 4.59MB)
  -12.04MB  1.31%  1.31%   -12.04MB  1.31%  github.com/tetratelabs/wazero/internal/engine/compiler.(*runtimeValueLocationStack).push (inline)
   -9.80MB  1.07%  2.38%    -9.80MB  1.07%  bytes.growSlice
    5.73MB  0.62%  1.76%     6.73MB  0.73%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeDataSegment
   -5.55MB  0.61%  2.36%    -5.55MB  0.61%  github.com/tetratelabs/wazero/internal/asm/amd64.(*nodePool).allocNode (inline)
```

**darwin/arm64**
```
Type: cpu
Time: May 18, 2023 at 12:03am (PDT)
Duration: 5.69s, Total samples = 5.31s (93.39%)
Showing nodes accounting for -0.13s, 2.45% of 5.31s total
Dropped 8 nodes (cum <= 0.03s)
     1.61s 30.32% 30.32%      1.61s 30.32%  encoding/binary.littleEndian.PutUint32 (inline)
    -1.44s 27.12%  3.20%     -1.44s 27.12%  runtime.memmove
    -0.06s  1.13%  2.07%     -0.16s  3.01%  bytes.(*Buffer).Write
    -0.06s  1.13%  0.94%      0.60s 11.30%  github.com/tetratelabs/wazero/internal/asm/arm64.(*AssemblerImpl).encodeLoadOrStoreWithConstOffset
```
```
Type: alloc_space
Time: May 18, 2023 at 12:06am (PDT)
Showing nodes accounting for -45.46MB, 5.12% of 887.49MB total
Dropped 3 nodes (cum <= 4.44MB)
  -22.27MB  2.51%  2.51%   -22.27MB  2.51%  bytes.growSlice
  -16.13MB  1.82%  4.33%   -16.13MB  1.82%  github.com/tetratelabs/wazero/internal/asm/arm64.(*nodePool).allocNode (inline)
   -5.50MB  0.62%  4.95%    -5.50MB  0.62%  debug/dwarf.(*Data).parseAbbrev
    4.95MB  0.56%  4.39%     4.95MB  0.56%  github.com/tetratelabs/wazero/internal/bitpack.newDeltaArray[...] (inline)
   -4.03MB  0.45%  4.84%    -4.03MB  0.45%  github.com/tetratelabs/wazero/internal/engine/compiler.(*runtimeValueLocationStack).push (inline)
```

Finally, I ran `make fuzz fuzz_timeout_seconds=180` to increase confidence in the change since it is touching a lot of core pieces of the compiler:
```
Done 358310 runs in 181 second(s)
```

## Comments

Overall, this is a lot of code change for a ~10% improvement, which is likely a sign that we are entering the land of diminishing returns in this direction. I wanted to measure how much the removal of the intermediary assembler buffer would impact, so now we know! But I've been too close to the issue to emit a fair judgment on whether the gains are worth the change.

Feel free to chime in with any comments or feedback!

### DIFF
--- a/internal/asm/amd64/impl.go
+++ b/internal/asm/amd64/impl.go
@@ -583,7 +583,7 @@ func (a *AssemblerImpl) encodeOnNextJmpNOPPAdding(buf asm.Buffer, n *nodeImpl) e
 	return a.encodeNOPPadding(buf, instructionLen)
 }
 
-// maybeNOPPadding maybe appends NOP instructions before the node `n`.
+// encodeNOPPadding maybe appends NOP instructions before the node `n`.
 // This is necessary to avoid Intel's jump erratum:
 // https://www.intel.com/content/dam/support/us/en/documents/processors/mitigations-jump-conditional-code-erratum.pdf
 func (a *AssemblerImpl) encodeNOPPadding(buf asm.Buffer, instructionLen int32) error {
--- a/internal/asm/amd64/impl.go
+++ b/internal/asm/amd64/impl.go
@@ -923,6 +923,19 @@ func errorEncodingUnsupported(n *nodeImpl) error {
 }
 
 func (a *AssemblerImpl) encodeNoneToNone(buf asm.Buffer, n *nodeImpl) (err error) {
+	// Throughout the encoding methods, we use this pair of base offset and
+	// code buffer to write instructions.
+	//
+	// The code buffer is allocated at the end of the current buffer to a size
+	// large enough to hold all the bytes that may be written by the method.
+	//
+	// We use Go's append builtin to write to the buffer because it allows the
+	// compiler to generate much better code than if we made calls to write
+	// methods to mutate an encapsulated byte slice.
+	//
+	// At the end of the method, we truncate the buffer size back to the base
+	// plus the length of the code buffer so the end of the buffer points right
+	// after the last byte that was written.
 	base := buf.Len()
 	code := buf.Append(4)[:0]
 
--- a/internal/asm/amd64/impl.go
+++ b/internal/asm/amd64/impl.go
@@ -688,7 +688,7 @@ func (a *AssemblerImpl) padNOP(buf asm.Buffer, num int) {
 		if singleNopNum > len(nopOpcodes) {
 			singleNopNum = len(nopOpcodes)
 		}
-		buf.Write(nopOpcodes[singleNopNum-1][:singleNopNum])
+		buf.AppendBytes(nopOpcodes[singleNopNum-1][:singleNopNum])
 		num -= singleNopNum
 	}
 }
--- a/internal/asm/amd64/impl.go
+++ b/internal/asm/amd64/impl.go
@@ -481,31 +481,46 @@ func (a *AssemblerImpl) initializeNodesForEncoding() {
 	}
 }
 
-func (a *AssemblerImpl) encode(buf asm.Buffer) (err error) {
+func (a *AssemblerImpl) encode(buf asm.Buffer) error {
 	for n := a.root; n != nil; n = n.next {
 		// If an instruction needs NOP padding, we do so before encoding it.
+		//
+		// This is necessary to avoid Intel's jump erratum; see in Section 2.1
+		// in for when we have to pad NOP:
 		// https://www.intel.com/content/dam/support/us/en/documents/processors/mitigations-jump-conditional-code-erratum.pdf
-		if err = a.maybeNOPPadding(buf, n); err != nil {
-			return
+		//
+		// This logic used to be implemented in a function called maybeNOPPadding,
+		// but the complexity of the logic made it impossible for the compiler to
+		// inline. Since this function is on a hot code path, we inlined the
+		// initial checks to skip the function call when instructions do not need
+		// NOP padding.
+		switch info := nopPaddingInfo[n.instruction]; {
+		case info.jmp:
+			if err := a.encodeJmpNOPPadding(buf, n); err != nil {
+				return err
+			}
+		case info.onNextJmp:
+			if err := a.encodeOnNextJmpNOPPAdding(buf, n); err != nil {
+				return err
+			}
 		}
 
 		// After the padding, we can finalize the offset of this instruction in the binary.
 		n.offsetInBinary = uint64(buf.Len())
 
-		if err = a.encodeNode(buf, n); err != nil {
-			return
+		if err := a.encodeNode(buf, n); err != nil {
+			return err
 		}
 
 		if n.forwardJumpOrigins != nil {
-			if err = a.resolveForwardRelativeJumps(buf, n); err != nil {
-				err = fmt.Errorf("invalid relative forward jumps: %w", err)
-				break
+			if err := a.resolveForwardRelativeJumps(buf, n); err != nil {
+				return fmt.Errorf("invalid relative forward jumps: %w", err)
 			}
 		}
 
 		a.maybeFlushConstants(buf, n.next == nil)
 	}
-	return
+	return nil
 }
 
 var nopPaddingInfo = [instructionEnd]struct {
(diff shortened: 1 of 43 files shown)


internal/engine/compiler/engine.go
```diff
@@ -538,43 +531,40 @@ func (e *engine) CompileModule(_ context.Context, module *wasm.Module, listeners
 	asmNodes := new(asmNodes)
 	offsets := new(offsets)
 
-	// The executable code is allocated in memory mappings of executableLength,
+	// The executable code is allocated in memory mappings of held by executable,
```
codefromthecrypt, line 534: this reads awkwardly

internal/asm/buffer.go
```diff
@@ -0,0 +1,188 @@
...
+	b, err := platform.MmapCodeSegment(size)
+	if err != nil {
+		return err
+	}
+	seg.code = b
+	seg.size = 0
+	return nil
+}
+
+func (seg *CodeSegment) Unmap() error {
```
codefromthecrypt, line 46: If Map is called multiple times, I might suggest renaming this to Close instead.

internal/asm/buffer.go
```diff
@@ -0,0 +1,188 @@
...
+//
+// To construct code segments, the program must call Next to obtain a buffer
+// view capable of writing data at the end of the segment. Next must be called
+// before generating the code of a function because it aligns the next write on
+// 16 bytes.
+//
+// Instances of CodeSegment hold references to memory which is NOT managed by
+// the garbage collector and therefore must be released *manually* by calling
+// their Unmap method to prevent memory leaks.
+type CodeSegment struct {
```
codefromthecrypt, line 24: should this file have independent tests?

internal/asm/amd64/impl.go
```diff
@@ -549,55 +556,52 @@ var nopPaddingInfo = [instructionEnd]struct {
...
+
+func (a *AssemblerImpl) encodeOnNextJmpNOPPAdding(buf asm.Buffer, n *nodeImpl) error {
+	instructionLen, err := a.fusedInstructionLength(buf, n)
+	if err != nil {
+		return err
 	}
+	return a.encodeNOPPadding(buf, instructionLen)
+}
 
+// maybeNOPPadding maybe appends NOP instructions before the node `n`.
```
codefromthecrypt, line 586:
```suggestion
// encodeNOPPadding maybe appends NOP instructions before the node `n`.
```

codefromthecrypt (approved):
Thanks. I think that any change like this would affect a lot of files, so wouldn't necessarily suggest based on LOC we are at diminishing returns, yet. That said, appreciate for sure enduring them!

codefromthecrypt:
ps thanks for pre-empting all my rigor requests by doing them all and proving it in the PR description!

internal/asm/buffer.go
achille-roussel: I chose `Unmap` because the value may still be used after unmapping it, I felt like `Close` implied that the value wouldn't be usable anymore (e.g. like closing a file or a stream).

Programs also tend to conventionally ignore errors returned by `Close() error`, I wanted to trigger the reader's attention around how they should handle errors returned when unmapping code segments.

I'll enhance the documentation but let me know if you feel like naming it `Close` has upsides that I'm missing out on.

internal/asm/buffer.go
```diff
@@ -0,0 +1,220 @@
...
+func (seg *CodeSegment) Bytes() []byte {
+	return seg.code
+}
+
+// Next returns a buffer pointed at the end of the code segment to support
+// writing more code instructions to it.
+//
+// Buffers are passed by value, but they hold a reference to the code segment
+// that they were created from.
+func (seg *CodeSegment) Next() Buffer {
```
evacchi, line 110: this method name (`Next`) was pretty opaque to me until I had read the comment on `CodeSegment`. Is this a common name, or may I suggest to think of a better name? I can't really think of a good one :D maybe `StartSection`? `BeginAligned`? `NextAligned`?

evacchi (commented):
leaving a few comments!

internal/asm/arm64/impl.go
```diff
@@ -765,18 +764,19 @@ func errorEncodingUnsupported(n *nodeImpl) error {
...
 	switch n.instruction {
 	case UDF:
-		a.buf.Write([]byte{0, 0, 0, 0})
+		buf.Write4Bytes(0, 0, 0, 0)
+		return nil
 	case NOP:
+		return nil
 	default:
-		err = errorEncodingUnsupported(n)
+		return errorEncodingUnsupported(n)
```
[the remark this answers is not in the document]
achille-roussel: Modestly intended, as I restructured things a few changes of this kind made it through.

internal/asm/buffer.go
achille-roussel: See the name change in https://github.com/tetratelabs/wazero/pull/1481/commits/202b640c728e8576dc95f1cd261784baa8344f03

codefromthecrypt:
Thanks for the continued good work!

PR #1482, opened 2023-05-18
wasi: Move ReadDir out of the FileEntry structure.

evacchi:
Decouple file fd from ReadDir so that it is possible to return
multiple ReadDir structs for a given fd (one per call to ReadDir).

Future work necessary to support e.g. directory-entry-stream
see https://github.com/WebAssembly/wasi-filesystem/blob/main/wit/types.wit#L787

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -836,7 +836,7 @@ func fdReaddirFn(_ context.Context, mod api.Module, params []uint64) syscall.Err
 	}
 	rd := f.File
 	// Discard the bool value because we validated the fd already.
-	dir, _ := fsc.LookupReadDir(fd)
+	dir, _ := fsc.LookupReaddir(fd)
 
 	if cookie == 0 && dir.CountRead > 0 {
 		// This means that there was a previous call to the dir, but cookie is reset.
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
So here's how I was thinking the refactoring would eventually occur, the order isn't too important

ReadDir is renamed to Readdir ;)

Readdir gets managed independent of file (this PR)

Logic that populates Readdir fields is moved to functions of Readdir, the fields are unexported for some sort of iterator API. Temporarily, these functions need to pass an fsapi.File which is used as needed to implement rewind, etc. (e.g. via File.Seek)

An interface is extracted from sys.Readdir: fsapi.Readdir package. This becomes the result of File.Readdir. the File parameter is removed from functions on the interface.

The File.Readdir implementation is made for both fsFile and osFile. e.g sysfs.osReaddir and sysfs.fsReaddir. When File.Readdir, returns fsapi.Readdir, it populates a field with the calling file, and this field is reset on rewind for the fsFile impl (which doesn't support Seek)

This is later extended to prove File.Readdir can be called multiple times and the results don't interfere with eachother, but if called only once, it re-uses the FD of the current file vs re-opening one always.

I hope these notes help!

evacchi:
thanks for the detailed writeup, I am renaming ReadDir => Readdir and ReadDirTable => ReaddirTable etc in this PR

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -870,9 +870,7 @@ func fdReaddirFn(_ context.Context, mod api.Module, params []uint64) syscall.Err
 
 	// Add entries for dot and dot-dot as wasi-testsuite requires them.
 	if cookie == 0 && dirents == nil {
-		if f, ok := fsc.LookupFile(fd); !ok {
```
evacchi, line 872: this is unnecessary because we already set `f` from the corresponding `fd` earlier:

```
	f, errno := openedDir(fsc, fd)
```

codefromthecrypt (approved):
Thanks for starting on this

PR #1483, opened 2023-05-19
wasi: move Readdir logics from fdReaddirFn to Readdir methods

evacchi:
As suggested [in this comment](https://github.com/tetratelabs/wazero/pull/1482#issuecomment-1554008113), the logic that populates the fields in `Readdir` is moved to methods of the `Readdir` struct. We are not exporting an interface yet.

- The fields are no longer exported. 
- There are new fields that are necessary to maintain the position of the cursor, and `func` fields to avoid embedding a `FileEntry`.
- We export 2 constructors: one takes a `FileEntry`, and it is defined in terms of the other; the other takes 2 `func`s `dirInit` and `dirReader` that are invoked when the initializing, resetting, advancing the cursor and filling the internal buffer.

Notably, for simplicity, entries are read in batches; `dirents` is always at most large `direntBufSize = 16` (currently), which is usually the (fixed) size of each batch, unless it is the initial fetch, or it is the last fetch:
- the initial fetch is `dirReader(n)`, where `n := direntBufSize - len(initial)`, and `len(initial)` is the size of the slice returned by `dirInit()`
- all fetches may return `len(batch) <= direntBufSize`; if `len(dirent) < direntBufSize`, then we have reached the end of the listing for this directory.

The cursor/iterator-like API has `Peek()`, `Advance()` methods. It also provides `Skip(n)`, `Reset()`, `Cookie()` and `Rewind(cookie)`.

- `Advance()` advances the `cursor` field and the `countRead` field. The `countRead` is the absolute index of the current item; `cursor` is the index in the current window (`dirent`)
- `Rewind(cookie)` allows seeking backwards to the position of the given cookie. The cookie corresponds to the internal `countRead`. When the cookie is 0, `Rewind(0)` is equivalent to `Reset()`. It is not possible to seek forward (beyond the cursor). It is not possible to seek backwards to a value outside the current window.
- `Skip(n)` is mostly used for testing, the other methods are more useful and used in `fdReaddirFn` too, and it is a short-hand for invoking `Advance()` `n` times.

Tests have been migrated to unit tests of `Readdir` when it made sense, and they were kept in the wasi namespace when it was simpler.

### DIFF
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -3,6 +3,7 @@ package sys
 import (
 	"io"
 	"io/fs"
+	"path"
 	"syscall"
 
 	"github.com/tetratelabs/wazero/internal/descriptor"
@@ -53,12 +54,27 @@ type FileEntry struct {
 	File fsapi.File
 }
 
+const direntBufSize = 16
+
 // Readdir is the status of a prior fs.ReadDirFile call.
 type Readdir struct {
-	// CountRead is the total count of files read including Dirents.
-	CountRead uint64
+	// cursor is the current position in the buffer.
+	cursor uint64
 
-	// Dirents is the contents of the last platform.Readdir call. Notably,
+	// countRead is the total count of files read including Dirents.
+	//
+	// Notes:
+	//
+	// * countRead is the index of the next file in the list. This is
+	//   also the value that Cookie returns, so it should always be
+	//   higher or equal than the cookie given in Rewind.
+	//
+	// * this can overflow to negative, which means our implementation
+	//   doesn't support writing greater than max int64 entries.
+	//   countRead uint64
+	countRead uint64
+
+	// dirents is a fixed buffer of size direntBufSize. Notably,
 	// directory listing are not rewindable, so we keep entries around in case
 	// the caller mis-estimated their buffer and needs a few still cached.
 	//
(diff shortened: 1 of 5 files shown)


evacchi:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`: expected one entry', src/bin/fd_readdir.rs:167:5
```

eh fair enough 🤷 😛

internal/sys/fs_test.go
```diff
@@ -274,3 +274,89 @@ func TestFSContext_Renumber(t *testing.T) {
...
+		// fixme? we are no longer using a sliding window
+		//{
+		//	name: "cookie is before current entries",
+		//	f: &Readdir{
+		//		countRead: 5,
+		//		//dirents:   testDirents,
+		//	},
+		//	cookie:        1,
+		//	expectedErrno: syscall.ENOSYS, // not implemented
+		//},
```
evacchi, line 333: I have kept this test here because it would not have been clear from the diff that it was deleted. I will delete it if we're ok with this: basically, it makes no longer sense because we are no longer using a sliding window. Should we go back to a sliding window instead?

codefromthecrypt: If I'm misreading, my mistake. a bit jet lagged.

First, I'd like to see what this test behavior is with the current change as it is hard to guess.

The sliding window was there to avoid caching a potentially massive directory listing, yet still allow lookup of recent entries. It is indeed simpler to not do a sliding window, but yeah that re-introduces this problem which we fixed earlier.

If the main issue is about diff size, maybe there's a way to extract some of this PR into another change, retaining the windowing. OTOH I'm also fine to have this wait until windowing is here. What we shouldn't do is merge partial implementation of something already supported as we should be able to ship main whenever we need to. I hope this makes sense.

codefromthecrypt (commented):
Definitely getting cleaner! let's make sure we retain sliding window so that a huge directory doesn't OOM. Ack it is more work, but worth it.

internal/sys/fs_test.go
evacchi: It returns ESUCCESS instead of ENOSYS. The test was trying to seek to a position before the window by sending a cookie with a lower value.

internal/sys/fs.go
```diff
@@ -53,20 +54,188 @@ type FileEntry struct {
...
+	dirents, errno := d.dirReader()
+	if errno != 0 {
+		return errno
+	}
+	d.dirents = append(d.dirents, dirents...)
+	return 0
+}
+
+// NewReaddirFromFileEntry is a constructor for Readdir that takes a FileEntry to initialize.
+func NewReaddirFromFileEntry(f *FileEntry) (*Readdir, syscall.Errno) {
```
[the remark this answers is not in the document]
evacchi: oh great point

internal/sys/fs_test.go
evacchi: restored.

evacchi:
restored the sliding window, cleaned up, squashed, rebased

codefromthecrypt (approved):
so much cleaner than what I had before. Great job!

PR #1484, opened 2023-05-19
compiler: allow inlining of internal/asm package functions

achille-roussel:
Go uses a scoring mechanism to determine which functions get inlined. Each statement gets assigned a scope and anything with a total below or equal to 80 gets inlined.

This PR modifies the `asm.CodeSegment` and `asm.Buffer` types to reduce the cost of functions that were on the hot path and allow inlining.

Chasing down inlining scores is somewhat of a dark art, but I was able to get the function scores by embedding `*CodeSegment` in `Buffer` so the call sites can see the inline-able leaf functions of `CodeSegment`, and the API remains identical from the standpoint of `Buffer`.

For the record, here's one way to ask the Go compiler to show functions that cannot be inlined in a package:
```
$ go build -gcflags '-m -m' ./internal/asm/ 2>&1 | grep 'cannot inline'
internal/asm/buffer.go:48:6: cannot inline (*CodeSegment).Map: function too complex: cost 153 exceeds budget 80
internal/asm/buffer.go:66:6: cannot inline (*CodeSegment).Unmap: function too complex: cost 111 exceeds budget 80
internal/asm/buffer.go:184:6: cannot inline (*CodeSegment).Grow: function too complex: cost 114 exceeds budget 80
internal/asm/buffer.go:175:6: cannot inline (*CodeSegment).growToSize: marked go:noinline
internal/asm/buffer.go:148:6: cannot inline (*CodeSegment).AppendBytes: function too complex: cost 86 exceeds budget 80
internal/asm/buffer.go:157:6: cannot inline (*CodeSegment).AppendUint32: function too complex: cost 82 exceeds budget 80
internal/asm/impl.go:42:6: cannot inline (*BaseAssemblerImpl).FinalizeJumpTableEntry: function too complex: cost 313 exceeds budget 80
```

This shrinks a couple of percentage points on CPU benchmarks, for example:
```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.60GHz
                                 │ /tmp/bench.1 │            /tmp/bench.2            │
                                 │    sec/op    │   sec/op     vs base               │
Compilation/without_extern_cache    5.051m ± 2%   4.899m ± 1%  -3.02% (p=0.005 n=12)

                                 │ /tmp/bench.1 │          /tmp/bench.2          │
                                 │     B/op     │     B/op      vs base          │
Compilation/without_extern_cache   597.8Ki ± 0%   597.8Ki ± 0%  ~ (p=0.291 n=12)

                                 │ /tmp/bench.1 │         /tmp/bench.2          │
                                 │  allocs/op   │  allocs/op   vs base          │
Compilation/without_extern_cache    1.074k ± 0%   1.074k ± 0%  ~ (p=1.000 n=12)
```
on darwin/amd64 the change did not seem to have any effect, so I avoided a few unsafe optimizations in favor of safer code (see comment in `AppendUint32`):
```
goos: darwin
goarch: arm64
                                 │ /tmp/bench.1 │         /tmp/bench.2          │
                                 │    sec/op    │   sec/op     vs base          │
Compilation/without_extern_cache    1.850m ± 2%   1.819m ± 2%  ~ (p=0.178 n=12)

                                 │ /tmp/bench.1 │          /tmp/bench.2          │
                                 │     B/op     │     B/op      vs base          │
Compilation/without_extern_cache   565.6Ki ± 0%   565.6Ki ± 0%  ~ (p=0.640 n=12)

                                 │ /tmp/bench.1 │          /tmp/bench.2          │
                                 │  allocs/op   │ allocs/op   vs base            │
Compilation/without_extern_cache     934.0 ± 0%   934.0 ± 0%  ~ (p=1.000 n=12) ¹
```

### DIFF
--- a/internal/asm/arm64/impl.go
+++ b/internal/asm/arm64/impl.go
@@ -390,12 +390,12 @@ func (a *AssemblerImpl) maybeFlushConstPool(buf asm.Buffer, endOfBinary bool) {
 			skipOffset = 0
 		}
 
-		buf.Append4Bytes(
+		buf.AppendUint32(u32(
 			byte(skipOffset),
 			byte(skipOffset>>8),
 			byte(skipOffset>>16),
 			0x14,
-		)
+		))
 
 		// Then adding the consts into the binary.
 		for _, c := range a.pool.Consts {
@@ -763,7 +763,7 @@ func errorEncodingUnsupported(n *nodeImpl) error {
 func (a *AssemblerImpl) encodeNoneToNone(buf asm.Buffer, n *nodeImpl) error {
 	switch n.instruction {
 	case UDF:
-		buf.Append4Bytes(0, 0, 0, 0)
+		buf.AppendUint32(0)
 		return nil
 	case NOP:
 		return nil
@@ -789,13 +789,13 @@ func (a *AssemblerImpl) encodeJumpToRegister(buf asm.Buffer, n *nodeImpl) error 
 		return fmt.Errorf("invalid destination register: %w", err)
 	}
 
-	buf.Append4Bytes(
+	buf.AppendUint32(u32(
 		0x00|(regBits<<5),
 		0x00|(regBits>>3),
 		0b000_11111|(opc<<5),
 		0b1101011_0|(opc>>3),
-	)
-	return err
+	))
+	return nil
 }
 
 func (a *AssemblerImpl) relativeBranchFinalize(code []byte, n *nodeImpl) error {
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
Thanks, and I agree it is dark arts, especially as you can't add a directive to fail if something can't be inlined. Things like this take paying attention to, but I think we have reviewers here that would, or could optimize again if things drift out of optimal later.

PR #1486, opened 2023-05-20
updates benchmark library versions

codefromthecrypt:
Verified output, here's an example

```
wazero-compiler 136.71  ns/op
wasmedge        2918.44 ns/op
Benchmark/memory.i32/Call-16
wazero-compiler 309.90  ns/op
wasmedge        8938.40 ns/op
Benchmark/memory.i64/Call-16
wazero-compiler 312.56  ns/op
wasmedge        8651.29 ns/op

wazero-compiler 142.29  ns/op
wasmer          3153.76 ns/op
Benchmark/memory.i32/Call-16
wazero-compiler 321.99  ns/op
wasmer          1476.40 ns/op
Benchmark/memory.i64/Call-16
wazero-compiler 318.96  ns/op
wasmer          1554.10 ns/op

wazero-compiler 144.44  ns/op
wasmtime        6294.14 ns/op
Benchmark/memory.i32/Call-16
wazero-compiler 317.62  ns/op
wasmtime        3766.36 ns/op
Benchmark/memory.i64/Call-16
wazero-compiler 308.69  ns/op
wasmtime        3844.60 ns/op
```


internal/integration_test/vs/wasmedge/wasmedge.go
```diff
@@ -1,4 +1,4 @@
-//go:build amd64 && cgo && !windows && wasmedge
```
codefromthecrypt, line 1: nice to have another exception gone. I verified the lib exists even if it is pinned to specific macos

PR #1487, opened 2023-05-22
Remove threads support

anuraaga:
Spec is still not in a great state


codefromthecrypt:
considering the state of the threads proposal and the mosquito-bites of broken builds, hacks, and other cognitive load on various devs since this was merged...

I'm wondering if it we aren't digging a hole. Could it be better to revert it and try again later when things normalize? https://github.com/WebAssembly/threads/issues/198

anuraaga:
Sure I think that makes sense too

mathetake:
At least the state makes me feel like we should put on hold the implementation on compiler backend at least until it reaches the phrase 4.

codefromthecrypt:
thanks folks. I think we learned a lot and will be better off on the redo. @anuraaga do you want to raise the revert (or re-purpose this one?)

anuraaga:
Sure, repurposed this PR

codefromthecrypt (approved):
Thanks for the understanding @anuraaga!

mathetake (approved):
🫡 👍

PR #1493, opened 2023-05-29
wasi: add support for sockets

evacchi:
This adds wasip1 sockets, and tests them on darwin, linux and windows via C, Go and Rust programming languages.

To use sockets, you can add the `-listen=host:port` via the CLI. You can also set the log scopes "sock,poll" to see activity related to socket accepts and transmissions.

As this is a new API, programmatic access is via an experimental package using context-based configuration.

For example, the below configures wazero to open a new random port for each module instantiated with the give context.
```go
sockCfg := sock.NewConfig().WithTCPListener("127.0.0.1", 0)
ctx = sock.WithConfig(ctx, sockCfg)
```


evacchi:
Now how are you supposed to get the `fd` that you can `accept()` at runtime? Do you just guess? 😅

achille-roussel:
> Now how are you supposed to get the fd that you can accept() at runtime? Do you just guess?

Pretty much; you can iterate over the preopens until one is not a directory, or just establish a convention between the host and guest on which fd number will be a socket to accept connections on.

evacchi:
🫠

and can you actually enumerate preopens in a standard way or you just `for i:=3; ;i++` ? 😅 

in the meantime I just noticed that if we attach this config to a module, this line:

```
		wasi_snapshot_preview1.MustInstantiate(ctx, rt)
		_, err = rt.InstantiateModule(ctx, code, conf)
```

will cause the listener to be instantiated twice 😬 


```
❯ wazero run --tcplisten :0 imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.wasm 
2023/05/29 22:47:00 Listening on: :0, fd=7
2023/05/29 22:47:00 Listening on: :0, fd=9
```

codefromthecrypt:
@achille-roussel do you know in practice is ever any more than one listener socket used?

achille-roussel:
I think the use cases are very limited, having more than one accepting socket seems like an edge case. 

That being said, it looks like wasmtime allows the creation of multiple sockets, the `--tcplisten` option can be provided multiple times.

experimental/net/tcplistener.go
```diff
@@ -0,0 +1,20 @@
+package net
+
+import "fmt"
+
+// TcpListenerKey is a context key for experimental
+type TcpListenerKey struct{}
```
codefromthecrypt, line 6: maybe re-use godoc style of FunctionListenerFactoryKey

codefromthecrypt, line 6:
```suggestion
type TcpListenersKey struct{}
```

experimental/net/tcplistener.go
```diff
@@ -0,0 +1,20 @@
...
+
+// TcpListenerColl is a collection of TcpListenerAddr.
+// It is the context value to the TcpListenerKey.
+type TcpListenerColl = []TcpListenerAddr
```
codefromthecrypt, line 10:
```suggestion
type TcpListeners = []TcpListener
```

experimental/net/tcplistener.go
```diff
@@ -0,0 +1,20 @@
...
+
+// TcpListenerAddr is an entry in a TcpListenerColl.
+type TcpListenerAddr struct {
```
codefromthecrypt, line 13:
```suggestion
type TcpListener struct {
```

experimental/net/tcplistener.go
```diff
@@ -0,0 +1,20 @@
...
+import "fmt"
+
+// TcpListenerKey is a context key for experimental
+type TcpListenerKey struct{}
+
+// TcpListenerColl is a collection of TcpListenerAddr.
+// It is the context value to the TcpListenerKey.
+type TcpListenerColl = []TcpListenerAddr
+
+// TcpListenerAddr is an entry in a TcpListenerColl.
```
codefromthecrypt, line 12:
```suggestion
// TcpListener is a host port to pre-open.
```

experimental/net/tcplistener.go
```diff
@@ -0,0 +1,20 @@
+package net
+
+import "fmt"
+
+// TcpListenerKey is a context key for experimental
+type TcpListenerKey struct{}
+
+// TcpListenerColl is a collection of TcpListenerAddr.
```
codefromthecrypt, line 8: // TcpListeners are TCP sockets to pre-open when instantiating a api.Module.

experimental/net/tcpconfig.go
```diff
@@ -0,0 +1 @@
+package net
```
codefromthecrypt, line 1: please simplify into one file, like net.go

codefromthecrypt (commented):
Some polishing comments for starters. I'd love to make a second pass once we have something in wasi_stdlib_test.go!

ncruces:
> Now how are you supposed to get the `fd` that you can `accept()` at runtime? Do you just guess? 😅

Convention seems about right, e.g. `systemd` socket activation (which is pretty close to this) uses environment variables (`$LISTEN_FDS` and `$LISTEN_FDNAMES`) to notify services of preopen sockets.

ncruces:
I bet it you can design your host so you can use `systemd` socket activation libraries in the guest.

experimental/net/tcpconfig.go
evacchi: but we're still keeping the subpackage right?

evacchi:
I don't know what's going on Linux... 🤔 at the beginning I thought I was opening too many ports, but now... 

```
--- FAIL: Test_sockAccept (0.03s)
    --- FAIL: Test_sockAccept/sock_accept (0.02s)
        require.go:330: expected no error, but was dial tcp [::]:37907: connect: cannot assign requested address
            /home/runner/work/wazero/wazero/imports/wasi_snapshot_preview1/sock_test.go:80
    --- FAIL: Test_sockAccept/sock_accept_(nonblock) (0.01s)
        require.go:330: expected no error, but was dial tcp [::]:37907: connect: cannot assign requested address
            /home/runner/work/wazero/wazero/imports/wasi_snapshot_preview1/sock_test.go:80
--- FAIL: Test_sockShutdown (0.02s)
```

achille-roussel:
Maybe you want to set the SO_REUSEADDR / SO_REUSEPORT socket options?

Or it may be a problem with `localhost` resolving to IPv4 vs IPv6? Try using `127.0.0.1` explicitly?

experimental/net/net.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+	return &netConfig{}, nil
+}
+
+type netConfig struct {
+	tcpListeners TCPListeners
+}
+
+func (c *netConfig) ToTCPListeners() (listeners []*net.TCPListener, err error) {
```
codefromthecrypt, line 34: move code like this directly into /runtime.go as this is a config type and it is easier to test when only doing that. For example, `FSConfig` doesn't have logic to open files, so this shouldn't have logic to create the listeners, only to specify them.

On chat what I meant was this basically, this object just holds the listener strings and documents what to put in (like host port rules).

just under where fs stuff happens, you'd do something like this.

```go
	var listeners []*net.TCPListener // normal go net package

	if n, ok := ctx.Value(internalnet.NetConfigKey{}).(*internalnet.NetConfig); ok {
		if listeners, err = f.ToTCPListeners(); err != nil {
			return
		}
	}
```

Then,`listeners` is passed to `internalsys.NewContext` as the last parameter and inserted with tests to make sure that when the context is closed, they are too.

codefromthecrypt (commented):
I hope this clarifies what I meant on slack. Good luck with the real work meanwhile (getting the socket tests passing)

achille-roussel:
I don't know if this is immediately useful, tho it might become at some point so I'm mentioning it just so the info is shared when it becomes of value.

We have added tests for the wasip1 socket APIs in Go https://github.com/golang/go/blob/094c75219a160be1667c9757f363ffad8926632b/src/runtime/internal/wasitest/tcpecho_test.go#L46-L51

Currently, these tests are disabled for wazero but hopefully, we'll be able to enable them once your work completes!

codefromthecrypt:
@achille-roussel haha so there's a reason to stick with `tcplisten` as the flag name ;)

codefromthecrypt:
ah nevermind I see the switch statement is about wasmtime. We can use anything we want.

@evacchi since this is merged, maybe what we can do is patch that file @achille-roussel mentioned in our CI (apply a diff) to add a wazero angle to it. Then our CI can magically show that we pass, and the diff can be contributed upstream after?

internal/sys/fs.go
```diff
@@ -47,6 +48,9 @@ type FileEntry struct {
 	// IsPreopen is a directory that is lazily opened.
 	IsPreopen bool
 
+	// IsConnection is a tcp connection.
```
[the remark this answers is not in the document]
evacchi: indeed, horrible hack. Will check for better alternatives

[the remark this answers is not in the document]
evacchi: Actually I just noticed this is not even used. I guess this was not necessary after all.

experimental/net/net.go
```diff
@@ -0,0 +1,90 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+)
+
+// NetConfig is a configuration for TCPListeners.
+type NetConfig interface {
```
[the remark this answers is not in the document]
evacchi: I completely misunderstood you were suggesting to put some `Config` _in_ the context key. I just went straight to (more or less) copying what was being done for Function Listeners! I suppose I have read Context + Listeners and I jumped straight to that conclusion!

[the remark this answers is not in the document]
evacchi: but I'll need `NewNetConfig()` in `experimental/net`, otherwise it won't be public. And if I need that to be there, then I'll need the impl here too to `return &netConfig{}`. Have I misunderstood again? 😅 

`netConfig` and the context key can all be unexported though, because now you would set them through the public API (I'll probably need a public way to extract it from the context though, unless I export the key)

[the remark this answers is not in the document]
codefromthecrypt: if NewNetConfig returns an interface, it shouldn't matter if its implementation is in internal/net

evacchi:
@achille-roussel:

> Or it may be a problem with localhost resolving to IPv4 vs IPv6? Try using 127.0.0.1 explicitly?

yes! 

> The Docker containers on the runners don’t get IPv6, even on the loopback device.

https://github.com/orgs/community/discussions/26331#discussioncomment-3251506

evacchi:
```
// Unimplemented functions.

func (fd *netFD) dup() (*os.File, error) {
	// TODO: Implement this
	return nil, syscall.EWINDOWS
}
```

ugh

evacchi:
oh that explains it LOL

```
func Recvfrom(fd Handle, p []byte, flags int) (n int, from Sockaddr, err error) {
	return 0, nil, EWINDOWS
}
```

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
```diff
@@ -336,7 +341,46 @@ func testOpen(t *testing.T, cmd string, bin []byte) {
...
+	// Close the listener and let it wazero create it instead.
+	ln.Close()
+	time.Sleep(1 * time.Second)
+	moduleConfig := wazero.NewModuleConfig().WithArgs("wasi", "socket").WithSysNanosleep()
+	// Instruct wazero to create the listener using the addr:port pair that was created and destroyed earlier.
+	// We assume that nobody stole that port in the meantime.
+	netCfg, err := experimentalnet.NewNetConfig().
+		WithTCPListenerFromString(addr)
+	require.NoError(t, err)
+	ctx := experimentalnet.WithNetConfig(testCtx, netCfg)
```
evacchi, line 370: the signature change in all the other tests is to explicitly pass a `ctx`. Otherwise we would have to modify the global to inject the listeners, and that would cause all tests to open a port for listening!

evacchi:
I can't believe I am already undrafting this. It may need a few more tests but in way better shape than I expected at this point!

internal/sysfs/sock.go
```diff
@@ -0,0 +1,66 @@
...
+		return syscall.EINVAL // FIXME: better errno?
+	}
+	syscallConn, err := tcpConn.SyscallConn()
+	if err != nil {
+		return platform.UnwrapOSError(err)
+	}
+	var actualErr error
+	// Control does not allow to return an error, but it is blocking;
+	// so it is ok to modify the external environment and setting
+	// `err` directly.
```
achille-roussel, line 33: I think this is actually referring to `actualErr` being set from the callback? If so it's a bit confusing to refer to `err` since there's another variable named `err` in the function.
```suggestion
	// `actualErr` directly.
```
Let me know if I'm mistaken.

internal/sysfs/sock.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if err != nil {
+		return platform.UnwrapOSError(err)
+	}
+	return 0
+}
+
+// SockShutdown exposes syscall.Shutdown on POSIX systems.
+func SockShutdown(f fsapi.File, how int) syscall.Errno {
+	// FIXME: can userland shutdown listeners?
```
achille-roussel, line 48: Can listening sockets be shutdown? They can't be read or written to so it's a bit unclear what that would do, the man page seems to indicate that they can't?
```
DESCRIPTION
     The shutdown() call causes all or part of a full-duplex connection on the socket associated with socket to be shut down.
```

```
     [ENOTCONN]         The specified socket is not connected.
```

internal/sysfs/sock_windows.go
```diff
@@ -0,0 +1,74 @@
...
+		uintptr(unsafe.Pointer(_p0)),
+		uintptr(len(buf)),
+		uintptr(flags),
+		0, // from *sockaddr (optional)
+		0) // fromlen *int (optional)
+	n = int(r0)
+	if n == -1 {
+		return n, e1
+	}
+	return
```
achille-roussel, line 73: Is this the same as?
```suggestion
	return int(r0), e1
```

achille-roussel (commented):
Gave it a first pass, the code is looking great overall 🙌

internal/sysfs/sock.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil, platform.UnwrapOSError(err)
+	}
+	return conn, 0
+}
+
+// SockSetNonblock exposes syscall.SetNonblock on POSIX systems.
+func SockSetNonblock(conn net.Conn) syscall.Errno {
+	tcpConn, ok := conn.(*net.TCPConn)
+	if !ok {
+		return syscall.EINVAL // FIXME: better errno?
```
codefromthecrypt, line 24: I think this should be a panic because our preconditions should have guaranteed this is a TCPConn. I'll look closer to see why this call site is unsure of the type.

internal/sysfs/sock.go
codefromthecrypt: I may try to rewrite this part as it is a shell game for sure.

experimental/net/net.go
```diff
@@ -0,0 +1,57 @@
...
+// NetConfig configures the host to set up host:port listeners
+// that the embedding host will allow the wasm guest to access.
+//
+// These will result in preopened socket file-descriptors at run-time.
+type NetConfig interface {
+	// WithTCPListener configures the host to set up the given host:port listener.
+	WithTCPListener(host string, port uint16) NetConfig
+	// WithTCPListenerFromString parses a string of the form "host:port"
+	// into a host:port pair.
+	WithTCPListenerFromString(conn string) (NetConfig, error)
```
codefromthecrypt, line 21: in a commit, I'll remove this as there's no need to add encoding logic at this level of abstraction.

codefromthecrypt (commented):
Thanks! I'll pull this locally for a cleanup pass.

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
codefromthecrypt: I have a change I will push to remove this footgun. Basically, a host module shouldn't instantiate listeners ever.

internal/sysfs/sock.go
codefromthecrypt: My guess is we have to see what happens. For example, wasmtime just pass directly to the underlying socket rather than guess if something should work or not. https://github.com/bytecodealliance/wasmtime/pull/4776/files#diff-ce08e37280f7832c5b5b018feed315bb2acf9eca12941698adb462b141aa590bR299

codefromthecrypt: similar in wasmer https://github.com/wasmerio/wasmer/blob/3adc37381d7a32aa25c4a27740e671ee50ba9fec/lib/wasi/src/syscalls/wasix/sock_shutdown.rs#L25

here's wasmedge https://github.com/WasmEdge/WasmEdge/blob/bc3c5641547fb179bc563f34abcb2b975a80b599/lib/host/wasi/inode-macos.cpp#L1260

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
```diff
@@ -336,7 +341,44 @@ func testOpen(t *testing.T, cmd string, bin []byte) {
...
+	require.NoError(t, err)
+	// Ensure that the address is correctly handled in CI.
+	// GitHub Actions may not hand over an IPv6 Address,
+	// so we force the IPv4 loopback explicitly.
+	port := ln.Addr().(*net.TCPAddr).Port
+	addr := fmt.Sprintf("127.0.0.1:%d", port)
+	// Close the listener and let it wazero create it instead.
+	ln.Close()
+	time.Sleep(1 * time.Second)
+	moduleConfig := wazero.NewModuleConfig().WithArgs("wasi", "socket").WithSysNanosleep()
```
codefromthecrypt, line 364: we should have a comment why this needs real sleep to work

evacchi: I suspect it was another attempt to "fix" CI. I can remove it locally. We can try to push again without it and see what happens

internal/sysfs/sock.go
codefromthecrypt: new code doesn't need to cast anymore

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -401,7 +406,20 @@ func fdFilestatGetFunc(mod api.Module, fd int32, resultBuf uint32) syscall.Errno
...
 
-	return writeFilestat(buf, &st)
+	filetype := getExtendedWasiFiletype(f.File, st.Mode)
+	return writeFilestat(buf, &st, filetype)
+}
+
+func getExtendedWasiFiletype(file fsapi.File, fm fs.FileMode) (ftype uint8) {
+	ftype = getWasiFiletype(fm)
+	if ftype == wasip1.FILETYPE_UNKNOWN {
+		if _, ok := file.(netapi.Sock); ok {
```
codefromthecrypt, line 416: clever

internal/sysfs/sock.go
```diff
@@ -0,0 +1,154 @@
...
+func (*tcpListenerFile) IsDir() (bool, syscall.Errno) {
+	// We need to override this method because WASI-libc prestats the FD
+	// and the default impl returns ENOSYS otherwise.
+	return false, 0
+}
+
+// Stat implements the same method as documented on File.Stat
+func (f *tcpListenerFile) Stat() (fs fsapi.Stat_t, errno syscall.Errno) {
+	// The mode is not really important, but it should be neither a regular file nor a directory.
+	fs.Mode = os.ModeIrregular
```
codefromthecrypt, line 44: good comment

codefromthecrypt (approved):
excellent work. I will do some polishing work only as this is basically ready to ship!

codefromthecrypt:
ok I added gotip tests from offline hints by @evacchi. now looking to see if they fail on windows 🤞

codefromthecrypt:
not sure if we should block on this, but latest zig-cc seems to not work without adjustments.

```bash
$ zig version
0.11.0-dev.3334+cd1417dbd
$ make build.examples.zig-cc
$ make test
--snip--
--- FAIL: Test_Sock (1.73s)
    --- FAIL: Test_Sock/zig-cc (0.81s)
        require.go:330: expected "wazero
            OK
            ", but was "ERR: accept: Success
            "
            /Users/adrian/oss/wazero/imports/wasi_snapshot_preview1/wasi_stdlib_test.go:396
            /Users/adrian/oss/wazero/imports/wasi_snapshot_preview1/wasi_stdlib_test.go:366
FAIL
--snip--
```

codefromthecrypt:
cool. gotip passed on windows and skipped without fail on the scratch tests

codefromthecrypt:
I'm concerned about the zig-cc because it is returning -1 from accept even though we returned success..

```
	==> wasi_snapshot_preview1.sock_accept(fd=3,flags=0,result.fd=1044432)
	<== errno=ESUCCESS
```

codefromthecrypt:
so basically, I think we need to verify if this is something about wasi-libc recent versions acting strangely. We use zig-cc to implicitly get c+wasi without a complicated install.

codefromthecrypt:
so now 2.5/3 integration tests work
* some older version of zig-cc works, but not latest (0.5)
* gotip GOOS=wasip1 works
* cargo-wasi rust works

evacchi:
just fetch zig's latest build `0.11.0-dev.3336+6c2f37455` and that seems to work for me!

evacchi:
the issue seems to arise only when the `-O` flag is set to > `-O0` (we are using `-Oz` but I have tested below that and it won't work). I assume some UB in libc maybe? the test fails on `wasmtime` 9.0.2 too

ncruces:
`wasi-sdk` is not a terribly complicated install? It's one a download, just larger. I moved away from `zig` to it when `wasi-sdk` published a much more recent, stable version.

evacchi:
@ncruces indeed, it's just curl, tar, export PATH; with zig cc we could avoid having yet another tool to install; yet...

codefromthecrypt:
Thanks to @evacchi for figuring out that for some reason `-Oz` breaks the code. leaving that out things work fine 🤷

codefromthecrypt:
fwiw I don't really want to toss zig quite yet, because it is really helpful and no-brainer install on the platforms, self-contained etc. This was annoying for sure, but let's wait before adding more tools to install.

codefromthecrypt:
@evacchi can you take a final pass and review for merge? while there are some other sorts of polish we can do, everything public looks good to me, especially for experimental.

I just added the logging scope in the last commit:
```bash
$ wazero run -hostlogging=sock -listen=127.0.0.1:8888 main.wasm
==> wasi_snapshot_preview1.sock_accept(fd=3,flags=)
<== (fd=4,errno=ESUCCESS)
==> wasi_snapshot_preview1.sock_recv(fd=4,ri_data=1044344,ri_data_len=1,ri_flags=)
<== (ro_datalen=0,ro_flags=ro_flags=,errno=ESUCCESS)
foo

OK
```

config.go
```diff
@@ -649,6 +651,8 @@ type moduleConfig struct {
 	environKeys map[string]int
 	// fsConfig is the file system configuration for ABI like WASI.
 	fsConfig FSConfig
+	// sockConfig is the network listener configuration for ABI like WASI.
+	sockConfig *internalsock.Config
```
evacchi, line 655: I see this is now sock instead fo net :)

evacchi:
changes look good to me ( can't self-approve ;) )

codefromthecrypt:
I rewrote the desc, shipping the 🧦

config.go
codefromthecrypt: yeah conventionally sock won, both in syscall layer and in most of go's type names

Issue #1500, opened 2023-06-02
support non-blocking files


codefromthecrypt:
**Is your feature request related to a problem? Please describe.**

Currently, we support non-blocking stdio and sockets, but not files. This leads to an awkward and unnecessary [skip in go's tests](https://github.com/golang/go/blob/5f695f7072848e425289ca94bc460da813de66da/src/runtime/internal/wasitest/nonblock_test.go#L41).

**Describe the solution you'd like**

Go supports non-blocking access and we support native access to files, so we should be able to implement this. If there are any caveats about windows, support can begin with a limited set of platforms.

**Describe alternatives you've considered**

continue to not do this

**Additional context**

our file type already has a SetNonblock setting, so that's a start. We also have a separate file implementation for raw os-backed files. So, we should be able to figure out why we aren't implementing non-blocking access for os files, then figure out how to implement it.

evacchi:
a quick summary of what's required to make this works.

In theory, we can easily enable SetNonblock() for regular files, but effectively it would only work on *NIX. In fact, the `syscall.SetNonblock()` fun on Windows is a no-op.

Now, even if we did support that only on *NIX, we should still fix poll_oneoff; because as much as it would still work without it (read() returns EAGAIN and you could just retry after a little), existing code may still assume that `select` works.

so:

- we can enable syscall.Setnonblock() on regular files
- it won't work on Windows unless we explicitly implement something
- read() now does not block and may return EAGAIN
- we should fix poll_oneoff because it is currently special-cased for stdin.

now the good news about `poll_oneoff` is that there actually _is_ a `select()` implementation that works on Windows
... the bad news it only works on sockets. Better than nothing.

So in short:

- regular files:
- easy to set the flag for nonblocking
- requires reworking poll_oneoff (but this can be done in a separate PR)
  - won't work for free on Windows
  - would work almost for free on Windows+Sockets (via WinSock select)

codefromthecrypt:
some trouble in HTTP land. I began troubleshooting with @chriso but I think we should all be able to enjoy a weekend and I don't think this will quickly resolve. https://github.com/tetratelabs/wazero/pull/1503

codefromthecrypt:
[...]

PR #1502, opened 2023-06-02
wasi: non-blocking i/o on files on *nix

evacchi:
Addresses #1500.

We enable non-blocking I/O on files (including FIFOs); this allows [nonblock_test.go](https://github.com/golang/go/blob/5f695f7072848e425289ca94bc460da813de66da/src/runtime/internal/wasitest/nonblock_test.go) to run to completion. I am pushing here a simpler test in C, but I have tested the one in the Go tree separately. We can work at adding the same test if we want.

This is not supported on Windows (yet?) because 
1) no Mkfifo means we have to rethink the test ;) 
2) `syscall.SetNonblock()` on Windows has the following implementation :D

```go
func SetNonblock(fd Handle, nonblocking bool) (err error) {
	return nil
}
```


achille-roussel (approved):
🙌

ncruces (approved):

codefromthecrypt:
thanks @evacchi for another epic evening. When you have a chance, add to #1500 something like:

update: blah blah works
next step: copy paste notes for windows

that way it is easier to see well, what's next in the original thingy, as I'll link to it in the next release notes

mathetake:
`imports/wasi_snapshot_preview1/example/testdata/zig-cc/cat.wasm` why in this example only the binary has been changed while the source left intact?

codefromthecrypt:
good eye. I should have caught this. as I've noticed folks are sometimes building wasm manually or with different versions I would suggest us all look carefully and ask each other.

1. if you meant to re-build, did you use the [make targets like our examples workflow](https://github.com/tetratelabs/wazero/blob/main/.github/workflows/examples.yaml#L86-L105)?
2. if you meant to re-build, did you verify the version of the [toolchain is same as our examples workflow](https://github.com/tetratelabs/wazero/blob/main/.github/workflows/examples.yaml#L19-L21)?

evacchi:
My bad. I rebuilt using make, without checking the version was aligned and committed all by mistake 🙏

PR #1503, opened 2023-06-03
wasi: fix nonblocking sockets on *NIX (gotip net/http)

codefromthecrypt:
This PR fixes nonblocking I/O when handling sockets with `net/http`. 

In this PR we further separate the Unix implementation from the Windows implementation; in the original implementation, we rely on Go's stdlib and the `net` package: this implementation is now Windows only.

On other Unix platforms, we instead go straight to the syscall layer and we pass-through most parameters.

We also add a `baseSockFile` struct that implements base behavior for sock files; this is extended as follows:

- unix: `tcpListenerFile`, `tcpConnFile`
- windows: `winTcpListenerFile`, `winTcpConnFile`
- unsupported: `unsupportedSockFile` 
  (implements only `socketapi.TCPSock.Accept()`, returns `ENOSYS`: no need for `TCPConn`, as nothing is able to return it)

### DIFF
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -556,9 +556,9 @@ func Test_fdFdstatSetFlags(t *testing.T) {
 	// Let's remove O_APPEND.
 	requireErrnoResult(t, wasip1.ErrnoSuccess, mod, wasip1.FdFdstatSetFlagsName, uint64(fd), uint64(0))
 	require.Equal(t, `
-==> wasi_snapshot_preview1.fd_fdstat_set_flags(fd=4,flags=0)
+==> wasi_snapshot_preview1.fd_fdstat_set_flags(fd=4,flags=)
 <== errno=ESUCCESS
-`, "\n"+log.String())
+`, "\n"+log.String()) // FIXME? flags==0 prints 'flags='
 	log.Reset()
 
 	// Without O_APPEND flag, the data is written at the beginning.
@@ -568,9 +568,9 @@ func Test_fdFdstatSetFlags(t *testing.T) {
 	// Restore the O_APPEND flag.
 	requireErrnoResult(t, wasip1.ErrnoSuccess, mod, wasip1.FdFdstatSetFlagsName, uint64(fd), uint64(wasip1.FD_APPEND))
 	require.Equal(t, `
-==> wasi_snapshot_preview1.fd_fdstat_set_flags(fd=4,flags=1)
+==> wasi_snapshot_preview1.fd_fdstat_set_flags(fd=4,flags=APPEND)
 <== errno=ESUCCESS
-`, "\n"+log.String())
+`, "\n"+log.String()) // FIXME? flags==1 prints 'flags=APPEND'
 	log.Reset()
 
 	// with O_APPEND flag, the data is appended to buffer.
--- a/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
+++ b/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
@@ -77,26 +77,7 @@ func mainHTTP() error {
 	defer ln.Close()
 
 	// Serve middleware that echos the request body to the response.
-	if err = http.Serve(&onlyOnceListener{Listener: ln}, echo{}); err == nil || err == http.ErrServerClosed {
-		return nil
-	} else {
-		return err
-	}
-}
-
-type onlyOnceListener struct {
-	net.Listener
-
-	accepted bool
-}
-
-func (ln *onlyOnceListener) Accept() (net.Conn, error) {
-	// TODO: this is a problem because we stop before the in-flight request is handled.
-	if ln.accepted {
-		return nil, http.ErrServerClosed
-	}
-	ln.accepted = true
-	return ln.Listener.Accept()
+	return http.Serve(ln, echo{})
 }
 
 type echo struct{}
--- a/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
+++ b/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"os"
@@ -86,7 +87,7 @@ func (echo) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Copy up to 32 bytes from the request to the response, appending a newline.
 	// Note: the test should write: "wazero", so that's all we should read.
 	var buf [32]byte
-	if n, err := r.Body.Read(buf[:]); err != nil {
+	if n, err := r.Body.Read(buf[:]); err != nil && err != io.EOF {
 		panic(err)
 	} else if n, err = w.Write(append(buf[:n], '\n')); err != nil {
 		panic(err)
--- a/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
+++ b/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
@@ -78,14 +78,14 @@ func mainHTTP() error {
 	defer ln.Close()
 
 	// Serve middleware that echos the request body to the response once, then quits.
-	h := &echoOnce{ch: make(chan interface{}, 1)}
+	h := &echoOnce{ch: make(chan struct{}, 1)}
 	go http.Serve(ln, h)
 	<-h.ch
 	return nil
 }
 
 type echoOnce struct {
-	ch chan interface{}
+	ch chan struct{}
 }
 
 func (e echoOnce) ServeHTTP(w http.ResponseWriter, r *http.Request) {
--- a/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
+++ b/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
@@ -77,13 +77,18 @@ func mainHTTP() error {
 	}
 	defer ln.Close()
 
-	// Serve middleware that echos the request body to the response.
-	return http.Serve(ln, echo{})
+	// Serve middleware that echos the request body to the response once, then quits.
+	h := &echoOnce{ch: make(chan interface{}, 1)}
+	go http.Serve(ln, h)
+	<-h.ch
+	return nil
 }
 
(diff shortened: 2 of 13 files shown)


imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
```diff
@@ -43,3 +50,64 @@ func mainSock() error {
...
+}
+
+type onlyOnceListener struct {
+	net.Listener
+
+	accepted bool
+}
+
+func (ln *onlyOnceListener) Accept() (net.Conn, error) {
+	// TODO: this is a problem because we stop before the in-flight request is handled.
```
codefromthecrypt, line 94: the main issue in this test is whether we do it in the guest, or externally, we want the test to stop after one request. Currently, the TCP test stops after one request, and I wanted to do the same here.

Right now, this approach doesn't work because it closes the server before the in-flight request completes. I also tried adding a channel, but go into a deadlock and feel maybe I'm digging a hole.

codefromthecrypt: Actually I think the request is never handled. I'll revert the "only once" stuff, so that we can tell if anything is working first.

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
```diff
@@ -396,3 +404,59 @@ func testSock(t *testing.T, bin []byte) {
...
+
+	// Give a little time for _start to complete
+	sleepALittle()
+
+	// Now, send a POST to the address which we had pre-opened.
+	body := bytes.NewReader([]byte("wazero"))
+	req, err := http.NewRequest(http.MethodPost, "http://"+tcpAddr.String(), body)
+	require.NoError(t, err)
+
+	// TODO: test hangs here
```
codefromthecrypt, line 450: here's the hang. from the log, the request is never read.

evacchi: it looks like it's stuck in a loop here:

```go
// poll/fd_unix.go
func (fd *FD) Accept() (int, syscall.Sockaddr, string, error) {
	...
	for {
		...
		case syscall.EAGAIN:
			if fd.pd.pollable() {
				if err = fd.pd.waitRead(fd.isFile); err == nil {
					continue
				}
			}
		...
```

evacchi: ok I am now calling `SetDeadline()`, catching the timeout error, and returning EAGAIN instead. This _I think_ causes the Wasm module to do the loop above 😬 so the wasm module is effectively unstuck but I think it's still looping until the connection becomes available; OTOH this lets other goroutines interleave; in fact I can see a poll_oneoff happening:

```
	==> wasi_snapshot_preview1.clock_time_get(id=monotonic,precision=0)
	<== (timestamp=81477104823666,errno=ESUCCESS)
	==> wasi_snapshot_preview1.sock_accept(fd=3,flags=)
	<== (fd=,errno=EAGAIN)
	==> wasi_snapshot_preview1.poll_oneoff(in=21422080,out=21446656,nsubscriptions=2)
	<== (nevents=2,errno=ESUCCESS)
	==> wasi_snapshot_preview1.sock_accept(fd=3,flags=)
	<== (fd=,errno=EAGAIN)
	==> wasi_snapshot_preview1.poll_oneoff(in=21422080,out=21446656,nsubscriptions=2)
	<== (nevents=2,errno=ESUCCESS)
	==> wasi_snapshot_preview1.sock_accept(fd=3,flags=)
	<== (fd=,errno=EAGAIN)
	==> wasi_snapshot_preview1.poll_oneoff(in=21422080,out=21446656,nsubscriptions=2)
	<== (nevents=2,errno=ESUCCESS)
	==> wasi_snapshot_preview1.sock_accept(fd=3,flags=)
	<== (fd=,errno=EAGAIN)
	==> wasi_snapshot_preview1.poll_oneoff(in=21422080,out=21446656,nsubscriptions=2)
	<== (nevents=2,errno=ESUCCESS)
	==> wasi_snapshot_preview1.sock_accept(fd=3,flags=)
	<== (fd=,errno=EAGAIN)
	==> wasi_snapshot_preview1.poll_oneoff(in=21422080,out=21446656,nsubscriptions=2)
	<== (nevents=2,errno=ESUCCESS)
	==> wasi_snapshot_preview1.sock_accept(fd=3,flags=)
	<== (fd=,errno=EAGAIN)
	==> wasi_snapshot_preview1.poll_oneoff(in=21422080,out=21446656,nsubscriptions=2)
	<== (nevents=2,errno=ESUCCESS)
	==> wasi_snapshot_preview1.sock_accept(fd=3,flags=)
```

evacchi:
I have pushed a change to use straight syscalls on *Nix, and fallbacks to the original code on Windows.
I haven't had time to layout the code properly, so it should be better move things in the right `_unix` and `_windows` files. Apologies for this not being great.

I have also updated the examples, hopefully correctly: even after the accept() was made pollable, the example failed on io.EOF, which however I think should be fine, so I modified the example to not fail on io.EOF. In that case, the buffer is read correctly and it contains headers+the string `wazero`.

On Windows I am skipping the new gotip test because it won't work, since we don't support nonblocking sockets (tested and yep, it hangs)

ncruces (approved):

codefromthecrypt (commented):
I agree with this approach and looking forward to having this shipped. We can do a patch even.

evacchi:
pushed a cleaner version. This may be good to go.

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
```diff
@@ -396,3 +402,67 @@ func testSock(t *testing.T, bin []byte) {
...
+	b, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+	require.Equal(t, "wazero\n", string(b))
+
+	cancelFunc()
+
+	// FIXME: this hangs because the example listens forever;
+	// 	      however mod.Close() does not result in a clean shudown.
+	// console := <-ch
+	// require.Equal(t, "", console)
```
evacchi, line 467: solving by serving only 1 request and then exiting the Wasm routine

evacchi:
Added description to the head post, marking ready for review.

achille-roussel (approved):
🙌

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
codefromthecrypt: hah I was thinking about this route also.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -556,9 +556,9 @@ func Test_fdFdstatSetFlags(t *testing.T) {
 	// Let's remove O_APPEND.
 	requireErrnoResult(t, wasip1.ErrnoSuccess, mod, wasip1.FdFdstatSetFlagsName, uint64(fd), uint64(0))
 	require.Equal(t, `
-==> wasi_snapshot_preview1.fd_fdstat_set_flags(fd=4,flags=0)
+==> wasi_snapshot_preview1.fd_fdstat_set_flags(fd=4,flags=)
 <== errno=ESUCCESS
-`, "\n"+log.String())
+`, "\n"+log.String()) // FIXME? flags==0 prints 'flags='
```
codefromthecrypt, line 561: that's fine IMHO because there are no flags and the formatted flags are pipe delimited. that said it is easy to change to special case none differently. @achille-roussel @ncruces any preferences on this one?

codefromthecrypt: this default applies to every flag type we have in wasi, btw.

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -568,9 +568,9 @@ func Test_fdFdstatSetFlags(t *testing.T) {
 	// Restore the O_APPEND flag.
 	requireErrnoResult(t, wasip1.ErrnoSuccess, mod, wasip1.FdFdstatSetFlagsName, uint64(fd), uint64(wasip1.FD_APPEND))
 	require.Equal(t, `
-==> wasi_snapshot_preview1.fd_fdstat_set_flags(fd=4,flags=1)
+==> wasi_snapshot_preview1.fd_fdstat_set_flags(fd=4,flags=APPEND)
 <== errno=ESUCCESS
-`, "\n"+log.String())
+`, "\n"+log.String()) // FIXME? flags==1 prints 'flags=APPEND'
```
codefromthecrypt, line 573: according to docs, flags is an fdflags type, so usually in log formatting we format the flag name vs have people need to look up what it was and do the bit flag math in their head. Maybe it was surprising that the logging "fixed" in a different change? https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-fdflags-flagsu16

internal/sysfs/sock_test.go
```diff
@@ -18,18 +20,26 @@ func TestTcpConnFile_Write(t *testing.T) {
 	require.NoError(t, err)
 	defer tcp.Close() //nolint
 
-	file := tcpConnFile{tc: tcp}
-	n, errno := file.Write([]byte("wazero"))
+	file := newTcpConn(tcp)
+	errno := syscall.Errno(0)
+	// Ensure we don't interrupt until we get a non-zero errno,
```
codefromthecrypt, line 25: great comment

imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
```diff
@@ -43,3 +51,52 @@ func mainSock() error {
...
+	// Copy up to 32 bytes from the request to the response, appending a newline.
+	// Note: the test should write: "wazero", so that's all we should read.
+	var buf [32]byte
+	if n, err := r.Body.Read(buf[:]); err != nil && err != io.EOF {
+		panic(err)
+	} else if n, err = w.Write(append(buf[:n], '\n')); err != nil {
+		panic(err)
+	}
+	// Once one request was served, close the channel.
+	close(e.ch)
```
codefromthecrypt, line 101: I think this is what I didn't get right when I tried to have the handler quit after one call and not loop back into the accept loop. glad you got it!

codefromthecrypt (commented):
thanks so much for sorting this out!

PR #1504, opened 2023-06-05
compiled: fix compileMemoryAccessOffsetSetup comment

anuraaga:
I've been looking at the assembler just for fun, and after a while trying to grok this function, I believe it is actually an issue with the comment and not my understanding. But if not, let me know and close this PR :)

The function does compute `"base + offsetArg + targetSizeInBytes"` for the purpose of a bounds check, but what it returns is `"base + offsetArg"`, I think


mathetake (approved):
👍

PR #1505, opened 2023-06-05
compiler(amd64): generate trap exit code once

ncruces:
I've been looking at compiled code (amd64) and noticed out-of-bounds checks in `compileMemoryAccessCeilSetup` account for quite a bit of runtime, so I set out to try and optimize them a bit.

Basically on every instruction that might trap, we currently compile the trap and skip over it if a condition is met. This blows up compiled code significantly if there are a lot of instructions that might trap.

I kept pretty much the same logic, but cache a trap's address the first time it's compiled, and on future appearances directly jump into the previously compiled trap if the condition is *not* met.

This should not affect code generated for functions that either don't trap at all, or that trap only once (I tried not to make those worse). Functions that trap a lot get the benefit.

I'm making this a draft because I'm unsure if this the best a approach. I don't want to detour @mathetake from his efforts in overhauling the compiler (this is but a stopgap) but would appreciate comments.

Tested on my [`speedtest1`](https://github.com/ncruces/go-sqlite3/tree/main/vfs/tests/speedtest1) SQLite benchmark:
```sh
speedtest1 % (cd ~/wazero; git switch main)
speedtest1 % go test -bench=. -benchtime=50x -count=10 --memdb | tee before.txt
speedtest1 % (cd ~/wazero; git switch traps)
speedtest1 % go test -bench=. -benchtime=50x -count=10 --memdb | tee trap.txt
```

Compiled code size goes down from 10.1MiB to 8.1MiB (-20%).
Benchstat results:
```sh
goos: darwin
goarch: amd64
               │ before.txt  │              trap.txt               │
               │   sec/op    │   sec/op     vs base                │
_speedtest1-12   87.27m ± 3%   75.95m ± 1%  -12.96% (p=0.000 n=10)
```

@codefromthecrypt any ideas what other benchmarks I should try?

### DIFF
--- a/internal/engine/compiler/impl_amd64.go
+++ b/internal/engine/compiler/impl_amd64.go
@@ -93,6 +93,7 @@ type amd64Compiler struct {
 	stackPointerCeil uint64
 	// assignStackPointerCeilNeeded holds an asm.Node whose AssignDestinationConstant must be called with the determined stack pointer ceiling.
 	assignStackPointerCeilNeeded asm.Node
+	compiledTrapTargets          [nativeCallStatusModuleClosed]asm.Node
 	withListener                 bool
 	typ                          *wasm.FunctionType
 	// locationStackForEntrypoint is the initial location stack for all functions. To reuse the allocated stack,
@@ -869,13 +870,9 @@ func (c *amd64Compiler) compileCallIndirect(o *wazeroir.UnionOperation) error {
 	// tmp = &module.Tables[0] + Index*8 = &module.Tables[0] + sizeOf(*TableInstance)*index = module.Tables[o.TableIndex].
 	c.assembler.CompileMemoryToRegister(amd64.MOVQ, tmp, int64(tableIndex*8), tmp)
 
-	// Then, we need to check if the offset doesn't exceed the length of table.
+	// Then, we need to trap if the offset exceeds the length of table.
 	c.assembler.CompileMemoryToRegister(amd64.CMPQ, tmp, tableInstanceTableLenOffset, offset.register)
-	notLengthExceedJump := c.assembler.CompileJump(amd64.JHI)
-
-	// If it exceeds, we return the function with nativeCallStatusCodeInvalidTableAccess.
-	c.compileExitFromNativeCode(nativeCallStatusCodeInvalidTableAccess)
-	c.assembler.SetJumpTargetOnNext(notLengthExceedJump)
+	c.compileTrapFromNativeCode(amd64.JLS, nativeCallStatusCodeInvalidTableAccess)
 
 	// next we check if the target's type matches the operation's one.
 	// In order to get the type instance's address, we have to multiply the offset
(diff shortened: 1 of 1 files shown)


ncruces:
One alternative I considered, but which I didn't implement, is to compile each trap once per module, regardless of it being used.

I tried doing it in the function preamble, but this makes functions that don't trap worse, which I wanted to avoid. Results for functions that _do_ trap are similar, slightly _better_ even.

But since doing it once per module requires `engine.go` changes, and I don't have an `arm64`, I did this for a start.

mathetake (approved):
Excellent and thank you for the detailed note. Could you do the same to arm64 as well if you have time?

ncruces:
The force push is just a rebase, to make sure the tests pass after @evacchi fix.

I tried to extend this idea to the "trapping float to int conversions" and I thought it _might_ be clearer if the condition was kept inverted (because there's always trapping and non-trapping branches), that leads to this patch instead:
https://github.com/tetratelabs/wazero/compare/main...ncruces:wazero:trap2

I honestly don't know which one reads better.

Float to int conversions aren't common enough to make a performance difference, so we may as well ignore them: 95% of the benefit is from memory access, and the other 5% from table access for indirect calls.

mathetake:
let's do the same to trapping int to float as well for consistency

ncruces:
- added most trapping int to float
- renamed variables in `compileTrapFromNativeCode` for readability
- fixed some comments for clarity
- tested again to confirm no regressions

Benchstat results:
```
goos: darwin
goarch: amd64
               │ before.txt  │              trap.txt               │
               │   sec/op    │   sec/op     vs base                │
_speedtest1-12   88.91m ± 2%   75.64m ± 2%  -14.92% (p=0.000 n=10)
```

internal/engine/compiler/impl_amd64.go
```diff
@@ -2276,13 +2250,13 @@ func (c *amd64Compiler) emitUnsignedI32TruncFromFloat(isFloat32Bit, nonTrapping
...
-	if !nonTrapping {
-		c.compileExitFromNativeCode(nativeCallStatusIntegerOverflow)
-	} else {
+	if nonTrapping {
 		err = c.assembler.CompileStaticConstToRegister(amd64.MOVL, c.maximum32BitUnsignedInt, result)
 		if err != nil {
 			return err
 		}
+	} else {
+		c.compileExitFromNativeCode(nativeCallStatusIntegerOverflow)
```
ncruces, line 2259: I skipped this one since I couldn't make sense of the above logic, but can try a bit harder if you want me to.

internal/engine/compiler/impl_amd64.go
```diff
@@ -2403,13 +2374,13 @@ func (c *amd64Compiler) emitUnsignedI64TruncFromFloat(isFloat32Bit, nonTrapping
...
-	if !nonTrapping {
-		c.compileExitFromNativeCode(nativeCallStatusIntegerOverflow)
-	} else {
+	if nonTrapping {
 		err = c.assembler.CompileStaticConstToRegister(amd64.MOVQ, c.maximum64BitUnsignedInt, result)
 		if err != nil {
 			return err
 		}
+	} else {
+		c.compileExitFromNativeCode(nativeCallStatusIntegerOverflow)
```
ncruces, line 2383: Ditto.

mathetake (approved):

PR #1506, opened 2023-06-05
wasi: try to make Test_Nonblock less flaky

evacchi:
We wait a little less to print each dot on the console in the wasm binary.
We add a blocking wait on a channel before we start sleeping.


evacchi:
added a blocking wait on a channel before the actual sleep, this may be good enough

ncruces (approved):

PR #1507, opened 2023-06-07
vs: fixes wasmedge version

mathetake:
`v0.12.1` is not existent


codefromthecrypt:
sorry about this. I think I was double checking the GitHub action section and noticed it was 0.12.1 latest, and then forgot the project isn't consistent.

PR #1508, opened 2023-06-08
compiler(amd64): generate trap exit code once

ncruces:
This finishes of #1505 for consistency, and (maybe) future reference.

Basically, this addresses the last few traps that weren't being reused, because control flow was a bit harder to parse.

This instructions are much less common, so the performance impact won't be measurable.


mathetake (approved):
cc @evacchi this also should be ported to arm64

mathetake:
Thanks @ncruces you rock!

PR #1509, opened 2023-06-09
compiler(amd64): eliminate a bounds check in memory copy

ncruces:
This uses a conditional move to find the maximum and removes one of the bounds checks thanks to it.

The difference is pretty negligible in my [`speedtest1`](https://github.com/ncruces/go-sqlite3/tree/main/vfs/tests/speedtest1) SQLite benchmark, but measurable:
```sh
speedtest1 % (cd ~/wazero; git switch main)
speedtest1 % go test -bench=. -benchtime=50x -count=10 --memdb | tee before.txt
speedtest1 % (cd ~/wazero; git switch traps)
speedtest1 % go test -bench=. -benchtime=50x -count=10 --memdb | tee trap.txt
```

Benchstat results:
```sh
goos: darwin
goarch: amd64
               │ before.txt  │             after.txt              │
_speedtest1-12   77.29m ± 1%   76.53m ± 2%  -0.99% (p=0.023 n=10)
```


mathetake (approved):
nice and sweet!

PR #1511, opened 2023-06-10
site: add wasi-go to the list of community projects

achille-roussel:
Add https://github.com/stealthrocket/wasi-go to the list of community projects built on wazero.

### DIFF
--- a/site/content/community/users.md
+++ b/site/content/community/users.md
@@ -15,9 +15,10 @@ considering their efforts before starting your own!
 | [go-pdfium][23]  | [PDFium][24] bindings to do PDF operations in Go, also available as end application [pdfium-cli][25] |
 | [go-re2][7]      | high performance regular expressions                                                                 |
 | [go-sqlite3][11] | [SQLite][12] bindings, `database/sql` driver                                                         |
-| [mjml-go][19]    | Compile [MJML][20] to HTML directly in Go                                                            |
+| [wasi-go][33]    | WASI host module for Wazero with experimental support for socket extensions                          |
 | [wazergo][29]    | Generics library for type-safe and high performance wazero host modules                              |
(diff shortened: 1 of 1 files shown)


site/content/community/users.md
```diff
@@ -15,9 +15,10 @@ considering their efforts before starting your own!
 | [go-pdfium][23]  | [PDFium][24] bindings to do PDF operations in Go, also available as end application [pdfium-cli][25] |
 | [go-re2][7]      | high performance regular expressions                                                                 |
 | [go-sqlite3][11] | [SQLite][12] bindings, `database/sql` driver                                                         |
-| [mjml-go][19]    | Compile [MJML][20] to HTML directly in Go                                                            |
+| [wasi-go][33]    | WASI host module for Wazero with experimental support for socket extensions                          |
 | [wazergo][29]    | Generics library for type-safe and high performance wazero host modules                              |
 | [Wetware][28]    | Simple, secure & scalable clusters                                                                   |
+| [mjml-go][19]    | Compile [MJML][20] to HTML directly in Go                                                            |
```
achille-roussel, line 21: I moved this line to follow the lexicographic order.

codefromthecrypt: I'm actually wondering if we shouldn't have both wasi-go and wazergo in an Extensions section? As they are a bit different than the others..

codefromthecrypt (approved):
Thanks and works for me!

At some point maybe we will want to refactor the index, but for now it is in a good place as any.

site/content/community/users.md
achille-roussel: I refrained from adding more sections because I fell like there were many already, but I agree that this index probably needs a periodic review to make sure the structure and content remain meaningful 👍

PR #1513, opened 2023-06-12
compiler(amd64): emit smaller instructions

ncruces:
This leads to an about 0.5% reduction in generated code size, and no change in runtime.
But it's a trivial change.


mathetake (approved):

mathetake:
cool!

PR #1514, opened 2023-06-12
compiler(amd64): avoid unnecessary jump in signed remainder

ncruces:
This trades an unneeded XOR for one less jump.
Confirmed clang does the same tradeoff.

At this point, it looks like I'm playing for commit count.
OTOH I assume these code blocks _will_ be reused in the new compiler (SSA optimization works at a different level).


mathetake (approved):
Nice and reasonable choice. Thank you for your research on the existing compiler's behavior!

PR #1515, opened 2023-06-13
compiler(amd64): generate exit code once

ncruces:
So I was trying to reuse the trap code for returns as well, and it turns out that that's simple enough.

Unfortunately, much to my disappointment, I figured #1505 is invalid, and probably needs to be reverted.

The issue is that the below cannot be reused if we want the line number in the stack trace:
https://github.com/tetratelabs/wazero/blob/e3755ce4af832a89e8557a5607bab2ab5df496bc/internal/engine/compiler/impl_amd64.go#L4882-L4887

I've tried to capture the address before the jump, but limited registers make it hard, and I'm running out of ideas.

It's unfortunate because this basically amounts to the entire 10 to 20% gain I'd got in #1505: traps are a significant chunk of compiled code, making the happy path leaner was the entire point.

If you have any suggestions I can try and work them out in this PR, otherwise I can open a revert PR.
Sorry for the wasted bandwidth.

### DIFF
--- a/internal/integration_test/spectest/v1/spec_test.go
+++ b/internal/integration_test/spectest/v1/spec_test.go
@@ -17,6 +17,13 @@ func TestCompiler(t *testing.T) {
 	spectest.Run(t, Testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(api.CoreFeaturesV1))
 }
 
+func TestCompilerNoDebug(t *testing.T) {
+	if !platform.CompilerSupported() {
+		t.Skip()
+	}
+	spectest.Run(t, Testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(api.CoreFeaturesV1).WithDebugInfoEnabled(false))
+}
+
 func TestInterpreter(t *testing.T) {
 	spectest.Run(t, Testcases, context.Background(), wazero.NewRuntimeConfigInterpreter().WithCoreFeatures(api.CoreFeaturesV1))
 }
--- a/internal/integration_test/spectest/v1/spec_test.go
+++ b/internal/integration_test/spectest/v1/spec_test.go
@@ -17,13 +17,6 @@ func TestCompiler(t *testing.T) {
 	spectest.Run(t, Testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(api.CoreFeaturesV1))
 }
 
-func TestCompilerNoDebug(t *testing.T) {
-	if !platform.CompilerSupported() {
-		t.Skip()
-	}
-	spectest.Run(t, Testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(api.CoreFeaturesV1).WithDebugInfoEnabled(false))
-}
-
 func TestInterpreter(t *testing.T) {
 	spectest.Run(t, Testcases, context.Background(), wazero.NewRuntimeConfigInterpreter().WithCoreFeatures(api.CoreFeaturesV1))
 }
--- a/internal/integration_test/spectest/v2/spec_test.go
+++ b/internal/integration_test/spectest/v2/spec_test.go
@@ -24,13 +24,6 @@ func TestCompiler(t *testing.T) {
 	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(enabledFeatures))
 }
 
-func TestCompilerNoDebug(t *testing.T) {
-	if !platform.CompilerSupported() {
-		t.Skip()
-	}
-	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(enabledFeatures).WithDebugInfoEnabled(false))
-}
-
 func TestInterpreter(t *testing.T) {
 	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigInterpreter().WithCoreFeatures(enabledFeatures))
 }
--- a/internal/integration_test/spectest/v2/spec_test.go
+++ b/internal/integration_test/spectest/v2/spec_test.go
@@ -24,6 +24,13 @@ func TestCompiler(t *testing.T) {
 	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(enabledFeatures))
 }
 
+func TestCompilerNoDebug(t *testing.T) {
+	if !platform.CompilerSupported() {
+		t.Skip()
+	}
+	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(enabledFeatures).WithDebugInfoEnabled(false))
+}
+
 func TestInterpreter(t *testing.T) {
 	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigInterpreter().WithCoreFeatures(enabledFeatures))
 }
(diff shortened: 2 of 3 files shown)


ncruces:
Maybe the correct decision is to only reuse trap sections if `WithDebugInfoEnabled`?

mathetake:
> Unfortunately, much to my disappointment, I figured https://github.com/tetratelabs/wazero/pull/1505 is invalid, and probably needs to be reverted. 

do you happen to have any repro test case to prove that? I think regardless of reverting #1505 or not, we should add that to out test suit

mathetake:
fyi @evacchi is working on the arm64 porting of #1505, so maybe he has some thoughts on this

achille-roussel:
Do you have a sense of whether the runtime gains could be maintained if the traps were all collocated at the end of the function?

Basically, keep one trap per location so there's a PC that can be mapped to each source instruction, but we put them out of the critical code paths.

The hypothesis would be that most of the throughput gains came from having more efficient use of caches because more of the _useful_ code is collocated and not interleaved with sections that never get executed.

I'm not sure how to prove this tho, just food for thoughts!

evacchi:
heh, and I just got started. Let me see if I can figure something out while I am looking into the port 🤔

ncruces:
> do you happen to have any repro test case to prove that? I think regardless of reverting #1505 or not, we should add that to out test suit

No, but it should be simple to create one: a `tiny.go` file with 2 panics; first one is conditionally skipped, second one traps.

ncruces:
> Do you have a sense of whether the runtime gains could be maintained if the traps were all collocated at the end of the function?
> 
> Basically, keep one trap per location so there's a PC that can be mapped to each source instruction, but we put them out of the critical code paths.
> 
> The hypothesis would be that most of the throughput gains came from having more efficient use of caches because more of the _useful_ code is collocated and not interleaved with sections that never get executed.
> 
> I'm not sure how to prove this tho, just food for thoughts!

Not sure (this is my first compiler), but the current assembler seems to make this hard to do (the `pc` is captured at runtime with `CompileReadInstructionAddress`) and I think it's wise to avoid structural changes to the assembler at this point?

That said, having the traps packed at the end of the function makes intuitive sense.

ncruces:
My proposal would be:
- keep ba843f99ea3b01f57f903ac136b639412685413b, this is innocuous for returns
- do nothing about calls back into Go (they stay how they've always been)
- partially revert #1505 for traps, disable the optimization when `WithDebugInfoEnabled` (by default)
- rename things according to the new status quo (this is no longer about traps but returns, etc)
- add tests to verify and check benchmarks again
- some tests need to run twice now

The rational is, if there's no debug info for line numbers, it's legal to merge these, otherwise it's not. We already have a config option for this, let's use it. Note though that this is only about line numbers. The stack trace is there, with function names, only the line number might be wrong.

More complex changes should be left to the new compiler, which should be able to eliminate _some_ traps (repeated bounds checks, division by a constant, etc), but which needs to keep in mind that the happy path needs love, because correct code basically never traps.

ncruces:
This `benchstat` this PR vs 1.2.0:
```
goos: darwin
goarch: amd64
               │ before.txt  │             debug.txt              │             nodebug.txt             │
               │   sec/op    │   sec/op     vs base               │   sec/op     vs base                │
_speedtest1-12   89.54m ± 1%   87.34m ± 1%  -2.45% (p=0.000 n=10)   76.91m ± 2%  -14.11% (p=0.000 n=10)
```

evacchi:
Let me throw an idea out there, but I may be missing some details and/or misinterpreted the code: consider I am new to generating asm... anyway...

Would it make sense to move the logic to store `ce.returnAddress` out of `compileExitFromNativeCode()` into its own method?

https://github.com/tetratelabs/wazero/blob/e3755ce4af832a89e8557a5607bab2ab5df496bc/internal/engine/compiler/impl_arm64.go#L412-L417
 
This method, say `compileStoreReturnAddress()` would have to be invoked explicitly before calling `compileExitFromNativeCode()` and/or(?) `compileTrapFromNativeCode()`.

- `compileTrap()` may emit the instructions or re-use the no-op to jump into the right address, 
- regardless `ce.returnAddress` has been prepared _before_ entering `compileExitFromNativeCode()` 
- thus, `ce.returnAddress` contains the address of the original trapping instruction.

does this make any sense?

ncruces:
We can factor it out, for sure.

One reason I think it's inlined there (and I kept it there) is for **readablity**:
`CompileReadInstructionAddress` reads the address after `beforeAcquisitionTargetInstruction`; in your snippet, the instruction _after_ `STRD`. It helps that you can visualize where `STRD` is, to figure out what instruction you're referring to, and to make sure there **is** an instruction _after_ `STRD`.

WRT factoring it out **and** calling it before `compileExitFromNativeCode`.

To get the full benefit from #1505 you need to not only reuse the code for traps, but also make the vast majority of traps a single conditional jump into the trap, with everything else (including saving the PC) staying out of the happy path. So I don't see where we could call `compileStoreReturnAddress`?

Ideally, as @achille-roussel pointed out, the first trap would also be outside of the happy path. But there are a lot of traps (20% of generated code for SQLite was bounds checking trap handling), and making 90% of them a single never-taken jump likely gives you 90% of the benefit.

evacchi:
> with everything else (including saving the PC) staying out of the happy path. So I don't see where we could call `compileStoreReturnAddress`?

great point. The change I was proposing would require always saving the PC unconditionally 👎

internal/integration_test/spectest/v2/spec_test.go
```diff
@@ -24,6 +24,13 @@ func TestCompiler(t *testing.T) {
 	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(enabledFeatures))
 }
 
+func TestCompilerNoDebug(t *testing.T) {
+	if !platform.CompilerSupported() {
+		t.Skip()
+	}
+	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(enabledFeatures).WithDebugInfoEnabled(false))
+}
+
```
mathetake, line 33: I guess spectest doesn't have debuginfo (dwarf) since it's coming from wat, not a real language, so I think we do not need to add this tests

```suggestion
```

internal/integration_test/spectest/v1/spec_test.go
```diff
@@ -17,6 +17,13 @@ func TestCompiler(t *testing.T) {
 	spectest.Run(t, Testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(api.CoreFeaturesV1))
 }
 
+func TestCompilerNoDebug(t *testing.T) {
+	if !platform.CompilerSupported() {
+		t.Skip()
+	}
+	spectest.Run(t, Testcases, context.Background(), wazero.NewRuntimeConfigCompiler().WithCoreFeatures(api.CoreFeaturesV1).WithDebugInfoEnabled(false))
+}
```
mathetake, line 25: ditto 
```suggestion
```

mathetake (approved):
do you want to make change to 
https://github.com/tetratelabs/wazero/blob/main/internal/testing/dwarftestdata/testdata/tinygo/main.go so that it would come with conditional? 

I think you can use `func trap()` to directly inline `unreachable` instruction  in a conditional branch. See https://github.com/tinygo-org/tinygo/blob/release/src/runtime/panic.go#L8-L12

internal/integration_test/spectest/v2/spec_test.go
ncruces: You're right, this is not the way.

I found these useful while developing the feature and wanted them to cover both cases.

Will remove.

ncruces:
I don't have the TinyGo tool chain installed, and pretty strapped for time today.

So it's mostly a matter of this sits for today, or it goes in and I add the test in a separate PR in a couple days.

ncruces:
If there's no change in consensus, I'll merge this today to unblock #1516, and change the TinyGo test in a couple of days.

codefromthecrypt (approved):
fire away!

PR #1517, opened 2023-06-13
wasi: add nonblock_test.go from gotip, fix nonblock read on Unix-like

evacchi:
Adapt and add the test from [Go's WASI test suite nonblock.go](https://github.com/golang/go/blob/0fcc70ecd56e3b5c214ddaee4065ea1139ae16b5/src/runtime/internal/wasitest/testdata/nonblock.go
) for non-blocking I/O (pipes) to our own suite. Go (almost) straight to the syscall layer for `read()` on supported platforms; i.e., this is currently Unix-only. This also adds the field `fd` to the struct `osFile` because apparently `f.file.Fd()` resets the `O_NONBLOCK` flag on the fd /o\.

### DIFF
--- a/internal/sysfs/file_unix.go
+++ b/internal/sysfs/file_unix.go
@@ -1,0 +1,21 @@
+//go:build unix || darwin || linux
+
+package sysfs
+
+import (
+	"syscall"
+
+	"github.com/tetratelabs/wazero/internal/platform"
+)
+
+const NonBlockingFileIoSupported = true
+
+// readFd exposes syscall.Read.
+func readFd(fd uintptr, buf []byte) (int, syscall.Errno) {
+	if len(buf) == 0 {
+		return 0, 0 // Short-circuit 0-len reads.
+	}
+	n, err := syscall.Read(int(fd), buf)
+	errno := platform.UnwrapOSError(err)
+	return n, errno
+}
--- a/internal/sysfs/file_unsupported.go
+++ b/internal/sysfs/file_unsupported.go
@@ -1,0 +1,12 @@
+//go:build !unix && !linux && !darwin
+
+package sysfs
+
+import "syscall"
+
+const NonBlockingFileIoSupported = false
+
+// readFd returns ENOSYS on unsupported platforms.
+func readFd(fd uintptr, buf []byte) (int, syscall.Errno) {
+	return -1, syscall.ENOSYS
+}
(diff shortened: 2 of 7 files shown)


achille-roussel (approved):
👏

evacchi:
ew, I guess this is a bad case of WorksOnMyMachine™️

evacchi:
weird, on Linux it does set nonblock = true, but it still blocks on read??

evacchi:
OK I think I found a solution that is more solid; pretty much, again, like for sockets, going _almost_ straight to the syscall layers for platforms that we support for nonblocking I/O (i.e. Unix for now). Probably better to add some specific test case before finalizing this, but this should make the gotip test work everywhere.

evacchi:
ok apparently something I did (maybe the `f.fd` in place of `f.file.Fd()`) ? broke Windows. I guess it's a good time to call it a day :D

codefromthecrypt:
Thanks for the focused help on this stuff @evacchi!

evacchi:
added tiny test case for `readFd` (all platforms) and nonblocking `f.Read()` (non-Windows)

internal/sysfs/file_unix.go
```diff
@@ -0,0 +1,21 @@
+//go:build unix || darwin || linux
```
ncruces, line 1: nit: unix already includes darwin and linux, does it not?

internal/sysfs/file_unsupported.go
```diff
@@ -0,0 +1,12 @@
+//go:build !unix && !linux && !darwin
```
ncruces, line 1: nit: same

ncruces (approved):

internal/sysfs/file_unix.go
codefromthecrypt: as long as unix works in go 1.18 sounds good to remove it. I forget when it was added.

evacchi: interesting I was overly specific because I recalled sometimes just using `unix` didn't work. I just tried and it _does_ work, but if I just rely on the naming convention `file_unix.go` then it won't

evacchi:
the timeouts are back, let's try to revert the commit with the build flags and see if it's related

internal/sysfs/file_unix.go
evacchi: I think it _doesn't_ work with go 1.18? 🤔 they're all timing out 
 https://github.com/tetratelabs/wazero/actions/runs/5265550045/jobs/9518327188

evacchi: reverted the change as apparently this won't work yet on 1.18

ncruces: You're right: https://tip.golang.org/doc/go1.19#go-unix
🤦

codefromthecrypt (approved):
Thanks for fixing this!

PR #1528, opened 2023-06-19
compiler: test traps report correct line numbers

ncruces:
Had to add a `zig cc` target. Tried doing this with TinyGo, but most traps happen inside the runtime, so I couldn't get 2 traps in the same function to be incorrectly deduped. Maybe it's possible (I tried panics, div by zero, and some memory accesses), but I eventually gave up.

Validated that reverting #1515 (so, applying the optimization in all cases) fails this test.


evacchi:
fuzzing failed but I don't think it's related; I feel it's more likely it was c43a169a25c57b30890e32a3b3534c39fb623984 🤔

mathetake:
`fatal error: runtime: out of memory` in fuzzing error can happen frequently actually without passing the proper flag, so please ignore it for now

mathetake (approved):
super cool, thanks!

PR #1529, opened 2023-06-19
wasi: introduce fsapi.Readdir and update File.Readdir()

evacchi:
Follow up to https://github.com/tetratelabs/wazero/pull/1482#issuecomment-1554008113 and https://github.com/tetratelabs/wazero/pull/1483

This PR introduces the `fsapi.Readdir` interface and updates all the usages of `File.Readdir()` to be nullary and return `fsapi.Readdir` instead of `[]Dirent`. 

`Readdir` instances are detached from the originating `File` and have their own lifecycle. In fact, it is also **necessary** to invoke `Close` on each `Readdir` instance that is created. This is particularly important e.g. on Windows where releasing the resource is necessary to allow cleanup operations and/or concurrent accesses.

The `fsapi.Readdir` interface is an iterator over entries of a directory. 
- It provides a `Next()` method to advance the internal cursors 
- an `Offset()` method returning the current position. 
- `Rewind(offset) is used to seek back to a previous position. Implementations may disallow seeking back to a position if it's "too in the past" (e.g. if a buffer or a window is maintained).
- A method `Close()` that **must** be invoked to ensure the underlying open directory is released. 

Moreover, we export the following utilities, for testing and internal use:

- `sysfs.NewReaddir(... fsapi.Dirent)` allows to create an instance of an `fsapi.Readdir`. 
- `sysfs.NewConcatReaddir(fsapi.Readdir, fsapi.Readdir)` allows to concat two instances 
- `sysfs.ReaddirAll(fsapi.Readdir) []fsapi.Dirents` exhausts the contents of a `Readdir` and returns a slice of `fsapi.Dirent`

Other subtypes of `Readdir` are also available (`windowedReaddir` and `emptyReaddir`) but these are for internal use and are not exported.


evacchi:
memo. Files on Windows must be re-opened, but the flow looks slightly twisted at the moment. I'll look into it tomorrow

evacchi:
Situation improved on windows (still patchy code), but there is a reliable failure on cleanup of the temp dir on Windows+go1.20 😢

evacchi:
close to resolving, but still messy code. Tried to refactor broke it again 😬 , will look tomorrow. Pretty close now.

internal/fsapi/file.go
```diff
@@ -225,7 +225,7 @@ type File interface {
 	//   - For portability reasons, no error is returned at the end of the
 	//     directory, when the file is closed or removed while open.
 	//     See https://github.com/ziglang/zig/blob/0.10.1/lib/std/fs.zig#L635-L637
-	Readdir(n int) (dirents []Dirent, errno syscall.Errno)
+	Readdir() (dirs Readdir, errno syscall.Errno)
```
evacchi, line 228: the comment below ⬇️  probably can be deleted now 🤔

internal/sys/fs.go
```diff
@@ -56,209 +56,6 @@ type FileEntry struct {
 	File fsapi.File
 }
 
-const direntBufSize = 16
-
-// Readdir is the status of a prior fs.ReadDirFile call.
-type Readdir struct {
```
evacchi, line 62: this struct is now the `windowedReaddir`

internal/sys/fs.go
```diff
@@ -56,209 +56,6 @@ type FileEntry struct {
...
-	// cursor is the current position in the buffer.
-	cursor uint64
```
evacchi, line 64: the cursor is no longer needed in the new version, because the `window` is a `sliceReaddir` which keeps its own internal cursor.

internal/sys/fs.go
```diff
@@ -56,209 +56,6 @@ type FileEntry struct {
...
-	countRead uint64
-
-	// dirents is a fixed buffer of size direntBufSize. Notably,
-	// directory listing are not rewindable, so we keep entries around in case
-	// the caller mis-estimated their buffer and needs a few still cached.
-	//
-	// Note: This is wasi-specific and needs to be refactored.
-	// In wasi preview1, dot and dot-dot entries are required to exist, but the
-	// reverse is true for preview2. More importantly, preview2 holds separate
-	// stateful dir-entry-streams per file.
```
evacchi, line 86: the note no longer appears in `windowedReaddir` because it no longer applies: the same effect is obtained by decorating the `windowedReaddir` with a `concatReaddir`; this `concatReaddir` tacks a `sliceReaddir` with the dot entries before a `windowedReaddir` that fetches the directory listing (see `LookupReaddir`).

🤔 maybe better turn this into a comment to that method.

internal/fsapi/file.go
```diff
@@ -380,3 +380,46 @@ type File interface {
...
+	Cookie() uint64
+	// Rewind seeks the internal cursor to the state represented by the cookie.
+	// It returns a syscall.Errno if the cursor was reset and an I/O error occurred while trying to re-init.
+	Rewind(cookie int64) syscall.Errno
+	// Peek emits the current value.
+	// It returns syscall.ENOENT when there are no entries left in the directory.
+	Peek() (*Dirent, syscall.Errno)
+	// Advance advances the internal counters and indices to the next value.
+	// It also empties and refill the buffer with the next set of values when the internal cursor
+	// reaches the end of it.
```
evacchi, line 400: FIXME: IIRC it also returns ENOENT when it can no longer advance. Must check

internal/sys/fs.go
```diff
@@ -363,6 +167,32 @@ func (c *FSContext) LookupReaddir(idx int32, f *FileEntry) (*Readdir, syscall.Er
 	}
 }
 
+// dotDirents returns "." and "..", where "." because wasi-testsuite does inode
+// validation.
+func dotDirents(f *FileEntry) ([]fsapi.Dirent, syscall.Errno) {
```
evacchi, line 172: function restored from a previous revision 34324031cb6d1f4b96ff706c72acd054fc70d6fd in wasi_snapshot_preview1/fs.go

internal/sys/fs_test.go
```diff
@@ -274,87 +274,3 @@ func TestFSContext_Renumber(t *testing.T) {
 		require.Equal(t, syscall.ENOTSUP, fsc.Renumber(3, 3))
 	})
 }
-
-func TestReaddDir_Rewind(t *testing.T) {
```
evacchi, line 278: moved to file_test.go

internal/sysfs/dir_test.go
```diff
@@ -78,20 +84,25 @@ func TestReaddir(t *testing.T) {
...
-				require.Equal(t, 1, len(dirents2))
 
 				// read exactly the last entry
-				dirents3, errno := dirF.Readdir(1)
+				errno = dirs.Advance()
+				require.EqualErrno(t, 0, errno)
+				dirent3, errno := dirs.Peek()
 				require.EqualErrno(t, 0, errno)
-				require.Equal(t, 1, len(dirents3))
+				// require.Equal(t, 1, len(dirents3))
```
evacchi, line 103: must delete

internal/sysfs/dir_test.go
```diff
@@ -134,9 +148,11 @@ func TestReaddir(t *testing.T) {
 		require.EqualErrno(t, 0, errno)
 		defer dirF.Close()
 
-		dirents, errno := dirF.Readdir(1)
+		dirs, errno := dirF.Readdir()
+		require.EqualErrno(t, 0, errno)
+		dirents, errno := fsapi.Collect(dirs)
 		require.EqualErrno(t, 0, errno)
-		require.Equal(t, 1, len(dirents))
+		_ = dirents
```
evacchi, line 155: fixme

evacchi:
left a few notes for review (and also to myself). Tomorrow I'll do the last round of cleanup.

internal/fsapi/file.go
evacchi: deleted and moved partially to the top of the `Readdir` interface

evacchi:
it should be good to go now

internal/fsapi/file.go
```diff
@@ -380,3 +375,59 @@ type File interface {
 	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/close.html
 	Close() syscall.Errno
 }
+
+// Readdir is the status of a prior fs.ReadDirFile call.
+//
+// TODO:
+//   - Verify if it is possible to reduce the need to reopen the whole file.
+//   - Verify if it is useful to expose a Close() method.
```
codefromthecrypt, line 383: This feels like you should do it as adding io.Closer later is dodgy. It seems unlikely we'll not end in a position of needing to close an underlying fd, right?

internal/fsapi/file.go
```diff
@@ -380,3 +375,59 @@ type File interface {
 	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/close.html
 	Close() syscall.Errno
 }
+
+// Readdir is the status of a prior fs.ReadDirFile call.
```
codefromthecrypt, line 379: Seems this is out-of date? My thought would be that this is the result of readdir, not a status, which is effectively a cursor over a directory.

codefromthecrypt (commented):
Thanks for getting this working. Made some polishing notes to help get things into similar shape to other things. It feels like the functionality is here, which is the most important part!

My comments are mostly bikeshed. In general (and I know I'm not perfect at this), I try to re-use names in go stdlib, also if possible to use the same style of text in comments when in doubt. I think Peek/Next are in use and you can also look at types like `syntax.scanner` and `scanner.Scanner` for ideas on how docs are written for a baseline and go from there (usually with notes about why, like this is for WASI, or only used by GOOS=js and might be removed when we remove that).

Also, you'll notice on public interfaces, we make a particular style so that we can relate to how things are used. Doing the same as existing code prevents inconsistency or the need for follow-up commits to do the same thing. While this is internal, we do want to eventually make things external.

So, try to follow forms like this.

```
	// SetNonblock toggles the non-blocking mode (O_NONBLOCK) of this file.
	//
	// # Errors
	//
	// A zero syscall.Errno is success. The below are expected otherwise:
	//   - syscall.ENOSYS: the implementation does not support this function.
	//   - syscall.EBADF: the file or directory was closed.
	//
	// # Notes
	//
	//   - This is like syscall.SetNonblock and `fcntl` with O_NONBLOCK in
	//     POSIX. See https://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html
```

internal/sysfs/file.go
```diff
@@ -432,3 +441,357 @@ func pwrite(w io.WriterAt, buf []byte, off int64) (n int, errno syscall.Errno) {
...
+// sliceReaddir implements fsapi.Readdir
+//
+// sliceReaddir iterates a given slice of fsapi.Dirent
+type sliceReaddir struct {
+	// cursor is the current position in the buffer.
+	cursor  uint64
+	dirents []fsapi.Dirent
+}
+
+// Reset implements the method of the same name in fsapi.Readdir.
```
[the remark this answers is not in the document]
evacchi: ha, I would have sworn we were using both phrasings.

internal/fsapi/file.go
evacchi: the thing is, the underlying FD is owned by the file you open, it is not created contextually to the Readdir instance. i.e., you do 

```
f, _ := sysfs.OpenFile()
rd := f.Readdir()
```

so the `Close()` method would only serve the purpose of "closing" the Readdir instance, which at this time I feel is kinda redundant. We can still add a `Close()` method that currently won't do anything in case we need it in further refactoring maybe?

codefromthecrypt:
strange..  I thought in certain cases of rewind we had to reopen the file. also it seemed internally directory reads dupe the fd. I guess I misremembered?

On Fri, Jun 23, 2023, at 3:12 PM, Edoardo Vacchi wrote:
> 
> 
> ***@***.**** commented on this pull request.
> 
> 
> In internal/fsapi/file.go <https://github.com/tetratelabs/wazero/pull/1529#discussion_r1239425993>:
> 
> > @@ -380,3 +375,59 @@ type File interface {
>  	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/close.html
>  	Close() syscall.Errno
>  }
> +
> +// Readdir is the status of a prior fs.ReadDirFile call.
> +//
> +// TODO:
> +//   - Verify if it is possible to reduce the need to reopen the whole file.
> +//   - Verify if it is useful to expose a Close() method.
> 
> the thing is, the underlying FD is owned by the file you open, it is not created contextually to the Readdir instance. i.e., you do
> 
> `f, _ := sysfs.OpenFile()
> rd := f.Readdir()
`

codefromthecrypt:
seems almost impossible to implement multiple calls to readdir without opening the file again.. maybe what's happening is this isnt implemented or tested.. so only one readdir instance can be created at a time?

On Fri, Jun 23, 2023, at 3:25 PM, Adrian Cole wrote:
> strange..  I thought in certain cases of rewind we had to reopen the file. also it seemed internally directory reads dupe the fd. I guess I misremembered?
> 
> On Fri, Jun 23, 2023, at 3:12 PM, Edoardo Vacchi wrote:
>> 
>> 
>> ***@***.**** commented on this pull request.
>> 
>> 
>> In internal/fsapi/file.go <https://github.com/tetratelabs/wazero/pull/1529#discussion_r1239425993>:
>> 
>> > @@ -380,3 +375,59 @@ type File interface {
>>  	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/close.html
>>  	Close() syscall.Errno
>>  }
>> +
>> +// Readdir is the status of a prior fs.ReadDirFile call.
>> +//
>> +// TODO:
>> +//   - Verify if it is possible to reduce the need to reopen the whole file.
>> +//   - Verify if it is useful to expose a Close() method.
>> 
>> the thing is, the underlying FD is owned by the file you open, it is not created contextually to the Readdir instance. i.e., you do
>> 
>> `f, _ := sysfs.OpenFile()
>> rd := f.Readdir()
`

codefromthecrypt:
I think what's happening is that the lifecycle of the Readdir instance is tied to the file that created it. So closing the file releases the resources of the cursor and there is no explicit way to do otherwise.

so basically we need to replace the TODO with a godoc on the interface saying something like this. then wasip>1 I think we may need to refactor as it allows more than one directory stream per FD. to implement that we would need to make this closeable.

internal/fsapi/file.go
```diff
@@ -380,3 +375,59 @@ type File interface {
...
+//   - Verify if it is useful to expose a Close() method.
+type Readdir interface {
+	// Reset seeks the internal cursor to 0 and refills the buffer.
+	Reset() syscall.Errno
+
+	// Skip is equivalent to calling n times Advance.
+	Skip(n uint64)
+
+	// Cookie returns a cookie representing the current state of the ReadDir struct.
+	Cookie() uint64
```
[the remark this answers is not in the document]
evacchi: for WASI the cookie value is supposed to be opaque, but indeed we are using it as a cursor offset.

codefromthecrypt:
right so we shouldn't bleed wasi specific into this interface imho. it should bias to posix, syscall and go first. of course it should be able to implement wasi ;)

internal/gojs/fs.go
```diff
@@ -325,10 +325,18 @@ func syscallReaddir(_ context.Context, mod api.Module, name string) (*objectArra
...
 
-	if dirents, errno := f.Readdir(-1); errno != 0 {
+	dirs, errno := f.Readdir()
+	if errno != 0 {
 		return nil, errno
-	} else {
-		entries := make([]interface{}, 0, len(dirents))
+	}
+
+	dirents, errno := fsapi.Collect(dirs)
```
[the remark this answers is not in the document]
evacchi: currently, it is also being used in `func (d *openRootDir) readdir() syscall.Errno`, not necessarily because it is right, but because the logic there seemed a bit articulated and I didn't want to touch too many places at once.

I can already update the tests to use a simplified version that takes the file instead, though.

internal/sysfs/file.go
```diff
@@ -432,3 +441,357 @@ func pwrite(w io.WriterAt, buf []byte, off int64) (n int, errno syscall.Errno) {
...
+func (e emptyReaddir) Rewind(int64) syscall.Errno { return 0 }
+
+// Peek implements the method of the same name in fsapi.Readdir.
+func (e emptyReaddir) Peek() (*fsapi.Dirent, syscall.Errno) { return nil, syscall.ENOENT }
+
+// Advance implements the method of the same name in fsapi.Readdir.
+func (e emptyReaddir) Advance() syscall.Errno { return syscall.ENOENT }
+
+// NewReaddirFromSlice is a constructor for fsapi.Readdir that only takes a []fsapi.Dirent.
+func NewReaddirFromSlice(dirents []fsapi.Dirent) fsapi.Readdir {
```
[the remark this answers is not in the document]
evacchi: are you sure about the name change?

internal/sysfs/file.go
```diff
@@ -432,3 +441,357 @@ func pwrite(w io.WriterAt, buf []byte, off int64) (n int, errno syscall.Errno) {
...
+	return &sliceReaddir{dirents: dirents}
+}
+
+// sliceReaddir implements fsapi.Readdir
```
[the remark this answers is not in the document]
evacchi: does that comment become redundant then?

codefromthecrypt:
I dont think it is typical to use the word slice in a function name. if you feel strongly about it we can revisit in a later PR

evacchi:
> strange..  I thought in certain cases of rewind we had to reopen the file. 

You remember right. `Fseek()` does `reopen()` the file in certain cases: basically on error; and this error appears to occur on Windows + go1.18 (if I understood correctly), where Fseek'ing directory won't work. I think this is also the main reason why it took me a while to get this working :D 

> also it seemed internally directory reads dupe the fd. I guess I misremembered?

I think there are some cases where Fd() dups but I found that only in sockets, [with Windows being LOLJK](https://github.com/golang/go/blob/25e46693a1085f604a9c92ee26afed2b4d45a9dd/src/net/fd_windows.go#L202-L205)

```go
func (fd *netFD) dup() (*os.File, error) {
	// TODO: Implement this
	return nil, syscall.EWINDOWS
}
```

but I didn't investigate for dirs; from a cursory glance, it looks like when the directory is opened with File.ReadDir() it is indeed re-opened from path, but its lifecycle is still tied to the owning File instance.

It is a good idea to untie those two, though. In fact, if we make Readdir unrelated from File probably we solve a lot of issues 😅 . I'll have to investigate if it's enough to call `OpenFile()` (my main concern: I am not sure we have the `path` string at hand at all times...)

(*) probably we "only" really need to do it for `osFile`, `fsFile` and `windowsOsFile`

evacchi:
> I dont think it is typical to use the word slice in a function name. if you feel strongly about it we can revisit in a later PR

no I was mostly hinting at that not being a generic `NewReaddir` but being a `Readdir` of a specific kind: if we use `NewReaddir()` as a name once we cannot really overload it with different arguments. It could be `NewReaddirWithDirents()` or something like that.

codefromthecrypt:
> no I was mostly hinting at that not being a generic NewReaddir but being a Readdir of a specific kind: if we use NewReaddir() as a name once we cannot really overload it with different arguments. It could be NewReaddirWithDirents() or something like that.

my 2p is I think we don't always have to repeat the type of parameters in the function name. If there's no obvious neat name (`NewReaddirWithDirents` is a mouthful) and nothing using the simplest name, and no plan for something to use that, go for it ;)

codefromthecrypt:
food for thought from the crypt

often if I encounter a type that's too strange to name publicly, I try to avoid naming it publicly. For example, there may not be more than one call site that needs to make a directory out of a fixed list of dirents. If that's the case here (I haven't audited), then you get the uber naming trick, which is to not actually export a name. Rarely, but it has happened, I'll even repeat a simple type twice because it is too awkward to name publicly, or otherwise something I wouldn't ever want to commit publicly.

internal/fsapi/file.go
evacchi: I have experimented today with actually cloning the File instance https://github.com/tetratelabs/wazero/pull/1529#issuecomment-1603951828 and I ended up again in the Windows/go1.18 world of pain. So I think for the time being I am going to add and document the Close method, but I am going to refrain from actually adding a meaningful implementation (will just return 0 everywhere), but I will do a follow-up PR with a proper implementation

the main issue is that if we do clone the FD, it is extremely important that the file is closed properly, otherwise the tests on Windows will often fail with a "TempDir RemoveAll cleanup: remove C:\Some\Dir: The process cannot access the file because it is being used by another process."

evacchi:
ok, it should be better now. I have moved back to draft for now because I wanted to look again at docs etc. Notably, now the Readdir instance is properly detached from the original file; however, now extra attention is required on Windows: where you really need to close; this shows up in tests because the cleanup routine errors out (maybe we can obviate by opening with FILE_SHARE_DELETE?)

evacchi:
some further cleanup done, improved godocs, unmarked from draft :)

internal/fsapi/file.go
codefromthecrypt: ps remember Go 1.18 will fall off our support list in a couple months when go 1.21 is out!

internal/fsapi/file.go
```diff
@@ -380,3 +375,158 @@ type File interface {
...
+	//   - This is conceptually similar to invoking Seek(0, io.SeekStart) on io.Seeker
+	//     and `fseek` in POSIX over an open directory. See
+	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/fseek.html
+	//   - However implementations will generally reopen the underlying directory
+	//     for portability reasons.
+	Reset() syscall.Errno
+
+	// Rewind seeks the internal cursor to the given offset.
+	//
+	// Rewinding to offset 0 usually should produce the same result
```
codefromthecrypt, line 422: I would love to have only one function. Can you give me the summary of what would happen if we deleted `Reset`?

internal/fsapi/file.go
```diff
@@ -380,3 +375,158 @@ type File interface {
...
+	// # Notes
+	//
+	//   - Ultimately, implementations may fetch data using a `readdir`-like API, thus
+	//     errors will reflect file system errors similarly to the POSIX call. See
+	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/readdir.html
+	Peek() (*Dirent, syscall.Errno)
+
+	// Next advances the internal cursor to the next value.
+	//
+	// This is equivalent to invoking Peek() and then advancing the internal
```
codefromthecrypt, line 464: I think this is backwards. The normal case of things like this is Next with Peek being the odd one. Can you either remove this note.

Then, simplify peek to say that this is like Next, except it doesn't advance the internal cursor? (on that case don't repeat the Errors and Notes section on Peek because the docs will say look at Next for details.

internal/fsapi/file.go
```diff
@@ -380,3 +375,158 @@ type File interface {
...
+	//     errors will reflect file system errors similarly to the POSIX call. See
+	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/readdir.html
+	Next() (*Dirent, syscall.Errno)
+
+	// Close closes the underlying file.
+	//
+	// # Errors
+	//
+	// A zero syscall.Errno is success. The below are expected otherwise:
+	//   - syscall.ENOSYS: the implementation does not support this function.
```
codefromthecrypt, line 486: Sorry our docs on fsapi.File Close are wrong. We shouldn't (and don't) return an error on not implemented close. Can you correct that, also?

`A zero syscall.Errno is returned if unimplemented or success.`

internal/fsapi/unimplemented.go
```diff
@@ -152,7 +152,7 @@ func (UnimplementedFile) Seek(int64, int) (int64, syscall.Errno) {
 }
 
 // Readdir implements File.Readdir
-func (UnimplementedFile) Readdir(int) (dirents []Dirent, errno syscall.Errno) {
+func (UnimplementedFile) Readdir() (dirs Readdir, errno syscall.Errno) {
```
codefromthecrypt, line 155:
```suggestion
func (UnimplementedFile) Readdir() (Readdir, syscall.Errno) {
```

codefromthecrypt (approved):
Looking good. Here are some final round notes:

* seems the PR description is out of date. you only need to talk about new exported symbols.
* I would love to be able to delete Reset, so asked if we can for Rewind(0).
* I made a mistake documenting fsapi.File Close which I think was propagated here. Let's fix both here, if that's ok (don't return ENOSYS).

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -905,7 +908,7 @@ const largestDirent = int64(math.MaxUint32 - wasip1.DirentSize)
 //
 // See https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#fd_readdir
 // See https://github.com/WebAssembly/wasi-libc/blob/659ff414560721b1660a19685110e484a081c3d4/libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c#L44
-func maxDirents(dir *sys.Readdir, bufLen uint32) (dirents []fsapi.Dirent, bufused, direntCount uint32, writeTruncatedEntry bool) {
+func maxDirents(dir fsapi.Readdir, bufLen uint32) (dirents []fsapi.Dirent, bufused, direntCount uint32, writeTruncatedEntry bool) {
```
codefromthecrypt, line 911: It appears that the only non-test usage of Peek is here. I haven't checked carefully, but I'm wondering if we can use Rewind in the case of not enough space vs Peek/advance always (plus needing to define and expose Peek)

internal/fsapi/file.go
evacchi: yeah that thought kept me sane 🤣

internal/fsapi/file.go
evacchi: just a useless check for offset == 0 every time you call it, but it's ok. Will do!

internal/fsapi/file.go
evacchi: TBH this was wrong too, it actually returns 0 also here.

imports/wasi_snapshot_preview1/fs.go
evacchi: so my immediate reaction was "no" because if you are at the end of a window, Next() will discard that window and you will lose access to the last item; however, I may rework the window so that it actually slides and so that index 0 always contains the last item of the previous batch. index 1 would be the first item of the new batch. Extra care is needed when you are on batch #1 where index 0 would be meaningless.

codefromthecrypt: clever approach on making dirent -1 is always available internally for a rewind(pos-1)

I feel like when things move to a native approach, they'll use Seek and those impls won't need to hold state possibly, rather seek offset relative -1. That said, wishing things doesn't make them true ;)

internal/fsapi/file.go
```diff
@@ -371,12 +366,155 @@ type File interface {
...
+//     over a directory. Because reading in batches requires to hold a reference
+//     to an open directory, callers should pay extra care to make sure
+//     that such references are released, by invoking Close.
+//   - This is especially important on the Windows platform, where keeping
+//     a file handle open may prevent concurrent processes from accessing
+//     (e.g. deleting) the underlying directory. This may conflict with cleanup
+//     procedures (such as `testing/TB.Cleanup()`).
+type Readdir interface {
+	// Skip is equivalent to calling Next n times.
+	Skip(n uint64)
```
codefromthecrypt, line 397: locally I only see this used in tests. Can we delete anything only used in tests for something similar inlined?

(ack this is not something you added, rather moved)

internal/fsapi/file.go
```diff
@@ -371,12 +366,155 @@ type File interface {
...
+	//   - syscall.EBADF: the file or directory was closed.
 	//
 	// # Notes
 	//
 	//   - This is like syscall.Close and `close` in POSIX. See
 	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/close.html
 	Close() syscall.Errno
 }
+
+// ReaddirAll reads eagerly all the values returned by the given
```
codefromthecrypt, line 484: This is only used in tests and gojs which will be removed in less than a half year. Can you either inline it or move it to the sysfs package?

imports/wasi_snapshot_preview1/fs.go
evacchi: there is unfortunately another caveat, the way . and .. are now prepended is through another Readdir, so it's not as easy as moving the window... you have also to take into account the prepended Readdir...

codefromthecrypt: FWIW, back in the day I was thinking to make "bool addDotEntries" a parameter of the readDir function for this reason. Composition is difficult externally.

evacchi: I think I managed to make it work 😬

evacchi:
I tried to make `Next()` work without `Peek()` but things break in very subtle ways. 

The thing is, if we default to Next() we really need to be able to Rewind() the cursor in many cases; this becomes especially annoying if you are within a window and you want to read an element that comes before that. I thought I had something working (it passed wasi-testsuite too), but some tests for Zig and Go stdlib started failing. 

I think it's better to get this merged at this point and circle back to it again.

evacchi:
Alright, Peek is no more. I'll backfill some more finer-grained tests, but I think this is the best I can do for now 😬

codefromthecrypt:
I'll take over from here, thanks for all the hard work @evacchi!

codefromthecrypt:
TL;DR; I'm closing this as it isn't the correct path forward. What we had is more relevant as it is special cased to wasip1/wasix: wasi-filesystem preview 2 will not use it and we have no idea yet the final form of preview 3.

I'll put the below rationale in a future PR and also try to land cleanups discovered over the last week, thinking hard (again) about the issue of readdir.

---

So this issue led me to a lot of soul searching on if this is the right path to go. I suggested it, but I believe it is wrong. In short there are problems in making this type, even if we made it more POSIX like (using words like tell, seek, etc.). What's wrong about it is that this isn't relevant for most usage.

* `fd_readdir` in wasip1 (and so also wasix) is like `getdents` in Linux, not `readdir` in POSIX. `getdents` is more like the `Readdir` we had (based on Go's `os.File.Readdir`).
* `directory-entry-stream` in wasi-filesystem preview2 is defined in [component model](https://github.com/WebAssembly/wasi-filesystem/blob/ef9fc87c07323a6827632edeb6a7388b31266c8e/example-world.md#directory_entry_stream), not an ABI, but in wasmtime it is a [consuming iterator](https://github.com/bytecodealliance/wasmtime/blob/b741f7c79d72492d17ab8a29c8ffe4687715938e/crates/wasi/src/preview2/preview2/filesystem.rs#L286-L296). A consuming iterator is easy to support with anything (like `Readdir(1)`), even if it is inefficient as you cannot bulk read, nor skip.
  * The implementation of the preview1 adapter (uses preview2) confirms this. They use a [dirent cache](https://github.com/bytecodealliance/preview2-prototyping/blob/e4c04bcfbd11c42c27c28984948d501a3e168121/crates/wasi-preview1-component-adapter/src/lib.rs#L2131-L2137) (similar in concept to our cache prior to this change). As there is no seek concept, in preview2, they interpret the cookie as numeric and [read on repeat entries](https://github.com/bytecodealliance/preview2-prototyping/blob/e4c04bcfbd11c42c27c28984948d501a3e168121/crates/wasi-preview1-component-adapter/src/lib.rs#L936) when a cache wasn't available.
* `directory-entry-stream` is documented to [change significantly](https://github.com/WebAssembly/WASI/blob/ddfe3d1dda5d1473f37ecebc552ae20ce5fd319a/docs/WitInWasi.md#Streams) in wasi-filesystem preview3, based on [async streams](https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE/edit#slide=id.g1270ef7d5b6_0_662). As preview 2 isn't scheduled until at least fall 2023, a future design could land possibly in 2024. So, we can't decide yet how this will affect the ABI.

In summary, we should clean up our prior design and special-case the dirent cache for wasip1 as we did before. A better use of energy would be to make portable host-module specific state, which could contain the dirent cache in a way that isn't used if a future wasp1 is defined (note gojs doesn't use it either!). Such state could also allow us to handle whatever changes preview3 may have next year or whenever it completes.

Issue #1534, opened 2023-06-23
Memory Leak when Recompiling Modules and Reinstantiating Runtimes


abraithwaite:
**Describe the bug**
We are recompiling modules frequently because our users may change their code.
We're also reinstantiating runtimes as a way to easily clear the module cache and start with a "clean slate" (we think/thought) for isolation purposes.
We're using the python interpreter guest.
We're closing all references to all wazero structs before reinstantiating things.
The memory continues to grow between each close/create cycle.

**To Reproduce**
Reproducable, just run main.go:
https://github.com/runreveal/wazero_memleak

**Expected behavior**
We expect memory to not grow when we're not holding any references to wazero objects any longer.

**Environment (please complete the relevant information):**
- Go version: 1.20.5
- wazero Version: v1.2.1
- Host architecture: amd64, arm64 (m2)
- Runtime mode: compiler mode

inkeliz:
I was able to reproduce that issue, and I run some PPROF and take a look in the code. So, my suspicious was: "some mmap is leaking" and that seems o be the case, but that was odd. Because Wazero uses `runtime.SetFinalizer`. 

However, the `releaseCompiledModule` is never called. So, I tried to hunt who is holding the reference. Then, I notice the following:

```
	compiledModule struct {
		executable        asm.CodeSegment
		functions         []compiledFunction
		source            *wasm.Module
		ensureTermination bool
	}
```

If you take a look at `compliledFunction`, you get:

```
compiledFunction struct {
		executableOffset uintptr
		stackPointerCeil uint64

		index           wasm.Index
		goFunc          interface{}
		listener        experimental.FunctionListener

		parent          *compiledModule                          ////  << HERE?!

		sourceOffsetMap sourceOffsetMap
	}
```

Each `compiledFunction` holds the pointer to `compiledModule`. So, I GUESS, that is the reason for the leak.

----------


I patch the wazero with that change: https://github.com/tetratelabs/wazero/pull/1535/commits/7367c2ea24b3127280bf7210be71cb972b55aa00


See if that helps. 👍

abraithwaite:
This is great!  Thank you!! 🙏

Testing with our contrived test seems to have a stable memory profile and runtime! 🎉

abraithwaite:
Grrr, fat fingered "enter"...

ANYWAY, I was going to say that we tried it in our environment and actually started getting panics for what seems like a related reference.

```
[...]

PR #1535, opened 2023-06-25
engine: mitigates memory leak

inkeliz:
Previouly, each compiledFuction retain one pointer of compiledModule.

### DIFF
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -496,6 +496,13 @@ func (e *engine) Close() (err error) {
 	e.mux.Lock()
 	defer e.mux.Unlock()
 	// Releasing the references to compiled codes including the memory-mapped machine codes.
+
+	for i := range e.codes {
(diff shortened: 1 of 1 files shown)


inkeliz:
That might not be the "perfect solution", I'm justing sending here to make easier to test.  :)

internal/engine/compiler/engine.go
```diff
@@ -496,6 +496,13 @@ func (e *engine) Close() (err error) {
 	e.mux.Lock()
 	defer e.mux.Unlock()
 	// Releasing the references to compiled codes including the memory-mapped machine codes.
+
+	for i := range e.codes {
```
mathetake, line 500: Would you mind adding tests?

mathetake (commented):
this change makes sense and actually fixes a bug introduced in #1377 which is a root cause to #1534. Thank @inkeliz  for tracking down!

codefromthecrypt (approved):
nice catch!

PR #1537, opened 2023-06-27
Adds failing test for cache closed while function that panics is in-flight

codefromthecrypt:
See #1534


codefromthecrypt:
closing as not really the issue,

PR #1542, opened 2023-06-28
wasi: improve stdin support for nonblocking, fix stdout

### LINKED ISSUE #1538: Non-blocking stdio isn't working correctly
**Describe the bug**

Non-blocking stdio isn't working correctly.

**To Reproduce**

Use [`gotip`](https://pkg.go.dev/golang.org/dl/gotip) that includes at least https://github.com/golang/go/commit/13529cc5f443ef4e242da3716daa0032aa8d34f2 (note that this will be part of the Go 1.21 release in ~August).

If I compile the following module:

```go
package main

import (
	"io"
	"os"
	"time"
)

func main() {
	go func() {
		time.Sleep(1 * time.Second)
		os.Stdout.WriteString("waiting for stdin...\n")
	}()

	b, err := io.ReadAll(os.Stdin)
	if err != nil {
		panic(err)
	}
	os.Stdout.Write(b)
	return
}
```

using:

```
$ GOOS=wasip1 GOARCH=wasm gotip build -o stdio.wasm stdio.go
```

and then run the script with wazero, it returns errno=28 (EINVAL) from `poll_oneoff` which causes Go to crash:

```
$ (sleep 3; echo foo) | wazero run stdio.wasm 
waiting for stdin...
errno= 28  len(pollsubs)= 1
fatal error: poll_oneoff failed

runtime stack:
runtime.throw({0x17c29, 0x12})
	/Users/chris/sdk/gotip/src/runtime/panic.go:1077 +0x3 fp=0xeb778 sp=0xeb750 pc=0x12250003
runtime.netpoll(0xffffffffffffffff)
	/Users/chris/sdk/gotip/src/runtime/netpoll_wasip1.go:218 +0x45 fp=0xeb7e8 sp=0xeb778 pc=0x11f30045
```

**Expected behavior**

[wasi-go](https://github.com/stealthrocket/wasi-go) does the right thing here:

```
$ (sleep 3; echo foo) | wasirun stdio.wasm                  
waiting for stdin...
foo
```

```
$ gotip version
go version devel go1.21-13529cc Tue Jun 27 01:09:18 2023 +0000 darwin/arm64
$ uname -a
Darwin MBP.local 22.5.0 Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:20 PDT 2023; root:xnu-8796.121.3~7/RELEASE_ARM64_T6000 arm64
$ wazero version
v1.2.1
```

evacchi:
- Allow nonblocking stdin to go through the "normal" file descriptor flow in `poll_oneoff`.
- Ensure stdio FDs are not closed on `SetAppend`: the default behavior for SetAppend() is to reopen, that is close and then open again: this causes an error on std streams. Just ignore the call on them.

Fixes #1538.

### DIFF
--- a/imports/wasi_snapshot_preview1/wasi_stdlib_test.go
+++ b/imports/wasi_snapshot_preview1/wasi_stdlib_test.go
@@ -481,7 +481,7 @@ func testStdin(t *testing.T, bin []byte) {
 	r, w, err := os.Pipe()
 	require.NoError(t, err)
 	moduleConfig := wazero.NewModuleConfig().
-		WithSysWalltime().WithSysNanotime(). // HTTP middleware uses both clocks
+		WithSysNanotime(). // poll_oneoff requires nanotime.
 		WithArgs("wasi", "stdin").
 		WithStdin(r).WithStdout(os.Stdout)
 	ch := make(chan string, 1)
--- a/imports/wasi_snapshot_preview1/wasi_stdlib_test.go
+++ b/imports/wasi_snapshot_preview1/wasi_stdlib_test.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"net"
 	"net/http"
+	"os"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


imports/wasi_snapshot_preview1/wasi_stdlib_test.go
```diff
@@ -457,3 +458,38 @@ func testHTTP(t *testing.T, bin []byte) {
...
+			testStdin(t, bin)
+		})
+	}
+}
+
+func testStdin(t *testing.T, bin []byte) {
+	r, w, err := os.Pipe()
+	require.NoError(t, err)
+	moduleConfig := wazero.NewModuleConfig().
+		WithSysWalltime().WithSysNanotime(). // HTTP middleware uses both clocks
```
achille-roussel, line 484: This comment doesn't seem relevant here, was it leftover from a copy/paste?
```suggestion
		WithSysWalltime().WithSysNanotime().
```

achille-roussel (approved):

imports/wasi_snapshot_preview1/wasi_stdlib_test.go
evacchi: leftover from copypaste 😬

codefromthecrypt:
nice work

PR #1548, opened 2023-06-30
Simplifies filesystem implementations

codefromthecrypt:
This performs a number of simplifying changes to files, particularly around directories. Some of them against my earlier advice to @evacchi!

* use POSIX errors only for readdir.
* stop using a separate subtype for windows osFile as the code is simpler with an if statement.
* stop using a separate table for directory stream descriptors. We aren't yet testing this works, so it is distracting. It was my bad advice to start doing this.

### DIFF
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/tetratelabs/wazero/internal/descriptor"
 	"github.com/tetratelabs/wazero/internal/fsapi"
-	"github.com/tetratelabs/wazero/internal/platform"
 	socketapi "github.com/tetratelabs/wazero/internal/sock"
 	"github.com/tetratelabs/wazero/internal/sysfs"
 )
@@ -53,6 +52,35 @@ type FileEntry struct {
 
 	// File is always non-nil.
 	File fsapi.File
+
+	// openDir is nil until OpenDir was called.
+	openDir *Readdir
+}
+
+// OpenDir lazy creates a directory stream for this file.
+//
+// # Parameters
+//
+// When `addDotEntries` is true, navigational entries "." and ".." precede
+// any other entries in the directory. Otherwise, they will be absent.
+//
+// # Errors
+//
+// # This returns the same errors as fsapi.File Readdir
+//
+// Notes:
+//   - In the future, this may be refactored to allow multiple open directory
+//     streams per file (likely in wasip>1).
+func (f *FileEntry) OpenDir(addDotEntries bool) (dir *Readdir, errno syscall.Errno) {
+	if dir = f.openDir; dir != nil {
+		return dir, 0
+	} else if dir, errno = newReaddirFromFileEntry(f, addDotEntries); errno != 0 {
+		// not a directory or error reading it.
+		return nil, errno
+	} else {
+		f.openDir = dir
+		return dir, 0
+	}
 }
 
 const direntBufSize = 16
@@ -134,11 +162,14 @@ func (d *Readdir) init() syscall.Errno {
 }
 
 // newReaddirFromFileEntry is a constructor for Readdir that takes a FileEntry to initialize.
-func newReaddirFromFileEntry(f *FileEntry) (*Readdir, syscall.Errno) {
-	// Generate the dotEntries only once and return it many times in the dirInit closure.
-	dotEntries, errno := synthesizeDotEntries(f)
-	if errno != 0 {
-		return nil, errno
+func newReaddirFromFileEntry(f *FileEntry, addDotEntries bool) (*Readdir, syscall.Errno) {
+	var dotEntries []fsapi.Dirent
+	if addDotEntries {
+		// Generate the dotEntries only once and return it many times in the dirInit closure.
+		var errno syscall.Errno
+		if dotEntries, errno = synthesizeDotEntries(f); errno != 0 {
+			return nil, errno
+		}
 	}
 	dirInit := func() ([]fsapi.Dirent, syscall.Errno) {
 		// Ensure we always rewind to the beginning when we re-init.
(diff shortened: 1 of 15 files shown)


internal/sysfs/osfile.go
```diff
@@ -4,15 +4,20 @@ import (
...
 )
 
-func newDefaultOsFile(openPath string, openFlag int, openPerm fs.FileMode, f *os.File) fsapi.File {
-	return &osFile{path: openPath, flag: openFlag, perm: openPerm, file: f, fd: f.Fd()}
+func newOsFile(openPath string, openFlag int, openPerm fs.FileMode, f *os.File) fsapi.File {
+	// Windows cannot read files written to a directory after it was opened.
+	// This was noticed in #1087 in zig tests. Use a flag instead of a
+	// different type.
+	reopenDir := runtime.GOOS == "windows"
+	return &osFile{path: openPath, flag: openFlag, perm: openPerm, reopenDir: reopenDir, file: f, fd: f.Fd()}
```
codefromthecrypt, line 20: notably this allows us to portably cast to *osFile if we ever need to get access to the underlying file. We couldn't before due to wrapping which was introduced for only one special case.

codefromthecrypt:
made a mistake. I will debug on my windows box and mark undraft after

codefromthecrypt:
oops zig test failed. I'll backfill one so we don't have to wait so long

```
1105/2509 test.Dir.Iterator but dir is deleted during iteration... reached unreachable codeerror instantiating wasm binary: module[] function[_start] failed: wasm error: unreachable
wasm stack trace:
	.os.abort()
		0x48ee: /home/runner/zig-source/lib/std/os.zig:611:32
	.builtin.default_panic(i32,i32,i32,i32)
		0x44c: /home/runner/zig-source/lib/std/builtin.zig:894:25
	.fs.IterableDir.Iterator__struct_94278.nextWasi(i32,i32)
		0xfc93ce: /home/runner/zig-source/lib/std/fs.zig:825:39
	.fs.IterableDir.Iterator__struct_94278.next(i32,i32)
		0xfc8a06: /home/runner/zig-source/lib/std/fs.zig:797:37
	.fs.test.test.Dir.Iterator but dir is deleted during iteration() i32
		0xfcd3f7: /home/runner/zig-source/lib/std/fs/test.zig:310:36
	.test_runner.mainTerminal()
		0x3e04: /home/runner/zig-source/lib/test_runner.zig:175:28
	.test_runner.main()
		0x396: /home/runner/zig-source/lib/test_runner.zig:35:28
	._start()
		0x1d: /home/runner/zig-source/lib/std/start.zig:599:22 (inlined)
		      /home/runner/zig-source/lib/std/start.zig:232:42
```

internal/sys/fs.go
```diff
@@ -336,77 +394,18 @@ func (c *FSContext) SockAccept(sockFD int32, nonblock bool) (int32, syscall.Errn
...
-}
-
 // CloseFile returns any error closing the existing file.
-func (c *FSContext) CloseFile(fd int32) syscall.Errno {
+func (c *FSContext) CloseFile(fd int32) (errno syscall.Errno) {
 	f, ok := c.openedFiles.Lookup(fd)
 	if !ok {
 		return syscall.EBADF
 	}
+	// Close the file
```
achille-roussel, line 403: Maybe remove this comment, it doesn't add much to what's already clear in the code
```suggestion
```

achille-roussel (commented):
I can't help but wonder if we are making the right trade off by prioritizing portability with the file reopen approach.

It might be a potential security vulnerability, the guest not knowing that a file was reopened and assuming that it can trust that open file handle while it actually was rerouted to a different path due to the directory structure having changed in between. It's a stretch but there have been security vulnerabilities exploiting weirder edge cases.

I'm not opposed to moving forward with this change tho, we already have this mechanism in place so we're not really introducing a new risk, but we're expanding the surface area a bit now that we also do this on directory reads.

internal/sysfs/open_file_windows.go
```diff
@@ -154,46 +148,3 @@ func open(path string, mode int, perm uint32) (fd syscall.Handle, err error) {
...
-		return errno
-	} else if !isDir {
-		return syscall.ENOTDIR
-	}
-
-	// On Windows, once the directory is opened, changes to the directory are
-	// not visible on ReadDir on that already-opened file handle.
-	//
-	// To provide consistent behavior with other platforms, we re-open it.
-	if errno := f.osFile.Close(); errno != 0 {
```
codefromthecrypt, line 189: @achille-roussel so this PR is moving this code and conditional to where it happens under an if statement instead of a subtype. So, really the change should be cosmetic, not actually behavioral. OTOH, it is good advice to contain this to windows, and I think we can!

internal/sysfs/osfile.go
```diff
@@ -4,15 +4,20 @@ import (
...
 	"github.com/tetratelabs/wazero/internal/platform"
 )
 
-func newDefaultOsFile(openPath string, openFlag int, openPerm fs.FileMode, f *os.File) fsapi.File {
-	return &osFile{path: openPath, flag: openFlag, perm: openPerm, file: f, fd: f.Fd()}
+func newOsFile(openPath string, openFlag int, openPerm fs.FileMode, f *os.File) fsapi.File {
+	// Windows cannot read files written to a directory after it was opened.
+	// This was noticed in #1087 in zig tests. Use a flag instead of a
+	// different type.
+	reopenDir := runtime.GOOS == "windows"
```
codefromthecrypt, line 19: so this state is only true when the OS is windows

codefromthecrypt:
self-note: the test I need to add instead of wait for zig to fail is readdir after close, like this..

https://github.com/ziglang/zig/blob/master/lib/std/fs/test.zig#L290-L317

```
==> wasi_snapshot_preview1.fd_readdir(fd=4,buf=1045884,buf_len=1024,cookie=0)
<== (bufused=,errno=EINVAL)
```

internal/sysfs/osfile.go
achille-roussel: Ah! I missed this, thanks for the correction 😄

internal/sysfs/open_file_windows.go
achille-roussel: I crossed context with discussions on Slack and misunderstood the goal of this change.

I'm glad there's no increase to the risk surface, and the scope of reopening on directory reads only on Windows is a more comfortable trade-off 👍

codefromthecrypt: ps your feedback dramatically affects #1529 which was on the path to always re-open. I'll make sure we don't do that, and probably it is better to just make changes like this so each part are visible.

PR #1552, opened 2023-07-02
configure golangci-lint

faddat:
## Context

Hi, I'm Jacob, and I'm researching ways to replace wasmer with wazero 
in cosmos-sdk blockchains that use https://github.com/cosmwasm/wasmd

Here's some info about the research work:

* https://github.com/CosmWasm/wasmd/pull/1475

* https://hackmd.io/18jrYUmQTdaEKP4J9q1Klg#

Usually when I am trying to get to know a new codebase, I lint it and 
try to otherwise improve it, so I can figure out what is going on 
under the hood.  

I've started up a twitter DM chat called "cgo considered harmful" if 
you'd like to join.  

https://twitter.com/gadikian

## work so far

* golangci-lint isn't configured in wazero, so it has been skipping 
tests.  
* I've enabled gci, which can replace gosimports
* I've made golangci-lint a part of the format command, and removed 
references to gofumpt in the makefile, since gofumpt is a part of 
golangci-lint.  
* make lint will work as it always did
* make format will also work as it always did, just we will now be 
able to control how the linter works, including having it run gofumpt 
and format imports for us.

- add golangci-lint configurations
- configure golangci-lint and update Makefile


codefromthecrypt:
thanks for looking closer at things! one thing to note is that `make check` runs lint. If there's an issue with lint, it should be caught by that. This also helps use a specific version declared in Makefile (for consistency). Can you look from that angle first? We are trying to avoid duplication between Makefile and GH actions.

codefromthecrypt:
hi, I'm going to close this as it is incompatible with how the maintainers work at the moment, as well our pre-check process `make check`. We intentionally define version info inside Makefile which this PR removes. Feel free to join gophers slack to discuss more and best luck on your porting. I hope to see you around! 

https://wazero.io/community/

Issue #1553, opened 2023-07-03
WithFS(fs.FS) directory read returns unexpected results


GeorgeMac:
> Nice to meet you all. I love this project 👋 

**Describe the bug**

I have found some obscure behaviour when I provide custom implementations of `fs.FS`.
I see inconsistent behaviour when compared with `os.DirFS` instance.
I can see they flow through very different paths in the codebase (I the see adaptor for `fs.FS`).
So I am not surprised there can be differences.

However, if I write a project in Go and compile with go1.21 wasip1 target, I get inconsistent results from reading directory entries when the FS added is not an `os.DirFS`. Only the root `.` entry gets returned.

I have created a minimal reproduction.

**To Reproduce**

Take the following Go code to be compiled to wasm using the `wasip1` target:

```go
package main

import (
	"fmt"
	"io/fs"
	"os"
)

func main() {
	fs.WalkDir(os.DirFS("."), ".", func(path string, d fs.DirEntry, err error) error {
		if err != nil {
			return err
		}

		fmt.Println(path, d)

		return nil
	})
}
```

I have this in a local directory called `cmd/walk/main.go` and compile it to `walk.wasm`:

```
GOOS=wasip1 GOARCH=wasm gotip build -o walk.wasm cmd/walk/main.go
```

Now take the following Go program using wazero to invoke it:

```go
package main

import (
	"context"
	"fmt"
	"io/fs"
	"log"
	"os"

	"github.com/tetratelabs/wazero"
	"github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1"
	"github.com/tetratelabs/wazero/sys"
)

func main() {
	ctx := context.Background()

	r := wazero.NewRuntime(ctx)
	defer r.Close(ctx) // This closes everything this Runtime created.

	config := wazero.NewModuleConfig().
		WithStdout(os.Stdout).WithStderr(os.Stderr).WithFS(FS{os.DirFS(".")})

	wasi_snapshot_preview1.MustInstantiate(ctx, r)

	wasm, err := os.ReadFile(os.Args[1])
	if err != nil {
		log.Panicln(err)
	}

	if _, err = r.InstantiateWithConfig(ctx, wasm, config.WithArgs("wasi")); err != nil {
		if exitErr, ok := err.(*sys.ExitError); ok && exitErr.ExitCode() != 0 {
			fmt.Fprintf(os.Stderr, "exit_code: %d\n", exitErr.ExitCode())
		} else if !ok {
[...]

PR #1555, opened 2023-07-04
fix(sysfs): return non-zero ino when adapting fs.FS

GeorgeMac:
Fixes https://github.com/tetratelabs/wazero/issues/1553

This is probably a really stupid idea. However, I found (in the issue linked) that the Go wasip1 implementation will just drop directory entries on the floor with a `0` ino. So `Readdir` just always returns nothing when your supply wazero with an implementation of `fs.FS`.

https://github.com/golang/go/blob/master/src/syscall/dirent.go#L78-L80

So my stupid idea was to just hash the name of the file (with parent for dirent) to a uint64 to represent the ino.
This is probably madness, but it does work for me and my `fs.FS` implementations can now support reading directories when I do this and the binary has been compiled from Go 1.21 with the `wasip1` target.

I branched off of `v1.2.1` to validate this as currently when I work from the HEAD of `main` I consistently get:
```
panic: open .: Bad file number

goroutine 1 [running]:
main.main()
	/Users/georgemac/github/georgemac/wasm-invoke/cmd/walk/main.go:11 +0x14
	==> wasi_snapshot_preview1.proc_exit(rval=2)
```

That is just from `main` itself, even with `os.DirFS`. Something seems up there.

Anyway, this is mostly food for thought and open to alternative ideas. My first approach was to just have a package level uint64 and read it with `atomic.AddUint64(&inos, 1)`.

### DIFF
--- a/internal/sysfs/dir_test.go
+++ b/internal/sysfs/dir_test.go
@@ -29,7 +29,7 @@ func TestReaddir(t *testing.T) {
 		expectIno bool
 	}{
 		{name: "os.DirFS", fs: dirFS, expectIno: runtime.GOOS != "windows"}, // To test readdirFile
-		{name: "fstest.MapFS", fs: fstest.FS, expectIno: false},             // To test adaptation of ReadDirFile
+		{name: "fstest.MapFS", fs: fstest.FS, expectIno: true},              // To test adaptation of ReadDirFile
 	}
 
 	for _, tc := range tests {
--- a/internal/sysfs/file.go
+++ b/internal/sysfs/file.go
@@ -1,9 +1,14 @@
 package sysfs
 
 import (
+	"fmt"
+	"hash"
+	"hash/fnv"
 	"io"
 	"io/fs"
 	"os"
+	"path"
+	"sync"
 	"syscall"
 
 	"github.com/tetratelabs/wazero/internal/fsapi"
@@ -175,6 +180,7 @@ func (f *fsFile) Stat() (st fsapi.Stat_t, errno syscall.Errno) {
 		return
 	} else {
 		st = StatFromDefaultFileInfo(t)
+		st.Ino = hashString(f.name)
 		return f.cacheStat(st)
 	}
 }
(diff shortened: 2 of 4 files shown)


codefromthecrypt:
If your underlying files are `os.File` and you are not using windows, we may be able to special case this differently.

> Go wasip1 implementation will just drop directory entries on the floor with a 0

This seems like a bug or draconian. Can you refer to the line in Go you are referring to for the wasp1 impl? cc @achille-roussel

codefromthecrypt:
Notes from slack chat with @GeorgeMac @achille-roussel and @ncruces

Problems with inodes are common in windows and virtual filesystems because often tests around these are only run on unix operating systems (linux, freebsd, darwin, etc).
* This problem exists until we expose our underlying FS type, which is unstable until wasip2 which isn't scheduled before fall 2023.
* `fs.FileInfo` has no stable means to expose an inode. `syscall.Stat_t` is not defined on windows either.
* `os.SameFile` uses inodes, but the fields are hidden, so we can't return them.

The underlying heuristic in Go that drops dirents with no inode has a historical context in POSIX, moreso Linux `getdents`, but not windows
* https://github.com/golang/go/blob/master/src/os/dir_unix.go#L92-L94
* https://www.spinics.net/lists/linux-fsdevel/msg95504.html
* https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getdirentries.2.html

There are alternatives to this approach:
* fan-out like wasi-libc instead of dropping dirents without an inode
* don't drop dirents without an inode.

@achille-roussel raised a change which may compensate for this issue by the time Go 1.21 is final.
* https://go-review.googlesource.com/c/go/+/507915

Thanks all for the feedback. If we proceed with this avenue, lets make it a FSConfig and re-summarize any notes in RATIONALE.md

codefromthecrypt:
I opened a new issue related to this https://github.com/tetratelabs/wazero/issues/1559

PR #1556, opened 2023-07-04
feat: add `ExportMemory` and `ExportMemoryWithMax` to `HostModuleBuilder` interface

timwu20:
Reintroduces `ExportMemory` and `ExportMemoryWithMax` which was removed in `v1.0.0-pre.2`.  

We require this feature in our project as mentioned in #1575.


codefromthecrypt:
made a comment here. I want you to work, ideally without re-introducing the problem in general motivating us removing this https://github.com/tetratelabs/wazero/issues/1575#issuecomment-1629859983

PR #1557, opened 2023-07-05
gojs: drops HTTP support to be compatible with Go 1.21

codefromthecrypt:
Go 1.21 introduced numerous incompatible changes in `GOOS=js`. This is ok for them, as this GOOS is considered experimental. This makes changes accordingly, including a signature change and removal of HTTP support.

Note: wazero drops this after Go 1.22 is released, so consider the code involved in this PR will last less than a year. However, we need to make this change to allow our CI to work on Go 1.21.

Before, the host module for `GOOS=js` was named "go". It was renamed to "gojs". Since we already require the caller to compile the guest, this adds the compiled module to the parameter list of
`gojs.MustInstantiate`. Doing so is required because the app compiled with wazero could be compiled with a different version of Go than the wasm it is running.

```diff
-               gojs.MustInstantiate(ctx, rt)
+               gojs.MustInstantiate(ctx, rt, guest)
```

One of the most complex pieces of `GOOS=js` involved an attempt to support `http.RoundTripper`. I added this before `wasip1` support started, even before we thought it could start. It was one of the more complex features and hairy to implement due to trying to emulate a callback based JS api.

This was made worse when the `toString` function was added recently, because this requires errors be treated as an object, which adds even more nesting.

See https://github.com/golang/go/commit/a64919945bdb65202783c5f9e078c3e735efc58f

It was made even worse when trying to detect if node is in use, heuristically by assuming a host function can return the arg0 of the actual host binary, and that being "node".

See https://github.com/golang/go/commit/30b17f4f97d37f654c5090001c5b44c331b00d0c

Knowing this module will last less than a year, the correct choice is to not dig the hole further, and instead remove HTTP support from gojs. This implied changing the example to `cat`, like the wasi one.


achille-roussel:
@johanbrandhorst this change might be of interest to you

codefromthecrypt:
ps if Go decides to switch the name back from "gojs" to "go", this PR will still work because it isn't correlating the host module name with the go version. Just in that case we would delete the branch for "gojs".

codefromthecrypt:
another note: there are no uses of gojs RoundTripper visible to github, except forks of wazero's example https://github.com/search?q=wazero+WithRoundTripper&type=code

PR #1558, opened 2023-07-05
Allows wrapped filesystems to return dirent.ino except in windows

### LINKED ISSUE #1553: WithFS(fs.FS) directory read returns unexpected results

codefromthecrypt:
This allows `fs.File` backed implementations to return inode data the same way regardless of it returns an `os.File` or a `fs.File` implementing `Readdir`. Notably, this will not work in windows.

This proves implementations are the same by adding tests with a wrapper implementing functions we use from os.File.

This might close #1553 depending on if @GeorgeMac is ok with the limitations. This would not help with virtual files as the `fs.FileInfo` is still special cased to platform-specific data.


internal/sysfs/adapter_test.go
```diff
@@ -105,9 +107,7 @@ func TestAdapt_Open_Read(t *testing.T) {
 
 	// We can't correct operating system portability issues with os.DirFS on
 	// windows. Use syscall.DirFS instead!
-	if runtime.GOOS != "windows" {
-		testOpen_Read(t, testFS, true)
-	}
+	testOpen_Read(t, testFS, true, runtime.GOOS != "windows")
```
codefromthecrypt, line 110: notice we now make sure we do return inode from stat on windows. This saves us from doing a useless wasi-libc stat fan-out, except in the case of virtual files.

internal/sysfs/stat_windows.go
```diff
@@ -50,17 +50,19 @@ func statPath(createFileAttrs uint32, path string) (fsapi.Stat_t, syscall.Errno)
...
-
-	// ERROR_INVALID_HANDLE happens before Go 1.20. Don't fail as we only
-	// use that approach to fill in inode data, which is not critical.
-	//
-	// Note: statHandle uses UnwrapOSError which coerces
-	// ERROR_INVALID_HANDLE to EBADF.
-	if err != syscall.EBADF {
-		return st, err
+func statFile(f fs.File) (fsapi.Stat_t, syscall.Errno) {
+	if osF, ok := f.(*os.File); ok {
```
codefromthecrypt, line 54: this was the only case we actually used os.File, so I special cased it as it is windows only

codefromthecrypt (commented):
notes

codefromthecrypt:
windows related:

I need to look into this one, and also tighten the assertions because go 1.18 isn't reliable for the others.
```
 --- FAIL: TestReadFS_Open_Read (0.02s)
    --- FAIL: TestReadFS_Open_Read/mask(os.DirFS) (0.00s)
        --- FAIL: TestReadFS_Open_Read/mask(os.DirFS)/readdir_._opens_root (0.00s)
            require.go:330: expected to not equal 0x0: {Dev:0 Ino:0 Uid:0 Gid:0 Mode:drwxrwxrwx Nlink:1 Size:0 Atim:1688529616090722400 Mtim:1688529616090722400 Ctim:1688529616090722400}
                D:/a/wazero/wazero/internal/sysfs/sysfs_test.go:90
```

codefromthecrypt:
github is out-of-date looking at commits, so it hasn't re-run actions after I fixed a glitch. I'll test locally on my windows machine until things restore https://www.githubstatus.com/

ncruces:
LGTM. Some further work to close #1553, IMO.
Wrote those notes on Slack.

codefromthecrypt:
> Some further work to close https://github.com/tetratelabs/wazero/issues/1553, IMO.

#1553 was opened about os.DirFS being wrapped has different behaviors than not wrapped. This proves they have the same behavior.  After this change adapted os.FS work the same as not adapted. If linux/darwin/freebsd, the dirents will have inodes. Nothing else will.

There are indeed tangents that came out of all of this, but I think both need separate issues to track, the first in Go and the second in our repo. If I missed something, let me know!

### Getting go `wasip1` to be lenient when `fd_readdir` returns ino=0

We are the first to notice Go throwing away dirents with 0 inodes, which is not surprising as few use wasip1 for reasons including it isn't released and also doesn't support exports. TinyGo does not throw away dirents with 0 inodes, neither clang, rust or zig. The issue in Go is a problem on a not explicitly supported OS and windows, as well virtual files. However, this is being tracked already https://go-review.googlesource.com/c/go/+/507915. Since our contributors are on that change, I think we don't need to design a solution until it resolves.

### Backfill missing ino on zero

We could also start an effort to virtualize or otherwise backfill inodes. https://github.com/tetratelabs/wazero/pull/1555 starts down a path of virtualizing inodes. This should be optional because it has a performance penalty, it is also a source of bugs as making a heuristic is full of edge cases. For these reasons, the underlying API should not require always returning inodes and be more like wasi where the result is still optional. Config could indicate whether they will be forced non-zero, in other words.

This config is mostly needed when we don't expose our FS api, as otherwise wrapping would be easy to backfill with some fake value even -1 if someone feels comfortable with that.

At the moment, this is a nice to have. Other runtimes don't even support virtual filesystems, and don't bother testing on windows either (wasi-testsuite skips both). Forcing us to handle this now is a bad idea because `fs.FileInfo` has no portable means (well catalogued in our RATIONALE.md). Our replacement filesystem isn't ready yet. We really ought to defer this as long as possible, IMHO. 

If https://go-review.googlesource.com/c/go/+/507915 closes without allowing a late fan-out, probably the better choice is racing forward our filesystem api vs hit the known brick wall with `fs.FS`. Still, the generation heuristic would be a big thing to discuss, and deserves its own issue.

GeorgeMac:
Nice one @codefromthecrypt 🙏 

Just going to re-iterate what I think this means for my custom implementations of `fs.FS`.
This allows me to fabricate inodes in the meantime by adjusting my implementation in the following way:

I need to implement: https://github.com/tetratelabs/wazero/blob/942766707958fc58f52a0b9922a6eb00b4e2bfd0/internal/sysfs/file.go#L427

This way I can skip `ReadFile` and have each of the fs.FileInfo have their `Sys()` method called here:
https://github.com/tetratelabs/wazero/blob/7498ad335f62d17eeda66d9bf5bc3cc73846fec0/internal/sysfs/stat_bsd.go#L34-L39

Allowing me to return e.g. a `*syscal.Stat_t` where I can fabricate the `Ino`?
Depending on the platforms I want my implementation to work on, I will have to adjust accordingly?

This great and works for me. I will take a stab at an implementation later.

codefromthecrypt:
> Allowing me to return e.g. a *syscall.Stat_t where I can fabricate the Ino?
Depending on the platforms I want my implementation to work on, I will have to adjust accordingly?

@GeorgeMac I think you have it right and this could work for now!

The main glitch is `syscall.Stat_t` is not a portable type in Go. It doesn't exist in all operating systems (namely windows maybe something else) and is in different lengths.

Once we expose `fsapi` you won't have to play these games (because it has its own Stat_t type for reason above). I'll try my best to pare this down and get it exposed minimally experimental soon.

GeorgeMac:
Thanks again @codefromthecrypt 🙏

PR #1562, opened 2023-07-06
wasi: adds Go readdir integration tests for `GOOS=wasip1` and TinyGo

codefromthecrypt:
This adds the `ls` test which we hadn't formerly tested on any Go compiler. This led to #1553 by @GeorgeMac. This backfills the tests, and starts with using real files instead of virtual ones.

A later PR will add dimensions for virtual files, something we expect `GOOS=wasip1` and possibly also TinyGo to drop unless we do something to set a non-zero inode. The others won't drop as before this PR, they were already tested with virtual files.

See https://go-review.googlesource.com/c/go/+/507915 for GOOS=wasip1


imports/wasi_snapshot_preview1/wasi_stdlib_test.go
```diff
@@ -48,37 +58,52 @@ var wasmZigCc []byte
...
+	t.Helper()
+
 	moduleConfig := wazero.NewModuleConfig().
-		WithFS(fstest.MapFS{
-			"-":   {},
-			"a-":  {Mode: fs.ModeDir},
-			"ab-": {},
-		})
+		WithFSConfig(wazero.NewFSConfig().
+			WithReadOnlyDirMount(path.Join(rootDir, "dir"), "/"))
```
codefromthecrypt, line 101: So all compilers use our sysfs.DirFS for `ls` tests right now. It is no longer `fs.FS` backed. A later PR will re-add `fs.FS` as a separate dimension, likely masking only `fs.ReadDirFS` to force the zero inode thing.

PR #1563, opened 2023-07-06
fs: pares down to functions used in wasip1 and wasip2

codefromthecrypt:
This removes support for any functions only used in gojs. It retains functions used in wasip1 and a couple additions that we already test that may end up in wasip2, notably:
* cache dev (we already cache ino) for `is_same_object`
  * See https://github.com/WebAssembly/wasi-filesystem/pulls
* chmod for `change-file-permissions-at` and `change-directory-permissions-at`
  * See https://github.com/WebAssembly/wasi-filesystem/blob/1c65f71df8e4fd6eda6970035cd276cba03c5bcb/wit/types.wit#L651-L685


.github/workflows/integration.yaml
```diff
@@ -291,34 +289,6 @@ jobs:
             -f ../.github/wasi_testsuite_skip.json \
             -r ../.github/wasi_testsuite_adapter.py
 
-  build_gojs_test_binaries:
```
codefromthecrypt, line 294: I stopped this because it only takes 2s to build the test binary, and the cache setup was distracting.

.github/workflows/integration.yaml
```diff
@@ -259,9 +259,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install wazero
-        run: |
-          go install ./cmd/wazero
-        shell: bash
```
codefromthecrypt, line 264: we set bash to default globally in the same file

internal/gojs/fs.go
```diff
@@ -464,16 +464,16 @@ type jsfsFchmod struct{}
 
 func (jsfsFchmod) invoke(ctx context.Context, mod api.Module, args ...interface{}) (interface{}, error) {
 	fd := goos.ValueToInt32(args[0])
-	mode := custom.FromJsMode(goos.ValueToUint32(args[1]), 0)
+	_ /* mode */ = args[1]
```
evacchi, line 467: intended?

evacchi, line 467: I'll be the nitpicker here; as I am seeing this is repeated I assume it's intended: is it a common style? I would normally use

```
_ = args[1] // mode
```

(same for the rest of the file)

evacchi (approved):
> honk if you love windows+go1.18 and inodes!

internal/gojs/fs.go
codefromthecrypt: works for me!

Issue #1564, opened 2023-07-06
panic: Bad file number when opening file with Go 1.21 wasip1


GeorgeMac:
**Describe the bug**

I consistently get a panic when trying to open any file using a wasm binary built with `gotip`.
Specifically, this only starts occuring after this commit was introduced in to wazero main: https://github.com/tetratelabs/wazero/commit/53ce5eea830b3100af7ec8caa13bb38310c0afbf#diff-4fb3ce1d291b912fd6aa5e7f790040234a92683e313d1c757c0fce7be9e43320L397-L400 

It does not fail in the parent.

**To Reproduce**
Take the following Go code and compile it with `gotip` to `wasm` and `wasip1` platform/architecture:

```go
package main

import (
	"fmt"
	"os"
)

func main() {
	_, err := os.Open(".")
	if err != nil {
		panic(err)
	}
}
```

```
GOOS=wasip1 GOARCH=wasm gotip build -o open.wasm main.go
```

The run with `wazero run`:

```
wazero run --hostlogging=all open.wasm
```

**Expected behavior**

No error to be produced.

**Screenshots**

```
--> ._rt0_wasm_wasip1()
	==> wasi_snapshot_preview1.random_get(buf=1191392,buf_len=32)
	<== errno=ESUCCESS
	==> wasi_snapshot_preview1.random_get(buf=1191064,buf_len=8)
	<== errno=ESUCCESS
	==> wasi_snapshot_preview1.clock_time_get(id=monotonic,precision=0)
	<== (timestamp=1625363670137041,errno=ESUCCESS)
	==> wasi_snapshot_preview1.args_sizes_get(result.argc=1210940,result.argv_len=1210936)
	<== errno=ESUCCESS
	==> wasi_snapshot_preview1.args_get(argv=21020680,argv_buf=21020688)
	<== errno=ESUCCESS
	==> wasi_snapshot_preview1.environ_sizes_get(result.environc=1210928,result.environv_len=1210932)
	<== errno=ESUCCESS
	==> wasi_snapshot_preview1.clock_time_get(id=monotonic,precision=0)
	<== (timestamp=1625363674881875,errno=ESUCCESS)
	==> wasi_snapshot_preview1.clock_time_get(id=monotonic,precision=0)
	<== (timestamp=1625363674888833,errno=ESUCCESS)
	==> wasi_snapshot_preview1.clock_time_get(id=monotonic,precision=0)
[...]

PR #1565, opened 2023-07-06
fs: empty guest path should bind to /

evacchi:
Fixes https://github.com/tetratelabs/wazero/issues/1564.

Commit https://github.com/tetratelabs/wazero/commit/53ce5eea830b3100af7ec8caa13bb38310c0afbf#diff-4fb3ce1d291b912fd6aa5e7f790040234a92683e313d1c757c0fce7be9e43320L397-L400 altered the effective behavior of `.WithFS()`; whereas prior to this patch a single, non-composite FS would always bind to "/" :

https://github.com/tetratelabs/wazero/blob/97d0d70b7321e5c16e584bbb6b3169032208bdec/internal/sys/fs.go#L469-L476

now, the `guestPath` is always used:

https://github.com/tetratelabs/wazero/blob/485b9e3ece346c2dd0912e1a54703d70fa10cdcf/internal/sys/fs.go#L420-L425

we resolve by overriding `guestPath` to `"/"` when `StripPrefixesAndTrailingSlash(guestPath) == ""`

### DIFF
--- a/internal/sys/fs.go
+++ b/internal/sys/fs.go
@@ -413,6 +413,8 @@ func (c *Context) InitFSContext(
 		guestPath := guestPaths[i]
 
 		if StripPrefixesAndTrailingSlash(guestPath) == "" {
+			// Default to bind to '/' when guestPath is effectively empty.
+			guestPath = "/"
 			c.fsc.rootFS = fs
 		}
 		c.fsc.openedFiles.Insert(&FileEntry{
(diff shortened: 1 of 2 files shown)


achille-roussel (approved):
Thanks for the fix!

codefromthecrypt (approved):
Thanks, this makes perfect sense

PR #1566, opened 2023-07-07
gojs: removes user/group behavior

codefromthecrypt:
This should have been removed in #1563, but I missed it. We are only supporting features used in wasip1/p2. This also never worked in windows.


achille-roussel (approved):
🔪

PR #1567, opened 2023-07-07
Exposes sys.Stat_t as a portable alternative to syscall.Stat_t

codefromthecrypt:
This exposes our internal `Stat_t` type for use in `fs.FileInfo.Sys`. This allows customizers to provide raw stat details from host functions or mask inodes with fake ones.


codefromthecrypt:
if folks are ok with this, I'll add a RATIONALE.md section for each and every field as well an example_test.go for how to use this to embelish an fs.FS with inodes cc @GeorgeMac

sys/stat_test.go
```diff
@@ -0,0 +1,156 @@
...
+			ModTime: osDirInfo.ModTime(),
+		},
+		"dir/file": {
+			Data:    fileData,
+			Mode:    osFileInfo.Mode(),
+			ModTime: osFileInfo.ModTime(),
+		},
+		"dir/file-sys": {
+			// intentionally skip other fields to prove sys is read.
+			Sys: &osFileSt,
```
codefromthecrypt, line 46: this is an example of using our stat type in lieu of parsing FileInfo

sys/stat_test.go
```diff
@@ -0,0 +1,156 @@
...
+					require.Equal(t, uint64(1), st.Nlink)
+				}
+			} else { // Nlink is possibly wrong, but not zero.
+				require.Equal(t, uint64(1), st.Nlink)
+			}
+
+			require.Equal(t, tc.expectedSize, st.Size)
+
+			if tc.expectAtimCtime && sysParseable {
+				// We don't validate times strictly because it is os-dependent
```
codefromthecrypt, line 144: I tried hard to do some validation on times, but it was a fools errand. it is really hard to get these to change especially without sleeping, in a portable way.

sys/stat.go
```diff
@@ -0,0 +1,97 @@
...
+// Ino is the file serial number, or zero if unknown.
+//
+// Any constant value will invalidate functions that use Ino for equivalence,
+// such as os.SameFile.
+//
+// When zero is returned by a `readdir`, some compilers will attempt to
+// get a non-zero value with `lstat`. Those using Ino for darwin's definition
+// of `getdirentries` conflate zero `d_fileno` with a deleted file and skip the
+// entry. See /RATIONALE.md for more on this.
+type Ino = uint64
```
achille-roussel, line 14: I like that this is an alias, it gives room for documentation without really changing the semantics of the `Stat_t.Ino` field in relation to the other fields of the struct.

achille-roussel (approved):

sys/stat_example_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"io/fs"
+	"math"
+
+	"github.com/tetratelabs/wazero"
+	"github.com/tetratelabs/wazero/sys"
+)
+
+type sysStatFS struct{ fs.FS }
```
codefromthecrypt, line 11: @evacchi @GeorgeMac @ncruces I have to dash.. I spiked this just to put something out there. If I messed this up feel free to comment or correct. I'll fix anything including adding more details to RATIONALE.md sometime not tonight.

GeorgeMac: Nice, this all tracks 👍

internal/sysfs/stat_linux.go
```diff
@@ -10,52 +10,33 @@ import (
...
 	}
 }
 
-func statFile(f fs.File) (fsapi.Stat_t, syscall.Errno) {
+func statFile(f fs.File) (sys.Stat_t, syscall.Errno) {
 	return defaultStatFile(f)
 }
 
-func inoFromFileInfo(_ string, t fs.FileInfo) (ino fsapi.Ino, err syscall.Errno) {
+func inoFromFileInfo(_ string, t fs.FileInfo) (ino sys.Ino, err syscall.Errno) {
```
GeorgeMac, line 37: Should this function also be attempting to cast to `sys.Stat_t`? Just trying to understand the distinction.
Infact can all these `stat_*` implementations be single sourced if they're going through `sys.Stat_t`?

codefromthecrypt: this is a little more complicated because we need to get the ino, and we have a path to get it in windows (pun intended)

That said, later I'll try to see if we can unwind and still do that. Good to revisit it!

sys/stat_unsupported.go
```diff
@@ -0,0 +1,11 @@
+//go:build (!((amd64 || arm64 || riscv64) && linux) && !((amd64 || arm64) && (darwin || freebsd)) && !((amd64 || arm64) && windows)) || js
```
ncruces, line 1: Wondering if we need to do this by CPU when it's a OS thing, but it's likely convention between "supported" (we test it) vs. "unsupported" (we don't).

ncruces (approved):

sys/stat_unsupported.go
codefromthecrypt: I agree it is mostly about what we are testing, as there would be some stat_t which would be compatible with our conversion logic.

I think this was also trying to say "64-bit" since the stat lengths are different on 32. I'll backfill a comment which I suppose we could support later (even if not test it) if someone needed precise stat on a 32-bit platform.

internal/sysfs/stat_linux.go
codefromthecrypt: I see what you meant originally and you are right!

ncruces (approved):

codefromthecrypt:
ok everything done imho except RATIONALE, which I'm adding now

fsconfig.go
```diff
@@ -118,6 +118,18 @@ type FSConfig interface {
 	// advise using WithDirMount instead. There will be behavior differences
 	// between os.DirFS and WithDirMount, as the latter biases towards what's
 	// expected from WASI implementations.
+	//
+	// # Custom fs.FileInfo
```
codefromthecrypt, line 122: Here is the integration doc, though obviously our underlying FS will be better when it is out. That said fs.FS may still be easier for some even after that, and so I don't think it is a problem to document this

fsconfig_example_test.go
```diff
@@ -14,7 +14,7 @@ var testdataIndex embed.FS
 var moduleConfig wazero.ModuleConfig
 
 // This example shows how to configure an embed.FS.
-func Example_withFSConfig_embedFS() {
```
codefromthecrypt, line 17: this is a bug as godoc won't actually render it

internal/sysfs/adapter_test.go
```diff
@@ -168,18 +169,69 @@ func TestAdapt_HackedWrites(t *testing.T) {
...
+	} else if fs.ZeroIno {
+		return &zeroInoOsFile{osF}, nil
 	} else {
 		return struct{ methodsUsedByFsAdapter }{osF}, nil
 	}
 }
 
+// zeroInoOsFile wraps an os.File to override functions that can return
+// fs.FileInfo.
+type zeroInoOsFile struct{ *os.File }
```
codefromthecrypt, line 192: I chose to not make a top-level example for this, as we can point the few who do this to the unit test. Reason being it is complicated and godoc examples can't render anything outside the function body. Instead, we should document the real fs once we expose it, shortly.

mathetake (approved):

codefromthecrypt:
added RATIONALE, so I think done

codefromthecrypt:
I'll handle more feedback post merge if folks have it. Meanwhile, merging as we prepare for v1.3.0 soon

PR #1568, opened 2023-07-07
explicitly documents ModuleConfig defaults to _start

codefromthecrypt:
wazero defines a helper function, `Instantiate`, which doesn't accept a `ModuleConfig`. This highlights that the default will call wasip1's `_start` function if it exists.

This does not fix #1561 but it documents the footgun we allow until that is fixed.

PR #1569, opened 2023-07-07
Adds more links to rationale about start functions

codefromthecrypt:
follow-up to #1568

### DIFF
--- a/config.go
+++ b/config.go
@@ -490,6 +490,8 @@ type ModuleConfig interface {
 	// WithStartFunctions configures the functions to call after the module is
 	// instantiated. Defaults to "_start".
 	//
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -490,6 +490,8 @@ type ModuleConfig interface {
 	// WithStartFunctions configures the functions to call after the module is
 	// instantiated. Defaults to "_start".
 	//
+	// Clearing the default is supported, via `WithStartFunctions()`.
```
codefromthecrypt, line 493: here's a reminder you can clear this. we clear it when we explicitly time `_start`, but most frameworks that use WASI or FFI won't clear it as it makes init require more logic and error handling, hence this config.

evacchi (approved):

PR #1570, opened 2023-07-09
wasi: nonblocking pipes on Windows

evacchi:
Improve #1500 with a simple observation, i.e. we really need non-blocking I/O only for pipes and sockets.

Now, for sockets we can use WinSock select, as for pipes, select_windows.go already imports PeekNamedPipe.

We combine PeekNamedPipe with a blocking Read only for those cases when PeekNamedPipe returns n > 0, which means a Read of n bytes won't block.

We special case n == 0 to return EAGAIN and
ERROR_BROKEN_PIPE as a non-failure (EOF reached).

We also introduce `isNonblock(f)` function, specialized on Windows to also check if the file is ModeNamedPipe; otherwise we would default to blocking anyway.

### DIFF
--- a/imports/wasi_snapshot_preview1/wasi_stdlib_test.go
+++ b/imports/wasi_snapshot_preview1/wasi_stdlib_test.go
@@ -498,18 +498,42 @@ func Test_Stdin(t *testing.T) {
 }
 
 func testStdin(t *testing.T, bin []byte) {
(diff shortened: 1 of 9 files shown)


ncruces (approved):

codefromthecrypt (approved):
thanks for the progress!

PR #1571, opened 2023-07-10
Changes build to test go 1.21rc2

codefromthecrypt:
This is in preparation for the upcoming Go 1.21 release, which will also raise the floor Go version to 1.19.

### DIFF
--- a/internal/sysfs/futimens.go
+++ b/internal/sysfs/futimens.go
@@ -50,16 +50,11 @@ func Utimens(path string, times *[2]syscall.Timespec, symlinkFollow bool) syscal
 	return platform.UnwrapOSError(err)
 }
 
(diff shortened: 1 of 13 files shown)


internal/sysfs/futimens.go
```diff
@@ -50,16 +50,11 @@ func Utimens(path string, times *[2]syscall.Timespec, symlinkFollow bool) syscal
 	return platform.UnwrapOSError(err)
 }
 
-var _zero uintptr //nolint:unused
```
codefromthecrypt, line 53: thanks @evacchi for figuring this out. Futimes tests failed on 1.21 until this was replaced with nil

mathetake (approved):

PR #1572, opened 2023-07-10
Fix drift in comment in mmap_other

anuraaga:
Noticed this comment which seems to have drifted

- `syscall.Mmap` is called, not `Mprotect`
- The code is for darwin and freebsd, not linux

I tried updating to where I suspect the drift might have happened, but happy to tweak further or remove if it seems unneeded


codefromthecrypt (approved):
thanks!

Issue #1197, opened 2023-03-04
Module closure notification 


ncruces:
Is there an idiomatic way to get notified of module closure (e.g. a `panic`, or WASI `proc_exit`)?

My use case is releasing resources (closing a file) when the module crashes. I could check the return value of every function call but im wondering if there's a better way.

On a related note, and I haven't tested it yet, does WASI close files when the module closes? I'd consider it a bug if not (but haven't read the specs on it), and looking over the code I couldn't find anything ensuring `FSContext.Close` gets called.

codefromthecrypt:
right now, the only thing that shows you if something exited is in the experimental logger, which has LogScopeExit. This logs that the function was called, but it isn't about the fact that the underlying module was closed.

To know a module was closed even when not from a function (ex. defer closing it), we'd need a non-function hook like #1150

In any case, files are closed when the module is (syscontext does this or we'd have windows errors galore)

ncruces:
If I understood your second point correctly, this gets called *always* when the module closes:
https://github.com/tetratelabs/wazero/blob/a9b3301862be7bb33f8123898c84c63242f9e596/internal/wasm/call_context.go#L165-L181

I guess my RFE, if it makes sense, would be to be able to inject something like `CodeCloser` in there.

Unfortunately, I open my own files, from Go, not just because WASI overhead was considerable, but because I need file locking.

As for your first suggestion, it seems the idiomatic way (if this was not an experimental API) would be to follow this [example](https://pkg.go.dev/github.com/tetratelabs/wazero@v1.0.0-rc.1/experimental#example-package-CustomListenerFactory), and check for errors on `After`.

I wonder what's the overhead of doing that, vs. wrapping all calls in my own function, which I think I'm already doing (mostly for ergonomic reasons).

Thanks, as always!

codefromthecrypt:
> I guess my RFE, if it makes sense, would be to be able to inject something like CodeCloser in there.

I think this plumbing will be similar in nature to the request around a memory.grow hook. Basically the scope is *anything*, so it will get notified twice for folks using WASI, once for explicit close inside `proc_exit` and an implicit redundant close due to a `defer rt.Close(ctx)`, unless we choose to not notify after closed.

> I wonder what's the overhead of doing that, vs. wrapping all calls in my own function, which I think I'm already doing (mostly for ergonomic reasons).

[...]

PR #1574, opened 2023-07-10
experimental: adds close notification hook

codefromthecrypt:
`CloseNotifier` is experimental progress towards #1197, allowing notification when a module is closed.


ckaznocha (commented):
The gist of this looks good to me!

This is how I would probably use this:
```go
closeCh := make(chan struct{})
ctx = close.WithNotification(
	ctx,
	close.NotificationFunc(func(context.Context, uint32){ close(closeCh} ),
)

// ... create module, do some work. Sometime later in another goroutine:

select {
	case <- closeCh:
		// do some cleanup
	default:
		// do some more work with the module
}
```

I don't know that this adds an advantage over new `Module.IsClosed` but it's nice to have the inverted control flow.

---
On the impl itself, I may be missing something but I don't see where `ModuleInstance.CloseNotification` gets set. This is still a draft so sorry if I'm giving feedback on something you didn't intend to be fully functional yet.

ncruces (approved):
This solves the issue for me. Not sure about the API, mostly nits about names.

I mentioned `Notification` seeming too generic on Slack, but I had missed that the package name `close`, which makes it OK. Remains to be seen if this would be kept in its own  package once out of experimental. Also `OnXXX` seemed jarring for Go.

There was once an [`http.CloseNotifier`](https://pkg.go.dev/net/http#CloseNotifier), before it got replaced by `Context`.

Would the name be more idiomatic (and moving it up to `experimental`)? Idk…

internal/close/close.go
```diff
@@ -0,0 +1,11 @@
+package close
+
+import "context"
+
+// NotificationKey is a context.Context Value key. Its associated value should
+// be a Notification.
+type NotificationKey struct{}
+
+type Notification interface {
```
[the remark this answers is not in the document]
codefromthecrypt: I'll add a note, mainly it was about hiding the context key and also at one point there were cyclic deps (there aren't any anymore at the moment)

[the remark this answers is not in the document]
codefromthecrypt: When we move it up to experimental package, this becomes a way to avoid cyclic deps again ;)

internal/wasm/module_instance.go
```diff
@@ -146,6 +146,12 @@ func (m *ModuleInstance) setExitCode(exitCode uint32, flag exitCodeFlag) bool {
 // ensureResourcesClosed ensures that resources assigned to ModuleInstance is released.
 // Multiple calls to this function is safe.
 func (m *ModuleInstance) ensureResourcesClosed(ctx context.Context) (err error) {
+	if notification := m.CloseNotification; notification != nil { // experimental
+		closed := atomic.LoadUint64(&m.Closed)
+		notification.OnClose(ctx, uint32(closed>>32))
+		m.CloseNotification = nil
+	}
```
mathetake, line 153: shouldn't this be guarded by sync.Once?

codefromthecrypt: yep! ensureResourcesClosed is externally guarded atomically to make sure it is only called once

codefromthecrypt: I'll add the missing comment

mathetake: aha!

codefromthecrypt:
I'll add an example test, but I've already used this in gojs

experimental/close.go
```diff
@@ -0,0 +1,63 @@
...
+	//   - This does not return an error because the module will be closed
+	//     unconditionally.
+	//   - Do not panic from this function as it doing so could cause resource
+	//     leaks.
+	//   - While this is only called once per module, if configured for
+	//     multiple modules, it will be called for each, e.g. on runtime close.
+	CloseNotify(ctx context.Context, exitCode uint32)
+}
+
+// ^-- Note: This might need to be a part of the listener or become a part of
```
codefromthecrypt, line 28: I added a lot of text here because we still need host module state which is associated with an instantiated guest. I don't know if we will repurpose this design as an evolution towards that, or just make another. In any case, both will need to handle close though possibly host module state doesn't need a notification model (rather it can be explicitly invoked)

internal/gojs/run/gojs.go
```diff
@@ -11,33 +11,26 @@ import (
 	"github.com/tetratelabs/wazero/sys"
 )
 
-func RunAndReturnState(
-	ctx context.Context,
-	r wazero.Runtime,
-	compiled wazero.CompiledModule,
-	moduleConfig wazero.ModuleConfig,
-	config *config.Config,
-) (*gojs.State, error) {
```
codefromthecrypt, line 20: first use is getting rid of this return, which I implemented with a notifier instead

internal/gojs/compiler_test.go
```diff
@@ -58,14 +59,13 @@ func compileAndRunWithRuntime(ctx context.Context, r wazero.Runtime, arg string,
 		WithStderr(&stderrBuf).
 		WithArgs("test", arg))
 
-	var s *internalgojs.State
-	s, err = run.RunAndReturnState(ctx, r, guest, mc, c)
-	if err == nil {
+	ctx = experimental.WithCloseNotifier(ctx, experimental.CloseNotifyFunc(func(ctx context.Context, exitCode uint32) {
+		s := ctx.Value(internalgojs.StateKey{})
 		if want, have := internalgojs.NewState(c), s; !reflect.DeepEqual(want, have) {
```
codefromthecrypt, line 64: ps I also checked with a debugger this was actually invoked ;)

mathetake (approved):

runtime.go
```diff
@@ -320,6 +320,10 @@ func (r *runtime) InstantiateModule(
 		return
 	}
 
+	if closeNotifier, ok := ctx.Value(internalclose.NotifierKey{}).(internalclose.Notifier); ok {
```
codefromthecrypt, line 323: note: this will notify on host and guest modules, so if you use the same context you can get multiple notifications on runtime.Close. I added this note to the godoc as well.

codefromthecrypt:
cc @abraithwaite we are evolving close notification, so you might be interested

abraithwaite:
>  cc @abraithwaite we are evolving close notification, so you might be interested

Thanks for the CC!  I think this is not really what we're looking for but appreciate the thought to make us aware of such feature.

We are really just concerned with being able to close the runtime (and all modules within it) from a single host function.  We figured that wasn't happening with the memory leak from earlier, but that's been cleared up a few weeks now :-)

PR #1579, opened 2023-07-12
wasi: nonblocking I/O for sockets and pipes on Windows

evacchi:
Further addresses #1500. 

Further work to improve the support for nonblocking I/O on Windows. This drops the special-casing for stdin, and instead checks if a handle is a named pipe (which includes sockets and other kinds of pipes); moreover, it further improves `_select` by introducing a proper wrapper for Winsock's select, which is BSD socket's select, and a compatible implementation of FdSet.

Because we need to distinguish between sockets, pipes, and regular files, the `FdSet` struct is actually wrapping 3 `WinSockFdSet`s, one for each type; the correct bucket is picket by probing the fd (win Handle) when it's passed to the `FdSet` method (e.g. `FdSet.Set(fd)`).

I think this can be further simplified, but I think now it is addressing some of the issues with the current implementation. Further work would be working on `poll_oneoff` and possibly remove `File.PollRead()` if the `_select()` implementations are now good enough for all platforms.

### DIFF
--- a/internal/platform/fdset_windows.go
+++ b/internal/platform/fdset_windows.go
@@ -1,0 +1,209 @@
+package platform
+
+import (
+	"syscall"
+	"unsafe"
+)
+
+var procGetNamedPipeInfo = kernel32.NewProc("GetNamedPipeInfo")
+
+// Maximum number of fds in a WinSockFdSet.
+const _FD_SETSIZE = 64
+
+// WinSockFdSet implements the FdSet representation that is used internally by WinSock.
+//
+// Note: this representation is quite different from the one used in most POSIX implementations
+// where a bitfield is usually implemented; instead on Windows we have a simpler array+count pair.
+// Notice that because it keeps a count of the inserted handles, the first argument of select
+// in WinSock is actually ignored.
+//
+// The implementation of the Set, Clear, IsSet, Zero, methods follows exactly
+// the real implementation found in WinSock2.h, e.g. see:
+// https://github.com/microsoft/win32metadata/blob/ef7725c75c6b39adfdc13ba26fb1d89ac954449a/generation/WinSDK/RecompiledIdlHeaders/um/WinSock2.h#L124-L175
+type WinSockFdSet struct {
+	// count is the number of used slots used in the handles slice.
+	count uint64
+	// handles is the array of handles. This is called "array" in the WinSock implementation
+	// and it has a fixed length of _FD_SETSIZE.
+	handles [_FD_SETSIZE]syscall.Handle
+}
+
+// FdSet implements the same methods provided on other plaforms.
+//
+// Note: the implementation is very different from POSIX; Windows provides
+// POSIX select only for sockets. We emulate a select for other APIs in the sysfs
+// package, but we still want to use the "real" select in the case of sockets.
+// So, we keep a separate FdSet of sockets, so that we can pass it directly
+// to the winsock select implementation
+type FdSet struct {
+	sockets WinSockFdSet
+	pipes   WinSockFdSet
+	regular WinSockFdSet
+}
+
+// Sockets returns a WinSockFdSet containing the handles in this FdSet that are sockets.
+func (f *FdSet) Sockets() *WinSockFdSet {
+	if f == nil {
+		return nil
+	}
+	return &f.sockets
+}
+
+func (f *FdSet) SetSockets(s WinSockFdSet) {
+	f.sockets = s
+}
+
+// Regular returns a WinSockFdSet containing the handles in this FdSet that are regular files.
+func (f *FdSet) Regular() *WinSockFdSet {
+	if f == nil {
+		return nil
+	}
+	return &f.regular
+}
+
+func (f *FdSet) SetRegular(r WinSockFdSet) {
+	f.regular = r
+}
+
+// Regular returns a WinSockFdSet containing the handles in this FdSet that are pipes.
+func (f *FdSet) Pipes() *WinSockFdSet {
+	if f == nil {
+		return nil
+	}
+	return &f.pipes
+}
+
+func (f *FdSet) SetPipes(p WinSockFdSet) {
+	f.pipes = p
+}
+
+func (f *FdSet) getFdSetFor(fd int) *WinSockFdSet {
+	h := syscall.Handle(fd)
+	t, err := syscall.GetFileType(h)
+	if err != nil {
+		return nil
+	}
+	switch t {
+	case syscall.FILE_TYPE_CHAR, syscall.FILE_TYPE_DISK:
+		return &f.regular
+	case syscall.FILE_TYPE_PIPE:
+		if isSocket(h) {
+			return &f.sockets
+		} else {
+			return &f.pipes
+		}
+	default:
+		return nil
+	}
+}
+
+// Set adds the given fd to the set.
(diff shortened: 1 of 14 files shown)


internal/sysfs/select_windows.go
```diff
@@ -37,69 +33,181 @@ func syscall_select(n int, r, w, e *platform.FdSet, timeout *time.Duration) (int
...
 // The duration may be nil, in which case it will wait undefinely. The given ctx is
-// used to allow for cancellation. Currently used only in tests.
-func pollNamedPipe(ctx context.Context, pipeHandle syscall.Handle, duration *time.Duration) (bool, error) {
+// used to allow for cancellation, and it is currently used only in tests.
+//
+// As indicated in the man page for select [1], r, w, e are modified upon completion:
+//
+// "Upon successful completion, the pselect() or select() function shall modify the objects pointed to by the readfds,
+// writefds, and errorfds arguments to indicate which file descriptors are ready for reading, ready for writing,
+// or have an error condition pending, respectively, and shall return the total number of ready descriptors in all the output sets"
```
codefromthecrypt, line 54: As this is emulation with goroutines, it is worthwhile to explain what is in windows, but not possible to use. Also, any downside such as this implies GOMAXPROCS>1 if that's true. This explanation could be in the RATIONALE.md like "Why do we emulate POSIX select on windows?"

internal/sysfs/select_windows.go
```diff
@@ -37,69 +33,181 @@ func syscall_select(n int, r, w, e *platform.FdSet, timeout *time.Duration) (int
...
+	// winsock_select mutates the given references, so we create copies.
+	rs, ws, es := r.Sockets().Copy(), w.Sockets().Copy(), e.Sockets().Copy()
+
 	// Short circuit when the duration is zero.
 	if duration != nil && *duration == time.Duration(0) {
-		bytes, err := peekNamedPipe(pipeHandle)
-		return bytes > 0, err
+		nsocks, errno = winsock_select(rs, ws, es, duration)
+	} else {
+		// Ticker that emits at every pollInterval.
```
codefromthecrypt, line 83: I presume this wasn't extracted to a function because the state "rs, ws, es, duration" or whatever else is needed here, isn't easy to create (and unit test). If it isn't or isn't harder, I might suggest extracting this part and testing it directly vs indirectly via selectAllHandles.

Regardless, if you don't mind doing a coverage pass. It feels like `selectAllHandles` accepts a context param which would allow reaching some edge cases here (e.g. context already done), but not sure if it is or not.. coverage could tell for sure.

Finally, any area where we rely on goroutines can have some subtle issues and we should be careful. In general stateful concurrent code is the sort of last resort, and I understand this may be the only way out for windows (vs say not supported). Maybe use the hammer test (for the high level function) to make sure the windows calls used below don't have any one-at-a-time things that should be locked. That or say too much.. in any case please consider what may be needed about any concurrency issues here.

codefromthecrypt (approved):
I'm marking this as approve, knowing that we may be digging a hole and some things may not be tenable to support on all platforms. Emulation via concurrency on windows is a high skill low population thing, and we have the skills (the edo), but should still consciously think through things as POSIX emulation get gnarlier and gnarlier.

internal/sysfs/select_windows.go
evacchi: 💯 on adding a section to RATIONALE. The tl;dr: is that I don't think there is one single API that does everything `select` does on POSIX, and, because winsock select is blocking under certain conditions we need to spin a goroutine to avoid blocking everything

another important note: TBH even UNIX select is not perfect for our use cases, because it won't honor the clock config of the module; so if you have a "poll 1 sec fd 4", and you implemented it with `select`, you would end up blocking for literally 1 sec. So if you wanted to do it properly, I think you may have to use some concurrency primitives there as well.

OTOH when this type of I/O is involved I don't even know if it composes with simulated clocks at all.

codefromthecrypt: agree on the tension of real clocks and real files. I found myself in that spaghetti before, thanks for clarifying!

internal/sysfs/select_windows.go
evacchi: I have reworked all this bit and did a coverage pass, adding a few more tests. With the `select(timeout:=0)` trick there is no longer a user-controlled goroutine, just channels with ticks and timeouts as before.

internal/platform/fdset_windows_test.go
```diff
@@ -0,0 +1,156 @@
...
+		{
+			name: "should clear all even bits",
+			init: allSet,
+			exec: func(fdSet *WinSockFdSet) {
+				for fd := 0; fd < _FD_SETSIZE; fd += 2 {
+					fdSet.Clear(fd)
+				}
+				for fd := 0; fd < _FD_SETSIZE; fd++ {
+					isSet := fdSet.IsSet(fd)
+					if fd&0x1 == 0x1 {
```
codefromthecrypt, line 98: require.Equals would be nicer

internal/platform/fdset_windows.go
```diff
@@ -0,0 +1,239 @@
...
+		if f.handles[i] == h {
+			return true
+		}
+	}
+	return false
+}
+
+// Zero clears the set.
+func (f *WinSockFdSet) Zero() {
+	if f == nil {
```
codefromthecrypt, line 199: unless we use things as nil (later re-assign the pointer) I would leave out the receiver == nil thing for internal types

internal/sysfs/select_windows.go
codefromthecrypt: thank you!

internal/sysfs/select_windows_test.go
```diff
@@ -2,40 +2,66 @@ package sysfs
...
+
 	t.Run("peekNamedPipe should report the correct state of incoming data in the pipe", func(t *testing.T) {
 		r, w, err := os.Pipe()
 		require.NoError(t, err)
 		rh := syscall.Handle(r.Fd())
 		wh := syscall.Handle(w.Fd())
 
-		// Ensure the pipe has data.
+		// Ensure the pipe has no data.
 		n, err := peekNamedPipe(rh)
```
codefromthecrypt, line 62: nit if errno return as the name errno and then use require.EqualErrno as it has better formatting on fail

codefromthecrypt (approved):
Really good work here. I appreciate the tenacity to eliminate a the goroutine.

I'm not going to subject you to a nit pass. if you feel like anything you can put it in another PR or part of something else, because probably this code will be touched again anyway.

internal/platform/fdset_windows.go
evacchi: the receiver can be nil in many places because select accepts fdreads, fdwrites, fdexcepts to be nil. By allowing the receiver to be nil, some error checks can be avoided e.g. `f.Count()` returning 0 when f == nil, or in this case, zeroing out a nil receiver, which is also useful in `select`. To be fair, maybe `select` can be rewritten in a different way so that the nil checks are centralized at the top of the function (maybe replacing nils with empty structs)

evacchi:
will do a follow up for the nits

internal/platform/fdset_windows.go
codefromthecrypt: I think what made me look was not all functions check nil ;) Anyway it is cool to know at least some of it is helpful!

PR #1581, opened 2023-07-13
fs: nonblocking writes should use syscall.Write

### LINKED ISSUE #1538: Non-blocking stdio isn't working correctly

evacchi:
resolves #1538 on non-Windows platforms. Windows will be addressed in #1579.

Using blocking `file.Write` APIs together with nonblocking I/O, causes stdout corruption under certain conditions. 

In particular, a reliable way to reproduce the issue is to capture the stdout of a sub-process running wazero. This PR fixes the writing issue by introducing `writeFd` (similar to `readFd` introduced in #1517), a function that just invokes the syscall directly when I/O is non-blocking. 

It also introduces the constants `nonBlockingFileReadSupported`, `nonBlockingFileWriteSupported`, superseding the generic `nonBlockingFileIoSupported`: the reason is still related to #1579, where Windows-related concerns will be addressed; for now we ignore the issue on that platform. However, because the test is _not skipped_ on Windows, it looks like that platform is not affected (also verified on my Windows machine).

### DIFF
--- a/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
+++ b/imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -8,6 +9,8 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -53,6 +56,8 @@ func main() {
 		}
 	case "stdout":
 		mainStdout()
+	case "largestdout":
+		mainLargeStdout()
 	}
 }
 
@@ -259,3 +264,46 @@ func mainStdin() error {
 func mainStdout() {
 	os.Stdout.WriteString("test")
 }
+
+func mainLargeStdout() {
+	const ntest = 1024
+
+	var decls, calls bytes.Buffer
+
+	for i := 1; i <= ntest; i++ {
+		s := strconv.Itoa(i)
+		decls.WriteString(strings.Replace(decl, "$", s, -1))
+		calls.WriteString(strings.Replace("call(test$)\n\t", "$", s, -1))
+	}
+
+	program = strings.Replace(program, "$DECLS", decls.String(), 1)
+	program = strings.Replace(program, "$CALLS", calls.String(), 1)
+	fmt.Print(program)
+}
+
+var program = `package main
+
+var count int
+
+func call(f func() bool) {
+	if f() {
+		count++
+	}
+}
+
+$DECLS
+
+func main() {
+	$CALLS
+	if count != 0 {
+		println("failed", count, "case(s)")
+	}
+}
+`
+
+const decl = `
+type T$ [$]uint8
+func test$() bool {
+	v := T${1}
+	return v == [$]uint8{2} || v != [$]uint8{1}
+}`
--- a/internal/sysfs/file_unix.go
+++ b/internal/sysfs/file_unix.go
@@ -8,7 +8,10 @@ import (
 	"github.com/tetratelabs/wazero/experimental/sys"
 )
 
-const nonBlockingFileIoSupported = true
+const (
+	nonBlockingFileReadSupported  = true
+	nonBlockingFileWriteSupported = true
+)
 
 // readFd exposes syscall.Read.
 func readFd(fd uintptr, buf []byte) (int, sys.Errno) {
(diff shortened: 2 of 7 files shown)


internal/sysfs/file_unix.go
```diff
@@ -19,3 +19,13 @@ func readFd(fd uintptr, buf []byte) (int, syscall.Errno) {
 	errno := platform.UnwrapOSError(err)
 	return n, errno
 }
+
+// writeFd exposes syscall.Write.
```
codefromthecrypt, line 23: this is a pretty cool way to ensure the logic is shared. nice!

evacchi: (actually I though you did this first...)

codefromthecrypt: uhh anyway thanks 😎

internal/sysfs/file_unix.go
```diff
@@ -19,3 +19,13 @@ func readFd(fd uintptr, buf []byte) (int, syscall.Errno) {
...
+func writeFd(fd uintptr, buf []byte) (int, syscall.Errno) {
+	if len(buf) == 0 {
+		return 0, 0 // Short-circuit 0-len reads.
```
ydnar, line 26: Reads?

internal/sysfs/osfile.go
```diff
@@ -213,7 +213,12 @@ func (f *osFile) Readdir(n int) (dirents []fsapi.Dirent, errno syscall.Errno) {
 
 // Write implements the same method as documented on fsapi.File
 func (f *osFile) Write(buf []byte) (n int, errno syscall.Errno) {
-	if n, errno = write(f.file, buf); errno != 0 {
+	if len(buf) == 0 {
+		return 0, 0 // Short-circuit 0-len reads.
```
ydnar, line 217: Reads?

ydnar (changes requested):

internal/sysfs/file_unix.go
evacchi: heh, lazy copypaste :)

evacchi:
the windows failure I know to handle (i.e. for now nothing, I'll include the fix in #1579)

imports/wasi_snapshot_preview1/testdata/gotip/wasi.go
```diff
@@ -259,3 +264,46 @@ func mainStdin() error {
 func mainStdout() {
 	os.Stdout.WriteString("test")
 }
+
+func mainLargeStdout() {
```
evacchi, line 268: this is the same test as [bug449.go](https://github.com/golang/go/blob/b88bd917b8dd8aabc471b1de41b3ee8c0d6eeabe/test/fixedbugs/bug449.go) except here it doesn't show the issue, because I cannot reproduce the same exact setup. I think the trick is that the test case in the Go repo is run via `exec` on a `wazero` binary, while here we are instantiating wazero without forking.

evacchi: resolved by actually forking

mathetake (approved):

Issue #1578, opened 2023-07-12
Plan9 compilation support


SlashScreen:
I am attempting to compile a program using Wazero to Plan9, since Go's compiler supports the target, however, pretty much everything related to Errno and Syscalls are broken, filled with Undefineds.
If it would be possible, I would like to have a mode where it can successfully compile to GOOS=plan9.
I am not adverse to implementing this change myself in a PR, however, I have no idea if this would be possible or not, nor where to start. I understand my use case is niche, but if I could get this working, it would potentially allow programs compiled to WASM to run on another OS, and I would be grateful.

codefromthecrypt:
So, I'm really happy you brought this up! I figured we wouldn't be lucky enough to expose syscall.X types in our 
portable internal interface.

Here's what I think needs to be done. In internal/fsapi/constants.go add `Errno` which defaults to `syscall.Errno`, where possible. Then forward all the values we use to that type. e.g. ` EAGAIN = syscall.EAGAIN` where syscall.Errno is around, and similar to how we treat `O_DIRECTORY` on windows, make fake values for plan9 that doesn't define them.

Then, change the interfaces in the fsapi package to return `sys.Errno` instead of `syscall.Errno` and I think we're ok.

cc @evacchi @achille-roussel for other input.

```bash
$ GOARCH=amd64 GOOS=plan9 go build ./...
# github.com/tetratelabs/wazero/internal/wasip1
internal/wasip1/errno.go:266:28: undefined: syscall.Errno
internal/wasip1/errno.go:272:15: undefined: syscall.EAGAIN
internal/wasip1/errno.go:274:15: undefined: syscall.EBADF
internal/wasip1/errno.go:278:15: undefined: syscall.EFAULT
internal/wasip1/errno.go:288:15: undefined: syscall.ELOOP
internal/wasip1/errno.go:294:15: undefined: syscall.ENOSYS
internal/wasip1/errno.go:298:15: undefined: syscall.ENOTEMPTY
internal/wasip1/errno.go:300:15: undefined: syscall.ENOTSOCK
internal/wasip1/errno.go:302:15: undefined: syscall.ENOTSUP
internal/wasip1/errno.go:306:15: undefined: syscall.EROFS
internal/wasip1/errno.go:306:15: too many errors
# github.com/tetratelabs/wazero/internal/fsapi
internal/fsapi/file.go:46:25: undefined: syscall.Errno
internal/fsapi/file.go:60:28: undefined: syscall.Errno
[...]

PR #1582, opened 2023-07-14
adds experimental sys.Errno to begin decoupling from the syscall package

codefromthecrypt:
This cherry-picks errors we use in our wasip1 implementation as `fsapi.Errno`. This is to start progress towards compiling on all GOOS values (such as plan9), to make future changes a lot easier.

See #1578


internal/fsapi/syscall_errno.go
```diff
@@ -0,0 +1,98 @@
+package fsapi
+
+import "syscall"
+
+func syscallToErrno(errno syscall.Errno) Errno {
```
codefromthecrypt, line 5: in the future, we can make a build constraint on this for GOOS that don't have syscall.Errno. I really don't know which that is.

experimental/sys/errno.go
```diff
@@ -0,0 +1,95 @@
...
+import "strconv"
+
+// Errno is a subset of POSIX errno used by wazero interfaces. Zero is not an
+// error. Other values should not be interpreted numerically, rather by constants
+// prefixed with 'E'.
+//
+// See https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html
+type Errno uint16
+
+// ^-- Note: This will eventually move to the public /sys package. It is
```
codefromthecrypt, line 12: Here's the note on why, but TL;DR; we already have an internal/sys package but we need one like the public sys one (shared symbols regardless of filesystem, socket, proc, etc.) By placing this into experimental, we can use the same package name and in a lot of cases refer to these errors as `sys.Errno` and especially in case of docs, this name will prevent drift mistakes, as eventually it will be the public package `sys`

imports/wasi_snapshot_preview1/sock.go
```diff
@@ -50,7 +51,7 @@ var sockRecv = newHostFunc(
 	"fd", "ri_data", "ri_data_len", "ri_flags", "result.ro_datalen", "result.ro_flags",
 )
 
-func sockRecvFn(_ context.Context, mod api.Module, params []uint64) syscall.Errno {
+func sockRecvFn(_ context.Context, mod api.Module, params []uint64) sys.Errno {
```
codefromthecrypt, line 54: I noticed we haven't audited POSIX, wasi-libc and wasix to figure out all the errno possibilities these should return. I'll do that afterwards. Particularly our errno should have enough values to support outbound sockets like https://github.com/stealthrocket/net

OTOH we shouldn't go too crazy on adding every definable one, as it makes a lot of work for those who aren't using posix systems. We should only add ones our apis say are returnable iotw, at least until our APIs are stable. The, as these are added, it is easier in PR review to ask "did you test this? where is this consumed in wasip1, wasix etc?" This question is harder when every errno is predefined.

internal/wasip1/errno.go
```diff
@@ -263,47 +264,47 @@ var errnoToString = [...]string{
 // Note: Coercion isn't centralized in sys.FSContext because ABI use different
 // error codes. For example, wasi-filesystem and GOOS=js don't map to these
 // Errno.
-func ToErrno(errno syscall.Errno) Errno {
+func ToErrno(errno sys.Errno) Errno {
```
codefromthecrypt, line 267: This file as all the errnos we currently use. This plus the ones in GOOS=js were the ones defined so far.

mathetake (approved):

PR #1584, opened 2023-07-15
sysfs: cleanup windows rename

codefromthecrypt:
This makes windows rename implementation cheaper and also consolidates errno conversion. It is helpful to consolidate more before we expose our own Errno in #1582

Minor note, but this also unexports a few symbols that were only used in the same package.


codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/1582 now passes (dependent on this PR)

PR #1585, opened 2023-07-17
wasi: add WASI FS constants for wasip1 and plan9

evacchi:
add some missing constants to improve cross-platform build support

### DIFF
--- a/internal/fsapi/constants.go
+++ b/internal/fsapi/constants.go
@@ -1,4 +1,4 @@
-//go:build !windows && !js && !illumos && !solaris
+//go:build !windows && !js && !illumos && !solaris && !plan9 && !wasip1
 
 package fsapi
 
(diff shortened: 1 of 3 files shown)


mathetake:
wanna add plan9 to https://github.com/tetratelabs/wazero/blob/main/Makefile#L211-L230 at least for testing?

codefromthecrypt:
right now I think it is a bit confusing in our docs to have partially os.O_ flags, partially syscall.O_ flags and partially package scoped ones. I had a WIP locally to define them all.. all the ones we accept. as it is less messy (though agree this could get things to compile). Either way I don't think we should add plan 9 constants until the whole thing can compile (GOOS=plan9 in the Makefile like the others)

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/1586 takes a different approach. When we later solve plan9 we can add constants for its oflags here. However, let's first finish decoupling from syscall first. Once that's done, plan9 can be solved for all things not just oflag.

PR #1586, opened 2023-07-19
fsapi: adds Oflag to decouple from syscall package

codefromthecrypt:
This adds `fsapi.Oflag` which includes oflag values needed to support `fsapi.FS`. Notably, this covers more area than Go does. Go defines portable abstractions for most open flags, but not all. Flags like `O_DIRECTORY` are needed by wasip1 and not defined in Go as `os.O_DIRECTORY`. Also, some operating systems don't define these either. Instead we have a new `fsapi.Oflag` type which includes all flags needed, even if notably in windows behavior is emulated.


internal/fsapi/oflag.go
```diff
@@ -0,0 +1,70 @@
...
+// interpreted numerically. Instead, use by constants prefixed with 'O_' with
+// special casing noted below.
+//
+// # Notes
+//
+//   - O_RDONLY, O_RDWR and O_WRONLY are mutually exclusive, while the other
+//     flags can coexist bitwise.
+//   - This is like `flag` in os.OpenFile and `oflag` in POSIX. See
+//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/open.html
+type Oflag uint32
```
codefromthecrypt, line 13: here is the new type

evacchi (approved):

PR #1587, opened 2023-07-21
Moves UTIME constants to fsapi package

codefromthecrypt:
This begins, but doesn't complete, syscall package decoupling around times. The first step is to move the constants `UTIME_NOW` and `UTIME_OMIT` to the fsapi package. While the real values vary a bit, at the cost of banishing updates to exclude one or two nanoseconds before epoch, we can portably handle unknown operating systems without codebase updates.


codefromthecrypt:
abandoning this approach for os.Chtimes emulation

codefromthecrypt:
obviated by #1602

PR #1588, opened 2023-07-21
Emulates AT_SYMLINK_NOFOLLOW instead of sometimes implementing it

codefromthecrypt:
futimes `AT_SYMLINK_NOFOLLOW` isn't universally supported. Before, we attempted to support it, but it was a bit hairy. It is better to take the hit of opening the file in the case we need to *not* follow symlinks than deal with something not very portable by default.

### DIFF
--- a/internal/sysfs/futimens_darwin.go
+++ b/internal/sysfs/futimens_darwin.go
@@ -6,22 +6,18 @@ import (
 )
 
 const (
-	_AT_FDCWD               = -0x2
-	_AT_SYMLINK_NOFOLLOW    = 0x0020
-	_UTIME_NOW              = -1
-	_UTIME_OMIT             = -2
-	SupportsSymlinkNoFollow = true
+	_AT_FDCWD            = -0x2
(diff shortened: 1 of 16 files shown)


internal/sysfs/futimens_darwin.go
```diff
@@ -6,22 +6,18 @@ import (
...
+	_UTIME_NOW           = -1
+	_UTIME_OMIT          = -2
 )
 
 //go:noescape
 //go:linkname utimensat syscall.utimensat
 func utimensat(dirfd int, path string, times *[2]syscall.Timespec, flags int) error
 
-func utimens(path string, times *[2]syscall.Timespec, symlinkFollow bool) error {
+func utimens(path string, times *[2]syscall.Timespec) error {
```
codefromthecrypt, line 19: a later PR will switch this to os.Chtimes which is basically the same logic (once we remove the special OMIT and NOW constants)

evacchi (approved):

PR #1591, opened 2023-07-22
docs: replace relative path

sho-hata:
If you run the example of trying the wazero CLI via Docker as is, it will fail.
```
docker: Error response from daemon: create ./testdata/: "./testdata/" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'
```
This is because relative paths cannot be specified for docker volumes.
Specifying the `$PWD` environment variable allows the reader to try it out smoothly.


achille-roussel (approved):
Thanks for the fix!

PR #1593, opened 2023-07-24
Fix grammar in comment about peeking

anuraaga:
Very minor just happened to notice while reading. I think this was meant to say `, not pop`, but looking through the file, there are far more comments just of the `Note this is peek!` form so aligned with that


mathetake (approved):
🍧 ☀️

PR #1596, opened 2023-07-26
sysfs: implements PollRead with poll instead of select

evacchi:
In all usages, it looks like `poll` should be preferred to `select`: the API is simpler and easier to emulate on other platforms; above all, it does not use `FdSet`s and won't change pointers in-place. 

We remove all usages of `select` and `FdSet` and instead expose and use `poll` and `[]PollFd`, which is comparatively much more straightforward. We also rewrite the `select` emulation on Windows, and instead emulate `poll`, based on the same principle: we use `WSAPoll` for sockets, `PeekNamedPipes` for pipes, and report all regular files as "ready". 

This in turn should eventually make it possible to remove `File.PollRead` or at the very least simplify further `poll_oneoff`.


internal/sysfs/poll.go
```diff
@@ -0,0 +1,35 @@
...
+//
+//	Because this is a blocking syscall, it will also block the carrier thread of the goroutine,
+//	preventing any means to support context cancellation directly.
+//
+//	There are ways to obviate this issue. We outline here one idea, that is however not currently implemented.
+//	A common approach to support context cancellation is to add a signal file descriptor to the set,
+//	e.g. the read-end of a pipe or an eventfd on Linux.
+//	When the context is canceled, we may unblock a Select call by writing to the fd, causing it to return immediately.
+//	This however requires to do a bit of housekeeping to hide the "special" FD from the end-user.
+func Poll(fds []PollFd, timeout *time.Duration) (int, sys.Errno) {
```
achille-roussel, line 27: Did you consider using a negative value as "no timeout" instead of a pointer?

internal/sysfs/poll_unsupported.go
```diff
@@ -0,0 +1,23 @@
+//go:build !linux && !darwin && !windows
```
achille-roussel, line 1: Did we support FreeBSD at some point?

internal/sysfs/poll.go
evacchi: TBH I just used the duration ptr only because select was like that. Since the underlying call uses negative values that way, I guess it's doable

internal/sysfs/poll_unsupported.go
evacchi: We never did for select, but fair point

internal/sysfs/poll.go
mathetake: that's using non pointer would be worth a shot since that way we could reduce allocation

codefromthecrypt: I'm not sure how I messed this up as certainly the original API should used normal numeric type for duration, as that maps to the underlying syscall. I'll fix PollRead in another PR

internal/sysfs/poll.go
```diff
@@ -0,0 +1,35 @@
+package sysfs
+
+import (
+	"time"
+
+	"github.com/tetratelabs/wazero/experimental/sys"
+)
+
+// Poll exposes the poll(2) syscall.
```
codefromthecrypt, line 9: please follow the documentation style of File.PollRead including how it uses links and refers to functions like `poll` not poll(2). I forget who said this, but there's an implicit goal to write as if everything is written by the same person. This means same style, tests, documentation form. This makes it easier to navigate and also easier to tell if other code was looked at etc (e.g. if not you could ask.. did you look at X which is similar?) It doesn't prevent changing style. For example, I went back and changed all links to better open group ones earlier, then all, not just a few we coherently done. So, general rule of thumb is to write like the others, then if something becomes lifted into the main API, no one needs to re-do it, also.

internal/sysfs/poll.go
```diff
@@ -0,0 +1,35 @@
...
+//
+// # Notes on Parameters
+//
+//	We currently expose a pointer to a time.Duration instead of a plain integer.
+//	A nil pointer means "wait forever" whereas poll(2) uses a negative value instead.
+//	This may change in the future.
+//
+// # Notes on the Syscall
```
codefromthecrypt, line 17: I would use the same doc style as others, but add a special heading
```suggestion
// # Impact of blocking
```

codefromthecrypt (approved):
I messed up the other API, signature wise (duration not int64) and will fix in another PR. 

Meanwhile, this looks good! I was verbose about style stuff so that we can try to be coherent with other style by default. This helps the project grow and feel to users as if it was written by one person, even if that's never the case.

Cheers.

codefromthecrypt:
I opened https://github.com/tetratelabs/wazero/pull/1597 to rebase the existing implementation, which will drift this. If it merges today, I will rebase your PR!

codefromthecrypt (approved):
OK I reworked this to pare down the change around implementing PollRead with `poll` instead of `select`, which is the main part. Main thing I think waiting here is to revisit the RATIONALE which is about select instead of poll. More work can happen in future PRs, but be wary of platforms that don't support syscalls such as plan9. More on that below!

A future change can implement a multi-poll, but that wouldn't seem to make sense on a file because it has only one file descriptor. In fact I would probably suggest not doing this until forced to, because there are a lot of edge cases we have to test. Even here we have logic to handle many types of inputs, but the only one tested is a single fd on read. In other words, unsolicited 2p is to not let the impl grow too far in front of reqiuirements and lazy grow the impl only when tests and ABI moves to require them.

internal/sysfs/poll_unsupported.go
codefromthecrypt: ps generally speaking it is tricky because we only compile on freebsd (in `make check`) because the VM setup for freebsd was really flakey. What this means is that tests don't run against it for things and you can get ENOSYS just like you could for solaris or plan9. Because there is some support for freebsd (e.g. mmap on the compiler side), it is more confused basically.

internal/sysfs/poll_windows.go
```diff
@@ -0,0 +1,235 @@
...
+	// _POLLRDNORM subscribes to normal data for read.
+	_POLLRDNORM = 0x0100
+	// _POLLRDBAND subscribes to priority band (out-of-band) data for read.
+	_POLLRDBAND = 0x0200
+	// _POLLIN subscribes a notification when any readable data is available.
+	_POLLIN = (_POLLRDNORM | _POLLRDBAND)
+)
+
+// pollFd is the struct to query for file descriptor events using poll.
+type pollFd struct {
```
codefromthecrypt, line 27: note when I refactored things, I unexported everything that can be unexported. Let's do this always please!

codefromthecrypt:
build break seems unrelated

```
 /usr/bin/sh: line 1: osslsigncode: command not found
mingw32-make: *** [makefile:302: build/wazero_windows_amd64/wazero.exe.signed] Error 127
```

internal/sysfs/poll_windows.go
```diff
@@ -0,0 +1,235 @@
...
+// If n==0 it will wait for the given timeout duration, but it will return sys.ENOSYS if timeout is nil,
+// i.e. it won't block indefinitely. The given ctx is used to allow for cancellation,
+// and it is currently used only in tests.
+//
+// The implementation actually polls every 100 milliseconds (pollInterval) until it reaches the
+// given timeout (in millis).
+//
+// The duration may be negative, in which case it will wait indefinitely. The given ctx is
+// used to allow for cancellation, and it is currently used only in tests.
+func pollWithContext(ctx context.Context, fds []pollFd, timeoutMillis int32) (n int, errno sys.Errno) {
```
codefromthecrypt, line 67: I know this was lifted from the select impl, but I would highly recommend removing the context based part here as it is dead code. There is a lot of code to maintain and things like this can be distracting. Especially in internal code, let's resist adding code that looks cool but never used.

evacchi: I thought I used in tests as select was... Fair however; the Unix impl doesn't come with s context either...

evacchi:
further work would be generalize the use of poll in poll_oneoff (right now it is only used for stdin).

codefromthecrypt (approved):
RATIONALE LGTM

internal/sysfs/poll_windows.go
evacchi: removed

PR #1597, opened 2023-07-28
sysfs: changes PollRead to accept int32 timeoutMillis

codefromthecrypt:
This changes `PollRead` to accept int32 timeoutMillis instead of using a pointer. I made a mistake earlier on this API, which wasn't being direct about units and also using nil where it isn't used that way at the syscall layer.

See https://github.com/tetratelabs/wazero/pull/1596

### DIFF
--- a/imports/wasi_snapshot_preview1/poll_test.go
+++ b/imports/wasi_snapshot_preview1/poll_test.go
@@ -555,8 +555,12 @@ type neverReadyTtyStdinFile struct {
 }
 
 // PollRead implements the same method as documented on fsapi.File
-func (neverReadyTtyStdinFile) PollRead(timeout *time.Duration) (ready bool, errno experimentalsys.Errno) {
-	time.Sleep(*timeout)
+func (neverReadyTtyStdinFile) PollRead(timeoutMillis int32) (ready bool, errno experimentalsys.Errno) {
+	switch {
+	case timeoutMillis == 0 || timeoutMillis < 0:
+		return
+	}
+	time.Sleep(time.Duration(timeoutMillis) * time.Millisecond)
 	return false, 0
 }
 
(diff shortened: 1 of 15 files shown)


internal/sysfs/osfile.go
```diff
@@ -183,17 +182,22 @@ func (f *osFile) Seek(offset int64, whence int) (newOffset int64, errno experime
...
 // PollRead implements the same method as documented on fsapi.File
-func (f *osFile) PollRead(timeout *time.Duration) (ready bool, errno experimentalsys.Errno) {
+func (f *osFile) PollRead(timeoutMillis int32) (ready bool, errno experimentalsys.Errno) {
 	fdSet := platform.FdSet{}
 	fd := int(f.fd)
 	fdSet.Set(fd)
 	nfds := fd + 1 // See https://man7.org/linux/man-pages/man2/select.2.html#:~:text=condition%20has%20occurred.-,nfds,-This%20argument%20should
-	count, err := _select(nfds, &fdSet, nil, nil, timeout)
+
+	// Coerce timeout to what's
```
codefromthecrypt, line 191:
```suggestion
	// Coerce negative timeout to -1 as that's defined in POSIX
```

imports/wasi_snapshot_preview1/poll_test.go
```diff
@@ -555,8 +555,12 @@ type neverReadyTtyStdinFile struct {
 }
 
 // PollRead implements the same method as documented on fsapi.File
-func (neverReadyTtyStdinFile) PollRead(timeout *time.Duration) (ready bool, errno experimentalsys.Errno) {
-	time.Sleep(*timeout)
+func (neverReadyTtyStdinFile) PollRead(timeoutMillis int32) (ready bool, errno experimentalsys.Errno) {
+	switch {
+	case timeoutMillis == 0 || timeoutMillis < 0:
```
mathetake, line 560:
```suggestion
	case timeoutMillis <= 0:
```

mathetake (approved):

PR #1599, opened 2023-07-29
fsapi: migrates PollRead to Poll with Pflag

codefromthecrypt:
This migrates PollRead to Poll so that it can support read, write or a combination via flags. This should allow `poll_oneoff` to be implemented even if internally it may need to use a deadline decrementing approach to cycle through each file instead of via multiple at the same time.

The benefit of doing things this way is we don't need to top-level a concept of file descriptor, which complicates the design of `fsapi.FS` which is far easier to implement if there are no "multi-file" apis on it.

This intentionally doesn't implement the approach at the same time, as the goal is to create the api to implement. Doing so also puts a lot less pressure on the task to complete migration off syscall types.

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -616,6 +616,134 @@ act differently and document `ModuleConfig` is more about emulating, not necessa
 
 ## File systems
 
+### Motivation on `sys.FS`
+
+The `sys.FS` abstraction in wazero was created because of limitations in
+`fs.FS`, and `fs.File` in Go. Compilers targeting `wasip1` may access
+functionality that writes new files. The ability to overcome this was requested
+even before wazero was named this, via issue #21 in March 2021.
+
+A month later, golang/go#45757 was raised by someone else on the same topic. As
+of July 2023, this has not resolved to a writeable file system abstraction.
+
+Over the next year more use cases accumulated, consolidated in March 2022 into
+#390. This closed in January 2023 with a milestone of providing more
+functionality, limited to users giving a real directory. This didn't yet expose
+a file abstraction for general purpose use. Internally, this used `os.File`.
+However, a wasm module instance is a virtual machine. Only supporting `os.File`
+breaks sand-boxing use cases. Moreover, `os.File` is not an interface. Even
+though this abstracts functionality, it does allow interception use cases.
+
+Hence, a few days later in January 2023, we had more issues asking to expose an
+abstraction, #1013 and later #1532, on use cases like masking access to files.
+In other words, the use case requests never stopped, and aren't solved by
+exposing only real files.
+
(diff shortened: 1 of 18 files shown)


internal/sysfs/poll_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows || linux || darwin
```
codefromthecrypt, line 1: this was missing. we really need to finish #1578 because it is very easy to break it more, with each change to things that can use syscalls.

evacchi:
fixed the Windows build. 

Without a lower-level concept of fd/handle, we improve encapsulation/portability but we won't be able to multiplex I/O (e.g. in this specific case, we won't be able to `Poll([]{fd_1,...fd_n})`. Probably being concerned with performance is not worth it at this time (we are not really benchmarking it), but it's worth noting. Maybe something to add to `RATIONALE.md` (e.g. "why we encapsulate FDs" or something?)

evacchi (approved):

codefromthecrypt:
@evacchi I think by "why we encapsulate FDs" what you are asking is why do we expose a replacement for `fs.FS/fs.File`?

The primary use case of this abstraction is to be able to write delete, etc files. Currently, we continually break portability and if we don't stop doing that we can never expose it. I'll make notes on this.

codefromthecrypt:
Also, it would be helpful when discussing what won't work, is to be very concrete. For example, some existing case (real not forged, like test we are failing) that cannot be solved as a poll loop or as a goroutine externally. We have a tendency to over-solve a lot and by making failing tests it helps conserve energy.

codefromthecrypt:
@mathetake can you take a look? I had planned to do code and still have time today. I spent the last 3 hours explaining why our abstraction doesn't need to up front interfaces for every niche features, using multi-poll as an example of that.

RATIONALE.md
```diff
@@ -616,6 +616,134 @@ act differently and document `ModuleConfig` is more about emulating, not necessa
...
+In other words, the use case requests never stopped, and aren't solved by
+exposing only real files.
+
+In summary, the primary motivation for exposing a replacement for `fs.FS` and
+`fs.File` was around repetitive use case requests for years, around
+interception and the ability to create new files, both virtual and real files.
+While some use cases are solved with real files, not all are. Regardless, an
+interface approach is necessary to ensure users can intercept I/O operations.
+
+### Why doesn't `sys.File` have a `Fd()` method?
```
codefromthecrypt, line 648: note: I'm using `sys.File` as even though we may have forgotten, we did promise to expose our filesystem abstraction in the next release. Even if that ends up `experimentalsys.File`, the end name would be this.

mathetake (approved):

PR #1602, opened 2023-07-30
sysfs: decouples FS and File from the syscall package

codefromthecrypt:
This decouples FS and File from the syscall package by passing through atim and mtim as numbers instead of structs. This makes it easier to integrate with `os.Chtimes` as well possible to compile with plan9.

In doing so, this unsupports the `UTIME_NOW` constant, which was problematic anyway. wasip1 defines `UTIME_NOW` as the value of the real-time clock, which isn't necessarily the same as the filesystem clock. By unsupporting `UTIME_NOW` we instead read from the correct clock.

### DIFF
--- a/internal/sysfs/poll_windows.go
+++ b/internal/sysfs/poll_windows.go
@@ -42,9 +42,6 @@ const pollInterval = 100 * time.Millisecond
 
 // _poll implements poll on Windows, for a subset of cases.
 //
-// pollWithContext emulates the behavior of POSIX poll(2) on Windows, for a subset of cases,
-// and it supports context cancellation.
-//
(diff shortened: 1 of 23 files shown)


mathetake (approved):

internal/sysfs/poll_windows.go
```diff
@@ -42,9 +42,6 @@ const pollInterval = 100 * time.Millisecond
 
 // _poll implements poll on Windows, for a subset of cases.
 //
-// pollWithContext emulates the behavior of POSIX poll(2) on Windows, for a subset of cases,
-// and it supports context cancellation.
-//
```
evacchi, line 47: thanks for noticing this

evacchi (approved):

PR #1604, opened 2023-07-30
sysfs: restore check for EINTR in poll test, fix Windows tests

### LINKED ISSUE #1601: Flakey test: Test_poll/should_wait_for_the_given_duration
We should probably look out for interrupt and make sure the godoc declares it.

```
--- FAIL: Test_poll (0.00s)
    --- FAIL: Test_poll/should_wait_for_the_given_duration (0.00s)
        require.go:331: expected Errno 0x0(success), but was 0x6(interrupted function)
            /home/runner/work/wazero/wazero/internal/sysfs/poll_test.go:30
FAIL
```

failed on scratch Go 1.21rc3

https://github.com/tetratelabs/wazero/actions/runs/5703851221/job/15456906175?pr=1599

evacchi:
EINTR used to be ignored when `select` was used, but the check was
deleted during the migration to `poll`: looks like this made some tests flaky.

Some time-based tests for Windows also seem flakier, so we try to make them slightly more robust.

Fixes #1601 

### DIFF
--- a/internal/sysfs/poll_test.go
+++ b/internal/sysfs/poll_test.go
@@ -8,40 +8,56 @@ import (
 	"testing"
 	"time"
 
+	"github.com/tetratelabs/wazero/experimental/sys"
 	"github.com/tetratelabs/wazero/internal/testing/require"
 )
 
 func Test_poll(t *testing.T) {
 	t.Run("should return immediately with no fds and duration 0", func(t *testing.T) {
-		n, err := _poll([]pollFd{}, 0)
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
I recall once us looping on EINTR internally, such as Go does in functions that call `os.ignoringEINTR`, for reasons including avoiding excess host calls. I wonder if that's appropriate in osFile or not for File.Poll? Such an adjustment can be now or later or never

PR #1606, opened 2023-07-31
wasi: use File.Poll for all blocking FDs in poll_oneoff

### LINKED ISSUE #1500: support non-blocking files

evacchi:
This includes stdin, pipes but also sockets. 

- This also updates RATIONALE.md. 
- It also fixes some issues in the fsapi.File wrappers for sockets on POSIX and Windows (esp. the nonblocking flag was set correctly at all times)
- I think with this we could also close #1500. 

Further refinements are possible (e.g. supporting poll for reading), but most of the work is ok now.

## Rationale
Instead of special-casing for stdin, we can now use `File.Poll()` on basically every file. I am limiting to those that report  `blocking = true`, because otherwise `Read()` is expected... not to block (i.e. it is allowed to return EAGAIN).

Notice that, previously, we knew we only handled stdin, so we always invoked `Poll` only on that one File. 

- This uncovered another interesting issue: in some cases, the `event` struct could be written at a wrong offset, because we precomputed the value. If for some reason one of the events was not written back, then we would leave an empty gap; this was hard to notice because of the special treatment reserved only for blocking stdin, and because in many cases we did not test more than one fd at a time (we did not simulate an unreported event).
- The other issue is that, with a nonzero timeout, the `poll` syscall blocks an OS thread until it returns. 

Now, we iterate on each `File` that has reported as "blocking" to invoke its method `Poll()`.  However:

- we no longer issue a blocking call to `sysfs.poll()` with the given timeout. 
- instead, we emulate `time.After()` and `time.Tick()` but we use `sys.Nanosleep()`
- we repeatedly call `sysfs.poll()` with a zero timeout, similarly to how this is handled on the Windows side for `WSAPoll`, every time the tick goes off until the timeout is reached
- This allows us to honor the given `sys.Nanosleep()` config instead of relying on `poll` regardless of the config settings (in fact, some test cases were broken because they did not configure `sys.Nanosleep()` properly if at all.

Alternative ways to do this would be:

- scatter/gather by spawning a goroutine for each `File.Poll()` with a given timeout; however, if the given timeout _does_ leverage `sysfs.poll()` this in turns issues a blocking call, taking over the underlying OS thread until the timeout is reached (which eventually may consume all resources).
- call `poll({fd_1, ..., fd_n})` which however, as already discussed in RATIONALE.md, is not an abstraction at the right level.

In fact, for our intents, we may also consider exporting `File.Poll(Flag)` with no timeouts, defaulting to 0ms, avoiding the risk of blocking altogether.

Finally, because emulating `poll(2)` on all platforms is not a goal, it might be possible to further refine this by replacing the Windows implementation of `sysfs.poll()`  with ad-hoc versions basically remove the wrapper (e.g. without handling an arbitrary timeout, since this would be now handled in `poll_oneoff` and/or for specific file types instead of detecting them automatically _in_ `sysfs.poll()`; in other words possibly going straight from `WinTcp*File.Poll()` to `WSAPoll()` etc.)


imports/wasi_snapshot_preview1/poll.go
```diff
@@ -161,44 +166,34 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno
...
 
-	// If there are blocking stdin subscribers, check for data with given timeout.
-	if len(blockingStdinSubs) > 0 {
-		stdin, ok := fsc.LookupFile(internalsys.FdStdin)
-		if !ok {
-			return sys.EBADF
-		}
-		// Wait for the timeout to expire, or for some data to become available on Stdin.
-		stdinReady, errno := stdin.File.Poll(sys.POLLIN, int32(timeout.Milliseconds()))
+	// There are no blocking subscribers, we just wait for the given timeout.
```
codefromthecrypt, line 171: I know this is old logic, but is the intent to wait for the timeout? is the assumption only clock events left, or does this loop? Or does wasi-libc expect the same? Might be worth clarifying why an unconditional timeout?

evacchi: yes the assumption is that only clock events are left, the timeout is initialized earlier at lines ~122

https://github.com/tetratelabs/wazero/blob/1f8c908f1c61c964192c731af495caf1d720418c/imports/wasi_snapshot_preview1/poll.go#L123-L126

I will update the comment to be more explicit about it.

evacchi: I have now refactored all the logic and added a better explanation

evacchi:
I added more test cases, and I realized some issues with file sock on both Windows and POSIX re: setting the nonblocking flag. I also added a wasi test case with zig-cc. Tomorrow I'll try to figure out if I can write another implementation in Rust and/or Go that _actually_ invokes poll with multiple FDs (I am not sure the higher-level APIs actually do it)

codefromthecrypt:
Thanks for digging into the integration tests. This type of code/behavior is hard to pin down and exactly where the extra tests come in: to establish an "implementation quorum" please ping back when you feel things are settled or need a hand from someone else (even if technical over my head ;))

evacchi:
Ok, I added a test for gotip,  and I have also figured out something for Rust (using [tokio-rs/mio](https://github.com/tokio-rs/mio)).

- I verified (with `--hostlogging poll` and also via debugger) they actually exercise `poll_oneoff`. 
- The tests are not all the same but they are ~similar. 
- The one that's actually most different is the one for gotip, because I can't tell if there is a straightforward way to make sure that all FDs are checked at once (spoiler: they aren't!), but at least it's checking more than one subscription at a time -- but they will all be in nonblocking mode, so it doesn't really follow the new code path, except for the timers.

I think at this point this is ready for review. It may still lack a bit of polish but your feedback is welcome.

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -125,24 +128,25 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno
...
 			if file, ok := fsc.LookupFile(fd); !ok {
 				evt.errno = wasip1.ErrnoBadf
-				writeEvent(outBuf, evt)
-				readySubs++
-				continue
-			} else if fd == internalsys.FdStdin && !file.File.IsNonblock() {
-				// if the fd is Stdin, and it is in non-blocking mode,
-				// do not ack yet, append to a slice for delayed evaluation.
-				blockingStdinSubs = append(blockingStdinSubs, evt)
+				writeEvent(outBuf[outOffset:], evt)
```
evacchi, line 140: writeEvents has been simplified, we pass the buffer at the right offset already

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -154,47 +158,46 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno
...
 		default:
 			return sys.EINVAL
 		}
 	}
 
-	// If there are subscribers with data ready, we have already written them to outBuf,
-	// and we don't need to wait for the timeout: clear it.
-	if readySubs != 0 {
-		timeout = 0
+	sysCtx := mod.(*wasm.ModuleInstance).Sys
```
evacchi, line 168: the section below has been reordered for clarity

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -125,24 +128,25 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno
...
+			} else if file.File.IsNonblock() {
+				writeEvent(outBuf[outOffset:], evt)
+				nevents++
 			} else {
-				writeEvent(outBuf, evt)
-				readySubs++
+				// If the fd is blocking, do not ack yet,
+				// append to a slice for delayed evaluation.
+				fe := &filePollEvent{f: file, e: evt}
+				blockingSubs = append(blockingSubs, fe)
```
evacchi, line 149: these have been reordered for clarity; first we avoid the double negation (`!IsNonblock()`), second the two immediate writes are in the `if + else if`, while `else` handles the special case of "delayed" processing.

imports/wasi_snapshot_preview1/poll_test.go
```diff
@@ -522,6 +810,10 @@ func fdReadSubFd(fd byte) []byte {
 		0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, // userdata
 		wasip1.EventTypeFdRead, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
 		fd, 0x0, 0x0, 0x0, // valid readable FD
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0,
```
evacchi, line 816: `fdReadSubFd` is now being used in other poll tests (see above); e.g. to create multiple records; in order for such records to be valid, we zero-pad the byte slice to the right size (32 bytes)

internal/sysfs/sock_windows.go
```diff
@@ -100,7 +100,6 @@ func syscallConnControl(conn syscall.Conn, fn func(fd uintptr) (int, sys.Errno))
 // because they are sensibly different from Unix's.
 func newTCPListenerFile(tl *net.TCPListener) socketapi.TCPSock {
 	w := &winTcpListenerFile{tl: tl}
-	_ = w.SetNonblock(true)
```
evacchi, line 103: we no longer default to nonblock on sockets (it's not necessary)

internal/sysfs/sock_windows.go
```diff
@@ -116,14 +115,11 @@ type winTcpListenerFile struct {
...
-	})
-
-	// Otherwise return immediately.
-	if n == 0 || errno != 0 {
-		return nil, sys.EAGAIN
+	// Ensure we have an incoming connection using winsock_select, otherwise return immediately.
+	if f.nonblock {
+		if ready, errno := f.Poll(sys.POLLIN, 0); !ready || errno != 0 {
+			return nil, sys.EAGAIN
+		}
```
evacchi, line 122: we can rewrite this in terms of f.Poll()

evacchi:
oh, I was almost forgetting: while running `make build.examples.zig-cc` I noticed that the DWARF example was mistakingly being overwritten (I think the new build might have stripped the DWARF symbols) so we'll need to check that

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -233,10 +236,60 @@ func processClockEvent(inBuf []byte) (time.Duration, sys.Errno) {
...
+}
+
+// pollFileEventsUntil repeatedly invokes pollFileEventsOnce until the given timeout is reached.
+// The poll interval is currently fixed at 100 millis.
+func pollFileEventsUntil(sysCtx *internalsys.Context, timeout time.Duration, blockingSubs []*filePollEvent, outBuf []byte) (n uint32, errno sys.Errno) {
+	timeoutCh := make(chan struct{}, 1)
+	go closeChAfter(sysCtx, timeout, timeoutCh)
+
+	pollInterval := 100 * time.Millisecond
+	ticker := time.NewTicker(pollInterval)
```
codefromthecrypt, line 280: fyi closeChAfter we are intentionally using the context clock, but this will use a real one..

RATIONALE.md
```diff
@@ -1457,15 +1453,18 @@ which plays nicely with the rest of the Go runtime.
...
-not currently implemented. A common approach to support context cancellation is
-to add a signal file descriptor to the set, e.g. the read-end of a pipe or an
-eventfd on Linux. When the context is canceled, we may unblock a Select call by
-writing to the fd, causing it to return immediately. This however requires to
-do a bit of housekeeping to hide the "special" FD from the end-user.
+Because this is a blocking syscall, invoking it with a nonzero timeout will also
+block the carrier thread of the goroutine, preventing any means
+to support context cancellation directly.
+
+We obviate this by invoking `poll` with a 0 timeout repeatedly,
```
codefromthecrypt, line 1460: I don't fully understand this. why not poll with 100ms vs poll zero+sleep? Are you suggesting that the poll implementation blocks too long even if you put 100ms? If so maybe the above paragraph needs to clarify this?

imports/wasi_snapshot_preview1/poll_test.go
codefromthecrypt:
```suggestion
		0x0, 0x0, 0x0, 0x0, // pad to record size (32 bytes)
```

codefromthecrypt (commented):
This looks good, just I might be missing a decision point around the poll interval. Right now, we seem to be polling via poll immediate followed by 100ms sleep.

100ms is a very long time, so I would suspect this should be a lot shorter, probably 100us even could be too long). It is worse because external sleep approach guarantees it will take that long.

So, I'm wondering mainly why not use poll with a short timeout isn't used, if it is a defect or we are trying for the native side to use the fake clock.

IMHO I think that since timeout is a parameter of poll, in the Poll api, how timeout is implemented is up to the backend, which may choose to use a real or fake clock to sleep or a native poll. If in any case we are not able to trust the implementation of the poll timeout and avoiding using it for that reason, I would try to make it very clear why not, because in worst case it can feel like "spaghetti around a problem" to do externnal orchestration of a feature defined in the poll documentation (timeout)

imports/wasi_snapshot_preview1/poll.go
evacchi: whoops!

RATIONALE.md
evacchi: yes, if we put a 100ms timeout, then the syscall will block for 100ms, which means it will also block the underlying OS thread. I will add a clarification.

evacchi:
> 100ms is a very long time, [...]
> So, I'm wondering mainly why not use poll with a short timeout isn't used, if it is a defect or we are trying for the native side to use the fake clock.

100ms is completely arbitrary, I picked that because that's what I picked in the Windows impl (which now is irrelevant, since it is being invoked here with 0ms). We could certainly use a smaller delay, I am not an expert at all here. 

> IMHO I think that since timeout is a parameter of poll, in the Poll api, how timeout is implemented is up to the backend, which may choose to use a real or fake clock to sleep or a native poll. If in any case we are not able to trust the implementation of the poll timeout and avoiding using it for that reason, I would try to make it very clear why not, because in worst case it can feel like "spaghetti around a problem" to do externnal orchestration of a feature defined in the poll documentation (timeout)

yeah the main issue is that the real poll is actually hogging a real OS thread, that means e.g. if you invoke poll_oneoff with a long delay it won't return control to the Go runtime until the underlying `poll` has returned. This way we avoid relying on the syscall native delay, and we give the Go runtime more chances to take over if necessary (e.g. schedule other goroutines)

so this has more to do with the interaction between Go and the underlying syscall, than how the syscall is actually implemented at the OS-level. I will add some notes to the RATIONALE

codefromthecrypt:
>  This way we avoid relying on the syscall native delay

What I'm probably missing here is that this is a timeout. What I'm thinking and could be wrong, but timeout is the worst case. A delay means it is blocked regardless.

So say 100us and the "file is ready to write" event happens at +10us. Using a clock sleep approach you have to wait extra 90us anyway. If this is correct then it seems we are trying too hard to not use poll's timeout even when Go uses it. In other words, I feel we are trying too hard to not use it, and in the process force a longer delay than necessary. What am I missing?

ncruces:
My understanding of the problem/solution is this:

If you poll with timeout:
- the guest unblocks as soon as an event happens; _but while waiting_
- the host has a goroutine blocked from doing anything else;
- the host can only cancel the guest at timeout intervals.

If you poll zero, and sleep the timeout:
- the guest can only unblock at timeout intervals; _but while sleeping_
- the host goroutine is free to do other things;
- the host can cancel the guest at any time.

Basically this looks like a choice/balance between giving priority to host or guest resources.

evacchi:
since this is introducing a significant change in how we handle `poll_oneoff`, I will close this PR for now and instead contribute the tests, small cleanups and fixes that were part of it, without modifying how `poll_oneoff` handles FDs, we can always revisit this :)

codefromthecrypt:
https://github.com/tetratelabs/wazero/pull/1606#issuecomment-1665475516 from @ncruces has basically the concern which was at the crux of this issue. I would say that the code abandoned and also in the description seems to both say that using blocking via ssyscall is bad, and I believe this is actually what go does in net poll. It makes me wonder how anyone would be able to choose a good value to block the client. Also why anyone would do this if there was only one blocking event.

If the project as a whole wants to prevent use of the timeout parameter in the syscall, basically to always block for no time and guess an interval to sleep for (with a syscall each guess).. I feel like the API should change, and remove the ability to give a timeout. It makes even less sense to have a timeout parameter and not use it, than how I currently perceive the sense in intentionally preventing an event from unblocking as defined in poll.

I think personally I've said enough on this and I'll leave any decision up to you all, just maybe decide once and for all in a couple months? because we are exposing the filesystem api and it should make sense why poll would be exposed and never use the timeout val here or in a potential multi-poll scenario.

codefromthecrypt:
The thing I believe whatever action to take is, that if we believe wazero should not support integrated syscall pollers, we should be very loud about it. It is a different direction than I expected considering the syscall layer otherwise acts like sys calls. I was expecting a comment like what go has on an emulated thing vs emulating poll behavior and intentionally not allowing native polling, by doing so above where someone has control of the impl (above the fs API) 

Basically, I would suggest study the topic in go, like [here](https://github.com/golang/go/blob/5d5305026ec51ff3aaa6e7a7c18a815a3ac787c7/src/runtime/netpoll_stub.go#L37-L38), make a decision and then try to rationalize both in API and also in the RATIONALE why specifically here only in poll we are doing like this, where other places like blocking Read will still block the calling thread etc. I've an idea that folks can figure out a justification, just I don't want to moderate it.

paralin:
See #1500 I think this is worth re-visiting.

PR #1607, opened 2023-07-31
sysfs: Makes ReadFS and AdaptFS embeddable

codefromthecrypt:
This exports the experimental `sysfs.ReadFS` and `sysfs.AdaptFS` types, to developers can choose to embed them. This makes certain decoration use cases easier.

Note: We can't do the same without a risk of drift in `sysfs.DirFS`, but that isn't likely a problem as people are used to `os.DirFS` which also doesn't export its type.

### DIFF
--- a/cmd/wazero/wazero.go
+++ b/cmd/wazero/wazero.go
@@ -419,9 +419,9 @@ func validateMounts(mounts sliceFlag, stdErr logging.Writer) (rc int, rootPath s
 			fmt.Fprintf(stdErr, "invalid mount: path %q is not a directory\n", dir)
 		}
 
-		root := sysfs.NewDirFS(dir)
+		root := sysfs.DirFS(dir)
 		if readOnly {
-			root = sysfs.NewReadFS(root)
+			root = &sysfs.ReadFS{FS: root}
 		}
 
 		config = config.(sysfs.FSConfig).WithSysFSMount(root, guestPath)
--- a/experimental/sys/unimplemented.go
+++ b/experimental/sys/unimplemented.go
@@ -10,16 +10,6 @@ import (
 // This should be embedded to have forward compatible implementations.
 type UnimplementedFS struct{}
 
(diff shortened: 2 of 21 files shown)


cmd/wazero/wazero.go
```diff
@@ -419,9 +419,9 @@ func validateMounts(mounts sliceFlag, stdErr logging.Writer) (rc int, rootPath s
 			fmt.Fprintf(stdErr, "invalid mount: path %q is not a directory\n", dir)
 		}
 
-		root := sysfs.NewDirFS(dir)
+		root := sysfs.DirFS(dir)
```
codefromthecrypt, line 422: reamed to be more like `os.DirFS`

experimental/sys/unimplemented.go
```diff
@@ -80,11 +70,6 @@ func (UnimplementedFS) Utimens(path string, atim, mtim int64) Errno {
 	return ENOSYS
 }
 
-// Truncate implements FS.Truncate
-func (UnimplementedFS) Truncate(string, int64) Errno {
```
codefromthecrypt, line 84: fuzz as we removed this from FS a while ago

internal/sysfs/adapter.go
```diff
@@ -51,24 +46,56 @@ func (a *adapter) Stat(path string) (sys.Stat_t, experimentalsys.Errno) {
...
+func (a *AdaptFS) Utimens(string, int64, int64) experimentalsys.Errno {
+	return experimentalsys.ENOSYS
 }
 
 func cleanPath(name string) string {
 	if len(name) == 0 {
 		return name
 	}
-	// fs.ValidFile cannot be rooted (start with '/')
+	// FS.ValidFile cannot be rooted (start with '/')
```
codefromthecrypt, line 98:
```suggestion
	// fs.ValidFile cannot be rooted (start with '/')
```

evacchi (approved):

PR #1608, opened 2023-08-01
compiler: fix compiledModule leak

ncruces:
Fixes #1600.

I'm not 100% sure why this is needed, but it's being done as well in:
https://github.com/tetratelabs/wazero/blob/1f8c908f1c61c964192c731af495caf1d720418c/internal/engine/compiler/engine.go#L495-L508

Somehow, not doing this prevents the `releaseCompiledModule` finalizer from running.

Issue #1600, opened 2023-07-29
binary.decodeCode cause Memory leak ?


panchen66:
**Describe the bug**
I suspect make binary.decodeCode cause Memory leak

**To Reproduce**
I will reinitialize the instance when I catch wasm execution exceptions, but after continuous initialization, I find that memory continues to grow

```(pprof) top
Showing nodes accounting for 2560.42MB, 92.16% of 2778.31MB total
Dropped 223 nodes (cum <= 13.89MB)
Showing top 10 nodes out of 68
      flat  flat%   sum%        cum   cum%
 1231.40MB 44.32% 44.32%  1231.40MB 44.32%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeCode
  323.56MB 11.65% 55.97%   411.41MB 14.81%  github.com/tetratelabs/wazero/internal/engine/compiler.(*engine).CompileModule
  283.46MB 10.20% 66.17%   349.47MB 12.58%  github.com/tetratelabs/wazero/internal/wasm.(*Module).buildFunctionDefinitionsOnce
  194.70MB  7.01% 73.18%   195.20MB  7.03%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeDataSegment
  187.75MB  6.76% 79.94%  1419.15MB 51.08%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeCodeSection
   88.01MB  3.17% 83.10%    88.01MB  3.17%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeUTF8
   82.12MB  2.96% 86.06%    96.15MB  3.46%  github.com/tetratelabs/wazero/internal/wasm.addFuncs
   71.23MB  2.56% 88.62%   156.24MB  5.62%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeFunctionNames
   65.50MB  2.36% 90.98%    65.50MB  2.36%  strings.(*Builder).WriteString
   32.69MB  1.18% 92.16%   227.89MB  8.20%  github.com/tetratelabs/wazero/internal/wasm/binary.decodeDataSection
(pprof) list binary.decodeCode
Total: 2.71GB
ROUTINE ======================== github.com/tetratelabs/wazero/internal/wasm/binary.decodeCode in /Users/pc/go/pkg/mod/github.com/tetratelabs/wazero@v1.3.0/internal/wasm/binary/code.go
    1.20GB     1.20GB (flat, cum) 44.32% of Total
[...]

PR #1609, opened 2023-08-02
test memory usage of the wazero engine

achille-roussel:
I'm opening this PR just to demonstrate that the test added in https://github.com/tetratelabs/wazero/pull/1608/commits/129e49ae84a1b635d767ff48246aefdb3015da19 fails when executed on `main`. 

The test reproduces the memory leak reported in #1600

### DIFF
--- a/internal/integration_test/engine/memleak_test.go
+++ b/internal/integration_test/engine/memleak_test.go
@@ -1,0 +1,51 @@
+package adhoc
+
+import (
+	"context"
+	"log"
+	"runtime"
+	"testing"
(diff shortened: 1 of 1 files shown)


mathetake:
ack!

achille-roussel:
For the record:
```
--- FAIL: TestMemoryLeak (5.15s)
    memleak_test.go:19: running memory leak test for 5s
    memleak_test.go:35: wazero used more than 100 MiB after running the test for 5s (alloc= 139795240)
```

PR #1611, opened 2023-08-04
Implement Emscripten jmp support in indirect function calls (invoke_xxx)

jerbob92:
This PR implements Emscripten jmp support from indirect function calls, which we have missed before. We missed this mainly due to me lazily patching out Emscripten methods in standalone builds, in this case `_emscripten_throw_longjmp`, which is a pretty vital part of doing setjmp/longjmp from indirect function calls.

How that seems to work in the JS version:
- C calls `longjmp` from an indirect function call
- This causes the C code to call `_emscripten_throw_longjmp`, which is a host method  (https://github.com/emscripten-core/emscripten/blob/6b6f7e7ed4aa8c1f2232443403fc8491056b01f6/system/lib/compiler-rt/emscripten_setjmp.c#L69),
- The host method then throws something identifiable ([JS chose for Infinit](https://github.com/emscripten-core/emscripten/blob/6b6f7e7ed4aa8c1f2232443403fc8491056b01f6/src/library.js#L1209)y, I chose a Go error)
- In the invoke function it implements stackSave/stackRestore that are exported by Emscripten, when the Indirect function call fails, it restores the stack, and when the error does not match the `_emscripten_throw_longjmp` type, it rethrows the error
- If the error does match up, it calls `setThrew`, probably to reset the throw status in C, because it wasn't actually a throw, it was a longjmp

I have tried to implement it in a very similar way in Wazero, but I have to say that I'm not aware of any edge cases where this does not work.

Sample C code:
```
#include <stdio.h>
#include <stdlib.h>
#include <setjmp.h>

void jmpfunction(jmp_buf env_buf) {
   longjmp(env_buf, 2);
}

int main() {
   int val;
   jmp_buf env_buffer;

   /* save calling environment for longjmp */
   val = setjmp( env_buffer );
   
   if( val != 0 ) {
      printf("Returned from a longjmp() with value = %d\n", val);
      exit(0);
   }
   
   printf("Jump function call\n");
   jmpfunction( env_buffer );

   return(0);
}
```


jerbob92:
I think the tests currently fail because the `invoke.wat`/`invoke.wasm` do not contain the needed methods (stackSave/stackRestore/setThrew) right now, however, they are always included when you have invoke_ methods, so we can assume that they are always there.

internal/emscripten/emscripten.go
```diff
@@ -89,8 +109,17 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 		panic(err)
 	}
 
-	err = f.CallWithStack(ctx, stack)
+	stackSave, err := mod.ExportedFunction("stackSave").Call(ctx)
```
ncruces, line 112: nit: I'd call this `stackPointer`, or `savedStack`; overloading the function name threw me off a bit.

internal/emscripten/emscripten.go
```diff
@@ -89,8 +109,17 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
...
 	if err != nil {
 		panic(err)
 	}
+
+	err = f.CallWithStack(ctx, stack)
+	if err != nil {
+		_, _ = mod.ExportedFunction("stackRestore").Call(ctx, stackSave[0])
```
ncruces, line 119: I'm not sure we can call this for all errors.
I know that's what the JS code does, but I'm not sure the module is always in a state where we can call this.
Like what happens if your sample calls `exit(1)`, do we go through this? Should we?

ncruces (commented):
Looks good. I assume the same from the failing tests.

`stackSave` seems easy to stub out (do nothing), `stackRestore` I have no idea if it's safe to just ignore.

ncruces:
Can you attach a wasm from built from your sample so can avoid figuring out how to build it? 😳

jerbob92:
> Looks good. I assume the same from the failing tests.
> 
> `stackSave` seems easy to stub out (do nothing), `stackRestore` I have no idea if it's safe to just ignore.

I think in the current tests both are safe to stub out, as the tests won't end up in the error branch. If we would make a separate test for the jumps we would have to implement them.

internal/emscripten/emscripten.go
```diff
@@ -89,8 +109,17 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
...
+		if !errors.Is(err, ThrowLongjmpError) {
```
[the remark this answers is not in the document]
jerbob92: Yeah it is, by this line:
```
	// At this point we expect the error was from a function defined by HostModuleBuilder that intentionally called panic.
	if runtimeErr, ok := recovered.(error); ok { // e.g. panic(errors.New("whoops"))
		return fmt.Errorf("%w (recovered by wazero)\nwasm stack trace:\n\t%s", runtimeErr, stack)
	}
```

internal/emscripten/emscripten.go
jerbob92: Changed :)

internal/emscripten/emscripten.go
jerbob92: We could check for specific errors, but I wanted the implementation to be the same as the JS one, and they always call this method, so I assume that the C implementation takes care of things like that.

However, I think calling just `exit(1)` would not make you end up in the error branch of `CallWithStack`

ncruces: Are you sure? Doesn't it produce a `sys.ExitError` here:

	if exitErr, ok := recovered.(*sys.ExitError); ok { // Don't wrap an exit error!
		return exitErr
	}

ncruces:
I _think_ a better implementation would be as such:

```go
	// Lookup the table index we will call.
	t := m.Tables[0] // Note: Emscripten doesn't use multiple tables
	f, err := m.Engine.LookupFunction(t, typeID, tableOffset)
	if err != nil {
		panic(err)
	}

	// Call: savedStack = stackSave()
	var savedStack [2]uint64
	err = mod.ExportedFunction("stackSave").CallWithStack(ctx, savedStack[:])
	if err != nil {
		panic(err) // stackSave failed
	}

	err = f.CallWithStack(ctx, stack)
	if err != nil {
		// Module closed: any calls will just fail with the same error
		if _, ok := err.(*sys.ExitError); ok {
			panic(err)
		}

		// Call: stackRestore(savedStack)
		if err := mod.ExportedFunction("stackRestore").CallWithStack(ctx, savedStack[:]); err != nil {
			panic(err) // stackRestore failed
		}
		if !errors.Is(err, ThrowLongjmpError) {
			panic(err) // f failed
		}

		// Call: setThrew(1, 0)
		savedStack[0] = 1
		savedStack[1] = 0
		err = mod.ExportedFunction("setThrew").CallWithStack(ctx, savedStack[:])
		if err != nil {
			panic(err) // setThrew failed
		}
	}
```

This uses `CallWithStack` to avoid a few allocs (to the cost of some obfuscation). We can further keep the `[2]uint64` in `InvokeFunc` to avoid one alloc per call.

More importantly: `stackRestore` and `setThrew` [here](https://github.com/emscripten-core/emscripten/blob/main/src/library_dylink.js) (hopefully that's the correct source) are in the catch block, so their exceptions interrupt flow. We should do the same.

And it's pointless to call `stackRestore` with a `sys.ExitError`, because it'll just do the throw out the `sys.ExitError` again.

ncruces:
My changes, including updating tests so `make test` passes:
https://github.com/jerbob92/wazero/compare/feature/implement-emscripten-jmp...ncruces:wazero:ncruces-emscripten-jmp

internal/emscripten/emscripten.go
jerbob92: Thanks for checking!

jerbob92:
> OK, so, sorry for the noise, but here's another idea: what if we just skip the entire dance if `mod.ExportedFunction("stackSave")` returns `nil`?

Fine with me, but it should never be nil if it comes into the invoke call (except for not real emscripten like the Wazero tests).



> I _think_ a better implementation would be as such:
> 
> ```go
> 	// Lookup the table index we will call.
> 	t := m.Tables[0] // Note: Emscripten doesn't use multiple tables
> 	f, err := m.Engine.LookupFunction(t, typeID, tableOffset)
> 	if err != nil {
> 		panic(err)
> 	}
> 
> 	// Call: savedStack = stackSave()
> 	var savedStack [2]uint64
> 	err = mod.ExportedFunction("stackSave").CallWithStack(ctx, savedStack[:])
> 	if err != nil {
> 		panic(err) // stackSave failed
> 	}
> 
> 	err = f.CallWithStack(ctx, stack)
> 	if err != nil {
> 		// Module closed: any calls will just fail with the same error
> 		if _, ok := err.(*sys.ExitError); ok {
> 			panic(err)
> 		}
> 
> 		// Call: stackRestore(savedStack)
> 		if err := mod.ExportedFunction("stackRestore").CallWithStack(ctx, savedStack[:]); err != nil {
> 			panic(err) // stackRestore failed
> 		}
> 		if !errors.Is(err, ThrowLongjmpError) {
> 			panic(err) // f failed
> 		}
> 
> 		// Call: setThrew(1, 0)
> 		savedStack[0] = 1
> 		savedStack[1] = 0
> 		err = mod.ExportedFunction("setThrew").CallWithStack(ctx, savedStack[:])
> 		if err != nil {
> 			panic(err) // setThrew failed
> 		}
> 	}
> ```
> 
> This uses `CallWithStack` to avoid a few allocs (to the cost of some obfuscation). We can further keep the `[2]uint64` in `InvokeFunc` to avoid one alloc per call.
> 
> More importantly: `stackRestore` and `setThrew` [here](https://github.com/emscripten-core/emscripten/blob/main/src/library_dylink.js) (hopefully that's the correct source) are in the catch block, so their exceptions interrupt flow. We should do the same.
> 
> And it's pointless to call `stackRestore` with a `sys.ExitError`, because it'll just do the throw out the `sys.ExitError` again.

These changes sound good to me!

ncruces:
Want to integrate them in your branch? Not sure I can.

internal/emscripten/emscripten.go
```diff
@@ -89,8 +110,60 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
...
+			panic(err)
+		}
+
+		// If we encounter ThrowLongjmpError, this means that the C code did a
+		// longjmp, which in turn called _emscripten_throw_longjmp and that is
+		// a host function that panics with ThrowLongjmpError. In that case we
+		// ignore the error because we have restored the stack to what it was
+		// before the indirect function call, so the program can continue.
+		// This is the equivalent of the "if (e !== e+0) throw e;" line in the
+		// JS implementation, which checks if the error is Infinity, which is
```
ncruces, line 155: which checks if the error is not a number, which is what the JS implementation throws (Infinity for _emscripten_throw_longjmp, a memory address for C++ exceptions).

ncruces (approved):

internal/emscripten/emscripten.go
jerbob92: Changed :) Also looked into implementing C++ exceptions but that looks a lot more complex.

internal/emscripten/emscripten.go
```diff
@@ -89,8 +110,61 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
...
+	//  } catch(e) {
+	//    stackRestore(sp);
+	//    if (e !== e+0) throw e;
+	//    _setThrew(1, 0);
+	//  }
+	//}
+
+	// This is the equivalent of "var sp = stackSave();".
+	// We reuse savedStack to save allocations.
+	var savedStack [2]uint64
```
mathetake, line 130: maybe add a comment of reference to why `sp = stackSave` becomes two variables? like a source code link in emscripten?

internal/emscripten/emscripten.go
```diff
@@ -89,8 +110,61 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
...
+	err = mod.ExportedFunction("stackSave").CallWithStack(ctx, savedStack[:])
```
mathetake, line 131: is `stackSave` exported since the beginning? If this was introduced recently, I am concerned that this might break the existing codes

mathetake:
it seems the failing tests legit but not directly relevant to this change (I guess it's something to do with unsafe.pointer/uintptr cast thingy)

internal/emscripten/emscripten.go
jerbob92: I have added a comment, it's not 2 variables for `stackSave()`, but it's 2 so that we can use it for `stackSave`, `stackRestore` and `setThrew`, `setThrew` needs 2 because it has 1 input and 1 output variable.

internal/emscripten/emscripten.go
jerbob92: Yes it was, it's not a recent addition to Emscripten, it's just something that we missed during the implementation of our invoke_xxx method since it worked for us in most cases, we just missed the case where a longjmp was used inside the invoke method, which is often used as a throw/catch mechanism in C, all of this is needed to make that work.

mathetake: good to know!

internal/emscripten/emscripten.go
mathetake: aha makes sense

mathetake:
https://github.com/tetratelabs/wazero/pull/1611/commits/dd9833845ecaaf9ee212cfa214ab7f1fd425cd72 yeah this fixed a bug

mathetake:
no luck so far. The only thing I can say is this is not compiler specific (happens to interpreter as well), but I hanve't been able to reproduce

internal/emscripten/emscripten.go
```diff
@@ -82,15 +104,69 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
...
 	// Lookup the table index we will call.
 	t := m.Tables[0] // Note: Emscripten doesn't use multiple tables
 	f, err := m.Engine.LookupFunction(t, typeID, tableOffset)
 	if err != nil {
 		panic(err)
 	}
-
 	err = f.CallWithStack(ctx, stack)
+	runtime.KeepAlive(t)
 	if err != nil {
```
mathetake, line 139: I think this block hasn't been covered in the tests. Do you want to add it?

jerbob92: @mathetake Sure, but I'm not that great in writing wat files, I can convert the sample C into a test and add the Emscripten compiled one to the test folder? Would that suffice?

mathetake: I would prefer Wat, so it would be great if you could do it. If it's too difficult, @evacchi or I would take over so feel free to tell us!

ncruces (approved):

evacchi (approved):

PR #1612, opened 2023-08-04
wasi: add more test cases to poll_oneoff, cleanup impl

evacchi:
follow up to #1606. Cleanup for some logic in `poll_oneoff` and add more test cases+fix nonblocking flags in sockets.

### DIFF
--- a/imports/wasi_snapshot_preview1/poll_test.go
+++ b/imports/wasi_snapshot_preview1/poll_test.go
@@ -2,6 +2,7 @@ package wasi_snapshot_preview1_test
 
 import (
 	"io/fs"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


imports/wasi_snapshot_preview1/poll_test.go
```diff
@@ -192,7 +199,6 @@ func Test_pollOneoff_Stdin(t *testing.T) {
 			mem: concat(
 				clockNsSub(20*1000*1000),
 				fdReadSub,
-				singleton('?'),
```
evacchi, line 195: these really aren't necessary as mem is pre-touched with '?' anyway.

internal/sysfs/sock_windows.go
```diff
@@ -90,6 +90,16 @@ func syscallConnControl(conn syscall.Conn, fn func(fd uintptr) (int, sys.Errno))
 	return
 }
 
+func _pollSock(conn syscall.Conn, flag sys.Pflag, timeoutMillis int32) (bool, sys.Errno) {
```
evacchi, line 93: this utility may be also useful if we expose File.Poll() in sockets too

codefromthecrypt (approved):
good work. thanks!

PR #1614, opened 2023-08-07
Adds support for build on GOOS=js

mathetake:
Fixes:

```
$ GOOS=js GOARCH=wasm go build ./...
# github.com/tetratelabs/wazero/internal/sock
internal/sock/sock_supported.go:8:22: undefined: syscall.SHUT_RD
internal/sock/sock_supported.go:9:22: undefined: syscall.SHUT_RDWR
internal/sock/sock_supported.go:10:22: undefined: syscall.SHUT_WR

```


internal/sock/sock_unsupported.go
```diff
@@ -0,0 +1,10 @@
+//go:build plan9 || js
+
+package sock
+
+// plan9/js doesn't declare these constants
+const (
+	SHUT_RD = iota
+	SHUT_RDWR
```
achille-roussel, line 8: Should we make SHUT_RDWR equal SHUT_RD|SHUT_WR?

Probably better if SHUT_RD isn't zero as well otherwise it's difficult to not shut the read end of the connection.

mathetake: https://github.com/tetratelabs/wazero/pull/1614/commits/0b5b6fed86700db715b29280ed8fb700c01ef732

mathetake: good catch!

achille-roussel (approved):

Issue #1496, opened 2023-05-31
Implement initial optimizing compiler backend


mathetake:
> _the description here will be updated over time_

This issue is the first tracking issue for our frequently requested feature: optimizing the backend. For background, see https://github.com/tetratelabs/wazero/issues/777#issuecomment-1546775472.

The ultimate goal is to develop the optimizing compiler backend which has support for arm64 and amd64 and has an equivalent performance as the state-of-the-art compilers like wasmtime or wasmer. But that work is never-ending considering that every compiler evolves over time. 

Thus, this issue exists for tracking the work around building the foundation for the never-ending iterative development of our new optimizing compiler. 

## Goal (to close this issue): 

1. Develops the compiler backend packages in a way that anyone can simultaneously add optimizations and do an experiment on the backend -- this requires thorough code documentation on the rationale of any non-trivial logic. Cite the research papers etc as much as possible in the code.
2. No third-party dependency - we previously thought about re-using Wasm-compiled Cranelift and embedding it in wazero to emit machine codes, but that would be more-- therefore everything will be written entirely from scratch in Go.
3. Passes all the Wasm V1 specification test on arm64 -- neither amd64 nor SIMD is a goal here.

## Non-Goal

- amd64
- V2 specification tests on arm64

## Tasks

TBD 

## Note

- I am not planning to reuse any portion of the current single-pass compiler code, so for those easger to contribute, please fight your urge to do so until my first PoC is out (I am expecting it up within July).
- Due to the limitation posed by go runtime, the resulting backend may not be the fastest for some cases, but in anyway no one should expect it to win all bench cases. Even after this finishes, some can find a losing case for sure.

mathetake:
So we call this new backend `wazevo` for convenience like a prefix to all related PRs etc because "new optimizing compiler" is verbose. But it's just temporarily used and we can change it later whenever we want and necessary.

mathetake:
FYI anyone following this thread: the first branch is here https://github.com/tetratelabs/wazero/compare/wazevoinit?expand=1 

The current status is that I'm working on the frontend (Wasm->ssa IR), especially getting around to implementing the control flow transitions.

mathetake:
so now that it's passing the entire v1 spectests, the things I need to work on but left are ensuring the stability around register allocations, and I would like to run a bit of fuzzing. Meanwhile, @evacchi you can start working on adding the v2-related instruction supports (non-trapping etc) 👍

mathetake:
[...]

PR #1615, opened 2023-08-08
wazevo: initial impl of the new optimizing backend

mathetake:
This commit introduces the initial implementation of our new optimizing
backend called "wazevo" (#1496) which I have worked on in the last 2 months.

The new backend is written completely from scratch without reusing any line of the 
existing compiler, assembler etc. The notable difference is that we employ the traditional
way of compilation pipeline, e.g. SSA-level IR, optimization passes, instruction selections,
register allocations, etc as opposed to the current single pass compiler. As a result, 
for example, we can now use the register-based calling conventions etc which will definitely
contribute to the perf improvement. As a sneak preview of the perf improvement, 
the following is the bench result of `recursive_fibonacci(30)`.

```
Benchmark_wazevo
Benchmark_wazevo/old
Benchmark_wazevo/old-10         	     106	  10891053 ns/op
Benchmark_wazevo/wazevo
Benchmark_wazevo/wazevo-10      	     286	   4175793 ns/op
```

As expected, this is far from completion, and instead, the purpose of this commit
is to set up the foundation and code base that multiple people can work 
on collaboratively. For now, it only has the AArch64 backend, and the plan is to make 
it pass all spec tests before we start introducing x64 backend. That way we could 
have a better abstraction layer over ISA-specific code. In any case, 
this leaves tens of TODOs intentionally, and we will iterate on this 
new codebase over the next few months in the subsequent PRs.

### Pipeline

Basically, we only have three components:

1. Frontend Compiler is in charge of lowering WebAssembly-level functions into SSA-level functions.
2. SSA Builder provides the interface between frontend and backend and is decoupled from Wasm-level concepts. It provides the necessary logic and functions to the Frontend Compiler to lower Wasm-level functions to neutral SSA-IR functions. After the construction of IR-functions, it also can be used to access the constructions functions by Backend.
3. Backend Compiler lowers the SSA-IR level functions to the ISA-specific instruction sequences. This takes a SSA Builder as its input. This does some optimization during the course of lowering as well as the register allocations.

After going through all three components for all functions in a module, we also resolve the "relocation" of functions. In other words, we resolve local function calls as relative address "direct branches', rather than indirect jumps. 

The following diagram illustrates the overview of the pipeline:

```mermaid
graph TD;
    SSA_Builder[[SSA Builder]]
    Frontend_Compiler(Frontend Compiler)
    Backend_compiler(Backend Compiler)
    SSA_Function(SSA Function)

    foo.wasm-- for each function -->Frontend_Compiler;
    Frontend_Compiler-->SSA_Function;
    SSA_Function-->Backend_compiler;
    Backend_compiler-- after all functions --> Relocations
    SSA_Builder---Frontend_Compiler;
    SSA_Builder---SSA_Function;
    SSA_Builder---Backend_compiler;
    ssa_passes[["CFG Analysis, 
SSA-level optimization passes,
Block Ordering"]]
    ssa_passes--->SSA_Function
    backend_passes[["Instruction Selection
Register Allocation
Branch Resolution,
Binary Encoding"]]
    backend_passes--->Backend_compiler

```

####

### Frame Pointer usage

As a simple example of compilation result, the Wasm function which swaps two params:

```wat
(func (param i32 i32) (result i32 i32)
    local.get 1
    local.get 0
)
```

is compiled as

```asm
str x30, [sp, #-0x10]!
mov x0, x3
mov x1, x2
ldr x30, [sp], #0x10
ret
```

where you can see we only use `sp`, but not `fp` which is kinda weird to other compilers (not that x0 and x1 params are implementation specific and not relevant here. The actual Wasm param starts at x2/v2). Basically, FP is only 
necessary when we want to integrate native debugger and profiles, which is in our case not relevant. Moreover, the typical way of saving/restoring frame pointer (at prologue/epilogue) cannot be applied here because our stack is Go-allocated byte slice therefore it "moves". As a consequence, saving the absolute address of FP is absolutely dangerous therefore we do not store it at all. Speaking of SP, the caller/callee always knows how much space needed and hence we only just need to add/sub relative amount. (Of course that can be applied to FP but it's costly and not necessary at all!).


codefromthecrypt:
well I have to say, thank god it isn't 24k lines 😎

internal/engine/wazevo/engine.go
```diff
@@ -0,0 +1,206 @@
...
+
+		// Run SSA-level optimization passes.
+		ssaBuilder.RunPasses()
+
+		// Finalize the layout of SSA blocks which might use the optimization results.
+		ssaBuilder.LayoutBlocks()
+
+		// Now our ssaBuilder contains the necessary information to further lower them to
+		// machine code.
+		body, rels, goPreambleSize, err := be.Compile()
```
mathetake, line 108: this is the outermost of the pipeline

internal/engine/wazevo/engine.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	cm.executable = executable
+
+	for i, b := range bodies {
+		offset := cm.functionsOffsets[i]
+		copy(executable[offset:], b)
+	}
+
+	// Resolve relocations for local function calls.
+	machine.ResolveRelocations(e.refToBinaryOffset, executable, e.rels)
```
mathetake, line 144: after compiling all local functions, we can resolve the relocations (relative offers between the call instructions and the target functions).

internal/engine/wazevo/frontend/lower.go
```diff
@@ -0,0 +1,842 @@
...
+			fmt.Println("--------- Translated " + wasm.InstructionName(op) + " --------")
+			fmt.Println("Stack: " + c.loweringState.String())
+			fmt.Println(c.formatBuilder())
+			fmt.Println("--------------------------")
+		}
+		c.loweringState.pc++
+	}
+}
+
+func (c *Compiler) lowerOpcode(op wasm.Opcode) {
```
mathetake, line 147: this is where lower Wasm instructions into SSA instructions

internal/engine/wazevo/ssa/pass.go
```diff
@@ -0,0 +1,266 @@
+package ssa
+
+// RunPasses implements Builder.RunPasses.
+//
+// The order here matters; some pass depends on the previous ones.
+//
+// Note that passes suffixed with "Opt" are the optimization passes, meaning that they edit the instructions and blocks
+// while the other passes are not, like passEstimateBranchProbabilities does not edit them, but only calculates the additional information.
+func (b *builder) RunPasses() {
```
mathetake, line 9: this is where IR-level optimization passes run

internal/engine/wazevo/backend/machine.go
```diff
@@ -0,0 +1,98 @@
+package backend
+
+import (
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/backend/regalloc"
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/ssa"
+)
+
+type (
+	// Machine is a backend for a specific ISA machine.
+	Machine interface {
```
mathetake, line 10: This is the interface implemented per ISA (of course currently only arm64 tho)

internal/engine/wazevo/backend/regalloc/api.go
mathetake, line 1: Interfaces in this file are implemented per ISA so that register allocation can run without ISA specifics

internal/engine/wazevo/backend/isa/arm64/machine.go
```diff
@@ -0,0 +1,461 @@
...
+
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/backend"
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/backend/regalloc"
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/ssa"
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/wazevoapi"
+)
+
+type (
+	// machine implements backend.Machine.
+	machine struct {
```
mathetake, line 16: arm64 implementation of Machine interface

internal/engine/wazevo/backend/isa/arm64/machine.go
```diff
@@ -0,0 +1,461 @@
...
+		//          |   ............  |      |
+		//          |   spill slot 2  |      |
+		//          |   spill slot 1  |   <<-+
+		//          |   clobbered N   |
+		//          |   ...........   |
+		//          |   clobbered 1   |
+		//          |   clobbered 0   |
+		//   SP---> +-----------------+
+		//             (low address)
+		//
```
mathetake, line 74: diagram to explain the stack layout

internal/engine/wazevo/backend/isa/arm64/abi_go_entry.go
mathetake, line 1: this file implements the "trampoline" prologue executed right after being jumped from Go.

mathetake (commented):
notes

mathetake:
now that 1.4.0 cut, I will land this!

PR #1620, opened 2023-08-09
Use atomic.Uint64.

ncruces:
Now that we floor on Go 1.19, take advantage of atomic types:
https://tip.golang.org/doc/go1.19#atomic_types

Context #1299, #1321.

### DIFF
--- a/experimental/wazerotest/wazerotest.go
+++ b/experimental/wazerotest/wazerotest.go
@@ -24,7 +24,6 @@ const (
 // WebAssembly module.
 type Module struct {
 	internalapi.WazeroOnlyType
-	exitStatus uint64
 
 	// The module name that will be returned by calling the Name method.
 	ModuleName string
@@ -39,6 +38,8 @@ type Module struct {
 	// The program memory. If non-nil, the memory is automatically exported as
 	// "memory".
 	ExportMemory *Memory
+
+	exitStatus atomic.Uint64
 
 	once                        sync.Once
 	exportedFunctions           map[string]api.Function
@@ -111,7 +112,7 @@ func (m *Module) Close(ctx context.Context) error {
 
 // CloseWithExitCode implements the same method as documented on api.Closer.
 func (m *Module) CloseWithExitCode(ctx context.Context, exitCode uint32) error {
-	atomic.CompareAndSwapUint64(&m.exitStatus, 0, exitStatusMarker|uint64(exitCode))
+	m.exitStatus.CompareAndSwap(0, exitStatusMarker|uint64(exitCode))
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


experimental/wazerotest/wazerotest.go
```diff
@@ -40,6 +39,8 @@ type Module struct {
 	// "memory".
 	ExportMemory *Memory
 
+	exitStatus atomic.Uint64
```
evacchi, line 42: this may be a silly question, but why are we moving this field?

internal/engine/compiler/engine.go
```diff
@@ -966,13 +966,13 @@ func newEngine(enabledFeatures api.CoreFeatures, fileCache filecache.Cache) *eng
 //
 // By declaring these values as `var`, slices created via `make([]..., var)`
 // will never be allocated on stack [1]. This means accessing these slices via
-// raw pointers is safe: As of version 1.18, Go's garbage collector never relocates
+// raw pointers is safe: As of version 1.21, Go's garbage collector never relocates
```
evacchi, line 969: unless I am misunderstanding, this may stay at 1.18, because the fact we are updating does not change that Go's GC has that behavior since 1.18

internal/wasm/store.go
```diff
@@ -116,6 +106,14 @@ type (
 		//	  security implications.
 		Sys *internalsys.Context
 
+		// Closed is used both to guard moduleEngine.CloseWithExitCode and to store the exit code.
+		//
+		// The update value is closedType + exitCode << 32. This ensures an exit code of zero isn't mistaken for never closed.
+		//
+		// Note: Exclusively reading and updating this with atomics guarantees cross-goroutine observations.
+		// See /RATIONALE.md
+		Closed atomic.Uint64
```
evacchi, line 115: same question about reordering. But maybe the note about keeping it at the top was only because it was a plain `uint64` to begin with?

evacchi (commented):
basically just nits

internal/engine/compiler/engine.go
ncruces: No, not necessary to change, but I think it adds information: it's 1.21 and Go is still a non moving GC.

internal/wasm/store.go
ncruces: Yes. Heap pointers at the start of an allocation (start of heap allocated struct) are the only way to ensure 64-bit alignment before Go 1.19.

ncruces:
BTW the struct where the `uint64` **wasn't** at the top of struct was a latent bug that this fixes.

experimental/wazerotest/wazerotest.go
evacchi: no that's ok, thanks for clarifying

internal/engine/compiler/engine.go
evacchi: I think I misread "as of" as "since" :)

evacchi (approved):
lgtm

internal/engine/compiler/engine.go
achille-roussel: I can picture us on the Go 1.42 release "the year is 2044, Go still doesn't have a moving GC"

PR #1629, opened 2023-08-14
Fix broken link in example README

philippgille:
Hi :wave: , Just noticed a broken link in the example README


philippgille:
Sorry for the forced push, a sign-off was missing

PR #1636, opened 2023-08-18
Lifts LookupFunction into ModuleInstance from ModuleEngine

### LINKED ISSUE #1626: A host function that calls another host function loses access to the memory
**Describe the bug**
So I'm not entirely sure this is a bug report or a feature request. I came across this issue while implementing C++ exceptions support for Emscripten in Wazero. Apparently it's possible that Emscripten uses the generated invoke_xxx methods (which is a host function) to call another host function, without a "proxy" through the guest. I then found out that it's also possible to reproduce it without using C++ exceptions. 

This issue causes the second host function not to have access to the memory, because the second host function is being called by invoke_xxx, which is in the env module, and receives the env `api.Module`, while the invoke_xxx is also in the env module, it does receive the main application memory, and this is also mentioned in the comments of `GoModuleFunction`:

```
// This function can be non-deterministic or cause side effects. It also
// has special properties not defined in the WebAssembly Core specification.
// Notably, this uses the caller's memory (via Module.Memory). See
// https://www.w3.org/TR/wasm-core-1/#host-functions%E2%91%A0
```

While it does make sense to give this the caller's memory (if we didn't, the host functions would be way less usable), but in my case, this means that the second host function call does not have access to the memory, just because it was called from another host method and not from the guest, while it would have access to it if it would have been called by a "proxy" function in the guest. 

I'm not sure what the best solution to this would be. Perhaps it would be possible to have an option to override the moduleInstance or memoryInstance in `m.Engine.LookupFunction` or `f.CallWithStack` so that we can use it in `InvokeFunc.Call()` to maintain access to the memory when doing host to host calls.

Given the sample code in the next block, the following two things will happen:
 - invoke_i (host)
 - you_could_say_im_a_host_function_proxy (guest)
 - you_could_say_im_a_host_function (host)
 - m.Memory().ReadByte(200)

Works :tada: 

 - invoke_i (host)
 - you_could_say_im_a_host_function (host)
 - m.Memory().ReadByte(200)

Breaks :bomb: (`failed: runtime error: invalid memory address or nil pointer dereference`)

**To Reproduce**

<details>

<summary>C code compiled with Emscripten</summary>

```c
#include <stdio.h>
#include <stdlib.h>
#include <setjmp.h>

extern int you_could_say_im_a_host_function();

void jmpfunction(jmp_buf env_buf) {
   longjmp(env_buf, 42);
}

int you_could_say_im_a_host_function_proxy() {
  return you_could_say_im_a_host_function();
}

int main () {
   int val;
   jmp_buf env_buffer;

   val = setjmp( env_buffer );

   if( val != 0 ) {
      printf("Returned from a longjmp() with value = %d\n", val);
      you_could_say_im_a_host_function_proxy();
      you_could_say_im_a_host_function();
      exit(0);
   }

   printf("Jump function call\n");
   jmpfunction(env_buffer);

   return(0);
}

```
</details>

`emcc -sERROR_ON_UNDEFINED_SYMBOLS=0 -sSUPPORT_LONGJMP=emscripten -g invoke_host.c -o invoke_host.wasm`

<details>

<summary>Go code with the host function implemented</summary>

```go
package main

import (
	"context"
	_ "embed"
	"github.com/tetratelabs/wazero/experimental"
	"github.com/tetratelabs/wazero/experimental/logging"
	"github.com/tetratelabs/wazero/imports/emscripten"
	"log"
	"os"
)

//go:embed invoke_host.wasm
var wasm []byte

func main() {
	ctx := context.WithValue(context.Background(), experimental.FunctionListenerFactoryKey{}, logging.NewLoggingListenerFactory(os.Stdout))
	r := wazero.NewRuntime(ctx)

	defer r.Close(ctx)

	if _, err := wasi_snapshot_preview1.Instantiate(ctx, r); err != nil {
		log.Fatal(err)
	}

	compiledModule, err := r.CompileModule(ctx, wasm)
	if err != nil {
		log.Fatal(err)
	}

	functionExporter, err := emscripten.NewFunctionExporterForModule(compiledModule)
	if err != nil {
		log.Fatal(err)
	}

	envModule := r.NewHostModuleBuilder("env")
	functionExporter.ExportFunctions(envModule)
	envModule.NewFunctionBuilder().WithGoModuleFunction(api.GoModuleFunc(func(ctx context.Context, m api.Module, stack []uint64) {
		m.Memory().ReadByte(200)
		stack[0] = 42
	}), []api.ValueType{}, []api.ValueType{api.ValueTypeI32}).Export("you_could_say_im_a_host_function")
	_, err = envModule.Instantiate(ctx)
	if err != nil {
		log.Fatal(err)
	}

	moduleConfig := wazero.NewModuleConfig().
		WithStartFunctions("_start").
		WithStdout(os.Stdout).
		WithStderr(os.Stderr)

	_, err = r.InstantiateModule(ctx, compiledModule, moduleConfig)
	if err != nil {
		log.Fatal(err)
	}
}
```

</details>

Full sample zip: 
[invoke_host.zip](https://github.com/tetratelabs/wazero/files/12327066/invoke_host.zip)

- Go version: 1.20.2
- wazero Version: v1.4.0
- Host architecture: amd64
- Runtime mode: both

**Additional context**

This was also discussed a bit on Slack: https://gophers.slack.com/archives/C040AKTNTE0/p1691699092428169

mathetake:
Fixes #1626

### DIFF
--- a/internal/emscripten/emscripten.go
+++ b/internal/emscripten/emscripten.go
@@ -106,10 +106,7 @@ func (v *InvokeFunc) Call(ctx context.Context, mod api.Module, stack []uint64) {
 
 	// Lookup the table index we will call.
 	t := m.Tables[0] // Note: Emscripten doesn't use multiple tables
-	f, err := m.Engine.LookupFunction(t, typeID, tableOffset)
-	if err != nil {
-		panic(err)
-	}
+	f := m.LookupFunction(t, typeID, tableOffset)
 
 	// The Go implementation below mimics the Emscripten JS behaviour to support
 	// longjmps from indirect function calls. The implementation of these
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -681,24 +681,20 @@ func (e *moduleEngine) newFunction(f *function) api.Function {
 }
 
 // LookupFunction implements the same method as documented on wasm.ModuleEngine.
-func (e *moduleEngine) LookupFunction(t *wasm.TableInstance, typeId wasm.FunctionTypeID, tableOffset wasm.Index) (f api.Function, err error) {
+func (e *moduleEngine) LookupFunction(t *wasm.TableInstance, typeId wasm.FunctionTypeID, tableOffset wasm.Index) (*wasm.ModuleInstance, wasm.Index) {
 	if tableOffset >= uint32(len(t.References)) || t.Type != wasm.RefTypeFuncref {
-		err = wasmruntime.ErrRuntimeInvalidTableAccess
-		return
+		panic(wasmruntime.ErrRuntimeInvalidTableAccess)
 	}
 	rawPtr := t.References[tableOffset]
 	if rawPtr == 0 {
-		err = wasmruntime.ErrRuntimeInvalidTableAccess
-		return
+		panic(wasmruntime.ErrRuntimeInvalidTableAccess)
 	}
 
 	tf := functionFromUintptr(rawPtr)
 	if tf.typeID != typeId {
-		err = wasmruntime.ErrRuntimeIndirectCallTypeMismatch
-		return
+		panic(wasmruntime.ErrRuntimeIndirectCallTypeMismatch)
 	}
-	f = e.newFunction(tf)
-	return
+	return tf.moduleInstance, tf.parent.index
 }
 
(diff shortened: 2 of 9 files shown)


jerbob92:
Tested with my C++ exceptions branch and works indeed, thanks so much :partying_face:

PR #1637, opened 2023-08-18
Adds LookupFunction in experimental/table package

mathetake:


mathetake:
PTAL @jerbob92

jerbob92:
@mathetake Looks good to me! Shall I use this branch to change my implementation to so that we can validate it? I'm also not completely sure I always know the params/return type when I'm looking up the function, so have to try that out.

mathetake:
Cool that sounds great! Please let me know if you have any issue

jerbob92:
@mathetake looks like I can't reliably know the types in all cases, it is possible to move the typeID validation somewhere else or make it optional?

mathetake:
genuine question then: how can you know the `stack` param to be passed to Call API then?

jerbob92:
> genuine question then: how can you know the `stack` param to be passed to Call API then?

Yeah you're right, but it's pretty decoupled in embind since the setup of the invokers is in a completely different place from calling it, and it doesn't always know the types when it's setting up the invokers.

But I think I was able to work around it :)

mathetake:
cool, keep me posted !

jerbob92:
@mathetake Got it fully working outside the Wazero repo: https://github.com/jerbob92/wazero-emscripten-embind

So, LGTM! :+1:

mathetake:
Awesome!

PR #1635, opened 2023-08-17
wazevo: adds bit count instructions clz, ctz

evacchi:
Continues #1496.

This introduces the front-end and back-end for clz, ctz (in terms of rbit+clz).

It originally also introduced popcnt for the front-end, but this requires 
VCNT in the backend, which is probably worth its own separate PR.

Contextually, this also introduces a partial implementation for bitRR.

First PR to wazevo, so I am not entirely sure that the generated
code for the SSA and the backend is correct, let me know if there are errors,
I am really figuring out a lot of stuff, including the framework :)

Signed-off-by: Edoardo Vacchi <evacchi@users.noreply.github.com>

PR #1638, opened 2023-08-18
wazevo: add bit count instruction Popcnt

evacchi:
Moves further #1496 and follows-up to #1635.

We introduce support to `popcnt` with the required dependencies for its implementation: we use vector instructions CNT and UADDLV as we do in the current compiler. This PR also implements `movToVec`, `movFromVec` and it lays down the first steps to extend further with the missing related operations in `vecMisc`, `vecLanes` (originally unimplemented).

### DIFF
--- a/internal/engine/wazevo/backend/backend_test.go
+++ b/internal/engine/wazevo/backend/backend_test.go
@@ -1426,9 +1426,19 @@ L1 (SSA Block: blk0):
 	clz w4?, w2?
 	rbit w27, w2?
 	clz w5?, w27
-	clz x6?, x3?
+	mov d11?.8b, x2?.8b
+	cnt d11?.16B, d11?.16B
+	uaddlv d11?.8B, d11?.8B
+	mov x6?.8b, d11?.8b
+	clz x7?, x3?
 	rbit x27, x3?
-	clz x7?, x27
+	clz x8?, x27
+	mov d10?.8b, x3?.8b
+	cnt d10?.16B, d10?.16B
+	uaddlv d10?.8B, d10?.8B
+	mov x9?.8b, d10?.8b
+	mov x5, x9?
+	mov x4, x8?
 	mov x3, x7?
 	mov x2, x6?
 	mov x1, x5?
@@ -1438,12 +1448,18 @@ L1 (SSA Block: blk0):
 			afterFinalizeARM64: `
 L1 (SSA Block: blk0):
 	str x30, [sp, #-0x10]!
+	mov x8, x3
 	clz w0, w2
 	rbit w27, w2
 	clz w1, w27
+	cnt x2.16B, x2.16B
+	uaddlv x2.8B, x2.8B
 	clz x2, x3
 	rbit x27, x3
 	clz x3, x27
+	cnt x8.16B, x8.16B
+	uaddlv x8.8B, x8.8B
+	mov x5.8b, x8.8b
 	ldr x30, [sp], #0x10
 	ret
 `,
--- a/internal/engine/wazevo/backend/backend_test.go
+++ b/internal/engine/wazevo/backend/backend_test.go
@@ -1426,17 +1426,17 @@ L1 (SSA Block: blk0):
 	clz w4?, w2?
 	rbit w27, w2?
 	clz w5?, w27
-	mov d11?.8b, x2?.8b
-	cnt d11?.16B, d11?.16B
-	uaddlv d11?.8B, d11?.8B
-	mov x6?.8b, d11?.8b
+	ins v13?.D[0], x2?
+	cnt v14?.16B, v13?.16B
+	uaddlv v15?.8B, v14?.8B
+	mov v6?.D[0], x6?
 	clz x7?, x3?
 	rbit x27, x3?
 	clz x8?, x27
-	mov d10?.8b, x3?.8b
-	cnt d10?.16B, d10?.16B
-	uaddlv d10?.8B, d10?.8B
-	mov x9?.8b, d10?.8b
+	ins v10?.D[0], x3?
+	cnt v11?.16B, v10?.16B
+	uaddlv v12?.8B, v11?.8B
+	mov v9?.D[0], x9?
 	mov x5, x9?
 	mov x4, x8?
 	mov x3, x7?
@@ -1452,14 +1452,17 @@ L1 (SSA Block: blk0):
 	clz w0, w2
 	rbit w27, w2
 	clz w1, w27
-	cnt x2.16B, x2.16B
-	uaddlv x2.8B, x2.8B
-	clz x2, x3
-	rbit x27, x3
-	clz x3, x27
-	cnt x8.16B, x8.16B
-	uaddlv x8.8B, x8.8B
-	mov x5.8b, x8.8b
+	ins v13.D[0], x2
+	cnt v14.16B, v13.16B
+	uaddlv v15.8B, v14.8B
+	mov v6.D[0], x2
+	clz x3, x8
+	rbit x27, x8
+	clz x4, x27
+	ins v10.D[0], x8
+	cnt v11.16B, v10.16B
+	uaddlv v12.8B, v11.8B
+	mov v9.D[0], x5
 	ldr x30, [sp], #0x10
 	ret
 `,
--- a/internal/engine/wazevo/backend/backend_test.go
+++ b/internal/engine/wazevo/backend/backend_test.go
@@ -1429,14 +1429,14 @@ L1 (SSA Block: blk0):
 	ins v13?.D[0], x2?
 	cnt v14?.16B, v13?.16B
 	uaddlv v15?.8B, v14?.8B
-	mov v6?.D[0], x6?
+	mov x6?, v15?.D[0]
 	clz x7?, x3?
 	rbit x27, x3?
 	clz x8?, x27
 	ins v10?.D[0], x3?
 	cnt v11?.16B, v10?.16B
 	uaddlv v12?.8B, v11?.8B
-	mov v9?.D[0], x9?
+	mov x9?, v12?.D[0]
 	mov x5, x9?
 	mov x4, x8?
 	mov x3, x7?
@@ -1455,14 +1455,14 @@ L1 (SSA Block: blk0):
 	ins v13.D[0], x2
 	cnt v14.16B, v13.16B
 	uaddlv v15.8B, v14.8B
-	mov v6.D[0], x2
+	mov x2, v15.D[0]
 	clz x3, x8
 	rbit x27, x8
 	clz x4, x27
 	ins v10.D[0], x8
 	cnt v11.16B, v10.16B
 	uaddlv v12.8B, v11.8B
-	mov v9.D[0], x5
+	mov x5, v12.D[0]
 	ldr x30, [sp], #0x10
 	ret
 `,
--- a/internal/engine/wazevo/backend/backend_test.go
+++ b/internal/engine/wazevo/backend/backend_test.go
@@ -1426,17 +1426,17 @@ L1 (SSA Block: blk0):
 	clz w4?, w2?
 	rbit w27, w2?
 	clz w5?, w27
-	ins v13?.D[0], x2?
-	cnt v14?.16B, v13?.16B
-	uaddlv v15?.8B, v14?.8B
-	mov x6?, v15?.D[0]
+	ins v13?.d[0], x2?
+	cnt v14?.16b, v13?.16b
+	uaddlv h15?, v14?.8b
+	mov x6?, v15?.d[0]
 	clz x7?, x3?
 	rbit x27, x3?
 	clz x8?, x27
-	ins v10?.D[0], x3?
-	cnt v11?.16B, v10?.16B
-	uaddlv v12?.8B, v11?.8B
-	mov x9?, v12?.D[0]
+	ins v10?.d[0], x3?
+	cnt v11?.16b, v10?.16b
+	uaddlv h12?, v11?.8b
+	mov x9?, v12?.d[0]
 	mov x5, x9?
 	mov x4, x8?
 	mov x3, x7?
@@ -1452,17 +1452,17 @@ L1 (SSA Block: blk0):
 	clz w0, w2
 	rbit w27, w2
 	clz w1, w27
-	ins v13.D[0], x2
-	cnt v14.16B, v13.16B
-	uaddlv v15.8B, v14.8B
-	mov x2, v15.D[0]
+	ins v8.d[0], x2
+	cnt v8.16b, v8.16b
+	uaddlv h8, v8.8b
+	mov x2, v8.d[0]
 	clz x3, x8
 	rbit x27, x8
 	clz x4, x27
-	ins v10.D[0], x8
-	cnt v11.16B, v10.16B
-	uaddlv v12.8B, v11.8B
-	mov x5, v12.D[0]
+	ins v8.d[0], x8
+	cnt v8.16b, v8.16b
+	uaddlv h8, v8.8b
+	mov x5, v8.d[0]
 	ldr x30, [sp], #0x10
 	ret
 `,
--- a/internal/engine/wazevo/backend/isa/arm64/cond.go
+++ b/internal/engine/wazevo/backend/isa/arm64/cond.go
@@ -224,6 +224,34 @@ const (
 	vecArrangement8B
 	// vecArrangement16B is an arrangement of 16 bytes (128-bit vector)
 	vecArrangement16B
+	// vecArrangement4H is an arrangement of 4 half precisions (64-bit vector)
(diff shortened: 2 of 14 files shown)


internal/engine/wazevo/backend/isa/arm64/instr.go
```diff
@@ -814,7 +832,7 @@ func (i *instruction) String() (str string) {
 	case cCmpImm:
 		panic("TODO")
 	case fpuMov64:
-		str = fmt.Sprintf("mov %s.8b, %s.8b", formatVRegSized(i.rd.nr(), 128), formatVRegSized(i.rn.nr(), 128))
+		str = fmt.Sprintf("mov %s.8b, %s.8b", formatVRegSized(i.rd.nr(), 64), formatVRegSized(i.rn.nr(), 64))
```
evacchi, line 835: this changes fpuMov64 to 128bit vregs for printing. 

the issue is that popcnt MOVs a 64-bit GPR to a vector register.

Was assuming 128 bits here incorrect in the first place or do we need to add stronger logic to autodetect? 

- if it was incorrect, then I will fix the test cases that I broke; 
- otherwise, how to do this without breaking the String() routine? Can we detect the size of the register "automatically" instead of asserting it here?

The offending line is the panic in `formatVRegSized()`:

```
		switch r.RegType() {
		case regalloc.RegTypeInt:
			switch size {
			case 32:
				ret = fmt.Sprintf("w%d?", r.ID())
			case 64:
				ret = fmt.Sprintf("x%d?", r.ID())
			default:
				panic("BUG: invalid register size: " + strconv.Itoa(int(size))) /// <--- fault occurs here
			}

```

internal/engine/wazevo/backend/isa/arm64/instr.go
```diff
@@ -890,9 +908,21 @@ func (i *instruction) String() (str string) {
...
+		)
 	case vecMisc:
-		panic("TODO")
+		str = fmt.Sprintf("%s %s.%s, %s.%s",
+			vecOp(i.u1),
+			formatVRegSized(i.rd.nr(), 64),
+			vecArrangement(i.u2),
+			formatVRegSized(i.rn.nr(), 64),
+			vecArrangement(i.u2),
+		)
```
evacchi, line 925: the 64-bit size has the same potential flaw as the change above

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -855,6 +869,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
...
+	switch op {
+	case vecOpUaddlv:
+		q, size := vecArrangementParams(arr)
+		return q<<29 | 0b101110<<24 | size<<22 | 0b110000001110<<10 | rn<<5 | rd
+	default:
+		panic("unsupported or illegal vecOp: " + op.String())
+	}
+}
+
+func encodeVecMisc(op vecOp, rd, rn uint32, arr vecArrangement) uint32 {
```
evacchi, line 882: I can't believe it but I think I got both the encodings right on my first shot 🎉 
need to add unit tests though

internal/engine/wazevo/backend/isa/arm64/instr.go
mathetake: yeah you are right, this should be `vX.b8` instead of `qX.b8`. I will fix it in a separate PR shortly so that we can keep this small

internal/engine/wazevo/backend/isa/arm64/instr.go
mathetake: we can fix it later, once we are at the phase when implementing SIMD spec. for now let's just not get concerned that much

mathetake: vector operations need arrangement to correctly print. Could you add `formatVRegSizedVec` or something like that which accepts `vecArrangement` as well? While there, could you add all the arrangement needed in the future? you can check the existing compiler's code which arrangement exists

internal/engine/wazevo/backend/isa/arm64/instr.go
mathetake: 128 bit is correct.  See how `fpuMov64` is encoded: 


- https://shell-storm.org/online/Online-Assembler-and-Disassembler/?opcodes=f21fbf0e&arch=arm64&endianness=little&baddr=0x00000000&dis_with_addr=True&dis_with_raw=True&dis_with_ins=True#disassembly
- https://github.com/tetratelabs/wazero/blob/main/internal/engine/wazevo/backend/isa/arm64/instr_encoding_test.go#L219

If you need to move GP to Vector, then use unimplemented `movToVec`. `fpuMov64` is not for them.

internal/engine/wazevo/backend/isa/arm64/cond.go
```diff
@@ -213,3 +213,29 @@ func condFlagFromSSAFloatCmpCond(c ssa.FloatCmpCond) condFlag {
...
+type vecArrangement byte
+
+const (
+	// vecArrangementNone is an arrangement indicating no data is stored.
+	vecArrangementNone vecArrangement = iota
+	// vecArrangement8B is an arrangement of 8 bytes (64-bit vector)
+	vecArrangement8B
+	// vecArrangement16B is an arrangement of 16 bytes (128-bit vector)
+	vecArrangement16B
+)
```
mathetake, line 227: let's add all of the arrangement now: https://github.com/tetratelabs/wazero/blob/eb11a2caf83e196d318999bee28040637f1be808/internal/asm/arm64/consts.go#L941-L979


and implement formatVRegSizedVec as I suggested ^^

internal/engine/wazevo/backend/isa/arm64/instr.go
mathetake: so in other words, you do not need to modify `fpuMov64`, and implement `movToVec` or `movFromVec`

mathetake:
https://github.com/tetratelabs/wazero/commit/9e944dcc83d9c8b348040139032ee809549f7b1c is landed, so could you rebase onto main before making this ready btw

evacchi:
moved forward, a bit of cleanup, still needs tests and docs :) pushing for feedback 🙏

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -251,6 +279,57 @@ func encodeFpuCSel(rd, rn, rm uint32, c condFlag, _64bit bool) uint32 {
...
+		panic("Unsupported arrangement " + arr.String())
+	}
+
+	return 0b01001110000<<21 | imm5<<16 | 0b000111<<10 | rn<<5 | rd
+}
+
+// encodeMoveFromVec encodes as "Move vector element to a general-purpose register." in
+// https://developer.arm.com/documentation/ddi0596/2020-12/SIMD-FP-Instructions/UMOV--Unsigned-Move-vector-element-to-general-purpose-register-?lang=en
+// https://developer.arm.com/documentation/ddi0596/2020-12/SIMD-FP-Instructions/MOV--to-general---Move-vector-element-to-general-purpose-register--an-alias-of-UMOV-?lang=en
+func encodeMoveFromVec(rd, rn uint32, arr vecArrangement, index vecIndex) uint32 {
```
mathetake, line 310: maybe let's add the bounds check of the index

mathetake (commented):
yeah this is right on the track!  💯

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -849,6 +928,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
...
+func encodeVecMisc(op vecOp, rd, rn uint32, arr vecArrangement) uint32 {
+	switch op {
+	case vecOpCnt:
+		var q, size uint32
+		if arr == vecArrangement8B {
+			q = uint32(0b0)
+			size = uint32(0b00)
+		} else {
+			q = uint32(0b1)
+			size = uint32(0b00)
```
mathetake, line 950: unnecessary `uint32()`

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -849,6 +928,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
...
+	}
+}
+
+func encodeVecMisc(op vecOp, rd, rn uint32, arr vecArrangement) uint32 {
+	switch op {
+	case vecOpCnt:
+		var q, size uint32
+		if arr == vecArrangement8B {
+			q = uint32(0b0)
+			size = uint32(0b00)
```
mathetake, line 947: same

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -849,6 +928,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
 	return _64bit<<31 | opc<<29 | 0b100110<<23 | _64bit<<22 | immr<<16 | imms<<10 | rn<<5 | rd
 }
 
+func encodeVecRRR(op vecOp, rd, rn uint32, arr vecArrangement) uint32 {
+	switch op {
+	case vecOpUaddlv:
+		q, size := vecArrangementParams(arr)
+		return q<<29 | 0b101110<<24 | size<<22 | 0b110000001110<<10 | rn<<5 | rd
```
mathetake, line 935: this should be the outer most return of this function. In other words, encoding rule must be the same for all vecOp passed to `encode**` function like others.

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -849,6 +928,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
...
+	default:
+		panic("unsupported or illegal vecOp: " + op.String())
+	}
+}
+
+func vecArrangementParams(arr vecArrangement) (q uint32, size uint32) {
+	switch arr {
+	case vecArrangement8B:
+		q = uint32(0b0)
+		size = uint32(0b00)
```
mathetake, line 963: same

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -849,6 +928,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
...
+	case vecArrangement16B:
+		q = uint32(0b1)
+		size = uint32(0b00)
```
mathetake, line 966: same

internal/engine/wazevo/backend/backend_test.go
```diff
@@ -1426,9 +1426,19 @@ L1 (SSA Block: blk0):
...
+	uaddlv v15?.8B, v14?.8B
+	mov v6?.D[0], x6?
+	clz x7?, x3?
 	rbit x27, x3?
-	clz x7?, x27
+	clz x8?, x27
+	ins v10?.D[0], x3?
+	cnt v11?.16B, v10?.16B
+	uaddlv v12?.8B, v11?.8B
+	mov v9?.D[0], x9?
```
mathetake, line 1439: seems incorrect (e.g. where's v12 used?)

internal/engine/wazevo/backend/isa/arm64/instr.go
```diff
@@ -659,6 +667,36 @@ func (i *instruction) asFpuMov128(rd, rn regalloc.VReg) {
...
+
+func (i *instruction) asMovFromVec(rd, rn regalloc.VReg, arr vecArrangement, index vecIndex) {
+	i.kind = movFromVec
+	i.rd = operandNR(rd)
+	i.rn = operandNR(rn)
+	i.u1 = uint64(arr)
+	i.u2 = uint64(index)
+}
+
+func (i *instruction) asVecRRR(op vecOp, rd, rn regalloc.VReg, arr vecArrangement) {
```
mathetake, line 686: this should take `rm` as well (see `aluRRR`).
```suggestion
func (i *instruction) asVecRRR(op vecOp, rd, rn, rm regalloc.VReg, arr vecArrangement) {
```

internal/engine/wazevo/backend/isa/arm64/instr.go
```diff
@@ -881,9 +939,17 @@ func (i *instruction) String() (str string) {
 	case vecMiscNarrow:
 		panic("TODO")
 	case vecRRR:
-		panic("TODO")
+		str = fmt.Sprintf("%s %s, %s",
+			vecOp(i.u1),
+			formatVRegSizedVec(i.rd.nr(), 64, vecArrangement(i.u2), vecIndexNone),
+			formatVRegSizedVec(i.rn.nr(), 64, vecArrangement(i.u2), vecIndexNone),
+		)
```
mathetake, line 946: should be fixed accordingly (taking rm and rn)

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -849,6 +928,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
 	return _64bit<<31 | opc<<29 | 0b100110<<23 | _64bit<<22 | immr<<16 | imms<<10 | rn<<5 | rd
 }
 
+func encodeVecRRR(op vecOp, rd, rn uint32, arr vecArrangement) uint32 {
```
mathetake, line 931:
```suggestion
func encodeVecRRR(op vecOp, rd, rn, rm uint32, arr vecArrangement) uint32 {
```

internal/engine/wazevo/backend/isa/arm64/instr.go
```diff
@@ -182,6 +186,10 @@ var useKinds = [numInstructionKinds]useKind{
 	loadFpuConst64:  useKindNone,
 	cSel:            useKindRNRM,
 	fpuCSel:         useKindRNRM,
+	movToVec:        useKindRN,
+	movFromVec:      useKindRN,
+	vecMisc:         useKindRN,
+	vecRRR:          useKindRN,
```
mathetake, line 192:
```suggestion
	vecRRR:          useKindRNRM,
```

mathetake (commented):
OK, `VecRRR` is not the right place for `UADDLV`, but instead you should use `vecLanes`:

```go
	// vecLanes represents a vector instruction across lanes.
	vecLanes
```

which should be encoded following "Advanced SIMD across lanes" in https://developer.arm.com/documentation/ddi0596/2020-12/Index-by-Encoding/Data-Processing----Scalar-Floating-Point-and-Advanced-SIMD?lang=en

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -849,6 +928,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
...
+	case vecOpCnt:
+		var q, size uint32
+		if arr == vecArrangement8B {
+			q = uint32(0b0)
+			size = uint32(0b00)
+		} else {
+			q = uint32(0b1)
+			size = uint32(0b00)
+		}
+		return q<<30 | 0b001110<<24 | size<<22 | 0b100000010110<<10 | rn<<5 | rd
```
mathetake, line 952: this  also should be the outermost return, and the encoding can be shared with other "Advanced SIMD two-register miscellaneous".
https://developer.arm.com/documentation/ddi0596/2020-12/Index-by-Encoding/Data-Processing----Scalar-Floating-Point-and-Advanced-SIMD?lang=en

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
```diff
@@ -849,6 +928,48 @@ func encodeAluRRImm(op aluOp, rd, rn, amount, _64bit uint32) uint32 {
...
 
+func encodeVecRRR(op vecOp, rd, rn uint32, arr vecArrangement) uint32 {
+	switch op {
+	case vecOpUaddlv:
+		q, size := vecArrangementParams(arr)
+		return q<<29 | 0b101110<<24 | size<<22 | 0b110000001110<<10 | rn<<5 | rd
+	default:
+		panic("unsupported or illegal vecOp: " + op.String())
+	}
+}
```
mathetake, line 939:
```suggestion
```

internal/engine/wazevo/backend/backend_test.go
evacchi: likely bad String() for mov 🤔  (rd printed twice)

mathetake (approved):
Awesome!

PR #1642, opened 2023-08-21
wazevo(arm64): do not use x18 reserved by apple

mathetake:
The wazevo tests have started failing somehow only from Go 1.21 on GOOS=dawrin,
and it turns out we used the reserved "do not use" register, x18 by Apple:
https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Respect-the-purpose-of-specific-CPU-registers

Since AArch64 has lots of GP regs, we conservatively stop using x18 for all platforms (incl. linux),
but that can be relaxed later.

#1496


mathetake:
Cool. Luckily we are not using it in the current compiler as well, so we are secured 😄

PR #1648, opened 2023-08-22
wasi: fix symlink-related issues with path handling

evacchi:
- Fixes a bug in handling the buffer when the write goes over the given limit. ERANGE should be returned.
- Removes dependency from the syscall package in `fs.go`, `experimentalsys.O_WRONLY` should be used instead (portability)

### DIFF
--- a/experimental/sys/errno.go
+++ b/experimental/sys/errno.go
@@ -35,6 +35,7 @@ const (
 	ENOENT
 	ENOSYS
 	ENOTDIR
+	ERANGE
 	ENOTEMPTY
 	ENOTSOCK
 	ENOTSUP
@@ -79,6 +80,8 @@ func (e Errno) Error() string {
 		return "functionality not supported"
 	case ENOTDIR:
 		return "not a directory or a symbolic link to a directory"
+	case ERANGE:
+		return "result too large"
 	case ENOTEMPTY:
 		return "directory not empty"
 	case ENOTSOCK:
--- a/experimental/sys/syscall_errno.go
+++ b/experimental/sys/syscall_errno.go
@@ -40,6 +40,8 @@ func syscallToErrno(err error) (Errno, bool) {
 		return ENOSYS, true
 	case syscall.ENOTDIR:
 		return ENOTDIR, true
+	case syscall.ERANGE:
+		return ERANGE, true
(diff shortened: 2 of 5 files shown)


evacchi:
- `symlink_filestat`

      ==> wasi_snapshot_preview1.path_filestat_get(fd=5,flags=,path=symlink)
      <== (filestat={filetype=SYMBOLIC_LINK,size=4,mtim=1692707577959674593},errno=ESUCCESS)
      thread 'main' panicked at 'assertion failed: `(left == right)`
        left: `1692707577959674593`,
       right: `1692707577959674393`: symlink mtim should change', src/bin/symlink_filestat.rs:46:5


          // Check that symlink mtim motification worked
          let modified_sym_stat = wasi::path_filestat_get(dir_fd, 0, "symlink")
              .expect("reading file stats after path_filestat_set_times");
          assert_eq!(
              modified_sym_stat.mtim, sym_new_mtim,
              "symlink mtim should change"
          );


I noticed that [`utimens` actually supports a NOFOLLOW flag](https://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html#:~:text=fcntl.h%3E%3A-,AT_SYMLINK_NOFOLLOW,-If%20path%20names), which might allow us to avoid reopening the file (which I don't think is working as expected anyway; it looks like it is always dereferencing the link:

https://github.com/tetratelabs/wazero/blob/967d8df56d018386a821a3c340d4f085aa9750bd/imports/wasi_snapshot_preview1/fs.go#L1461-L1471

also notice that we are using syscall.O_WRONLY while I think it should be experimentalists.O_WRONLY. if I pass O_NOFOLLOW however I get back an ELOOP (weird?)

....however [adding a flag to Utimens will require us to break the API of the experimental VFS (`type FS` in `sys/fs.go`)](https://github.com/tetratelabs/wazero/blob/967d8df56d018386a821a3c340d4f085aa9750bd/experimental/sys/fs.go#L292). I think it may be acceptable since it is experimental, but breaking a public API is never nice. I am open to suggestions.

evacchi:
repurposing this PR to only fix the actual issues, dropping the edge cases for path handling in symlinks for now.

mathetake (approved):

mathetake:
good call!

PR #1664, opened 2023-08-28
wazevo: passes i32/i64 spectests 

mathetake:
#1496


internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -463,33 +534,25 @@ func (m *machine) lowerBitwiseAluOp(si *ssa.Instruction, op aluOp) {
 }
 
 func (m *machine) lowerRotl(si *ssa.Instruction) {
```
mathetake, line 536: fixed your bug FYI @evacchi

internal/engine/wazevo/backend/isa/arm64/cond.go
```diff
@@ -213,86 +213,3 @@ func condFlagFromSSAFloatCmpCond(c ssa.FloatCmpCond) condFlag {
 		panic(c)
 	}
 }
-
-// vecArrangement is the arrangement of data within a vector register.
```
mathetake, line 217: also I moved them from this file (these are nothing to do with "cond" ) FYI @evacchi

PR #1667, opened 2023-08-28
wazevo: passes float_exprs, local_set/tee spectests

evacchi:
follow up to https://github.com/tetratelabs/wazero/issues/1496.

- spec tests `local_set` and `local_tee` only missed one convert+one demote instruction to pass
- we also implement the other float conv instructions used in `float_exprs` so that we pass that spec test too


internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -205,6 +205,14 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
...
 		result := instr.Return()
 		m.lowerPopcnt(x, result)
+	case ssa.OpcodeFcvtToSint:
+		x := instr.Arg()
+		result := instr.Return()
+		m.lowerFpuToInt(result, x, true, x.Type() == ssa.TypeF64, result.Type().Bits() == 64)
+	case ssa.OpcodeFcvtToUint:
+		x := instr.Arg()
+		result := instr.Return()
+		m.lowerFpuToInt(result, x, false, x.Type() == ssa.TypeF64, result.Type().Bits() == 64)
```
mathetake, line 215: these are not simple in reality - these must check the range and NaN-ness like here https://github.com/tetratelabs/wazero/blob/main/internal/engine/compiler/impl_arm64.go#L2241-L2268

the trap assertions are part of "traps" spectest  which is yet to be run, but I will do the follow up. Meanwhile,

mathetake (approved):
I will do the follow up on the trap assertion

evacchi:
Thanks !

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
evacchi: Ah right

PR #1669, opened 2023-08-29
wazevo: passes traps, conversions spectest

mathetake:
#1496

### DIFF
--- a/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
+++ b/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
@@ -216,11 +216,15 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
 	case ssa.OpcodeFcvtFromSint:
 		x := instr.Arg()
 		result := instr.Return()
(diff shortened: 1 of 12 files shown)


internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -415,17 +427,47 @@ func (m *machine) lowerFpuUniOp(op fpuUniOp, in, out ssa.Value) {
 	m.insert(neg)
 }
 
-func (m *machine) lowerFpuToInt(dst, src ssa.Value, signed, src64bit, dst64bit bool) {
-	rn := m.getOperand_NR(m.compiler.ValueDefinition(src), extModeNone)
-	rd := operandNR(m.compiler.VRegOf(dst))
+func (m *machine) lowerFpuToInt(rd, rn operand, ctx regalloc.VReg, signed, src64bit, dst64bit bool) {
```
mathetake, line 430: fyi @evacchi this is the correct impl of FpuToInt

PR #1678, opened 2023-09-01
wazevo(regalloc): support for reg spills

mathetake:
This make it pass "skip-stack-guard-page" spectest!!

#1496


internal/engine/wazevo/e2e_test.go
```diff
@@ -99,7 +99,7 @@ func TestSpectestV1(t *testing.T) {
 		{name: "stack"},
 		{name: "store"},
 		{name: "switch"},
-		//{name: "skip-stack-guard-page"},
+		{name: "skip-stack-guard-page"},
```
mathetake, line 102: 🍺

PR #1685, opened 2023-09-04
wazevo: add nontrapping fp conversions (wasm v2)

evacchi:
moves forward #1496 with support for non-trapping floating point conversions to integer.

This also changes the signature of `lowerOpcode(op)` to `lowerCurrentOpcode()`. In order to parse `wasm.OpcodeMisc*TruncSat*` etc. it is necessary to parse `wasm.OpcodeMiscPrefix` first, and then advance to the next opcode, which will be in the range `[0x00..0x07]` They cannot be parsed otherwise, because then these opcodes would be mismatched in the `switch` statement for other unrelated opcodes.

Now `lowerCurrentOpcode()` instead moves `pc`itself; for this reason, all the related logic such as the debugging statements are now moved to this procedure, and  `return` statements had had to be changed to `break`s, otherwise the final `pc++` would be skipped in many cases.

### DIFF
--- a/internal/engine/wazevo/backend/backend_test.go
+++ b/internal/engine/wazevo/backend/backend_test.go
@@ -1730,21 +1730,13 @@ L1 (SSA Block: blk0):
 			afterFinalizeARM64: `
 L1 (SSA Block: blk0):
 	str x30, [sp, #-0x10]!
-	msr fpsr, xzr
 	fcvtzs x0, d0
-	msr fpsr, xzr
 	fcvtzs x1, s1
-	msr fpsr, xzr
 	fcvtzs w2, d0
-	msr fpsr, xzr
 	fcvtzs w3, s1
-	msr fpsr, xzr
 	fcvtzu x4, d0
-	msr fpsr, xzr
 	fcvtzu x5, s1
-	msr fpsr, xzr
 	fcvtzu w6, d0
-	msr fpsr, xzr
 	fcvtzu w7, s1
 	ldr x30, [sp], #0x10
 	ret
--- a/internal/engine/wazevo/backend/backend_test.go
+++ b/internal/engine/wazevo/backend/backend_test.go
@@ -1727,37 +1727,6 @@ L1 (SSA Block: blk0):
 		{
 			name: "nontrapping_float_conversions",
 			m:    testcases.NonTrappingFloatConversions.Module,
-			afterLoweringARM64: `
-L1 (SSA Block: blk0):
-	mov x0?, x0
-	mov v2?.8b, v0.8b
-	mov v3?.8b, v1.8b
-	msr fpsr, xzr
-	fcvtzs x4?, d2?
-	msr fpsr, xzr
-	fcvtzs x5?, s3?
-	msr fpsr, xzr
-	fcvtzs w6?, d2?
-	msr fpsr, xzr
-	fcvtzs w7?, s3?
-	msr fpsr, xzr
-	fcvtzu x8?, d2?
-	msr fpsr, xzr
-	fcvtzu x9?, s3?
-	msr fpsr, xzr
-	fcvtzu w10?, d2?
-	msr fpsr, xzr
-	fcvtzu w11?, s3?
-	mov x7, x11?
-	mov x6, x10?
-	mov x5, x9?
-	mov x4, x8?
-	mov x3, x7?
-	mov x2, x6?
-	mov x1, x5?
-	mov x0, x4?
-	ret
-`,
 			afterFinalizeARM64: `
 L1 (SSA Block: blk0):
 	str x30, [sp, #-0x10]!
--- a/internal/engine/wazevo/backend/backend_test.go
+++ b/internal/engine/wazevo/backend/backend_test.go
@@ -1725,6 +1725,63 @@ L1 (SSA Block: blk0):
 `,
 		},
 		{
+			name: "nontrapping_float_conversions",
+			m:    testcases.NonTrappingFloatConversions.Module,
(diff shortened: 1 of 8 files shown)


internal/engine/wazevo/backend/backend_test.go
```diff
@@ -1713,6 +1713,63 @@ L1 (SSA Block: blk0):
 	fcvt d1, s1
 	ldr x30, [sp], #0x10
 	ret
+`,
+		},
+		{
+			name: "nontrapping_float_conversions",
+			m:    testcases.NonTrappingFloatConversions.Module,
+			afterLoweringARM64: `
```
mathetake, line 1721: nothing interesting here, so let's not have `afterLoweringARM64` and only have `afterFinalizeARM64` (the assertion on `afterLoweringARM64` will be skipped if it is empty).

internal/engine/wazevo/frontend/lower.go
mathetake, line 1: could you revert this change?

evacchi: this was intended, `lowerOpcode(op)` operates on a given opcode, but in order to parse `wasm.OpcodeMisc*TruncSat*` etc. it is necessary to parse `wasm.OpcodeMiscPrefix` and then advance to the next opcode. Otherwise the opcodes for `wasm.OpcodeMisc*TruncSat*` clash with other unrelated opcodes. 

So I moved the logic to advance `pc` inside `lowerOpcode(op)`, and renamed it to `lowerCurrentOpcode()`. I also moved all the related logic such as the debugging statements.

I can do it differently, this seemed the most straightforward way.

mathetake: ah ok make sense! thank you for clarifying

evacchi: adding this to the top post for future reference

internal/engine/wazevo/frontend/lower.go
```diff
@@ -475,18 +470,49 @@ func (c *Compiler) lowerOpcode(op wasm.Opcode) {
...
 		).Insert(builder).Return()
 		state.push(ret)
+	case wasm.OpcodeMiscPrefix:
+		c.loweringState.pc++
+		// A misc opcode is encoded as an unsigned variable 32-bit integer.
+		miscOpUint, num, err := leb128.LoadUint32(c.wasmFunctionBody[c.loweringState.pc:])
+		if err != nil {
+			// fixme return err
+			panic(fmt.Sprintf("failed to read misc opcode: %v", err))
+		}
```
evacchi, line 490: in current compiler the procedure returns an error, should we do it here too? @mathetake

mathetake: no, this comes after function validation, so panic should be fine and shouldn't happen. See the code pass of wazero.Runtime.CompileModule!

internal/engine/wazevo/backend/backend_test.go
mathetake: @evacchi I believe this hasn't be resolved.

evacchi: maybe I reset it by mistake

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -427,7 +429,7 @@ func (m *machine) lowerFpuUniOp(op fpuUniOp, in, out ssa.Value) {
 	m.insert(neg)
 }
 
-func (m *machine) lowerFpuToInt(rd, rn operand, ctx regalloc.VReg, signed, src64bit, dst64bit bool) {
+func (m *machine) lowerFpuToInt(rd, rn operand, ctx regalloc.VReg, signed, src64bit, dst64bit, nonTrapping bool) {
 	// First of all, we have to clear the FPU flags.
```
mathetake, line 433: I believe this shouldn't be neede when non trapping

evacchi: that's ok, I only copied it because the original compiler is doing it

https://github.com/tetratelabs/wazero/blob/ea634990050bbe9af5ce230cce9811d54566c60f/internal/engine/compiler/impl_arm64.go#L2189-L2192

at this point I'd rather create a `lowerFpuToIntNontrapping()` since it becomes three lines. WDYT? or maybe add `lowerFpuToIntTrapping()` which invokes `lowerFpuToInt()` that becomes the 3-liner:


```
func lowerFpuToIntTrapping(...) {
   // clear flags
   lowerFpuToInt(...)
   // handle trap
}
func lowerFpuToInt(...) {
	cvt := m.allocateInstr()
	cvt.asFpuToInt(rd, rn, signed, src64bit, dst64bit)
	m.insert(cvt)
}
```

mathetake: up to you!

mathetake: either works for me 🍷

evacchi: I am keeping it simple (just if statements) because I guess I am still too jet-lagged to refactor it right :D basically the ctx register becomes useless for the nontrapping version, but that requires a few more changes to bubble up the chain and something was going wrong at the regalloc stage. I can refactor it later.

evacchi: I am marking this as ready, but if you'd rather me do the refactor I can do push it here later

mathetake (approved):
💯

PR #1689, opened 2023-09-07
fuzz: make it possible to fuzz wazevo

mathetake:
now we can run the fuzzing by passing the env:
```
$ WAZERO_FUZZ_WAZEVO=1 cargo fuzz run no_diff
```

#1496


mathetake:
cc @evacchi whenever you want to change the vive, you can run the fuzzing ^ and soon sure you will be able to find another bug you can work on 😄

PR #1690, opened 2023-09-07
wazevo: adds initial support for SIMD vectors

mathetake:
This makes wazevo pass the simd_const spectest.

#1496


internal/engine/wazevo/e2e_test.go
```diff
@@ -129,6 +129,7 @@ func TestSpectestV2(t *testing.T) {
 		name string
 	}{
 		{"conversions"}, // includes non-trapping conversions.
+		{"simd_const"},
```
mathetake, line 132: 👍

PR #1711, opened 2023-09-14
Use setup-go output for standard library test key

anuraaga:
Allows respecting patch version

Interestingly, it seems that based on the randomly selected runner, the go version can be different

https://github.com/tetratelabs/wazero/actions/runs/6180785720/job/16777777783?pr=1709
https://github.com/tetratelabs/wazero/actions/runs/6180785720/job/16778006464?pr=1709

One way to not have this happen is `check-latest`. It's funny how they recommend `false` for stability :P

https://github.com/actions/setup-go#check-latest-version

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -335,7 +335,8 @@ jobs:
           - "1.21"  # Current Go version && The only version that supports wasip1.
 
     steps:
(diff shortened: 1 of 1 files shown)


mathetake (approved):
works for me! thank you so much

PR #1712, opened 2023-09-14
wazevo: passes vec_arith2 spectests

evacchi:
#1496

- add missing simd_i8x16_arith
- move negative ops higher in the switch statement to respect the definition order
- add min_s, min_u, max_s, max_u, avgr_u, abs, vector popcnt correspoding to the arm64:
   - smin, umin, smax, umax
   - urhadd
   - abs
   - cnt (only for i8x16)


internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
mathetake, line 1: encoding tests?

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -317,13 +317,66 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
...
+		rm := m.getOperand_NR(m.compiler.ValueDefinition(y), extModeNone)
+		rd := operandNR(m.compiler.VRegOf(instr.Return()))
+		sub.asVecRRR(vecOpUmax, rd, rn, rm, arr)
+		m.insert(sub)
+	case ssa.OpcodeVAvgRound:
+		x, y, lane := instr.Arg2WithLane()
+		arr := ssaLeneToArrangement(lane)
+		sub := m.allocateInstr()
+		rn := m.getOperand_NR(m.compiler.ValueDefinition(x), extModeNone)
+		rm := m.getOperand_NR(m.compiler.ValueDefinition(y), extModeNone)
```
mathetake, line 361: could you share the code?

internal/engine/wazevo/backend/isa/arm64/instr_encoding.go
evacchi: Right! Will do!

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
evacchi: refactored to share. I could unit test the refactored function but it might be sort-of redundant, given there is already the encoding tests and the spec tests

mathetake (approved):
🏅

evacchi:
I forgot to update the description, this also passes add/sub sat

PR #1715, opened 2023-09-15
wazevo: passes simd_bitwise spec tests

evacchi:
follow-up to #1496. Passes simd_bitwise.

### DIFF
--- a/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
+++ b/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
@@ -323,9 +323,15 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
 		ins := m.allocateInstr()
 		rn := m.getOperand_NR(m.compiler.ValueDefinition(x), extModeNone)
 		rm := m.getOperand_NR(m.compiler.ValueDefinition(y), extModeNone)
-		rd := m.getOperand_NR(m.compiler.ValueDefinition(c), extModeNone)
-		ins.asVecRRR(vecOpBsl, rd, rn, rm, vecArrangement16B)
+		creg := m.getOperand_NR(m.compiler.ValueDefinition(c), extModeNone)
+		ins.asVecRRR(vecOpBsl, creg, rn, rm, vecArrangement16B)
 		m.insert(ins)
+
+		rd := m.compiler.VRegOf(instr.Return())
+		mov := m.allocateInstr()
+		mov.asFpuMov128(rd, creg.nr())
+
+		m.insert(mov)
 	case ssa.OpcodeVIadd:
 		x, y, lane := instr.Arg2WithLane()
 		arr := ssaLaneToArrangement(lane)
--- a/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
+++ b/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
@@ -346,10 +346,6 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
 		rd := m.getOperand_NR(m.compiler.ValueDefinition(c), extModeNone)
 		ins.asVecRRR(vecOpBsl, rd, rn, rm, vecArrangement16B)
 		m.insert(ins)
-	case ssa.OpcodeVanyTrue, ssa.OpcodeVallTrue:
-		m.lowerVcheckTrue(instr)
-	case ssa.OpcodeVhighBits:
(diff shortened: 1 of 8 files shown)


evacchi:
originally I wanted to tackle simd_logic too, but highbits seemed easier, so I'll get to that in another PR. This still needs encoding tests

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -299,28 +299,77 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
...
+		rn := m.getOperand_NR(m.compiler.ValueDefinition(x), extModeNone)
+		rm := m.getOperand_NR(m.compiler.ValueDefinition(y), extModeNone)
+		rd := m.getOperand_NR(m.compiler.ValueDefinition(c), extModeNone)
+		ins.asVecRRR(vecOpBsl, rd, rn, rm, vecArrangement16B)
+		m.insert(ins)
 	case ssa.OpcodeVIadd:
-		m.lowerVecRRR(vecOpAdd, instr)
+		x, y, lane := instr.Arg2WithLane()
+		arr := ssaLaneToArrangement(lane)
+		m.lowerVecRRR(vecOpAdd, x, y, instr.Return(), arr)
```
evacchi, line 332: shared the `lowerVecRRR()`'s signature to share it in more places

internal/engine/wazevo/ssa/instructions.go
```diff
@@ -851,6 +869,15 @@ var instructionSideEffects = [opcodeEnd]sideEffect{
 	OpcodeFabs:               sideEffectNone,
 	OpcodeFcopysign:          sideEffectNone,
 	OpcodeVconst:             sideEffectNone,
+	OpcodeVbor:               sideEffectNone,
+	OpcodeVbxor:              sideEffectNone,
+	OpcodeVband:              sideEffectNone,
+	OpcodeVbandnot:           sideEffectNone,
+	OpcodeVbnot:              sideEffectNone,
+	OpcodeVbitselect:         sideEffectStrict,
```
evacchi, line 877: this took a me little to figure out :^)

mathetake: why do you need strict? how is this order dependent?

internal/engine/wazevo/frontend/lower.go
```diff
@@ -1347,6 +1347,66 @@ func (c *Compiler) lowerCurrentOpcode() {
...
 			ret := builder.AllocateInstruction().AsVconst(lo, hi).Insert(builder).Return()
 			state.push(ret)
+		case wasm.OpcodeVecV128Load:
+			_, offset := c.readMemArg()
+			if state.unreachable {
+				break
+			}
+			opSize := uint64(16)
+			baseAddr := state.pop()
+			addr := c.memOpSetup(baseAddr, uint64(offset), opSize)
```
mathetake, line 1357:
```suggestion
			baseAddr := state.pop()
			addr := c.memOpSetup(baseAddr, uint64(offset), 16)
```

internal/engine/wazevo/ssa/instructions.go
evacchi: ha, good point. I picked "strict" only because it does not really trap, it's only effectful. I did not realize that was the most general one.

evacchi: fixed

mathetake (approved):
Thank you for following the discussion with me! do it once CI goes green ✅

internal/engine/wazevo/ssa/instructions.go
evacchi: ok for reference: it wasn't really fixed, and it was not due to the op being effectful. Well, not in the meaning I intended.

BSL overwrites one of its arguments (register `rd` which both contains the input mask _and_ is used as the output register); however, ssa being an ssa _always_ want an assignment (well, duh); however, register v5 was being killed because now:

```
[[[after lowering for [4/5] "bitselect" ]]]
L1 (SSA Block: blk0):
	mov v2?.16b, v0.16b
	mov v3?.16b, v1.16b
	mov v4?.16b, v2.16b
	bsl v4?.16b, v2?.16b, v3?.16b
	mov v0.16b, v5?.16b
	ret
```

with v5 being never defined.

so the fix is to add a  `mov v5, v4` so that now v5 is defined once (then the optimizer clears out all redundant moves)

PR #1716, opened 2023-09-18
experimental: removes Parameters API from StackIterator

mathetake:
as discussed offline in the Gophers slack, the Parameters API on 
the experimental StackIterator hasn't been used in practice, and 
also is difficult to implement in the optimizing compiler.
Therefore, we've decided to delete it from the API.


achille-roussel:
The best is we'll actually be able to drop some unnecessary code in https://github.com/stealthrocket/wzprof :D 

image

Thanks @mathetake !

PR #1718, opened 2023-09-18
wazevo(arm64): fixes VBitselect translation

mathetake:
follow up on #1715 

#1496


mathetake:
FYI @evacchi, I will work on adding test case which actually fails on the current impl tomorrow

mathetake:
@evacchi please take a look at https://github.com/tetratelabs/wazero/pull/1718/commits/d555169979574c9175e43ce5bb82f80591506275 when you have time, and let's keep that in our mind for the future logics

PR #1723, opened 2023-09-19
Supports compilation with GOOS=aix

ncruces:
Going through `go tool dist list` on `go version go1.21.1 darwin/amd64` this was the only target that didn't compile.

The specific issue is that on AIX `ENOTEMPTY` is the same as `EEXIST` so they can't both be in a switch.
But we decided not to fix similar issues either for `plan9` in #1603.
Since this is in `experimental`, I think that's the correct approach.


mathetake:
https://github.com/tetratelabs/wazero/blob/2686665e2a30faf91f28fd9af917920fa962523d/Makefile#L233


could you add test like ^^?

evacchi:
👁️ 🐝 Ⓜ️

evacchi (approved):

PR #1730, opened 2023-09-22
wazevo: passes simd_cmp spec tests

evacchi:
continues #1496, passes simd_cmp family of tests.
also checked against `DeterministicCompilationVerifierEnabled`

### DIFF
--- a/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
+++ b/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
@@ -154,6 +154,8 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
 		m.lowerCall(instr)
 	case ssa.OpcodeIcmp:
 		m.lowerIcmp(instr)
+	case ssa.OpcodeVIcmp:
+		m.lowerVIcmp(instr)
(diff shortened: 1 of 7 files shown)


internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -1012,6 +1014,73 @@ func (m *machine) lowerIcmp(si *ssa.Instruction) {
...
+	x, y, c, lane := si.VIcmpData()
+	flag := condFlagFromSSAIntegerCmpCond(c)
+	arr := ssaLaneToArrangement(lane)
+
+	var ext extMode
+	if c.Signed() {
+		ext = extModeSignExtend64
+	} else {
+		ext = extModeZeroExtend64
+	}
```
mathetake, line 1027: I don't believe this is needed. Could you give me the rationale

evacchi: it was originally there in `lowerIcmp()` for size == 64 (on which this was originally based; for a little, I tried to share the same function but it made little sense), but it is probably useless; tests still pass

mathetake (approved):

PR #1738, opened 2023-09-28
wazevo: simd bit shifts (wip)

evacchi:
continues #1496

PR #1740, opened 2023-10-01
wazevo: fix simd bit shift test temp reg usage

evacchi:
- continues #1496 
- addresses an incorrect usage of registers in #1738, removes useless instruction


mathetake (approved):
Awesome! Thank you for the fast turnaround!

PR #1755, opened 2023-10-04
Removes unnecessary mutex on Table/Memory Instance

mathetake:
This was _somehow_ introduced in the past, but the
reality is that the mutex only used during Grow doesn't
solve any concurrent issue at all and still could result in
crash when the module was accessed incorrectly concurrently.
The proper synchronization mechanism must be provided by
threads proposal, and this mutex usage can be removed.

### DIFF
--- a/internal/wasm/memory.go
+++ b/internal/wasm/memory.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"math"
 	"reflect"
-	"sync"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


ncruces:
Yes, using this only on Grow makes no sense, everything else would need to take the read lock.

Still, anything that touches memory (tables?) concurrently without the threads proposal is asking for trouble.

ncruces (approved):

mathetake:
@ncruces thank you for the look!

PR #1766, opened 2023-10-08
wazevo: passes simd load/store spec tests

evacchi:
continues #1496


internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -674,6 +674,33 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
...
+			rn = m.getOperand_NR(m.compiler.ValueDefinition(x), extModeNone)
+		default:
+			panic("unsupported lane " + lane.String())
+		}
+
+		// movToVec does not zero-out the other lanes,
+		// so we first zero-out explicitly.
+		mov0 := m.allocateInstr()
+		mov0.asLoadFpuConst64(rd.nr(), 0)
+		m.insert(mov0)
```
evacchi, line 698: does this make sense or is there an instruction that does zero + movToVec all at once?

internal/engine/wazevo/ssa/instructions.go
```diff
@@ -1748,6 +1742,15 @@ func (i *Instruction) AsVSshr(x, amount Value, lane VecLane) *Instruction {
 	return i
 }
 
+// AsScalarToVector initializes this instruction as move scalar to vector instruction with OpcodeScalarToVector.
```
mathetake, line 1745: could you fix the comment on `OpcodeScalarToVector` to reflect that it needs to zero out the higher bits?

internal/engine/wazevo/ssa/instructions.go
```diff
@@ -712,8 +712,7 @@ const (
 	// OpcodeBitcast is a bitcast operation: `v = bitcast MemFlags, x`.
 	OpcodeBitcast
 
-	// OpcodeScalarToVector ...
-	// `v = scalar_to_vector s`.
+	// OpcodeScalarToVector moves the given scalar value to a vector: `v = scalar_to_vector s`.
```
mathetake, line 715: could you reflect in the comment that it needs to zero out the higher bits?

internal/engine/wazevo/frontend/lower.go
```diff
@@ -1705,6 +1705,156 @@ func (c *Compiler) lowerCurrentOpcode() {
...
+			if state.unreachable {
+				break
+			}
+			baseAddr := state.pop()
+			addr := c.memOpSetup(baseAddr, uint64(offset), 8)
+			load := builder.AllocateInstruction().
+				AsLoad(addr, offset, ssa.TypeI64).
+				Insert(builder).Return()
+			ret := builder.AllocateInstruction().AsSplat(load, ssa.VecLaneI64x2).Insert(builder).Return()
+			state.push(ret)
```
mathetake, line 1856: this can use the `LoadSplat` which we should add and can be lowered as LD1R instruction at the backend https://developer.arm.com/documentation/ddi0596/2021-12/SIMD-FP-Instructions/LD1R--Load-one-single-element-structure-and-Replicate-to-all-lanes--of-one-register--


See the existing compiler https://github.com/tetratelabs/wazero/blob/991f4642c8f8e5f90be4c535d0eaa2cbb51035fa/internal/engine/compiler/impl_vec_arm64.go#L220-L246

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
mathetake: interesting. then why are we passing the tests with the old compiler which seems to use one load? https://github.com/tetratelabs/wazero/blob/main/internal/engine/compiler/impl_vec_arm64.go#L247-L262

internal/engine/wazevo/frontend/lower.go
```diff
@@ -1705,6 +1705,156 @@ func (c *Compiler) lowerCurrentOpcode() {
...
+			}
+			baseAddr := state.pop()
+			addr := c.memOpSetup(baseAddr, uint64(offset), 4)
+			load := builder.AllocateInstruction().
+				AsExtLoad(ssa.OpcodeUload32, addr, offset, false).
+				Insert(builder).Return()
+			ret := builder.AllocateInstruction().
+				AsScalarToVector(load, ssa.VecLaneI32x4).
+				Insert(builder).Return()
+			state.push(ret)
```
mathetake, line 1767: as I commented ^, it seems there should be a way to lower this as one instruction https://github.com/tetratelabs/wazero/blob/main/internal/engine/compiler/impl_vec_arm64.go#L247-L262

internal/engine/wazevo/frontend/lower.go
```diff
@@ -1705,6 +1705,156 @@ func (c *Compiler) lowerCurrentOpcode() {
...
+			}
+			baseAddr := state.pop()
+			addr := c.memOpSetup(baseAddr, uint64(offset), 8)
+			load := builder.AllocateInstruction().
+				AsLoad(addr, offset, ssa.TypeI64).
+				Insert(builder).Return()
+			ret := builder.AllocateInstruction().
+				AsScalarToVector(load, ssa.VecLaneI64x2).
+				Insert(builder).Return()
+			state.push(ret)
```
mathetake, line 1782: same

internal/engine/wazevo/frontend/lower.go
mathetake: maybe we won't need `OpcodeScalarToVector` in the first place

mathetake:
could you remove this skips and ensure they are passing? https://github.com/tetratelabs/wazero/blob/main/internal/integration_test/fuzzcases/fuzzcases_test.go#L53-L57

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
evacchi: Yeah I noticed and I am confused. I will investigate

evacchi:
re-drafted, will add LoadSplat ssa instruction

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
evacchi: it should be ok now

internal/engine/wazevo/frontend/lower.go
evacchi: done

internal/engine/wazevo/frontend/lower.go
evacchi: removed

evacchi:
what the hell is happening on GHA

> internal/engine/wazevo/backend/isa/arm64/lower_instr.go:683:37: cannot use regalloc.RegTypeInt (constant 1 of type regalloc.RegType) as ssa.Type value in argument to m.compiler.AllocateVReg

PR #1775, opened 2023-10-12
fuzz: documents to use --sanitizer=none

mathetake:
FYI @evacchi


evacchi:
should we update the Makefile too?

mathetake:
Oh yeah good catch

PR #1792, opened 2023-10-16
wazevo: add fuzz test cases

evacchi:
continues #1496

- v128.const should skip on state.unreachable 
- OpcodeVhighBits returns V128 
- OpcodeVFcmp returns V128
- Fix / Improve some error/log messages


internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -437,3 +437,43 @@ func Test1777(t *testing.T) {
 		require.Equal(t, []uint64{18446626425965379583, 4607736361554183979}, res)
 	})
 }
+
+// Test2000 tests that v128.const i32x4 is not skipped when state is unreachable.
```
evacchi, line 441: I was not sure how the tests were originally numbered because from what I can see the fuzzer names the tests as a hex string... so I just took a high integer after 1777 (2000),

mathetake:
spec test is failing

internal/integration_test/fuzzcases/fuzzcases_test.go
evacchi: can do. Since it's 3 tests I could change it to a string and use "1792a", "1792b", "1793c"

mathetake: sg

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -1516,10 +1516,6 @@ func (m *machine) lowerVFcmp(si *ssa.Instruction) {
 		cmp.asVecRRR(vecOpFcmge, rd, rm, rn, arr) // rm, rn are swapped
 		m.insert(cmp)
 	}
-
-	cset := m.allocateInstr()
-	cset.asCSet(rd.reg(), flag)
-	m.insert(cset)
```
mathetake, line 1522: lol

mathetake (approved):

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
evacchi: yeah, same thought 🤣

mathetake (approved):

PR #1793, opened 2023-10-17
wazevo: fuzz OpcodeVIcmp, OpcodeVFcmp, allTrue

evacchi:
continues #1496

- wazevo: fuzz, fix allTrue registers
- fix OpcodeVIcmp, OpcodeVFcmp
- fix lowerVIcmp
- fix lowerVShift

the common trait is that most of these I did initially when I was still using `rd` freely as a temp register 😅 maybe I'll review the instructions to fix this issue in most places...


mathetake (approved):
Sounds like it was fun right 😄

PR #1802, opened 2023-10-19
wazevo: add VRegTable data structure

achille-roussel:
This PR explores changing the data structure used for _last uses_ in block info.

The change takes the compilation of Python 3.11 with Wazevo from 40s to 28s.

I added a bitset to perform fast checks of virtual registers on the hot code path of `beginUpAndMarkStack`, which yielded most of the gains (changing from a map to a simple array only gave a ~17% improvement).

Note that even with this change, the hottest code path is still `beginUpAndMarkStack`, but we're starting to see a more balanced time spent in various parts:

image

Based on what I'm seeing in the current profile, we're starting to see the impact of indirect function calls through method interfaces and such (e.g., 7%+ spent in calling `Block.ID`). There is likely some room for reducing the number of those calls on the hottest code paths to amortize their cost.

My only concern is that I don't understand the lifecycle of `blockInfo` values nor how virtual register IDs are allocated, which might increase the memory footprint. We can do more investigation as part of this PR, or we might look into this in a follow-up.

Let me know what you think!

### DIFF
--- a/internal/engine/wazevo/backend/regalloc/reg.go
+++ b/internal/engine/wazevo/backend/regalloc/reg.go
@@ -67,7 +67,7 @@ func (v VReg) Valid() bool {
 // virtual registers.
 //
 // We store the min values + 1 so the zero-value of the VRegIDMinSet is valid.
-type VRegIDMinSet [numRegTypes]VRegID
+type VRegIDMinSet [NumRegType]VRegID
 
 func (mins *VRegIDMinSet) Min(t RegType) VRegID {
 	return mins[t] - 1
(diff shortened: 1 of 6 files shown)


internal/engine/wazevo/backend/regalloc/reg.go
```diff
@@ -96,6 +213,7 @@ const (
 	RegTypeInt
 	RegTypeFloat
 	RegTypeNum
+	numRegTypes // keep last
```
mathetake, line 216: I think we can use `RegTypeNum` for numRegTypes :D 
```suggestion
	RegTypeNum
```

mathetake (approved):
Thank you so much @achille-roussel, this looks good and we can merge after the test is fixed! As for the concern about the lifecycle, I see no problem with this change

mathetake:
btw can you do the same thing to liveIns, liveOuts and defs? Does that help in the performance?

achille-roussel:
That was going to be be next thing I wanted to try 👍

internal/engine/wazevo/backend/regalloc/reg.go
achille-roussel: Thanks for pointing out that the constant already existed. I got confused at first and assumed `RegTypeNum` was a type of registry (like _number_ 🤷) so I took the liberty to rename the constant to `NumRegType` to differentiate it and make it more clear that it's not representing a registry type. I can revert if you prefer the other spelling; let me know!

PR #1803, opened 2023-10-19
wazevo: reduce memory footprint of regalloc kills

achille-roussel:
This a small optimization that I noticed while going through the code: the regalloc `kills` are always derived from `lastUses`, which can avoid tracking the program counters in `kills` and turn the collection into a set that is only tracking virtual registers. When we need the `kills` program counters, we can look them back up into `lastUses`.

On the Python compilation, this saves 55MB of a 248MB cumulative allocation:
```
github.com/tetratelabs/wazero/internal/engine/wazevo/backend/regalloc/regalloc.go

  Total:    248.01MB   248.01MB (flat, cum)  1.99%
    249            .          .           			if _, ok := outs[use]; !ok { 
    250     248.01MB   248.01MB           				info.kills[use] = pc 
    251            .          .           			} 
    252            .          .           		}) 
    253            .          .            
    254            .          .           		if wazevoapi.RegAllocLoggingEnabled { 
    255            .          .           			fmt.Printf("\nfinalized info for block[%d]:\n%s\n", blk.ID(), info.Format(a.regInfo)) 
```
```
github.com/tetratelabs/wazero/internal/engine/wazevo/backend/regalloc/regalloc.go

  Total:    -55.36MB   -55.36MB (flat, cum)  0.22%
    249            .          .           			if _, ok := outs[use]; !ok { 
    250     -55.36MB   -55.36MB           				info.kills[use] = struct{}{} 
    251            .          .           			} 
    252            .          .           		}) 
    253            .          .            
    254            .          .           		if wazevoapi.RegAllocLoggingEnabled { 
    255            .          .           			fmt.Printf("\nfinalized info for block[%d]:\n%s\n", blk.ID(), info.Format(a.regInfo)) 
```

As a side note, this also makes the compilation of Python ~1s faster on a M2.

There is likely a lot of room for improvement here by replacing the maps with a bitset, but I wanted to make this optimization first to get a baseline.

### DIFF
--- a/internal/engine/wazevo/backend/regalloc/regalloc.go
+++ b/internal/engine/wazevo/backend/regalloc/regalloc.go
@@ -79,7 +79,7 @@ type (
 		liveIns  map[VReg]struct{}
 		defs     map[VReg]programCounter
 		lastUses VRegTable
-		kills    map[VReg]programCounter
+		kills    map[VReg]struct{}
 		// Pre-colored real registers can have multiple live ranges in one block.
 		realRegUses [vRegIDReservedForRealNum][]programCounter
 		realRegDefs [vRegIDReservedForRealNum][]programCounter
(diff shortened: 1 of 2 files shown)


mathetake (approved):
good catch

Issue #1495, opened 2023-05-30
WASIX Support


moloch--:
This is a longer term feature request, but would be great to have WASIX support in Wazero; 
* https://wasmer.io/posts/announcing-wasix
* https://wasix.org/

ncruces:
The important bit to track is:
https://wasix.org/docs/api-reference#wasix-functions

The thread related stuff seems to depend on the [phase 3](https://github.com/WebAssembly/proposals/blob/main/README.md) [threads proposal](https://github.com/WebAssembly/threads) (which we deferred in #1487).

codefromthecrypt:
@moloch-- so we currently have a table of supported functions in [wasi_snapshot_preview1](https://github.com/tetratelabs/wazero/blob/main/site/content/specs.md#webassembly-system-interface-wasi-wasi), so I imagine this would require a table for the extensions like [clock_time_set](https://wasix.org/docs/api-reference/wasix/clock_time_set).

I only just saw the docs today.. do you know what module name the extensions go into? Also, when we implemented WASI in the first place, we did things priority order instead of big bang. Can you mention which specific functions you need implemented and how you are using them (rust or c)?

codefromthecrypt:
Also, there's recently a project that is more experimentation friendly https://github.com/stealthrocket/wasi-go One way to proceed is to develop the features there (as there's no back compat or windows requirements), then once stable we can add them here.

ncruces:
> do you know what module name the extensions go into?

It's `wasix_32v1`, [apparently](https://github.com/wasmerio/wasmer/blob/554b20273aa71e25b28f6f6133740fb3d19bea1e/lib/wasi/src/utils/mod.rs#L131-L132).

codefromthecrypt:
@moloch--  by the way, as wasix is a superset of wasip1, I expect its socket support to be "mix and match" capable. watch releases as I am writing notes about v1.2.0 which [adds experimental support for the socket functions](https://github.com/tetratelabs/wazero/pull/1493) defined in wasip1. There are glitches around non-blocking I/O, still, so if interested you can follow [this](https://github.com/tetratelabs/wazero/issues/1500#issuecomment-1574623676). It would be good to know which functions python uses, as I'm not currently sure if it uses the existing wasip1 ones or not..

PR #1808, opened 2023-10-20
Add experimental support for snapshot/restore

anuraaga:
This allows a user to opt-in to having snapshot/restore functionality exposed to host functions. This can be used to implement host-based exception handling, for example with `wasix`'s implementation of `setjmp` and `longjmp`. The basic machinery can also likely be reused if implementing WebAssembly exceptions in the future.

This is intended to stay in experimental as there is no official standard for exceptions yet, but as there are compilers such as wasix SDK that can compile them, this unblocks allowing wazero to run such programs. For example, the sqlc tool is popular in Go but has cgo issues, this is a step towards using wasix to compile it's Postgres parser and have a pure Go sqlc binary.

Discussed in this slack thread: https://gophers.slack.com/archives/C04CG4A2NKX/p1697555612115509?thread_ts=1697441543.851059&cid=C04CG4A2NKX

wasix
stack_checkpoint: https://wasix.org/docs/api-reference/wasix/stack_checkpoint
stack_restore: https://wasix.org/docs/api-reference/wasix/stack_restore

Example usage in wasix built app:
https://github.com/wasilibs/go-pgquery/blob/main/internal/wasix_32v1/wasix.go#L130


experimental/checkpoint_test.go
```diff
@@ -0,0 +1,122 @@
...
+	require.NoError(t, err)
+	// snapshot returned zero
+	require.Equal(t, uint64(0), res[0])
+
+	// Fails, snapshot and restore are called from different wasm invocations. Currently, this
+	// results in a panic.
+	err = require.CapturePanic(func() {
+		_, _ = mod.ExportedFunction("restore").Call(ctx, snapshotPtr)
+	})
+	require.EqualError(t, err, "unhandled snapshot restore, this generally indicates restore was called from a different "+
```
anuraaga, line 120: Assuming `ctx` is propagated through a nested invocation, I think it'd be possible to handle this as an error, but I don't know if it's worth it since it would be an error either way but add some tricky handling

ncruces:
Issue #1495 is relevant.

internal/engine/compiler/engine.go
```diff
@@ -1065,12 +1074,27 @@ entry:
...
 
 			fn := calleeHostFunction.parent.goFunc
-			switch fn := fn.(type) {
-			case api.GoModuleFunction:
-				fn.Call(ctx, ce.callerModuleInstance, stack)
-			case api.GoFunction:
-				fn.Call(ctx, stack)
-			}
+			func() {
+				defer func() {
```
mathetake, line 1078: this defer causes allocation every time a Go host func is called. would it be possible to avoid when snapshot is disabled?

anuraaga:
@ncruces Just to refresh memory :)

https://gophers.slack.com/archives/C04CG4A2NKX/p1698068979958989?thread_ts=1697441543.851059&cid=C04CG4A2NKX

The biggest reason for holding off at the time was that a "real world test", aka using wasix-sdk, requires enabling threads. Now that wazero has threads support, the issue is gone or at least reduced.

internal/engine/compiler/engine.go
```diff
@@ -1065,12 +1076,29 @@ entry:
...
-				fn.Call(ctx, ce.callerModuleInstance, stack)
-			case api.GoFunction:
-				fn.Call(ctx, stack)
-			}
+			func() {
+				if snapshotEnabled {
+					defer func() {
+						if r := recover(); r != nil {
+							if s, ok := r.(*snapshot); ok {
+								if s.ce == ce {
```
ncruces, line 1084: Can we do a single `ok && s.ce == ce` and merge the else clauses?

anuraaga:
Great point, that is indeed my intention so I should verify with a replace directive, I'll check it tomorrow.

experimental/checkpoint.go
```diff
@@ -0,0 +1,25 @@
...
+// Snapshotter allows host functions to snapshot the WebAssembly execution environment.
+type Snapshotter interface {
+	// Snapshot captures the current execution state.
+	Snapshot() Snapshot
+}
+
+// EnableSnapshotterKey is a context key to indicate that snapshotting should be enabled.
+// The context.Context passed to a exported function invocation should have this key set
+// to a non-nil value, and host functions will be able to retrieve it using SnapshotterKey.
+type EnableSnapshotterKey struct{}
```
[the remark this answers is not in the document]
anuraaga: I followed the pattern of function listener when writing this, I could change it to the sock pattern instead though. Any suggestion on what package the internal context key would go in?

internal/engine/compiler/engine.go
anuraaga: Sweet, thanks!

anuraaga (commented):
@ncruces Here I confirm replacing wazerox with wazero

https://github.com/wasilibs/go-pgquery/pull/23

As for the juggling, I guess it's mostly just stubbing out unused functions that make it into the binary

https://github.com/wasilibs/go-pgquery/blob/main/internal/wasix_32v1/wasix.go

I don't think we can say it's really easy to use wasix with wazero but I guess it's not terrible either

ncruces:
> I don't think we can say it's really easy to use wasix with wazero but I guess it's not terrible either

This is good enough. I mean, we can't correctly extend wasi into wasix outside wazero because em we'd need access to internals of wasi implementation (e.g. filesystem).

But you can usefully stub stuff emitted by wasix-sdk, or you can likely implement wasix from scratch outside wazero (same as stealthrocket/wasi-go).

Good enough for me.

ncruces (approved):
I'd like a second opinion on the enabling API, but other than that, LGTM.

experimental/checkpoint_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"context"
+	"testing"
+
+	"github.com/tetratelabs/wazero"
+	"github.com/tetratelabs/wazero/api"
+	"github.com/tetratelabs/wazero/experimental"
+	"github.com/tetratelabs/wazero/internal/testing/require"
+)
+
+func TestSnapshotNestedWasmInvocation(t *testing.T) {
```
mathetake, line 13: could you move this to internal/integration_test/adhoc.go? That's where we ensure the consistency across engines.

internal/engine/compiler/engine.go
```diff
@@ -787,7 +787,13 @@ func (ce *callEngine) call(ctx context.Context, params, results []uint64) (_ []u
 		defer done()
 	}
 
-	ce.execWasmFunction(ctx, m)
+	snapshotEnabled := false
+	if ctx.Value(experimental.EnableSnapshotterKey{}) != nil {
+		ctx = context.WithValue(ctx, experimental.SnapshotterKey{}, ce)
+		snapshotEnabled = true
+	}
```
mathetake, line 794:
```suggestion
	snapshotEnabled := ctx.Value(experimental.EnableSnapshotterKey{}) != nil
	if snapshotEnabled {
		ctx = context.WithValue(ctx, experimental.SnapshotterKey{}, ce)
	}
```

experimental/checkpoint_test.go
anuraaga: Sure I can try it. Two issues

- This reuses the same snapshot.wasm used for the example test. Should I just copy the wasm?
- It's highly experimental, is it ok to have tests in the core area?

internal/engine/compiler/engine.go
anuraaga: 🤦

experimental/checkpoint_test.go
mathetake: ok let's leave it as-is now!

mathetake (approved):

PR #1809, opened 2023-10-20
config: add support for cancellable Nanosleep

evacchi:
It has been reported that in some cases (e.g. in `poll_oneoff`) the context is not honored on **sleep**. We introduce support for a configurable cancellable context, and we make it the default for `ModuleConfig.WithSysNanosleep()`.

We also introduce `WithCancellableNanosleep(sys.CancellableNanosleep)` alongside `WithNanosleep(sys.Nanosleep)` which should be a source-compatible change (advice welcome on whether this is a correct assumption).

Changing the default for `ModuleConfig.WithSysNanosleep()` is a behavioral change that should have no effect (it is actually a bug if you relied on the context not being cancellable), but it might be considered "breaking", so we can revert that if we'd rather keep the old default.

We can also deprecate `WithNanosleep(sys.Nanosleep)`, but for now I am just modifying the documentation to nudge users towards the new API.

### DIFF
--- a/imports/wasi_snapshot_preview1/poll.go
+++ b/imports/wasi_snapshot_preview1/poll.go
@@ -49,7 +49,7 @@ type event struct {
 	errno     wasip1.Errno
 }
 
-func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno {
+func pollOneoffFn(ctx context.Context, mod api.Module, params []uint64) sys.Errno {
 	in := uint32(params[0])
 	out := uint32(params[1])
 	nsubscriptions := uint32(params[2])
@@ -166,7 +166,7 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno 
 		// We only need to observe the timeout (nonzero if there are clock subscriptions)
 		// and return.
 		if timeout > 0 {
-			sysCtx.Nanosleep(int64(timeout))
+			sysCtx.Nanosleep(ctx, int64(timeout))
 		}
 		return 0
 	}
--- a/config.go
+++ b/config.go
@@ -606,11 +606,32 @@ type ModuleConfig interface {
 	//
 	// # Notes:
 	//   - This does not default to time.Sleep as that violates sandboxing.
+	//   - Some compilers implement sleep by looping on sys.Nanotime (e.g. Go).
+	//   - If you set this, you should probably set WithNanotime also.
+	//   - This function does not honor a context cancellation: when in doubt,
+	//     you should always prefer WithCancellableNanosleep, especially if used
+	//     together with WithCloseOnContextDone.
+	WithNanosleep(sys.Nanosleep) ModuleConfig
+
+	// WithCancellableNanosleep configures the how to pause the current goroutine for at
+	// least the configured nanoseconds. Defaults to return immediately.
+	//
+	// This example uses a custom sleep function:
+	//	moduleConfig = moduleConfig.
+	//		WithCancellableNanosleep(func(ctx context.Context, ns int64) {
+	//			select {
+	//			case <-ctx.Done(): // Return on context done.
+	//			case <-time.After(time.Duration(ns)):
+	//			}
+	//		})
+	//
+	// # Notes:
+	//   - This does not default to time.Sleep as that violates sandboxing.
 	//   - This is used to implement host functions such as WASI `poll_oneoff`.
 	//   - Some compilers implement sleep by looping on sys.Nanotime (e.g. Go).
 	//   - If you set this, you should probably set WithNanotime also.
 	//   - Use WithSysNanosleep for a usable implementation.
-	WithNanosleep(sys.Nanosleep) ModuleConfig
+	WithCancellableNanosleep(sys.CancellableNanosleep) ModuleConfig
 
 	// WithOsyield yields the processor, typically to implement spin-wait
 	// loops. Defaults to return immediately.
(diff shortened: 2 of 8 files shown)


imports/wasi_snapshot_preview1/poll.go
```diff
@@ -166,7 +166,7 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno
 		// We only need to observe the timeout (nonzero if there are clock subscriptions)
 		// and return.
 		if timeout > 0 {
-			sysCtx.Nanosleep(int64(timeout))
+			sysCtx.Nanosleep(ctx, int64(timeout))
```
ncruces, line 169: Is this really enough? This is not keeping the typical promise of sleep: sleeping _at least_ `n`ns.

Are we sure wazero immediately errors out on the context after this? Or is it possible for guest code to continue assuming poll just timed out?

evacchi: I should probably add an integration test

codefromthecrypt:
random drive by:

Let's be specific about reports, for example if they are from end users or integrators, and what is the log or code leading up to it. If there's a trace of sleep you can see how impactful it is, because there is another decision to make.. whether to do this at all or not, and that's hard to do without a sense of impact. This is why I tend to recommend having end users raise their own issues, as you can get some of this info, and if it isn't worth it to them to raise an issue, maybe adding a new API is too early. For example, does the end user plan to actually use this? What is the cost benchmark wise of doing it like this? We don't need to frontline work and design iterations if this is not going to be used, or only used by one person. Anything that changes the public API should be very justified, IOTW.

If we get past the point above, then we can start about the other things this does, such as introduce a concept we don't currently have "deprecation". We have simpler APIs which have more complex or advanced alternatives available (such as FSConfig), but unless I missed something, we have no deprecation concept here and doing so would hint at a wazero 2 which may never happen.

fariedt:
I asked about the problem in the Gophers slack, @codefromthecrypt.  I didn't open an issue because I didn't know if I was using wazero incorrectly, or if it was a problem with wazero itself.

Here's my message on Slack, with a little more detail:

I have a web server that loads a few wasm modules and exposes them over http.  I use this config:
```
config := wazero.NewRuntimeConfig().WithCompilationCache(cache).WithCloseOnContextDone(true)
wruntime = wazero.NewRuntimeWithConfig(context.Background(), config)
[...]
modConfig := wazero.NewModuleConfig().WithSysWalltime().WithStdout(os.Stdout).WithStderr(os.Stderr).WithSysNanosleep()
```

I need `.WithSysNanosleep()` so that my wasm code can sleep.  I use tinygo for testing, so that's `time.Sleep(...)` in my guest code.  (In production, I won't have any control over which tool generates the wasm code -- anything that targets wasi should work.)

I instantiate the wasm modules and close them in the http handler with something like this:
```
ctx := r.Context()
timeoutctx, cancel := context.WithTimeout(ctx, 5*time.Second)
defer cancel()

mod, _ := wruntime.InstantiateModule(ctx, compiled, modConfig)
defer mod.Close(ctx)

callfn := mod.ExportedFunction("call")
ret, err := callfn.Call(timeoutctx, 0, 0)
if err { ... }
```
If the wasm code a forever loop (`for { }`) it's cancelled after five seconds, as I'd expect.  However, if it calls `time.Sleep(20*time.Second)`, it continues to run past the limit.

ncruces:
Just for the sake of condensing the Slack discussion here for future reference (please also add your comments here if you feel they're relevant), I'm worried that this creates an expectation for _other_ WASI syscalls, and host function calls in general.

For WASI syscalls, sleep was the easy one. In general getting Go to timeout on a `Context` for things like (file/network) `Read` and `Write` is _much_ harder. For host functions in general we could at best warn the user they should respect `Context` cancellation if it matters to _them_. Anything more complicated is over engineering, IMO.

So, again IMO, it boils down to: is sleep special enough that we should special case it, also because we can? If so, LGTM, _if we can ensure the guest does not run anymore code beyond this cancelled sleep_ (my comment above).

Or we decide that the host being able to kill the guest wherever it might be stuck (at least for WASI) is important enough. But that's a huge undertaking, IMO.

evacchi:
[for further record, I am not particularly attached to this PR, I just opened it because it was simple enough and as a way to start the discussion (as we are doing)]

codefromthecrypt:
In summary, at most do it in experimental, but resist complicating the whole codebase trying to make everything else do similar (yield, poll, etc). If only handling sleep works out for significant shared use cases, it could be worth promoting and documenting despite being a half measure.

---

Thanks for the feedback folks, particularly @fariedt for piping in. My sentiment was well summarized by @ncruces:

Breaking out of a large magnitude sleep is something we could address as a special case, but it wouldn't make problems of unblocking go away (think blocking read). In other words, we would patch one of many ways a guest could block.

A different workaround could be to wrap a sleep function with something that reduces large magnitude to something smaller. like `maxSleep(sys.Nanotime, 50*milliseconds)`. Something like this could be used while a special case context feature popularizes and lands.

If this PR goes anywhere I think the maximum blast it should have is being an experimental thing. Experimental lets folks try this knowing it doesn't solve sandboxing WASI in general. Checking back about success (I guess code that only sleeps yet doesn't do I/O), could result in a sensible promotion from experimental. After that, we could get into nuance of how to talk about both ways to configure sleep without making things confusing, as well the relationship to other things like sleep (such as Osyield or poll)

fariedt:
My end goal is to create and document a plugin facility for a service written in Go.  Plugins using wasm will let (trusted) end users add functionality without learning Go, modifying the server's code, restarting it, etc.  I ran into the sleep issue while experimenting with ways that I could catch runaway processes.  (Another issue that I have to think about is limiting disk usage by a guest -- also not a wazero problem.)

Given the discussion above, I am okay with my guest code not being able to sleep; I would rather not have it special cased in wazero.  I think documenting this problem somewhere would help future users of the library.

Thanks!

evacchi:
we can close this for now. As an action item, I'll add/consolidate a note about syscalls+blocking in RATIONALE.md

PR #1814, opened 2023-10-23
wazevo: fuzz, select does not return the correct value for vectors

evacchi:
#1496

### DIFF
--- a/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
+++ b/internal/engine/wazevo/backend/isa/arm64/lower_instr.go
@@ -1873,21 +1873,40 @@ func (m *machine) lowerSelect(c, x, y, result ssa.Value) {
 }
 
 func (m *machine) lowerSelectVec(rc, rn, rm, rd operand) {
-	tmp := operandNR(m.compiler.AllocateVReg(ssa.TypeI64))
+	ifNonZero := m.allocateLabel()
+	ifNonZeroPos := m.allocateLabelPosition()
+	end := m.allocateLabel()
+	endPos := m.allocateLabelPosition()
 
-	// Sets all bits to 1 if rc is not zero.
-	alu := m.allocateInstr()
-	alu.asALU(aluOpSub, tmp, operandNR(xzrVReg), rc, true)
-	m.insert(alu)
+	cbr := m.allocateInstr()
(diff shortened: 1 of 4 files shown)


internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -1873,21 +1873,40 @@ func (m *machine) lowerSelect(c, x, y, result ssa.Value) {
...
+	ifnzTgt := m.allocateInstr()
+	ifnzTgt.asNop0WithLabel(ifNonZero)
+	m.insert(ifnzTgt)
+
+	mov := m.allocateInstr()
+	mov.asFpuMov128(rd.nr(), rn.nr())
+	m.insert(mov)
+
+	ifNonZeroPos.begin, ifNonZeroPos.end = ifnzTgt, ifnzTgt
+	m.labelPositions[ifNonZero] = ifNonZeroPos
```
mathetake, line 1904: plz use `insertBrTargetLabel`  instead to avoid these lines

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -1873,21 +1873,40 @@ func (m *machine) lowerSelect(c, x, y, result ssa.Value) {
...
+
+	endTgt := m.allocateInstr()
+	endTgt.asNop0WithLabel(end)
+	m.insert(endTgt)
+
+	endPos.begin, endPos.end = endTgt, endTgt
+	m.labelPositions[end] = endPos
```
mathetake, line 1911: same

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
evacchi: I wanted to use `insertBrTargetLabel()` but isn't that only for jump backwards? it inserts a labeled nop immediately, while here I need to allocate a label, point to it in the `condBr`, and then insert the `nop` at the right spot.

evacchi: done, let me know if this is good.

mathetake: I'm going to sleep but I believe you should be able to whoever what you want without adding new change like that. Basically checking other places in the lower_instr would help

PR #1815, opened 2023-10-24
compiler: add munmap finalizer on cache hits to avoid memory leak

valpackett:
Not sure how this was never noticed, trying to run the same thing over and over in response to request with cache enabled would OOM very quickly…

Screenshot from 2023-10-23 19-08-13

### DIFF
--- a/internal/engine/compiler/engine_cache.go
+++ b/internal/engine/compiler/engine_cache.go
@@ -46,6 +46,9 @@ func (e *engine) getCompiledModule(module *wasm.Module, listeners []experimental
 				cm.functions[i].listener = listeners[i]
 			}
 		}
+
+		// As this uses mmap, we need to munmap on the compiled machine code when it's GCed.
+		e.setFinalizer(cm, releaseCompiledModule)
 	}
 	return
 }


ncruces:
Thanks for the report. I don't think this is the correct/best place to fix?

I think this should be added somewhere to this function:
https://github.com/tetratelabs/wazero/blob/610fdbd6642867a553ef19ce74df2c0396db8f14/internal/engine/compiler/engine_cache.go#L133

cc: @mathetake

valpackett:
Seemed like the place with least duplication and seemed good that it's the same function as the non-cached path. Inside `getCompiledModule` I thought in some way does make sense too, but there would need to be duplicated for the memory hit and disk hit paths.

ncruces:
The memory cache hit doesn't create a new instance, so it doesn't _need_ one?

You're actually replacing the finalizer that's already there, which probably causes no harm, but is unnecessary.

valpackett:
Oh, makes sense. I see wazevo actually does set the finalizer and it's in `getCompiledModule` for the disk branch, so this should match that now.

mathetake (approved):
Thanks! I actually noticed this before but yeah completely it slipped my mind

PR #1816, opened 2023-10-24
filecache: use rename for an atomic write

valpackett:
This prevents e.g. getting killed during writing from leaving incomplete/corrupt data in the cache directory.

### DIFF
--- a/internal/filecache/file_cache.go
+++ b/internal/filecache/file_cache.go
@@ -69,12 +69,28 @@ func (fc *fileCache) Add(key Key, content io.Reader) (err error) {
 	fc.mux.Lock()
 	defer fc.mux.Unlock()
 
-	file, err := os.Create(fc.path(key))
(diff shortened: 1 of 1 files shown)


internal/filecache/file_cache.go
```diff
@@ -69,12 +69,28 @@ func (fc *fileCache) Add(key Key, content io.Reader) (err error) {
 	fc.mux.Lock()
 	defer fc.mux.Unlock()
 
-	file, err := os.Create(fc.path(key))
+	// Use rename for an atomic write (on Unix, at least; Windows needs special direct syscall code)
```
ncruces, line 72: Since https://go.dev/cl/6140 I'd say `os.Rename` is "atomic enough" on Windows (using `MoveFileEx`), so I don't think we really need this comment.

ncruces (approved):

mathetake (approved):

PR #1824, opened 2023-10-31
wazevo: fuzz, fix f64x2.pmin

evacchi:
#1496

- BSL overwrites the condition register, fix is to reproduce the codegen for `OpcodeVbitselect`.
- I am also renaming the test case 1818 because the PR was actually #1820


internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -667,13 +667,13 @@ func Test1817(t *testing.T) {
 	})
 }
 
-// Test1817 tests that i16x8.narrow_i32x4_u assigns the dest register correctly.
-func Test1818(t *testing.T) {
+// Test1820 tests that i16x8.narrow_i32x4_u assigns the dest register correctly.
+func Test1820(t *testing.T) {
```
evacchi, line 671: rename the test case for #1820 (the doc comment was incorrect too)

internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -682,3 +682,22 @@ func Test1818(t *testing.T) {
...
+		require.NoError(t, err)
+		m := mod.(*wasm.ModuleInstance)
+		_, err = m.ExportedFunction("").Call(ctx)
+		require.NoError(t, err)
+		require.Equal(t, uint64(17282609607625994159), m.Globals[0].Val)
+		require.Equal(t, uint64(4671060543367625455), m.Globals[0].ValHi)
+	})
+}
+
+//1823
```
mathetake, line 703:
```suggestion
```

internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -682,3 +682,22 @@ func Test1818(t *testing.T) {
 		require.Equal(t, uint64(0xFFFF), m.Globals[1].ValHi)
 	})
 }
+
+// Test1820 tests that f64x2.pmin lowers to BSL with the right register usage
```
mathetake, line 686: 23?

PR #1845, opened 2023-11-23
wazevo: fuzz, fix regalloc ensure phi spills are consistent across siblings.

evacchi:
Fuzz case. Consider:

```mermaid
graph TD
    blk1 --> blk3
    blk2 --> blk3
```

where blk1 spills a given vReg, say `v183` like in this case, and blk2 does not. Then if blk3 allocates its registers starting from blk1, then it will expect v183 to be on the stack. 

We need to ensure that `blk2` correctly spills `v183` even if locally it wouldn't be necessary, so that it can be unspilled by `blk3` regardless if the flow originates in `blk1` or `blk2`.

This is an attempt to fix this, and it works in the given fuzz case without breaking any other tests, but it could be horribly misguided. 

Feel free to close the PR if the solution is horrifying. I suspect I am missing some bits.


internal/integration_test/vs/wasmtime/go.sum
evacchi, line 1: Huh I guess this was committed by mistake

internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -719,7 +720,7 @@ func Test1825(t *testing.T) {
 	})
 }
 
-// Test1825 tests that lowerFcopysignImpl allocates correctly the temporary registers.
+// Test1826 tests that lowerFcopysignImpl allocates correctly the temporary registers.
```
evacchi, line 723: This was incorrect

internal/integration_test/fuzzcases/fuzzcases_test.go
```diff
@@ -734,3 +735,20 @@ func Test1826(t *testing.T) {
 		require.Equal(t, uint64(0), m.Globals[0].ValHi)
 	})
 }
+
+// Test1845 tests regalloc ensures that when one predecessor Y1 for a block X
```
evacchi, line 739:
```suggestion
// Test1845 tests regalloc and ensures that when one predecessor Y1 for a block X
```

PR #1847, opened 2023-11-24
interpreter: fixes V128Mul for f32x4 lane

mathetake:


internal/engine/interpreter/interpreter.go
```diff
@@ -2973,6 +2973,7 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, m *wasm.ModuleInstance
 				retHi = x1hi * x2hi
 				retLo = x1lo * x2lo
 			case wazeroir.ShapeF32x4:
+				fmt.Println(mulFloat32bits(uint32(x1hi>>32), uint32(x2hi>>32)) << 32)
```
mathetake, line 2976: this is crazy why on earth this fmt.Println fixes the test????????

mathetake: seems like GOARCH=amd64 doesn't need this line (it passes the test without this print line), it seems highly likely that this is a bug of Go compiler on arm64

PR #1849, opened 2023-11-27
fuzz: ignores the stack detail during stack overflow

mathetake:


internal/integration_test/fuzz/wazerolib/nodiff.go
```diff
@@ -349,10 +349,13 @@ func ensureInvocationError(compilerErr, interpErr error) error {
 		interpErrMsg = interpErrMsg[:strings.Index(interpErrMsg, "\n")]
 	}
 
-	if strings.Contains(compilerErrMsg, "stack overflow") && strings.Contains(interpErrMsg, "unreachable") {
+	if compiledStackOverFlow := strings.Contains(compilerErrMsg, "stack overflow"); compiledStackOverFlow && strings.Contains(interpErrMsg, "unreachable") {
```
evacchi, line 352: fyi: the fuzzer found a pathological case, where there are 2 functions.

1. The first function "" returns unreachable + stack overflow
2. but if a second function "2" depends on the counter that caused "unreachable" earlier, now the interpreter and the compiler will return different values/errors. In my case, the interpreter did not fail, the compiler returned `unreachable`. This is fixed by bumping the stack ceiling further (it failed with ~91k, so bumping to 100k succeeded) see the shrunk case for reference https://gist.github.com/evacchi/cb54e1f1e0632d0b7a8d612502253d4e

mathetake: oh yeah I noticed that too, maybe I should've left comments somewhere. That's due to the stack space usage diff between old engines vs the new one and unavoidable stuff in my opinion. so I think as long as we share this knowledge, it should be fine for now. sounds good?

evacchi: Yeah, let's just keep it in the back of our heads for now :)

PR #1851, opened 2023-11-28
wazevo: initial work on constant folding

evacchi:
#1496 

This implements an initial, simple constant folding pass for add/sub/mul for i32/i64/f32/f64.
It also implements a new case for `passNopInstElimination` where one of the args is const 0.

- is it ok to mutate instructions in-place that way?
- do we need to add checks for integer underflow/overflow? the behavior right now is essentially whatever Go does (i.e. let integers wrap, let floats go to ±inf)
- todo: add a few more test cases

### DIFF
--- a/internal/engine/wazevo/ssa/pass.go
+++ b/internal/engine/wazevo/ssa/pass.go
@@ -361,8 +361,6 @@ func passNopInstElimination(b *builder) {
 }
 
 func ensureValueIdToInstructionInit(b *builder) {
-	b.valueIDToInstruction = b.valueIDToInstruction[:0]
-
 	if int(b.nextValueID) >= len(b.valueIDToInstruction) {
 		b.valueIDToInstruction = append(b.valueIDToInstruction, make([]*Instruction, b.nextValueID)...)
 	}
--- a/internal/engine/wazevo/ssa/pass.go
+++ b/internal/engine/wazevo/ssa/pass.go
@@ -20,6 +20,8 @@ func (b *builder) RunPasses() {
 	passRedundantPhiEliminationOpt(b)
 	// The result of passCalculateImmediateDominators will be used by various passes below.
 	passCalculateImmediateDominators(b)
+
+	passCollectValueIdToInstructionMapping(b)
 	passNopInstElimination(b)
 
 	// TODO: implement either conversion of irreducible CFG into reducible one, or irreducible CFG detection where we panic.
@@ -312,8 +314,6 @@ func (b *builder) clearBlkVisited() {
 
 // passNopInstElimination eliminates the instructions which is essentially a no-op.
 func passNopInstElimination(b *builder) {
-	ensureValueIdToInstructionInit(b)
-
 	for blk := b.blockIteratorBegin(); blk != nil; blk = b.blockIteratorNext() {
 		for cur := blk.rootInstr; cur != nil; cur = cur.next {
 			op := cur.Opcode()
@@ -361,7 +361,7 @@ func passNopInstElimination(b *builder) {
 	}
 }
 
-func ensureValueIdToInstructionInit(b *builder) {
+func passCollectValueIdToInstructionMapping(b *builder) {
 	if int(b.nextValueID) >= len(b.valueIDToInstruction) {
 		b.valueIDToInstruction = append(b.valueIDToInstruction, make([]*Instruction, b.nextValueID)...)
 	}
@@ -382,8 +382,6 @@ func ensureValueIdToInstructionInit(b *builder) {
 // passConstFoldingOpt scans all instructions for arithmetic operations over constants,
 // and replaces them with a const of their result.
 func passConstFoldingOpt(b *builder) {
-	ensureValueIdToInstructionInit(b)
-
 	isFixedPoint := false
 	for !isFixedPoint {
 		isFixedPoint = true
--- a/internal/engine/wazevo/ssa/pass.go
+++ b/internal/engine/wazevo/ssa/pass.go
@@ -399,14 +399,26 @@ func passConstFoldingOpt(b *builder) {
 						// Clear the references to operands.
 						cur.v, cur.v2 = ValueInvalid, ValueInvalid
 						// We assume all the types are consistent.
-						xc, yc := xDef.ConstantVal(), yDef.ConstantVal()
-						switch op {
-						case OpcodeIadd:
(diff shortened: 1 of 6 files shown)


internal/engine/wazevo/ssa/pass.go
```diff
@@ -309,26 +311,15 @@ func (b *builder) clearBlkVisited() {
 
 // passNopInstElimination eliminates the instructions which is essentially a no-op.
 func passNopInstElimination(b *builder) {
-	if int(b.nextValueID) >= len(b.valueIDToInstruction) {
-		b.valueIDToInstruction = append(b.valueIDToInstruction, make([]*Instruction, b.nextValueID)...)
-	}
+	ensureValueIdToInstructionInit(b)
```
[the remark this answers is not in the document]
evacchi: on second thought, it looks like it has to be reset each time (and/or we must ensure it is consistent across passes)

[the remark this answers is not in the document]
mathetake: not it shouldn't be necessary to do each time on each pass - should be fine with valueIDToInstruction once.

internal/engine/wazevo/ssa/pass.go
```diff
@@ -348,6 +339,88 @@ func passNopInstElimination(b *builder) {
...
+				if definingInst.Constant() && definingInst.ConstantVal() == 0 {
+					b.alias(cur.Return(), x)
+				}
+			}
+		}
+	}
+}
+
+func ensureValueIdToInstructionInit(b *builder) {
+	b.valueIDToInstruction = b.valueIDToInstruction[:0]
```
mathetake, line 364: as I said this initializes and iterates over instructions per invocation of this function.

internal/engine/wazevo/ssa/pass.go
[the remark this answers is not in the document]
evacchi: I wonder if it makes sense to refactor the remove the instructions in `passDeadCodeEliminationOpt` and invoke this instead.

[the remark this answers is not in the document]
evacchi: > on second thought, it looks like it has to be reset each time (and/or we must ensure it is consistent across passes)

> no it shouldn't be necessary to do each time on each pass

I got misled because passes are also invoked individually. So, I refactored this init procedure in its own pass ( `passCollectValueIdToInstructionMapping`) which is expected to execute before the dependant passes (nopInstElim and constFold for now).

internal/engine/wazevo/ssa/pass.go
evacchi: should be fixed now

internal/engine/wazevo/ssa/pass.go
```diff
@@ -348,6 +330,138 @@ func passNopInstElimination(b *builder) {
...
+						// We assume all the types are consistent.
+						if x.Type().Bits() == 64 {
+							xc, yc := int64(xDef.ConstantVal()), int64(yDef.ConstantVal())
+							switch op {
+							case OpcodeIadd:
+								cur.u1 = uint64(xc + yc)
+							case OpcodeIsub:
+								cur.u1 = uint64(xc - yc)
+							case OpcodeImul:
+								cur.u1 = uint64(xc * yc)
```
mathetake, line 410: I would like to see the overflow is handled correctly for 32-bits in a unit test

internal/engine/wazevo/ssa/pass.go
```diff
@@ -348,6 +330,138 @@ func passNopInstElimination(b *builder) {
...
+							}
+						} else {
+							xc, yc := int32(xDef.ConstantVal()), int32(yDef.ConstantVal())
+							switch op {
+							case OpcodeIadd:
+								cur.u1 = uint64(xc + yc)
+							case OpcodeIsub:
+								cur.u1 = uint64(xc - yc)
+							case OpcodeImul:
+								cur.u1 = uint64(xc * yc)
```
mathetake, line 420: I would like to see the overflow is handled correctly for 32-bits in a unit test

mathetake (commented):
I think this really needs fuzzing yeah

internal/engine/wazevo/ssa/pass.go
```diff
@@ -348,6 +330,138 @@ func passNopInstElimination(b *builder) {
...
+}
+
+// passConstFoldingOpt scans all instructions for arithmetic operations over constants,
+// and replaces them with a const of their result.
+func passConstFoldingOpt(b *builder) {
+	for blk := b.blockIteratorBegin(); blk != nil; blk = b.blockIteratorNext() {
+		for cur := blk.rootInstr; cur != nil; cur = cur.next {
+			// The fixed point is reached through a simple iteration over the list of instructions.
+			// Note: Instead of just an unbounded loop with a flag, we may also add an upper bound to the number of iterations.
+			isFixedPoint := false
```
mathetake, line 379: yeah I think I would suggest to see how much overhead costs will be added like by compiling zig stdlib binray and let's have a reasonable O(1) upper bound

internal/engine/wazevo/ssa/pass.go
evacchi: I started adding some cases https://github.com/tetratelabs/wazero/pull/1851/files#diff-96af5251a7ec1727e33ba8e3e622a79b77ba1c0753d099188b20841e230436caR507

internal/engine/wazevo/ssa/pass.go
evacchi: even without an upper bound the pass seems to account for less than 1% CPU time against the Wasm library.

Screenshot 2023-11-30 at 18 27 49

To be fair, that might not be surprising: e.g., if the wasm binary went through `wasm-opt` it might be already minimized, and if it weren't, operations between constants may not naturally occur in a Wasm binary very often (after all, it's already a compilation target)

evacchi: ...in fact, I am not even sure how often this pass will be useful, I don't know how often we'll find cases in the wild. Anyway it's a useful playground for me...

evacchi: yeah agreed (yeah that's size in bytes)

evacchi:
I pushed some new simple substitutions extending nop elimination. Algebraic simplification need more research to be effective.

evacchi:
closing for now as no significant benefit so far

Issue #1862, opened 2023-12-12
` fd_fdstat_set_flags` truncates file


yagehu:
**Describe the bug**
If a file is opened with [`oflags::trunc`](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-oflags-record), a call to [`fd_fdstat_set_flags`](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-fd_fdstat_set_flagsfd-fd-flags-fdflags---result-errno) will truncate the file.

**To Reproduce**

Compile this snippet with [`wasi-sdk`](https://github.com/WebAssembly/wasi-sdk) and run it:

```bash
clang wazero-write.c
mkdir -p tmp
wazero run -mount tmp a.out
stat tmp/a 
```

```c
#include <fcntl.h>
#include <string.h>
#include <stdio.h>
#include <unistd.h>

int main(void) {
  int fd = open("tmp/a", O_RDWR | O_CREAT | O_TRUNC);
  if (fd == -1) {
    perror("open");
    return 1;
  }

  char * buf = "abc";

  int written = write(fd, buf, 3);

  printf("written %d\n", written);

  int fcntl_ret = fcntl(fd, F_SETFL, 0);
  if (fcntl_ret != 0) {
    perror("fcntl");
    return 1;
  }

  return 0;
}
```

**Expected behavior**
The file should be size 3 because we wrote 3 bytes to it. But it is size 0.

**Environment (please complete the relevant information):**
- Go version: 1.21.4
- wazero Version: `5796897f37852bf2042cd63959ce4ff673c4366e`
- Host architecture: amd64
- Runtime mode: compiler

**Additional context**
Compiling the same snippet with native clang or gcc and running it will produce a 3-byte file as expected.

evacchi:
confirmed that even with other runtimes the byte size is still 3 after stat

PR #1863, opened 2023-12-12
Fix `fd_fdstat_set_flags` truncating with O_TRUNC

yagehu:
This commit fixes a buf when `fd_fdstat_set_flags` would erroneously truncate an open file if it is opened with the O_TRUNC flag.  This happens because the O_TRUNC flag causes the file to be truncated when it is opened.

This fixes https://github.com/tetratelabs/wazero/issues/1862

### DIFF
--- a/internal/sysfs/osfile.go
+++ b/internal/sysfs/osfile.go
@@ -83,8 +83,8 @@ func (f *osFile) SetAppend(enable bool) (errno experimentalsys.Errno) {
 		f.flag &= ^experimentalsys.O_APPEND
 	}
 
-	// Clear any create flag, as we are re-opening, not re-creating.
-	f.flag &= ^experimentalsys.O_CREAT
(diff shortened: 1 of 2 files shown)


internal/sysfs/osfile.go
```diff
@@ -86,6 +86,9 @@ func (f *osFile) SetAppend(enable bool) (errno experimentalsys.Errno) {
 	// Clear any create flag, as we are re-opening, not re-creating.
 	f.flag &= ^experimentalsys.O_CREAT
 
+	// Clear trunc flag, as we are re-opening.
+	f.flag &= ^experimentalsys.O_TRUNC
```
achille-roussel, line 90: Since this comment repeats what's above, how about merging the two?
```suggestion
	f.flag &= ^(experimentalsys.O_CREAT | experimental.O_TRUNC)
```

evacchi (approved):
I think this looks good, thanks for your contribution!

Issue #1865, opened 2023-12-12
[WASI] `fd_fdstat_set_flags` with `fdflags::append` resets fd offset


yagehu:
**Describe the bug**
Calling `fd_fdstat_set_flags` with `fdflags::append` resets the fd offset such that calling `fd_tell` will always return 0.

**To Reproduce**

```bash
clang wazero-tell.c
mkdir -p tmp
wazero run -mount tmp a.out
```

```c
#include <fcntl.h>
#include <string.h>
#include <stdio.h>
#include <unistd.h>

int main(void) {
  int fd = open("tmp/a", O_CREAT | O_RDWR);
  if (fd == -1) {
    perror("open");
    return 1;
  }

  char * buf = "abc";

  int written = write(fd, buf, 3);
  printf("written %d bytes\n", written);

  int ret = fcntl(fd, F_SETFL, O_APPEND);
  if (ret == -1) {
    perror("fcntl");
    return 1;
  }

  off_t pos = lseek(fd, 0, SEEK_CUR);
  printf("offset %lld\n", pos);
}
```

**Expected behavior**
This should print `offset 3` since the snippet writes 3 bytes.  This is the case with other runtimes (Wasmtime, WasmEdge, WAMR) and native.  However, Wazero prints `offset 0`.

**Environment (please complete the relevant information):**
- Go version: 1.21.4
- wazero Version: 99ededcefd564ff52255fd25cedb014c83860f23
- Host architecture: amd64
- Runtime mode: compiler

mathetake:
by #1866

PR #1866, opened 2023-12-12
wasi: resets offsets in reopen

yagehu:
This commit fixes `fd_fdstat_set_flags` resetting the fd offset due to it reopening the file every time.  This fix gets the current offset before reopening and restore the offset with seek. This addresses https://github.com/tetratelabs/wazero/issues/1865.

NB: A couple of existing tests are actually spec'd wrong. To fix, I introduced addition `Seek` calls before checking removing append flag.

### DIFF
--- a/internal/sysfs/osfile.go
+++ b/internal/sysfs/osfile.go
@@ -98,9 +98,38 @@ func (f *osFile) reopen() (errno experimentalsys.Errno) {
 	// Clear any create flag, as we are re-opening, not re-creating.
 	f.flag &= ^experimentalsys.O_CREAT
 
+	var (
+		isDir  bool
+		offset int64
+		err    error
+	)
+
+	isDir, errno = f.IsDir()
+	if errno != 0 {
+		return errno
+	}
+
+	if !isDir {
+		offset, err = f.file.Seek(0, 1)
+		if err != nil {
+			return experimentalsys.UnwrapOSError(err)
+		}
+	}
+
 	_ = f.close()
 	f.file, errno = OpenFile(f.path, f.flag, f.perm)
-	return
+	if errno != 0 {
+		return errno
+	}
+
+	if !isDir {
+		_, err = f.file.Seek(offset, 0)
+		if err != nil {
+			return experimentalsys.UnwrapOSError(err)
+		}
+	}
+
+	return 0
 }
 
 // IsNonblock implements the same method as documented on fsapi.File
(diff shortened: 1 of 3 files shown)


evacchi (commented):
as you can see, unfortunately this is failing on Windows

yagehu:
> as you can see, unfortunately this is failing on Windows

Fixed!

internal/sysfs/osfile.go
```diff
@@ -98,9 +98,32 @@ func (f *osFile) reopen() (errno experimentalsys.Errno) {
 	// Clear any create flag, as we are re-opening, not re-creating.
 	f.flag &= ^experimentalsys.O_CREAT
 
+	var (
+		offset int64
+		err    error
+	)
```
evacchi, line 104: I don't think we normally use this style for inline declarations; besides err is only used in the nested scope, so maybe you can declare `var offset int64` here...

internal/sysfs/osfile.go
```diff
@@ -98,9 +98,32 @@ func (f *osFile) reopen() (errno experimentalsys.Errno) {
...
+
+	if isDir, errno := f.IsDir(); errno == 0 && !isDir {
+		offset, err = f.file.Seek(0, 1)
```
evacchi, line 107: ...declare error here...
```suggestion
		var err error
		offset, err = f.file.Seek(0, 1)
```

internal/sysfs/osfile.go
```diff
@@ -98,9 +98,32 @@ func (f *osFile) reopen() (errno experimentalsys.Errno) {
...
+
 	_ = f.close()
 	f.file, errno = OpenFile(f.path, f.flag, f.perm)
-	return
+	if errno != 0 {
+		return errno
+	}
+
+	if isDir, errno := f.IsDir(); errno == 0 && !isDir {
+		_, err = f.file.Seek(offset, 0)
```
evacchi, line 120:
```suggestion
		_, err := f.file.Seek(offset, 0)
```

use the short form here

evacchi (approved):

evacchi:
thanks!

evacchi:
I see you force-pushed new changes, let me know when the PR is in a stable state @yagehu ! :)

yagehu:
> I see you force-pushed new changes, let me know when the PR is in a stable state @yagehu ! :)

I'm done. Just a quick fix to avoid double `IsDir` calls. Thanks!

evacchi (approved):

internal/sysfs/osfile.go
```diff
@@ -98,9 +98,38 @@ func (f *osFile) reopen() (errno experimentalsys.Errno) {
...
+		err    error
+	)
+
+	isDir, errno = f.IsDir()
+	if errno != 0 {
+		return errno
+	}
+
+	if !isDir {
+		offset, err = f.file.Seek(0, 1)
```
ncruces, line 113: Maybe too late for a nit, but please constants for the whence: `io.SeekStart`, etc…

evacchi: Good point, let's fix it in a follow up PR

PR #1869, opened 2023-12-13
Introduces the new optimizing compiler (wazevo) in experimental pkg

### LINKED ISSUE #1496: Implement initial optimizing compiler backend

mathetake:
This finally closes #1496 which aims to introduce an optimizing compiler in wazero as 
an alternative backend for the current single pass compiler.
Even though this is released under experimental, it already has been well-tested with
long-running fuzzing, Zig/TinyGo standard library tests etc.
It's only available yet on arm64, but we will soon start working on amd64 as well led by @evacchi.

The optimizing compiler is written entirely from scratch, and the initial result is pretty amazing.
The rough numbers on the performance aspect is like **2x faster** on compilation and
**3x faster** on runtime compared to the existing single passs compiler when running 
Zig's standard library tests which is a huge Wasm binary with all tests.
This is a huge win given that this new back end, even though it is "optimizing", but the implementation 
is pretty naive currently, and there's huge room for improvements (we welcome contributions!).

Once amd64 is implemented and it's been adopted by communities, then this will be the default
compiler and we will delete the existing single pass is the plan for now.

You can enable the optimizing compiler by replacing `wazero.NewRuntimeConfigCompiler()` or `wazero.NewRuntimeConfig()` with the new API like
```diff
-       c := wazero.NewRuntimeConfigCompiler()
+       c := opt.NewRuntimeConfigOptimizingCompiler()
-       c := wazero.NewRuntimeConfig()
+       c := opt.NewRuntimeConfigOptimizingCompiler()
```

In CLI, you can use `-optimizing-compiler` flag to enable it:
```
wazero run -optimizing-compiler myapp.wasm
```

Thank you for all your support!


ncruces (approved):
Ran tests on QEMU and real hardware, all green!

Issue #1870, opened 2023-12-13
`fd_filestat_set_size` wraps instead of error on large file size


yagehu:
**Describe the bug**
Calling [`fd_filestat_set_size`](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-fd_filestat_set_sizefd-fd-size-filesize---result-errno) with a very large size results in Wazero returning `0` (success) and setting the file size to a different value.  It's probably wrapping or truncating the number.

**To Reproduce**

```bash
clang wazero.c
mkdir -p tmp
wazero run -mount tmp a.out
```

```c
#include <fcntl.h>
#include <string.h>
#include <stdio.h>
#include <unistd.h>

int main(void) {
  int fd = open("tmp/a", O_CREAT | O_RDWR);
  if (fd == -1) {
    perror("open");
    return 1;
  }

  int ret = ftruncate(fd, 2815222840989488694);  // Very large size
  if (ret == -1) {
    perror("ftruncate");
    return 1;
  }

  return 0;
}
```

**Expected behavior**

The snippet should error with "File too large". Instead, Wazero successfully sets the file to a size of 415143478 on my machine.

**Environment (please complete the relevant information):**
- Go version: 1.21.4
- wazero Version: f46d5b943da5337ae1ebf6d102f352f52858d8ec
- Host architecture: amd64
- Runtime mode: compiler

PR #1871, opened 2023-12-13
wasi: Handle EFBIG from OS on`fd_filestat_set_size`

yagehu:
This commit handles potential `EFBIG` returned from the OS and maps it to WASI's `fbig`.  It also fixes an integer truncation bug in the `fd_filestat_set_size` handler.

fixes https://github.com/tetratelabs/wazero/issues/1870


evacchi (approved):
Makes sense; `off_t` in POSIX on 64-bit systems seems to be usually an alias to a signed int 64

mathetake:
cool

Issue #1891, opened 2023-12-26
Feature Request: Rust tokio async interoperability through pollOneOff


gaukas:
Merry Christmas and Happy Holidays! 🎄

**Is your feature request related to a problem? Please describe.**
I have a WebAssembly binary generated by compiling from Rust, in which async I/O operations are performed using tokio. 

However, the WebAssembly panics when performing async read operations: 

```
thread '<unnamed>' panicked at /path/to/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.35.1/src/runtime/io/driver.rs:157:23:
unexpected error when polling the I/O driver: Os { code: 58, kind: Uncategorized, message: "Not supported" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Ref: [tokio-1.35.1/src/runtime/io/driver.rs:157](https://github.com/tokio-rs/tokio/blob/46ff36386d11e52faae038d5afe8d2f7a39dfe39/tokio/src/runtime/io/driver.rs#L149-L158)

I noticed this error was actually returned by `mio::Poll::poll()`([Docs](https://docs.rs/mio/latest/mio/struct.Poll.html#method.poll)), which calls to `Selector::select()`([Source](https://docs.rs/mio/latest/src/mio/sys/unix/selector/kqueue.rs.html#93-121)). So it is likely the invoked `syscall(kevent)` has triggered the OS error. 

**Describe the solution you'd like**
Since the given operation can be executed perfectly fine on at least one other popular WebAssembly runtime (wasmtime) with rest of them untested, the best option is seemingly adding the related syscall support in full in order to support complete async/nonblocking operations. 

**Describe alternatives you've considered**
Or, it could be due to a mistake on my end for non-standard use of wazero (see Additional Context). I would be rather happy to see an existing example that works with `tokio::select` and `tokio::io::AsyncReadExt::read()`.

**Additional context**
So I am indeed doing something weird (#1857). I patched several interfaces in wazero to allow me insert `net.TCPConn` into WebAssembly with access through a file descriptor. 

<details>
<summary>(*Context).InsertTCPConn</summary>
<br>

```
func (c *Context) InsertTCPConn(conn *net.TCPConn) (key int32, ok bool) {
	return c.fsc.openedFiles.Insert(&FileEntry{
		IsPreopen: true,
		File:      fsapi.Adapt(sysfs.NewTCPConnFile(conn)),
	})
}
```

</details>

[...]

PR #1892, opened 2023-12-28
wasi: sockets, use Conn.Control on Unix, too; align to Windows

evacchi:
The current implementation of WASI sockets differs between Unix and Windows. On the Unix version, we save underlying FD to the struct, while on Windows we keep the Listener/Connection as a field, and invoke `Conn.Control(func(fd))`.

The reason for the difference on Windows, is that Windows won't return a valid file handle, because the Dup implementation is a no-op. However, further somewhat related testing (see #1891) highlighted that the Unix strategy does not really work: the FD that is being saved, is not really the underlying FD of the Listener/Connection struct, but rather the FD of a duplicated File. 

Because the duplicate is not being saved, but only the FD, it *looks like* eventually the new File instance is collected, invalidating the underlying FD. This is only visible in long-running tests, where we probably give a chance to the runtime to run a GC cycle.

The implementation on `main` is incorrect anyway, so in this PR I am aligning the implementation strategy to the Windows version: we keep the Conn/Listener around in our own File struct, and only access the FD through `Conn.Control`. This in turn allows to share a lot more of the original code, further cleaning up the implementation.

### DIFF
--- a/internal/sysfs/file_windows.go
+++ b/internal/sysfs/file_windows.go
@@ -46,10 +46,10 @@ func writeFd(fd uintptr, buf []byte) (int, sys.Errno) {
 	return -1, sys.ENOSYS
 }
 
-func readSocket(h syscall.Handle, buf []byte) (int, sys.Errno) {
+func readSocket(h uintptr, buf []byte) (int, sys.Errno) {
 	var overlapped syscall.Overlapped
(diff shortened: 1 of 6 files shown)


gaukas:
Thank you, it works perfectly.

PR #1896, opened 2024-01-05
threads: add spectests for Wasm threads proposal

anuraaga:
@mathetake Trying to keep PRs small but let me know if this is too small :)


mathetake (approved):
Good! 🎉

PR #1898, opened 2024-01-05
threads: add function validation of atomics

anuraaga:


mathetake (approved):
I slightly remember that memory must have min==max, but where is it validated?

anuraaga:
@mathetake Hmm, don't think I've seen that anywhere. For example in libpgquery wasm, memory looks like this

`(memory (;0;) 15 65536 shared)`

mathetake:
aha ok

mathetake:
then what happens when growing memory (changing the base address would be dangerous I guess when multiple threads are executing load/store?)

anuraaga:
@mathetake Yup you're right. We discussed a bit but basically the only way to support it, without allocating tons of memory up front, is to use mmap. I think I need to clean up the implementation a bit before sending the PR for here, but currently it looks like this

https://github.com/wasilibs/wazerox/blob/main/internal/wasm/memory.go#L74

Will it be ok?

mathetake:
yeah looks good provided the behavior is well-documented on the API to toggle threads instructions. IIRC mmap itself doesn't immediately allocate physical pages immediately so it makes sense

ncruces:
I wonder, do tests pass on 32-bit platforms, with mmaping 65536 pages (4GB)?

PR #1903, opened 2024-01-07
wasi: sockets, use WSAPoll and GetOverlappedResult on Windows

gaukas:
Current `readSocket` function for WASI sockets is implemented incorrectly on Windows and WASI polling an async socket for a long time will eventually get `ERROR_WORKING_SET_QUOTA` (1453):

```
Insufficient quota to complete the requested service.
```

A call to `GetOverlappedResult` is expected after each `ERROR_IO_PENDING` returned by `ReadFile` and that is what's missing in the current implementation. Besides that, `ReadFile` would create a background worker to keep on trying reading from the handle into the `buf`, which could result in data loss if we don't keep tracking of it. So we add a call to `WSAPoll` before calling `ReadFile` to observe how many bytes are ready to be read. 

In this pull request, we modify how `readSocket` in `internal/sysfs/file_windows.go` is implemented by adding the following changes: 
- [x] Call `WSAPoll` before calling `ReadFile`, abort if ~no bytes are available~ no fd is ready to be read. 
- [x] Call `GetOverlappedResult` if `ReadFile` returns `ERROR_IO_PENDING`. 
- [x] ~Limit the read buffer's size according to how many bytes are ready to be read (see review comments)~ 
- [x] Cancel the ongoing overlapped read if it cannot be completed immediately.

### DIFF
--- a/internal/sysfs/file_windows.go
+++ b/internal/sysfs/file_windows.go
@@ -81,7 +81,6 @@ func readSocket(h uintptr, buf []byte) (int, sys.Errno) {
 	//
 	// We are currently skipping checking if hFile was opened with FILE_FLAG_OVERLAPPED but using
 	// both lpOverlapped and lpNumberOfBytesRead.
-
 	var overlapped syscall.Overlapped
 
 	// Create an event to wait on.
@@ -98,12 +97,6 @@ func readSocket(h uintptr, buf []byte) (int, sys.Errno) {
 		if errno != nil {
 			return 0, sys.UnwrapOSError(errno) // This is a fatal error. CancelIo failed.
 		}
-		// // Temp Fix: This is to make sure that I/O is cancelled before we wait for it.
-		// // Otherwise, we may still see ERROR_IO_PENDING.
-		// //
-		// // Disabled: This is not a good fix since it does not address the root cause.
-		// // And the alleged bug cannot be reproduced in any test.
-		// time.Sleep(1 * time.Nanosecond)
 
 		done, errno = getOverlappedResult(syscall.Handle(h), &overlapped, true) // wait for I/O to complete(cancel or finish). Overwrite done and errno.
 		if errors.Is(errno, syscall.ERROR_OPERATION_ABORTED) {
--- a/internal/sysfs/file_windows.go
+++ b/internal/sysfs/file_windows.go
@@ -3,6 +3,7 @@ package sysfs
 import (
 	"errors"
 	"syscall"
+	"time"
 	"unsafe"
 
 	"github.com/tetratelabs/wazero/experimental/sys"
@@ -23,6 +24,8 @@ var (
 	procPeekNamedPipe = kernel32.NewProc("PeekNamedPipe")
 	// procGetOverlappedResult is the syscall.LazyProc in kernel32 for GetOverlappedResult
 	procGetOverlappedResult = kernel32.NewProc("GetOverlappedResult")
+	// procCreateEventW is the syscall.LazyProc in kernel32 for CreateEventW
+	procCreateEventW = kernel32.NewProc("CreateEventW")
 )
 
 // readFd returns ENOSYS on unsupported platforms.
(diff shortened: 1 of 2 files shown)


internal/sysfs/file_windows.go
```diff
@@ -47,20 +55,49 @@ func writeFd(fd uintptr, buf []byte) (int, sys.Errno) {
...
+	//  function to get the actual number of bytes read. If the hFile parameter is associated
+	//  with an I/O completion port, you can also get the number of bytes read by calling the
+	//  GetQueuedCompletionStatus function.
+	//
+	// We are currently skipping checking if hFile was opened with FILE_FLAG_OVERLAPPED but using
+	// both lpOverlapped and lpNumberOfBytesRead.
+
 	var overlapped syscall.Overlapped
 	var done uint32
 	errno := syscall.ReadFile(syscall.Handle(h), buf, &done, &overlapped)
```
gaukas, line 85: Not sure if it is good to use the original input `buf` here, or better using a temporary buffer with the size up to `n` returned by `wsaPoll`. Considering it is possible for the buffer to be in use even after `syscall.ReadFile` and `getOverlappedResult` returns and by the next time caller calls this function we already lost the information in `overlapped`.

Any thoughts? @evacchi

evacchi: Actually n just returns the number of FDS that are ready, so it must be >=0. Instead of readfile i think you can just use syscall.Read and we avoid the complexity of overlapped i/o altogether because we will do a read only when there's some data ready

gaukas: `syscall.Read` very unfortunately returns an error. I think it has something to do with internally calling `syscall.ReadFile` with `overlapped` being `nil`.

gaukas: But I see what you meant there. So since `wsaPoll` does not really tell you how many bytes are ready, I am good with the current implementation.

internal/sysfs/file_windows.go
```diff
@@ -47,20 +55,49 @@ func writeFd(fd uintptr, buf []byte) (int, sys.Errno) {
...
-	if errno == syscall.ERROR_IO_PENDING {
-		errno = sys.EAGAIN
+	if errors.Is(errno, syscall.ERROR_IO_PENDING) {
+		nBytesTransferred, errno := getOverlappedResult(syscall.Handle(h), &overlapped, false)
```
evacchi, line 87: I wonder if this will ever happen, and what to do in this case. If we are at this point, `poll` has returned n >= 0 and `ReadFile` hopefully won't hang.

As we discovered together offline, ERROR_IO_PENDING means that ReadFile is holding a reference to `buf` and will complete the write to it once `getOverlappedResult()` (or related APIs such as `WaitForSingleObject`, `WaitForMultipleObjects`) returns success.

the `false` in `getOverlappedResult(syscall.Handle(h), &overlapped, false)` means that it will return immediately, with `true` meaning it will block until done. But if `false` and the I/O is still pending, we cannot really return, because `buf` could be still in an invalid state.

So, either we use `true` or we just remove `getOverlappedResult()` altogether. I am tempted to suggest the latter.

PS now I am also wondering if we need to really set the nonblocking state on the socket since we are simulating it anyway, but that's maybe for another time.

evacchi (changes requested):
I would add a test case.

You can modify https://github.com/tetratelabs/wazero/blob/009ee7056c1f7afb597781851cc8b0f660c5bbc5/internal/sysfs/sock_test.go#L50 to ensure that it also checks the conditions that caused this PR in the first place (You might need to modify the WASI test case)

You can also add a unit test to `readSocket` (add a test to `file_test.go`, there is none currently.)

You can do either or both. Thanks!

internal/sysfs/file_windows.go
gaukas: What happens if two concurrent Reads on the same fed happened at once? I assume after both get pass wsaPoll, the slower one will start an asynchronous read... So I'm skeptical about dropping `getOverlappedResult`.

evacchi: Good point, either way if we call with false you are still going to potentially leave the buffer in an invalid state (maybe we should create a new buffer and then copy over if no error is returned? But then we're back to requiring a table of buffers) with true we will block. Maybe, assuming that concurrent access is rare, blocking is the simplest option, and we can improve in the future if necessary

gaukas: I agree, blocking might be a more proper behavior. I will also check if there's a proper way to cancel if it became async read, which might align with Unix's async reading better?

evacchi: ok

internal/sysfs/file_windows.go
```diff
@@ -47,20 +57,68 @@ func writeFd(fd uintptr, buf []byte) (int, sys.Errno) {
...
+		errno = syscall.CancelIo(syscall.Handle(h))
+		if errno != nil {
+			return 0, sys.UnwrapOSError(errno) // This is a fatal error. CancelIo failed.
+		}
+		// // Temp Fix: This is to make sure that I/O is cancelled before we wait for it.
+		// // Otherwise, we may still see ERROR_IO_PENDING.
+		// //
+		// // Disabled: This is not a good fix since it does not address the root cause.
+		// // And the alleged bug cannot be reproduced in any test.
+		// time.Sleep(1 * time.Nanosecond)
```
evacchi, line 106: to be removed?

internal/sysfs/file_windows.go
```diff
@@ -47,20 +57,68 @@ func writeFd(fd uintptr, buf []byte) (int, sys.Errno) {
...
+	//  - The lpOverlapped parameter must point to a valid and unique OVERLAPPED structure,
+	//  otherwise the function can incorrectly report that the read operation is complete.
+	//  - The lpNumberOfBytesRead parameter should be set to NULL. Use the GetOverlappedResult
+	//  function to get the actual number of bytes read. If the hFile parameter is associated
+	//  with an I/O completion port, you can also get the number of bytes read by calling the
+	//  GetQueuedCompletionStatus function.
+	//
+	// We are currently skipping checking if hFile was opened with FILE_FLAG_OVERLAPPED but using
+	// both lpOverlapped and lpNumberOfBytesRead.
+
```
evacchi, line 84: nit, redundant whitespace

evacchi (changes requested):
clean up the commented out bits, if the tests pass then it should be good to go!

internal/sysfs/file_windows.go
gaukas: I first thought to keep this as an internal bug documentation, but since we could not find a reproduce I figured it is better to just remove it! 😄

evacchi (approved):

internal/sysfs/sock_test.go
```diff
@@ -83,6 +85,51 @@ func TestTcpConnFile_Read(t *testing.T) {
...
+
+	// Test #2: Read from a TCP connection asynchronously (i.e., with
+	// the non-blocking flag set explicitly).
+	tcpAddr2, err := net.ResolveTCPAddr("tcp", listen.Addr().String())
+	require.NoError(t, err)
+	tcp2, err := net.DialTCP("tcp", nil, tcpAddr2)
+	require.NoError(t, err)
+	defer tcp.Close() //nolint
+
+	// use a goroutine to asynchronously write to the TCP connection
```
mathetake, line 97:
```suggestion
	// Use a goroutine to asynchronously write to the TCP connection
```

PR #1915, opened 2024-01-12
threads: implement threads proposal in interpreter

anuraaga:


internal/integration_test/spectest/spectest.go
```diff
@@ -336,7 +340,7 @@ func Run(t *testing.T, testDataFS embed.FS, ctx context.Context, config wazero.R
 
 	// If the go:embed path resolution was wrong, this fails.
 	// https://github.com/tetratelabs/wazero/issues/247
-	require.True(t, len(caseNames) > 1, "len(caseNames)=%d (not greater than one)", len(caseNames))
+	require.True(t, len(caseNames) > 0, "len(caseNames)=%d (not greater than zero)", len(caseNames))
```
anuraaga, line 343: There is only one JSON file in threads/testdata

internal/integration_test/spectest/threads/spec_test.go
```diff
@@ -26,6 +26,5 @@ func TestCompiler(t *testing.T) {
 }
 
 func TestInterpreter(t *testing.T) {
-	t.Skip() // TODO: Delete after implementing interpreter support
 	spectest.Run(t, testcases, context.Background(), wazero.NewRuntimeConfigInterpreter().WithCoreFeatures(enabledFeatures))
```
anuraaga, line 29: Notice this is passing but I haven't added the threads feature flag yet, since that is public API. It means after this PR threads code technically is runnable without setting a flag.

Let me know if I should add the flag in this PR or a followup, it will touch a few files so I think a followup would be better if that temporary situation is ok

mathetake: let's do the flag addition after compiler is done!

internal/integration_test/spectest/spectest.go
mathetake: haha ok

mathetake:
@achille-roussel @ncruces could you take a look?

internal/wasm/memory.go
```diff
@@ -279,3 +297,109 @@ func (m *MemoryInstance) writeUint64Le(offset uint32, v uint64) bool {
...
+
+// Wait32 suspends the caller until the offset is notified by a different agent.
+func (m *MemoryInstance) Wait32(offset uint32, exp uint32, timeout int64) uint64 {
+	w := m.getWaiters(offset)
+	w.mux.Lock()
+
+	addr := unsafe.Add(unsafe.Pointer(&m.Buffer[0]), offset)
+	// Must be atomic even though we have the lock since it can be modified
+	// concurrently by a non-wait instruction.
+	cur := atomic.LoadUint32((*uint32)(addr))
```
ncruces, line 309: Is your expectation that unlocking `MemoryInstance.Mux` makes those changes visible here?

internal/wasm/memory.go
```diff
@@ -39,6 +49,14 @@ type MemoryInstance struct {
 	Shared        bool
 	// definition is known at compile time.
 	definition api.MemoryDefinition
+
+	// Mux is used in interpreter mode to prevent overlapping calls to atomic instructions,
+	// introduced with WebAssembly threads proposal.
+	Mux sync.RWMutex
```
ncruces, line 55: You seem to be using this as a `sync.Mutex` only, which I don't disagree with (unless it's *proven* that we have many concurrent readers, few writers *and* this helps, a plain `Mutex` is probably cheaper).

If this becomes a bottleneck, intuition tells me having (e.g.) `[4]sync.Mutex` and using (e.g.) bits 5 and 6 of the address to index that array would be better than a `RWMutex`.

ncruces (approved):
I think I brought it this up before, and I don't know how to do better, so… 👍

But I think it should be documented somehow that operations (particularly `OperationKindAtomicRMW`) are atomic between WASM guests, but not between guest and host.

anuraaga:
Thanks for remembering @ncruces - yeah that doc will be on the feature flag that will be added after implementing compiler

internal/wasm/memory.go
anuraaga: Ah thanks, for now I'll switch to normal mutex, I just resurrected this which used to be used with Grow without noticing

internal/wasm/memory.go
anuraaga: I see what you mean let me fix

ncruces: I'm actually not sure; as since we're assuming for fence that a mutex implies a barrier/fence, maybe this works (under the same assumption).

If that's the rational, I'd document it, though (just as you did on fence).

anuraaga: Thanks, I have updated the comment. I do think mux unlock will make changes visible to future atomic reads. I did think that perhaps there is still an issue that we won't read the value from an atomic write whose mux has not been unlocked yet - does that seem OK? Otherwise we would need to acquire the lock in interpreter mode, and unlock after reading this value, before potentially blocking. It makes the code more hairy but I wonder if it's required, what do you think?

ncruces: I need a functioning brain to answer that.
Let me try again tomorrow.

anuraaga: Thanks @ncruces - I went ahead and pushed a commit too that while a bit annoying, isn't so bad, and it makes sure cur value is read in interpreter in the same way as other memory operations. We can revert if it is seems unnecessary

anuraaga: BTW this may be useful reading

https://github.com/WebAssembly/threads/issues/211

Other alternatives could be simpler code if the current value read didn't need to be under the waiters lock

ncruces: LGTM

ncruces:
> Thanks for remembering @ncruces - yeah that doc will be on the feature flag that will be added after implementing compiler

I wonder if the answer here is to provide (a subset of?) atomic access methods in the memory API.

Can a memory be shared between an interpreter and a compiled instance?

Anyway, this is out-of-scope here.

anuraaga:
@ncruces Thanks for the review! @mathetake just in case it's hard to follow, we got @ncruces's LGTM.

> I wonder if the answer here is to provide (a subset of?) atomic access methods in the memory API.

I think so, perhaps it could be a separate type `AtomicMemory` or something

> Can a memory be shared between an interpreter and a compiled instance?

Not sure I understand the question but because compiler or interpreter is a `Runtime` setting, I don't think so, all modules created by that runtime will be the same including memory.

mathetake (approved):

PR #1916, opened 2024-01-12
wazevo(amd64): encode unaryRmR, not, neg, mulHi, shiftR

evacchi:
- arm64: encode unaryRmR
- amd64: encode not,neg
- amd64: introduce encodeEncEnc, rexInfo.encode() accepts arbitrary uint8
- amd64: mulHi, fix formatting for amode + _64

some refactoring as well. E.g. introduce encodeEncEnc and encodeEncMem, which are generalized versions of, respectively encodeRegReg and encodeRegMem, where instead of a `regEnc` a `uint8` is accepted. 

Notice that in some cases it is still necessary to treat the uint8 as registers; instead of arbitrarily casting I have added accessors `regEncoding(uint8)` and `regRexBit(uint8)`; open to suggestions if this makes no sense / alternative approaches are preferred.

### DIFF
--- a/internal/engine/wazevo/backend/isa/amd64/instr_encoding.go
+++ b/internal/engine/wazevo/backend/isa/amd64/instr_encoding.go
@@ -493,21 +493,28 @@ func (i *instruction) encode(c backend.Compiler) (needsLabelResolution bool) {
 
 	case not:
 		var prefix legacyPrefixes
-		var opcode uint32
-
-		src, dst := regEncodings[i.op1.r.RealReg()], regEncodings[i.op2.r.RealReg()]
-
+		src := regEncodings[i.op1.r.RealReg()]
 		rex := rexInfo(0)
 		if i.b1 { // 64 bit.
 			rex = rexInfo(0).setW()
 		} else {
 			rex = rexInfo(0).clearW()
 		}
-
-		encodeRegReg(c, prefix, opcode, 1, src, dst, rex)
+		subopcode := regEnc(2)
+		encodeRegReg(c, prefix, 0xf7, 1, subopcode, src, rex)
 
 	case neg:
-		panic("TODO")
+		var prefix legacyPrefixes
+		src := regEncodings[i.op1.r.RealReg()]
+		rex := rexInfo(0)
+		if i.b1 { // 64 bit.
+			rex = rexInfo(0).setW()
+		} else {
+			rex = rexInfo(0).clearW()
+		}
+		subopcode := regEnc(3)
+		encodeRegReg(c, prefix, 0xf7, 1, subopcode, src, rex)
+
 	case div:
 		panic("TODO")
 	case mulHi:
--- a/internal/engine/wazevo/backend/isa/amd64/instr_encoding.go
+++ b/internal/engine/wazevo/backend/isa/amd64/instr_encoding.go
@@ -518,7 +518,36 @@ func (i *instruction) encode(c backend.Compiler) (needsLabelResolution bool) {
 	case div:
 		panic("TODO")
 	case mulHi:
-		panic("TODO")
+		var prefix legacyPrefixes
+		rex := rexInfo(0)
+		if i.b1 { // 64 bit.
+			rex = rexInfo(0).setW()
+		} else {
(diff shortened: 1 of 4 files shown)


internal/engine/wazevo/backend/isa/amd64/operands.go
```diff
@@ -120,22 +120,22 @@ func newAmodeRipRelative(label backend.Label) amode {
 }
 
 // String implements fmt.Stringer.
-func (a *amode) String() string {
+func (a *amode) format(_64 bool) string {
```
mathetake, line 123: why do you need _64 flag for addressing mode? shouldn't it be always 64-bit address on x64?

evacchi: does `mull 123(%edi)` make any sense at all?

internal/engine/wazevo/backend/isa/amd64/instr_encoding_test.go
```diff
@@ -100,6 +100,126 @@ func TestInstruction_format_encode(t *testing.T) {
...
+		},
+		{
+			setup:      func(i *instruction) { i.asMulHi(newOperandMem(newAmodeImmReg(123, rdiVReg)), true, false) },
+			want:       "f76f7b",
+			wantFormat: "imull 123(%edi)",
+		},
+		{
+			setup:      func(i *instruction) { i.asMulHi(newOperandMem(newAmodeImmReg(123, rdiVReg)), false, false) },
+			want:       "f7677b",
+			wantFormat: "mull 123(%edi)",
```
mathetake, line 151: this is weird as I said the addressing mode should be 64-bit address

evacchi:
(commit message should have read "operandKindMem && !_64 makes no sense")

internal/engine/wazevo/backend/isa/amd64/instr_encoding.go
```diff
@@ -463,15 +463,102 @@ func (i *instruction) encode(c backend.Compiler) {
...
+		var subopcode uint8
+		if signed {
+			subopcode = 5
+		} else {
+			subopcode = 4
+		}
+
+		// src1 is implicitly rax,
+		// dst_lo is implicitly rax,
+		// dst_hi is implicitly rdx.
```
evacchi, line 550: it could also make sense to make these explicitly if it simplifies tracking allocated registers later

internal/engine/wazevo/backend/isa/amd64/operands.go
mathetake: let's revert this change on amode

internal/engine/wazevo/backend/isa/amd64/instr_encoding.go
```diff
@@ -463,15 +463,102 @@ func (i *instruction) encode(c backend.Compiler) {
...
+		} else {
+			rex = rexInfo(0).clearW()
+		}
+		op1 := i.op1
+		if op1.kind == operandKindReg {
+			src := regEncodings[op1.r.RealReg()]
+			encodeRegReg(c, prefix, opcode, opcodeNum, dst, src, rex)
+		} else if i.op1.kind == operandKindMem {
+			m := i.op1.amode
+			encodeRegMem(c, prefix, opcode, opcodeNum, dst, m, rex)
```
mathetake, line 500: I believe this line is not covered in the tests

internal/engine/wazevo/backend/isa/amd64/instr_encoding.go
```diff
@@ -909,11 +1037,19 @@ func (ri rexInfo) encodeForIndex(c backend.Compiler, encR regEnc, encIndex regEn
...
+	return regRexBit(byte(r))
 }
 
 func (r regEnc) encoding() byte {
-	return byte(r) & 0x07
+	return regEncoding(byte(r))
+}
+
+func regRexBit(r byte) byte {
+	return r >> 3 & 1
```
mathetake, line 1048: & 1 seems unneeded 
```suggestion
	return r >> 3
```

evacchi:
rebased and force pushed

mathetake (approved):

PR #1920, opened 2024-01-15
wazevo(amd64): encode cmovcc, condcc, setcc, cmpRmiR, xmmRmiReg

evacchi:
- arm64: encode xmmRmiReg
- arm64: encode cmpRmiR
- arm64: encode setcc + fix rex.encode()
- arm64: encode condcc
- arm64: encode cmovcc

Note: it looks like `rex.encode()` should emit the prefix when `ri&0x02 != 0`, not when `== 1`. This in turn modified the result for `mov.b %rdi, 1048576(%rax)`, I think that should be correct.


mathetake (approved):

internal/engine/wazevo/backend/isa/amd64/instr_encoding_test.go
```diff
@@ -1866,7 +1866,7 @@ func TestInstruction_format_encode(t *testing.T) {
 				a := newOperandMem(newAmodeImmReg(1<<20, raxVReg))
 				i.asMovRM(rdiVReg, a, 1)
 			},
-			want:       "88b800001000",
+			want:       "4088b800001000",
```
mathetake, line 1869: good catch

PR #1923, opened 2024-01-16
threads: use mmap to allocate shared memory

anuraaga:
Actually this is my first time implementing "use mmap when shared _and compiler_" so not sure if it's the best approach. Open to any suggestions to improve

### DIFF
--- a/internal/platform/mmap_windows.go
+++ b/internal/platform/mmap_windows.go
@@ -57,6 +57,20 @@ func virtualProtect(address, size, newprotect uintptr, oldprotect *uint32) error
 	return nil
 }
 
+func mmapMemory(size int) ([]byte, error) {
+	p, err := allocateMemory(uintptr(size), windows_PAGE_READWRITE)
+	if err != nil {
+		return nil, err
+	}
+
+	var mem []byte
+	sh := (*reflect.SliceHeader)(unsafe.Pointer(&mem))
+	sh.Data = p
+	sh.Len = size
+	sh.Cap = size
+	return mem, err
+}
+
 func mmapCodeSegmentAMD64(size int) ([]byte, error) {
 	p, err := allocateMemory(uintptr(size), windows_PAGE_EXECUTE_READWRITE)
 	if err != nil {
--- a/internal/platform/mmap_windows.go
+++ b/internal/platform/mmap_windows.go
@@ -2,7 +2,6 @@ package platform
 
 import (
 	"fmt"
-	"reflect"
 	"syscall"
 	"unsafe"
 )
@@ -63,12 +62,7 @@ func mmapMemory(size int) ([]byte, error) {
 		return nil, err
 	}
 
-	var mem []byte
-	sh := (*reflect.SliceHeader)(unsafe.Pointer(&mem))
-	sh.Data = p
-	sh.Len = size
-	sh.Cap = size
-	return mem, err
+	return unsafe.Slice((*byte)(unsafe.Pointer(p)), size), nil
 }
 
 func mmapCodeSegmentAMD64(size int) ([]byte, error) {
@@ -77,12 +71,7 @@ func mmapCodeSegmentAMD64(size int) ([]byte, error) {
 		return nil, err
 	}
 
-	var mem []byte
-	sh := (*reflect.SliceHeader)(unsafe.Pointer(&mem))
-	sh.Data = p
-	sh.Len = size
-	sh.Cap = size
-	return mem, err
+	return unsafe.Slice((*byte)(unsafe.Pointer(p)), size), nil
 }
 
 func mmapCodeSegmentARM64(size int) ([]byte, error) {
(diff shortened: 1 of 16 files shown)


internal/wasm/memory.go
```diff
@@ -56,19 +57,63 @@ type MemoryInstance struct {
...
 	capacity := MemoryPagesToBytesNum(memSec.Cap)
+
+	var buffer []byte
+	var cap uint32
+	mmappedMemory := false
+	if memSec.IsShared && compilerEngine {
+		// When using compiler, memory accesses can happen at the same time that memory is resized, meaning
+		// we cannot have the memory base move during operation. mmap allows allocating memory virtually so
+		// we can grow without changing the base. The spec requires max for shared memory currently because
+		// all threads implementations are effectively expected to use mmap for shared memory.
```
mathetake, line 77: Is it really the only issue for compiler? I think interpreter can also have the same problem e.g. what happens if memory grow happens right before https://github.com/tetratelabs/wazero/blob/bfd9a00cd64e4380d5ab308fc85f9650a6370d9d/internal/engine/interpreter/interpreter.go#L1624 ?

mathetake:
I would like an input from @ncruces

mathetake:
btw if it is really only the case for compiler, then the introduction of IstCompiler on Engine interface seems good to me

internal/wasm/memory.go
```diff
@@ -56,19 +57,63 @@ type MemoryInstance struct {
...
+		max := MemoryPagesToBytesNum(memSec.Max)
+		b, err := platform.MmapMemory(int(max))
+		if err != nil {
+			panic(fmt.Errorf("unable to mmap memory: %w", err))
+		}
+		sp := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+		sp.Len = int(MemoryPagesToBytesNum(memSec.Min))
+		buffer = b
```
ncruces, line 85: Can't we just do `buffer = b[:MemoryPagesToBytesNum(memSec.Min)]`, instead of using `sp`?

internal/platform/mmap_windows.go
```diff
@@ -57,6 +57,20 @@ func virtualProtect(address, size, newprotect uintptr, oldprotect *uint32) error
...
 }
 
+func mmapMemory(size int) ([]byte, error) {
+	p, err := allocateMemory(uintptr(size), windows_PAGE_READWRITE)
+	if err != nil {
+		return nil, err
+	}
+
+	var mem []byte
+	sh := (*reflect.SliceHeader)(unsafe.Pointer(&mem))
```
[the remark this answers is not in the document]
mathetake: woah I thought unsafe.Slice was introduced like 1.20 or newer but yeah seems like we can use it right now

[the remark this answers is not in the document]
ncruces: Only the opposite conversion was 1.20.

internal/wasm/memory.go
anuraaga: Ah I forgot when I first wrote this code, Grow was guarded by the same mutex but currently it's not. With that lock there could be no accessed I think.

Since interpreter can be used on platforms without mmap, should I restore the lock on grow, only when mmap isn't used? Let me know if it makes sense

mathetake: even grow was guarded,  accessing `memoryInst.Buffer` can be unsafe as e.g., what if grow happens right after dereferencing []byte? loads/stores are not guarded by mutex in the first place

anuraaga: Ah that's true, it's not thread-safe code but also shouldn't crash. I guess it means we need to use mmap with interpreter too. I'll remove the checks in this code and in the future we can return an error is user specifies threads flag and mmap isn't supported?

mathetake: I think we simply use `make([]byte, maxpages)` always for interpreter, wdyt? cc @ncruces

anuraaga: maxpages will often be something like `4GB` since it's assuming use of mmap. It's because it's hard to predict the actual required max memory when compiling a binary since it can be run in many different ways.

https://github.com/wasix-org/wasix-libc/blob/main/.github/workflows/files/clang-wasix.cmake_toolchain#L5
https://github.com/wasilibs/go-re2/blob/main/buildtools/re2/Dockerfile#L51

mathetake: I think I'm fine with it until someone screams assuming few people will use the thread proposal on interpreter env

anuraaga: Ok I went ahead and went with that. One use case I heard of interpreter was compiling sqlc itself to wasm to run. But maybe we can tweak later

ncruces: I agree this seems the only sane way to do it. Users that scream have remedy: edit their WASM to reduce max memory.

Does this mean we can avoid `IsCompiler`, use `mmap` for shared where we can, huge slice where we can't?

anuraaga: Yeah that's also an option, want to confirm it's OK to use mmap with interpreter when it's supported as I believe so far we haven't used much syscall in that codepath.

mathetake (approved):

PR #1926, opened 2024-01-17
wazevo(amd64): Iadd, Isub lowering

evacchi:
add admits various operand combinations, but `asAluRmiR()` accepts only `rm operand, rd reg` (at least for now).

This PR also sets up clobber/declobber in the prologue/epilogue as it is made necessary in the e2e test case.


internal/engine/wazevo/backend/isa/amd64/machine.go
```diff
@@ -138,11 +138,70 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
...
+		panic("BUG?")
+	}
+
+	_64 := x.Type().Bits() == 64
+
+	xDef := m.c.ValueDefinition(x)
+	yDef := m.c.ValueDefinition(y)
+
+	rn := m.getOperand(xDef)
+	rm := m.getOperand(yDef)
```
mathetake, line 162: AluRmiR can have a memory operand, so jsut using getOperand seems too simple

internal/engine/wazevo/backend/isa/amd64/instr.go
```diff
@@ -1480,6 +1514,7 @@ type useKind byte
 const (
 	useKindNone useKind = iota + 1
 	useKindOp1
+	useKindOp1Reg
```
mathetake, line 1517: As I said aluRmiR can have reg, imm and memory, so this should be generalized regardless of reg or not.
```suggestion
	useKindOp1Op2
```

internal/engine/wazevo/backend/isa/amd64/machine.go
mathetake: so in other words, rm can be any of imm, mem, reg operands

internal/engine/wazevo/backend/isa/amd64/instr.go
```diff
@@ -239,6 +239,11 @@ func (i *instruction) Uses(regs *[]regalloc.VReg) []regalloc.VReg {
 	*regs = (*regs)[:0]
 	switch uk := useKinds[i.kind]; uk {
 	case useKindNone:
+	case useKindOp1Reg:
+		//if op.kind != operandKindReg {
+		//	panic("BUG: unsupported operand")
+		//}
+		*regs = append(*regs, i.op1.r, i.op2.r)
```
mathetake, line 246: should be generalized op2 can be any of imm, mem, reg

internal/engine/wazevo/backend/isa/amd64/machine_pro_epi_logue.go
```diff
@@ -46,15 +46,38 @@ func (m *machine) SetupPrologue() {
...
+		//          |    Caller_RBP   |      ====>       |    Caller_RBP   |
+		//   SP---> +-----------------+                  +-----------------+ <----- RBP
+		//             (low address)                     |   clobbered M   |
+		//                                               |   clobbered 1   |
+		//                                               |   ...........   |
+		//                                               |   clobbered 0   |
+		//                                               +-----------------+ <----- RSP
+		//
+		for _, vr := range regs {
+			cur = linkInstr(cur, m.allocateInstr().asPush64(newOperandReg(vr)))
```
mathetake, line 79: same

internal/engine/wazevo/backend/isa/amd64/machine.go
```diff
@@ -179,6 +184,25 @@ func (m *machine) lowerAluRmiROp(si *ssa.Instruction, op aluRmiROpcode) {
...
 }
 
+func (m *machine) getOperand_Mem_Imm32_Reg(def *backend.SSAValueDefinition) (op operand) {
+	if def.IsFromBlockParam() {
+		return newOperandReg(def.BlkParamVReg)
+	}
+
+	instr := def.Instr
+	switch instr.Opcode() {
+	case ssa.OpcodeLoad:
```
mathetake, line 194: just simply matching on opcode is troublesome. look at arm64 carefully and you will know why

internal/engine/wazevo/backend/isa/amd64/machine_pro_epi_logue.go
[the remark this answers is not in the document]
mathetake: I am taking care of this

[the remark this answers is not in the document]
mathetake: https://github.com/tetratelabs/wazero/pull/1931 here we go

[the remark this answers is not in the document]
evacchi: ok thanks for the help!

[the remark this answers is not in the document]
evacchi: rebased

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -0,0 +1,164 @@
...
+					// Sign-extension of a 32-bit constant can be merged into the offset.
+					offset += int64(int32(inputDef.Instr.ConstantVal())) // sign-extend!
+				default:
+					// Cannot fold into a constant, ignore.
+					continue
+				}
+				def.Instr.MarkLowered()
+				continue
+			}
+			// Note: case Ishl x, y could be handled too when the offset amount is <= 3.
```
evacchi, line 110: I am trying to do this; basically, instead of returning VReg we can return a pair `(VReg, byte)`, but then folding complicates a bit; e.g. when a single `(VReg, byte)` is present, we should either use the missing `amodeImmRegShift` or zero-out some register so that we can use `amodeRegRegImmShift`.

Even when we `add` together all the remaining addends extra care becomes necessary because in some cases we might want to substitute `(add x (shl y amt))` with `(add x [y+amt])`; but even in this case `[y+amt]` is either  `amodeRegRegImmShift` or needs a zero reg.

evacchi: I have added this to commit 6c687e40 but we can revert it if that adds too much noise

mathetake:
will take it from here

PR #1927, opened 2024-01-18
threads: for arm64, when shared memory ignore memory length cache

anuraaga:
When looking at arm64 code for threads, I believe this is the only invasive change to current code, then it will be just implementing new ops. So sending this separately first.

Currently, we cache the buffer length into the call engine on every function preamble, which is also updated when calling memory.Grow, but this does not work when there are concurrent accesses. So now, when memory is shared, it is assumed there may be concurrent grows and length is read directly every time.

Unlike wazerox, I have made this happen only with shared memory (there I actually removed the length cache from the call engine) so this code is somewhat new to me, notably the large-ish `MemoryType` change.

### DIFF
--- a/internal/wasm/memory.go
+++ b/internal/wasm/memory.go
@@ -266,11 +266,7 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 		sp := (*reflect.SliceHeader)(unsafe.Pointer(&m.Buffer))
 		if m.Shared {
 			// Use atomic write to ensure new length is visible across threads.
-			if math.MaxInt == math.MaxInt32 {
-				atomic.StoreInt32((*int32)(unsafe.Pointer(&sp.Len)), int32(MemoryPagesToBytesNum(newPages)))
-			} else {
-				atomic.StoreInt64((*int64)(unsafe.Pointer(&sp.Len)), int64(MemoryPagesToBytesNum(newPages)))
-			}
+			atomic.StoreUintptr((*uintptr)(unsafe.Pointer(&sp.Len)), uintptr(MemoryPagesToBytesNum(newPages)))
 		} else {
 			sp.Len = int(MemoryPagesToBytesNum(newPages))
 		}
--- a/internal/wasm/memory.go
+++ b/internal/wasm/memory.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"reflect"
 	"sync"
+	"sync/atomic"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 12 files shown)


internal/wasm/memory.go
```diff
@@ -263,7 +264,16 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 		return currentPages, true
 	} else { // We already have the capacity we need.
 		sp := (*reflect.SliceHeader)(unsafe.Pointer(&m.Buffer))
-		sp.Len = int(MemoryPagesToBytesNum(newPages))
+		if m.Shared {
+			// Use atomic write to ensure new length is visible across threads.
+			if math.MaxInt == math.MaxInt32 {
```
anuraaga, line 269: TBH I guess I've never ever tested this branch

anuraaga:
@ncruces Pre-pinging you as I guess @mathetake would :)

internal/wasm/memory.go
ncruces: I think you can use `atomic.StoreUintptr` here.

ncruces (commented):
TBH I don't really have context here.
Can you give me a primer on what `compileLoadMemoryBufferLen` used for?

Otherwise, LGTM.

anuraaga:
It loads the current length of the memory buffer, most commonly for bounds checks on memory accesses I think

ncruces:
> It loads the current length of the memory buffer, most commonly for bounds checks on memory accesses I think

I get that, but I guess it depends on _usage_ what's the required “happens before” relationship. Do we need atomic because we're dereference a pointer and memory must have grown before the pointer being created?

anuraaga:
While I'm not sure, one example that comes to mind is

Thread 1 malloc's, growing, and creates pointer
Thread 1 stores pointer in given location in memory, a pointer pointer
Thread 2 reads pointer pointer than dereferences.

Does it seem like a valid example of happens before?

ncruces:
Makes sense, and atomic load/store of the length should be enough to fix that (no need for CAS, or anything more complicated), as long as the pointer is also moved race-free (which it needs to, otherwise that's a bug in the guest).

internal/wazeroir/compiler.go
```diff
@@ -213,6 +213,14 @@ func (c *Compiler) resetUnreachable() {
...
 }
 
+type MemoryType byte
+
+const (
+	MemoryTypeNone MemoryType = iota
+	MemoryTypeStandard
+	MemoryTypeShared
+)
+
```
mathetake, line 223: could you add Go doc comments

mathetake (approved):

ncruces (approved):

internal/wazeroir/compiler.go
anuraaga: Oops, fixed

PR #1929, opened 2024-01-19
threads: compile threads proposal instructions for arm64

anuraaga:


anuraaga:
I thought that there may be some way to send a followup PR to another one but against a fork, nah guess not.

Feel free to add comments in https://github.com/anuraaga/wazero/pull/1/files before I send it here if you feel like

mathetake (approved):

PR #1930, opened 2024-01-20
threads: compile threads proposal instructions for amd64

anuraaga:


mathetake (approved):
cool, thank you for all the hard work! I guess this is the last one? then maybe let's create the experimental/features package to expose the thread-proposal flag

PR #1940, opened 2024-01-23
wazevo(amd64): amode lowering

evacchi:
add address mode lowering/folding with tests

### DIFF
--- a/internal/engine/wazevo/backend/isa/amd64/lower_mem.go
+++ b/internal/engine/wazevo/backend/isa/amd64/lower_mem.go
@@ -6,7 +6,7 @@ import (
 	"github.com/tetratelabs/wazero/internal/engine/wazevo/ssa"
 )
 
-var addendsMatchOpcodes = [5]ssa.Opcode{ssa.OpcodeUExtend, ssa.OpcodeSExtend, ssa.OpcodeIadd, ssa.OpcodeIconst, ssa.OpcodeIshl}
+var addendsMatchOpcodes = [...]ssa.Opcode{ssa.OpcodeUExtend, ssa.OpcodeSExtend, ssa.OpcodeIadd, ssa.OpcodeIconst, ssa.OpcodeIshl}
 
 type (
 	addend64 struct {
@@ -147,13 +147,10 @@ func (m *machine) lowerAddendFromInstr(instr *ssa.Instruction) addend64 {
 		// If the addend is a shift, we can only handle it if the shift amount is a constant.
 		x, amount := instr.Arg2()
 		amountDef := m.c.ValueDefinition(amount)
-		if !amountDef.IsFromInstr() || !amountDef.Instr.Constant() || amountDef.Instr.ConstantVal() > 3 {
-			return addend64{m.getOperand_Reg(m.c.ValueDefinition(instr.Return())).r, 0, 0}
+		if amountDef.IsFromInstr() && amountDef.Instr.Constant() && amountDef.Instr.ConstantVal() <= 3 {
+			return addend64{m.getOperand_Reg(m.c.ValueDefinition(x)).r, 0, uint8(amountDef.Instr.ConstantVal())}
 		}
-		instr.MarkLowered()
-		amountDef.Instr.MarkLowered()
-		return addend64{m.getOperand_Reg(m.c.ValueDefinition(x)).r, 0, uint8(amountDef.Instr.ConstantVal())}
-
+		return addend64{m.getOperand_Reg(m.c.ValueDefinition(x)).r, 0, 0}
 	}
 	panic("BUG: invalid opcode")
 }
--- a/internal/engine/wazevo/backend/isa/amd64/lower_mem.go
+++ b/internal/engine/wazevo/backend/isa/amd64/lower_mem.go
@@ -1,10 +1,99 @@
 package amd64
 
 import (
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/backend/regalloc"
 	"github.com/tetratelabs/wazero/internal/engine/wazevo/ssa"
 )
 
+var addendsMatchOpcodes = [5]ssa.Opcode{ssa.OpcodeUExtend, ssa.OpcodeSExtend, ssa.OpcodeIadd, ssa.OpcodeIconst, ssa.OpcodeIshl}
+
 // lowerToAddressMode converts a pointer to an addressMode that can be used as an operand for load/store instructions.
 func (m *machine) lowerToAddressMode(ptr ssa.Value, offsetBase uint32) (am amode) {
-	panic("TODO")
+	if !lower32willSignExtendTo64(uint64(offsetBase)) {
+		tmpReg := m.c.AllocateVReg(ssa.TypeI64)
+		m.lowerIconst(tmpReg, uint64(offsetBase), true)
+		return newAmodeImmReg(0, tmpReg)
+	}
+
+	offBase := int32(offsetBase)
+	def := m.c.ValueDefinition(ptr)
+	switch op := m.c.MatchInstrOneOf(def, addendsMatchOpcodes[:]); op {
+	case ssa.OpcodeIadd:
+		x, y := def.Instr.Arg2()
+		xDef, yDef := m.c.ValueDefinition(x), m.c.ValueDefinition(y)
+		rx, offx := m.lowerAddend(xDef.Instr)
+		ry, offy := m.lowerAddend(yDef.Instr)
+
+		imm32 := uint32(offBase + offx + offy)
+		if imm32 != 0 && lower32willSignExtendTo64(uint64(imm32)) {
+			switch {
+			case rx != regalloc.VRegInvalid && ry != regalloc.VRegInvalid:
+				return newAmodeRegRegShift(imm32, rx, ry, 0)
+			case rx != regalloc.VRegInvalid && ry == regalloc.VRegInvalid:
+				return newAmodeImmReg(imm32, rx)
+			case rx == regalloc.VRegInvalid && ry != regalloc.VRegInvalid:
+				return newAmodeImmReg(imm32, ry)
+			case rx == regalloc.VRegInvalid && ry == regalloc.VRegInvalid:
+				tmpReg := m.c.AllocateVReg(ssa.TypeI64)
+				m.lowerIconst(tmpReg, uint64(offBase+offx+offy), true)
+				return newAmodeImmReg(0, tmpReg)
+			}
+		} else {
+			tmpReg := m.c.AllocateVReg(ssa.TypeI64)
+			m.lowerIconst(tmpReg, uint64(offBase), true)
+			return newAmodeImmReg(0, tmpReg)
+		}
+
+	default:
+		r, off := m.lowerAddend(def.Instr)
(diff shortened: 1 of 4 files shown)


internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,116 @@
...
+		} else {
+			ry = tmpReg
+		}
+	}
+
+	u32 := uint32(u64)
+	switch {
+	// We assume rx, ry are valid iff offx, offy are 0.
+	case rx != regalloc.VRegInvalid && ry != regalloc.VRegInvalid:
+		return newAmodeRegRegShift(u32, rx, ry, 0)
```
mathetake, line 58: non zero shift pattern is difficult to implement?

evacchi: doable, it complicates a bit the flow, not sure how worth it is; looking into it.

mathetake: you can write a ssa level pass that counts a patter we are looking for and run it against Zig stdlib on aarch64 maybe :)

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,116 @@
...
+	case rx != regalloc.VRegInvalid && ry == regalloc.VRegInvalid:
+		return newAmodeImmReg(u32, rx)
+	case rx == regalloc.VRegInvalid && ry != regalloc.VRegInvalid:
+		return newAmodeImmReg(u32, ry)
+	default: // Both are invalid: use the offset.
+		// TODO: if offset == 0, xor vreg, vreg.
```
mathetake, line 64: it's already done inside lowerIconst
```suggestion
```

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,116 @@
...
+	if x.IsFromBlockParam() {
+		return x.BlkParamVReg, 0
+	}
+	return m.lowerAddendFromInstr(x.Instr)
+}
+
+// lowerAddendFromInstr takes an instruction returns a Vreg and an offset that can be used in an address mode.
+// The Vreg is regalloc.VRegInvalid if the addend cannot be lowered to a register.
+// The offset is 0 if the addend can be lowered to a register.
+func (m *machine) lowerAddendFromInstr(instr *ssa.Instruction) (regalloc.VReg, int64) {
```
mathetake, line 81: instr.MarkLowered() can be called here

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,116 @@
...
+	switch op := instr.Opcode(); op {
+	case ssa.OpcodeIconst:
+		instr.MarkLowered()
+		u64 := instr.ConstantVal()
+		if instr.Return().Type().Bits() == 32 {
+			return regalloc.VRegInvalid, int64(int32(u64)) // sign-extend.
+		} else {
+			return regalloc.VRegInvalid, int64(u64)
+		}
+	case ssa.OpcodeUExtend, ssa.OpcodeSExtend:
```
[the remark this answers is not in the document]
evacchi: this function is meant to be called (transitively) by `lowerToAddressMode()` which has already used `MatchInstr`, so when we get here we should be sure that the value is referenced only once.

If you instead mean something like `UExtend(UExtend(...))` then we are only looking ahead once, which means we will fall in the default case, if we want to look ahead further, then the function should probably become ~recursive.

...or maybe I am missing something else :)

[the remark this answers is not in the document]
evacchi: clarified offline

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
evacchi: I added Ishl in b6606bba21abc976ad3981cb1216dbeffa25007f, I can always do the SSA level pass later (that should be fun). We can also revert back this commit if it's too significant.

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,159 @@
...
+		}
+	case x.r == regalloc.VRegInvalid && y.r != regalloc.VRegInvalid:
+		x, y = y, x
+		fallthrough
+	case x.r != regalloc.VRegInvalid && y.r == regalloc.VRegInvalid:
+		if x.shift != 0 {
+			zero := m.c.AllocateVReg(ssa.TypeI64)
+			m.lowerIconst(zero, 0, true)
+			return newAmodeRegRegShift(u32, zero, x.r, x.shift)
+		}
```
evacchi, line 95: this can be either lowered to a shift instruction, or be merged into RegRegShift with a zero reg; I decided for the latter since we know it's already <= 3

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,159 @@
 package amd64
 
 import (
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/backend"
+	"github.com/tetratelabs/wazero/internal/engine/wazevo/backend/regalloc"
 	"github.com/tetratelabs/wazero/internal/engine/wazevo/ssa"
 )
 
+var addendsMatchOpcodes = [5]ssa.Opcode{ssa.OpcodeUExtend, ssa.OpcodeSExtend, ssa.OpcodeIadd, ssa.OpcodeIconst, ssa.OpcodeIshl}
```
mathetake, line 9:
```suggestion
var addendsMatchOpcodes = [...]ssa.Opcode{ssa.OpcodeUExtend, ssa.OpcodeSExtend, ssa.OpcodeIadd, ssa.OpcodeIconst, ssa.OpcodeIshl}
```

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,159 @@
...
+			}
+		}
+	case ssa.OpcodeIshl:
+		// If the addend is a shift, we can only handle it if the shift amount is a constant.
+		x, amount := instr.Arg2()
+		amountDef := m.c.ValueDefinition(amount)
+		if !amountDef.IsFromInstr() || !amountDef.Instr.Constant() || amountDef.Instr.ConstantVal() > 3 {
+			return addend64{m.getOperand_Reg(m.c.ValueDefinition(instr.Return())).r, 0, 0}
+		}
+		instr.MarkLowered()
```
mathetake, line 153: redundant ?

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,159 @@
...
+		amountDef.Instr.MarkLowered()
```
mathetake, line 154: consts are not needed to call MarkLowered

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,159 @@
...
+				return addend64{regalloc.VRegInvalid, int64(int32(inputDef.Instr.ConstantVal())), 0} // sign-extend!
+			default:
+				return addend64{m.getOperand_Reg(inputDef).r, 0, 0}
+			}
+		}
+	case ssa.OpcodeIshl:
+		// If the addend is a shift, we can only handle it if the shift amount is a constant.
+		x, amount := instr.Arg2()
+		amountDef := m.c.ValueDefinition(amount)
+		if !amountDef.IsFromInstr() || !amountDef.Instr.Constant() || amountDef.Instr.ConstantVal() > 3 {
```
mathetake, line 150: can you revert the branch?
```suggestion
		if amountDef.IsFromInstr() && amountDef.Instr.Constant() && amountDef.Instr.ConstantVal() <= 3 {
```

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,159 @@
...
+			return addend64{regalloc.VRegInvalid, int64(int32(u64)), 0} // sign-extend.
+		} else {
+			return addend64{regalloc.VRegInvalid, int64(u64), 0}
+		}
+	case ssa.OpcodeUExtend, ssa.OpcodeSExtend:
+		input := instr.Arg()
+		inputDef := m.c.ValueDefinition(input)
+		switch input.Type().Bits() {
+		case 64:
+			return addend64{m.getOperand_Reg(inputDef).r, 0, 0}
```
mathetake, line 134: does this branch happen?

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
```diff
@@ -1,10 +1,159 @@
...
 
+var addendsMatchOpcodes = [5]ssa.Opcode{ssa.OpcodeUExtend, ssa.OpcodeSExtend, ssa.OpcodeIadd, ssa.OpcodeIconst, ssa.OpcodeIshl}
+
+type (
+	addend64 struct {
+		r     regalloc.VReg
+		off   int64
+		shift byte
+	}
+)
```
mathetake, line 17: 64 seems unnecessary


```suggestion
type addend struct {
	r     regalloc.VReg
	off   int64
	shift byte
}
```

mathetake (commented):
looking almost there!!

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
evacchi: ah yes, good point, I wanted to tell. It happens only in this case:

https://github.com/tetratelabs/wazero/blob/4b85945ee971264cb3a1d9276d7d64f1de867898/internal/engine/wazevo/backend/isa/amd64/lower_mem_test.go#L100-L110

I have included it because it is also there on `collectAddends()` on arm64, but TBH it seems very redundant, I am not even sure it lowers from valid wasm...

evacchi:
https://github.com/tetratelabs/wazero/actions/runs/7631628885/job/20790017217?pr=1940 ??

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
mathetake: No, I mean that shouldn't happen in the first place right?

mathetake: at the frontend I meant

mathetake: > also there on collectAddends() on arm64

OK that should be deleted then :D

mathetake:
OK seems unrelated but yeah seems like it must be fixed

internal/engine/wazevo/backend/isa/amd64/lower_mem.go
evacchi: > at the frontend I meant

yeah I don't think so

> OK that should be deleted then :D

:D

PR #1941, opened 2024-01-23
fuzz: enables threads by default

mathetake:
cc @anuraaga

### DIFF
--- a/internal/integration_test/fuzz/fuzz/build.rs
+++ b/internal/integration_test/fuzz/fuzz/build.rs
@@ -27,9 +27,9 @@ fn main() {
     let mut command = process::Command::new("go");
     command.current_dir(&wazero_fuzz_lib_dir);
     command.arg("build");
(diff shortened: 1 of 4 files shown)


mathetake:
@anuraaga do you want to fix the bug and make the impl more solid until like 4 hours of fuzzing run without issues?

anuraaga:
Fuzzing is indeed the best, that's really cool. I can fix the bug, but if you're asking if I can run a 4 hour fuzz, I don't have a always-on separate computer. I can see if the fuzz command in the background doesn't kill my Mac while doing normal work.

mathetake:
> I can see if the fuzz command in the background doesn't kill my Mac while doing normal work.

yeah usually running 2 or 4 jobs on m1 mac while doing other work could be possible (I am doing that alot)

mathetake:
also I would appreciate it if you could go ahead and fix the bug

mathetake:
ok now at least minimal fuzzing job (3m per target) passing

PR #1959, opened 2024-01-25
wazevo(amd64): lowering for popcnt, ctz, clz + cpuid refactor

evacchi:
Refactors cpuid so that it can be referred by non-arch specific code (e.g. for testing); then uses it for ctz/clz.


internal/platform/cpuid_amd64.go
```diff
@@ -69,11 +47,11 @@ func loadCpuFeatureFlags() CpuFeatureFlags {
...
-func (f *cpuFeatureFlags) Has(cpuFeature uint64) bool {
-	return (f.flags & cpuFeature) != 0
+func (f *cpuFeatureFlags) Has(cpuFeature CpuFeature) bool {
+	return (f.flags & uint64(cpuFeature)) != 0
 }
 
 // HasExtra implements the same method on the CpuFeatureFlags interface
-func (f *cpuFeatureFlags) HasExtra(cpuFeature uint64) bool {
-	return (f.extraFlags & cpuFeature) != 0
+func (f *cpuFeatureFlags) HasExtra(cpuFeature CpuFeature) bool {
```
mathetake, line 55:
```suggestion
func (f cpuFeatureFlags) HasExtra(cpuFeature CpuFeature) bool {
```

mathetake (commented):
could you make sure that the e2e passes when forcing this huge else branch on your machine? then good to go

evacchi:
ah, right. In fact, my m1 defaults to the "elses", but you reminded me that I had to boot my amd64 to check the default branch: it works!

mathetake (approved):

PR #1960, opened 2024-01-25
wazevo(amd64): lowering for div/rem

evacchi:
Not finished yet, I must more tests to verify all the code paths.

### DIFF
--- a/internal/engine/wazevo/backend/isa/amd64/machine.go
+++ b/internal/engine/wazevo/backend/isa/amd64/machine.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"math"
 	"strings"
 
 	"github.com/tetratelabs/wazero/internal/engine/wazevo/backend"
@@ -258,6 +259,10 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
 		m.lowerAluRmiROp(instr, aluRmiROpcodeSub)
 	case ssa.OpcodeImul:
 		m.lowerAluRmiROp(instr, aluRmiROpcodeMul)
+	case ssa.OpcodeSdiv, ssa.OpcodeUdiv, ssa.OpcodeSrem, ssa.OpcodeUrem:
+		isDiv := op == ssa.OpcodeSdiv || op == ssa.OpcodeUdiv
+		isSigned := op == ssa.OpcodeSdiv || op == ssa.OpcodeSrem
+		m.lowerIDivRem(instr, isDiv, isSigned)
 	case ssa.OpcodeBand:
 		m.lowerAluRmiROp(instr, aluRmiROpcodeAnd)
 	case ssa.OpcodeBor:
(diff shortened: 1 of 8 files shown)


internal/engine/wazevo/backend/isa/amd64/machine.go
```diff
@@ -621,8 +626,6 @@ func (m *machine) lowerShiftR(si *ssa.Instruction, op shiftROp) {
 		m.insert(alu)
 
 	} else {
-		println("::::::SONO QUI")
```
evacchi, line 624: leftover garbage...

mathetake:
maybe you might want to do rebase as I merged some bugs

internal/engine/wazevo/backend/isa/amd64/machine.go
```diff
@@ -964,3 +978,127 @@ func (m *machine) frameSize() int64 {
...
+	// If we are compiling a Rem instruction, when the divisor is -1 we land at the end of the function.
+	if ifRemNeg1 != nil {
+		nop, end := m.allocateBrTarget()
+		m.insert(nop)
+		ifRemNeg1.asJmpIf(condZ, newOperandLabel(end))
+	}
+
+	// Restore rax, rdx.
+	m.copyTo(raxTmp, raxVReg)
+	m.copyTo(rdxTmp, rdxVReg)
```
mathetake, line 1103: please think deeply about why do you think you need these restore?

internal/engine/wazevo/backend/isa/amd64/machine.go
```diff
@@ -940,8 +945,17 @@ func (m *machine) copyTo(src regalloc.VReg, dst regalloc.VReg) {
...
+		} else {
+			tmp = m.c.AllocateVReg(ssa.TypeV128)
+		}
+	} else {
+		typ := m.c.TypeOf(v)
+		tmp = m.c.AllocateVReg(typ)
+	}
 	m.copyTo(v, tmp)
 	return tmp
 }
```
mathetake, line 961: plase revert this change.

evacchi:
rebased, did a quick+dirty change, will review, add more tests. Promising!

internal/engine/wazevo/e2e_test.go
```diff
@@ -122,6 +122,128 @@ func TestE2E(t *testing.T) {
...
+				},
+				{ // This now succeeds.
+					params: []uint64{1, 1, 21, 0},
+					expErr: "wasm error: integer divide by zero",
+				},
+				//{ // This now fails.
+				//	params: []uint64{0x80000000, 0xffffffff, 1, 1},
+				//	expErr: "wasm error: integer overflow",
+				//},
+				{
```
evacchi, line 220: it seems like if I have more than one error trap in my function, and the generated code is long enough than the error handers down low are generated incorrectly and they SIGSEGV. In fact, this same e2e test above has DivS,RemS and (rem 21%0) crashed. This test case here (Rems, DivS) does not crash and raises the error correctly, but fails on integer overflow (which was handled correctly in the example above)

mathetake (approved):

PR #1964, opened 2024-01-26
wazevo: adds support for memory wait

anuraaga:
Not too sure what I'm doing so sending out lowering of one atomic op to understand the pattern, hopefully it will be enough to do the rest in the later PR.

I'm trying to compare to wasmtime but not sure if I'm doing something wrong. For this wat file

```
(module
  (import "console" "log" (func $log (param i32)))
  (func
    (i32.const 5)
    (i32.const 0)
    (i64.const 10)
    memory.atomic.wait32
    (call $log)
  )
  (memory 1 1 shared)
  (export "" (func 0))
)
```

I get

```
❯ target/debug/clif-util wasm --target aarch64-apple-darwin test.wat -p -t
function u0:1(i64 vmctx) fast {
    gv0 = vmctx
    gv1 = load.i64 notrap aligned readonly gv0
    sig0 = (i32, i64 vmctx) fast
    fn0 = u0:0 sig0

                                block0(v0: i64):
@0038                               v1 = iconst.i32 5
@003a                               v2 = iconst.i32 0
@003c                               v3 = iconst.i64 10
@003e                               v4 = iconst.i32 0xffff_ffff
@0042                               call fn0(v4, v0)  ; v4 = 0xffff_ffff
@0044                               jump block1

                                block1:
@0044                               return
}
```

I initially used the same wat as my test case but thought some optimization may be happening so tried adding the `call $log`, but still somehow it is just `v4 = iconst.i32 0xffff_ffff`. Any idea what I'm doing wrong?

### DIFF
--- a/internal/engine/wazevo/call_engine.go
+++ b/internal/engine/wazevo/call_engine.go
@@ -82,6 +82,8 @@ type (
 		memmoveAddress uintptr
 		// framePointerBeforeGoCall holds the frame pointer before calling a Go function. Note: only used in amd64.
 		framePointerBeforeGoCall uintptr
+		// memoryWaitAddress holds the address of memory wait trampoline function.
+		memoryWaitAddress *byte
 	}
 )
 
--- a/internal/engine/wazevo/call_engine.go
+++ b/internal/engine/wazevo/call_engine.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"reflect"
+	"sync/atomic"
 	"unsafe"
 
 	"github.com/tetratelabs/wazero/api"
@@ -82,8 +83,10 @@ type (
 		memmoveAddress uintptr
 		// framePointerBeforeGoCall holds the frame pointer before calling a Go function. Note: only used in amd64.
 		framePointerBeforeGoCall uintptr
-		// memoryWaitAddress holds the address of memory wait trampoline function.
-		memoryWaitAddress *byte
+		// memoryWait32TrampolineAddress holds the address of memory_wait32 trampoline function.
+		memoryWait32TrampolineAddress *byte
+		// memoryWait32TrampolineAddress holds the address of memory_wait64 trampoline function.
+		memoryWait64TrampolineAddress *byte
 	}
 )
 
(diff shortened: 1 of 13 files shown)


internal/engine/wazevo/ssa/instructions.go
```diff
@@ -648,6 +648,10 @@ const (
 	// `fence`.
 	OpcodeFence
 
+	// OpcodeMemoryWait ...
+	// `v = memory_wait MemFlags, t, x, p`.
+	OpcodeMemoryWait
```
anuraaga, line 653: Got pointer that I should be able to define host func call without a separate opcore will check it out

anuraaga:
Thanks @mathetake - can you check this is going in a reasonable direction? I think what's left is defining the trampoline for the wait function

internal/engine/wazevo/ssa/instructions.go
```diff
@@ -200,6 +200,9 @@ const (
 	// OpcodeExitIfTrueWithCode exits the execution immediately if the value `c` is not zero.
 	OpcodeExitIfTrueWithCode
 
+	// OpcodeExitIfEqWithCode exits the execution immediately if the condition flag is set.
+	OpcodeExitIfCondWithCode
```
mathetake, line 204: You won't need this

internal/engine/wazevo/ssa/instructions.go
```diff
@@ -323,6 +326,9 @@ const (
 	// OpcodeIcmpImm compares an integer value with the immediate value on the given condition: `v = icmp_imm Cond, x, Y`.
 	OpcodeIcmpImm
 
+	// OpcodeIandsImm performs a bitwise AND with the immediate value and updates condition flags.
+	OpcodeIandsImm
```
mathetake, line 330: This as well. There's no concept in the flag and you are supposed to use Icmp and ExitTrueIfTrueWithCond

mathetake (changes requested):

internal/engine/wazevo/ssa/instructions.go
mathetake: See other various places for doing similar thing eg memory bound check

mathetake:
Basically you should be able to achieve lowering without adding any new instruction

internal/engine/wazevo/call_engine.go
```diff
@@ -82,6 +82,8 @@ type (
 		memmoveAddress uintptr
 		// framePointerBeforeGoCall holds the frame pointer before calling a Go function. Note: only used in amd64.
 		framePointerBeforeGoCall uintptr
+		// memoryWaitAddress holds the address of memory wait trampoline function.
+		memoryWaitAddress *byte
```
mathetake, line 86: memoryWaitTrampolineAddress (memmove isn't many trampoline hence it doesn't have it in the name)

internal/engine/wazevo/ssa/instructions.go
anuraaga: Currently we can do alignment check with just one instruction for both archs

https://github.com/tetratelabs/wazero/blob/main/internal%2Fengine%2Fcompiler%2Fimpl_threads_arm64.go#L40
https://github.com/tetratelabs/wazero/blob/main/internal%2Fengine%2Fcompiler%2Fimpl_threads_amd64.go#L654

I'm not too sure how to do it with Icmp or should we use two instructions, OpcodeBmask then Icmp?

mathetake: Yes two instructions

mathetake: Yes two instructions

anuraaga:
Thanks for the patience @mathetake I think this is on the right way now

mathetake:
want to add a e2e test case in https://github.com/tetratelabs/wazero/blob/main/internal/engine/wazevo/e2e_test.go#L39 ?

anuraaga:
Added a simple e2e for now for the "not equals" case, other ones will be trickier / require notify to be implemented.

And fixed a bug

mathetake:
looking good! will take a look tomorrow

mathetake (approved):
💯

mathetake:
lastly could you fix the conflict

anuraaga:
Thanks, resolved conflicts

mathetake (approved):

PR #1971, opened 2024-01-29
wazevo(amd64): float rounding, min/max

evacchi:


mathetake (approved):
this sparks my nasty memory I implemented this logic in the old compiler 😄

PR #2004, opened 2024-02-02
wazevo(amd64): simd arith spec tests

evacchi:

PR #1997, opened 2024-02-01
wazevo(amd64): simd boolean, bitwise spec tests

evacchi:
implements anytrue, alltrue, bitmask, and, or, xor, andnot, bselect

PR #2005, opened 2024-02-02
wazevo(amd64): simd arith2 spec tests

evacchi:
this is based on #2004, ~~and #1997~~ for easier merging


evacchi:
rebased and fixed regalloc issues

evacchi:
rebased, added missing encoding test, fix vreg naming, allow mem args

evacchi:
I swapped mem/reg args where appropriate to use copyToTmp(), but in some cases (unaries: abs/popcnt) I think we still need to do an explicit copy.

evacchi:
force-push to retrigger CI which didn't run last time...

evacchi:
(I suspect CI won't run until the PR targets main)

mathetake (approved):

PR #2022, opened 2024-02-07
wazevo(regalloc): map-free liveness for less memory pressure

mathetake:
this reduces the memory pressure/allocations drastically as you can see in the following result for zig stdlib test compilation: 

```
goos: darwin
goarch: arm64
                                        │   old.txt   │          new_old.txt           │
                                        │   sec/op    │   sec/op     vs base           │
Zig/optimizing/Compile/test-opt.wasm-10   6.200 ± ∞ ¹   4.788 ± ∞ ¹  ~ (p=1.000 n=1) ²

                                        │    old.txt    │           new_old.txt            │
                                        │     B/op      │     B/op       vs base           │
Zig/optimizing/Compile/test-opt.wasm-10   745.6Mi ± ∞ ¹   583.6Mi ± ∞ ¹  ~ (p=1.000 n=1) ²

                                        │   old.txt    │           new_old.txt           │
                                        │  allocs/op   │  allocs/op    vs base           │
Zig/optimizing/Compile/test-opt.wasm-10   2.249M ± ∞ ¹   1.249M ± ∞ ¹  ~ (p=1.000 n=1) ²

```

which shows that allocation counts gets halved and the memory space gets reduced by like 160MB!


mathetake:
fuzzing time

PR #2024, opened 2024-02-08
wazevo(ssa): avoids allocation during natural order sort

mathetake:
This starts using `slices.SortFunc` (thanks to @achille-roussel) to avoid
allocations during `ssa.passSortSuccessors`. As a result for zig stdlib compilation,
huge amount of allocations has been eliminated.

```
                                        │    old.txt    │             new.txt              │
                                        │     B/op      │     B/op       vs base           │
Zig/optimizing/Compile/test-opt.wasm-10   513.2Mi ± ∞ ¹   506.6Mi ± ∞ ¹  ~ (p=1.000 n=1) ²

                                        │    old.txt    │             new.txt             │
                                        │   allocs/op   │  allocs/op    vs base           │
Zig/optimizing/Compile/test-opt.wasm-10   1201.3k ± ∞ ¹   781.8k ± ∞ ¹  ~ (p=1.000 n=1) ²

```

### DIFF
--- a/internal/engine/wazevo/ssa/basic_block_sort.go
+++ b/internal/engine/wazevo/ssa/basic_block_sort.go
@@ -1,0 +1,23 @@
+//go:build go1.21
+
+package ssa
+
+import (
+	"slices"
+)
+
+func sortBlocks(blocks []*basicBlock) {
+	slices.SortFunc(blocks, func(i, j *basicBlock) int {
+		if j.ReturnBlock() {
+			return 1
+		}
+		if i.ReturnBlock() {
+			return -1
+		}
+		iRoot, jRoot := i.rootInstr, j.rootInstr
+		if iRoot == nil || jRoot == nil { // For testing.
+			return 1
+		}
+		return j.rootInstr.id - i.rootInstr.id
+	})
+}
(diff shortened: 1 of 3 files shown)


internal/engine/wazevo/ssa/basic_block_sort.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func sortBlocks(blocks []*basicBlock) {
+	slices.SortFunc(blocks, func(i, j *basicBlock) int {
+		if j.ReturnBlock() {
+			return 1
+		}
+		if i.ReturnBlock() {
+			return -1
+		}
```
achille-roussel, line 16: Should the function return zero when `i.ReturnBlock() && j.ReturnBlock()` ?

I honestly don't know if it matters, but it could confuse the sorting algorithm if `f(i, h) == 1` but also `f(j, i) == 1`.

The comparison function has three states bascially for less than, equal, and greater than, which is slightly different from the "less" function that was historically used in the `sort` package.

mathetake: yeah it should return zero but it doesn't make any difference in the result at the end of the day. Would you mind raising a fix if you like? 🏃

achille-roussel: These things make me nervous, I've seen quick-sort turn into O(N^2) because the comparison function wasn't producing a coherent output.

Sending a PR 👍

PR #2026, opened 2024-02-08
return 0 is compare function of sortBlocks when blocks are equal

achille-roussel:
Follow-up to https://github.com/tetratelabs/wazero/pull/2024; it's possible that this isn't important, but it could prevent pathological cases where the sort algorithm gets confused because the compare function doesn't provide total ordering.


mathetake (approved):
💯

PR #2028, opened 2024-02-08
wazevo: lower atomic fence to ssa

anuraaga:


internal/engine/wazevo/frontend/lower.go
```diff
@@ -3393,6 +3393,13 @@ func (c *Compiler) lowerCurrentOpcode() {
 			c.memAlignmentCheck(addr, size)
 			res := builder.AllocateInstruction().AsAtomicCas(addr, exp, repl, size).Insert(builder).Return()
 			state.push(res)
+		case wasm.OpcodeAtomicFence:
+			order := c.readByte()
+			c.readI32u()
+			if state.unreachable {
+				break
+			}
+			builder.AllocateInstruction().AsFence(order).Insert(builder)
```
mathetake, line 3402: I see this in the interpreter 
>  case wazeroir.OperationKindAtomicFence:
			// Memory not required for fence only
			if memoryInst != nil {
				// An empty critical section can be used as a synchronization primitive, which is what
				// fence is. Probably, there are no spectests or defined behavior to confirm this yet.
				memoryInst.Mux.Lock()
				memoryInst.Mux.Unlock() //nolint:staticcheck
			}



so could you optimize out if the current wasm module doesn't have memory?

anuraaga: Nice idea, fixed

mathetake (approved):

PR #2031, opened 2024-02-09
wazevo(arm64): support for atomic cas

anuraaga:


internal/engine/wazevo/backend/isa/arm64/instr.go
```diff
@@ -212,7 +213,11 @@ const (
 	useKindAMode
 	useKindRNAMode
 	useKindCond
-	useKindVecRRRRewrite
+	// useKindRDRewrite indicates an instruction where RD is used both as a source and destination.
```
anuraaga, line 216: atomic CAS has the same register behavior as vec rewrite. I tried to find a name and comment to clarify this but happy to change as needed

internal/engine/wazevo/testcases/testcases.go
```diff
@@ -2318,6 +2318,50 @@ var (
 			}}},
 		},
 	}
+
+	AtomicCasConst0 = TestCase{
```
anuraaga, line 2322: I added this test case to check the behavior regards to

https://github.com/tetratelabs/wazero/blob/main/internal/engine/compiler/impl_threads_arm64.go#L411

I am assuming `tmp := operandNR(m.compiler.AllocateVReg(si.Return().Type()))` and the `mov` is what ensures we don't run into zero register issues with this backend but let me know if that's not the case

mathetake:
will review next week!

mathetake (approved):
perfect...

PR #2040, opened 2024-02-11
fuzz: adds unaligned global f32 case

evacchi:
fails consistently on amd64, but it won't crash rosetta2 on macOS/Arm
EDIT: and unfortunately it's not crashing qemu either...

```
evacchi@lima-amd64:/Users/evacchi/Devel/github.com/tetratelabs/wazero/internal/integration_test/fuzzcases$ ./fuzz.test -test.v -test.run Test2040
=== RUN   Test2040
--- PASS: Test2040 (0.00s)
PASS
```


mathetake:
Cool thanks... this is weird...

mathetake:
I will take this from here. I managed to reproduce with EPYC-v3 CPU on qemu.

evacchi:
oh cool!

mathetake:
https://github.com/tetratelabs/wazero/pull/2042 temporarily allows passing this test case, but I will do the more fundamental fix later. anyway thank you for noticing this crash!

mathetake (approved):

PR #2042, opened 2024-02-11
wazevo: aligns globals at 16 byte

mathetake:


internal/engine/wazevo/module_engine.go
```diff
@@ -74,6 +74,17 @@ type (
...
 
+func newAlignedOpaque(size int) moduleContextOpaque {
+	// Check if the size is a multiple of 16.
+	if size%16 != 0 {
+		panic("size must be a multiple of 16")
+	}
+	type _16 [16]byte
+	buf := make([]_16, size/16)
+	slice := unsafe.Slice(&buf[0][0], size)
+	return *(*moduleContextOpaque)(&slice)
```
mathetake, line 85: @achille-roussel @ncruces teach me if there's a better way to align []byte at 16 bytes

ncruces: Tbh, I'm not entirely sure that ensues alignment…

mathetake: omg

ncruces: For up to 8 bytes there _are_ tricks, that work according to the spec (e.g. `_ [0]atomic.Int64` field in a struct forces alignment of the struct).

To ensure 16 byte alignment, you can heap alloc 32 bytes, but that's an [implementation detail](https://go.dev/src/runtime/mksizeclasses.go), not specified.

I guess we can over alloc and add padding ourselves? 🫤

There's an open [proposal](https://github.com/golang/go/issues/19057) to allow this.

mathetake: ok thanks, will get back to this tomorrow!

mathetake: https://github.com/tetratelabs/wazero/pull/2043 fixed

PR #2052, opened 2024-02-13
wazevo(arm64): support for atomic load/store

anuraaga:


internal/engine/wazevo/frontend/lower.go
```diff
@@ -3266,8 +3266,7 @@ func (c *Compiler) lowerCurrentOpcode() {
 
 			addr := c.memOpSetup(baseAddr, uint64(offset), size)
 			c.memAlignmentCheck(addr, size)
-			res := builder.AllocateInstruction().AsAtomicStore(addr, val, size).Insert(builder).Return()
-			state.push(res)
+			builder.AllocateInstruction().AsAtomicStore(addr, val, size).Insert(builder)
```
anuraaga, line 3269: Oops, the frontend test case didn't catch this since it didn't have any return values

mathetake (approved):

PR #2059, opened 2024-02-14
wazevo(ssa): load memory bounds directly when shared

anuraaga:


internal/engine/wazevo/frontend/lower.go
```diff
@@ -3829,13 +3829,25 @@ func (c *Compiler) getMemoryLenValue(forceReload bool) ssa.Value {
...
 		builder.InsertInstruction(loadBufSizePtr)
 
 		ret = loadBufSizePtr.Return()
 	} else {
 		load := builder.AllocateInstruction()
-		load.AsExtLoad(ssa.OpcodeUload32, c.moduleCtxPtrValue, c.offset.LocalMemoryLen().U32(), true)
+		if c.memoryShared {
+			lenOffset := builder.AllocateInstruction().AsIconst64(c.offset.LocalMemoryLen().U64()).Insert(builder).Return()
+			addr := builder.AllocateInstruction().AsIadd(c.moduleCtxPtrValue, lenOffset).Insert(builder).Return()
+			load.AsAtomicLoad(addr, 8, ssa.TypeI64)
```
anuraaga, line 3847: I'm a bit confused on whether `8` is the correct value here. I see that the length seems to be `uint64`

https://github.com/tetratelabs/wazero/blob/main/internal/engine/wazevo/module_engine.go#L48

But below it is a `ssa.OpcodeUload32`

internal/engine/wazevo/frontend/frontend_test.go
mathetake, line 1: why so many non shared cases got affected?

anuraaga: They're all shared cases. Probably it could have been better to do this PR before adding so many :P

internal/engine/wazevo/frontend/lower.go
mathetake: good point, and I should fix the below so that I could use the normal 64bit load. 8 sounds correct for this

mathetake:
is this an optimization or a fix for a bug?

anuraaga:
I think fix for possible bug, it's the wazevo version of the PR with this discussion

https://github.com/tetratelabs/wazero/pull/1927#issuecomment-1898583351

mathetake (approved):

Issue #2064, opened 2024-02-15
flaky test: TestPoll_Windows/poll_should_wait_for_the_given_duration


mathetake:
https://github.com/tetratelabs/wazero/actions/runs/7923603740/job/21633642739

Issue #1994, opened 2024-01-31
flaky test: wasi_snapshot_preview1/wasi_stdlib_unix_test Test_NonblockingFile


mathetake:
https://github.com/tetratelabs/wazero/actions/runs/7733181655/job/21084725277?pr=1993

PR #2068, opened 2024-02-18
wasi: fix flaky tests

evacchi:
I could not reproduce them because they are quite time-sensitive, this should address #2064, #1994.


imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.c
```diff
@@ -171,6 +171,9 @@ void main_nonblock(char* fpath) {
   ssize_t newLen = 0;
   while (newLen == 0) {
     newLen = read(fd, buf, sizeof(buf));
+    if strlen(buf) == 0 {
+      newLen = 0;
+    }
```
mathetake, line 176: comment on the rationale?

internal/sysfs/poll_windows_test.go
mathetake, line 1: sorry could you teach me what this changes?

imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.c
evacchi: oh right, I added a comment inline. If buf contains an empty string (like in the case of #1994) the length of the buffer might be 1 (because of the terminator `\0`) but we want to discard the element and stay in the loop.

internal/sysfs/poll_windows_test.go
evacchi: right, in the original version we were getting the underlying file handle at the beginning of the block and then using `syscall.Write()` directly; the flake was reporting an invalid file handle, which indicates a race where the handle has become invalid between the time it has been retrieved and the time it's being used. 

The rationale is that the handle is internal, so we don't control its lifecycle; instead we use the `*os.File` directly and pass it as a writer to `write(w, buf)` which internally just calls `Writer.Write()`; the idea is that this should handle more gracefully error situations.

I still suspect the flake has to do with timers going off too early and closing file handles, making them invalid, but I am trying to avoid relaxing the timers because then the test would take longer.

imports/wasi_snapshot_preview1/testdata/zig-cc/wasi.c
evacchi: (and apparently I also forgot to push the wasm)

internal/sysfs/poll_windows_test.go
evacchi: actually I am redrafting because I think this failure I can reproduce.

evacchi: it looks one of the issues might have been the file handles for os.Pipe() were not closed correctly in all tests, because now I can no longer reproduce.

mathetake (approved):

PR #2085, opened 2024-02-21
Changes Read to receive uint64 byte count.

ncruces:
Since #2074 changed the type of `Size() uint64` to we should probably change the type of `Read(offset uint32, byteCount uint64) ([]byte, bool)`. Otherwise, there's no API that gets entire buffer, if it covers the entire address range.

As discussed on [slack](https://gophers.slack.com/messages/C04CG4A2NKX/p1708414573291929?thread_ts=1708401526.868449).

### DIFF
--- a/api/wasm.go
+++ b/api/wasm.go
@@ -635,7 +635,7 @@ type Memory interface {
 	// shared. Those who need a stable view must set Wasm memory min=max, or
 	// use wazero.RuntimeConfig WithMemoryCapacityPages to ensure max is always
 	// allocated.
-	Read(offset, byteCount uint32) ([]byte, bool)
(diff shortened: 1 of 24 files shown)


mathetake:
looking good. can you proceed?

ncruces:
I got a bit stuck with the test failures, then work/life got in the way. Let me try again.

mathetake:
yeah cool, no rush anyway!

ncruces:
This will definitely cause more churn than #2074, but if we want to do it, v2 is the right time.

mathetake (approved):
@evacchi this must be explained in the release note together with https://github.com/tetratelabs/wazero/pull/2074 👍

Issue #2039, opened 2024-02-10
flaky test: TestFileCache_{compiler, wazevo}


mathetake:
We recently are having the flaky seg fault as in 

* https://github.com/tetratelabs/wazero/actions/runs/7853356149/job/21432882550?pr=2036
* https://github.com/tetratelabs/wazero/actions/runs/7856559936/job/21439463102

since the failure is happening for both wazevo and old compiler, I am pretty confident that 
this is something to do with life cycle issue around mmap/finalizer in file caches.

mathetake:
relevant: https://github.com/tetratelabs/wazero/pull/1815

evacchi:
Unrelated, but I am noticing that if the function returns early (line 36) then the finalizer is not set.

https://github.com/tetratelabs/wazero/blob/99c057bcb5f3f4311815c884ba7113c8d52b3316/internal/engine/compiler/engine_cache.go#L35-L54

As for the actual issue. I still don't know, but one thing that I notice is that before this commit the original code was a little more defensive see [here](https://github.com/tetratelabs/wazero/commit/9780f0f4a019ebf8c657fd74f65ce25e2046e507#diff-45ddcb29b654e977deee8ee51939c95ca6320460c241f4ace76116c316f43df6R544-R547) and [especially here](https://github.com/tetratelabs/wazero/commit/9780f0f4a019ebf8c657fd74f65ce25e2046e507#diff-45ddcb29b654e977deee8ee51939c95ca6320460c241f4ace76116c316f43df6L476-L485)

e.g. originally:

```
	e := cm.executable
	if e == nil {
		return // already released
	}

	// Setting this to nil allows tests to know the correct finalizer function was called.
	cm.executable = nil
	if err := platform.MunmapCodeSegment(e); err != nil {
		// munmap failure cannot recover, and happen asynchronously on the finalizer thread. While finalizer
		// functions can return errors, they are ignored.
		panic(fmt.Errorf("compiler: failed to munmap code segment: %w", err))
	}
```

and I wonder if the defensive check against nil was actually needed. WDYT @achille-roussel could this be related at all?

achille-roussel:
My understanding of this code is that we don't need to set the finalizer on values we got from the in-memory table of compiled modules because they have already been attached a finalizer if needed (e.g., if they were obtained from the cache).

[...]

PR #2088, opened 2024-02-23
filecache: use os.CreateTemp to create files

achille-roussel:
This might be a fix for #2039 

Since https://github.com/tetratelabs/wazero/pull/1816, the file cache appended `.tmp` to the file path name to avoid leaving incomplete files in the cache when an error occurs. However, this is the definition of `os.Create`:

```go
func Create(name string) (*File, error) {
	return OpenFile(name, O_RDWR|O_CREATE|O_TRUNC, 0666)
}
```

File creation does not include the `O_EXCL` flag, which means that when the file already exists, it is opened and truncated.

This PR changes the file cache to use `os.CreateTemp` instead, which guarantees that the file name will be unique by injecting a random string in the file name and opening the file with `O_EXCL` to prevent racing calls that would accidentally use the same name from both suceeding:

```go
func CreateTemp(dir, pattern string) (*File, error) {
	if dir == "" {
		dir = TempDir()
	}

	prefix, suffix, err := prefixAndSuffix(pattern)
	if err != nil {
		return nil, &PathError{Op: "createtemp", Path: pattern, Err: err}
	}
	prefix = joinPath(dir, prefix)

	try := 0
	for {
		name := prefix + nextRandom() + suffix
		f, err := OpenFile(name, O_RDWR|O_CREATE|O_EXCL, 0600)
		if IsExist(err) {
			if try++; try < 10000 {
				continue
			}
			return nil, &PathError{Op: "createtemp", Path: prefix + "*" + suffix, Err: ErrExist}
		}
		return f, err
	}
}
```

In #2039, we observe errors where the memory mapping containing the code was getting corrupted somehow. We first assumed that it was caused by a lifecycle issue on the compiled modules that hold these memory segments. The hypothesis I'm making with this change is that concurrent calls to `os.Create` for the same compiled module race each other; when the second call occurs, it truncates the content that was being written by the first call. After that, the program cannot rely on the content of the file anymore.

I don't have a test to verify the hypothesis, but regardless, I believe this change is useful to make because it addresses a potential race condition that will manifest sooner or later.

### DIFF
--- a/internal/filecache/file_cache.go
+++ b/internal/filecache/file_cache.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"sync"
 )
 
 // New returns a new Cache implemented by fileCache.
@@ -24,12 +23,6 @@ func newFileCache(dir string) *fileCache {
 // Note: this can be expanded to do binary signing/verification, set TTL on each entry, etc.
 type fileCache struct {
 	dirPath string
-	mux     sync.RWMutex
-}
-
-type fileReadCloser struct {
-	*os.File
-	fc *fileCache
 }
 
 func (fc *fileCache) path(key Key) string {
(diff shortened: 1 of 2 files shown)


achille-roussel:
I submitted a second commit after realizing that we could get rid of the in-memory synchronization in the file cache and rely exclusively on the atomicity guarantees offered by the file system. Besides the code simplification, relying on the file system also offers synchronization that works when multiple processes share the same cache directory.

achille-roussel:
Hardening of the file format sounds great to me, I would add the checksum even if it's more compute, the in-memory cache should amortize that cost quite effectively, something like CRC would probably be good enough?

That being said, this PR would still fix the problem as-is since it makes the insertion of files in the cache an atomic operation, so there is never a conditions where opening a file would cause the truncation of an existing one.

internal/filecache/file_cache.go
```diff
@@ -23,55 +23,28 @@ func newFileCache(dir string) *fileCache {
...
-	// TODO: take lock per key for more efficiency vs the complexity of impl.
-	fc.mux.Lock()
-	defer fc.mux.Unlock()
-
-	// Use rename for an atomic write
 	path := fc.path(key)
-	file, err := os.Create(path + ".tmp")
+	dirPath, fileName := filepath.Split(path)
+
+	file, err := os.CreateTemp(dirPath, fileName+".*.tmp")
```
ncruces, line 47: Can we do:

```go
	defer func() {
		if err != nil {
			_ = file.Close()
			_ = os.Remove(file.Name())
		}
	}()
```

And remove the `defer file.Close()`?
I'm not sure on Windows `os.Remove(…)` will work before `f.Close()` (and `defer`s run in reverse order).

ncruces (approved):
If we want to add a checksum, Castagnoli has hardware support in [`hash/crc32`](https://pkg.go.dev/hash/crc32).

But this PR should also fix the race, as long as no one corrupts the cache on purpose.

internal/filecache/file_cache.go
achille-roussel: I made the change in [52d6eb1](https://github.com/tetratelabs/wazero/pull/2088/commits/52d6eb198af742efe98d196c0b9b7ace0935dc5a)

I made an adjustment to close the file regardless of whether an error occurred.

PR #2090, opened 2024-02-24
wazevo(ssa): implements TailDuplication pass

mathetake:


mathetake:
found a bug in immediate dominator calculations:

```
(module
  (type (;0;) (func (param i32) (result i32)))
  (func (;0;) (type 0) (param i32) (result i32)
    loop (result i32) ;; label = @1
      block ;; label = @2
        local.get 0
        br_table 1 (;@1;) 0 (;@2;) 0 (;@2;)
      end
      i32.const 0
    end
    local.set 0
    loop (result i32) ;; label = @1
      block ;; label = @2
        local.get 0
        br_table 0 (;@2;) 1 (;@1;) 1 (;@1;)
      end
      i32.const 0
    end
  )
)
```

I need to fix this first

PR #2097, opened 2024-02-27
validation: returns error on unexpected end of function

mathetake:


mathetake:
found in https://github.com/tetratelabs/wazero/actions/runs/8057956650/job/22009991155

PR #2100, opened 2024-02-27
sysfs: fix compile-time check on osFile

karelbilek:
This is a tiny fix of a compile-time check on osFile.

It's not like it _matters_ all that much, as the missing method would be caught in other layers anyway, but when the check is there, it should be correct :)


karelbilek:
ugh DCO. OK

evacchi (approved):
ha! good eye!

PR #2103, opened 2024-02-29
wazevo(frontend): store module context for only imported calls

mathetake:
This cuts the unnecessary one store instruction per local function call, hence results in the executable size reduction (around 50kb)!


mathetake:
easy win!

PR #2105, opened 2024-03-01
spectest: ensure compiled functions can outlive module

ncruces:
Tentative fix for #2039 for the `compiler`, not `wazevo`.

In [`linking.wast`](https://github.com/tetratelabs/wazero/blob/0b0a23c28f52ee16d102d14453af3a4e4316f98c/internal/integration_test/spectest/v2/testdata/linking.wast#L421-L453) we have:
- a module `A` that exports its memory and table
- a module `B` that imports both
  - it initializes the (shared) memory, and 
  - it initializes the (shared) table, setting entry zero to one of its functions
  - then it fails while running `main`

The test then verifies that initializing the memory and the table _both_ succeeded _before_ main failed. It does so by reading from the memory, and **indirect calling** the function in the table.

Both the memory and table are owned by `A`, so still exist. But the function, or rather the compiled code (and `mmap`ed memory) for the function is/was owned by `B` and can be GCed, because we `SetFinalizer` on an object owned by `B`.

This change makes it so that the finalizer is set on the `compiledCode` rather than on the `compiledModule`. Since the [resolved function import](https://github.com/tetratelabs/wazero/blob/0b0a23c28f52ee16d102d14453af3a4e4316f98c/internal/engine/compiler/engine.go#L659-L663) is a `function`, and that owns a `compiledFunction`, which owns the `compiledCode`, this should fix the issue.

### DIFF
--- a/internal/engine/compiler/engine.go
+++ b/internal/engine/compiler/engine.go
@@ -498,9 +498,9 @@ func (s nativeCallStatusCode) String() (ret string) {
 	return
 }
 
-// releaseCompiledModule is a runtime.SetFinalizer function that munmaps the compiledModule.executable.
-func releaseCompiledModule(cm *compiledModule) {
-	if err := cm.executable.Unmap(); err != nil {
+// releaseCompiledCode is a runtime.SetFinalizer function that munmaps the compiledCode.executable.
+func releaseCompiledCode(code *compiledCode) {
+	if err := code.executable.Unmap(); err != nil {
 		// munmap failure cannot recover, and happen asynchronously on the
 		// finalizer thread. While finalizer functions can return errors,
 		// they are ignored.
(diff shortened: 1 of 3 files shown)


achille-roussel (approved):
The change looks good, do you want to merge this in independently of https://github.com/tetratelabs/wazero/pull/2102 ?

ncruces:
> The change looks good, do you want to merge this in independently of #2102 ?

I'd reviewed a previous version of that, and (I guess because the title was not updated) totally missed the progress.
Then went on and duplicate all the work. 🤦

Happy to shelve this. I guess there's only the test that I fixed, but @evacchi can easily fix it.

ncruces:
Everything is already in #2102.

Issue #1852, opened 2023-11-29
api.Memory.Size() overflows when the memory has the maximum 65536 pages


EclesioMeloJunior:
**Describe the bug**
Basically, when calling `memory.Grow` function passing a delta that grows the memory to the `MemoryLimitPages` (65536) returns successfully but when we call `memory.Size()` the number of pages mismatches the expected.

I think that occurs because uint32 cannot fit the new buffer len (since the current page is `65536` and the number of bytes per page is `65536` and the multiplication of these values overflows uint32)

**To Reproduce**
I've wrote a test that reproduces the error in this fork https://github.com/ChainSafe/wazero/pull/4

**Expected behavior**
Since the limit is `MemoryLimitPages` we should be able get the correct page size when growing up to that size

**Screenshots**
N/A

**Environment (please complete the relevant information):**
- Go version: 1.21.3
- wazero Version: 6a7e474e022e5d325261cf3073eb246cf4fb4238
- Host architecture: arm64
- Runtime mode:

**Additional context**

mathetake:
Thank you @EclesioMeloJunior for reporting this, and I confirmed that this is actually a bug in our API. As you described, when the memory has the maximum of 65536 pages, the buffer length becomes 2^32, which is larger than the maximum of uint32 which is the return type of api.Memory Size.

We could do either 

1. Changing the return type of api.Memory Size to uint64
    - This is somewhat breaking change, though the impact should be really minimum because this is just a integer type conversion.
2. Adds another for `SizeSafe()` or whatever it is while putting deprecation comments on the api.Memory Size.
    - This won't be a breaking change 


thoughts? @evacchi @ncruces @achille-roussel

mathetake:
> what about checking in the Grow method for this limit? I mean, instead of allowing the buffer size to reach 2^32 (which overflows uint32 only by 1) check and limit the buffer size/cap to be (2^32) - 1 then it will bound to uint32 limits (and the Size will not have problems to calc/return the correct size). I would assume that the problem with this approach is that the latest page will lose 1 byte (65535 instead of 65536), right?

no that's not our option as that means wazero is not compliant with the Wasm spec.

EclesioMeloJunior:
@mathetake @ncruces thanks for the clarifications, I took the freedom to create a draft PR that introduces the new method https://github.com/tetratelabs/wazero/pull/1856

mathetake:
[...]

PR #2106, opened 2024-03-02
api: reverts the uint32->uint64 breaking change for 2.0 release

mathetake:
This commit reverts the only/first ever breaking change we committed 
towards 2.0 release, #2085 and #2074.

These commits were made to fix a bug in our API which is described well in
https://github.com/tetratelabs/wazero/issues/1852 but the reality is that it will 
cause users to modify their existing code when they upgrade to 2.0 whose main
focus is to introduce the optimizing compiler just by updating wazero, not to fix
the bug. 

Note that this doesn't mean that we won't make any breaking change never, but this 
is more about dedicating 2.0 to the optimizing compiler, not fixing the bug which 
is highly likely that existing users will never encounter.


mathetake:
cc @ncruces

PR #2109, opened 2024-03-04
wazevo(ssa): recursive constant phi eliminations

mathetake:


mathetake:
This makes the entire perf really better like 2x faster in some cases. Especially, the sqlite speedtest executable size is like 30% smaller... amazing...

mathetake:
fuzzing time

PR #2117, opened 2024-03-05
wazevo(frontend): stricter BCE

mathetake:
## sqlite speedtest binary size
amd64: 8478767 bytes on main.
amd64: 8337167 bytes with predecessor check.
amd64: 8345407 bytes without predecessor check.

arm64: 8624468 on main.
arm64: 8465156 bytes with predecessor check.
arm64: 8474788 bytes without predecessor check.

### result
amd64: 8345407 - 8337167 = 141600 bytes: ~1.6% of 8.3MB
arm64: 8624468 - 8465156 = 159312 bytes: ~1.8% of 8.6MB

## zig stdlib binary size

amd64: 47128091 bytes on main.
amd64: 45244091 bytes with predecessor check.

arm64: 48182088 bytes on main.
arm64: 46195880 bytes with predecessor check.

### result
amd64: 47128091 - 45244091 = 1884000 bytes: ~4.0% of 47MB
arm64: 48182088 - 46195880 = 1986208 bytes: ~4.1% of 48MB

### DIFF
--- a/internal/engine/wazevo/frontend/frontend.go
+++ b/internal/engine/wazevo/frontend/frontend.go
@@ -203,6 +203,7 @@ func (c *Compiler) Init(idx, typIndex wasm.Index, typ *wasm.FunctionType, localT
 	c.wasmFunctionBody = body
 	c.wasmFunctionBodyOffsetInCodeSection = bodyOffsetInCodeSection
 	c.needListener = needListener
+	c.clearSafeBounds(true)
 }
 
 // Note: this assumes 64-bit platform (I believe we won't have 32-bit backend ;)).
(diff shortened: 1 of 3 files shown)


mathetake:
https://github.com/tetratelabs/wazero/pull/2118 will fix the fuzzing failure, which is not relevant to this

internal/engine/wazevo/frontend/frontend.go
```diff
@@ -438,12 +442,21 @@ func (c *Compiler) recordKnownSafeBound(v ssa.ValueID, safeBound uint64, absolut
...
+		for _, v := range c.knownSafeBoundsSet {
+			ptr := &c.knownSafeBounds[v]
+			ptr.bound = 0
+		}
+		c.knownSafeBoundsSet = c.knownSafeBoundsSet[:0]
+	} else {
+		for _, v := range c.knownSafeBoundsSet {
+			ptr := &c.knownSafeBounds[v]
+			ptr.absoluteAddr = ssa.ValueInvalid
+		}
```
achille-roussel, line 458: I would recommend splitting this method in two instead of passing a boolean, especially since it seems that we always call it with a constant true or false:

- the method body clearly shows it's doing two things, that's a good indication that it would be better suited to have the logic in two methods

- at the call site, true/false says nothing about the meaning of the literal; this can be improved if the two methods were named `clearSafeBounds` and `clearAbsoluteAddresses`

mathetake: agreed, thank you for the third eye!

PR #2119, opened 2024-03-05
wazevo(frontend): stricter BCE part2

mathetake:
This reduces the executable roughly 0.5%~0.7% for sqlite and zig stdlib binary.

### Zig stdlib executable size

arm64: 46195880 --> 45877240: 318640 bytes
amd64: 45244091 --> 44953739: 290352 bytes

### Sqlite speedtest executable size
arm64: 8465316 --> 8419684: 45632 bytes
amd64: 8337167 --> 8295535: 41632 bytes


mathetake:
fuzzing time

PR #2120, opened 2024-03-05
Handle negative grow pages to avoid shrinking memory

mathetake:
This has existed for the very long time but thanks to the wazevo's aggressive bounds check
optimization, this has surfaced just now through fuzzing. I would say this is one of the most
terrible (yet not harmful in practice) bugs in the history of wazero...


internal/wasm/memory.go
```diff
@@ -256,7 +256,7 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 
 	// If exceeds the max of memory size, we push -1 according to the spec.
 	newPages := currentPages + delta
-	if newPages > m.Max {
+	if newPages > m.Max || int32(delta) < 0 {
```
ncruces, line 259: Wouldn't `newPages > m.Max || newPages < currentPages` be clearer, or am I misunderstanding it?

mathetake: yeah clearer

PR #2134, opened 2024-03-07
fuzz: enables threads, fixes widening v128.load

mathetake:


internal/engine/wazevo/frontend/lower.go
```diff
@@ -1840,7 +1840,7 @@ func (c *Compiler) lowerCurrentOpcode() {
 			baseAddr := state.pop()
 			addr := c.memOpSetup(baseAddr, uint64(offset), 8)
 			load := builder.AllocateInstruction().
-				AsLoad(addr, offset, ssa.TypeV128).
+				AsLoad(addr, offset, ssa.TypeF64).
```
mathetake, line 1843: This load should only reads the 8 bytes, and then widens it to a full v128 below. Before, loads the entire vector, which might go beyond the memory bound (they are semantically equivalent and harmless tho)

PR #2140, opened 2024-03-08
chore: ensure all tinygo binaries have been built with 0.31.1

evacchi:


Makefile
```diff
@@ -52,12 +52,13 @@ build.examples.as:
...
 	@cd internal/testing/dwarftestdata/testdata/zig; zig build; mv zig-out/*/main.wasm ./ # Need DWARF custom sections.
 
-tinygo_sources := examples/basic/testdata/add.go examples/allocation/tinygo/testdata/greet.go examples/cli/testdata/cli.go imports/wasi_snapshot_preview1/example/testdata/tinygo/cat.go imports/wasi_snapshot_preview1/testdata/tinygo/wasi.go
+tinygo_sources := examples/basic/testdata/add.go examples/allocation/tinygo/testdata/greet.go examples/cli/testdata/cli.go imports/wasi_snapshot_preview1/example/testdata/tinygo/cat.go imports/wasi_snapshot_preview1/testdata/tinygo/wasi.go cmd/wazero/testdata/cat/cat.go
 .PHONY: build.examples.tinygo
 build.examples.tinygo: $(tinygo_sources)
 	@for f in $^; do \
 	    tinygo build -o $$(echo $$f | sed -e 's/\.go/\.wasm/') -scheduler=none --no-debug --target=wasi $$f; \
 	done
+	@mv cmd/wazero/testdata/cat/cat.wasm cmd/wazero/testdata/cat/cat-tinygo.wasm
```
mathetake, line 61: instead of adding only one exception, can we change the user side of this binary to `cat.wasm` in wazero_test.go?
```go
//go:embed testdata/cat/cat-tinygo.wasm
var wasmCatTinygo []byte
```

evacchi: sure, but there is both a cat-tinygo.wasm and a cat-go.wasm, that's why I kept the exception

mathetake: ok then let's change the source file name?

mathetake: aha now I get it

PR #2144, opened 2024-03-11
wazevo: support for experimental snapshotting

mathetake:


mathetake:
pgquery tests are crashing so it's draft now

PR #2152, opened 2024-03-12
fuzz: adds --no-trace-compares flag by default

mathetake:
This speeds up the fuzzing drastically at the cost of worse coverage (I think we don't care).


mathetake:
FYI @evacchi

PR #2157, opened 2024-03-24
mmap memory when it never grows

ncruces:
As discussed on Slack. Both branches of the if are very similar, but I just wanted to make sure this doesn't break anything before going further.


evacchi:
link to the discussion for reference https://gophers.slack.com/archives/C04CG4A2NKX/p1711139148569809

internal/wasm/memory.go
```diff
@@ -76,25 +77,33 @@ func NewMemoryInstance(memSec *Memory) *MemoryInstance {
...
+			b, err := platform.MmapMemory(int(max))
 			if err != nil {
 				panic(fmt.Errorf("unable to mmap memory: %w", err))
 			}
 			mmappedBuffer = b
+			buffer = b[:min]
 		} else {
 			// mmap not supported so we just preallocate a normal buffer. This will often be large, i.e. ~4GB,
-			// and likely isn't practical, but interpreter usage should be rare and the Wasm binary can be
+			// and likely isn't practical, but interpreter usage should be rare and the WASM binary can be
```
evacchi, line 89: "Wasm" is the right capitalization

ncruces: Ah, OK. 😬

ncruces:
This turns out to be a simple and pretty much painless change.

In light of comments on Slack we may prefer instead to create allow configuring a memory allocator for each runtime (roughly, that's @achille-roussel's suggestion) or provide a memory buffer on instantiation (@anuraaga's).

Both APIs are more explicit, which is good, IMO, but are a bigger change. I'm also not so sure if they can start as experimental, without further complication.

This is enough for my purposes/experimentation. I can leave it as a draft while we discuss alternatives, and I test my usage of it (I need to `mmap` files into guest memory). Or I can withdraw, if you don't like the PR hanging.

If we coalesce around an API with like to offer, I'm happy to implement.

ncruces:
Marking ready for review since I'm now confident that I need something like this. I don't mind at all if we decide we should do it another way.

anuraaga:
I don't see a problem with this for long lived modules. But if anyone is using WithCapacityFromMax in a e.g. module per request type of situation, all modules will allocate using syscalls rather than the Go GC which is a big change and not obviously for the better. I don't know if anyone does this but it doesn't seem like a bad thing to do. I'd feel more comfortable with an experimental flag to enable it

ncruces:
So, summary of options:
1. this;
2. similar to this but with an (experimental) flag rather than `WithMemoryCapacityFromMax`;
3. an (experimental) `RuntimeConfig` API to configure a memory allocator for the runtime;
4. an (experimental) `ModuleConfig` API to configure a memory buffer for the module.

I'm OK with any such option. I'm sensitive to the arguments against this (observability, performance for short lived modules), but these should apply similarly to shared memory. We could just as easily use Go memory to back shared memory.

anuraaga:
> We could just as easily use Go memory to back shared memory.

I don't think so since there is generally an expectation that a process only uses as much physical memory as is currently used, but the nature of thread safety requires either mmap to allocate virtually, or very complex stop-the-world logic in the compiler. Wasm shared memory requiring a specification for max is basically a crutch that it is expected all implementations mmap the max.

Lack of shared memory / atomics is one reason we have the module per request model in the first place and generally I'm comfortable assuming shared memory is in practice used for long lived modules. This may also be the case for the WithCapacityAsMax flag, I just don't have as much experience or context on it.

ncruces:
> I don't think so since there is generally an expectation that a process only uses as much physical memory as is currently used, but the nature of thread safety requires either mmap to allocate virtually, or very complex stop-the-world logic in the compiler. Wasm shared memory requiring a specification for max is basically a crutch that it is expected all implementations mmap the max.

I'd say thread safety only requires the memory to never move (which is the exact same requirement conveyed by `WithMemoryCapacityFromMax`). If the shared memory has a reasonable maximum size (say 64MB), we could prefer (for observability) in a short-lived module, to just `make([]byte, 64*1024*1024)`.

ncruces:
Not to derail the PR, I'm not sure if it is such an issue to do `make([]byte, fourGigabytesOfRAM)` (for shared memory) in the OSes we care about.

That said, my vote would go for option **3**, but I could live with **2** (even **1**, as a stopgap).

anuraaga:
option 3 sgtm

mathetake:
https://github.com/tetratelabs/wazero/pull/2170 obviates this PR, but discussion continues in slack!

PR #2159, opened 2024-03-25
compiler: use build tags instead of runtime.GOARCH to select ISA

orsinium:
Select the right ISA in `isa.go` statically using build tags instead of a runtime switch on `runtime.GOARCH`. It makes the build faster and is required for building wazero with tinygo.


orsinium:
> Unfortunate that this is needed, but fine if so.

Using build tags instead of switches better matches the rest of the codebase. The only other places with `switch runtime.GOARCH` are in tests. I think it's nice to consistently use build tags especially because it makes both runtime and compilation faster.

> TinyGo doesn't resolve runtime.GOOS at compile time?

Perhaps, dead code elimination happens after type checking. I don't know much about how TinyGo works. I simply run `tinygo build` on the project and see what explodes.

evacchi (approved):
I think this should be fine 🤔

Issue #1854, opened 2023-11-30
wazero running on embedded systems


deadprogram:
I would like to be able to do something like WAMR https://github.com/bytecodealliance/wasm-micro-runtime for embedded systems, but to use wazero as the code/runtime to do this.

Ideal scenario would be that I would be able to run:
```
cd /path/to/repo/wazero-micro
tinygo build -o wazero-micro.uf2 -target itsybitsy-m4 .
```
and get a binary I can then flash onto that MCU with the wazero runtime. The binary would contain a minimal version of the wazero runtime.

Pretty sure this could be done with WAMR, but it requires also using an RTOS like Zephyr. Wazero would be running bare metal on the hardware e.g. "The Go runtime is your RTOS"
Also, it would be great to have a pure Go stack able to do this without needing any other languages/compilers.

Specific hardware functionality can be exposed to WASM modules via host functions.
Deciding on how to load new WASM code onto the device can be part of the host program, for example via serial port/OTA updates/etc.

evacchi:
A few days ago I ran the same thought experiment. Wazero already builds fine to wasm using the plain Go compiler (and then you can even run wazero-on-wazero in interpreter mode).

I think the bulk of the work is in the syscalls (which are mostly in the WASI lib implementation). e.g. your command fails with

      ❯ tinygo build -o wazero-micro.uf2 -target itsybitsy-m4
      # github.com/tetratelabs/wazero/internal/platform
      internal/platform/mmap_unix.go:42:22: undefined: syscall.Syscall
      internal/platform/mmap_unix.go:42:38: undefined: syscall.SYS_MPROTECT

evacchi:
hey this is awesome!

evacchi:
using this issue to track the related PRs

- https://github.com/tetratelabs/wazero/pull/2159
- https://github.com/tetratelabs/wazero/pull/2160
- https://github.com/tetratelabs/wazero/pull/2161
- https://github.com/tetratelabs/wazero/pull/2162
- https://github.com/tetratelabs/wazero/pull/2163

deadprogram:
We made it! Thank you to @orsinium @evacchi @achille-roussel and everyone else who helped make this happen.

I will close this issue now. We can open smaller more focused individual issues for anything else that we want to work on related to this area.

Thanks again!

paralin:
This is awesome! Thanks for the incredible work!

ncruces:
[...]

PR #2160, opened 2024-03-25
wasi: support building internal/sysfs with TinyGo

orsinium:
Support for building internal/sysfs package with TinyGo. The changes are sufficient to compile the sysfs package for use in cmd/wazero and for [mechanoid](https://github.com/hybridgroup/mechanoid). More changes are needed to compile other packages, I'll deliver it in follow-up PRs.

I didn't implement some of the methods for tinygo version, like `dirFs.Link`. Somehow it compiles without them. Probably, they are not used by the relevant code. I'll provide the implementation later if needed.

related: https://github.com/tetratelabs/wazero/issues/1854


evacchi:
Maybe we should add a CI job to build against tinygo so that we can validate this. e.g. at the beginning we could just build the packages we know that work; at the end of the patch series we delete the package list and just `tinygo build` 

e.g. if I build this I am currently getting a few errors:

```
❯ tinygo build ./internal/sysfs/.
# github.com/tetratelabs/wazero/experimental/sys
experimental/sys/syscall_errno.go:19:15: undefined: syscall.EBADF
experimental/sys/syscall_errno.go:23:15: undefined: syscall.EFAULT
experimental/sys/syscall_errno.go:29:15: undefined: syscall.EIO
experimental/sys/syscall_errno.go:33:15: undefined: syscall.ELOOP
```

/cc @ncruces WDYT

ncruces:
Agree that some validation would be interesting.

orsinium:
Is it possible to build a non-main package? 🤔 

```
$ tinygo build ./internal/sysfs/.
panic: expected main package to have name 'main'

goroutine 1 [running]:
...
	/__w/tinygo/tinygo/main.go:1586 +0x368d
```

orsinium:
I'm going to deliver the rest of the changes very soon anyway, so we will be able to build with tinygo the whole project at once, without bothering with per-package builds. Running tests on tinygo will require much more effort, though, like using [tinytest](https://github.com/orsinium-labs/tinytest).

evacchi:
running the tests might be more challenging, but at least having a `tinygo build` in ci might be worth it

orsinium:
I've added TinyGo build on CI, see  #2163.

internal/sysfs/dirfs_go.go
evacchi, line 1: this naming convention looks sort-of weird to me, because "_go" should be the default. I understand the reason for doing it, but it's the first time I see it. @ncruces, @achille-roussel WDYT?

we have often adopted the `<base>_unsupported.go` convention for platforms that do not support a particular feature. See also the comment for sockets.

internal/sysfs/sock_go.go
```diff
@@ -0,0 +1,56 @@
+//go:build !tinygo
```
evacchi, line 1: will sockets be used at all in the tinygo version? If they won't maybe it's worth keeping `sock.go` the default (`//go:build !tinygo`) and then stub everything in `sock_tinygo.go` (`//go:build tinygo`)

I am afraid all these bifurcated code paths might become confusing

Notice that there is already a `sock_unsupported.go` with heading:

```
//go:build !linux && !darwin && !windows
```

maybe it's enough to add `|| tinygo` there?

deadprogram: > will sockets be used at all in the tinygo version?

I hope so, yes.

deadprogram: > If they won't maybe it's worth keeping sock.go the default (//go:build !tinygo) and then stub everything in sock_tinygo.go (//go:build tinygo)

That sounds like a good idea, since I do not know timeframe for the sockets support.

internal/sysfs/dirfs_go.go
ncruces: I've never seen that convention, if there is one.

deadprogram:
OK, so hopefully to summarize:

- rename `dirfs_go.go` to `dirfs_link_supported.go`
- add a file `dirfs_link_unsupported.go` for the tinygo build tag
- rename `sock_go.go` back to `sock.go`
- we already have `sock_tinygo.go` so no change needed there.

Is that correct?

evacchi:
yep that sounds good

deadprogram:
@evacchi we cannot just rename `sock_go.go` to `sock.go` as that file already exists, as pointed out to me by @orsinium 

How about we instead, we 
- rename `sock_go.go` to `sock_big.go`
- rename `sock_tinygo.go` to `sock_tiny.go`

orsinium:
About the "sock" naming (to add context on Ron's message above): "rename sock_go.go back to sock.go" is not right because there is already "sock.go". Putting sock_go.go contents back into sock.go makes these methods to be double implemented when building with TinyGo. And if I exclude socks.go from tinygo, then the rest of the socks.go content is reported as not implemented. The option that Eduardo suggested to bootstrap everything in sock_unsupported.go increases the amount of code a lot and makes sockets for unsupported platfroms to be not implemented at all instead of being implemented partially, which changes the existing logic. To preserve the same behavior, I'd have to literally copy-paste everything from sock.go into sock_unsupported.go, which is about 200 lines of code.

The decision we made to change `_go.go` and `_tinygo.go` to `_big.go` and `_tiny.go`. We consider refactoring the structure of these files later, when we work on the actual sockets support for TinyGo. For now, the goal is to have at least "wasm-unknown" modules working with tinygo-compiled wazero.

deadprogram:
@evacchi anything else needed for this PR?

evacchi:
please check this commit https://github.com/evacchi/wazero/commit/15c25d01ec82e1c6ca08a4f0460c5f174d5e86c0 
This approach won't require adding any new files for sockets, it just shuffles a few things around. Also, notice that `Link` no longer panics but it just returns ENOSYS. 

If that's ok I can push it to this PR, or you can cherry-pick. Double-check if that works for you.

deadprogram:
@evacchi that looks good! You can push it to this PR and then merge. Thank you!

evacchi:
looks like I can't just push here, you can merge this PR @orsinium  https://github.com/orsinium-forks/wazero/pull/2

orsinium:
@evacchi I've merged your PR and added you and @deadprogram to the fork repo. Since I use an organization for forks, Github doesn't allow upstream maintainers to push into PR branches. That's an interesting quirk of Github.

deadprogram:
@evacchi this branch now includes https://github.com/evacchi/wazero/commit/15c25d01ec82e1c6ca08a4f0460c5f174d5e86c0 so it should be ready. Thanks @orsinium and @evacchi !

evacchi (approved):
thanks!

PR #2161, opened 2024-03-25
refactor reflect.SliceHeader uses to allow tinygo cross-compilation

orsinium:
The `Len` and `Cap` fields of `reflect.SliceHeader` are `int` in Go and `uintptr` in TinyGo. The PR adjusts all places that set these fields to be compatible with both Go and TinyGo.

Related issue: https://github.com/tetratelabs/wazero/issues/1854


ncruces:
Do we need this now that we floor on 1.20?
I've been meaning to do a pass on APIs like [`unsafe.SliceData`](https://pkg.go.dev/unsafe#SliceData).

orsinium:
> Do we need this now that we floor on 1.20?

What does it mean? Do you ask me? Sorry, I'm out of the loop. I build my changes on top of the latest `main` branch.

> I've been meaning to do a pass on APIs like [unsafe.SliceData](https://pkg.go.dev/unsafe#SliceData).

It would be great to get rid of it eventually since it is deprecated but I'd be glad if we can get Wazero compiling to TinyGo without blocking it with refactoring. The fork we have for TinyGo is already far behind the upstream, and keeping it up-to-date is hard.

ncruces:
> What does it mean? Do you ask me? Sorry, I'm out of the loop. I build my changes on top of the latest `main` branch.

wazero has a [policy](https://github.com/tetratelabs/wazero/blob/253c0346e4c8d7bd604304447233c1a3ce0a71f7/go.mod#L3C1-L5C1) of supporting current Go version minus 2. Today, that means Go 1.20.

`unsafe.SliceData` was introduced in 1.20. It couldn't be used when the `reflect.SliceHeader` code was originally written, but it can now.

Does TinyGo support `unsafe.SliceData`? If so, that's the correct change. I don't mind doing that change myself if it helps you.

internal/wasm/memory.go
```diff
@@ -271,7 +271,7 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 			// Use atomic write to ensure new length is visible across threads.
 			atomic.StoreUintptr((*uintptr)(unsafe.Pointer(&sp.Len)), uintptr(MemoryPagesToBytesNum(newPages)))
 		} else {
-			sp.Len = int(MemoryPagesToBytesNum(newPages))
+			sp.Len = lengthMemoryPages(newPages)
```
ncruces, line 274: Here I'd rather do:

```go
m.Buffer = m.Buffer[:MemoryPagesToBytesNum(newPages)]
```

And move the `sp` declaration inside the `m.Shared` branch.

internal/engine/wazevo/backend/isa/amd64/stack.go
```diff
@@ -79,8 +77,7 @@ func GoCallStackView(stackPointerBeforeGoCall *uint64) []uint64 {
 	{
 		sh := (*reflect.SliceHeader)(unsafe.Pointer(&view))
 		sh.Data = uintptr(unsafe.Pointer(stackPointerBeforeGoCall)) + 8 // skips the(sliceSize.
-		sh.Len = int(size)
-		sh.Cap = int(size)
+		setSliceLimits(sh, uintptr(size))
```
ncruces, line 80: Here it should be OK to do:

```go
data := unsafe.Pointer(uintptr(unsafe.Pointer(stackPointerBeforeGoCall)) + 8)
size := *stackPointerBeforeGoCall / 8
return unsafe.Slice((*uint64)(data), int(size))
```

ncruces (approved):
OK… so, I'm sorry.

I see that in most places were using the deprecated `reflect.SliceHeader` to skirt around a `go vet` warning: _possible misuse of unsafe.Pointer_.

So we need to pick our poison. If we do stick with `reflect.SliceHeader`, we should remove the TODOs for updating to `unsafe.Slice`.

I'm still pretty sure this is skirting around the rules, not fixing anything, otherwise this would be a fine way to avoid the _"you're reviving a `uintptr` which might break the GC and you really shouldn't"_ problems we were having.

Anyway, this is a longer, hairier, discussion which shouldn't block this PR.

orsinium:
@ncruces applied code suggestions ✅

evacchi:
since `reflect_{go,tinygo}.go` are new I think we can rename `reflect_go.go` to `reflect.go` so that it is clearer that's the default

orsinium:
Applied suggestion ✅

ncruces (approved):

evacchi (approved):

evacchi:
merging this, but I am not sure why tinygo is reaching this sub-package at all. It would be worth investigating if it's possible to drop the entire surface of the compiler.

PR #2162, opened 2024-03-26
internal: skip platform-specific code when building with TinyGo

orsinium:
Use build tags to skip some platform-specific Go code when building the project using TinyGo.

Together with my other recent PRs into the project, this one completes the set of changes required to build the project with TinyGo. When all of them are merged, I'll make sure to do one final check that it works as intended and then I'll add the TinyGo builds on CI.

Related open PRs: #2161 and #2160

### DIFF
--- a/internal/engine/wazevo/entrypoint_amd64.go
+++ b/internal/engine/wazevo/entrypoint_amd64.go
@@ -1,3 +1,5 @@
+//go:build amd64 && !tinygo
+
 package wazevo
 
 import _ "unsafe"
--- a/internal/engine/wazevo/entrypoint_arm64.go
+++ b/internal/engine/wazevo/entrypoint_arm64.go
@@ -1,3 +1,5 @@
+//go:build arm64 && !tinygo
+
 package wazevo
 
 import _ "unsafe"
--- a/internal/engine/wazevo/entrypoint_others.go
+++ b/internal/engine/wazevo/entrypoint_others.go
@@ -1,4 +1,4 @@
-//go:build !arm64 && !amd64
+//go:build (!arm64 && !amd64) || tinygo
 
 package wazevo
 
--- a/internal/platform/cpuid_amd64.go
+++ b/internal/platform/cpuid_amd64.go
@@ -1,3 +1,5 @@
+//go:build amd64 && !tinygo
+
 package platform
 
 // CpuFeatures exposes the capabilities for this CPU, queried via the Has, HasExtra methods
--- a/internal/platform/cpuid_unsupported.go
+++ b/internal/platform/cpuid_unsupported.go
@@ -12,7 +12,3 @@ func (c *cpuFeatureFlags) Has(cpuFeature CpuFeature) bool { return false }
 
 // HasExtra implements the same method on the CpuFeatureFlags interface
 func (c *cpuFeatureFlags) HasExtra(cpuFeature CpuFeature) bool { return false }
-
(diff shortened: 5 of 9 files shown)


internal/engine/wazevo/entrypoint_amd64.go
```diff
@@ -1,3 +1,5 @@
+//go:build amd64 && !tinygo
```
ncruces, line 1: I dunno what style this project favours (or if this is part of a more general style guide), but duplicating the tag from the filename is unnecessary.

ncruces (approved):

internal/engine/wazevo/entrypoint_amd64.go
deadprogram: Adding the tag is to make sure the TinyGo compiler is able to completely optimize away everything it can. Every byte counts on such constrained devices.

ncruces: Not sure I understood. I'm saying since the file already has a `_amd64` the `amd64` in the tag is unnecessary, so this should do the same as the above:

```go
//go:build !tinygo
```

That said, I'm not sure if there's a convention where duplicating the tag and filename is considered good form.

deadprogram: TinyGo uses build tags because of supporting so many different hardware devices, not just processor families, and without any operating system on them. You can ALSO use TinyGo to build for Linux/Windows/Darwin targets.

Hence the addition of the build tag specifically for the TinyGo compiler, since the baremetal on MCU targets are the ones we are using Wazero compiled using TinyGo.

Hopefully that clarifies the intent here.

ncruces: I'm sorry, I'm still confused.
I'm 100% ok with `//go:build !tinygo`.
I'm just asking if we need/want `//go:build amd64 && !tinygo` (given that the file is `_amd64.go`).

evacchi: @deadprogram what @ncruces is saying is the file is already named `entrypoint_amd64.go` hence the build tag `amd64` is redundant. 

I am kind of wondering if these two tags are necessary since there is already the catchall `entrypoint_others.go` where the build condition is `!arm64 && !amd64`.

deadprogram: OK, now I think I understand.

So basically all we would need would be `//go:build !tinygo` since the file naming already handles using "Big Go" compiler, and the build tag handles TinyGo.

ncruces: That said, the duplication doesn't hurt. It's a matter of style, and I was more asking the team if a style guide applies.

I know, I'm the unwanted style enforcer. 🙂

evacchi: I am not aware of a specific guideline, but I agree we should try to be minimal when possible :) but I am also not 100% sure if explicitly indicating `//go:build` overrides the naming convention, in that case better keep both for clarity IMO

deadprogram: > in that case better keep both for clarity IMO

In which case, this PR is ready for merge! :smile_cat:

ncruces: No, they're always cumulative, with an "and" _i.e._ `_filename.go && (go:build tags)`.

evacchi (approved):

Issue #2158, opened 2024-03-24
compiler(arm64) overflow in call relocations


davidmdm:
**Describe the bug**

A wasm module that was being compiled and instantiated with wazero 1.6.0 successfully now segfaults with version 1.7.0.

**To Reproduce**
I am building my own experimental tool that embeds wazero for supporting wasm. Here is a link to the wasm file I was testing with:
https://github.com/davidmdm/flights/releases/download/v0.0.0-yokecd-test/pg.wasm

This module is an executable that should output some JSON to stdout.

This is the code I use to intantiate it:

```go
package wasi

import (
	"bytes"
	"context"
	"crypto/rand"
	"fmt"
	"io"

	"github.com/tetratelabs/wazero"
	"github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1"

	"github.com/davidmdm/x/xerr"
)

func Execute(ctx context.Context, wasm []byte, release string, stdin io.Reader, args ...string) (output []byte, err error) {
	cfg := wazero.
		NewRuntimeConfig().
		WithCloseOnContextDone(true)

	// Create a new WebAssembly Runtime.
	wasi := wazero.NewRuntimeWithConfig(ctx, cfg)
	defer func() {
		err = xerr.MultiErrFrom("", err, wasi.Close(ctx))
	}()

	wasi_snapshot_preview1.MustInstantiate(ctx, wasi)

	// Because we are running a binary directly rather than embedding in an application,
	// we default to wiring up commonly used OS functionality.
	mod, err := wasi.CompileModule(ctx, wasm)
	if err != nil {
		return nil, fmt.Errorf("failed to compile module: %w", err)
	}

	var (
		stdout bytes.Buffer
		stderr bytes.Buffer
	)

	moduleCfg := wazero.
		NewModuleConfig().
		WithStdout(&stdout).
		WithStderr(&stderr).
		WithRandSource(rand.Reader).
		WithSysNanosleep().
		WithSysNanotime().
		WithSysWalltime().
		WithArgs(append([]string{release}, args...)...)

	if stdin != nil {
		moduleCfg = moduleCfg.WithStdin(stdin)
	}

	if _, err := wasi.InstantiateModule(ctx, mod, moduleCfg); err != nil {
		details := stderr.String()
		if details == "" {
			details = "(no output captured on stderr)"
		}
		return nil, fmt.Errorf("failed to instantiate module: %w: stderr: %s", err, details)
	}

[...]

PR #2164, opened 2024-03-26
wazevo(frontend): BCE should not propagate the memory address to unsealed blocks

evacchi:
tl;dr: I propose to propagate the memory address only for **sealed** blocks, where we know that no predecessors will be added at a later time.

This originated from an issue raised by @jerbob92 for go-pdfium.

Essentially, it looks like block 4 of function `5011` (local ID = 4960) in go-pdfium references some values that *may* not be visibile on by some of its predecessors.

We are propagating known bounds _while_ we are creating basic blocks; which means that, at a give time, we _might not know_ all of the possible predecessors to a given block. In the case of that block 4, the SSA dump shows:

```
blk4: (v305:i32,v341:i32) <-- (blk3,blk62)
```

which means that one of the predecessors in `blk62`. However, when we enter function `initializeCurrentBlockKnownBounds()` we enter `case 1` because there is only one predecessor (`blk3`) at that time. 

When blocks have more than 1 predecessor, we should enter the `default` case instead: this never propagates `kb.bound` AND `kb.address`, but it always propagates `kb.bound` and `ssa.ValueInvalid` because it cannot assume that the address will be valid on all paths.

This small change, as opposed to other earlier attempts, where I was unconditionally propagating `ssa.ValueInvalid` in all cases (see the Slack threads), does not affect codegen in a lot of cases (e.g. test case `TestCompiler_LowerToSSA/bounds check in if-else` is left unaffected)

### DIFF
--- a/internal/engine/wazevo/frontend/frontend.go
+++ b/internal/engine/wazevo/frontend/frontend.go
@@ -524,7 +524,13 @@ func (c *Compiler) initializeCurrentBlockKnownBounds() {
 	case 1:
 		pred := currentBlk.Pred(0).ID()
 		for _, kb := range c.getKnownSafeBoundsAtTheEndOfBlocks(pred).View() {
-			c.recordKnownSafeBound(kb.id, kb.bound, kb.absoluteAddr)
+			// Unless the block is sealed, we cannot assume the absolute address is valid:
+			// later we might add another predecessor that has no visibility of that value.
+			addr := ssa.ValueInvalid
+			if currentBlk.Sealed() {
+				addr = kb.absoluteAddr
+			}
+			c.recordKnownSafeBound(kb.id, kb.bound, addr)
 		}
 	default:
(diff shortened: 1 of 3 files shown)


mathetake:
this is exactly what I was thinking is the right way to fix the issue, so looks good

evacchi:
Zig Stdlib (arm64)
```
main: 45877240
this: 45877128
```

Zig Stdlib (amd64)
```
main: 44953787
this: 44954715
```

apparently this is actually slightly _better_ on arm64... (I checked multiple times)

mathetake (approved):

mathetake:
once https://github.com/tetratelabs/wazero/issues/2158 is resolved as well, let's run the fuzzing for both arm64 and amd64 for like multiple days (let's not forget to add `--no-trace-compares` flag: https://github.com/tetratelabs/wazero/pull/2152)

PR #2169, opened 2024-04-02
wazevo(arm64): trampoline for far calls on relocation

### LINKED ISSUE #2158: compiler(arm64) overflow in call relocations

evacchi:
Fixes #2158, follows up to #2167. 

Draft for feedback on the approach; also needs tests.

- Essentially, this adds some padding (`r.TrampolineOffset`) at the end of a function to accommodate a trampoline for far jumps (using BRL).
- The trampoline will be written **if** the relative jump is outside the accepted range. 
- When such an invalid relative jump is detected, instead of jumping to the function we jump to the trampoline:
   * 64-bit mov to `tmp` reg -- movz/movk/movk/movk 
   * BRL to `tmp`
   * unconditional branch back to the origin
- Note: This is currently adding such a padding for _all_ relocated calls, so obviously it is not optimal (will continue working on it later then undraft. I think it should be doable with some minor refactoring); however there is only waste of space: we use the trampoline only for far jumps.
- Also note: this is using a BLR. In other places (e.g. `resolveRelativeAddress`) we use relative jumps but at that time 1) we don't know the absolute address yet and 2) we can inject instruction at any point within the function because we are resolving the addresses before encoding.

Moreover, BRL's 64-bit addressing should obviously be comfortable enough to reach any point in the executable.

This was manually tested against the reproducers (top post and others in the comments) in #2158 and verified to work, but I will add proper tests. 

### DIFF
--- a/internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
+++ b/internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
@@ -21,12 +21,13 @@ func (m *machine) ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, bina
 		if diff < -(1<<25)*4 || diff > ((1<<25)-1)*4 {
 			// If the diff is out of range, we need to use a trampoline.
 			diff = int64(r.TrampolineOffset) - instrOffset
-			// The trampoline invokes the function using the BLR instruction
+			// The trampoline invokes the function using the BR instruction
 			// using the absolute address of the callee function.
+			// The BR instruction will not pollute LR, leaving set to the
+			// PC at this location. Thus, upon return, the callee will
+			// transparently return to the actual caller, skipping the trampoline.
 			absoluteCalleeFnAddress := uint(base) + uint(calleeFnOffset)
-			// The trampoline should return to the next instruction after the branch instruction.
-			returnOffset := -diff + 8
-			encodeTrampoline(absoluteCalleeFnAddress, binary, r.TrampolineOffset, returnOffset)
+			encodeTrampoline(absoluteCalleeFnAddress, binary, r.TrampolineOffset)
 		}
 		// https://developer.arm.com/documentation/ddi0596/2020-12/Base-Instructions/BL--Branch-with-Link-
 		imm26 := diff / 4
@@ -46,11 +47,11 @@ func (m *machine) UpdateRelocationInfo(r backend.RelocationInfo, totalSize int, 
 	// But when we invoke this method the refToBinaryOffset is not set for all funcRefs.
 	r.Offset += int64(totalSize)
 	r.TrampolineOffset = totalSize + len(body)
-	body = append(body, make([]byte, 4*6)...)
+	body = append(body, make([]byte, 4*5)...) // 5 instructions for the trampoline.
 	return r, body
 }
 
-func encodeTrampoline(addr uint, binary []byte, instrOffset int, returnOffset int64) {
+func encodeTrampoline(addr uint, binary []byte, instrOffset int) {
 	// The tmpReg is safe to overwrite.
 	tmpReg := regNumberInEncoding[tmp]
 
@@ -63,8 +64,7 @@ func encodeTrampoline(addr uint, binary []byte, instrOffset int, returnOffset in
 		encodeMoveWideImmediate(movkOp, tmpReg, uint64(uint16(addr>>16)), 1, 1),
 		encodeMoveWideImmediate(movkOp, tmpReg, uint64(uint16(addr>>32)), 2, 1),
 		encodeMoveWideImmediate(movkOp, tmpReg, uint64(uint16(addr>>48)), 3, 1),
-		encodeUnconditionalBranchReg(tmpReg, true),
-		encodeUnconditionalBranch(false, returnOffset-6*4),
+		encodeUnconditionalBranchReg(tmpReg, false),
 	}
 
 	for i, inst := range instrs {
--- a/internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
+++ b/internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
@@ -18,7 +18,7 @@ func (m *machine) ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, bina
 		brInstr := binary[instrOffset : instrOffset+4]
 		diff := int64(calleeFnOffset) - (instrOffset)
 		// Check if the diff is within the range of the branch instruction.
-		if diff < -(1<<25)*4 || diff > ((1<<25)-1)*4 {
+		if r.TrampolineOffset > 0 {
 			// If the diff is out of range, we need to use a trampoline.
 			diff = int64(r.TrampolineOffset) - instrOffset
 			// The trampoline invokes the function using the BR instruction
@@ -42,13 +42,22 @@ func (m *machine) ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, bina
 	}
 }
 
-func (m *machine) UpdateRelocationInfo(r backend.RelocationInfo, totalSize int, body []byte) (backend.RelocationInfo, []byte) {
-	// FIXME: this should add padding conditionally based on refToBinaryOffset[r.FuncRef].
-	// But when we invoke this method the refToBinaryOffset is not set for all funcRefs.
-	r.Offset += int64(totalSize)
-	r.TrampolineOffset = totalSize + len(body)
-	body = append(body, make([]byte, 4*5)...) // 5 instructions for the trampoline.
-	return r, body
+const relocationTrampolineSize = 5 * 4
+
+func (m *machine) RelocationTrampolineSize(rels []backend.RelocationInfo) int {
+	return relocationTrampolineSize * len(rels)
+}
+
+func (m *machine) UpdateRelocationInfo(refToBinaryOffset map[ssa.FuncRef]int, trampolineOffset int, r backend.RelocationInfo) (backend.RelocationInfo, int) {
+	instrOffset := r.Offset
+	calleeFnOffset := refToBinaryOffset[r.FuncRef]
+	diff := int64(calleeFnOffset) - (instrOffset)
+	trampolineSize := 0
+	if diff < -(1<<25)*4 || diff > ((1<<25)-1)*4 {
+		r.TrampolineOffset = trampolineOffset
+		trampolineSize = relocationTrampolineSize
+	}
+	return r, trampolineSize
 }
 
 func encodeTrampoline(addr uint, binary []byte, instrOffset int) {
--- a/internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
+++ b/internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
@@ -9,8 +9,6 @@ import (
 )
 
 // ResolveRelocations implements backend.Machine ResolveRelocations.
-//
-// TODO: unit test!
 func (m *machine) ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, binary []byte, relocations []backend.RelocationInfo) {
 	base := uintptr(unsafe.Pointer(&binary[0]))
 	for _, r := range relocations {
@@ -18,7 +16,8 @@ func (m *machine) ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, bina
 		calleeFnOffset := refToBinaryOffset[r.FuncRef]
 		brInstr := binary[instrOffset : instrOffset+4]
 		diff := int64(calleeFnOffset) - (instrOffset)
-		// Check if the diff is within the range of the branch instruction.
+		// The diff was outside the range of the branch instruction, so we use a
+		// trampoline (see UpdateRelocationInfo).
 		if r.TrampolineOffset > 0 {
 			// If the diff is out of range, we need to use a trampoline.
 			diff = int64(r.TrampolineOffset) - instrOffset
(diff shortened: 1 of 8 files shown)


internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
```diff
@@ -32,3 +39,38 @@ func (m *machine) ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, bina
...
+	const movkOp = uint32(0b11)
+	// Note: for our purposes the 64-bit width of the reg should be enough.
+	//       however, larger values could be written to and loaded from a constant pool (see encodeBrTableSequence).
+	instrs := []uint32{
+		encodeMoveWideImmediate(movzOp, tmpReg, uint64(uint16(addr)), 0, 1),
+		encodeMoveWideImmediate(movkOp, tmpReg, uint64(uint16(addr>>16)), 1, 1),
+		encodeMoveWideImmediate(movkOp, tmpReg, uint64(uint16(addr>>32)), 2, 1),
+		encodeMoveWideImmediate(movkOp, tmpReg, uint64(uint16(addr>>48)), 3, 1),
+		encodeUnconditionalBranchReg(tmpReg, true),
+		encodeUnconditionalBranch(false, returnOffset-6*4),
```
evacchi, line 66: it's possible that we can resolve the width of the jump earlier and either generate a smaller trampoline, with less instructions, or use a different strategy for encoding the address (for now I just wanted to see if the general approach was working)

internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
```diff
@@ -11,14 +11,22 @@ import (
...
 		// Check if the diff is within the range of the branch instruction.
 		if diff < -(1<<25)*4 || diff > ((1<<25)-1)*4 {
-			panic(fmt.Sprintf("TODO: too large binary where branch target is out of the supported range +/-128MB: %#x", diff))
+			// If the diff is out of range, we need to use a trampoline.
+			diff = int64(r.TrampolineOffset) - instrOffset
+			// The trampoline invokes the function using the BLR instruction
+			// using the absolute address of the callee function.
+			absoluteCalleeFnAddress := uint(base) + uint(calleeFnOffset)
+			// The trampoline should return to the next instruction after the branch instruction.
+			returnOffset := -diff + 8
```
mathetake, line 28: I don't think this is necessary by using `BR`, and not `BLR` in the trampoline

internal/engine/wazevo/engine.go
```diff
@@ -264,7 +264,7 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
 		// At this point, relocation offsets are relative to the start of the function body,
 		// so we adjust it to the start of the executable.
 		for _, r := range rels {
-			r.Offset += int64(totalSize)
+			body = e.machine.UpdateRelocationInfo(&r, totalSize, body)
```
mathetake, line 267: this leaks `r` into heap I guess

mathetake (commented):
the direction looks good

evacchi:
>  did you do some research on how other linkers/comilers deal with this?

Yes, I will add some notes inline in the code. Here's my summary. 

On ARM there are [3 types of "veneer"][3types] (trampoline/thunk in ARM parlance)

- inline, which is ARM-specific and it's a combination of an absolute jump and a command to [switch between different instruction sets (Arm to Thumb; i.e. A32->T32 etc.)][ISAs] using the `BX` instruction; this is also called [interworking][book2] 
- short-range (still relative jump, essentially like those we generate e.g. in `resolveRelativeAddresses`)
- long-range (absolute, like this one)

The main issue is where to put them. My first thought was to tack it at the end of the current function for simplicity, but how do linkers do it? 

Because the job of a traditional linker is to shuffle code around, it's cheap for them to generate new code and then point the branch to it; essentially, they can generate a new (internal) symbol. [This online book ("Embedded Systems Security and TrustZone") gives a nice explanation of how this works][book]. But in our case adding a new symbol/code section would mean adding a new `functionOffset` and mmapping a new segment. 

Placing these new segments now represents a problem of its own; for instance:

- we could add these new segments as we need them, incrementing the `i` in `cm.functionOffsets[i]`; e.g. say we are processing function `i` and we discover we need `k` trampolines, then we would generate those `k` sections and add them at `i+1`, ..., `i+k`,
 but then we would lose the correspondence between `cm.functionOffsets[i]` and `module.CodeSection[i]` (as well as other similar cases), and that would cascade to the surrounding code.

- Instead, we could add them to the end of `functionOffsets`, with `j=len(functionOffsets), ...`. This however means that, depending on the order in which we generate and `mmap` them they could be _even further_ than the original function that needed to be invoked.

- Besides, this also adds more `mmapped` code to be collected and finalized.

In fact, these sections can be fused with the code of a function: this behavior can be controlled through linker scripts; for instance, [the veneer for interlinking is joined with the `.text` section by the default linker script][book2].

So, tacking it to the bottom of the original function is actually a behavior that can be controlled even in traditional linkers, and, in our case, it addresses the other problems; assuming that the bottom of the function is within range for a relative jump. 

I currently ruled out encoding the trampoline as a short-range thunk for simplicity: it's an optimization we can address later, and it was important to get to the bottom of this first. 

Other optimizations might be detecting multiple invocations to the same function and jump to the same veneer (that requires ).

Finally, the text itself of the veneer: how do you load the address? There are a few possible approaches; in my case, again, for simplicity, I am `movz/movk/movk/movk`'ing the address to the `tmp` reg which we know is safe to use. Another way to load the address is similar to `brTableSequence`, using `adr` and a constant pool. In some cases (when the address is especially large), `adrp`+`add` might need to be used instead (`adrp` addresses by 4KB-pages). E.g. the example in the [book][book] uses `ldr r0, [pc, #8]` which is essentially what `adr r0, #8` would translate to. 


[3types]: https://developer.arm.com/documentation/101754/0622/armlink-Reference/Image-Structure-and-Generation/Linker-generated-veneers/What-is-a-veneer-
[ISAs]: https://developer.arm.com/documentation/102374/0102/Instruction-sets-in-the-Arm-architecture
[book]: https://embeddedsecurity.io/sec-c#sec:c-veneers
[book2]: https://embeddedsecurity.io/sec-c#sec:fw-excursus-glue

internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
evacchi: clever!

evacchi:
I improved the PoC but it's not clean enough for reviewing, so I'll push the code later, but I'll write some notes here.

I wanted to make sure that we do not waste too much space for trampolines that are not needed. In the code here we always allocate the space for a trampoline for each relocation, regardless if this trampoline will be needed (i.e. when the relocation does not require a far jump).

Now, suppose that we have functions `fk`, `fk1`, `fk2` all adjacent to each other. Suppose that `fk` needs more space to accommodate a trampoline to `fk2`. If we do a linear scan, at this point, we can't know for certain the offset of `fk2` because we don't know if also `fk1` will require some trampolines (causing it to increase in size). 

The solution is obviously to iterate more than once; but how many times? We could compute a fixed point, but instead:

- we can do a first pass assuming the "worst case" scenario, where each relocation really needs a trampoline; 
  - this way, we can precompute the size of each function with all the trampolines (and thus their offsets) knowing that they can't grow further than that
- we iterate over the relocations and flag those that really need a trampoline:
  - now, for each function, if any of the offset diffs is **within** range, we know **for certain** that that trampoline **won't** be needed
  - conversely, if any of the offset diffs is **outside** the acceptable range, we will just assume that a trampoline will be needed to avoid further iterations (we can further optimize the edge cases in the future)
- we do a final pass where we actually allocate space at the end of each function and write the trampolines

internal/engine/wazevo/backend/machine.go
```diff
@@ -61,6 +61,12 @@ type (
 		// ResolveRelocations resolves the relocations after emitting machine code.
 		ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, binary []byte, relocations []RelocationInfo)
 
+		// UpdateRelocationInfo recomputes the RelocationInfo after emitting machine code and returns the updated
+		// RelocationInfo and the size of the trampoline if necessary.
+		UpdateRelocationInfo(refToBinaryOffset map[ssa.FuncRef]int, trampolineOffset int, r RelocationInfo) (res RelocationInfo, trampolineSize int)
+
+		RelocationTrampolineSize(relocations []RelocationInfo) int
```
evacchi, line 68: Not really in love with this, but I could not think of anything else. We need to know the size estimate of a given trampoline "section" at the end of a function.

note to self: need doc comment

internal/engine/wazevo/engine.go
```diff
@@ -220,6 +220,7 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
 	totalSize := 0 // Total binary size of the executable.
 	cm.functionOffsets = make([]int, localFns)
 	bodies := make([][]byte, localFns)
+	sourceOffsets := make([][]backend.SourceOffsetInfo, localFns)
```
evacchi, line 223: we need to append the sourceOffsets in the second stage, so we need to store `be.SourceOffsetInfo()` somewhere. I don't love this change, but I could not think of anything better at the moment...

evacchi:
Added a few test cases, will continue later.

internal/engine/wazevo/backend/isa/arm64/machine_relocation.go
```diff
@@ -32,3 +41,46 @@ func (m *machine) ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, bina
...
+		encodeUnconditionalBranchReg(tmpReg, false),
+	}
+
+	for i, inst := range instrs {
+		instrBytes := binary[instrOffset+i*4 : instrOffset+(i+1)*4]
+		instrBytes[0] = byte(inst)
+		instrBytes[1] = byte(inst >> 8)
+		instrBytes[2] = byte(inst >> 16)
+		instrBytes[3] = byte(inst >> 24)
+	}
```
mathetake, line 85: can we use `binary.LittleEndian.PutUint32` ?

internal/engine/wazevo/backend/machine.go
```diff
@@ -61,6 +61,12 @@ type (
 		// ResolveRelocations resolves the relocations after emitting machine code.
 		ResolveRelocations(refToBinaryOffset map[ssa.FuncRef]int, binary []byte, relocations []RelocationInfo)
 
+		// UpdateRelocationInfo recomputes the RelocationInfo after emitting machine code and returns the updated
+		// RelocationInfo and the size of the trampoline if necessary.
+		UpdateRelocationInfo(refToBinaryOffset map[ssa.FuncRef]int, trampolineOffset int, r RelocationInfo) (res RelocationInfo, trampolineSize int)
```
mathetake, line 66: nit: Make this accept pointer and let's fix the callsite to avoid the heap allocation (see various places in the codebase)

```suggestion
		UpdateRelocationInfo(refToBinaryOffset map[ssa.FuncRef]int, trampolineOffset int, r *RelocationInfo) trampolineSize int
```

mathetake (commented):
on the right track I guess!

internal/engine/wazevo/backend/machine.go
evacchi: I think it should be good now

mathetake:
would you mind comparing the binary size and the compilation time perf before/after for stdlib binaries (Zig, Go, and libsodium would be fine)?

internal/engine/wazevo/engine.go
```diff
@@ -264,17 +257,48 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
...
 		// so we adjust it to the start of the executable.
 		for _, r := range rels {
+			r.Caller = fref
 			r.Offset += int64(totalSize)
 			e.rels = append(e.rels, r)
 		}
 
 		bodies[i] = body
-		totalSize += len(body)
+		totalSize += len(body) + e.machine.RelocationTrampolineSize(rels)
```
mathetake, line 266: I think the logic is broken here. Inside RelocationTrampolineSize, it checks `rels[i].TrampolineOffset > 0` but TrampolineOffset is always zero at this point. In fact
```diff
--- a/internal/engine/wazevo/engine.go
+++ b/internal/engine/wazevo/engine.go
@@ -263,7 +263,11 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
                }
 
                bodies[i] = body
-               totalSize += len(body) + e.machine.RelocationTrampolineSize(rels)
+               s := e.machine.RelocationTrampolineSize(rels)
+               if s != 0 {
+                       panic("BUG")
+               }
+               totalSize += len(body) + s
                if wazevoapi.PrintMachineCodeHexPerFunction {
                        fmt.Printf("[[[machine code for %s]]]\n%s\n\n", wazevoapi.GetCurrentFunctionName(ctx), hex.EncodeToString(body))
                }
```

this obvious assertion doesn't get hit after running `wazero run pg.wasm`.

mathetake: This indicates that the current code is not starting with the worst case size, but instead "smallest size" scenario.

mathetake:
ok I think I will take this over. Thank you for your hard work and hints here @evacchi

mathetake:
superseded by #2181

PR #2170, opened 2024-04-04
threads: stop explicit use of mmap for shared memory

mathetake:
this simplifies the discussion around https://github.com/tetratelabs/wazero/pull/2157 cc @anuraaga @ncruces


anuraaga (approved):
Thanks!

PR #2173, opened 2024-04-04
Renames PageSize -> Pages

mathetake:
@ncruces  `PageSize` is not appropriate name as it's defined in the spec as 65536 bytes.


ncruces (approved):

mathetake:
I might revert the API in the first place before the next release (as I still not sure if anyone needs this) but this is anyway better than the original

ncruces:
We can always use `Grow(0)`. If so, we should update the documentation noting that workaround.

mathetake:
That sounds better I think

PR #2175, opened 2024-04-06
experimental: cleanup context keys

ncruces:
We follow 2 different models in wazero for context keys:
1. one where we make the key type public and the user is supposed to do `context.WithValue(ctx, key, value)`
2. another where we expose `experimental.WithBlargh(ctx, blargh)`

The second doesn't expose the key, and is more typesafe because it ensures `blargh` is of the expected type.

This PR refactors to use *2* everywhere.
Old code left deprecated, but we can remove (since it's experimental).

### DIFF
--- a/experimental/checkpoint.go
+++ b/experimental/checkpoint.go
@@ -25,12 +25,12 @@ type Snapshotter interface {
 // The context.Context passed to a exported function invocation should have this key set
 // to a non-nil value, and host functions will be able to retrieve it using SnapshotterKey.
 //
-// Deprecated: use [WithSnapshotter] to enable snapshots.
+// Deprecated: use WithSnapshotter to enable snapshots.
 type EnableSnapshotterKey = ctxkey.EnableSnapshotterKey
 
 // WithSnapshotter enables snapshots.
 // Passing the returned context to a exported function invocation enables snapshots,
-// and allows host functions to retrieve the [Snapshotter] using [SnapshotterKey].
+// and allows host functions to retrieve the Snapshotter using SnapshotterKey.
 func WithSnapshotter(ctx context.Context) context.Context {
 	return context.WithValue(ctx, ctxkey.EnableSnapshotterKey{}, struct{}{})
 }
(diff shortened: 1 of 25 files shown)


experimental/checkpoint.go
```diff
@@ -18,8 +24,25 @@ type Snapshotter interface {
 // EnableSnapshotterKey is a context key to indicate that snapshotting should be enabled.
 // The context.Context passed to a exported function invocation should have this key set
 // to a non-nil value, and host functions will be able to retrieve it using SnapshotterKey.
-type EnableSnapshotterKey struct{}
+//
+// Deprecated: use [WithSnapshotter] to enable snapshots.
+type EnableSnapshotterKey = ctxkey.EnableSnapshotterKey
```
mathetake, line 29: We are not using this `[symbol]` anywhere, so could you remove?

```suggestion
// Deprecated: use WithSnapshotter to enable snapshots.
type EnableSnapshotterKey = ctxkey.EnableSnapshotterKey
```

experimental/listener.go
```diff
@@ -23,10 +24,17 @@ type StackIterator interface {
 	ProgramCounter() ProgramCounter
 }
 
-// FunctionListenerFactoryKey is a context.Context Value key. Its associated value should be a FunctionListenerFactory.
+// FunctionListenerFactoryKey is a context.Context Value key.
+// Its associated value should be a FunctionListenerFactory.
 //
-// See https://github.com/tetratelabs/wazero/issues/451
-type FunctionListenerFactoryKey struct{}
+// Deprecated: use [WithSnapshotter] to enable snapshots.
```
mathetake, line 30: WithFunctionListenerFactory

experimental/checkpoint.go
ncruces: Yeah, we're no longer using it, users might be.
My idea was to keep it deprecated for one (?) version, then remove it.

But we can remove immediately, since it's experimental. Just to confirm, is that the intention?

mathetake: no what I meant is literal characters `[` and `]`. see the suggested change

ncruces: Oh, yeah, those are [doc links](https://tip.golang.org/doc/comment#doclinks) which I'm slowing converting my code bases to. Reverting.

experimental/listener.go
ncruces: Thanks!

mathetake (approved):

Issue #1331, opened 2023-04-01
Expose memory implementation


inkeliz:
**Is your feature request related to a problem? Please describe.**

I'm playing with mmap, and because of that I need to manipulate the memory from Wazero. Currently, I have the following code:

```
		// internalWazeroMemoryInstance is a private struct in wazero
		type internalWazeroMemoryInstance struct {
			Buffer        []byte
			Min, Cap, Max uint32
			mux           sync.RWMutex
		}

		// wasm.Memory is private, we need to use unsafe
		_hostMemoryInstance := mod.Memory()
		hostMemoryInstance := (*[2]*internalWazeroMemoryInstance)(unsafe.Pointer(&_hostMemoryInstance))[1]

		hostMemoryInstance.mux.Lock()
		defer hostMemoryInstance.mux.Unlock()

                //...
               
                // redefine wazero memory (including capacity)
		hostMemoryInstance.Buffer = mem
		hostMemoryInstance.Cap = f.pages
```

https://github.com/inkeliz/karmap/blob/b0ed8241fbc27fd4f3b915bdba8da9ebec17f8ba/wazero/wasmkarmap/karmap.go#L62-L74

That is necessaty because I need to modify the `Buffer` and also `Cap`. And I need to do it for two reasons:

1. I have a custom Buffer implementation, using mmap. 
2. I need to have a fixed size of memory (see https://github.com/tetratelabs/wazero/issues/1330).

**Describe the solution you'd like**
Maybe a simpler interface/struct of Memory, which could be:

```
type Memory struct {
Buffer []byte
Grow func(pages int) int
}
```

Or, using Interface:

```
type MemoryInterface interface {
 Buffer() []byte
 Grow(page int) int
}
```

In that case, anyone can have any custom implementation, which will be called when "Grow" (fixing https://github.com/tetratelabs/wazero/issues/1150), and also provides one way to implement `Buffer` using mmap, arenas (https://github.com/golang/go/issues/51317) or anything else.

The current Memory implementation is very large, and contains a lot of `Read*` and such, which makes very hard to implement and, since we have `[]byte` by default, anyone can read from that (so ReadFloat32Le can just take from .Buffer()).

**Describe alternatives you've considered**
Using unsafe, which works, but it's unsafe and unstable.

**Additional context**

mathetake:
https://github.com/tetratelabs/wazero/pull/2177

Issue #1150, opened 2023-02-21
Memory Grow Callback


samyfodil:
When used in a Serverless environment wazero needs to provide a mean to limit memory used by a module and its dependencies. A Call back before `grow` is executed would help implement this and many more use cases like debugging.

codefromthecrypt:
Thanks for raising this! TL;DR; we have an incomplete "memory" log scope which needs this anyway (in the experimental interceptor).

cc @ckaznocha @achille-roussel @inkeliz @danielpacak who might be interested, but I'm sure also interested in zero overhead unless active.

There have been many issues around this #417 #483 #500. Today, some recompile with flags (not every compiler has one), use wabin to rewrite the wasm with a max, or assign a different max with RuntimeConfig. However, none of these approaches are hooks. That said, emscripten and GOOS=js already insert memory hook host functions.

I think the best way forward is to finish the interceptor stuff, and augment the "memory" log scope to include this. This will prove it is interceptable.  However, the devil is in the details as "memory.grow" is actually a host function, but it isn't invoked by "call" or "call_indirect", rather well.. "memory.grow". There's some interesting work that would need to complete to make built-in host functions interceptable. That way, you can keep zero overhead, and also change the input or result like any other function.

Unless someone contributes, I doubt this will get into 1.0, however, if someone with an appetite for internal refactoring plus feedback cycles can start on this, it could. We won't ever consider "experimental" packages, which include the interceptors at the moment, a part of any compatibility promise. If no one volunteers until after 1.0 I think someone in the project team, possibly me, can take this on.

# some notes in general.

The memory.grow op returns a bool, which when false the guest interprets in a compiler-specific way to produce an error like "out of memory", so for example, I wouldn't suggest anyone panic in a hook as it prevents the guest from knowing a better message.

It is true that if there was a callback, you can do host logic, and while not sure precisely what debugging might mean (maybe you can elaborate) logging could be one form of that.

In fact we are overdue adding a logging hook for memory.grow, (using the existing interceptor) and the impl would allow the same as you mention. I think that could be a lightweight way to start this.

For the interceptor, I think the easiest way out is to make "memory.grow" a pseudo host function. In fact it is a host function, just it isn't in the function table. I've thought about that in the past, but it takes some thinking to do it right. The jist of it would be to make a pseudo-functions for "memory.grow" "table.grow" (there are a couple more I think). This way the interceptors can keep zero overhead when nothing is being intercepted. One of the major

[...]

PR #2177, opened 2024-04-06
experimental: configure custom memory allocator

ncruces:
Still draft, missing tests, but you can see it put to good use in https://github.com/ncruces/go-sqlite3/pull/71.
It passes extensive tests on Linux (`amd64` native, `arm64` emulated) and macOS (`amd64`/`arm64` native), and solves a long standing issue with SQLite (on those platforms).

Happy to redesign the API, putting the PR up precisely to collect early feedback!

### DIFF
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -18,6 +18,8 @@ type MemoryBuffer interface {
 	// Return the backing []byte for the memory buffer.
 	Buffer() []byte
 	// Grow the backing memory buffer to size bytes in length.
+	// To back a shared memory, Grow can't change the address
+	// of the backing []byte (only its length/capacity may change).
 	Grow(size uint64) []byte
 	// Free the backing memory buffer.
 	Free()
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -8,7 +8,7 @@ import (
 
 // MemoryAllocator is a memory allocation hook which is invoked
 // to create a new MemoryBuffer, with the given specification:
-// min is the initial and minimum length of the backing []byte,
+// min is the initial and minimum length (in bytes) of the backing []byte,
 // cap a suggested initial capacity, and max the maximum length
 // that will ever be requested.
 type MemoryAllocator func(min, cap, max uint64) MemoryBuffer
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -8,8 +8,8 @@ import (
 
 type MemoryAllocator interface {
 	Make(min, cap, max uint64) []byte
-	Grow(old []byte, more uint64) []byte
-	Free([]byte)
+	Grow(uint64) []byte
+	Free()
 }
 
 func WithMemoryAllocator(ctx context.Context, allocator MemoryAllocator) context.Context {
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -6,18 +6,20 @@ import (
 	"github.com/tetratelabs/wazero/internal/ctxkey"
 )
 
-// MemoryAllocator is a memory allocation hook.
-type MemoryAllocator interface {
-	// Make is invoked to create a new memory, with the given specification.
-	// Implementations must return a []byte min bytes in length,
-	// should return a []byte with at least cap capacity,
-	// and be prepared to allocate up to max bytes of memory.
-	Make(min, cap, max uint64) []byte
+// MemoryAllocator is a memory allocation hook which is invoked
+// to create a new MemoryBuffer, with the given specification:
+// min is the initial and minimum length of the backing []byte,
+// cap a suggested initial capacity, and max the maximum length
+// that will ever be requested.
+type MemoryAllocator func(min, cap, max uint64) MemoryBuffer
 
-	// Grow is invoked to grow the memory to size bytes in length.
+// MemoryBuffer is a memory buffer that backs a Wasm memory.
+type MemoryBuffer interface {
+	// Return the backing []byte for the memory buffer.
+	Buffer() []byte
+	// Grow the backing memory buffer to size bytes in length.
 	Grow(size uint64) []byte
-
-	// Free is invoked to free the memory.
+	// Free the backing memory buffer.
 	Free()
 }
 
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -6,12 +6,23 @@ import (
 	"github.com/tetratelabs/wazero/internal/ctxkey"
 )
 
+// MemoryAllocator is a memory allocation hook.
 type MemoryAllocator interface {
+	// Make is invoked to create a new memory, with the given specification.
+	// Implementations must return a []byte min bytes in length,
+	// should return a []byte with at least cap capacity,
+	// and be prepared to allocate up to max bytes of memory.
 	Make(min, cap, max uint64) []byte
-	Grow(uint64) []byte
+
+	// Grow is invoked to grow the memory to size bytes in length.
+	Grow(size uint64) []byte
+
+	// Free is invoked to free the memory.
 	Free()
 }
 
+// WithMemoryAllocator registers the given MemoryAllocator into the given
+// context.Context.
 func WithMemoryAllocator(ctx context.Context, allocator MemoryAllocator) context.Context {
 	if allocator != nil {
 		return context.WithValue(ctx, ctxkey.MemoryAllocatorKey{}, allocator)
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -9,6 +9,7 @@ import (
 type MemoryAllocator interface {
 	Make(min, cap, max uint64) []byte
 	Grow(old []byte, more uint64) []byte
+	Free([]byte)
 }
 
 func WithMemoryAllocator(ctx context.Context, allocator MemoryAllocator) context.Context {
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -1,0 +1,19 @@
+package experimental
+
+import (
+	"context"
+
+	"github.com/tetratelabs/wazero/internal/ctxkey"
+)
+
+type MemoryAllocator interface {
+	Make(min, cap, max uint64) []byte
+	Grow(old []byte, more uint64) []byte
+}
+
+func WithMemoryAllocator(ctx context.Context, allocator MemoryAllocator) context.Context {
+	if allocator != nil {
+		return context.WithValue(ctx, ctxkey.MemoryAllocatorKey{}, allocator)
+	}
+	return ctx
+}
--- a/internal/wasm/memory.go
+++ b/internal/wasm/memory.go
@@ -59,7 +59,7 @@ type MemoryInstance struct {
 	// with a fixed weight of 1 and no spurious notifications.
 	waiters sync.Map
 
-	allocator experimental.MemoryAllocator
+	allocator experimental.MemoryBuffer
 }
 
 // NewMemoryInstance creates a new instance based on the parameters in the SectionIDMemory.
@@ -67,7 +67,13 @@ func NewMemoryInstance(memSec *Memory, allocator experimental.MemoryAllocator) *
 	min := MemoryPagesToBytesNum(memSec.Min)
 	cur := MemoryPagesToBytesNum(memSec.Cap)
 	max := MemoryPagesToBytesNum(memSec.Max)
-	if memSec.IsShared {
+
+	var buffer []byte
+	var alloc experimental.MemoryBuffer
+	if allocator != nil {
+		alloc = allocator(min, cur, max)
+		buffer = alloc.Buffer()
+	} else if memSec.IsShared {
 		// Shared memory needs a fixed buffer, so allocate with the maximum size.
 		//
 		// The rationale as to why we can simply use make([]byte) to a fixed buffer is that Go's GC is non-relocating.
@@ -78,12 +84,7 @@ func NewMemoryInstance(memSec *Memory, allocator experimental.MemoryAllocator) *
 		// the memory buffer allocation here is virtual and doesn't consume physical memory until it's used.
 		// 	* https://github.com/golang/go/blob/8121604559035734c9677d5281bbdac8b1c17a1e/src/runtime/malloc.go#L1059
 		//	* https://github.com/golang/go/blob/8121604559035734c9677d5281bbdac8b1c17a1e/src/runtime/malloc.go#L1165
-		cur = max
-	}
-
-	var buffer []byte
-	if allocator != nil {
-		buffer = allocator.Make(min, cur, max)
+		buffer = make([]byte, min, max)
 	} else {
 		buffer = make([]byte, min, cur)
 	}
@@ -93,7 +94,7 @@ func NewMemoryInstance(memSec *Memory, allocator experimental.MemoryAllocator) *
 		Cap:       memoryBytesNumToPages(uint64(cap(buffer))),
 		Max:       memSec.Max,
 		Shared:    memSec.IsShared,
-		allocator: allocator,
+		allocator: alloc,
 	}
 }
 
@@ -232,8 +233,18 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 	if newPages > m.Max || int32(delta) < 0 {
 		return 0, false
 	} else if m.allocator != nil {
-		m.Buffer = m.allocator.Grow(MemoryPagesToBytesNum(newPages))
-		m.Cap = newPages
+		buffer := m.allocator.Grow(MemoryPagesToBytesNum(newPages))
+		if m.Shared {
+			if unsafe.SliceData(buffer) != unsafe.SliceData(m.Buffer) {
+				panic("shared memory cannot move")
+			}
+			// Use atomic to ensure maximum length is visible across threads.
+			atomicGrowLength(&m.Buffer, uintptr(len(buffer)))
+			m.Cap = memoryBytesNumToPages(uint64(cap(buffer)))
+		} else {
+			m.Buffer = buffer
+			m.Cap = newPages
+		}
 		return currentPages, true
 	} else if newPages > m.Cap { // grow the memory.
 		if m.Shared {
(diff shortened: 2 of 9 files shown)


experimental/memory.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"context"
+
+	"github.com/tetratelabs/wazero/internal/ctxkey"
+)
+
+type MemoryAllocator interface {
+	Make(min, cap, max uint64) []byte
+	Grow(old []byte, more uint64) []byte
+	Free([]byte)
```
ncruces, line 12: Since this is an interface, implementations can hold state, so `Grow` and `Free` could potentially _not_ receive the slice.

Also sizes are in bytes, but could just as easily be in pages. Still, page size would one more thing for implementations to mess up, so maybe this is better.

internal/wasm/memory.go
```diff
@@ -57,11 +58,15 @@ type MemoryInstance struct {
...
+
+	allocator experimental.MemoryAllocator
 }
 
 // NewMemoryInstance creates a new instance based on the parameters in the SectionIDMemory.
-func NewMemoryInstance(memSec *Memory) *MemoryInstance {
-	var size uint64
+func NewMemoryInstance(memSec *Memory, allocator experimental.MemoryAllocator) *MemoryInstance {
+	min := MemoryPagesToBytesNum(memSec.Min)
+	cur := MemoryPagesToBytesNum(memSec.Cap)
```
ncruces, line 68: Sadly, cap is taken as a name (needed bellow).

internal/wasm/memory.go
```diff
@@ -73,18 +78,22 @@ func NewMemoryInstance(memSec *Memory) *MemoryInstance {
...
+		cur = max
 	}
 
-	buffer := make([]byte, MemoryPagesToBytesNum(memSec.Min), size)
+	var buffer []byte
+	if allocator != nil {
+		buffer = allocator.Make(min, cur, max)
+	} else {
+		buffer = make([]byte, min, cur)
+	}
```
ncruces, line 89: Opted for `nil` checks rather than a default implementation, because the zero value is used in a bunch of tests.

experimental/memory.go
```diff
@@ -0,0 +1,31 @@
+package experimental
+
+import (
+	"context"
+
+	"github.com/tetratelabs/wazero/internal/ctxkey"
+)
+
+// MemoryAllocator is a memory allocation hook.
+type MemoryAllocator interface {
```
anuraaga, line 10: If this weren't an experimental API, I guess currently it would have to go on `ModuleConfig` rather than `RuntimeConfig` since it looks like it can only return a single buffer. An alternative that is runtime-scoped could return an interface

```
type Buffer interface {
  Bytes() []byte
}

type MemoryAllocator interface {
  Make(...) Buffer

  Grow(Buffer, uint64) Buffer

  Free(Buffer)
```

`Memory` would store a reference to both Buffer and `[]byte`, later for all memory operations and former for interacting with the allocator.

This somehow seems like what I would expect from an allocator but it's not clearly better. I think with the current one, I would at least rename it to `ModuleMemoryAllocator` to make it more obvious it is only scoped to a single instance of a module (of course not using multi-memory which I don't think we need to consider, but just as a FYI).

anuraaga: I realized another idea that's probably better than above is

```
type Buffer interface {
  Bytes() []byte

  Grow(uint64)

  Free()
}

type MemoryAllocator interface {
  Make(...) Buffer
```

This is only if making allocator a runtime config rather than a module config. Somehow these seem closer to how other APIs like malloc look

ncruces: Thanks! That's an improvement, but I still trying to come up with an API that's safe to use by shared memories. Unfortunately, it's likely not very ergonomic…

I mean, `Grow` needs to update the `[]byte` inside `MemoryInstance` atomically. The only way I see to do that, is to pass a `*[]byte` into `Grow`?

ncruces: Sorry for the churn. New attempt.

Also fixes a race (I think) in existing shared memory, where the size of shared memory could go back (become smaller).

anuraaga: Previously we had a lock around Grow to make it atomic but we removed it since it didn't seem to make sense if the guest doesn't itself lock around Grow, and it's at least the case with wasi-libc. So not sure if we need to be worried here either, or is it a bad assumption?

ncruces: Tbh, I don't know.
But the previous atomic write doesn't seem to be guaranteeing anything much?
I mean, yes, it makes the write visible to other threads, but it doesn't ensure length can't go down.
I don't know what wasi-libc is doing if it needs locks when multiple threads try to grow concurrently.
Do you have a code pointer?

internal/wasm/memory.go
```diff
@@ -222,6 +232,20 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 	newPages := currentPages + delta
 	if newPages > m.Max || int32(delta) < 0 {
 		return 0, false
+	} else if m.allocator != nil {
+		buffer := m.allocator.Grow(MemoryPagesToBytesNum(newPages))
+		if m.Shared {
+			if unsafe.SliceData(buffer) != unsafe.SliceData(m.Buffer) {
+				panic("shared memory cannot move")
```
ncruces, line 239: Would it make sense to `return 0, false` here?

experimental/memory.go
anuraaga: I think the code pointer is this one

https://github.com/WebAssembly/wasi-libc/blob/main/dlmalloc/src/malloc.c#L262

Every malloc call is surrounded by a lock. Since only malloc can grow, this implicitly locks grow. This seems to make sense to me as grow looks closer to sbrk (not allowed to be called concurrently AFAIK), then mmap. I'm not sure how to reason about the return value of grow if it could be called concurrently,

> but it doesn't ensure length can't go down.

Currently the code is assuming non-concurrent grow so only cares about the visibility and the length wouldn't go down (I think, let me know if I missed something different).

I filed https://github.com/WebAssembly/threads/issues/224 to see what the spec may have intended. I would personally still assume grow is called non-concurrently until hearing on it since I think it's been our assumption since https://github.com/tetratelabs/wazero/pull/1755

ncruces: So what does it mean that we care about visibility? Who/what views it?

If it's locked in the guest, which makes sense, that mutex's release takes care to publish the changes. We wouldn't need to do anything about it, right?

ncruces: I'm asking because I think the allocator needs the same treatment, and I'd like to understand what (if anything) I should document.

From what you're saying, custom the allocators for shared memories wouldn't need a mutex, and ISTM they don't need this atomic either.

anuraaga: Hmm tried looking for a conversation I thought we had about visibility but can't find it maybe it was an unmerged PR.

Basically, the mutex is around grow but memory length is used in bounds check and is read without the mutex. So it seems theoretically possible to malloc a pointer in a thread that called grow and read in a different one where bounds check fails.

Since this is only about the visibility of size to bounds checks, I believe this doesn't affect custom allocators as that part is managed by wazero.

internal/wasm/memory.go
```diff
@@ -267,6 +290,22 @@ func PagesToUnitOfBytes(pages uint32) string {
...
+	for {
+		if old := atomic.LoadUintptr(lenPtr); old >= length {
+			// Was already greater or equal.
+			break
+		} else if atomic.CompareAndSwapUintptr(lenPtr, old, length) {
+			// Successfully updated.
+			break
+		}
+	}
+}
```
mathetake, line 307: definitely need test

internal/wasm/memory.go
```diff
@@ -57,12 +58,22 @@ type MemoryInstance struct {
 	// waiters implements atomic wait and notify. It is implemented similarly to golang.org/x/sync/semaphore,
 	// with a fixed weight of 1 and no spurious notifications.
 	waiters sync.Map
+
+	allocator experimental.MemoryBuffer
```
mathetake, line 62: This is not `allocator`, but a buffer, which sounds confusing to me. I would suggest to rename it to `expBuffer`

```suggestion
	expBuffer experimental.MemoryBuffer
```

internal/wasm/memory.go
```diff
@@ -222,6 +232,20 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 	newPages := currentPages + delta
 	if newPages > m.Max || int32(delta) < 0 {
 		return 0, false
+	} else if m.allocator != nil {
+		buffer := m.allocator.Grow(MemoryPagesToBytesNum(newPages))
```
mathetake, line 236: This section also needs unit test.

internal/wasm/store.go
```diff
@@ -362,8 +363,13 @@ func (s *Store) instantiate(
 		return nil, err
 	}
 
+	var allocator experimental.MemoryAllocator
+	if ctx != nil {
+		allocator, _ = ctx.Value(ctxkey.MemoryAllocatorKey{}).(experimental.MemoryAllocator)
+	}
```
mathetake, line 369: We don't check ctx != nil elsewhere, so please simplify as 
```suggestion
	allocator := ctx.Value(ctxkey.MemoryAllocatorKey{}).(experimental.MemoryAllocator)
```

mathetake:
left some comments, but I think generally this looks like I was expecting to see. Only thing left it naming nits and unit tests.

ps I think this closes #1331

experimental/memory.go
ncruces: OK, that makes sense. Will revert to the simpler version.

internal/wasm/store.go
ncruces: Actually, there's a specific test case covering this:
https://github.com/tetratelabs/wazero/blob/f31be308fa39150af4939fec0e1f0bb005cd11f4/internal/wasm/module_instance_test.go#L87-L89

infogulch: nil contexts should not be common in Go. There's even a lint to use context.Background() instead of passing nil because a lot of places assume a valid context. It may be reasonable to change this test.

internal/wasm/memory.go
ncruces: @anuraaga convinced me to revert this, but will add a test for the new version anyway.

internal/wasm/memory.go
ncruces: No, at this point the memory has moved, the harm is done.
I'll document instead that if you want to use your allocator for shared memories, it can't move the memory.

anuraaga (commented):
Besides other comments this LGTM. Thanks!

internal/wasm/memory.go
ncruces: Done.

internal/wasm/module_instance.go
```diff
@@ -151,20 +151,24 @@ func (m *ModuleInstance) ensureResourcesClosed(ctx context.Context) (err error)
 	}
 
 	if sysCtx := m.Sys; sysCtx != nil { // nil if from HostModuleBuilder
-		if err = sysCtx.FS().Close(); err != nil {
-			return err
-		}
+		err = sysCtx.FS().Close()
```
ncruces, line 154: The early return here alarmed me, was it really intended?

ncruces:
> ps I think this closes #1331

It also closes #1150, IMO.

internal/wasm/memory_test.go
```diff
@@ -979,3 +993,25 @@ func requireChannelEmpty(t *testing.T, ch chan string) {
...
+type sliceBuffer struct {
+	buf []byte
+	max uint64
+}
+
+func (b *sliceBuffer) Free() {}
+
+func (b *sliceBuffer) Buffer() []byte { return b.buf }
+
+func (b *sliceBuffer) Grow(size uint64) []byte {
```
mathetake, line 1010: Please put comments on them like any other places in the code base
```suggestion
// Free implements ...
func (b *sliceBuffer) Free() {}

// Buffer implements ...
func (b *sliceBuffer) Buffer() []byte { return b.buf }

// Grow implements ...
func (b *sliceBuffer) Grow(size uint64) []byte {
```

mathetake (approved):

experimental/memory.go
```diff
@@ -0,0 +1,35 @@
...
+type MemoryBuffer interface {
+	// Return the backing []byte for the memory buffer.
+	Buffer() []byte
+	// Grow the backing memory buffer to size bytes in length.
+	// To back a shared memory, Grow can't change the address
+	// of the backing []byte (only its length/capacity may change).
+	Grow(size uint64) []byte
+	// Free the backing memory buffer.
+	Free()
+}
```
achille-roussel, line 26: Apologies because I'm a little late to the party, but since this is still in the `experimental` package it's maybe still OK to leave feedback; a few suggestions:

- Turn the memory allocator into an interface since it's intended to be an abstraction; single-method interfaces are a very common idiom in Go that developers are familiar with:
```go
type MemoryAllocator interface {
  Allocate(min, cap, max uint64) MemoryBuffer
}

// We can always add a helper to implement the interface on function values.
type MemoryAllocatorFunc func(min, cap, max uint64) MemoryBuffer
```

- The use of 3 uint64 as argument makes the signature error prone to programs that would mistakenly swap the position of these arguments. For example, my inuition would be for min/max to be back-to-back, I will for sure make this mistake at some point and the Go compiler cannot detect those kind of errors. One way to reduce this risk would be to introduce a configuration struct, which forces named arguments:
```go
type MemoryConfig struct {
  _ struct{} // forces named arguments
  Min uint64
  Max uint64
  Cap uint64
}
```
```go
// Allocation must always explicilty name parameters:
alloc(MemoryConfig{
  Max: 64 * 1024 * 1024,
})
```

- The `MemoryBuffer` type sound more like an region or linear memory, we have an opportunity to improve the name to be less generic I think ("memory buffer" can describe about anything in a program), for example:
```go
type LinearMemory interface {
  ...
}
```
```go
type MemoryArena interface {
   ...
}
```
etc...

experimental/memory.go
```diff
@@ -0,0 +1,35 @@
...
+// that will ever be requested.
+type MemoryAllocator func(min, cap, max uint64) MemoryBuffer
+
+// MemoryBuffer is a memory buffer that backs a Wasm memory.
+type MemoryBuffer interface {
+	// Return the backing []byte for the memory buffer.
+	Buffer() []byte
+	// Grow the backing memory buffer to size bytes in length.
+	// To back a shared memory, Grow can't change the address
+	// of the backing []byte (only its length/capacity may change).
```
achille-roussel, line 22: Does this requirement make it impossible to implement a version of this interface backed by the Go heap?

It seems like the `sliceBuffer` implementation using `append` would violate this constraint.

ncruces: No:
1. this requirement is **only** for _shared_ memories;
2. you're allowed to pre-allocate the `max` to avoid relocations if you need such behavior.

Note that my `mmap` [implementation](https://github.com/ncruces/go-sqlite3/blob/main/internal/util/alloc.go) _also_ reserves the `max` address space, because it depends on memory never moving.

PS: I can make this into a package (and, I guess, support Windows) if there's interest. It has interesting benefits. Makes startup/teardown slower, but can make growing memory faster, as it avoids copies. And with the trick of allocating 8GB, we could probably disable bounds checks altogether.

experimental/memory.go
ncruces: Thanks! I'm:
🤷 on the single method interface
👎 on the args struct (this is called in one place incident wazero, users implement it, don't call it)
👍 on renaming `MemoryBuffer`

But I'll do whatever. I'm just ecstatic that I was able to `mmap` files and do WAL after one year. 🥳

achille-roussel: Ah OK, your argument on the arg struct makes sense, if users won't be calling it directly the risk is low 👍

experimental/memory.go
achille-roussel: The API as-is is quite easy to misuse, the interface implementation has no way of knowing if it's being called to allocated a shared memory segment or not, and the user may also not know if the underlying memory is shared-ready or not.

In my experience, this calls for separating the concepts, maybe we can even get compile-time checks to ensure that the right kind of memory is being used:
```go
type MemoryAllocator interface {
  AllocateLinearMemory(...) LinearMemory
  AllocateSharedMemory(...) SharedMemory
}

// It could be useful for these two interfaces to share a common set of methods,
// but also differentiate so the compiler can guarantee that they're never misused.

type LinearMemory interface {
  ...
}

type SharedMemory interface {
  ...
}
```

ncruces: Yes, point taken on “the allocator doesn't know if the memory is shared.”

I thought of adding a parameter, but then decided on just documenting. An implementor can and should specify if memories move when grown. Mine don't regardless of sharing.

With Go duck-typing your suggested API only differentiates the “kind” of memory if we add a dummy method.

Note that shared memories can still grow, they just can't move when grown. With a large virtual address space it's perfectly feasible to reserve N×4GB of address space without committing any of it to memory, all on a 2GB “droplet" with no swap, and then organically decide at runtime which of the N gets 512MB when the other N-1 stay under 64MB.

achille-roussel: > I thought of adding a parameter, but then decided on just documenting. An implementor can and should specify if memories move when grown. Mine don't regardless of sharing.
>
> The allocator is configured at module instantiation, so you should know by then if you need shared memory. Same as I know I need a memory that I can mmap but this is not compile time checked.

That's a bit more reliance on developers "getting it right" that I'm usually comfortable with; assumptions change over time especially in an evolving world like WebAssembly, I'd err on the side of caution and figure out how to leverage the compiler and API to be more future-proof. That being said, that's not a hill I'm willing to die on either, things working is better than things being perfect, I'm just trying to emphasize that most developers aren't Nuno and will likely get things wrong if the API allows for it ;) 

> Also with Go duck-typing your suggested API only differentiates the “kind” of memory if we add a dummy method.

I don't know if I'm suggesting differentiating via a dummy method, I'll have to think more about it, maybe the differences between linear and shared are more profound and could be expressed in the API.

experimental/memory.go
```diff
@@ -0,0 +1,35 @@
...
+// to create a new MemoryBuffer, with the given specification:
+// min is the initial and minimum length (in bytes) of the backing []byte,
+// cap a suggested initial capacity, and max the maximum length
+// that will ever be requested.
+type MemoryAllocator func(min, cap, max uint64) MemoryBuffer
+
+// MemoryBuffer is a memory buffer that backs a Wasm memory.
+type MemoryBuffer interface {
+	// Return the backing []byte for the memory buffer.
+	Buffer() []byte
```
achille-roussel, line 19: Food for thoughts, this method could potentially be removed in favor of `Grow(0)`.

experimental/memory.go
ncruces: Future Nuno will thank you for it!
I'm all for expressing stuff with the type system, as long as it doesn't become clunky.

PR #2186, opened 2024-04-15
experimental: custom memory allocator API tweaks

ncruces:
Addresses post submit comments in #2177.

### DIFF
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -30,12 +30,12 @@ func (f MemoryAllocatorFunc) Allocate(cap, max uint64) LinearMemory {
 
 // LinearMemory is an expandable []byte that backs a Wasm linear memory.
 type LinearMemory interface {
-	// Grow the linear memory to size bytes in length.
+	// Reallocates the linear memory to size bytes in length.
 	//
 	// Notes:
-	//   - To back a shared memory, Grow can't change the address of the
+	//   - To back a shared memory, Reallocate can't change the address of the
 	//     backing []byte (only its length/capacity may change).
-	Grow(size uint64) []byte
+	Reallocate(size uint64) []byte
 	// Free the backing memory buffer.
 	Free()
 }
--- a/experimental/memory.go
+++ b/experimental/memory.go
@@ -7,17 +7,27 @@ import (
 )
 
 // MemoryAllocator is a memory allocation hook which is invoked
-// to create a new MemoryBuffer, with the given specification:
+// to create a new LinearMemory, with the given specification:
 // min is the initial and minimum length (in bytes) of the backing []byte,
 // cap a suggested initial capacity, and max the maximum length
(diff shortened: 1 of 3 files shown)


experimental/memory.go
```diff
@@ -7,17 +7,27 @@ import (
...
 
 // MemoryAllocator is a memory allocation hook which is invoked
-// to create a new MemoryBuffer, with the given specification:
+// to create a new LinearMemory, with the given specification:
 // min is the initial and minimum length (in bytes) of the backing []byte,
 // cap a suggested initial capacity, and max the maximum length
 // that will ever be requested.
-type MemoryAllocator func(min, cap, max uint64) MemoryBuffer
+type MemoryAllocator interface {
+	Allocate(min, cap, max uint64) LinearMemory
```
mathetake, line 15: Need comment.

experimental/memory.go
```diff
@@ -6,18 +6,34 @@ import (
...
+func (f MemoryAllocatorFunc) Allocate(min, cap, max uint64) LinearMemory {
+	return f(min, cap, max)
+}
 
-// MemoryBuffer is a memory buffer that backs a Wasm memory.
-type MemoryBuffer interface {
-	// Buffer returns the backing []byte for the memory buffer.
+// LinearMemory is an expandable []byte that backs a Wasm linear memory.
+type LinearMemory interface {
+	// Buffer returns the backing []byte for the linear memory.
```
achille-roussel, line 34: What were your thoughts on dropping this method in favor of Grow(0)?

ncruces: I don't mind either way. Smaller interfaces are good. OTOH we're not growing the memory.

I also thought grow might not return the buffer. Or… endless possibilities.

ncruces: This is, maybe, a little less intuitive. `Grow(size)` grows the memory to `size`.
So it's not `Grow(0)` it's `Grow(min)`.

And then it probably makes little sense to have `Allocate(min, cap, max uint64) LinearMemory`, rather than just `Allocate(cap, max uint64) LinearMemory`.

achille-roussel: Ah OK, `Grow` is maybe non-intuitive then because Go developers will probably be familiar with [`bytes.(*Buffer).Grow`](https://pkg.go.dev/bytes#Buffer.Grow) which takes the number of bytes to make space for, not the total size. Same for the [`memory.grow`](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Memory/Grow) instruction in WebAssembly.

How about renaming the method to `Resize`, or making it work like `bytes.(*Buffer).Grow`?

achille-roussel (approved):
🚢

ncruces:
New mmapped allocator:
https://github.com/ncruces/go-sqlite3/blob/future/internal/util/alloc.go

PR #2189, opened 2024-04-17
experimental: reflect custom allocated buffer capacity to internal buffer

anuraaga:
/cc @ncruces 

When trying to use the custom allocator interface, I was getting crashes since the capacity wasn't being applied to the internal `m.Buffer` itself

```
 [recovered]
	panic: runtime error: slice bounds out of range [::266304] with capacity 196608 (recovered by wazero)
wasm stack trace:
	wasi_snapshot_preview1.fd_write(i32,i32,i32,i32) i32
	.$21(i32,i32,i32) i32
	.$22(i32,i32,i32) i32
	.$31(i32,i32,i32)
	.$787(i32)
	.$783(i32,i32,i32)
	.$806(i32,i32,i32) i32

Go runtime stack trace:
goroutine 37 [running]:
runtime/debug.Stack()
	/opt/homebrew/Cellar/go/1.22.2/libexec/src/runtime/debug/stack.go:24 +0x6c
github.com/tetratelabs/wazero/internal/wasmdebug.(*stackTrace).FromRecovered(0xc000a0c798, {0x101482be0, 0xc000212048})
	/Users/anuraag/go/pkg/mod/github.com/tetratelabs/wazero@v1.7.1/internal/wasmdebug/debug.go:139 +0x1a8
github.com/tetratelabs/wazero/internal/engine/wazevo.(*callEngine).callWithStack.func1()
	/Users/anuraag/go/pkg/mod/github.com/tetratelabs/wazero@v1.7.1/internal/engine/wazevo/call_engine.go:256 +0x4b0
panic({0x101482be0?, 0xc000212048?})
	/opt/homebrew/Cellar/go/1.22.2/libexec/src/runtime/panic.go:770 +0x124
github.com/tetratelabs/wazero/internal/wasm.(*MemoryInstance).Read(0xc00013c7e0, 0x41030, 0x10)
	/Users/anuraag/go/pkg/mod/github.com/tetratelabs/wazero@v1.7.1/internal/wasm/memory.go:160 +0x104
github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1.writev({0x1014aedd0, 0xc00013c7e0}, 0x41030, 0x2, 0xc00002ece8)
```

### DIFF
--- a/internal/wasm/memory.go
+++ b/internal/wasm/memory.go
@@ -297,10 +297,10 @@ func PagesToUnitOfBytes(pages uint32) string {
 // Uses atomic write to update the length of a slice.
 func atomicStoreLengthAndCap(slice *[]byte, length uintptr, cap uintptr) {
 	slicePtr := (*reflect.SliceHeader)(unsafe.Pointer(slice))
-	lenPtr := (*uintptr)(unsafe.Pointer(&slicePtr.Len))
-	atomic.StoreUintptr(lenPtr, length)
 	capPtr := (*uintptr)(unsafe.Pointer(&slicePtr.Cap))
 	atomic.StoreUintptr(capPtr, cap)
+	lenPtr := (*uintptr)(unsafe.Pointer(&slicePtr.Len))
+	atomic.StoreUintptr(lenPtr, length)
 }
 
 // Uses atomic write to update the length of a slice.
--- a/internal/wasm/memory.go
+++ b/internal/wasm/memory.go
@@ -241,7 +241,7 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 			// We assume grow is called under a guest lock.
 			// But the memory length is accessed elsewhere,
 			// so use atomic to make the new length visible across threads.
-			atomicStoreLength(&m.Buffer, uintptr(len(buffer)))
+			atomicStoreLengthAndCap(&m.Buffer, uintptr(len(buffer)), uintptr(cap(buffer)))
 			m.Cap = memoryBytesNumToPages(uint64(cap(buffer)))
 		} else {
 			m.Buffer = buffer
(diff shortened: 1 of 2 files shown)


internal/wasm/memory.go
```diff
@@ -294,6 +294,15 @@ func PagesToUnitOfBytes(pages uint32) string {
 
 // Below are raw functions used to implement the api.Memory API:
 
+// Uses atomic write to update the length of a slice.
+func atomicStoreLengthAndCap(slice *[]byte, length uintptr, cap uintptr) {
```
anuraaga, line 298: Somehow duplicating with `atomicStoreLength` seemed cleanest of what I could come up with but happy to take suggestions

internal/wasm/memory_test.go
```diff
@@ -828,6 +828,14 @@ func Test_atomicStoreLength(t *testing.T) {
 	require.Equal(t, 15, len(slice))
 }
 
+func Test_atomicStoreLengthAndCap(t *testing.T) {
+	// Doesn't verify atomicity, but at least we're updating the correct thing.
+	slice := make([]byte, 10, 20)
+	atomicStoreLengthAndCap(&slice, 15, 20)
```
ncruces, line 834: Please change the capacity as well. Something like `atomicStoreLengthAndCap(&slice, 12, 18)`.

internal/wasm/memory.go
ncruces: Not sure if they both need to be atomic.
But if we're going to use atomic, which doesn't hurt, I'd switch order: first update `cap` then `len`.
This ensures it's impossible to observe `len>cap`.

ncruces (approved):
You're right! How did I miss this? Sorry this got to 1.7.1.

As a mitigation, while this is unreleased, I think your allocator can use fixed capacity.

Mine doesn't because bytes beyond `len` haven't been committed, and segfault if accessed, but this happens only if manually resliced, or if Wasm would access beyond `len`, neither of which should happen.

internal/wasm/memory_test.go
anuraaga: Oops, done

internal/wasm/memory.go
anuraaga: Ah makes sense

Issue #2198, opened 2024-04-30
Binary built with simd128 works with arm64 but fails with amd64


anuraaga:
**Describe the bug**

I am compiling [aho-corasick](https://github.com/BurntSushi/aho-corasick) which has some [code](https://github.com/BurntSushi/memchr/blob/20ef11fa92d8b393735f10906c436a8ce6e792a4/src/memmem/searcher.rs#L676) special-cased for wasm with simd. If I build with simd enabled, test cases that previously worked on both amd64 and arm64 fail on amd64, arm64 passes. This is a difference in behavior, not a crash.

This does not seem to be newly introduced with optimizing compiler.

**To Reproduce**

```
git clone https://github.com/anuraaga/go-aho-corasick.git --branch enablesimd
cd go-aho-corasick
GOARCH=amd64 go test ./...
```

From https://github.com/wasilibs/go-aho-corasick/pull/34

**Expected behavior**

Test passes

**Environment (please complete the relevant information):**
- Go version: [e.g. 1.19.1] 1.22.2
- wazero Version: 1.7.1 (also checked with 1.6.0, same behavior)
- Host architecture: amd64 (reproduces with Rosetta on M1 Mac)
- Runtime mode: compiler (interpreter works fine)

**Additional context**

Unfortunately this is not a crash, just the result is wrong. So I don't know if it's feasible to actually debug this 😓 But posting to at least track. No threads or anything, so at least the behavior is deterministic I think.

ncruces:
Just to confirm, interpreter passes the test, right?

anuraaga:
Yup, interpreter works fine, so only amd64 compiler is failing.

mathetake:
just went ahead and run the unit tests in the original Rust aho-corasick repo, and all passed 🤷‍♂️

mathetake:
well, I reproduced the error with the official unit tests with `--release` flag in `cargo test --no-run` 
```asm stack trace:
	.__rust_start_panic(i32,i32) i32
		0xa5a99: /rustc/ef71f1047e04438181d7cb925a833e2ada6ab390/library/panic_abort/src/lib.rs:96:17 (inlined)
		         /rustc/ef71f1047e04438181d7cb925a833e2ada6ab390/library/panic_abort/src/lib.rs:38:5
	.rust_panic(i32,i32)
		0xa5682: /rustc/ef71f1047e04438181d7cb925a833e2ada6ab390/library/std/src/panicking.rs:833:25
[...]

PR #2202, opened 2024-05-06
amd64: saves clobbered xmm8-xmm15 beyond runtime.memmove

mathetake:
fixes https://github.com/tetratelabs/wazero/issues/2198


mathetake:
@anuraaga could you double check the fix?

mathetake:
wait it's broken 😉 just a sec

mathetake:
ok looks like this is another bug than #2198😉

mathetake:
@anuraaga thank you for double checking

PR #2210, opened 2024-05-20
Revert tinygo reflect.SliceHeader specialization

ncruces:
In light of https://github.com/tinygo-org/tinygo/pull/4156, we need to revert #2161.

Also use `unsafe.Add` where feasible, and remove TODOs to use `unsafe.Slice` where doing so would raise a `go vet` warning about misuse of `unsafe.Pointer`.


ncruces:
Drafting while waiting for the TinyGo release, to update failed integration tests.

deadprogram:
We are in process of releasing v0.32 so this PR should be prepared, please.

ncruces:
Thanks for the heads up.

I think this is mostly ready, I just need to update CI to download the v0.32 binaries, and run tests.

ncruces:
Linux binaries still missing from [here](https://github.com/tinygo-org/tinygo/releases/tag/v0.32.0), will try again.

ncruces:
@deadprogram it seems TinyGo v0.32 lost the ability to build `--target=wasi` binaries with Go 1.20?

.github/workflows/examples.yaml
```diff
@@ -84,6 +84,7 @@ jobs:
 
       - name: Build TinyGo examples
         run: make build.examples.tinygo
+        if: matrix.go-version != '1.20' # fails with TinyGo v0.32.0
```
ncruces, line 87: Idk how we feel about skipping building tests with Go 1.20 + TinyGo v0.32.0, as that seems to be an unsupported combination.

mathetake (approved):
yeah looks like a bug in TinyGo that -target=wasi is now trying to use GOOS=wasip1, which is not existent in Go 1.20, for source resolution vs previously it didn't use wasip1 target internally (I remember it used arm target virtually). I think anyway this is harmless so I will go ahead and merge. cc @deadprogram my guess is that the latest TinyGo breaks wasi builds for all Go versions <= 1.20.

PR #2218, opened 2024-05-27
Maps ERROR_NOT_A_REPARSE_POINT to EINVAL

ncruces:
This fixes a Windows inconsistency where Readlink was returning EIO instead of EINVAL when the named file is not a symbolic link.

See:
https://man7.org/linux/man-pages/man2/readlink.2.html
https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--4000-5999-


mathetake (approved):
Great!

PR #2219, opened 2024-05-28
mmap: eliminates dead flag

mathetake:
platform.MmapSupported hasn't been used since a while


ncruces (approved):

mathetake:
@ncruces thank you for the look!

PR #2220, opened 2024-05-28
cache: includes cpu features in cache keys

mathetake:
there's a discussion about "offline" compilation, and if that happens, 
the CPU feature incompatibility becomes the problem.
This hardens the cache against it by including the cpu id in the file cache keys 
so that the content will be used exactly on the same CPUs.

Note that this impacts only amd64 target.

### DIFF
--- a/internal/platform/cpuid_amd64.go
+++ b/internal/platform/cpuid_amd64.go
@@ -59,4 +59,21 @@ func (f *cpuFeatureFlags) HasExtra(cpuFeature CpuFeature) bool {
 }
 
 // Raw implements the same method on the CpuFeatureFlags interface.
-func (f *cpuFeatureFlags) Raw() [2]uint64 { return [2]uint64{f.flags, f.extraFlags} }
+func (f *cpuFeatureFlags) Raw() uint64 {
+	// Below, we only set the first 4 bits for the features we care about,
+	// instead of setting all the unnecessary bits obtained from the CPUID instruction.
+	var ret uint64
+	if f.Has(CpuFeatureAmd64SSE3) {
+		ret = 1 << 0
+	}
+	if f.Has(CpuFeatureAmd64SSE4_1) {
+		ret |= 1 << 1
+	}
+	if f.Has(CpuFeatureAmd64SSE4_2) {
+		ret |= 1 << 2
+	}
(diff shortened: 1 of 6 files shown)


internal/platform/cpuid_amd64.go
```diff
@@ -48,12 +48,15 @@ func loadCpuFeatureFlags() CpuFeatureFlags {
...
 }
 
-// HasExtra implements the same method on the CpuFeatureFlags interface
+// HasExtra implements the same method on the CpuFeatureFlags interface.
 func (f *cpuFeatureFlags) HasExtra(cpuFeature CpuFeature) bool {
 	return (f.extraFlags & uint64(cpuFeature)) != 0
 }
+
+// Raw implements the same method on the CpuFeatureFlags interface.
+func (f *cpuFeatureFlags) Raw() [2]uint64 { return [2]uint64{f.flags, f.extraFlags} }
```
anuraaga, line 62: I wonder if it's too much effort to enumerate `CpuFeature` and return only those bits. Otherwise, it will be quite hard to have offline compilation work in practical environments, any architectural difference between CI and server machine would be incompatible (I think?). Of course, not crashing is more important so this would still be a fine change vs not.

anuraaga (approved):

internal/platform/cpuid_amd64.go
mathetake: I think I would like to have the input from @evacchi

mathetake: https://github.com/tetratelabs/wazero/pull/2220/commits/3de5bcb59d0bf62a2ef295b42fb866843bbc8612 you cool with this?

evacchi: looks good!

evacchi (approved):

anuraaga (approved):

mathetake:
thanks guys!

anuraaga:
@ncruces Thanks for mentioning the Go baseline - had the random thought that perhaps we shouldn't do any cupid detection. Since Go has `GOAMD64`, I think it is also a build tag. Maybe it's more "pure-Go" to just use the build tags instead of detection, wazero code is no better than Go code?

Anyways, I don't think it's an important, or even great, enhancement to implement, but maybe an idea for where there is a will and a want.

PR #2223, opened 2024-05-28
Disallow empty path on path_open

### LINKED ISSUE #2222: `path_open` with an empty path should fail
**Describe the bug**

Calling `path_open` on an empty path fails on host Linux as well as other runtimes (Wasmtime, Node, WAMR, WasmEdge, Wasmer). But Wazero succeeds and returns a fd.

As a reference, [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html) specifies that:

> A null pathname shall not be successfully resolved.

A rust example that uses `wasi` crate `0.11`.

```rust
fn main() {
    unsafe {
        let fd = wasi::path_open(
            3,
            0,
            "",
            0,
            wasi::RIGHTS_FD_READ | wasi::RIGHTS_FD_WRITE | wasi::RIGHTS_FD_SEEK,
            0,
            0,
        )
        .unwrap();

        eprintln!("fd: {fd}");
    }
}
```

**Expected behavior**
Wazero should error and not return a fd.

- Go version: go1.21.1 darwin/amd64
- wazero Version: d4a49031b671b59fb832c95741df63b96f144104
- Host architecture: amd64
- Runtime mode: [e.g. interpreter or compiler]

yagehu:
This commit makes `path_open` return `inval` when passed an empty path. This behavior is consistent with other Wasm runtimes (Wasmtime, Wasmer, Node, WAMR, WasmEdge).

fixes #2222

### DIFF
--- a/imports/wasi_snapshot_preview1/fs.go
+++ b/imports/wasi_snapshot_preview1/fs.go
@@ -1596,6 +1596,10 @@ func pathOpenFn(_ context.Context, mod api.Module, params []uint64) experimental
 		return errno
 	}
 
+	if pathLen == 0 {
+		return experimentalsys.EINVAL
+	}
+
 	fileOpenFlags := openFlags(dirflags, oflags, fdflags, rights)
 	isDir := fileOpenFlags&experimentalsys.O_DIRECTORY != 0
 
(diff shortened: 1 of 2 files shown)


evacchi (approved):
looks good!

PR #2244, opened 2024-06-10
wasi: `path_open` should accept a dir with RIGHT_FD_WRITE

### LINKED ISSUE #2243: Should not be able to open a directory with write
**Describe the bug**

`path_open`ing a directory with `rights::fd_write` should fail. Wazero does fail *if* the `oflags::directory` flag is not set. But will successfully open the directory and return a valid fd if the it is set.

**To Reproduce**

This Rust snippet can be compiled with Cargo and `wasi` crate v0.11. Wazero returns an FD while other runtimes (Wasmtime, Wasmer, WasmEdge, WAMR, and Node) all fails the call.

```rust
fn main() {
    unsafe {
        let base_fd = 3;
        let fd = wasi::path_open(
            base_fd,
            0,
            ".",
            wasi::OFLAGS_DIRECTORY,
            wasi::RIGHTS_FD_WRITE,
            0,
            0,
        )
        .unwrap();

        eprintln!("fd {fd}");
    }
}
```

- Go version: 1.22.3
- wazero Version: 5c8366f8d5f8f8e355c241c423b254eb1c10b02b
- Host architecture: amd64
- OS: Linux
- Runtime mode: compiler

yagehu:
This commit fixes a `path_open` behavior that allows a directory to be `path_open`ed with the `rights::fd_write` flag.  Since Wazero considers it to be an error to open a directory with the write flag normally, it is safe to treat this as a bug.  Although rights/capability system is being scrapped, we still use the `fd_read` and `fd_write` flag to determine if write permission needs to be set.

fixes #2243

### DIFF
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -3939,6 +3939,18 @@ func Test_pathOpen(t *testing.T) {
 <== (opened_fd=4,errno=ESUCCESS)
 `,
 		},
+		{
+			name:          "sysfs.DirFS file O_DIRECTORY RIGHTS_FD_WRITE",
+			fs:            writeFS,
+			path:          func(*testing.T) string { return fileName },
+			oflags:        wasip1.O_DIRECTORY,
+			rights:        wasip1.RIGHT_FD_WRITE,
+			expectedErrno: wasip1.ErrnoNotdir,
+			expectedLog: `
+==> wasi_snapshot_preview1.path_open(fd=3,dirflags=,path=file,oflags=DIRECTORY,fs_rights_base=FD_WRITE,fs_rights_inheriting=,fdflags=)
+<== (opened_fd=,errno=ENOTDIR)
+`,
+		},
 	}
 
 	for _, tt := range tests {
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -3951,6 +3951,18 @@ func Test_pathOpen(t *testing.T) {
 <== (opened_fd=,errno=ENOTDIR)
 `,
 		},
+		{
+			name:          "sysfs.DirFS dir O_DIRECTORY RIGHTS_FD_WRITE",
+			fs:            writeFS,
+			path:          func(*testing.T) string { return dirName },
+			oflags:        wasip1.O_DIRECTORY,
+			rights:        wasip1.RIGHT_FD_WRITE,
+			expectedErrno: wasip1.ErrnoIsdir,
+			expectedLog: `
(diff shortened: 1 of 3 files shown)


evacchi:
Interesting apparently this fails a test case in the zig stdlib!

yagehu:
> Interesting apparently this fails a test case in the zig stdlib!

I checked out main and the Zig stdlib test is still failing. Seems unrelated?

evacchi:
First, thanks anyway for the contribution!

Now, unfortunately I checked out both this branch and `main` and `main` succeeds while this branch reports 1 failed test. You can run Zig's test suite directly by installing `./cmd/wazero` and then:

    wazero run -mount . ./internal/integration_test/stdlibs/testdata/zig/test-opt.wasm

make sure you first build the zig test cases, see https://github.com/tetratelabs/wazero/blob/253c0346e4c8d7bd604304447233c1a3ce0a71f7/.github/workflows/integration.yaml#L77

Regardless, the PR should include a test for this case too.

yagehu:
Ahh OK it uses the built Wazero. I just had to rebuild wazero. Now I can repro. Thanks!

internal/sysfs/file.go
```diff
@@ -38,9 +38,6 @@ func NewStdioFile(stdin bool, f fs.File) (fsapi.File, error) {
 }
 
 func OpenFile(path string, flag experimentalsys.Oflag, perm fs.FileMode) (*os.File, experimentalsys.Errno) {
-	if flag&experimentalsys.O_DIRECTORY != 0 && flag&(experimentalsys.O_WRONLY|experimentalsys.O_RDWR) != 0 {
```
yagehu, line 41: This check is safe to remove because the write flag is now properly propagated to the underlying syscall instead of being masked by an early return in `func openFlags`.

yagehu: It is also necessary to remove this for correctness because this check sometimes erroneously return `isdir` when the target is not actually a directory.

evacchi: would you be able to add a (negative) test for when this happens, i.e. a case where `main` still returns `isdir` for something that is not a dir? so we won't erroneously restore the incorrect behavior in the future.

evacchi (commented):
just a minor change, overall looks fine!

imports/wasi_snapshot_preview1/fs_test.go
```diff
@@ -3937,6 +3937,30 @@ func Test_pathOpen(t *testing.T) {
 			expectedLog: `
 ==> wasi_snapshot_preview1.path_open(fd=3,dirflags=,path=file,oflags=,fs_rights_base=FD_READ|FD_WRITE,fs_rights_inheriting=,fdflags=)
 <== (opened_fd=4,errno=ESUCCESS)
+`,
+		},
+		{
+			name:          "sysfs.DirFS file O_DIRECTORY RIGHTS_FD_WRITE",
```
yagehu, line 3943: @evacchi This test ensures the erroneous `isdir` is not returned for regular file and the `O_DIRECTORY` flag is set.

evacchi (approved):

PR #2254, opened 2024-06-14
[1/2] Proper path resolution: add openat functionality

yagehu:
This PR is the first in a series that will implement the path resolution algorithm described by [wasi-filesystem](https://github.com/WebAssembly/wasi-filesystem/blob/main/path-resolution.md).  Since this new path resolution implementation requires the use of `openat`, this PR first introduces `openat(2)` and its Windows equivalent [`NtCreateFile`](https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntcreatefile) to Wazero.

- [x] *nix implementation with `openat(2)`
- [ ] Windows implementation with `NtCreateFile`
- [ ] Add more test cases to `internal/sysfs/dirfs_test.go`
- [ ] Add function docs

The next PR will stack on this one and use the `openat` functionality to implement path resolution.


experimental/sys/fs.go
```diff
@@ -58,6 +58,8 @@ type FS interface {
 	//     https://pubs.opengroup.org/onlinepubs/9699919799/functions/open.html
 	OpenFile(path string, flag Oflag, perm fs.FileMode) (File, Errno)
 
+	OpenFileAt(dir File, path string, flag Oflag, perm fs.FileMode) (File, Errno)
```
evacchi, line 61: I may be mistaken, but I don't think it is really necessary to expose this to the public interface, is it? can't we just adopt it internally for now?

yagehu: You're right. Removed from interface.

internal/sysfs/nt_windows.go
evacchi, line 1: most of the constants and I think some of the structs (e.g. ACL, unless I am mistaken) are not really used in the code; we can trim the constants we are not referencing and remove the structs we are not using (just replace them with generic nil pointers)

yagehu: Removed all unused constants and types.

yagehu:
FYI I can't stack PRs on top of this one because I can't push branches to this repo.

internal/sysfs/readfs.go
```diff
@@ -71,7 +72,7 @@ func (r *ReadFS) Utimens(path string, atim, mtim int64) experimentalsys.Errno {
 }
 
 // compile-time check to ensure readFile implements api.File.
-var _ experimentalsys.File = (*readFile)(nil)
+var _ fsapi.File = (*readFile)(nil)
```
evacchi, line 75: this was intentionally implementing only the experimental public VFS API. In fact I think this experimental VFS should not expose OpenAt

internal/fsapi/file.go
```diff
@@ -66,4 +70,22 @@ type File interface {
...
+	// # Notes
+	//
+	//   - This is like `openat` in POSIX.
+	//     See https://pubs.opengroup.org/onlinepubs/9699919799/functions/open.html
+	OpenAt(
+		fs experimentalsys.FS,
+		path string,
+		flag experimentalsys.Oflag,
+		mode fs.FileMode,
+	) (experimentalsys.File, experimentalsys.Errno)
```
evacchi, line 90: do we really need to expose this at this level?

the reason I am asking is that this obviously only applies to hierarchical FSs, but it won't work (as you can see later) on other types of file descriptors such as sockets or stdin/err/out. Could it be possible to keep this internal and limited to tree-like file systems?

Essentially, the final goal of this (series of) PR is to introduce an API that will reproduce the correct behavior when resolving paths, so it would great if we can scope it only to files for which "paths" make sense.

internal/sysfs/readfs.go
yagehu: Thanks for the context! I had thought it was an oversight.

internal/fsapi/file.go
yagehu: It definitely doesn't need to be exposed at this level.  I misunderstood your last comment and thought you wanted the method moved from the public interface to a private one.

evacchi: well, that was a fair assumption, but I was wondering if we could keep it as an implementation detail of Open for those FSs that need hierarchical file access 🤔

yagehu:
Quick update: I'm on a cross-country move at the moment. Will get around to this again after mid-August. Thanks for the speedy reviews @evacchi

Issue #2257, opened 2024-06-18
Panic when calling `ExportedFunction` with WASI function name


asubiotto:
**Describe the bug**
Panic when calling `ExportedFunction` on an instantiated WASI module. This might stem from a misunderstanding of how to use wazero.

**To Reproduce**
To reproduce run the following test:
```
func TestWasi(t *testing.T) {
	ctx := context.Background()
	wasiRuntime := wazero.NewRuntime(ctx)
	wasi_snapshot_preview1.MustInstantiate(ctx, wasiRuntime)
	wasiModule := wasiRuntime.Module(wasi_snapshot_preview1.ModuleName)
	_ = wasiModule.ExportedFunction("fd_write")
}
```

This panics with:
```
panic: runtime error: index out of range [8] with length 0 [recovered]
	panic: runtime error: index out of range [8] with length 0

goroutine 22 [running]:
testing.tRunner.func1.2({0x1016176e0, 0x140001b28a0})
	/opt/homebrew/opt/go/libexec/src/testing/testing.go:1631 +0x1c4
testing.tRunner.func1()
	/opt/homebrew/opt/go/libexec/src/testing/testing.go:1634 +0x33c
panic({0x1016176e0?, 0x140001b28a0?})
	/opt/homebrew/opt/go/libexec/src/runtime/panic.go:770 +0x124
github.com/tetratelabs/wazero/internal/engine/wazevo.(*moduleEngine).NewFunction(0x140001eec00, 0x1a)
	/Users/asubiotto/go/pkg/mod/github.com/tetratelabs/wazero@v1.7.2/internal/engine/wazevo/module_engine.go:187 +0x404
github.com/tetratelabs/wazero/internal/wasm.(*ModuleInstance).ExportedFunction(0x140000b6a00, {0x10120833b?, 0x1016ea000?})
	/Users/asubiotto/go/pkg/mod/github.com/tetratelabs/wazero@v1.7.2/internal/wasm/module_instance.go:210 +0x5c
```

**Expected behavior**
No panic an an `api.Function` object, or a clear error message as to why calling `ExportedFunction` in this case is not valid.

**Environment (please complete the relevant information):**
- Go version: `go version go1.22.4 darwin/arm64`
- wazero Version: `github.com/tetratelabs/wazero v1.7.2`
- Host architecture: `arm64`

mathetake:
[...]

PR #2259, opened 2024-06-20
doc: adds a note on direct use of host functions

mathetake:
I was under the impression that this was already written in the doc
somewhere, but turned out it wasn't. Thank you for reporting @asubiotto 

ref #2257

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -179,6 +179,9 @@ type HostFunctionBuilder interface {
 //     are deferred until Compile.
 //   - Functions are indexed in order of calls to NewFunctionBuilder as
 //     insertion ordering is needed by ABI such as Emscripten (invoke_*).
+//   - Because of the semantic of host function that it assumes the existence of "importing module" and it has access to
+//     the memory of the importing module for example, direct use of ExportedFunction interface is forbidden for host modules.
+//     Practically speaking, it is usually meaningless to directly call a host function from Go code as it is already somewhere in Go code.
 type HostModuleBuilder interface {
 	// Note: until golang/go#5860, we can't use example tests to embed code in interface godocs.
 


mathetake:
I will follow up in a separate PR to enforce this with panic.

mathetake:
I went ahead and implemented the enforcement of the rule

api/wasm.go
```diff
@@ -151,9 +151,13 @@ type Module interface {
...
-	// Note: The default wazero.ModuleConfig attempts to invoke `_start`, which
-	// in rare cases can close the module. When in doubt, check IsClosed prior
-	// to invoking a function export after instantiation.
+	// # Notes
+	//   - The default wazero.ModuleConfig attempts to invoke `_start`, which
+	//     in rare cases can close the module. When in doubt, check IsClosed prior
+	//     to invoking a function export after instantiation.
+	//   - Because of the semantic of host function that it assumes the existence of "importing module" and it has access to
+	//     the memory of the importing module for example, direct use of ExportedFunction is forbidden for host modules.
+	//     Practically speaking, it is usually meaningless to directly call a host function from Go code as it is already somewhere in Go code.
```
evacchi, line 160:
```suggestion
	//   - Because the semantics of host functions assumes the existence of an "importing module", and because the host function has access to
	//     the memory of the importing module, direct use of ExportedFunction is forbidden for host modules.
	//     Practically speaking, it is usually meaningless to directly call a host function from Go code as it is already somewhere in Go code.
```

nit, I think this slight rephrasing should be clearer. Same nit in builder.go

evacchi (approved):
One nit, otherwise 👍

mathetake:
tests failing hah

PR #2275, opened 2024-06-28
doc: adds comments on anonymous module names

mathetake:
Anonymous modules are introduced like a few years ago 
to allow multiple/concurrent instantiations from a single
binary. That was not documented enough well, so this
adds doc comments in the API.


ncruces:
I'm an “expert” (supposedly) and this tripped me multiple times already.

ncruces (approved):

mathetake:
Thanks for the look @ncruces

PR #2281, opened 2024-07-06
test: correctly makes all example tests runnable

mathetake:


sys/stat_example_test.go
mathetake, line 1: this is useless

PR #2290, opened 2024-07-19
Add support for Emscripten version 3.1.57

jerbob92:
Emscripten has removed some exported functions and replaced them with pure JS functions. They still reference native methods though, they just have a different name and the JS method is just an alias for the native function.

This PR adds the ability to add function aliases for exported Emscripten functions, since Emscripten renamed some exports on their side, see https://github.com/emscripten-core/emscripten/pull/21555

This change is backwards compatible.


mathetake (commented):
this is the backward incompatibility introduced by Emscripten, so I would prefer just remove the old names - the only promise we make is the API of our public Go package stability and this is different. in other words, this could potentially accumulate overtime and we have to remove the tech debt at some point future vs now

jerbob92:
@mathetake that would also be fine, but that does mean that any wasm file compiled with Emscripten < 3.1.57 will suddenly stop working after upgrading Wazero without a clear indication why.

mathetake:
yeah, that's unfortunate but as long as we put that in the release note, this should be good

jerbob92:
@mathetake I have removed the backwards compatible code. Can you tell me how you compile the .wat files? When I use wabt I get a different output, but perhaps that has to do with wabt options and/or version.

mathetake:
pushed the fix!

mathetake (approved):

PR #2298, opened 2024-08-05
Add experimental ImportResolver

### LINKED ISSUE #2294: Allow multiple instances of the same named import (e.g. javy_quickjs_provider_v2)

This is a more concrete follow up to #2293. Currently it's only possible to have one instance of a named import, so if you want multiple instances that requires the same import, you need to create multiple runtimes.

I guess there are many ways to go about this, one way could possibly be a new module config option, e.g.

```
wazero.NewModuleConfig().WithNamedImport("javy_quickjs_provider_v2", myQuickJsInstance)
```

I tested out 2 options:

1. One runtime per script (with a shared compilation cache)
2. Having the QuickJS runtime embedded in the JS script wasm.

The benchmark below shows the two options compared. The motivation in the case below is speeding up the relatively slow Katex by throwing more threads at it, but different scripts needing the QuickJS import will end up with the same problem.

Embedding the QuickJS runtime makes the startup much more effective, but 

1. The wasm files gets much larger (`katex.wasm` from 450 kb to 4.5 MB)
2. It makes it very unpractical if we want to embed the QuickJS runtime and have end users provide their plugin wasm files.

```bash
name                          old time/op    new time/op    delta
KatexStartStop/PoolSize1-10     19.2ms ± 6%    20.8ms ± 5%     ~     (p=0.057 n=4+4)
KatexStartStop/PoolSize8-10      116ms ± 4%      23ms ± 3%  -80.16%  (p=0.029 n=4+4)
KatexStartStop/PoolSize16-10     222ms ± 3%      32ms ± 5%  -85.58%  (p=0.029 n=4+4)

name                          old alloc/op   new alloc/op   delta
KatexStartStop/PoolSize1-10     12.2MB ± 0%    11.4MB ± 0%   -7.01%  (p=0.029 n=4+4)
KatexStartStop/PoolSize8-10     97.9MB ± 0%    49.5MB ± 0%  -49.43%  (p=0.029 n=4+4)
KatexStartStop/PoolSize16-10     196MB ± 0%      93MB ± 0%  -52.46%  (p=0.029 n=4+4)

name                          old allocs/op  new allocs/op  delta
KatexStartStop/PoolSize1-10      18.1k ± 0%     25.8k ± 0%  +42.74%  (p=0.029 n=4+4)
KatexStartStop/PoolSize8-10       144k ± 0%       32k ± 0%  -77.61%  (p=0.029 n=4+4)
KatexStartStop/PoolSize16-10      289k ± 0%       40k ± 0%  -86.20%  (p=0.029 n=4+4)
```

bep:
If set in context, the ImportResolver will be used as the first step in resolving imports.

Closes #2294


mathetake:
yeah this actually makes sense.

mathetake (approved):
💯

PR #2304, opened 2024-08-20
ci: Updates TinyGo to 0.33

mathetake:


mathetake:
image

ok so i will drop TinyGo build CI - we can't send anytime on this

PR #2306, opened 2024-08-22
fix: add wrapper for osfile.Truncate for TinyGo 0.33

deadprogram:
This PR adds a wrapper for the `osfile` `Truncate()` functions to restore ability to build Wazero on bare metal using TinyGo 0.33

UPDATE: I have added a second commit to this PR that adds back the TinyGo 0.33 to CI, which also should demonstrate that this fix does what it is supposed to.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -152,6 +152,20 @@ jobs:
         # This runs all tests compiled above in sequence. Note: This mounts /tmp to allow t.TempDir() in tests.
         run: find . -name "*.test" | xargs -Itestbin docker run --platform linux/${{ matrix.arch }} -v $(pwd)/testbin:/test -v $(pwd)/wazerocli:/wazero -e WAZEROCLI=/wazero --tmpfs /tmp --rm -t wazero:test
 
+  test_tinygo:
+    name: "TinyGo on Ubuntu"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:  # Use version consistent with TinyGo.
+          go-version: "1.23"
+      - uses: acifani/setup-tinygo@v2
+        with:
+          tinygo-version: "0.33.0"
+      - run: tinygo build ./cmd/wazero
+      - run: tinygo build -size short -target pico -stack-size=8kb ./cmd/wazero
+
   # This ensures that internal/integration_test/fuzz is runnable, and is not intended to
   # run full-length fuzzing while trying to find low-hanging frontend bugs.
   fuzz:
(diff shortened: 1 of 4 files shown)


evacchi (approved):
I think this should be fine @mathetake @ncruces any comments?

achille-roussel (commented):
@deadprogram is the absence of a `Truncate` method on `*os.File` by design in Tinygo?

deadprogram:
I think that it is one of the various file APIs that do not have any way to be resolved when running baremetal without any file system. Similar to what is being done in code such as https://github.com/tetratelabs/wazero/blob/main/internal/sysfs/datasync_tinygo.go

ncruces:
> I think that it is one of the various file APIs that do not have any way to be resolved when running baremetal without any file system. Similar to what is already being done in code such as https://github.com/tetratelabs/wazero/blob/main/internal/sysfs/datasync_tinygo.go

Well, no. It's not like `datasync`, as that's a Linux system call. Arguably the TinyGo fall back for that one is wrong, because the appropriate fallback is to call `File.Sync`.

The question that should be asked, IMO, is why isn't a fallback for `Truncate` not making sense on devices without a filesystem, to return `ENOSYS` in TinyGo? Why should every user of TinyGo work around this?

mathetake:
Please fix this in TinyGo and not by adding some workaround here.

mathetake:
Please fix this in TinyGo and not by adding some workaround here.

mathetake:
What you are trying to add here is like "We introduced a breaking change/bug in TinyGo 0.33, so please adda a workaround in wazero", and to be honest, that's not right at all. I will NEVER accept this.

mathetake:
And note thatwe clearly stated multiple times that interpreter is not the scope of this project, and let alone compiling support with TinyGo. When introduced the CI job, we said it's best effort and the last CI failure was the last straw. We are trying to support TinyGo as a compilation target, which is not a goal of this project in the first place.

ncruces:
I honestly think it's reasonable for us to have fallbacks for all `syscall` stuff, and put a reasonable effort to have the interpreter work with a current Go version standard library (latest minus 2, as per the support policy), no trickery, no unsafe stuff, etc.

I personally put in a moderate (small, actually) amount of effort to support z/OS, for instance (a big endian platform that's not supported out of the box by Go, but which has a toolchain, etc). It almost worked out-of-the-box, and it passed all wazero tests, and all SQLite tests.

But it doesn't make sense for *us* to add workarounds for the standard library being broken on *our* end.

If this is interesting to you, CI can be setup on your end; it doesn't need to break us moving forward.

And I guess compatibility fixes that have the interpreter work with the Go standard library _will_ have my upvote. But not this.

deadprogram:
Just to be clear: this was an attempt to address compiling Wazero using TinyGo for embedded targets, not to address using a WASM module compiled using TinyGo with wazero as host on a full OS.

In any case, no problem. Thanks for everyone's feedback.

PR #2311, opened 2024-09-06
Fix descriptor.Table buffer growth calc

emcfarlane:
Fixes InsertAt size calculation to avoid growing slices more than needed. On insert the mask should calculate the index in the slice not the number of elements in the table. Avoids a factor of 8 increase in table size. Small optimisation to use the capcity of the slice to reduce the average length whilst keeping allocations down. Clarified the grow method grows the table by n * 64 items.


internal/descriptor/table.go
```diff
@@ -109,10 +101,10 @@ func (t *Table[Key, Item]) InsertAt(item Item, key Key) bool {
 	if key < 0 {
 		return false
 	}
-	if diff := int(key) - t.Len(); diff > 0 {
```
emcfarlane, line 112: For example, with 3 items in the table calling `InsertAt(item, 64)` would call `grow(61)` growing the internal slices of masks and items to 61 and 3904. Should be 2 and 128.

mathetake (commented):
add unit tests that cover the fix?

emcfarlane:
@mathetake added a testcase which shows the error and panic on main on edgecases. Renamed the file but its causing a large diff then intended, let me know if you;d like a diff format,

internal/descriptor/table_sys_test.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, found := range []struct {
+		key int32
+		ok  bool
+	}{
+		{key: k0, ok: k0Found},
+		{key: k1, ok: k1Found},
+		{key: k2, ok: k2Found},
+	} {
+		if !found.ok {
+			t.Errorf("key not found while ranging over table: %v", found.key)
```
mathetake, line 80: could you please take a look at other places in the code base and use require?

internal/descriptor/table_sys_test.go
```diff
@@ -0,0 +1,136 @@
...
+	for i := 0; i < b.N; i++ {
+		table.Insert(entry)
+
+		if (i % 65536) == 0 {
+			table.Reset() // to avoid running out of memory
+		}
+	}
+}
+
+func BenchmarkFileTableLookup(b *testing.B) {
```
mathetake, line 114: maybe we can delete this

internal/descriptor/table_sys_test.go
```diff
@@ -0,0 +1,136 @@
...
+		if _, ok := table.Lookup(deletion.key); ok {
+			t.Errorf("item found after deletion of '%v'", deletion.key)
+		}
+		if n, want := table.Len(), 3-(i+1); n != want {
+			t.Errorf("wrong table length after deletion: want=%d got=%d", want, n)
+		}
+	}
+}
+
+func BenchmarkFileTableInsert(b *testing.B) {
```
mathetake, line 101: also this

internal/descriptor/table_sys_test.go
emcfarlane: This was already here.

internal/descriptor/table_sys_test.go
emcfarlane: This was already here.

internal/descriptor/table_sys_test.go
emcfarlane: This was already here.

mathetake: hah, then while there could you fix?

mathetake:
sorry what was the intent for renaming the file and adding another test? I would prefer make it in the existing _test.go

emcfarlane:
Its a `package descriptor_test` as it imports `sys` to avoid a cycle. Moved it to `_sys_test.go` to make it more clear not sure if theres a pattern I can follow?

mathetake:
hah i see

mathetake:
ok i will defer to @achille-roussel for the rest. Thanks for the work here!

mathetake:
pushed two commits to minimize the change - used the export_test pattern

internal/descriptor/table.go
```diff
@@ -37,23 +37,19 @@ func (t *Table[Key, Item]) Len() (n int) {
...
-	if n > len(t.masks) {
-		masks := make([]uint64, n)
-		copy(masks, t.masks)
-
-		items := make([]Item, n*64)
-		copy(items, t.items)
+	total := len(t.masks) + n
+	if total > cap(t.masks) {
+		t.masks = append(t.masks, make([]uint64, n)...)
+	}
```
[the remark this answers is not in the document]
ncruces: Oh. Exciting, that's on 1.21. Is there any benefit besides readability? There are more _important_ places we use this in. I see it's not done inside the runtime, but if the pattern used by the method is intrisified that'd be great.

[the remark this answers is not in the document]
achille-roussel: A bit more safety, and avoids reallocation if the original capacity is large enough:
```go
func Grow[S ~[]E, E any](s S, n int) S {
	if n < 0 {
		panic("cannot be negative")
	}
	if n -= cap(s) - len(s); n > 0 {
		s = append(s[:cap(s)], make([]E, n)...)[:len(s)]
	}
	return s
}
```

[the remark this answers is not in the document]
ncruces: The recognized pattern is sufficiently general, it seems: https://golang.org/cl/109517

To be clear: I support making this change here.

I'm just wondering about memory grow as well, now that I know we can use this.

ncruces (approved):

PR #2314, opened 2024-09-14
compiler: ResolveImportedFunction should infer the typeID from the importing module

### LINKED ISSUE #2313: Running into `wasm error: indirect call type mismatch` when using the compiler backend, but not the interpreter
**Describe the bug**
I am running a wasm module that uses `call_indirect` - when using the interpreter it runs successfully, but when using the compiler it fails with `wasm error: indirect call type mismatch` - it also runs successfully on wasmtime, so I think this could be a compiler bug.

**To Reproduce**
I've uploaded the wasm here: [wasm.zip](https://github.com/user-attachments/files/16985989/wasm.zip)

Then using the extism CLI: https://github.com/extism/cli/ (can be installed by cloning the repo and running `go install ./extism` from the root of the repo)
```
EXTISM_ENABLE_WASI_OUTPUT=1 extism call ./plugin.wasm req --allow-host '*' --input "https://extism.org" --wasi --link import_module=import_module.wasm --log-level info
```

Note: the current CLI uses wazero 1.7, but I've also tested using the latest code on `main` with the same result.

**Expected behavior**
The call should return successfully instead of throwing the type mismatch error

- Go version: 1.22.6
- wazero Version:tested on 1.7 and the latest main (`58488880a334e8bda5be0d715a24d8ddeb34c725`)
- Host architecture: amd64
- Runtime mode: compiler

**Additional context**
If I comment out lines 3558-3581 in `internal/engine/wazevo/frontend/lower.go` it works as expected.

evacchi:
Fixes #2313.

This changes the signature for `ResolveImportedFunction()` so that it takes an additional index into the typeID section of the "current" module, i.e. the subject of the import.

The current behavior is to search for the import recursively into the chain of dependencies; however, this resolves to a typeID that is specific to _that_ module. For instance:

- consider the case of a module `m` declaring an import `m2inst.sym`.
- let be `m2inst.sym` an import with type `t` with ID `tid` 
- let be `m2inst` be linked to an instance of module `m2` with type `t2` and ID `t2id`.

Now: we want `t` to be compatible with `t2`; this we verify at:

https://github.com/tetratelabs/wazero/blob/6bb9899ea4c745a4ae23179a34fac0540c2ce551/internal/wasm/store.go#L441-L447

The current behavior is for `ResolveImportedFunction()` to resolve recursively a reference to a function so that we can then jump precisely to its address at runtime; however, we should _not_ carry on its type ID, and instead use the type ID of the import.

In other words, even if the types are compatible, we should be still referring to type `t` with ID `tid` from `m`, and _not_ type ID `t2id`, because, of course, the index `t2id` might be different from the local index `tid`.

The solution is to add another `descFunc Index` (from the name of the field `Import.DescFunc`) to the signature of  `ResolveImportedFunction()`, so that the typeID is correctly resolved from `moduleEngine.module.TypeIDs[descFunc]` (i.e. the typeID defined in the module that is requiring the imported function).

The tests are adjusted accordingly; notably, this also means that, in order for `m.ResolveImportedFunction()` to work properly, `m` must be correctly initialized with an `m.module` being the module that is performing the import, so the initialization of `moduleEngine` is pushed down, after the declaration of the "`imported`" and "`importing`" modules.

Let me know if you think it is worth adding more tests!

@zshipko please double-check that this is fixing your issue (I have slightly changed my original approach)
@ncruces would you like to double-check this against your sqlite lib?
@anuraaga if you spare any time, would you take a look too? :P

### DIFF
--- a/internal/engine/interpreter/interpreter.go
+++ b/internal/engine/interpreter/interpreter.go
@@ -487,7 +487,7 @@ func (e *engine) setLabelAddress(op *uint64, label label, labelAddressResolution
 }
 
 // ResolveImportedFunction implements wasm.ModuleEngine.
-func (e *moduleEngine) ResolveImportedFunction(index, indexInImportedModule wasm.Index, importedModuleEngine wasm.ModuleEngine) {
+func (e *moduleEngine) ResolveImportedFunction(index, descFunc, indexInImportedModule wasm.Index, importedModuleEngine wasm.ModuleEngine) {
 	imported := importedModuleEngine.(*moduleEngine)
 	e.functions[index] = imported.functions[indexInImportedModule]
 }
--- a/internal/engine/wazevo/module_engine.go
+++ b/internal/engine/wazevo/module_engine.go
@@ -237,7 +237,7 @@ func (m *moduleEngine) putLocalMemory() {
 }
 
 // ResolveImportedFunction implements wasm.ModuleEngine.
-func (m *moduleEngine) ResolveImportedFunction(index, indexInImportedModule wasm.Index, importedModuleEngine wasm.ModuleEngine) {
+func (m *moduleEngine) ResolveImportedFunction(index, descFunc, indexInImportedModule wasm.Index, importedModuleEngine wasm.ModuleEngine) {
 	executableOffset, moduleCtxOffset, typeIDOffset := m.parent.offsets.ImportedFunctionOffset(index)
 	importedME := importedModuleEngine.(*moduleEngine)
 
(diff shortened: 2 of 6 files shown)


zshipko:
Just tested this and verified it fixes the issue on my end! :partying_face: 

Thank you!

ncruces:
Everything seems to [work fine](https://github.com/ncruces/go-sqlite3/actions/runs/10870979374).

ncruces (approved):

mathetake (approved):

PR #2315, opened 2024-09-14
Allow custom memory allocator Reallocate to fail

ncruces:
Hey! 👋

This is just an idea, but the change was small enough that I decided to just try it.

How would we feel about allowing the custom allocator to return `nil` to signal it can't grow the linear memory?

In Go, a failure to allocate memory is a fatal condition, so it makes sense that we don't think about this when using `append` to grow the linear memory.

But custom allocators can handle this, it doesn't need to be a fatal condition. Also Wasm modules can deal with it: e.g. for `wasi-lib`, if `sbrk` fails, `malloc` just returns `NULL`, and (e.g.) SQLite can deal with that by freeing caches and trying again (if configured to do so).

cc: @anuraaga

### DIFF
--- a/internal/wasm/memory.go
+++ b/internal/wasm/memory.go
@@ -239,14 +239,13 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 		return currentPages, true
 	}
 
-	// If exceeds the max of memory size, we push -1 according to the spec.
 	newPages := currentPages + delta
 	if newPages > m.Max || int32(delta) < 0 {
 		return 0, false
 	} else if m.expBuffer != nil {
 		buffer := m.expBuffer.Reallocate(MemoryPagesToBytesNum(newPages))
-		// If the allocator fails to grow, we push -1 according to the spec.
 		if buffer == nil {
+			// Allocator failed to grow.
 			return 0, false
 		}
 		if m.Shared {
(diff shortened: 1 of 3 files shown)


mathetake:
Sounds reasonable to me

ncruces:
OK. I'll add a test and document the behavior in the interface.

internal/wasm/memory.go
```diff
@@ -244,6 +245,10 @@ func (m *MemoryInstance) Grow(delta uint32) (result uint32, ok bool) {
 		return 0, false
 	} else if m.expBuffer != nil {
 		buffer := m.expBuffer.Reallocate(MemoryPagesToBytesNum(newPages))
+		// If the allocator fails to grow, we push -1 according to the spec.
```
anuraaga, line 248: I'm probably missing something but where is the -1 here?

mathetake: same comment

ncruces: I just copied the comment: 
https://github.com/tetratelabs/wazero/blob/20de8b01b5183c29a44ee0bd13ee0cbb15210f1d/internal/wasm/memory.go#L242-L250

But that's actually what happens [one layer deep](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Memory/Grow).

mathetake: heh could you remove the comment

mathetake (approved):

internal/wasm/memory.go
ncruces: Sure. Both?

mathetake: yeah

PR #2319, opened 2024-09-23
sysfs: reopening file doesn't update fd

### LINKED ISSUE #2318: fd_fdstat_set_flags can silently change a fd's number
**Describe the bug**
*osFile.reopen doesn't set f.fd, leaving the file with potentially the wrong file descriptor.

**To Reproduce**
```
package main

import (
        "io"
        "log"
        "os"
        "syscall"
)

func main() {
        f1, err := os.Open("/file")
        if err != nil {
                log.Fatal(err)
        }
        f2, err := os.Open("/file")
        if err != nil {
                log.Fatal(err)
        }
        f1.Close()
        syscall.SetNonblock(int(f2.Fd()), true)
        if _, err := io.ReadAll(f2); err != nil {
                log.Fatal(err)
        }
}
```

```
$ echo foo > file
$ wazero run -mount=.:/ main.wasm
2024/09/20 23:50:44 read /file: Bad file number
```

This program should read from "file", but the read fails because f2's FD has changed. (The "syscall.SetNonblock" is necessary to trigger the correct write path here.)

I initially thought this might be exploitable to write to files in read-only mapped directories, but that doesn't seem to be the case. You can write to an unexpected writeable file, though.

- Go version: master
- wazero Version: master
- Host architecture: darwin_arm64

ncruces:
Fixes #2318.

Reopening the file was erroneously not updating the `fd` field.
Also, reopening is risky, because the file have been deleted/renamed (yes, even on Windows, because we're using `FILE_SHARE_DELETE`).

In theory, on POSIX, we'd never need to reopen (hence why WASI provides this, it tries to offer much of what POSIX allows, neglecting that lots of it is hard to emulate). But... reopen is only is only really used in two places: seeking directories (to rewind them), and setting the file's append flag.

For directory rewind, it's _already_ just a fallback:
https://github.com/tetratelabs/wazero/blob/6bb9899ea4c745a4ae23179a34fac0540c2ce551/internal/sysfs/osfile.go#L196-L209

If seeking works, it's used. So, no point in trying to "fix" that one.

For setting the append flag, reopen is always used, for good reason:
https://github.com/tetratelabs/wazero/blob/6bb9899ea4c745a4ae23179a34fac0540c2ce551/internal/sysfs/osfile.go#L78-L92

That link is still valid in [1.23](https://github.com/golang/go/blob/go1.23.1/src/os/file_unix.go). Go's `os.File` caches this, most importantly, to return a specific error on `WriteAt`. There's a good reason for this. The Open Group specifies that `pwrite` should work with, and ignore, `O_APPEND`, but Linux does the opposite and appends, which is a [bug](https://linux.die.net/man/2/pwrite).

So (IMO) we do need to reopen for `SetAppend`, and the best we can do, is check if the new file is still the same file, and error out otherwise.

That's what this PR does for `osFile`:
1. fix the `fd` not being updated issue
2. check that reopen gets the same file

For `fsFile`, this is only used to rewind directories, so:
1. own that
2. check that the reopened file is still a directory

### DIFF
--- a/internal/sysfs/file.go
+++ b/internal/sysfs/file.go
@@ -418,13 +418,6 @@ func seek(s io.Seeker, offset int64, whence int) (int64, experimentalsys.Errno) 
 	return newOffset, experimentalsys.UnwrapOSError(err)
 }
 
-// reopenFile allows re-opening a file for reasons such as applying flags or
-// directory iteration.
-type reopenFile func() experimentalsys.Errno
-
-// compile-time check to ensure fsFile.reopen implements reopenFile.
-var _ reopenFile = (*fsFile)(nil).reopen
-
 // reopen implements the same method as documented on reopenFile.
 func (f *fsFile) reopen() experimentalsys.Errno {
 	_ = f.close()
--- a/internal/sysfs/file.go
+++ b/internal/sysfs/file.go
@@ -269,7 +269,7 @@ func (f *fsFile) Readdir(n int) (dirents []experimentalsys.Dirent, errno experim
 
 	if f.reopenDir { // re-open the directory if needed.
 		f.reopenDir = false
-		if errno = adjustReaddirErr(f, f.closed, f.reopen()); errno != 0 {
+		if errno = adjustReaddirErr(f, f.closed, f.rewindDir()); errno != 0 {
 			return
 		}
 	}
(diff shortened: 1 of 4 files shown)


internal/sysfs/file.go
```diff
@@ -418,13 +418,6 @@ func seek(s io.Seeker, offset int64, whence int) (int64, experimentalsys.Errno)
 	return newOffset, experimentalsys.UnwrapOSError(err)
 }
 
-// reopenFile allows re-opening a file for reasons such as applying flags or
-// directory iteration.
-type reopenFile func() experimentalsys.Errno
```
ncruces, line 423: This wasn't being used anywhere. The compile time check is not very useful, tbh.

internal/sysfs/file.go
```diff
@@ -418,19 +418,25 @@ func seek(s io.Seeker, offset int64, whence int) (int64, experimentalsys.Errno)
...
-// compile-time check to ensure fsFile.reopen implements reopenFile.
-var _ reopenFile = (*fsFile)(nil).reopen
-
-// reopen implements the same method as documented on reopenFile.
-func (f *fsFile) reopen() experimentalsys.Errno {
-	_ = f.close()
-	var err error
-	f.file, err = f.fs.Open(f.name)
-	return experimentalsys.UnwrapOSError(err)
+func (f *fsFile) rewindDir() experimentalsys.Errno {
```
ncruces, line 421: `fsFile` only really supports reopening dirs, to rewind them.
Own that, check that the reopened file is still a dir.

internal/sysfs/osfile.go
```diff
@@ -83,21 +83,12 @@ func (f *osFile) SetAppend(enable bool) (errno experimentalsys.Errno) {
 		f.flag &= ^experimentalsys.O_APPEND
 	}
 
-	// Clear any create or trunc flag, as we are re-opening, not re-creating.
-	f.flag &= ^(experimentalsys.O_CREAT | experimentalsys.O_TRUNC)
```
ncruces, line 87: Don't needlessly update the file/flags.

mathetake (approved):

mathetake:
awesome detective work!

PR #2322, opened 2024-09-25
sysfs: clean paths lexically before use

### LINKED ISSUE #2321: File operations do not defend against symlink traversal, permitting trivial escape from read-write mounts
(I have been asked to report this bug via the public issue tracker.)

wazero does not appear to defend against symlink traversal in its filesystem APIs.

Since the WASI filesystem API permits creating symlinks, this allows for trivial access to the full filesystem from any program with a read-write mount. For example:

```
package main

import (
	"fmt"
	"io"
	"log"
	"os"
)

func main() {
	if err := os.Symlink("../../../../../../etc/passwd", "a"); err != nil {
		log.Fatal(err)
	}
	f, err := os.Open("a")
	if err != nil {
		log.Fatal(err)
	}
	b, _ := io.ReadAll(f)
	fmt.Println(string(b))
}
```

Run this program with read-write access to some directory:

```
wazero run -mount=restricted:/ main.wasm
```

It prints out the contents of /etc/passwd. /etc/passwd hasn't been exciting for a couple decades, admittedly, but this can be used to read or write any file anywhere on the filesystem that the account running wazero has access to.

ncruces:
This is intended to close #2321.

W.r.t. security, after consulting the other maintainers, we must say that fully sandboxing our WASI implementation is beyond the current scope of the project. This will be made explicit in the documentation.

We consider Wasm a mature specification, and security issues against our implementation of it will be taken seriously. WASI is not as mature, nor is our implementation of WASI preview 1.

We also provide advanced enough APIs that you can use to implement your own, better locked down, alternative to our WASI implementation.

That said, this PR still sides with breaking POSIX path resolution semantics in WASI vs. allowing trivial root mount escapes, by using `path.Clean` to lexically clean all paths before use.

This is a judgement call.

Handling paths lexically is easier to reason about, easier to port, and idiomatic Go. Go provides  `path.Clean` and `filepath.Clean`; `fs.FS` is even stricter disallowing `..` entirely. It's also unlikely to break well, behaving, useful applications.

### DIFF
--- a/internal/sysfs/dirfs.go
+++ b/internal/sysfs/dirfs.go
@@ -3,6 +3,8 @@ package sysfs
 import (
 	"io/fs"
 	"os"
+	"path"
+	"strings"
 
 	experimentalsys "github.com/tetratelabs/wazero/experimental/sys"
 	"github.com/tetratelabs/wazero/internal/platform"
@@ -12,7 +14,7 @@ import (
 func DirFS(dir string) experimentalsys.FS {
 	return &dirFS{
 		dir:        dir,
-		cleanedDir: ensureTrailingPathSeparator(dir),
+		cleanedDir: strings.TrimSuffix(dir, "/"),
 	}
 }
 
@@ -84,16 +86,17 @@ func (d *dirFS) Utimens(path string, atim, mtim int64) experimentalsys.Errno {
 	return utimens(d.join(path), atim, mtim)
 }
 
-func (d *dirFS) join(path string) string {
-	switch path {
-	case "", ".", "/":
-		if d.cleanedDir == "/" {
-			return "/"
+func (d *dirFS) join(name string) string {
+	last := strings.HasSuffix(name, "/")
+	name = path.Clean("/" + name)
+	if name == "/" {
+		if d.cleanedDir != "" {
+			return d.cleanedDir
 		}
-		// cleanedDir includes an unnecessary delimiter for the root path.
-		return d.cleanedDir[:len(d.cleanedDir)-1]
+		return "/"
 	}
-	// TODO: Enforce similar to safefilepath.FromFS(path), but be careful as
-	// relative path inputs are allowed. e.g. dir or path == ../
-	return d.cleanedDir + path
+	if last {
+		return d.cleanedDir + name + "/"
+	}
+	return d.cleanedDir + name
 }
--- a/internal/sysfs/dirfs.go
+++ b/internal/sysfs/dirfs.go
@@ -87,16 +87,20 @@ func (d *dirFS) Utimens(path string, atim, mtim int64) experimentalsys.Errno {
 }
 
 func (d *dirFS) join(name string) string {
-	last := strings.HasSuffix(name, "/")
-	name = path.Clean("/" + name)
-	if name == "/" {
+	// Lexically clean `name` to enforce that it always stays within the root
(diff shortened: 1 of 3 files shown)


anuraaga:
Sorry for coming to this so late, was just randomly thinking about this for some reason even before seeing this new PR.

I think to correctly handle symlinks, we would need to recognize them when opening/statting and use the resolved name recursively through the WASI apis rather than directly opening, likely with a recursion limit if not real cycle detector though that's likely too much.

I think cleaning paths still doesn't prevent the case where a symlink outside the sandbox is present in the mounted files. Naturally this is a far smaller issue than being to create the symlink in wasm itself.

I understand that the WASI implementation won't be fully hardened which makes sense, so the above might be going too far. But just wanted to share in case it didn't seem too hard, but probably is.

anuraaga:
Sorry I noticed the comment is about 3) in the issue, and is probably too much so no worries

ncruces:
No worries. Yeah, that sounds like doing path resolution ourselves, maybe with some shortcut. But I honestly don't think there's any shortcut possible.

ISTM that to do path resolution, you really need to start at the root (there's no CWD in WASI preview 1), then for each path element do `readlinkat` to possibly resolve a symlink, then `openat(NO_FOLLOW)`, then `fstat` to see if it's a directory, and on to the next element. If you don't have `NO_FOLLOW`, you have a race. I have no idea about Windows.

That's a lot of syscalls, and a lot of stuff to get wrong, especially if you're doing this for sandboxing/security sake.

anuraaga:
> ISTM that to do path resolution, you really need to start at the root (there's no CWD in WASI preview 1)

I might not understand the syscalls right, but I didn't really expect CWD to be needed. I thought it could be something like this pseudocode similar to checking whether opening a file matches isDir

https://github.com/tetratelabs/wazero/blob/main/imports/wasi_snapshot_preview1/fs.go#L1619

```golang
f, ok := fsc.LookupFile(newFD)
if symlinkTarget := f.File.SymlinkTarget(); symlinkTarget != nil { // Calls `os.Lstat` or `os.Readlink`
  _ = fsc.CloseFile(newFD)
  return doPathOpen(symlinkTarget, etc...) // The contents of pathOpen after pathName is computed from wasm arg
}
```

The biggest problem is that it seems like a fairly large change to have to add symlink lstat to the `File` interface, and anyways indeed could be missing something here so something simpler seems fine, just wanted to make sure the idea got across.

ncruces:
CWD is not needed because WASI preview 1 has no concept of CWD.

By the time you get to those lines the file is already opened, and path resolution already happened. The file is opened here:
https://github.com/tetratelabs/wazero/blob/111c51a1bd5acfd1e91afe489d28e52136a3e6d5/imports/wasi_snapshot_preview1/fs.go#L1610

The lines you're quoting are really only checking if you opened what you wanted to open (a directory) _after the fact_. That's good because “ask for forgiveness instead of permission” avoids a race.

But this doesn't work for symlinks for two reasons:
1. You don't open a symlink; you stat it with `lstat`, or read it with `readlink`, but you can't `open` and then `fstat` it, because by then you opened the target; on some platforms, you can say you don't want to open the target, with `O_NOFOLLOW`, on others not even that.
2. Symlinks in the last path element are the _least_ of your worries; what you should really be worrying about is directory symlinks in the middle; and those can be symlinks that have been moved around to point to different things than when they were created.

PR #2338, opened 2024-11-14
Add BSD to CI.

ncruces:
This adds FreeBSD, OpenBSD, and NetBSD to CI.

The ~weird~ regex (`imports|sysfs`) skips the following packages:
- `internal/sysfs`
- `experimental/sysfs`
- `imports/assemblyscript`
- `imports/assemblyscript/example`
- `imports/emscripten`
- `imports/wasi_snapshot_preview1`
- `imports/wasi_snapshot_preview1/example`

Mostly, `sysfs` and WASI need further work, even on FreeBSD.

I propose we merge this as is, if for no other reason, to ensure no regressions. Later, we can figure out what needs to be fixed/skipped. Most of the issues are WASI details/intricacies.

Instead of using the regex, we should probably skip specific WASI tests. Either based on GOOS, or some `supports_X` (test?) flags. A bunch of tests fail because of non-portable `stat` fields; others because of reduced file time resolution; etc.

Users of these OSes that interested in fixing the actual behaviour are encouraged to do so, then whitelist tests.

Also: the fact wazero works and is used in these OSes with little to no complaints suggests these details don't matter _that_ much in actual usage. WASI is complicated, our implementation of it is already very complex. It's also a preview.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ We currently test Linux (Ubuntu and scratch), MacOS and Windows as packaged by
 * Interpreter
   * Linux is tested on amd64 (native) as well arm64 and riscv64 via emulation.
   * MacOS and Windows are only tested on amd64.
-  * NetBSD and OpenBSD are only tested on amd64.
+  * FreeBSD, NetBSD and OpenBSD are only tested on amd64.
 * Compiler
   * Linux is tested on amd64 (native) as well arm64 via emulation.
   * MacOS and Windows are only tested on amd64.


mathetake:
* `internal/integration_test/engine`
* `internal/integration_test/fuzzcases`

I think we need these two to pass to say the core is working - what's the failing tests in there?

ncruces:
A few more notes.

This uses [`cross-platform-actions/action`](https://github.com/cross-platform-actions/action) which in my experience is the best option for these OSes.

I'm planing a follow up PR that uses [`vmactions`](https://github.com/vmactions) for other OSes, but that is a bit slower than these, and adds a bit more infrastructure (we need a templated custom action for a matrix run). We can do DragonFly BSD, illumos and Solaris that way, though.

We can also use `cross-platform-actions/action` to test `arm64`. It would be slower (CPU emulation), which compounds with the interpreter for additional overhead.

I am not testing multiple Go versions, nor am I installing a specific Go version; instead, I use the preinstalled GitHub Go version. Both these actions start a VM and rsync stuff into the VM. They don't know what exactly to copy, and they aren't very configurable here. If we install Go, they end up copying the entire local Linux Go installation into the VM for no reason.

ncruces:
> * `internal/integration_test/engine`
> * `internal/integration_test/fuzzcases`
> 
> I think we need these two to pass to say the core is working - what's the failing tests in there?

`fuzzcases` repeatedly crashes OpenBSD with an out of memory: https://github.com/ncruces/wazero/actions/runs/11840657747/job/32995009277

`engine` does the same to all 3 OSes (two of them just die):
https://github.com/ncruces/wazero/actions/runs/11840579279/job/32994741104

I can try to isolate the test cases. If you have any hints/suspicions, it'd be helpful.

ncruces:
Updated the initial comment, for more visibility of the current state of play.

I'd like to highlight this:

**Users of these OSes that interested in fixing the actual WASI behaviour are encouraged to do so, _then_ whitelist tests.**

They can run the tests locally with `make test` and fix things much more easily that we can.

.github/workflows/commit.yaml
```diff
@@ -152,6 +152,44 @@ jobs:
...
+        uses: cross-platform-actions/action@v0.25.0
+        env:
+          WAZEROCLI: ./wazerocli
+        with:
+          operating_system: ${{ matrix.os.name }}
+          version: ${{ matrix.os.version }}
+          shell: bash
+          sync_files: runner-to-vm
+          environment_variables: WAZEROCLI
+          # This runs all tests compiled above in sequence. Note: This mounts /tmp to allow t.TempDir() in tests.
```
mathetake, line 190:
```suggestion
          # This runs all tests compiled above in sequence.
```

.github/workflows/commit.yaml
```diff
@@ -152,6 +152,44 @@ jobs:
...
+        os:
+          - name: freebsd
+            version: "14.1"
+          - name: openbsd
+            version: "7.5"
+          - name: netbsd
+            version: "10.0"
+
+    steps:
+
```
mathetake, line 170:
```suggestion
```

README.md
```diff
@@ -101,9 +101,11 @@ We currently test Linux (Ubuntu and scratch), MacOS and Windows as packaged by
 * Interpreter
   * Linux is tested on amd64 (native) as well arm64 and riscv64 via emulation.
   * MacOS and Windows are only tested on amd64.
+  * NetBSD and OpenBSD are only tested on amd64.
 * Compiler
   * Linux is tested on amd64 (native) as well arm64 via emulation.
   * MacOS and Windows are only tested on amd64.
+  * FreeBSD is only tested on amd64.
```
mathetake, line 108: wondering if Net/OpenBSD passes with compiler or not https://github.com/tetratelabs/wazero/blob/dc08732e57d575a240774bc64e87cbed59f81c34/config_supported.go#L8 ?

README.md
```diff
@@ -101,9 +101,11 @@ We currently test Linux (Ubuntu and scratch), MacOS and Windows as packaged by
 * Interpreter
   * Linux is tested on amd64 (native) as well arm64 and riscv64 via emulation.
   * MacOS and Windows are only tested on amd64.
+  * NetBSD and OpenBSD are only tested on amd64.
```
mathetake, line 104:
```suggestion
  * FreeBSD, NetBSD and OpenBSD are only tested on amd64.
```

ncruces: I'm sure it works, but are we testing the interpreter?
I mean, is that one of the tests?

README.md
ncruces: Yes. NetBSD works. I was keeping that feature change for after I have working tests.
OpenBSD fails for some unknown reason.

One of the reasons I am holding the change, is that illumos _also_ works, and I wanted to add CI for it _before_ enabling the compiler for both.

wazero is (as a dependency) on [pkgsrc](https://pkgsrc.org/) which is the package manager for NetBSD and illumos, so it'd be nice to enable the compiler there.

README.md
mathetake: yeah spectests are running both interpreter and compiler when available

README.md
mathetake: ok, let's keep it as-is for now, and later enable compilers there when you have time

README.md
ncruces: Ah, I'm away from keyboard, I'll update when I'm back. Doing it here apparently borks the signoff?

mathetake: taking care of it!

PR #2343, opened 2024-11-15
Enable compiler on NetBSD, DragonFly, illumos and Solaris

ncruces:
This culminates this series of changes, enabling the compiler on:
- NetBSD `amd64` and `arm64`
- DragonFly BSD `amd64`
- illumos `amd64`
- Solaris `amd64`

We're not currently testing `arm64` for either FreeBSD or NetBSD, but I've tested them. The emulation is slow, I'm not sure if it's worth it. The others are not supported Go `arm64` targets.

OpenBSD was not added because it failed with an SIGSEGV. If anyone wants to take a stab at this, I'd welcome it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,8 @@ We also test cross compilation for many `GOOS` and `GOARCH` combinations.
   * macOS is tested only on arm64.
 * Compiler
   * Linux is tested on amd64 (native) as well arm64 via emulation.
-  * Windows and FreeBSD are only tested on amd64.
+  * Windows, FreeBSD, NetBSD, DragonFly BSD, illumos and Solaris are
+    tested only on amd64.
   * macOS is tested only on arm64.
 
 wazero has no dependencies and doesn't require CGO. This means it can also be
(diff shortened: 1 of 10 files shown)


mathetake (approved):

ncruces:
Just adding a note that https://github.com/tetratelabs/wazero/pull/2345 essentially disables the compiler on `arm64` NetBSD.

It can be force enabled with `wazero.NewRuntimeConfigCompiler()` (which disables CPU detection), but you run the risk of crashes at runtime (if your CPU doesn't support the necessary instructions).

anuraaga:
@ncruces Just to understand why does #2345 (excellent pr number by the way) disable it? I figured it would use compiler as long as it's arm 8.1 even on NetBSD. Or does the OS actually expose feature flags differently?

ncruces:
This should offer a good explanation:
https://kernel.org/doc/Documentation/arm64/cpu-feature-registers.txt

Unlike the Intel CPUID instruction, these ARM registers are not made available to user space.

Linux and FreeBSD do make them available, somehow (which I don't fully understand), but on other OSes just running that code crashes the process.

Any other OS where this works, I'm OK with whitelisting, but these are the only two in the Go codebase.

ncruces:
This also means that post #2345 is not a regression vs `v1.8.1`: the only systems where we disable the compiler by default, are the ones were **sure** are missing instructions (`amd64` without SSE4.1 or `arm64` without atomics), and you can **still** force enable the compiler anyway.

PR #2345, opened 2024-11-19
compiler: require ARMv8.1

ncruces:
This checks that [Large System Extensions](https://learn.arm.com/learning-paths/servers-and-cloud-computing/lse/intro/) (atomic instructions) are supported on `arm64`. These are included in many processors, and are part of ARMv8.1, which is required by macOS and Windows.

This is similar to the check that the `amd64` CPUs support SSE4.1 (which is part of x86-64-v2, and available on most processors).

Biggest behavioral change is using `platform.CompilerSupported()` to decide whether to (not) use the compiler on `wazero.NewRuntimeConfig()`, but I think this makes sense.

### DIFF
--- a/internal/platform/cpuid_arm64.go
+++ b/internal/platform/cpuid_arm64.go
@@ -1,0 +1,76 @@
+//go:build gc
+
+package platform
+
+import "runtime"
+
+// CpuFeatures exposes the capabilities for this CPU, queried via the Has, HasExtra methods.
+var CpuFeatures = loadCpuFeatureFlags()
+
+// cpuFeatureFlags implements CpuFeatureFlags interface.
+type cpuFeatureFlags struct {
+	isar0 uint64
+	isar1 uint64
+}
+
+// implemented in cpuid_arm64.s
+func getisar0() uint64
+
+// implemented in cpuid_arm64.s
+func getisar1() uint64
+
+func loadCpuFeatureFlags() CpuFeatureFlags {
+	switch runtime.GOOS {
+	case "darwin", "windows":
+		// These OSes require ARMv8.1, which includes atomic instructions.
+		return &cpuFeatureFlags{
+			isar0: uint64(CpuFeatureArm64Atomic),
+			isar1: 0,
+		}
+	case "linux", "freebsd":
+		// These OSes allow reading the instruction set attribute registers.
+		return &cpuFeatureFlags{
+			isar0: getisar0(),
+			isar1: getisar1(),
+		}
+	default:
+		return &cpuFeatureFlags{}
+	}
+}
+
+// Has implements the same method on the CpuFeatureFlags interface.
+func (f *cpuFeatureFlags) Has(cpuFeature CpuFeature) bool {
+	return (f.isar0 & uint64(cpuFeature)) != 0
+}
+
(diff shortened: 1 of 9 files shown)


ncruces:
See https://github.com/wasilibs/go-re2/issues/136 for an instance of this happening.

internal/platform/cpuid_arm64.go
```diff
@@ -0,0 +1,76 @@
+//go:build gc
```
mathetake, line 1: why is this file suffixed with arm64 but it contains logic?

internal/platform/cpuid_arm64.s
```diff
@@ -0,0 +1,21 @@
+//go:build gc
+
+#include "textflag.h"
+
+// lifted from github.com/golang/sys and cpu/cpu_arm64.s
+
+// func getisar0() uint64
+TEXT ·getisar0(SB), NOSPLIT, $0-8
+	// get Instruction Set Attributes 0 into x0
+	// mrs x0, ID_AA64ISAR0_EL1 = d5380600
```
mathetake, line 10: could you put the reference to the arm64 manual?

internal/platform/cpuid_arm64.s
```diff
@@ -0,0 +1,21 @@
...
+	WORD $0xd5380600
+	MOVD R0, ret+0(FP)
+	RET
+
+// func getisar1() uint64
+TEXT ·getisar1(SB), NOSPLIT, $0-8
+	// get Instruction Set Attributes 1 into x0
+	// mrs x0, ID_AA64ISAR1_EL1 = d5380620
```
mathetake, line 18: also here

internal/platform/cpuid_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+// implemented in cpuid_arm64.s
+func getisar0() uint64
+
+// implemented in cpuid_arm64.s
+func getisar1() uint64
+
+func loadCpuFeatureFlags() CpuFeatureFlags {
+	switch runtime.GOOS {
+	case "darwin", "windows":
+		// These OSes require ARMv8.1, which includes atomic instructions.
```
mathetake, line 25: i get why darwin requires v8.1 by default (m series is so), but not sure about windows here. any reference or pointer?

internal/platform/cpuid_arm64.go
ncruces: Because I'm dumb. Thanks for noticing!

internal/platform/cpuid_arm64.s
ncruces: If you mean the constants (`0xd5380600`), I have none besides Go code for `runtime`, `x/sys/cpu` and `klauspost/cpuid`. Other assemblers apparently get this from a (assembly) macro, that the Go assembler doesn't support.

internal/platform/cpuid_arm64.go
ncruces: Windows 10 does not in fact require them, is it good enough to document this?
At least this change should not be a regression.

mathetake (approved):

mathetake:
cc @anuraaga fyi

PR #2346, opened 2024-11-23
arm64: account for imported functions when encoding relocation islands

### LINKED ISSUE #2340: Programs that succeed to execute with v1.6.0 sometimes hang indefinitely with v1.8.1 during instantiation.
**Describe the bug**
Programs that succeed to execute with v1.6.0 sometimes hang indefinitely with v1.8.1 during instantiation.
The same program crashes with v1.7.x

**To Reproduce**
Clone the example repository:
```bash
git clone https://github.com/davidmdm/wzbug.git
cd ./wzbug
./test.sh
```
**Expected behavior**
Program should behave the same as with the v1.6.0 runtime.

- Go version: 1.23.3
- wazero Version: v1.8.1
- Host architecture: Arm64 (M2 Mac)
- Runtime mode: Compiler

**Additional context**

Incredibly rare and difficult to reproduce. If I tweak the example program slightly, the bug disappears. If I don't use stdin the bug disappears. However I don't want to write programs fearing that the perfect storm will hang my execution.

evacchi:
Fixes #2340 

Thanks @davidmdm for reporting the issue. Please do confirm if this solves your problem!

Here's a summary for the lazy. On arm64, when we compute the offsets in a trampoline, the following function is given a mapping between **all** function refs (basically Indexes in the original Wasm file) and offsets (`refToBinaryOffset`).

```go
func encodeCallTrampolineIsland(refToBinaryOffset []int, islandOffset int, executable []byte) {
	for i := 0; i < len(refToBinaryOffset); i++ {
		trampolineOffset := islandOffset + trampolineCallSize*i
```

Wasm function indexes account for _all_ functions, including _imported_ functions. However, we do not need to allocate space for such functions in the "trampoline island". In fact, _we do not account for them_:

```go
	trampolineInterval, callTrampolineIslandSize, err := machine.CallTrampolineIslandInfo(localFns)
```

however:
1. because we do not account for them, 
2. because relocations are being written _after_ the code of each function has been laid out into the mmap'ed segment, 

if we write out of the expected bounds, we end up overwriting the beginning of the function that follows that island. This causes all kind of wacky behavior, because when you invoke _that_ function, it starts with some jumps to "random" places. 

The fix is relatively simple, and it can be rewritten in a few ways. Essentially, we need to account for the imported functions, and ignore those indexes when we write the island.

### DIFF
--- a/internal/engine/wazevo/backend/isa/amd64/machine.go
+++ b/internal/engine/wazevo/backend/isa/amd64/machine.go
@@ -2196,7 +2196,7 @@ func (m *machine) Encode(ctx context.Context) (err error) {
 }
 
 // ResolveRelocations implements backend.Machine.
-func (m *machine) ResolveRelocations(refToBinaryOffset []int, binary []byte, relocations []backend.RelocationInfo, _ []int) {
+func (m *machine) ResolveRelocations(refToBinaryOffset []int, _ int, binary []byte, relocations []backend.RelocationInfo, _ []int) {
 	for _, r := range relocations {
 		offset := r.Offset
 		calleeFnOffset := refToBinaryOffset[r.FuncRef]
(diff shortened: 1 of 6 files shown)


mathetake (approved):
💯

davidmdm:
I confirm that this has solved my issue !

PR #2354, opened 2024-12-25
fix: return proper nil on runtime.InstantiateModule errors (#2353)

### LINKED ISSUE #2353: runtime.InstantiateModule returns a "struct-wrapped" nil value
**Describe the bug**
`runtime.InstantiateModule` sometimes returns a "struct-wrapped" `nil` value, instead of returning actual `nil`.

More specifically, this line https://github.com/tetratelabs/wazero/blob/main/runtime.go#L318 assigns `mod` and returns it implicitly on L324.

This allows code like the following to panic with a nil pointer dereference because `m` is `nil`:

```go
m, _ := r.InstantiateModule(ctx, compiled, config)
if m != nil {      // this check passes
  defer m.Close()  // and yet this causes a nil pointer dereference
}
```

**To Reproduce**
Here is an example that mimics wazero's behavior: https://go.dev/play/p/66QShXidalq

**Expected behavior**
From https://go.dev/doc/faq#nil_error, the FAQ recommends:

> To return a proper nil error to the caller, the function must return an explicit nil

**Additional context**
Found on https://github.com/sqlc-dev/sqlc/pull/3759#issuecomment-2562013442

edman:
Fixes #2353

Previously InstantiateModule could return a `(*wasm.ModuleInstance)(nil)` instead of a proper `nil` value.

So callers could get a nil dereference panic in code like:

    m, _ := r.InstantiateModule(...)
    if m != nil {       // this check passes
        defer m.Close() // but this causes a panic
    }

### DIFF
--- a/runtime.go
+++ b/runtime.go
@@ -9,7 +9,6 @@ import (
 	experimentalapi "github.com/tetratelabs/wazero/experimental"
 	"github.com/tetratelabs/wazero/internal/expctxkeys"
 	internalsock "github.com/tetratelabs/wazero/internal/sock"
-	internalsys "github.com/tetratelabs/wazero/internal/sys"
 	"github.com/tetratelabs/wazero/internal/wasm"
 	binaryformat "github.com/tetratelabs/wazero/internal/wasm/binary"
 	"github.com/tetratelabs/wazero/sys"
@@ -289,8 +288,8 @@ func (r *runtime) InstantiateModule(
 	ctx context.Context,
 	compiled CompiledModule,
 	mConfig ModuleConfig,
-) (mod api.Module, err error) {
-	if err = r.failIfClosed(); err != nil {
+) (api.Module, error) {
+	if err := r.failIfClosed(); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


runtime.go
```diff
@@ -340,20 +339,20 @@ func (r *runtime) InstantiateModule(
 		if start == nil {
 			continue
 		}
-		if _, err = start.Call(ctx); err != nil {
+		if _, err := start.Call(ctx); err != nil {
 			_ = mod.Close(ctx) // Don't leak the module on error.
```
edman, line 343: Unrelated to the bug I found, but is this correct?

It looks like this shouldn't be closed in the success case below in L347.

ncruces:
This changes a lot more than the issue you had. If your issue was with line 318, why do you assume that returning the module _and_ the error is not intentional when a start function "fails"?

I'll defer to @mathetake but I can definitely see someone wanting to inspect the linear memory of a module that called `exit(1)` on a start function.

runtime.go
ncruces: I think if the module exited (even with exit code zero) it has already been closed. In that case, closing it again would not be an error.

Issue #2360, opened 2025-01-12
suspected bug: different results on x86 vs apple arm


faddat:
This is the same floating point math test. The only difference is that it is being run on an ubuntu instance and on an M1.

Ubuntu fails because I used the result from my m1.

* https://github.com/CosmWasm/wasmvm/actions/runs/12732434887/job/35487631744?pr=582

* https://github.com/CosmWasm/wasmvm/actions/runs/12732434887/job/35487631959?pr=582

My understanding is that the output should be exactly the same, regardless of platform.  Is that correct?

ncruces:
Do you have source for the Wasm? Also, even if you don't have source for the Wasm, some minimal Go code to show the difference would be helpful. A repo (or a [gist](https://gist.github.com/)) with a [MCVE](https://stackoverflow.com/help/minimal-reproducible-example) is _always_ helpful.

Also, how was the Wasm compiled? Does it use SIMD? Is it possible the difference is in SIMD code?

In terms of assurances of the output being equal, I'm reasonably sure it _should_ be equal, _if_ it's using IEEE 754 floats. I'm less sure if it is using SIMD.

anuraaga:
Also one thing to keep in mind is NaN isn't deterministic in Wasm.

https://github.com/WebAssembly/design/blob/main/Nondeterminism.md

Not sure but found that perhaps it is being used in the test case. Would be good to confirm it (though I would be surprised if Wasmer handled them more deterministically than Wazero).

https://github.com/CosmWasm/cosmwasm/blob/main/contracts/floaty/src/floats.rs#L27

ncruces:
@faddat if this is indeed an issue, it's something we'd want to fix. But we need more info/help.

mathetake:
yep, floating point is non deterministic and it's clearly documented in the spec (and each CPU has different behavior as it's undefined!) - unless you give us the real repro which doesn't involve NaN or Inf, i am closing

faddat:
Ok, thank you all for your feedback!

I have been afk for a few days, and I'll poke around and come back.

ncruces:
#2365 should not change anything, unless (again) there is a `NaN` (which is non-deterministic), or some SIMD that is allowed to have reduced precision, or of course, some _other_ form of non-determinism (a clock, whatever).

If you have a discrepancy to report, between interpreter and compiler or across platforms, please do so. But it should be (as much as possible) a reduced test case, which is not supposed to be non-deterministic or unspecified to begin with.

PR #2365, opened 2025-01-16
compiler: fix ARMv8 regression introduced in #2345

ncruces:
This is an attempt to fix the performance regression introduced in #2345 by requiring ARMv8.1 when it is only strictly needed for `experimental.CoreFeaturesThreads`.

When I did #2345 I was trying to fix https://github.com/wasilibs/go-re2/issues/136, and assumed (from the bug description) that these processors were really uncommon. It turns out that (e.g.) RaspberryPi 3 and 4 don't support these, and neither does the original Graviton (still in use by AWS).

Also, only Linux and FreeBSD allow us to figure out if the processor is an ARMv8.1 (we assume it is for macOS and take the chance on Windows, because 11 requires it). Still, the ARM compiler works fine on NetBSD too (I test this it in [SQLite](https://github.com/ncruces/go-sqlite3/wiki/Support-matrix), as well as the ARM FreeBSD compiler).

I'm happy to reconsider the approach if this seems confusing.


ncruces:
I clearly need to run tests with the race detector on. Sorry about that.

faddat:
What I am thinking about #2360 is that I am going to try out this branch, and see if it changes anything, then investigate the rest.

Thanks for your responses!

ncruces:
OK, so I guess this is ready.

This delays the determination of whether to use the compiler or interpreter to `NewRuntimeWithConfig`, then taking account of `enabledFeatures`.

Potentially `amd64` could run on all x64 processors if SIMD is disabled, but I didn't go there. I'm just stating the possibility.

anuraaga (approved):
Seems like a generally good simplification too

PR #2366, opened 2025-01-17
CI: use Linux ARM runners.

ncruces:
These just entered public preview:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -117,7 +117,7 @@ jobs:
         platform:
           - os: ubuntu-22.04
             arch: amd64
-          - os: ubuntu-24.04-arm64
+          - os: ubuntu-22.04-arm
             arch: arm64
           - os: ubuntu-22.04
             arch: riscv64
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -107,17 +107,21 @@ jobs:
 
   test_scratch:
     name: ${{ matrix.arch }}, Linux (scratch), Go-${{ matrix.go-version }}
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.platform.os }}
     strategy:
       fail-fast: false  # don't fail fast as sometimes failures are arch/OS specific
       matrix:  # Use versions consistent with wazero's Go support policy.
         go-version:
           - "1.23"  # Current Go version
           - "1.21"  # Floor Go version of wazero (current - 2)
-        arch:
-          - "amd64"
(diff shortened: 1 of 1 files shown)


ncruces:
OK, these may be less available than I thought.

I tested it a few times on my repo, and didn't experience the "long queue times" they mention in the blog post. This is a bit ridiculous.

We may still feel that it's better to test native vs emulated. Or not.

robherley:
@ncruces The new label is `ubuntu-24.04-arm`, I think there's a typo in the workflow definitions where you have `*-arm64` and it's not matching any labels

ncruces:
Nice catch. Just need to fix the DCO.

.github/workflows/commit.yaml
```diff
@@ -106,18 +106,22 @@ jobs:
...
           - "1.21"  # Floor Go version of wazero (current - 2)
-        arch:
-          - "amd64"
-          - "arm64"
-          - "riscv64"
+        platform:
+          - os: ubuntu-22.04
+            arch: amd64
+          - os: ubuntu-22.04-arm
+            arch: arm64
```
anuraaga, line 121: Shouldn't we add to the normal `test` task too?

Another note is personally, running this for every arch seems like overkill. amd64 was added for #385, while arm64 was here because it required qemu. Not it doesn't require qemu, so I'm wondering if we need both amd64 and arm64 in this matrix.

mathetake: yeah I would rather delete this for amd64 and arm64

ncruces: I used this because it's a “reduced” test suite compared to “full” (same way I used it for the VMs, although cross compilation helps there).

But I can drop this and run the full test suite on Linux ARM.

Or not at all (we stopped running any tests on macOS AMD64). I run a bunch of these other configurations (including BSD ARM) on my own repo, so it's not like it's untested.

mathetake: so up to you but we want to reduce the CI costs

ncruces: Well at least for this run, scratch+AMD was faster than base+AMD (and scratch+ARM is faster still).

I dunno what's cheaper by the minute? 

If we wanna reduce costs we can drop some of the targets I added, or move them to something we manually run (e.g. before a release).

If we make them manually triggered I can at least run them for free on a fork, that's easy enough to do.

mathetake (approved):

Issue #2371, opened 2025-02-04
Interpreter: Incorrect `i32.ne` results for value from host function (negative numbers only)


AndSDev:
**Describe the bug**
Incorrect `i32.ne` results for value from host function (negative numbers only).

Host func result put on stack as uint64 (see [internal/wasm.callGoFunc](https://github.com/tetratelabs/wazero/blob/v1.8.2/internal/wasm/gofunc.go#L135)).

For -1 its 0xFFFFFFFFFFFFFFFF. 
But on comaration with i32(-1) from wasm its comapres with 0x00000000FFFFFFFF.

**To Reproduce**

Golang code:
```Go
package main

import (
	"context"
	"encoding/base64"
	"log"

	"github.com/tetratelabs/wazero"
)

func main() {
	wasm, _ := base64.StdEncoding.DecodeString("AGFzbQEAAAABCgJgAX8Bf2AAAX8CDAEDZW52BGVjaG8AAAMCAQEHBwEDcnVuAAEKCwEJAEF/EABBf0cLABoEbmFtZQEMAgAEZWNobwEDcnVuAgUCAAABAA==")

	config := wazero.NewRuntimeConfigInterpreter()
	ctx := context.Background()
	r := wazero.NewRuntimeWithConfig(ctx, config)

	_, _ = r.NewHostModuleBuilder("env").
		NewFunctionBuilder().
		WithFunc(func(v int32) int32 {
			return v
		}).Export("echo").
		Instantiate(ctx)

	mod, _ := r.Instantiate(ctx, wasm)
	run := mod.ExportedFunction("run")
	result, err := run.Call(ctx)
	if err != nil {
		log.Fatalf("failed to call run: %v", err)
	}

	if result[0] == 0 {
		println("OK: -1 != -1 is false")
	} else {
		panic("FAIL, -1 != -1 is true")
	}
}
```

Wat code for WASM
```wat
(module
  (import "env" "echo" (func $echo (param i32) (result i32)))
  (func $run (result i32)
    i32.const -1
    call $echo
    i32.const -1
    i32.ne
  )
  (export "run" (func $run))
)
```

**Expected behavior**
After run, the message "OK: -1 != -1 is false" should be displayed, but an error occurs.

**Screenshots**
<details>
  <summary>Screenshot from debugger</summary>

![Image](https://github.com/user-attachments/assets/135ee688-96ac-4176-a32c-57cad44036cf)
</details>

**Environment (please complete the relevant information):**
[...]

PR #2374, opened 2025-02-10
fix: `HostFunctionBuilder.WithFunc` works with negative int32

AndSDev:
According to the description, `HostFunctionBuilder.WithFunc` should correctly convert int32 to wasm-type. But int32 is extended to int64, which is incorrect for negative numbers.

Description of HostFunctionBuilder.WithFunc:
https://github.com/tetratelabs/wazero/blob/4231c48a9c9adf451efcacecfd82278cab2c43a4/builder.go#L92-L94

The correct conversion of int32 to i32 is `uint64(uint32(input))`:
https://github.com/tetratelabs/wazero/blob/4231c48a9c9adf451efcacecfd82278cab2c43a4/api/wasm.go#L715-L718

### DIFF
--- a/internal/wasm/gofunc.go
+++ b/internal/wasm/gofunc.go
@@ -131,7 +131,9 @@ func callGoFunc(ctx context.Context, mod api.Module, fn *reflect.Value, stack []
 			stack[i] = math.Float64bits(ret.Float())
 		case reflect.Uint32, reflect.Uint64, reflect.Uintptr:
 			stack[i] = ret.Uint()
-		case reflect.Int32, reflect.Int64:
+		case reflect.Int32:
+			stack[i] = uint64(uint32(int32(ret.Int())))
+		case reflect.Int64:
 			stack[i] = uint64(ret.Int())
 		default:
 			panic(fmt.Errorf("BUG: result[%d] has an invalid type: %v", i, ret.Kind()))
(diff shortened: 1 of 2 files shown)


anuraaga (approved):

internal/wasm/gofunc.go
```diff
@@ -131,7 +131,9 @@ func callGoFunc(ctx context.Context, mod api.Module, fn *reflect.Value, stack []
 			stack[i] = math.Float64bits(ret.Float())
 		case reflect.Uint32, reflect.Uint64, reflect.Uintptr:
 			stack[i] = ret.Uint()
-		case reflect.Int32, reflect.Int64:
+		case reflect.Int32:
+			stack[i] = uint64(uint32(int32(ret.Int())))
```
ncruces, line 135: nit: given your comment (and source reference), there's no reason this shouldn't be `uint64(uint32(ret.Int()))` for consistency with that other code.

ncruces (approved):

internal/wasm/gofunc.go
anuraaga: Just for context, I had suggested the `int32` cast to match the `int32` parameter type of that code. But if it seems like overkill, no worries, TBH I'm not able to grok the int casts well enough to know if they're equivalent or not.

ncruces: Ah. OK. I had missed that. Sorry. It's a nit though, feel free to ignore.

AndSDev: Yes, converting to int32 is not necessary. But I think it is better for readability.

anuraaga:
Hi @AndSDev - discussing with Takeshi, I realized that we have a design in wazero where for `i32` the upper bits are meant to be ambiguous/opaque, so they should not affect results based on the implementations of the engines. We are reopening #2371 as an issue within the interpreter itself, that it needs to truncate somewhere (compilers have the truncation automatic/implicit by selecting 32-bit instructions). Sorry for the churn

PR #2379, opened 2025-02-18
ci: update Go floor to 1.22 and run CI with 1.22/1.24

anuraaga:


.github/workflows/examples.yaml
```diff
@@ -18,7 +18,7 @@ on:
 env:
   EMSDK_VERSION: "3.1.40"
   TINYGO_VERSION: "0.35.0"
-  ZIG_VERSION: "0.11.0"
```
anuraaga, line 21: I was having some issues in CI with this and it was easier to bump to the latest which I have installed to match it up

.github/workflows/integration.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
         with:
           # Use share the cache containing archives across OSes.
           enableCrossOsArchive: true
-          key: zig-stdlib-test-binary-${{ env.ZIG_VERSION }}
+          key: zig-stdlib-test-binary-${{ env.ZIG_VERSION }}-${{ hashFiles(format('{0}/Makefile', env.STDLIB_TESTS)) }}
```
anuraaga, line 50: I think it makes sense to rebuild if we edit the `Makefile`, we may add test targets for example

internal/integration_test/stdlibs/Makefile
```diff
@@ -1,7 +1,7 @@
 # Note: Some tests might fail if Go has been installed with homebrew because
 # the file system layout is different than what the tests expect.
 # Easiest fix is to install without using brew.
-goroot?=$(shell go env GOROOT)
+goroot?=$(shell go env GOROOT)/src
```
anuraaga, line 4: `go.mod` is in `/src`. I'm not sure why I seemed to only have huge issues on Windows but it would fail with something like "unable to find go.mod".

internal/integration_test/stdlibs/bench_test.go
```diff
@@ -91,20 +91,29 @@ var (
 			c = c.WithFSConfig(
 				wazero.NewFSConfig().
 					WithDirMount(sysroot, "/")).
-				WithEnv("PWD", normalizedTestdir)
+				WithEnv("PWD", normalizedTestdir).
+				WithEnv("GOWASIRUNTIME", "wazero")
 
 			args := []string{fname, "-test.short", "-test.v"}
 
+			// Some distributions of Go such as used by homebrew or GitHub actions do not
```
anuraaga, line 99: We have a comment about homebrew in `Makefile` but it seems now setup-go on GHA also can run into the same problem on all OSes

internal/integration_test/stdlibs/bench_test.go
```diff
@@ -91,20 +91,29 @@ var (
...
 				c = c.
 					WithEnv("GOROOT", normalizeOsPath(runtime.GOROOT()))
 
-				args = append(args,
-					"-test.skip=TestRenameCaseDifference/dir|"+
-						"TestDirFSPathsValid|TestDirFS|TestDevNullFile|"+
-						"TestOpenError|TestSymlinkWithTrailingSlash|TestCopyFS")
+				skip = append(skip, "TestRenameCaseDifference/dir", "TestDirFSPathsValid", "TestDirFS",
+					"TestDevNullFile", "TestOpenError", "TestSymlinkWithTrailingSlash", "TestCopyFS",
+					"TestRoot", "TestOpenInRoot", "ExampleAfterFunc_connection")
```
anuraaga, line 114: I think `TestRoot` / `TestOpenInRoot` are similar to `TestDirFS`. Not sure about `ExampleAfterFunc_connection` but it fails

internal/sysfs/osfile.go
```diff
@@ -275,6 +275,9 @@ func (f *osFile) Pwrite(buf []byte, off int64) (n int, errno experimentalsys.Err
 
 // Truncate implements the same method as documented on sys.File
 func (f *osFile) Truncate(size int64) (errno experimentalsys.Errno) {
+	if size < 0 {
```
anuraaga, line 278: Only Go 1.24 windows has a different result here. But I couldn't see a reason not to do parameter validation directly here

evacchi (approved):
I think this is fine!

ncruces (approved):

PR #2381, opened 2025-02-22
wazevo: concurrent local wasm function compilation

davidmdm:
This PR takes a stab at making local wasm function compilation run concurrently.

In general the executables produced by it seem to work and compile 50-70% faster on my machine with 8 logical CPUs.

However, a couple of tests have broken with regards to expected offsets, and I don't have the context on first pass to see where the problem has occurred.

Hopefully it is an easy fix, if not and this is not the correct approach I will close this PR.
Thanks!


davidmdm:
@mathetake 

In order to make this concurrent, I have created a new SSABuilder, FrontEnd and Backend for each local wasm compile, since in sequence it was running an Init and seemed to reset internal state between runs.

Are there any structures that need to be shared between runs? Is that why offsets have changed?

davidmdm:
Big update: the bug was in the sourcemap offsets not being captured from the local backend instance.

All tests are passing locally, and CI seems much happier. Unclear if the windows CI failure was a flaky test or related to this change.

evacchi:
humm, it might be a real issue 🤔 

> panic: BUG: MmapCodeSegment with zero length [recovered]
>	panic: BUG: MmapCodeSegment with zero length

it looks like `MmapCodeSegment` is called with 0 somehow. It might be platform-dependent because Windows uses a different strategy 

https://github.com/tetratelabs/wazero/blob/c397a402ad17e495a5a22fef709a1d4abef1f3ce/internal/platform/mmap_windows.go#L59-L75

but what fails is the call to `MmapCodeSegment`, so it might be also a concurrency issue that is not apparent in the Unix version 🤔

davidmdm:
Update: All tests are passing. 

I had wrongfully assumed that a canceled context would return an error from the local compile, and ignore context canceled errors, simply opting to end the worker routine.

The latest change made sure that any cancelation error was surfaced, and this fixed the mmap of length 0.

This likely occurred due to a canceled context, causing no worker to run, but no error to surface, so no machine code produced.

davidmdm:
@mathetake 

CC @evacchi @ncruces 

At this point the POC is generally done. Spawning workers to compile wasm code sections concurrently passes all tests.
On my machine with 4 cores (8 logical), I get on average 50-70% speed ups with a median of about ~60% speed ups. So more or less a 2.5x speed up in compilation time.

The one downside is that if you force the code to use a single worker, it is approximately ~10%  slower than the previous version. (Likely due to the buffering of compiled local wasm functions in memory before doing a single pass in order).

However, in my estimation constrained environments with a single core probably are not sensitive to compile times as they are already very long and the 10% loss in compile speed is unlikely to be meaningful.

In contrast, for development, and user facing applications that need to compile wasm modules, the speed up will be very noticeable. Especially for the poor souls (like myself) that tend to use the Go Toolchain to build their wasm modules (very large wasm module sizes).

Therefore, at this point I wonder what the path forward ought to be?

My gut would be to switch over to the concurrent implementation and expose a new concurrency configuration on the RuntimeConfig. Perhaps extend the interface with:

```go
WithCompilationWorkers(workers int) RuntimeConfig
```

And exposing it on the CLI such that we can control the number of workers (defaulting to runtime.NumCPUs)

```bash
wazero compile -workers 4 ./main.wasm
```

Alternatively, we could keep the old code path, and make the concurrent implementation opt in with an experimental engine/strategy. This would avoid the 10% slowdown on single core machines at the cost of a higher maintenance and duplication . This is not my preference, but if it is the project's preference, that is okay.

Any feedback would be great as I wish to align the PR with the project's goals and priorities as much as possible. 

Thanks!

davidmdm:
Added timing benchmarks from my machine with a couple large wasm files generated by the Go Toolchain:

```
# wzmain -> wazero built from main
# wzdev -> this experimental PR
# wz1.6.0 -> wazero v1.6.0 before the optimizing compiler
# wasmtime -> wasmtime 30.0.2

# redis.wasm is a big wasm file ~78mb
# basic.wasm is a big but less humongous ~15mb

wzmain compile ./redis.wasm  14.16s user 0.15s system 99% cpu 14.329 total
wzdev compile ./redis.wasm  28.98s user 1.55s system 631% cpu 4.839 total
wz1.6.0 compile ./redis.wasm  2.72s user 0.08s system 99% cpu 2.821 total
wasmtime compile ./redis.wasm  35.89s user 0.78s system 662% cpu 5.536 total

wzmain compile ./basic.wasm  2.82s user 0.04s system 93% cpu 3.073 total
wzdev compile ./basic.wasm  6.71s user 0.39s system 500% cpu 1.418 total
wz1.6.0 compile ./basic.wasm  0.62s user 0.03s system 74% cpu 0.874 total
wasmtime compile ./basic.wasm  7.42s user 0.18s system 641% cpu 1.183 total
```

We can notice that the new concurrent wazero implementation has similar system usage and performance to that of wasmtime.

davidmdm:
Added timing benchmarks from my machine with a couple large wasm files generated by the Go Toolchain:

```
# wzmain -> wazero built from main
# wzdev -> this experimental PR
# wz1.6.0 -> wazero v1.6.0 before the optimizing compiler
# wasmtime -> wasmtime 30.0.2

# redis.wasm is a big wasm file ~78mb
# basic.wasm is a big but less humongous ~15mb

wzmain compile ./redis.wasm  14.16s user 0.15s system 99% cpu 14.329 total
wzdev compile ./redis.wasm  28.98s user 1.55s system 631% cpu 4.839 total
wz1.6.0 compile ./redis.wasm  2.72s user 0.08s system 99% cpu 2.821 total
wasmtime compile ./redis.wasm  35.89s user 0.78s system 662% cpu 5.536 total

wzmain compile ./basic.wasm  2.82s user 0.04s system 93% cpu 3.073 total
wzdev compile ./basic.wasm  6.71s user 0.39s system 500% cpu 1.418 total
wz1.6.0 compile ./basic.wasm  0.62s user 0.03s system 74% cpu 0.874 total
wasmtime compile ./basic.wasm  7.42s user 0.18s system 641% cpu 1.183 total
```

We can notice that the new concurrent wazero implementation has similar system usage and performance to that of wasmtime.

ncruces:
Hi @davidmdm! Do you mind testing this loop instead:

```go
	type compiledFunc struct {
		body             []byte
		relsPerFunc      []backend.RelocationInfo
		idx              wasm.Index
		sourceOffsetInfo []backend.SourceOffsetInfo
	}

	compiledFuncs := make([]compiledFunc, len(module.CodeSection))
	var nextFunc atomic.Uintptr

	ctx, cancel := context.WithCancelCause(ctx)
	defer cancel(nil)

	workers := runtime.GOMAXPROCS(0)
	wg := sync.WaitGroup{}
	wg.Add(workers)

	for range workers {
		go func() {
			defer wg.Done()

			machine := newMachine()
			ssaBuilder := ssa.NewBuilder()
			fe := frontend.NewFrontendCompiler(module, ssaBuilder, &cm.offsets, ensureTermination, withListener, needSourceInfo)

			for i := int(nextFunc.Add(1) - 1); i < len(compiledFuncs); i = int(nextFunc.Add(1) - 1) {
				if ctx.Err() != nil {
					// Compilation cancelled.
					return
				}
				if wazevoapi.DeterministicCompilationVerifierEnabled {
					i = wazevoapi.DeterministicCompilationVerifierGetRandomizedLocalFunctionIndex(ctx, i)
				}

				fidx := wasm.Index(i + importedFns)

				if wazevoapi.NeedFunctionNameInContext {
					def := module.FunctionDefinition(fidx)
					name := def.DebugName()
					if len(def.ExportNames()) > 0 {
						name = def.ExportNames()[0]
					}
					ctx = wazevoapi.SetCurrentFunctionName(ctx, i, fmt.Sprintf("[%d/%d]%s", i, len(module.CodeSection)-1, name))
				}

				be := backend.NewCompiler(ctx, machine, ssaBuilder)
				needListener := len(listeners) > 0 && listeners[i] != nil

				body, relsPerFunc, err := e.compileLocalWasmFunction(ctx, module, wasm.Index(i), fe, ssaBuilder, be, needListener)
				if err != nil {
					cancel(err)
					return
				}
				compiledFuncs[i] = compiledFunc{
					body:             body,
					relsPerFunc:      relsPerFunc,
					idx:              fidx,
					sourceOffsetInfo: be.SourceOffsetInfo(),
				}
			}
		}()
	}

	wg.Wait()
	if err := context.Cause(ctx); err != nil {
		return nil, err
	}
```

Should be clear where this fits in, hopefully…

This:
1. reuses (some) compiler stuff across each goroutine (instead of recreating per function)
2. uses `GOMAXPROCS` instead of `NumCPU` to avoid unnecessary concurrency
3. replaces the `sections` channel with an atomic, which should be cheaper
4. removes the `errs` with the `WithCancelCause` context
5. calls `wg.Wait` immediately outside the loop
6. avoids some defers

I think this makes the intent, error flow, assured termination, all quite a bit clearer, and reduces quite a bit of overhead. However, concurrency is hard, and you may disagree, or find bugs.

With the above, compilation of my SQLite Wasm goes from about 1s, to 300ms in a 6 core machine; your version was closer to 400ms.

Problem: it's not immediately clear to me why the backend can't be shared and the rest (apparently) can. I can't, in good faith, approve this without understanding that.

But maybe both our takes help @mathetake chime in with a review?

davidmdm:
@ncruces 

I love the review, and I will update the PR accordingly once I have a minute. Thank you! Will update my pseudo scientific timing as well!

davidmdm:
After integrating changes from  🚀@ncruces 🚀  (The rockets are part of their official name as far as I am concerned)

my pseudo-scientific timings got even better results!

```
# wzmain -> wazero built from main
# wzdev -> this PR before @ncruces changes
# wzexp -> this PR with @ncruces changes
# wz1.6.0 -> wazero v1.6.0 before the optimizing compiler
# wasmtime -> wasmtime 30.0.2

# redis.wasm is a big wasm file ~78mb
# basic.wasm is a big but less humongous ~15mb

wzmain compile ./basic.wasm  2.89s user 0.05s system 100% cpu 2.920 total
wzdev compile ./basic.wasm  6.75s user 0.39s system 612% cpu 1.166 total
wzexp compile ./basic.wasm  5.17s user 0.15s system 630% cpu 0.845 total
wz1.6.0 compile ./basic.wasm  0.59s user 0.03s system 100% cpu 0.610 total
wasmtime compile ./basic.wasm  7.43s user 0.19s system 600% cpu 1.270 total

wzmain compile ./redis.wasm  14.25s user 0.14s system 99% cpu 14.418 total
wzdev compile ./redis.wasm  29.16s user 1.60s system 633% cpu 4.856 total
wzexp compile ./redis.wasm  26.61s user 0.73s system 676% cpu 4.042 total
wz1.6.0 compile ./redis.wasm  2.72s user 0.08s system 99% cpu 2.815 total
wasmtime compile ./redis.wasm  35.89s user 0.72s system 678% cpu 5.397 total
```

With these changes, we are consistently compiling quicker than wasmtime and approaching a stone's throw away from the old non-optimizing compiler.

evacchi:
> Problem: it's not immediately clear to me why the backend can't be shared and the rest (apparently) can. I can't, in good faith, approve this without understanding that.

I think that not sharing the ssa, front-end, backend and machine makes sense because they have internal state that gets modified as the function is being processed. e.g.:

https://github.com/tetratelabs/wazero/blob/9a19a04962881134301ddedf7dc84c50004dbedf/internal/engine/wazevo/backend/compiler.go#L17-L25

or:

https://github.com/tetratelabs/wazero/blob/70a9688a9f400e5997872e6f351b56eb0ca45dbe/internal/engine/wazevo/backend/isa/amd64/machine.go#L295-L297

davidmdm:
@evacchi @ncruces 

What is interesting here, is that we are able to extract the machine, fe, ssabuilder into the worker state so as to reuse it per worker instead of per loop. This gives us a good boost in performance and the tests work as before.

If we pull out the backend to be per-worker instead of per-loop for some reason the tests just hang forever.
However experimenting with this build, I think we could get even more performance out of wazero if we could manage to reuse the backend per-worker.

It is odd because in the previous serial version the backend was reused for all code sections and there was no issues with the tests. The local compile func also calls `be.Init()` internally, which should reset all the state on the backend, which should make it reusable...

So it is in fact kind of odd that its not working as expected.

internal/engine/wazevo/engine.go
```diff
@@ -227,27 +226,81 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
...
+
+			for i := range sections {
+				if err := ctx.Err(); err != nil {
+					// Compilation canceled!
+					return
+				}
+
+				if wazevoapi.DeterministicCompilationVerifierEnabled {
+					i = wazevoapi.DeterministicCompilationVerifierGetRandomizedLocalFunctionIndex(ctx, i)
+				}
```
evacchi, line 264: worth noticing that enabling this flag in `debug_options.go` give us further information about what could be happening. Do notice that this fails even without pushing `be` outside the loop.

```
❯ go run ./cmd/wazero run ../repro/redis-installer.wasm
BUG: Deterministic compilation failed for function[3/55834].internal_abi.__IntArgRegBitmap_.Set at scope="SSA".

This is mostly due to (but might not be limited to):
        * Resetting ssa.Builder, backend.Compiler or frontend.Compiler, etc doens't work as expected, and the compilation has been affected by the previous iterations.
        * Using a map with non-deterministic iteration order.
        
[... more details ...]
```

evacchi: I don't exclude that the check currently might not be concurrency-safe :)

davidmdm: Yeah, when I try and enable that check, I get a concurrency error:

```
fatal error: concurrent map read and map write
```

Setting up the verifierState. Unsure how you got it setup to verify anything :\

davidmdm: Okay sometimes it gives me that error and sometimes it gives me your error.

If I run with the verifier against main, it just hangs and never completes.

Hmmm.

internal/engine/wazevo/engine.go
```diff
@@ -841,3 +894,14 @@ func (cm *compiledModule) getSourceOffset(pc uintptr) uint64 {
 	}
 	return cm.sourceMap.wasmBinaryOffsets[index]
 }
+
+func sequence(size int) <-chan int {
```
james-lawrence, line 898: probably more efficient to just use atomics and a yield function. but this is a microoptimization.

internal/engine/wazevo/engine.go
james-lawrence: @davidmdm if the problem is deterministic builds being prevented due to the random order of the sequence generate I might have a few solutions you can try via seeding.

internal/engine/wazevo/engine.go
```diff
@@ -227,27 +226,81 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
...
+
+	workers := runtime.GOMAXPROCS(0)
+
+	wg := sync.WaitGroup{}
+	wg.Add(workers)
+
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer cancel(nil)
+
+	sections := sequence(len(module.CodeSection))
```
james-lawrence, line 246: based on the errors you guys mentioned, should be resolved with:
1. create a seeded PRNG.
2. shuffle the code sections using the PRNG.
3. create a thread safe queue to pop the module function to compile and its idx after shuffle
4. guard the compiledFuncs with a lock. and insert the result into the idx.

james-lawrence: (this approach should allow removing the conditional logic on deterministic builds across the code base as well while enabling PIE)

internal/engine/wazevo/engine.go
james-lawrence: yeah, rewrite this using a mutex and and a argument of a shuffle list of indices.

davidmdm: Ncruces and I tried removing the channel and using an atomic and didn't have any noticeable performance gains.

internal/engine/wazevo/engine.go
davidmdm: I am not sure I follow... especially the bit about guarding the compiled funcs with a lock. We want them to run concurrently as this is how we are getting our compilation performance gains. 

Do you mean only for the deterministic checking mode?

davidmdm: Thanks! I look forward to it. I knew I was not following :)

james-lawrence: https://github.com/tetratelabs/wazero/pull/2394 take a peek. just opened a PR to run the ci/cd to see if it works out. feel free to merge the changes into this PR and close it if it works out.

evacchi: mm I feel like what we really want is to make the build independent of the order; compiling each function should not really depend on shared state, if it does, then something icky must be going on :p

james-lawrence: agreed, the fix i opened that PR was to resolve the deterministic builds error you mentioned, not to ensure a particular order of compilation. by shuffling we ensure the compilation *is* randomized but can be deterministically reproduced by controlling the seed.

interestingly the errors from CI have changed dramatically from my changes, they likely identify any shared state now that the compiled functions are truly randomized.

evacchi:
I sent a PR to @davidmdm with a first approximation of a fix for the deterministic verifier. 

TBF indeed, the issues mainly seem to derive from the fact that the verifier was not designed with concurrency in mind in the first place, so I don't think there are fundamental issues with the code in this PR 

This is the commit https://github.com/evacchi/wazero/pull/62/commits/3003c902d925f1abdb8cceb680dda7160b6686bc

- there is one minor issue with `ctx` being a lambda capture and thus a mutable reference in the loop body, so it's better to `ctx := ctx` there, 
- besides that, most problems seem localized in `debug_options.go` and the way the internal state is accessed (the map and the slice of randomized indexes)

my PR ensures we assign the randomized indexes before we enter the loop to a local variable, and that we don't mutate it anymore; the map used by the verifier is now a sync.Map

I think it can be improved further, so everyone feel free to chime in!

davidmdm:
@evacchi @mathetake 

I am moving this PR as ready to review given that @evacchi has fixed the deterministic verifier to be concurrency safe.
I have also fixed the deadline exceeded tests to no longer be racy between a compilation or runtime deadline.

There is one failing test on windows that has to due with symlinks but I believe is unrelated to the changes here. I am willing to spend some time on it if it is blocking this PR.

With this PR wazero compilation speed becomes comparable to that of other runtimes such as wasmtime 🎉

mathetake (changes requested):
Did you check the memory usage as well? Could you check two binaries: Zig standard library tests and another some huge binary.
I believe the memory is also huge when compilation, automatically using all CPU cores can drastically increase it and maybe some case leads to OOM i guess. At least I definitely would like this to be an experimental flag, not enabled by default.

davidmdm:
Would you be amenable to a -workers flag? With default 1? This way users can determine their own memory / concurrency tradeoffs?

internal/engine/wazevo/engine.go
```diff
@@ -227,27 +227,83 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
...
+
+	compiledFuncs := make([]CompiledLocalFuncResult, len(module.CodeSection))
+
+	workers := experimental.GetCompilationWorkers(ctx)
+
+	wg := sync.WaitGroup{}
+	wg.Add(workers)
+
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer cancel(nil)
```
mathetake, line 245: i don't think we need this. compilations are not in-fact cancelable (we don't have any context checking in the compilation routines)

internal/engine/wazevo/engine.go
```diff
@@ -841,3 +897,14 @@ func (cm *compiledModule) getSourceOffset(pc uintptr) uint64 {
...
+
+func sequence(size int) <-chan int {
+	result := make(chan int)
+	go func() {
+		for i := range size {
+			result <- i
+		}
+		close(result)
+	}()
+	return result
```
mathetake, line 909: could you remove this sequence function ? i think it's overkill and don't want to spawn channel/goroutines unnecessarily for the worker=1 case (which is the default for existing projects.)

mathetake: sounds good.

james-lawrence:
@mathetake sorry for the ping but I believe everything you wanted addressed has been addressed?

internal/engine/wazevo/engine.go
```diff
@@ -227,27 +228,88 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
...
+				if wazevoapi.NeedFunctionNameInContext {
+					def := module.FunctionDefinition(fidx)
+					name := def.DebugName()
+					if len(def.ExportNames()) > 0 {
+						name = def.ExportNames()[0]
+					}
+					ctx = wazevoapi.SetCurrentFunctionName(ctx, i, fmt.Sprintf("[%d/%d]%s", i, len(module.CodeSection)-1, name))
+				}
+
+				be := backend.NewCompiler(ctx, machine, ssaBuilder)
```
mathetake, line 286: why `be` is created per function? I believe this could make the single worker case **much** slower and memory consuming. could you fix it to be reused?

mathetake (changes requested):

internal/engine/wazevo/engine.go
mathetake: could you check and see if there's really no impact on the regular compilation by doing the profile for like compiling zig's Stdlib or whatever huge binary as in the Standard Library Tests workflow

james-lawrence:
@davidmdm just bumping this for visibility, do you want some assistance getting this over the line?

davidmdm:
@mathetake 
> could you check and see if there's really no impact on the regular compilation by doing the profile for like compiling zig's Stdlib or whatever huge binary as in the Standard Library Tests workflow

I am not exactly sure how to perform this ask. Any assistance would be appreciated :)

internal/engine/wazevo/wazevoapi/debug_options.go
```diff
@@ -141,9 +135,9 @@ func VerifyOrSetDeterministicCompilationContextValue(ctx context.Context, scope
 	fn := ctx.Value(currentFunctionNameKey{}).(string)
 	key := fn + ": " + scope
 	verifierCtx := ctx.Value(verifierStateContextKey{}).(*verifierState)
-	oldValue, ok := verifierCtx.values[key]
+	oldValue, ok := verifierCtx.values.Load(key)
```
ncruces, line 138: If you're going to use a `sync.Map`, please use `LoadOrStore` here to actually make this change "atomic".

james-lawrence:
@MaanasSathaye @davidmdm I think the problem is related to the fact there are two backend compilers involved (hence 2x allocations). you have the one you're spinning up per routine and the one embedded in the engine.... I suspect the compiler embedded in the engine is the source of most of the problems, including why you can't just nil out those fields in Init.... because it gets shared whereas the routine ones dont.

MaanasSathaye:
Couple of out-of-scope notes. 

Personally don't know how the tests relate to execution in practice, but the assembly loop in `wazero/internal/engine/wazevo/engine.go` has repeated append() calls. Pre-allocating these doesn't reduce total allocations and arguably hurts readability, but reduces memory by ~100MB in the Zig test at least:

```
	totalRels := 0
	totalSourceOffsets := 0
	for i := range compiledFuncs {
		totalRels += len(compiledFuncs[i].RelsPerFunc)
		if needSourceInfo {
			totalSourceOffsets += 1 + len(compiledFuncs[i].SourceOffsetInfo)
		}
	}
	rels = make([]backend.RelocationInfo, 0, totalRels)
	if needSourceInfo {
		cm.sourceMap.executableOffsets = make([]uintptr, 0, totalSourceOffsets)
		cm.sourceMap.wasmBinaryOffsets = make([]uint64, 0, totalSourceOffsets)
	}

	for i := range module.CodeSection {
		fn := compiledFuncs[i]

```

```
     BenchmarkZig/Compile/test.wasm-10                 1        3717765500 ns/op        545416584 B/op    318332 allocp
     BenchmarkZig/Compile/test.wasm-10                 1        3607608167 ns/op        545365792 B/op    318110 allocs/op
     BenchmarkZig/Compile/test.wasm-10                 1        3624120125 ns/op        545355096 B/op    317946 allocs/op
```

Secondly, adding a capacity hint in the ResetMap, `m = make(map[K]V, 32)`, reduces allocations independent of concurrency. 

  1. Worker creates SSA builder (engine.go:254)
  2. Builder allocates basic blocks during lowering (builder.go:161)
  3. Block reset calls ResetMap (basic_block.go:275)
  4. ResetMap creates map
  5. DefineVariable writes to map 93,393 times (builder.go:219)

mathetake (changes requested):
Not following the discussion but let me clarify must-have requirements:
* DO NOT modify unrelated code in this PR, which makes it hard to verify and compare the results fairly. Revert the change in backend/compiler.go
* Use the same code path regardless of the workers count.  Otherwise, the code will _unnecessarily complex_ as well as it makes the future optimization so hard like imagine one works for worker=1 but not for worker=N case. 
* And yes, calling init correctly should be enough to fill the last mile gap (if i haven't missed anything)

davidmdm:
@mathetake updated stats (wzmem is this PR and suffix integers are workers).
Major shout out to @MaanasSathaye for carrying the brunt of the profiling.

TLDR: 4.42% increase in allocations for the single worker case.

```
goos: darwin
goarch: arm64
cpu: Apple M2
                            │ ./wz1.9.0.txt │             ./wzmain.txt              │               wzmem.txt               │               wzmem2.txt               │              wzmem4.txt               │              wzmem8.txt               │
                            │    sec/op     │    sec/op      vs base                │    sec/op      vs base                │    sec/op      vs base                 │    sec/op     vs base                 │    sec/op     vs base                 │
Zig/Compile/test-opt.wasm-8   2380.5m ± ∞ ¹   2301.7m ± ∞ ¹       ~ (p=1.000 n=1) ²   2324.0m ± ∞ ¹       ~ (p=1.000 n=1) ²   1198.1m ± ∞ ¹        ~ (p=1.000 n=1) ²   747.5m ± ∞ ¹        ~ (p=1.000 n=1) ²   556.6m ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Run/test-opt.wasm-8         17.34 ± ∞ ¹     16.81 ± ∞ ¹       ~ (p=1.000 n=1) ²     17.03 ± ∞ ¹       ~ (p=1.000 n=1) ²     16.99 ± ∞ ¹        ~ (p=1.000 n=1) ²    17.49 ± ∞ ¹        ~ (p=1.000 n=1) ²    17.23 ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Compile/test.wasm-8       3020.0m ± ∞ ¹   2855.0m ± ∞ ¹       ~ (p=1.000 n=1) ²   3000.9m ± ∞ ¹       ~ (p=1.000 n=1) ²   1533.9m ± ∞ ¹        ~ (p=1.000 n=1) ²   891.4m ± ∞ ¹        ~ (p=1.000 n=1) ²   753.6m ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Run/test.wasm-8             17.53 ± ∞ ¹     17.11 ± ∞ ¹       ~ (p=1.000 n=1) ²     18.22 ± ∞ ¹       ~ (p=1.000 n=1) ²     17.86 ± ∞ ¹        ~ (p=1.000 n=1) ²    17.84 ± ∞ ¹        ~ (p=1.000 n=1) ²    17.82 ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                         6.837           6.594        -3.55%                     6.821        -0.23%                     4.859        -28.94%                    3.798        -44.46%                    3.369        -50.73%

                            │ ./wz1.9.0.txt │             ./wzmain.txt              │               wzmem.txt               │              wzmem2.txt               │               wzmem4.txt               │               wzmem8.txt                │
                            │     B/op      │     B/op       vs base                │     B/op       vs base                │     B/op       vs base                │     B/op       vs base                 │      B/op       vs base                 │
Zig/Compile/test-opt.wasm-8   373.7Mi ± ∞ ¹   373.7Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   376.5Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   461.0Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   585.9Mi ± ∞ ¹        ~ (p=1.000 n=1) ²    717.9Mi ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Run/test-opt.wasm-8       741.8Mi ± ∞ ¹   741.8Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   741.8Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   741.8Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   741.8Mi ± ∞ ¹        ~ (p=1.000 n=1) ²    741.8Mi ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Compile/test.wasm-8       592.0Mi ± ∞ ¹   592.0Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   534.1Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   653.5Mi ± ∞ ¹       ~ (p=1.000 n=1) ²   817.2Mi ± ∞ ¹        ~ (p=1.000 n=1) ²   1035.8Mi ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Run/test.wasm-8           1.297Gi ± ∞ ¹   1.297Gi ± ∞ ¹       ~ (p=1.000 n=1) ²   1.297Gi ± ∞ ¹       ~ (p=1.000 n=1) ²   1.297Gi ± ∞ ¹       ~ (p=1.000 n=1) ²   1.297Gi ± ∞ ¹        ~ (p=1.000 n=1) ²    1.297Gi ± ∞ ¹        ~ (p=1.000 n=1) ³
geomean                       683.2Mi         683.2Mi        -0.00%                   667.1Mi        -2.36%                   738.0Mi        +8.02%                   828.6Mi        +21.28%                    925.0Mi        +35.39%
³ all samples are equal

                            │ ./wz1.9.0.txt │             ./wzmain.txt             │              wzmem.txt               │              wzmem2.txt               │              wzmem4.txt               │              wzmem8.txt               │
                            │   allocs/op   │  allocs/op    vs base                │  allocs/op    vs base                │  allocs/op    vs base                 │  allocs/op    vs base                 │  allocs/op    vs base                 │
Zig/Compile/test-opt.wasm-8    331.0k ± ∞ ¹   331.0k ± ∞ ¹       ~ (p=1.000 n=1) ²   339.3k ± ∞ ¹       ~ (p=1.000 n=1) ²   400.5k ± ∞ ¹        ~ (p=1.000 n=1) ²   503.8k ± ∞ ¹        ~ (p=1.000 n=1) ²   608.1k ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Run/test-opt.wasm-8        52.02k ± ∞ ¹   52.02k ± ∞ ¹       ~ (p=1.000 n=1) ³   52.02k ± ∞ ¹       ~ (p=1.000 n=1) ²   52.02k ± ∞ ¹        ~ (p=1.000 n=1) ²   52.02k ± ∞ ¹        ~ (p=1.000 n=1) ²   52.02k ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Compile/test.wasm-8        266.7k ± ∞ ¹   266.7k ± ∞ ¹       ~ (p=1.000 n=1) ³   309.3k ± ∞ ¹       ~ (p=1.000 n=1) ²   364.1k ± ∞ ¹        ~ (p=1.000 n=1) ²   460.8k ± ∞ ¹        ~ (p=1.000 n=1) ²   633.9k ± ∞ ¹        ~ (p=1.000 n=1) ²
Zig/Run/test.wasm-8            2.156M ± ∞ ¹   2.156M ± ∞ ¹       ~ (p=1.000 n=1) ²   2.156M ± ∞ ¹       ~ (p=1.000 n=1) ²   2.156M ± ∞ ¹        ~ (p=1.000 n=1) ²   2.156M ± ∞ ¹        ~ (p=1.000 n=1) ²   2.156M ± ∞ ¹        ~ (p=1.000 n=1) ³
geomean                        315.4k         315.4k        -0.00%                   329.4k        +4.42%                   357.6k        +13.37%                   401.7k        +27.35%                   456.0k        +44.56%
³ all samples are equal
```

evacchi:
also, in order to reduce the scope: the changes I originally submitted to the debug options/verifier to ensure that the verifier worked even when running in parallel... probably don't really make sense. I think we can just throw an error if the verifier runs with more than one worker or force a single worker when the verifier is enabled.

james-lawrence:
@ncruces sounds reasonable I will point out the 10% bump is entirely because @mathetake asked for a clean concurrency only PR. there are microoptimizations we can do that drop us 2-5% below the current main branch allocation rate. and iirc the only real controversial bit was the backend compiler changes which are gone.

```text
                                            │   main.out   │             conc.1.out              │              conc.8.out               │
                                            │  allocs/op   │  allocs/op   vs base                │  allocs/op    vs base                 │
Wasip1/Compile/archive_tar.test-16            239.2k ± ∞ ¹   230.8k ± 0%  -3.53% (n=1+6)           401.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/bufio.test-16                  181.4k ± ∞ ¹   174.4k ± 0%  -3.89% (n=1+6)           306.6k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/bytes.test-16                  232.1k ± ∞ ¹   224.7k ± 0%  -3.18% (n=1+6)           381.8k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/context.test-16                205.3k ± ∞ ¹   198.7k ± 0%       ~ (p=0.286 n=1+6)   336.0k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_ascii85.test-16       171.2k ± ∞ ¹   164.0k ± 0%  -4.20% (n=1+6)           300.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_asn1.test-16          186.7k ± ∞ ¹   178.8k ± 0%       ~ (p=1.714 n=1+6)   330.9k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_base32.test-16        173.9k ± ∞ ¹   166.8k ± 0%  -4.10% (n=1+6)           305.4k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_base64.test-16        174.5k ± ∞ ¹   166.9k ± 0%  -4.36% (n=1+6)           309.9k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_binary.test-16        180.4k ± ∞ ¹   172.6k ± 0%  -4.32% (n=1+6)           313.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_csv.test-16           215.3k ± ∞ ¹   207.2k ± 0%  -3.77% (n=1+6)           374.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_gob.test-16           213.0k ± ∞ ¹   204.5k ± 0%  -3.98% (n=1+6)           365.9k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_hex.test-16           173.5k ± ∞ ¹   166.3k ± 0%  -4.12% (n=1+6)           301.6k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_json.test-16          332.5k ± ∞ ¹   325.2k ± 0%  -2.19% (n=1+6)           529.9k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_pem.test-16           207.2k ± ∞ ¹   199.4k ± 0%       ~ (p=0.286 n=1+6)   361.8k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/encoding_xml.test-16           195.3k ± ∞ ¹   186.7k ± 0%  -4.38% (n=1+6)           336.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/errors.test-16                 175.1k ± ∞ ¹   167.4k ± 0%       ~ (p=1.714 n=1+6)   321.1k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/expvar.test-16                 244.1k ± ∞ ¹   236.8k ± 0%  -3.00% (n=1+6)           399.0k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/flag.test-16                   189.0k ± ∞ ¹   181.6k ± 0%       ~ (p=1.714 n=1+6)   328.0k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/fmt.test-16                    279.4k ± ∞ ¹   272.3k ± 0%       ~ (p=1.714 n=1+6)   433.1k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/hash.test-16                   175.5k ± ∞ ¹   168.3k ± 0%  -4.05% (n=1+6)           303.4k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/hash_adler32.test-16           171.7k ± ∞ ¹   164.6k ± 0%       ~ (p=1.714 n=1+6)   297.6k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/hash_crc32.test-16             173.6k ± ∞ ¹   166.4k ± 0%  -4.11% (n=1+6)           303.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/hash_crc64.test-16             172.4k ± ∞ ¹   165.2k ± 0%  -4.16% (n=1+6)           308.2k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/hash_fnv.test-16               175.0k ± ∞ ¹   167.9k ± 0%  -4.06% (n=1+6)           291.8k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/hash_maphash.test-16           187.4k ± ∞ ¹   180.5k ± 0%  -3.65% (n=1+6)           323.4k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/io.test-16                     184.6k ± ∞ ¹   177.7k ± 0%       ~ (p=1.714 n=1+6)   315.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/io_fs.test-16                  183.3k ± ∞ ¹   176.3k ± 0%       ~ (p=1.714 n=1+6)   310.3k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/io_ioutil.test-16              173.1k ± ∞ ¹   165.9k ± 0%  -4.14% (n=1+6)           293.8k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/log.test-16                    183.3k ± ∞ ¹   176.3k ± 0%  -3.78% (n=1+6)           310.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/log_syslog.test-16             174.5k ± ∞ ¹   167.3k ± 0%       ~ (p=1.714 n=1+6)   298.4k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/maps.test-16                   171.7k ± ∞ ¹   164.4k ± 0%       ~ (p=1.714 n=1+6)   290.6k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/math.test-16                   180.6k ± ∞ ¹   173.6k ± 0%       ~ (p=1.714 n=1+6)   305.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/math_big.test-16               280.8k ± ∞ ¹   273.5k ± 0%  -2.61% (n=1+6)           443.8k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/math_bits.test-16              175.4k ± ∞ ¹   168.3k ± 0%       ~ (p=1.714 n=1+6)   309.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/math_cmplx.test-16             173.4k ± ∞ ¹   166.3k ± 0%  -4.12% (n=1+6)           299.3k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/math_rand.test-16              271.4k ± ∞ ¹   264.3k ± 0%  -2.64% (n=1+6)           433.1k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/mime.test-16                   177.3k ± ∞ ¹   170.0k ± 0%  -4.10% (n=1+6)           310.9k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/mime_multipart.test-16         195.1k ± ∞ ¹   187.7k ± 0%  -3.80% (n=1+6)           335.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/mime_quotedprintable.test-16   175.2k ± ∞ ¹   168.1k ± 0%  -4.06% (n=1+6)           306.8k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/os.test-16                     247.9k ± ∞ ¹   241.9k ± 0%       ~ (p=1.714 n=1+6)   381.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/os_exec.test-16                298.3k ± ∞ ¹   292.3k ± 0%       ~ (p=0.286 n=1+6)   485.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/os_signal.test-16              169.7k ± ∞ ¹   162.5k ± 0%  -4.26% (n=1+6)           295.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/os_user.test-16                172.8k ± ∞ ¹   165.7k ± 0%  -4.14% (n=1+6)           295.6k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/path.test-16                   171.4k ± ∞ ¹   164.2k ± 0%  -4.19% (n=1+6)           301.4k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/reflect.test-16                433.4k ± ∞ ¹   416.1k ± 0%       ~ (p=1.714 n=1+6)   626.4k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/regexp.test-16                 209.2k ± ∞ ¹   199.9k ± 0%  -4.44% (n=1+6)           342.3k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/regexp_syntax.test-16          171.9k ± ∞ ¹   164.3k ± 0%  -4.44% (n=1+6)           307.9k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/runtime.test-16                353.4k ± ∞ ¹   340.2k ± 0%  -3.73% (n=1+6)           618.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/slices.test-16                 190.0k ± ∞ ¹   183.1k ± 0%  -3.64% (n=1+6)           310.6k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/sort.test-16                   187.4k ± ∞ ¹   180.6k ± 0%  -3.64% (n=1+6)           314.1k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/strconv.test-16                183.3k ± ∞ ¹   175.8k ± 0%       ~ (p=1.714 n=1+6)   322.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/strings.test-16                233.2k ± ∞ ¹   225.4k ± 0%       ~ (p=0.286 n=1+6)   386.4k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/sync.test-16                   205.5k ± ∞ ¹   198.1k ± 0%       ~ (p=1.714 n=1+6)   348.4k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/sync_atomic.test-16            182.8k ± ∞ ¹   175.9k ± 0%       ~ (p=1.714 n=1+6)   305.6k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/syscall.test-16                175.8k ± ∞ ¹   168.7k ± 0%  -4.03% (n=1+6)           297.7k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/testing.test-16                238.4k ± ∞ ¹   230.8k ± 0%       ~ (p=1.714 n=1+6)   389.6k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/testing_fstest.test-16         187.6k ± ∞ ¹   180.5k ± 0%  -3.79% (n=1+6)           316.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/testing_iotest.test-16         173.8k ± ∞ ¹   166.5k ± 0%       ~ (p=1.714 n=1+6)   303.8k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/testing_quick.test-16          178.0k ± ∞ ¹   169.8k ± 0%       ~ (p=1.714 n=1+6)   317.5k ± ∞ ¹        ~ (p=1.000 n=1) ²
Wasip1/Compile/time.test-16                   261.3k ± ∞ ¹   254.0k ± 0%  -2.81% (n=1+6)           418.0k ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                       201.0k         193.4k       -3.78%                   341.1k        +69.65%
```

ncruces:
My branch is https://github.com/wazero/wazero/compare/main...ncruces:wazero:conc. I encourage folks to "review" each commit there in isolation.

@evacchi I don't think those changes are necessarily controversial, I've actually included them in "fixes": https://github.com/wazero/wazero/commit/8a1019a0553caba752a2795343862b8173dfd3da. The only bit I might move to "prep work" is the `sync.Map`, but it's only unnecessary; it's not a maintenance burden, nor does it uglify anything, or make things measurably worse.

Even the prep work in https://github.com/wazero/wazero/commit/aeb4433c613b4b6be801f0998e5ba47440a8324b seems mostly uncontroversial to me, so we're really down to https://github.com/wazero/wazero/commit/892f11e071bcdcd0a8a18f2dda63acd775806148.

GitHub's diff is unhelpful here, this is really just a matter of making the independent part of the loop parallel (let's call it section 1), store some intermediate results in memory, and then wrap up the part of the loop than has interdependencies (section 2).

I see 3 possible solutions:
- eat the cost of this, it's not terrible after all
- refactor so sections 1 and 2 are reusable bits that can be called in both arrangements (concurrent or not)
- somehow introduce stalls so goroutines can do the parallel thing in parallel, then wait to do the serial bit before doing more parallel work

Just to put some numbers to it, the "problem" really is with just making things parallel this way (parallel, intermediate results, wrap up):
```
% benchstat main fix* conc 
goos: darwin
goarch: amd64
                             │    main    │             fix+prep             │               conc                │
                             │   sec/op   │   sec/op    vs base              │   sec/op     vs base              │
Zig/Compile/test-opt.wasm-12   4.551 ± 4%   4.555 ± 3%       ~ (p=0.818 n=6)   4.571 ± 26%       ~ (p=1.000 n=6)
Zig/Compile/test.wasm-12       5.504 ± 3%   5.432 ± 1%       ~ (p=0.240 n=6)   5.520 ±  3%       ~ (p=1.000 n=6)
geomean                        5.005        4.974       -0.62%                 5.023        +0.36%

                             │     main     │              fix+prep              │                conc                 │
                             │     B/op     │     B/op      vs base              │     B/op      vs base               │
Zig/Compile/test-opt.wasm-12   359.5Mi ± 0%   359.5Mi ± 0%       ~ (p=0.732 n=6)   362.2Mi ± 0%   +0.76% (p=0.002 n=6)
Zig/Compile/test.wasm-12       574.3Mi ± 0%   574.3Mi ± 0%       ~ (p=0.370 n=6)   516.3Mi ± 0%  -10.10% (p=0.002 n=6)
geomean                        454.4Mi        454.4Mi       +0.00%                 432.5Mi        -4.82%

                             │    main     │             fix+prep              │                conc                │
                             │  allocs/op  │  allocs/op   vs base              │  allocs/op   vs base               │
Zig/Compile/test-opt.wasm-12   336.2k ± 0%   336.2k ± 0%       ~ (p=0.913 n=6)   344.5k ± 0%   +2.47% (p=0.002 n=6)
Zig/Compile/test.wasm-12       289.8k ± 0%   289.8k ± 0%       ~ (p=0.446 n=6)   332.4k ± 0%  +14.71% (p=0.002 n=6)
geomean                        312.1k        312.1k       -0.00%                 338.4k        +8.42%
```

@james-lawrence other micro-optimizations we can always do or not do. I'm genuinely just trying to make this easier to review _because_ although I'm personally lost, it's not fair to @davidmdm that after 80 comments and so much work this is simply dropped _because_ I'm lost. But, as you may have noticed, the project is in maintenance mode, and we need to be very mindful of what features we take in so as to make it continuously useful, without increasing the maintenance cost.

mathetake (approved):

PR #2386, opened 2025-03-05
cache/memory: Disable cache for host modules

mqasimsarfraz:
I wanted to understand how cache grows overtime I realized it kept growing because of host modules. I assume this is because they [won't have same IDs across multiple runs](https://github.com/tetratelabs/wazero/blob/cc40233f33197942e79997c3e6fccbe453ddb535/internal/wasm/host.go#L69) so I wanted to propose to disable it like [we did for file system cache](https://github.com/tetratelabs/wazero/pull/949). Having said that I am not calling close on the runtime. Let me know if I am missing something perhaps there is a use-case I am over looking? 

Following is the screenshot from the debugger where I saw `compiledModules` kept on growing:

Screenshot from 2025-03-05 10-18-28

Screenshot from 2025-03-05 10-19-06

### DIFF
--- a/internal/engine/wazevo/engine_cache.go
+++ b/internal/engine/wazevo/engine_cache.go
@@ -43,8 +43,11 @@ func fileCacheKey(m *wasm.Module) (ret filecache.Key) {
 }
 
 func (e *engine) addCompiledModule(module *wasm.Module, cm *compiledModule) (err error) {
+	if module.IsHostModule {
+		return
+	}
 	e.addCompiledModuleToMemory(module, cm)
-	if !module.IsHostModule && e.fileCache != nil {
+	if e.fileCache != nil {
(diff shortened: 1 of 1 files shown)


mathetake:
if the host module compilation is a bottleneck, then i would strongly suggest to reconsider your design (like use context and single host. Usually people only initialize host modules on the startup, not creating with unbound cardinality). sorry this will not be merged.

mqasimsarfraz:
> if the host module compilation is a bottleneck, then i would strongly suggest to reconsider your design (like use context and single host. Usually people only initialize host modules on the startup, not creating with unbound cardinality). sorry this will not be merged.

Thanks for the quick feedback! host module compilation isn't really bottleneck but it didn't make sense to add them to cache (`compiledModules`) but I do agree this could be handled better on our side! :)

PR #2395, opened 2025-04-11
Optimize/streamline fill operations

ncruces:
This optimizes/streamlines fill operations.

#### Backstory:
- since 1.21 we have a [`clear`](https://tip.golang.org/doc/go1.21#language) builtin which is faster than `copy` when zeroing memory
- [`slices.Repeat`](https://github.com/golang/go/blob/go1.24.0/src/slices/slices.go#L514-L517) uses a `copy` loop very similar to the one we're using
- [`bytes.Repeat`](https://github.com/golang/go/blob/go1.24.0/src/bytes/bytes.go#L664-L673) does the same, but adds an 8KB maximum to the chunk size
- the existing generated code could be streamlined

#### Intuition:
- rare that tables are much larger than 8KB
- using `clear` (actually `runtime.memclrNoHeapPointers`) from native code is harder
- favor smaller/simpler code for tables, faster code for memory (like the standard library)

#### For the compiler:
- streamline/simplify `table.fill`
- optimize `memory.fill` using the 8KB maximum chunk size

#### For the interpreter:
- keep `table.fill` unchanged
- optimize `memory.fill` with both `clear` and maximum chunk size

#### For `table.Grow`:
- recognize that zero initialization is already done

#### Results:
I could measure an over 2x improvement filling megabytes of memory, with no degradation in performance at small sizes. No increase of generated code size.

#### Future work:
Using `memory.fill` to zero memory is _very_ common. If we could access `runtime.memclrNoHeapPointers` in the compiler, we could potentially gain another 25% there (especially if we optimized this when zero is known at compile time).

### DIFF
--- a/internal/engine/interpreter/interpreter.go
+++ b/internal/engine/interpreter/interpreter.go
@@ -1725,12 +1725,17 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, m *wasm.ModuleInstance
 			if fillSize+offset > uint64(len(memoryInst.Buffer)) {
 				panic(wasmruntime.ErrRuntimeOutOfBoundsMemoryAccess)
 			} else if fillSize != 0 {
-				// Uses the copy trick for faster filling buffer.
-				// https://gist.github.com/taylorza/df2f89d5f9ab3ffd06865062a4cf015d
+				// Uses the copy trick for faster filling the buffer with the value.
+				// https://github.com/golang/go/blob/go1.24.0/src/bytes/bytes.go#L664-L673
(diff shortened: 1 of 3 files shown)


mathetake (approved):
Fantastic

ncruces:
That was way faster than I expected. 😂
I kinda expected this to be torn to pieces.

It has since occured to me that `((i - 1) & 8191) + 1` might not be better than `min(i, 8192)`. Wdyt?

I know very little of the back ends to know what generates better code. “Branches are bad, conditional moves not so much” is the advanced-ness of my understanding.

PR #2403, opened 2025-06-14
Tail call

evacchi:
Implements the tail call extension proposal. Besides the included test cases, I have verified that this works with a simple factorial compiled with GHC, and also made an in-depth analysis on SQLite. Thanks to @ncruces's invaluable help, as usual <3, we discovered a lot of edge cases that were not covered in the spec test suite.

I understand we're all pretty busy these days, so I am obviously not calling for an urgent review :)

---

The PR adds **Feature Flags** modeled after those for Threads:

   - Added a new experimental feature flag `CoreFeaturesTailCall` to enable tail call support
   - Users need to explicitly enable this feature to use tail call instructions

The PR adds support in both the interpreter and compiler backends, I have also added a few e2e examples, some adhoc tests, and the spec testsuite.

## Caveats and Notable Changes

- Notably, the implementation in the compiler for both backends restores the stack as it would normally do in the epilogue (in fact, it invokes `setupEpilogueAfter()` in `postRegAlloc()`), then it jumps to the beginning of the function, thus it is not "just" a jump: it required adding corresponding pseudo-instructions to treat them specially.

- In particular, on `amd64`, indirect calls required to mov the function pointer to a safe (caller-saved) register (I picked `r11`) to make sure it was not overwritten in the epilogue (which occurred to me in my testing).

### Edge Cases

The following edge cases were only discovered through testing SQLite, and I have added e2e and ad-hoc test to reproduce them. SQLite (well, LLVM with tail calls enabled) does not necessarily make use of tail calls for recursion, but most often to optimize a call to a private function usually from a public, exported function. So this exposes a number of interesting situations that the spec tests do not account for (even though they are definitely not forbidden, according to the spec), like tail calling host functions.
 
1. In the **interpreter**,  I am pretty much just resetting the program counter and pointing the function body to the function that needs to be called; this does not work very well with **imported functions** (i.e. functions defined in a different module) and in particular this won't flight with _host functions_! However, since the tail call is mostly an optimization, I have decided that for now it is not worth optimizing these cases --this is experimental, after all; the interpreter will _fall back to a plain call_ in these cases.

2. In the **compiler** there are fewer limitations, with a _bigger constraint_: it is **hard to perform tail calls in the presence of stack arguments**  (i.e. given the types have to match because of validation, this occurs when the number of args is above a given threshold, different in the ARM and AMD calling conventions [*]) the reasons are 
  a. if the sizes of the required stack space do not match, then we might be writing garbage
  b. even if the sizes match, we would need special attention not to clear the stack completely before we jump into the callee
  c. even if we managed to do that, the spill slots are not necessarily matching so, we would still need some special care (notice that other runtimes have completely revolutionized their internal calling conventions _for all cals_ to accommodate tail calls)

So, again, in these cases I have decided to **fall back to plain calls**.

Now, this could be addressed in two ways:

  - in the _front-end_ by adding a fixed constant `maxArgs` that would be set to `min(regArgsAmd, regArgsArm) == 8` and calling it a day

    this has the only downside of surfacing into the back-end a constant that is effectively arch-specific. If you want to see how that looks, [take a look at this branch --- this lacks a lot of polish, but you'll get the idea](https://github.com/tetratelabs/wazero/compare/main...evacchi:wazero:tail-call-fixes-compiler#diff-2df377ff6055fa1b350d733ac8c0e131c4711f946a5471b5a30d474e6befaf5fR3660-R3711)

  - in the _back-end_ with proper arch-specific handling, which is the solution I went with in the end

      unfortunately,  because of the way and point in time we are emitting `Return`s, I think the best way to do this in the back-end is _always_ emitting returns, basically treating tail calls effectively as plain calls almost until we get to the encoding phase. This way, we get the regalloc phase to work as usual; then:
      
      - if the tail call **cannot** be handled, then we _just flip it to a traditional call_, and then all the return handling will be reachable code; 
      - if the tail call CAN be handled, THEN we just remove the useless instructions between the tail-call and the ret instruction during Finalize (postRegAlloc()), as it's dead code anyway.

[*] fun aside: this made isolating the issue particularly frustrating because the same set of test was failing on ARM and magically passing on AMD... because (surprisingly, to me) in AMD the calling convention allows to pass more args via registers than ARM before falling back to the stack


ncruces (approved):
Few comments left.

Also, if this gets submitted, I think some of the PR description should go into the `RATIONAL.md` (or maybe one in the experimental folder).

I've additionally verified this doesn't seem to make things measurably worse if unused (in the compiler at least). And my battery of tests passes on all platforms (not sure if @evacchi had already tested CI): https://github.com/ncruces/go-sqlite3/actions/runs/15657504181

evacchi:
I thought about RATIONALE.md too, but I realized we didn't have any specific section for the compiler internals, but I am 100% in favor of documenting these choices somewhere.

internal/engine/interpreter/interpreter.go
```diff
@@ -4336,13 +4338,80 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, m *wasm.ModuleInstance
...
+				// Return
+				ce.drop(op.Us[0])
+				// Jump to the function frame (return)
+				frame.pc = op.Us[1]
+				continue
+			}
+
+			ce.dropForTailCall(frame, tf)
+			if *tf == *(frame.f) {
+				// Short-circuit self-recursion.
```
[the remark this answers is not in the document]
evacchi: right 🤔 originally the code below was more articulate, so this made more sense

evacchi:
(failure obviously unrelated, it's a flaky test on Windows)

internal/engine/RATIONALE.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Interpreter
+
+There are 3 cases that needs to be handled in the interpreter:
+
+1. **Tail calls to the same function**: this is the simplest case, where we just reset the program counter to the start of the function body. This is straightforward and does not require any special handling.
+2. **Tail calls to a different function**: this is also straightforward, as we just reset the program counter to the start of the function body, then replace the function body with the new function's body. 
+3. **Tail calls to a host function**: in this case, we cannot simply reset the program counter, as the host function is not part of the WebAssembly module. Instead, we fall back to a plain call to the host function, which is the same as if we were calling it from a regular function.
+
+As a further restriction, currently the interpreter does not handle tail calls to any **imported functions** and falls back to a plain call in these cases too.
```
ncruces, line 20: This means functions from other modules? Are there more?

ncruces (commented):
I'm left wondering how the interpreter restrictions are handled in the compiler. Does it handle calls to host functions and to other modules if they're compatible?

evacchi:
> I'm left wondering how the interpreter restrictions are handled in the compiler. Does it handle calls to host functions and to other modules if they're compatible?

it does, maybe I should clarify. In the compiler indirect calls are all treated in the same way (it's a function pointer), so they don't require special handling

internal/engine/wazevo/ssa/instructions.go
```diff
@@ -2620,6 +2653,17 @@ func (i *Instruction) Format(b Builder) string {
 		instSuffix = fmt.Sprintf("_%d, %s, %s, %s", 8*i.u1, i.v.Format(b), i.v2.Format(b), i.v3.Format(b))
 	case OpcodeFence:
 		instSuffix = fmt.Sprintf(" %d", i.u1)
+	case OpcodeTailCallReturnCall, OpcodeTailCallReturnCallIndirect:
+		view := i.vs.View()
+		vs := make([]string, len(view))
```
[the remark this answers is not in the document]
evacchi: that's right, it's a copy-paste

mathetake (commented):
sorry about the delay. Did you run the fuzzer this feature enabled?

mathetake:
do you want to enable the flag by default on the fuzzer run (so that we can catch a bug on CI as well)

evacchi:
ah right, let me push it here

internal/engine/interpreter/interpreter.go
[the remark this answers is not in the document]
mathetake: I would prefer to drop this for simplicity (not sure on why we want to optimize interpreter for this code path alone)

internal/engine/interpreter/interpreter.go
```diff
@@ -4336,13 +4338,82 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, m *wasm.ModuleInstance
 			memoryInst.Mux.Unlock()
 			ce.pushValue(uint64(old))
 			frame.pc++
+		case operationKindTailCallReturnCall:
+			f := &functions[op.U1]
+			ce.dropForTailCall(frame, f)
+			if *f == *(frame.f) {
+				// Short-circuit self-recursion.
```
mathetake, line 4345: let's drop it

internal/engine/RATIONALE.md
mathetake: could you describe on why part of this restriction (that's the rationale i guess? this is not a rationale)

mathetake (commented):
Reviewed only interpreters

internal/engine/interpreter/interpreter.go
```diff
@@ -4336,13 +4338,82 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, m *wasm.ModuleInstance
...
+			}
+
+			// The compiler is currently allowing proper tail call only across functions
+			// that belong to the same module; thus, we can overwrite the frame in-place.
+			// For details, see internal/engine/RATIONALE.md
+			frame.f = f
+			frame.base = len(ce.stack)
+			frame.pc = 0
+			body = frame.f.parent.body
+			bodyLen = uint64(len(body))
```
mathetake, line 4357: also can you share this with non-tailcallindirect and tailcallindirect and share the comment

internal/engine/interpreter/interpreter.go
```diff
@@ -4336,13 +4338,82 @@ func (ce *callEngine) callNativeFunc(ctx context.Context, m *wasm.ModuleInstance
...
 	}
 	ce.popFrame()
 }
 
+func (ce *callEngine) dropForTailCall(frame *callFrame, f *function) {
+	base := frame.base - frame.f.funcType.ParamNumInUint64
+	paramCount := f.funcType.ParamNumInUint64
+	if len(ce.stack) < base+paramCount {
+		panic(fmt.Sprintf("tail call: stack underflow: have %d, need %d", len(ce.stack)-base, paramCount))
+	}
```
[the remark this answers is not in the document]
evacchi: I think this is a leftover from debugging! I'll double check!

internal/integration_test/fuzz/fuzz/fuzz_targets/validation.rs
```diff
@@ -18,6 +18,7 @@ fn run(data: &[u8]) -> Result<()> {
     // Generate the configuration with possibly invalid functions.
     let mut config = Config::arbitrary(&mut u)?;
     config.threads_enabled = true;
+    config.tail_call_enabled = true;
```
mathetake, line 21: this needs to be enabled for no_diff tests not only for the validation.

evacchi: should be done. If that is all (it's been a while 🙂) I reran the tests on arm64+amd64 and got no issues

evacchi:
Thanks. You will still find some duplicated code. It was originally intentional for clarity, I'll refactor for reuse if you believe it's better

PR #2410, opened 2025-07-22
Ensure interpreter.Engine is empty after closing

QuLogic:
If the compiler is not supported on the current platform, then `TestCompilationCache/non-host module` will load the interpreter instead. At the end of the test, it checks that the cache is empty, but that was not previously the case for the interpreter.

### DIFF
--- a/internal/engine/interpreter/interpreter.go
+++ b/internal/engine/interpreter/interpreter.go
@@ -43,6 +43,7 @@ func NewEngine(_ context.Context, enabledFeatures api.CoreFeatures, _ filecache.
 
 // Close implements the same method as documented on wasm.Engine.
 func (e *engine) Close() (err error) {
+	clear(e.compiledFunctions)
 	return
 }
 


mathetake (changes requested):
> If the compiler is not supported on the current platform, then TestCompilationCache/non-host module will load the interpreter instead. At the end of the test, it checks that the cache is empty, but that was not previously the case for the interpreter.

Fix itself looks alright but are you saying that the tests are not running on non-compiler platforms? i don't think that is the case as CI are running for RISCV as well. Could you create a repro unit test and push it?

QuLogic:
> Fix itself looks alright but are you saying that the tests are not running on non-compiler platforms? i don't think that is the case as CI are running for RISCV as well. Could you create a repro unit test and push it?

No, they don't appear to be. See this [scratch build with higher verbosity](https://github.com/QuLogic/wazero/actions/runs/16586101065/job/46911457682#logs).

Top-level tests are [built here](https://github.com/QuLogic/wazero/actions/runs/16586101065/job/46911457682#step:4:10)
```
$ go list -f '{{.Dir}}' ./... | egrep -v 'spectest' | xargs -t -Ipkg go test pkg -c -o pkg.test
go test /home/runner/work/wazero/wazero -c -o /home/runner/work/wazero/wazero.test
```
then they are run as:
```
  find . -name "*.test" | xargs -t -Itestbin docker run --platform linux/riscv64 -v $(pwd)/testbin:/test -v $(pwd)/wazerocli:/wazero -e WAZEROCLI=/wazero --tmpfs /tmp --rm -t wazero:test
```
but that `find` is in the current directory while the top-level tests are compiled into a test that is a directory above.

PR #2412, opened 2025-07-22
Add ppc64le&s390x Linux to supported stat implementation

QuLogic:
These platforms pass the stat tests just fine with the existing implementation.

### DIFF
--- a/internal/sysfs/stat_linux.go
+++ b/internal/sysfs/stat_linux.go
@@ -1,4 +1,4 @@
-//go:build (amd64 || arm64 || riscv64) && linux
+//go:build (amd64 || arm64 || ppc64le || riscv64 || s390x) && linux
 
 // Note: This expression is not the same as compiler support, even if it looks
 // similar. Platform functions here are used in interpreter mode as well.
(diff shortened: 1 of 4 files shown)


QuLogic:
Well, all I can tell you is that I've run the tests and they work. If you want to setup more cross-arch tests, then I can, or you can do downstream checks in Fedora with [Packit](https://packit.dev/).

ncruces:
For what's worth it, I test both these platforms (as well as `386` for 32-bit and `s390` for big endian) at [`ncruces/go-sqlite3`](https://github.com/ncruces/go-sqlite3/blob/0399f10c0640cfb7a335762dbe961c30ba269130/.github/workflows/test.yml#L192) with Qemu.

The interpreter should be fine, but WASI is completely untested. Given it's Linux, the bar shouldn't be too high to get things working, though. The major issue here is that the structs need to have the same layout (which is not a given across archs).

I'm of the opinion that, at least the WASI implementation, if it's not spun off, should start using `x/sys/unix` to take these matters as much as possible off our hands.

ncruces:
For Linux we probably just use `docker/setup-qemu-action` and run a subset of tests. I regularly test wazero that way. In the past I think we didn't do that because of concerns about quotas? What's the situation now?

Another solution for this is to simply depend on `x/sys/unix`: push the responsibility there, and assume things should work if they're tested on Linux/macOS/BSD.

This would be a breaking change with the "zero dependencies" motto, but: Go maintainers maintain `x/sys/unix`, and it's their stated goal that syscalls should go through there.

PR #2415, opened 2025-08-02
TST: Increase buffer size of TestWriteFdNonblock

QuLogic:
The default pipe size is 1M (see `man 7 pipe`), and this test is only trying to fill it with 100K, so increase the fill size so it isn't so flaky.

### DIFF
--- a/internal/sysfs/file_test.go
+++ b/internal/sysfs/file_test.go
@@ -99,8 +99,9 @@ func TestWriteFdNonblock(t *testing.T) {
 
 	// Create a buffer (the content is not relevant)
 	buf := make([]byte, 1024)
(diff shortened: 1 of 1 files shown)


internal/sysfs/file_test.go
```diff
@@ -100,7 +100,7 @@ func TestWriteFdNonblock(t *testing.T) {
 	// Create a buffer (the content is not relevant)
 	buf := make([]byte, 1024)
 	// Write to the file until the pipe buffer gets filled up.
-	numWrites := 100
+	numWrites := 1111
```
mathetake, line 103: I think this change itself seems fine, but could you put the documentation comment here on why we have 1111 here.

mathetake (approved):

PR #2422, opened 2025-09-23
chore: rename import paths to github.com/wazero/wazero

Xe:
Also handles the site with a lazy sed, may not be the best idea?


mathetake:
so we (all maintainers) agree with what @ncruces said. We won't do this right away but will take it as an opportunity where we can break "something". wazero has never been breaking any existing users before and that's our philosophy so we won't take this lightly

Issue #2336, opened 2024-11-06
Compiled modules may share state


emcfarlane:
**Describe the bug**
Closing a `CompiledModule` will affect another `CompiledModule` if they are compiled to the same runtime using the same wasm bytes.

**To Reproduce**

```go
package main

import (
	"context"
	"log"

	"github.com/tetratelabs/wazero"
)

func main() {
	ctx := context.Background()
	rt := wazero.NewRuntimeWithConfig(ctx, wazero.NewRuntimeConfig())

	wasm := []byte{0, 'a', 's', 'm', 1, 0, 0, 0}
	compiled1, _ := rt.CompileModule(ctx, wasm) // Original.
	compiled2, _ := rt.CompileModule(ctx, wasm) // Duplicate.

	// Instantiating the module then closing the compiled module should not affect the other compiled module.
	if _, err := rt.InstantiateModule(ctx, compiled1, wazero.NewModuleConfig().WithName("foo")); err != nil {
		log.Fatal(err)
	}
	_ = compiled1.Close(ctx)

	// This compiled2 module should still be usable.
	if _, err := rt.InstantiateModule(ctx, compiled2, wazero.NewModuleConfig().WithName("bar")); err != nil {
		log.Fatal(err)
	}
	_ = compiled2.Close(ctx)
}
```
This causes the following error on the second log line:
```
source module must be compiled before instantiation
 ```

https://github.com/tetratelabs/wazero/compare/main...emcfarlane:wazero:ed/dupCompCache

**Expected behavior**
The `CompiledModule` should still be able to be instantiated as the caller isn't aware that the module is shared between multiple instances.

**Screenshots**
N/A

**Environment (please complete the relevant information):**
- Go version: go version go1.23.2 darwin/arm64
- wazero Version: dc08732e57d575a240774bc64e87cbed59f81c34
- Host architecture: arm64
- Runtime mode: compiler

**Additional context**
The current behaviour makes it difficult to bound the number of active modules. The caller needs to hash the bytes (as there is no ModuleID provided in the interface of `CompiledModules`. 

Related issues https://github.com/tetratelabs/wazero/issues/1913

PR #2427, opened 2025-10-22
Introduce reference counting for compiled modules

burak-ok:
Hello,

This PR tries to take a stab at #2336.
I tried to find a good layer where to put this kind of reference counting to not have duplication across the engines. But I couldn't find one without any further race conditions.

I intentionally created this as a Draft to first to gather your opinions about this way of solving the problem outlined in the linked issue.
The RLock change to Lock can be circumvented when "doubling" this function so that one is specifically for increasing the ref and the other one only reading it

### DIFF
--- a/internal/engine/interpreter/interpreter.go
+++ b/internal/engine/interpreter/interpreter.go
@@ -37,7 +37,7 @@ type compiledFunctionWithCount struct {
 type engine struct {
 	enabledFeatures   api.CoreFeatures
 	compiledFunctions map[wasm.ModuleID]*compiledFunctionWithCount // guarded by mutex.
-	mux               sync.RWMutex
+	mux               sync.Mutex
 }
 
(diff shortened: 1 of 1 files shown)


anuraaga:
Thanks @burak-ok. From what I can tell with our current design, reference counting like this PR is the only solution and it seems ok to me.

One thought that came to mind is that the main issue is that explicitly compiled and implicitly compiled modules are implemented the same way, so ownership is all in the engine. A refactoring that comes to mind is to have the code not be owned by the engine but by CompiledModule itself, or the module otherwise (more like create or accept the CompiledModule). It's similar to how some RPC frameworks can accept an http client (connection pool) in which case they won't close it, or automatically create if not provided, in which case they do close it. Not sure if such a refactoring is worth it though - @ncruces what do you think?

ncruces:
I'm currently lacking the mental energy to "see" (and pursue) the refactoring suggested by @anuraaga. So this seems to make sense to me.

If anyone wants to take on that refactoring, and it streamlines things (makes things more maintainable, etc) I'm happy to review it!

PR #2432, opened 2025-10-28
Align preambles, trampolines to 16 bytes.

ncruces:
I'm afraid I may have messed up #2428.

I'm not aligning preambles, trampolines or listeners to 16 bytes, which we're otherwise doing for all functions.

This doesn't seem to affect performance much, and what I've read about it is inconclusive, but for consistency this is something we should either always do, or never do.

This fixes it by always aligning these. I'm not sure I'm doing perfmaps correctly. Is it important that the last few up to 15 zero filled bytes that are never run are not included in the size or is this a non-issue?

### DIFF
--- a/internal/engine/wazevo/engine_test.go
+++ b/internal/engine/wazevo/engine_test.go
@@ -80,6 +80,51 @@ func TestEngine_CompileModule(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


mathetake (commented):
can we somehow write some test so we won't mess up later

PR #2435, opened 2025-10-29
Use slices.Clone.

ncruces:
In #2381 I didn't notice we cloud use `slices.Clone` and copied it into `engine.go`.

This fixes that, uses it in other places I could find, as well as the `append(..., make(...))` idiom in some other places.


davidmdm:
Yeah I had thought of it, but I was concerned the `append` implementation of the slices package would not be as performant as `copy`.

I didn't benchmark the difference. I think it makes no difference in the end.

PR #2436, opened 2025-11-01
wazevo+wazevoapi: reduce mem by 10%

MaanasSathaye:
```
maanas@maanas wazero % ~/go/bin/benchstat baseline.txt current.txt
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
                         │ baseline.txt │         current.txt         │
                         │    sec/op    │   sec/op    vs base         │
Zig/Compile/test.wasm-10     3.464 ± 3%   3.452 ± 7%  ~ (p=1.000 n=6)

                         │ baseline.txt │            current.txt             │
Zig/Compile/test.wasm-10   466.9Mi ± 0%   420.5Mi ± 0%  -9.93% (p=0.002 n=6)

                         │ baseline.txt │            current.txt            │
                         │  allocs/op   │  allocs/op   vs base              │
Zig/Compile/test.wasm-10    228.2k ± 0%   225.8k ± 0%  -1.06% (p=0.002 n=6)
```

ResetSlice/ResetMap calls were allocating with zero capacity, causing multiple reallocations as slices grew.
Added capacity hints based on known sizes (e.g. number of functions, estimated source map entries) when calling ResetSlice/ResetMap to pre-allocate to appropriate sizes and avoid reallocation overhead.

These, imo, are the last remaining "gimmes". Profiling revealed VarLengthPool and slice growth as top allocators during compilation, consuming ~110MB and ~58MB respectively. Without addressing these, we're likely attempting to draw water from stone at this point.

### DIFF
--- a/internal/engine/wazevo/engine.go
+++ b/internal/engine/wazevo/engine.go
@@ -234,7 +234,7 @@ func (e *engine) compileModule(ctx context.Context, module *wasm.Module, listene
 
 	needSourceInfo := module.DWARFLines != nil
 	if needSourceInfo {
-		estimatedOffsets := localFns * 20
+		estimatedOffsets := localFns * 150
 		cm.sourceMap.executableOffsets = make([]uintptr, 0, estimatedOffsets)
 		cm.sourceMap.wasmBinaryOffsets = make([]uint64, 0, estimatedOffsets)
 	}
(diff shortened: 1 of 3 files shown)


MaanasSathaye:
Maybe I should pose a question first before I make a fool of myself in a further MR:

The frontend tracks memory bounds for values across basic block boundaries:
  - `finalizeKnownSafeBoundsAtTheEndOfBlock()` captures bounds at block exit
  - `initializeCurrentBlockKnownBounds()` propagates bounds from predecessors
  - mysterious intersection logic for merge points with multiple predecessors
  
Why was cross-block bounds check propagation implemented?  In my head I'm playing with keeping bounds local to each block.

Issue #2434, opened 2025-10-29
Change version policy to two versions and allow golang/x


anuraaga:
The topic of using golang/x, especially sys, has come up for improving portability. This issue is to push for it.

Much of golang/x is vendored into the Go compiler itself. So it's unusual to have compatibly issues with code supported by the Go team. We have traditionally taken a looser version policy though, and IIRC grpc was one example cited when we supported more than Go. Well that ended.

https://github.com/grpc/grpc-go/issues/7249

So I think in this case we shouldn't try to beat Google and go with the flow - if we reduce our supported Go versions I see no reason to avoid /x libraries since users shouldn't end up with any problems.

mathetake:
+1

ncruces:
If we all agree, I'd tag one more release under the current policy, and then switch.

As for when, I think we should consider fixing #2427, then tag `v1.10.0` if nothing else advises against it.

mathetake:
> If we all agree, I'd tag one more release under the current policy, and then switch.
> 
> As for when, I think we should consider fixing [#2427](https://github.com/wazero/wazero/pull/2427), then tag `v1.10.0` if nothing else advises against it.

sg @ncruces

PR #2443, opened 2025-11-10
Use golang.org/x/sys

ncruces:
As discussed in #2434. This adds `golang.org/x/sys` as a dependency:
- `golang.org/x/sys/cpu` for CPU detection
- `golang.org/x/sys/unix`/`windows` for (W^X) memory mapping
- `golang.org/x/sys/unix`/`windows` for WASI

----

~This is just so I can explore improvements that increasing the minimum Go version and allowing use of `golang.org/x/sys` permit.~

~It's particularly helpful to run CI across platforms for this effort.~

~Feel free to comment, give suggestions, but in principle I'll cherry pick commits from here for review if the benefits become clear.~

~First commit sets the tone and uses `x/sys/cpu` for CPU detection; despite a mismatch with one detected flag, the improvement is apparent: removes 4× the code it adds.~

### DIFF
--- a/RATIONALE.md
+++ b/RATIONALE.md
@@ -7,23 +7,18 @@ have heavy impact usually due to CGO. By avoiding CGO, wazero avoids
 prerequisites such as shared libraries or libc, and lets users keep features
 like cross compilation.
 
-Avoiding go.mod dependencies reduces interference on Go version support, and
-size of a statically compiled binary. However, doing so brings some
+Avoiding most `go.mod` dependencies reduces interference on Go version support,
+and size of a statically compiled binary. However, doing so brings some
 responsibility into the project.
 
-Go's native platform support is good: We don't need platform-specific code to
+Go's native platform support is good: we don't need platform-specific code to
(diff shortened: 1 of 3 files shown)


ncruces:
So OpenBSD still doesn't work. Maybe someone wants to tackle that.

ncruces:
I'd say in terms of maintenance, this is the actual benefit of `golang.org/x/sys`.

The stats (350 LoC removed) hides some improvements (that we've refused because they'd require _more_ LoC):
- Solaris & illumos now have `mprotect`
- NetBSD/ARM64 now supports atomics
- older Linux/ARM64 no longer crashes checking for atomics

The major negative for me, has been `x/sys/cpu` not having the explicit `ABM` bit (but I'm convinced my solution is correct, and that this actually brings CTZ to more CPUs; not that it's a big win).

There's also, for the future, the factor "it's there, we might as well use it."

We can always decided not to do this. @mathetake @anuraaga @evacchi

anuraaga (approved):
Less assembly code seems like a nice win anyways

PR #2447, opened 2025-11-14
Fix race condition in refCount initialization

jackorse:
This pr solves a race condition that occurs when multiple goroutines compile the same module.

Changes:
* Modified `addCompiledFunctions` in `interpreter.go` so that adding functions for an already cached function increments its reference count instead of replacing the entry.
* Modified `addCompiledModuleToMemory` in `engine_cache.go` so that it increments the reference count and returns the compiled module if it was already cached, otherwise it returns the newly added module.


ncruces (approved):
Seems reasonable.

It's a waste that we're compiling the module twice to then drop one of them, but maybe it's not worth the complexity to fix that.

PR #2451, opened 2025-11-30
Implement memory64

dzbarsky:
I'm not familiar with this codebase; this PR was AI-generated. I did look through it and fix up a few of the things that looked wrong, but it's certainly possible (probable) that there are more issues. The best way to address is probably to add additional tests :) This does seem to work for my test wasm64 blob, which I've included in this PR


ncruces:
A new Wasm feature needs to pass the spectest and be fuzzed. See: https://github.com/wazero/wazero/pull/2403

Also, it doesn't make sense for maintainers to spend more time reviewing AI generated code than contributors spent prompt engineering it.

PR #2458, opened 2025-12-06
Update Wasm 2.0 spec tests.

ncruces:
I updated the Wasm 2.0 spectests to the `wg-2.0` tag, and fixed a couple of simple things. I'm still stuck on a few others.

Specifically this one is failing (compiler crashes instead, but it's the same test, so probably the same issue):
```
--- FAIL: TestCompiler (2.42s)
    --- FAIL: TestCompiler/table_grow.wast (0.00s)
        --- FAIL: TestCompiler/table_grow.wast/module/line:117 (0.00s)
            require.go:331: expected no error, but was import table[grown-table.table]: minimum size mismatch: 2 > 1: table_grow.wast:117 module
                /usr/local/google/home/cruces/Documents/Personal/wazero/internal/integration_test/spectest/spectest.go:400
```

It's this wast:
```wast
(module $Tgt
  (table (export "table") 1 funcref) ;; initial size is 1
  (func (export "grow") (result i32) (table.grow (ref.null func) (i32.const 1)))
)
(register "grown-table" $Tgt)
(assert_return (invoke $Tgt "grow") (i32.const 1)) ;; now size is 2
(module $Tgit1
  ;; imported table limits should match, because external table size is 2 now
  (table (export "table") (import "grown-table" "table") 2 funcref)
  (func (export "grow") (result i32) (table.grow (ref.null func) (i32.const 1)))
)
(register "grown-imported-table" $Tgit1)
(assert_return (invoke $Tgit1 "grow") (i32.const 2)) ;; now size is 3
(module $Tgit2
  ;; imported table limits should match, because external table size is 3 now
  (import "grown-imported-table" "table" (table 3 funcref))
  (func (export "size") (result i32) (table.size))
)
(assert_return (invoke $Tgit2 "size") (i32.const 3))
```

Apparently, registering the `grown-table` module registers it but doesn't instantiate it? Idk.

@evacchi any ideas?

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -32,8 +32,8 @@ jobs:
     steps:
       - name: Install latest wast2json
         run: |  # Needed for build.spectest. wabt includes wast2json.
-          wabt_version=1.0.34
-          wabt_url=https://github.com/WebAssembly/wabt/releases/download/${wabt_version}/wabt-${wabt_version}-ubuntu.tar.gz
+          wabt_version=1.0.39
+          wabt_url=https://github.com/WebAssembly/wabt/releases/download/${wabt_version}/wabt-${wabt_version}-linux-x64.tar.gz
           curl -sSL ${wabt_url} | tar --strip-components 2 -C /usr/local/bin -xzf - wabt-${wabt_version}/bin/wast2json
 
       - uses: actions/checkout@v4


evacchi:
Do you mean that our runner is not instantiating it or that we're not supposed to instantiate it? I should check BTW

ncruces:
I mean that, debugging this, our runner is “registering” the module, but not “instantiating” it? That's what I could understand before life called. 🫣

ncruces:
I also updated the other spec tests and wabt.

PR #2464, opened 2026-01-10
Fix error when using host modules with ImportResolver.

mpoindexter:
When attempting to supply a host module created via HostModuleBuilder.Instantiate(), a panic occurred since hostModuleInstance could not be cast to *ModuleInstance.

This PR adds a method that can be used to test if the module supplied by the import resolver is a wrapper module, and if so, unwraps it before casting to *ModuleInstance.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -352,6 +352,10 @@ func (h hostModuleInstance) ExportedFunction(name string) api.Function {
 	panic("calling ExportedFunction is forbidden on host modules. See the note on ExportedFunction interface")
 }
 
+func (h hostModuleInstance) WrappedModule() api.Module {
+	return h.Module
+}
+
 // Instantiate implements HostModuleBuilder.Instantiate
 func (b *hostModuleBuilder) Instantiate(ctx context.Context) (api.Module, error) {
 	if compiled, err := b.Compile(ctx); err != nil {
--- a/experimental/importresolver_test.go
+++ b/experimental/importresolver_test.go
@@ -24,7 +24,7 @@ func TestImportResolver(t *testing.T) {
 		start := func(ctx context.Context) {
 			callCount++
 		}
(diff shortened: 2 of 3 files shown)


ncruces:
I don't know this bit very well, so I'll let other maintainers comment on this.

I'll say that _if we don't want to add an additional exported method_ (albeit to an unexported type), we could instead delete `hostModuleInstance` and do:
```go
func (m *ModuleInstance) ExportedFunction(name string) api.Function {
	if m.Source.IsHostModule {
		return nil
	}
	// ...
}
```

Then cleanup `hostModuleBuilder.Instantiate` and `runtime.Module`; and fix `TestNewHostModuleBuilder_Instantiate`  to not expect a panic.

anuraaga:
If the approach to remove the type entirely works that seems worth doing. Otherwise, I don't know how common it is in practice, but I think we can add an empty struct parameter defined anywhere under `internal` to keep the method effectively unusable outside of wazero.

mpoindexter:
I like what @ncruces suggested - I wasn't sure the exact motivation or history of the wrapper struct approach, was just trying to clean up a sharp edge I hit. I'll try to update this PR with that approach and see if there's any trouble with it.

mpoindexter:
I investigated this further, and the approach proposed by @ncruces won't work I think. https://github.com/wazero/wazero/blob/main/runtime.go#L358 calls `ExportedFunction` on `*ModuleInstance`, so that would break if we added the proposed check. That said, I think with https://github.com/wazero/wazero/pull/2475 using the exported function on a host module actually will no longer fail so maybe just removing this check entirely would make sense?

mpoindexter:
I think this approach is wrong given recent changes. I'd propose that the check is simply removed with the addition of https://github.com/wazero/wazero/pull/2475, but that's more of an api surface question for the maintainers of this project. In any case, I have been able to work around this, and it's more of an annoyance than anything, so closing this PR.

PR #2471, opened 2026-02-09
Don't use compiler when executable mmap is disabled 

### LINKED ISSUE #2470: Automatically select interpreter at runtime if codegen execution is not allowed

Some systems restrict executable mmap for security reasons. While we have build-time checks for platform support, I don't think we have a runtime check which could be useful for these situations.

Check if executable mmap works before selecting the compiler and select interpreter instead.

- An environment variable to select compiler or interpreter. Having to change code to select for corner cases has been a persistent pain point. This could be useful with or without the above

- A build flag to select compiler or interpreter. At least a rebuild without code change is easier, but it is still friction when needing to fallback for a corner case. That being said, the above env var would either itself need to be guarded carefully or otherwise prevents any future work on allowing excluding the compiler codebase from executables we've briefly discussed 

**Additional context**

https://github.com/wasilibs/go-re2/issues/223

anuraaga:
Fixes #2470

The compiler may be selected based on platform and CPU restrictions but still fail when trying to actually execute generated code, which can be disabled in security-sensitive environments using seccomp. As `NewRuntimeConfig` is meant to detect compiler support based on the environment, I think it's appropriate to check this too, as usual a user that really needs compiler can / probably should use `NewRuntimeConfigCompiler` which will error out as expected.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -105,7 +105,7 @@ jobs:
         run: bash <(curl -s https://codecov.io/bash)
 
   test_scratch:
-    name: ${{ matrix.platform.arch }}, Linux (scratch), Go-${{ matrix.go-version }}
+    name: ${{ matrix.platform.arch }}, Linux (scratch), Go-${{ matrix.go-version }}${{ matrix.platform.no_prot_exec && ', no-prot-exec' }}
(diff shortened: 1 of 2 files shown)


mathetake (approved):

danslo:
Thanks for this @anuraaga , much appreciated 🙏

PR #2476, opened 2026-02-16
sysfs: support Poll() on custom stdin implementations

### LINKED ISSUE #1500: support non-blocking files

paralin:
When using WithStdin() with a custom io.Reader, allow it to implement Poll() for proper async I/O in poll_oneoff. Previously, only os.Stdin (via NewStdioFile) had Poll support.

Add Pollable and PollableFile interfaces to experimental/sys so users can implement polling on their custom readers and filesystems. PollableFile extends File with Poll, IsNonblock, and SetNonblock without modifying the existing File interface, avoiding breakage for downstream File implementors.

As discussed in #2462, the current poll_oneoff implementation has special handling for stdin but doesn't allow custom stdin implementations to participate in polling. This prevents proper async I/O when using custom readers (e.g., virtual terminals, networked I/O).

With these changes, users can now:

- Pass a custom io.Reader to WithStdin() implementing experimental/sys.Pollable
- The poll_oneoff WASI function will call this implementation
- Implement experimental/sys.PollableFile for full file poll support

Changes:

- Add Pollable interface and Pflag type to experimental/sys
- Add PollableFile interface to experimental/sys (File + Pollable + nonblock methods)
- Make internal/fsapi.File a type alias for experimental/sys.PollableFile
- Add Pollable check in StdinFile.Poll() and fsFile.Poll()
- Check Pollable before POLLIN flag so impl. can handle non-POLLIN flags
- Re-export Pflag types from internal/fsapi via type aliases to experimental/sys canonical definitions
- Add unit tests for Poll using in-memory files

Fixes #1500


ncruces:
@paralin, can you look into the failures?

paralin:
@ncruces got it - sorry, forgot to commit the test data file. But actually, I can change it to use an in-memory file anyway, which is cleaner. Will do & update the PR shortly

paralin:
according to https://github.com/wazero/wazero/pull/2462#discussion_r2811156534 I will split File into File, PollableFile since we don't want to break downstream users of File, regardless of if its in experimental or not.

paralin:
all done and review-ready @ncruces  @evacchi

internal/fsapi/file.go
```diff
@@ -2,68 +2,10 @@ package fsapi
...
-	//     See https://pubs.opengroup.org/onlinepubs/9699919799/functions/poll.html
-	//   - No-op files, such as those which read from /dev/null, should return
-	//     immediately true, as data will never become available.
-	//   - See /RATIONALE.md for detailed notes including impact of blocking.
-	Poll(flag Pflag, timeoutMillis int32) (ready bool, errno experimentalsys.Errno)
-}
+// This alias exists so that internal code can continue to reference
+// fsapi.File without change, while the canonical definition lives in
+// the experimental/sys package for external implementors.
+type File = experimentalsys.PollableFile
```
evacchi: huh that's not exactly what I meant, I would have broken the two interfaces/make one embed the other. @ncruces does this break source compatibility?

paralin: Within the exported (public) interface there is File, Pollable, and PollableFile, this is just an internal alias. Have a look at how I set up those - was that what you were thinking of?

paralin: I took another look at this and I think I know what you mean. Well, it's ok as-is because there's already type assertions internally on Pollable. I was trying to keep this changeset small - but, for it to be maximally backwards-compatible and a proper full-feature fix for this topic, here's what I propose (and will put a second commit):

internal/fsapi exports only:

  - File
  - Pflag, POLLIN, POLLOUT
  - Adapt() = wraps a base File into PollableFile by adding stub Poll/IsNonblock/SetNonblock

We can replace all of this with references directly to the experimentalsys File and PollableFile exports.

Using type assertions we can check if a File supports Pollable and fallback to the old behavior if not, here are all the locations that use Pollable:

  - internal/sys/fs.go:335 -- SetNonblock on accepted socket
  - internal/sysfs/osfile.go:195,258 -- IsNonblock check
  - internal/sysfs/sock.go:107,126 -- IsNonblock check
  - WASI poll_oneoff host function -- Poll call

I'll get rid of the internal/fsapi aliases and use experimentalsys.File as the base type everywhere, then type-switch on PollableFile.

evacchi: yeah, that should work 🤔

paralin: @evacchi OK done!

imports/wasi_snapshot_preview1/fs.go
```diff
@@ -311,9 +311,14 @@ func fdFdstatSetFlagsFn(_ context.Context, mod api.Module, params []uint64) expe
...
 	} else {
 		nonblock := wasip1.FD_NONBLOCK&wasiFlag != 0
-		errno := f.File.SetNonblock(nonblock)
-		if errno != 0 {
-			return errno
+		if pf, ok := f.File.(experimentalsys.PollableFile); ok {
+			if errno := pf.SetNonblock(nonblock); errno != 0 {
+				return errno
+			}
+		} else if isDir, errno := f.File.IsDir(); errno == 0 && isDir {
```
paralin, line 318: I should mention why this change was necessary -

In the old code, `f.File` was always `PollableFile` (via `Adapt()`), so `SetNonblock` was called unconditionally on every file, even with `nonblock=false`. For `lazyDir`, `SetNonblock` always returned `EISDIR` regardless of the argument. This acted as a gatekeeper that blocked any flag modification on directories.

In the new code, `lazyDir` doesn't implement `PollableFile`, so the type assertion falls into the `else` branch. Without the `IsDir` check:

- `FD_NONBLOCK` on a directory: hits `else if nonblock` and returns `ENOSYS` instead of `EISDIR`
- `FD_APPEND` on a directory: `nonblock=false`, so it falls through entirely, reaches the `Stat()` check, directory isn't regular, returns 0 (success) instead of `EISDIR`

The `IsDir` check restores what `lazyDir.SetNonblock()` used to provide implicitly.

paralin:
Gentle nudge @ncruces @evacchi - I'm happy to spend as much time as needed adjusting this for you - this is the last thing remaining before I can drop my wazero fork and just use upstream, so would love to work with you to finalize it soon. Thanks!

internal/fsapi/unimplemented.go
evacchi, line 1: ok I was worried about all these deletions for a minute, but all of this was `internal` anyway; it took me also a minute to recall how this worked at the beginning (`unimplementedFile` was essentially the base of the type hierarchy for file implementations) and the change makes it more explicit

evacchi (approved):
I think this is ok, maybe if @ncruces or @anuraaga want to take a look to make sure this is good idiomatic Go; other than that I think I'm fine with it.

internal/fsapi/file.go
paralin: @evacchi all done :)

ncruces (approved):

PR #2477, opened 2026-02-21
examples: add pause/resume host function pattern for interpreter embedding

trevorprater:
## Summary

- Adds a new `examples/python-interpreter/` example demonstrating the **pause/resume host function pattern** used when embedding interpreters (e.g. Python, Lua) in WebAssembly via wazero
- Shows host-side orchestration of compile -> start -> resume loop with JSON-based data exchange through WASM linear memory
- Includes memory management patterns (alloc/dealloc) and status-code-driven control flow

## Background

When embedding an interpreter compiled to WASM, the guest module needs a way to call back into the host for I/O and application-specific functions. This example documents the cooperative execution pattern where:

1. WASM execution pauses by returning a status code
2. The host reads a JSON request from shared memory
3. The host dispatches the function call and writes the result back
4. The host resumes WASM execution

This pattern is used in production by projects like [monty-go](https://github.com/fugue-labs/monty-go) (Python-in-WASM) and is a common question in the wazero community.

## Test plan

- [x] `go build ./examples/python-interpreter/` compiles successfully
- [x] `go run ./examples/python-interpreter/` produces expected output matching README
- [x] README accurately describes the pattern and output

### DIFF
--- a/examples/python-interpreter/README.md
+++ b/examples/python-interpreter/README.md
@@ -109,7 +109,7 @@ The guest returns numeric status codes to indicate why execution paused:
 ### Real-World Usage
 
 This pattern is used by projects like
-[monty-go](https://github.com/aspect-build/monty-go), which embeds a Python
+[monty-go](https://github.com/fugue-labs/monty-go), which embeds a Python
 interpreter (RustPython compiled to WASM) in Go applications. The embedded
 Python can call Go-defined functions as if they were native Python functions,
 with the pause/resume loop handling the cross-boundary communication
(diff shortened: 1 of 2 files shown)


ncruces:
Maybe I'm too deep into Wasm, but I don't see this as a big insight?

This is just a regular host call: the guest calls the host and is obviously paused while waiting for the host to answer.

And Wasm host call is just like a system call: when you want to read from a socket you call into your operating system and your "app" (or at least that "thread") is "suspended" waiting for the answer.

I find calling this a pause/resume loop actually confusing.

PR #2478, opened 2026-02-26
Skip debug error handling for intentional errors

jerbob92:
This is a proposed change to skip error handling for intentional errors, such as the ThrowLongjmpError from the Emscripten code. This allows this code to be more efficient since it does not have to do debug collection, like looking up method names/parameters and DWARF information.

I'm not super happy with the implementation but maybe it's a good start for a better implementation.

Related Slack thread: https://gophers.slack.com/archives/C040AKTNTE0/p1772033734736119

### DIFF
--- a/sys/error.go
+++ b/sys/error.go
@@ -4,6 +4,8 @@ package sys
 import (
 	"context"
 	"fmt"
+
+	"github.com/tetratelabs/wazero/internal/internalapi"
 )
 
 // These two special exit codes are reserved by wazero for context Cancel and Timeout integrations.
@@ -80,4 +82,12 @@ func (e *ExitError) Is(err error) bool {
 		return true
 	}
 	return false
+}
+
+// IntentionalError is an error interface that can be used by intentional
+// internal errors that should skip default error handling.
+type IntentionalError interface {
+	internalapi.WazeroOnly
+	error
+	IsIntentional() bool
 }
--- a/sys/error.go
+++ b/sys/error.go
@@ -4,8 +4,6 @@ package sys
 import (
 	"context"
 	"fmt"
-
-	"github.com/tetratelabs/wazero/internal/internalapi"
 )
 
 // These two special exit codes are reserved by wazero for context Cancel and Timeout integrations.
@@ -84,10 +82,15 @@ func (e *ExitError) Is(err error) bool {
 	return false
 }
 
+func (e *ExitError) IsIntentional() bool {
+	// ExitError is always intentional and should not collect debug/DWARF
+	// information, because that information is never used anyway.
+	return true
+}
+
 // IntentionalError is an error interface that can be used by intentional
 // internal errors that should skip default error handling.
 type IntentionalError interface {
-	internalapi.WazeroOnly
 	error
 	IsIntentional() bool
 }
--- a/internal/emscripten/emscripten.go
+++ b/internal/emscripten/emscripten.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 
 	"github.com/tetratelabs/wazero/api"
+	"github.com/tetratelabs/wazero/internal/internalapi"
 	"github.com/tetratelabs/wazero/internal/wasm"
 	"github.com/tetratelabs/wazero/sys"
 )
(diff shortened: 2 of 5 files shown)


karelbilek:
I can confirm the code now works properly and does not "serialize" and does not wait on all program exits and long jumps in the same wazero runtime, when instantiating multiple modules.

I originally wanted to fix it by using sync.Map on the dwarf lines, but that only fixes the "serialization"; it would still compute the data that are not needed.

ncruces (approved):
This looks fine. I'm not sure about the term "intentional". I might prefer "sentinel" but that's overloaded? Idk.

sys/error.go
```diff
@@ -81,3 +81,16 @@ func (e *ExitError) Is(err error) bool {
...
+	// ExitError is always intentional and should not collect debug/DWARF
+	// information, because that information is never used anyway.
+	return true
+}
+
+// IntentionalError is an error interface that can be used by intentional
+// internal errors that should skip default error handling.
+type IntentionalError interface {
+	error
+	IsIntentional() bool
```
anuraaga, line 95: Can we add a key parameter in the internal package to this to keep it an implementation detail? I'm not sure it's actually a general concept yet

anuraaga (changes requested):
I'm a bit unsure what an intentional or even sentinel error is conceptually, in English terms I would expect both to have debug info. I guess this is a "control flow error", something forbidden by most Java textbooks? ;) If were honest about that it's fine, but I gave a suggestion to make sure we don't generalize this too much.

As long as it's implementation detail, I'm fine with landing it. But IMO emscripten is a hack and wazero supports it best effort - I am generally totally fine with performance being bad with it since people should use a WASI toolchain with wazero. If that means implementing exception handling, then we need to get to that, but we should make sure any emscripten workarounds can be deleted in some future.

karelbilek:
As I wrote; if this isn't a recommended way, putting sync.Map instead of
the mutexes works fine too.

anuraaga:
My comment is just that emscripten should be avoided as much as possible, not about such implementation details.

sys/error.go
```diff
@@ -81,3 +81,16 @@ func (e *ExitError) Is(err error) bool {
 	}
 	return false
 }
+
+func (e *ExitError) IsIntentional() bool {
```
anuraaga, line 85: By the way, if exit error is adding unnecessary debug overhead I think we should recognize the type directly without relying on an interface. It's a wazero detail.

jerbob92: @anuraaga it already does that in `stacktrace.FromRecovered`, reason I added the IsIntentional to it is so that we can re-use the same logic for both.

jerbob92:
> I'm a bit unsure what an intentional or even sentinel error is conceptually, in English terms I would expect both to have debug info. I guess this is a "control flow error", something forbidden by most Java textbooks? ;) If were honest about that it's fine, but I gave a suggestion to make sure we don't generalize this too much.

Yeah control flow error would be a good description. But that's basically what ExitError is as well. For ExitError the debug info is collected but never added anyway (check `stackTrace.FromRecovered`).

> As long as it's implementation detail, I'm fine with landing it. But IMO emscripten is a hack and wazero supports it best effort - I am generally totally fine with performance being bad with it

Sure, I'm also fine with that, but this is just something that was missed in the implementation of Emscripten and is quite an easy fix. I don't know why we would leave this in deliberately when we know how to fix it.

>  since people should use a WASI toolchain with wazero. If that means implementing exception handling, then we need to get to that, but we should make sure any emscripten workarounds can be deleted in some future.

It's basically the only reason I'm using Emscripten (and it does make compilation quite easy in some cases), if I could use WASI toolchains with Wazero I would do it, but everything I compiled needs setjmp support.

anuraaga:
> It's basically the only reason I'm using Emscripten

Yeah - so not committing to good experience with emscripten is a carrot I am going for to fix that ;)

In the meantime I gave the idea to make sure this doesn't leak into public ABI so as long as that's covered it's fine. Just was also trying to sync expectations, emscripten is hacky and exists for browsers, standard runtimes like wasmtime don't support it, and wazero probably should delete it eventually. It's cool that we have it as a stopgap but we shouldn't overcommit to supporting it.

ncruces:
Trying to unstuck this.

Is this an improvement for `ExitError` too? If so, we should definitely do _at least_ that part.

For `ThrowLongjmpErrorType`, can we do this without additional public API? Adding an interface may be the most Go way to do it, but it also leaks a promise which we may want to walk back on. I think that's @anuraaga's objection.

Unfortunately these are questions I can't answer without checking out the code (they may involve adding circular deps or something else that might stop me).

jerbob92:
> Trying to unstuck this.
> 
> Is this an improvement for `ExitError` too? If so, we should definitely do _at least_ that part.

Yeah it is.

> For `ThrowLongjmpErrorType`, can we do this without additional public API? Adding an interface may be the most Go way to do it, but it also leaks a promise which we may want to walk back on. I think that's @anuraaga's objection.
> 
> Unfortunately these are questions I can't answer without checking out the code (they may involve adding circular deps or something else that might stop me).

I'm fine with putting the IntentionalError interface in some internal package. I believe I had it before but there was some import cycle.

karelbilek:
why was this closed?

jerbob92:
That was actually me by accident, I was working on something else and reset my main to the wazero main, not realizing I used it as base for this PR. I'm now trying to get it back but I can't seem to re-open this PR. The commits are still here: https://github.com/jerbob92/wazero/tree/feature/skip-debug-error-handling-for-intentional-errors 

But I think they are less relevant for me now that I'm switching to native EH.

Issue #2480, opened 2026-03-04
poll_oneoff: multiplex blocking fd subscriptions


paralin:
This issue follows merging of PR #2481 

Currently `poll_oneoff` polls each blocking fd read subscription sequentially in a loop, tracking the remaining time budget so total wall time stays within the requested timeout.

While correct, this is suboptimal: a true multiplexed `poll(2)`-style call would wake on the first ready fd without waiting for each one. For example, if fd 3 has data but fd 4 does not, the current code must poll fd 4 first (consuming timeout) before discovering fd 3 is ready, depending on iteration order.

The fix would be to collect all pollable fds and issue a single multiplexed poll syscall. 

This is non-trivial as there's no way to simultaneously call Poll() on multiple fds at the same time.

PR #2481, opened 2026-03-04
poll_oneoff: generalize blocking fd polling for non-stdin fds

### LINKED ISSUE #1500: support non-blocking files

paralin:
poll_oneoff only polled stdin for blocking fd read subscriptions, silently dropping non-stdin pollable fds. Generalize the deferred polling to poll each blocking fd individually, tracking remaining time budget across iterations so total wall time never exceeds the requested timeout.

Changes:
- Generalize poll_oneoff blocking fd handling: track each deferred subscription's file and poll it individually instead of only stdin
- Track elapsed time per poll so N blocking fds complete within 1x timeout
- Add test for poll_oneoff with non-stdin pollable fd

Fixes: #1500

### DIFF
--- a/imports/wasi_snapshot_preview1/poll.go
+++ b/imports/wasi_snapshot_preview1/poll.go
@@ -83,8 +83,11 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno 
 
 	// Extract FS context, used in the body of the for loop for FS access.
 	fsc := mod.(*wasm.ModuleInstance).Sys.FS()
-	// Slice of events that are processed out of the loop (blocking stdin subscribers).
-	var blockingStdinSubs []*event
+	// blockingPollSubs are fd read subscriptions processed after the loop via polling.
+	var blockingPollSubs []struct {
+		evt  *event
+		file sys.File
+	}
 	// The timeout is initialized at max Duration, the loop will find the minimum.
 	var timeout time.Duration = 1<<63 - 1
 	// Count of all the subscriptions that have been already written back to outBuf.
@@ -135,9 +138,12 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno 
 				writeEvent(outBuf[outOffset:], evt)
 				nevents++
 			} else {
-				// if the fd is Stdin, and it is in blocking mode,
-				// do not ack yet, append to a slice for delayed evaluation.
-				blockingStdinSubs = append(blockingStdinSubs, evt)
+				// The fd is in blocking mode; do not ack yet, append
+				// to a slice for deferred polling evaluation.
+				blockingPollSubs = append(blockingPollSubs, struct {
+					evt  *event
+					file sys.File
+				}{evt, file.File})
 			}
 		case wasip1.EventTypeFdWrite:
 			fd := int32(le.Uint32(argBuf))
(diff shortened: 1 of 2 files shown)


paralin:
Filed #2480 as a follow-up issue on investigating calling Poll on multiple fds simultaneously.

ncruces (commented):
I will defer this one to @evacchi.

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -168,24 +174,29 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno
...
-	// Wait for the timeout to expire, or for some data to become available on Stdin.
-	if p, ok := stdin.File.(sys.Pollable); ok {
-		if stdinReady, errno := p.Poll(sys.POLLIN, int32(timeout.Milliseconds())); errno != 0 {
+	// Wait for the timeout to expire, or for data to become available on
+	// each blocking fd subscriber. The remaining time budget is tracked
+	// across iterations so total wall time never exceeds the timeout.
+	remaining := timeout
+	for _, sub := range blockingPollSubs {
+		p, ok := sub.file.(sys.Pollable)
+		if !ok {
```
evacchi, line 183: I am not sure this is intended, but if a blocking fd doesn't implement Pollable, its subscription is skipped entirely. It was the same for stdin in the old code, but now it applies to all blocking fds; maybe we should return ErrnoNotsup or least ErrnoSuccess with ready=false in these cases 🤔

imports/wasi_snapshot_preview1/poll.go
```diff
@@ -168,24 +174,29 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno
...
+		if errno != 0 {
 			return errno
-		} else if stdinReady {
-			// stdin has data ready to for reading, write back all the events
-			for i := range blockingStdinSubs {
-				evt := blockingStdinSubs[i]
-				evt.errno = 0
-				writeEvent(outBuf[nevents*32:], evt)
-				nevents++
-			}
```
evacchi, line 188: so this was obviously sort of a hack, but stdin is indeed treated specially because you want blocking in the interactive case. 

It is worth noticing that while here we were dispatching to all stdin subscribers, now we are doing it in the loop; that means that, in all cases, if two subs reference the same fd, the fd gets polled twice (with the second fd obviously getting less time, potentially 0ms).

not a bug, just mentioning

imports/wasi_snapshot_preview1/poll_test.go
```diff
@@ -537,6 +537,60 @@ func Test_pollOneoff_Zero(t *testing.T) {
 	require.Equal(t, uint32(1), nevents)
 }
 
+// Test_pollOneoff_NonStdinPollable verifies that poll_oneoff correctly polls
+// non-stdin fds that implement Pollable (e.g. custom fs.FS mounts).
+func Test_pollOneoff_NonStdinPollable(t *testing.T) {
```
evacchi, line 542: we might want to add the following cases:

- Timeout: e.g., two blocking fds where the first consumes most of the timeout; the second gets a short/zero poll
- verify behavior when some fd doesn't implement Pollable
- verify what happens if Poll returns ready=false

evacchi (changes requested):
a few notes and a smaller request, but overall looks good!
... and sorry about the delay!

paralin:
Relevant work in tinygo: https://github.com/tinygo-org/tinygo/commit/d01a932201bb4059eb6c84f1ee4e7e5e84f70040

evacchi:
I think we were just missing a few test cases mentioned above, otherwise IIRC this was good to go

imports/wasi_snapshot_preview1/poll_test.go
paralin: Done 👍🏽

imports/wasi_snapshot_preview1/poll.go
paralin: Done! Fixed 👍🏽

paralin:
@evacchi Just finished that - test cases added and I made sure all the review comments were addressed.

paralin:
The tests failed with a github error "Your account is suspended" when trying to checkout the repo.

Weird!

Edit: It's a major outage, apparently: https://www.githubstatus.com/incidents/gnftqj9htp0g

evacchi:
> Edit: It's a major outage, apparently: https://www.githubstatus.com/incidents/gnftqj9htp0g

surprisedpikachu.gif

evacchi (approved):
good to go when CI is green

ncruces (approved):
I think this run went fine: https://github.com/wazero/wazero/actions/runs/26447328131
But I couldn't re-trigger this one: https://github.com/wazero/wazero/actions/runs/26447328130

Maybe an empty commit re-triggers it? Otherwise, I vote lets merge it.

ncruces:
Oh this is absurd. I forgot to merge this one before the release.

Lets decide what to do about #2482 then I make a point release before the next big push, which should be GC.

Sorry @paralin.

Issue #2466, opened 2026-01-11
Very large performance impact of WithCloseOnContextDone(true)


zdandoh:
**Describe the bug**
First off, thank you for your work on this great library!

I have been using wazero for a project that I'm working on that requires running untrusted user code, I noticed that running wasm binaries with `WithCloseOnContextDone(true)` slows down the runtime performance of my C code by a factor of 10-20x, which was a surprising result. The docs state:

> Note that this comes with a bit of extra cost when enabled. The reason is that internally this forces the interpreter and compiler runtimes to insert the periodical checks on the conditions above. For that reason, this is disabled by default.

That makes sense, but the 10-20x performance hit seems quite extreme. Am I doing something wrong, or is the impact expected to be that large? I spent some time poking at the compiler internals and noticed that wazero calls back into Go for every native loop iteration when WithCloseOnContextDone is on. I looked at a couple ways to mitigate the performance impact and came up with a couple ideas:
- A native code counter that only calls into Go every X loop iterations. I implemented this, and when X=100 the performance hit was more like 60%, which seems like a reasonable perf/latency tradeoff.
- The current code opts to call into Go frequently to avoid deadlocks when the goroutine running WASM can't be preempted, could a more efficient native code approach be used when GOMAXPROCS > 1?

**To Reproduce**
Here is a repository with a minimal reproduction: https://github.com/zdandoh/wazero-perf-investigation
```
Ran test with contextEnabled=false, duration: 72.848667ms
Ran test with contextEnabled=true, duration: 1.266401583s
```
Note that this code is a bit of a contrived microbenchmark for simplicity reasons, the factor I was seeing in real-world code was closer to the 10x number than 20x.

**Expected behavior**
Less drastic slowdown when using `WithCloseOnContextDone(true)`

**Screenshots**
N/A

**Environment (please complete the relevant information):**
- Go version: [e.g. 1.19.1] 1.25.5
- wazero Version: 1.11.0
- Host architecture: arm64
- Runtime mode: compiler

**Additional context**
This test was run on an M4 macbook pro

mpoindexter:
An alternate idea to make this fast: work with the scheduler/runtime to make sure that when WASM code is executing other Go work can be scheduled.

[...]

PR #2482, opened 2026-03-07
Add WithInterruptCheckInterval to reduce WithCloseOnContextDone overhead

sbzimpart:
Addresses https://github.com/tetratelabs/wazero/issues/2466

Summary
- Adds experimental WithInterruptCheckInterval(interval uint64) to control how often the exit code check is performed in loop headers
- When WithCloseOnContextDone(true) is set, the compiler currently inserts a host call on every loop iteration, causing 10-20x slowdown in tight loop scenarios
- With this change, a counter-based bitmask approach checks only every Nth iteration (e.g., interval=32 reduces checks by ~32x)
- interval must be 0 (check every iteration, default) or a power of 2
- Context cancellation, timeout, and CloseWithExitCode still terminate loops


internal/engine/wazevo/engine.go
```diff
@@ -41,6 +41,9 @@ type (
 		sharedFunctions *sharedFunctions
 		// setFinalizer defaults to runtime.SetFinalizer, but overridable for tests.
 		setFinalizer func(obj interface{}, finalizer interface{})
+		// interruptCounter is the shared counter used for interrupt check interval optimization.
+		// All callEngine instances point to this counter to amortize context cancellation checks.
+		interruptCounter uint64
```
mathetake: this should be in call_engine. Engine is a global/shared construct

sbzimpart: done

internal/engine/wazevo/call_engine.go
```diff
@@ -90,6 +91,10 @@ type (
 		memoryWait64TrampolineAddress *byte
 		// memoryNotifyTrampolineAddress holds the address of the memory_notify trampoline function.
 		memoryNotifyTrampolineAddress *byte
+		// interruptCounter points to callEngine.interruptCounterValue, used for loop termination checking.
+		// When interruptCheckInterval is configured, the counter is incremented on each loop iteration
+		// and the exit code check is only performed when (counter & mask) == 0.
+		interruptCounter *uint64
```
mathetake: this is completely wrong and different than the existing pattern. Accessing a field on the call engine is via offset, not adding another guy here

internal/engine/wazevo/engine.go
```diff
@@ -41,7 +41,6 @@ type (
 		sharedFunctions *sharedFunctions
 		// setFinalizer defaults to runtime.SetFinalizer, but overridable for tests.
 		setFinalizer func(obj interface{}, finalizer interface{})
-
```
mathetake, line 44: irrelevant change

mathetake (changes requested):
can you remove the configuration knob in the root package and have it under experimental flag? we won't add things like this directly in the public stable API

sbzimpart:
>can you remove the configuration knob in the root package and have it under experimental flag? we won't add things like this directly in the public stable API

moved also addressed previous comments

jwriteclub:
After a long trail of debugging a strange issue I find myself at this PR. I would love to get this PR merged since it's also causing a significant slowdown on something I'm working on.

Is there anything I can do to help?

ncruces:
I'm sorry this fell through the cracks. I was under the impression changes had been requested but weren't yet addressed.

Meanwhile #2500 showed up, which IIUC is how other engines avoid runaway tasks, and is (strictly?) more powerful (also slower).

In Go, though, `Context` seems the obvious API for cancellation, so my gut tells me we should merge this one.

But I don't anticipate using either, so I'll leave up to other maintainers to comment. @mathetake are your issues addressed? @evacchi any say?

I don't mind fixing the conflicts myself if we're OK with merging this one.

evacchi:
I think at this point https://github.com/wazero/wazero/pull/2533 might show more promise

PR #2485, opened 2026-04-09
Added f4 to community/users.md

### LINKED ISSUE #2484: Project using you library
Hi, and thanks for the lib! It's great :)

If you plan to start maintaining the list of projects using your library, here is the one: [f4 file manager](https://github.com/unxed/f4), Far Manager clone in pure go. We use it in our plugin system to allow writing cross platform plugins and also allow future compatibility with Far3/far2l/far2m C plugins.

Thanks again!

unxed:
Fixes #2484

### DIFF
--- a/site/content/community/users.md
+++ b/site/content/community/users.md
@@ -42,6 +42,12 @@ considering their efforts before starting your own!
 |:-----------------------|----------------------------------------------------|
 | [yoke][47] | A WebAssembly-based package deployer for Kubernetes, enabling declarative and programmable deployments with Wasm |
 
+### Applications
+
+| Name               | Description                                   |
+|:-------------------|-----------------------------------------------|
+| [f4][50] | Far Manager clone in Go using wazero for plug-in system |
+
 ### Middleware
 
 | Name                   | Description                                        |
@@ -187,3 +193,5 @@ experience.
 [48]: https://github.com/klippa-app/go-libtiff
 
 [49]: https://libtiff.gitlab.io/libtiff/
+
+[50]: https://github.com/unxed/f4


mathetake (approved):
👍

unxed:
Oh, I am so sorry, we had to switch to another architecture. Pls revert

unxed:
You might find this amusing, but I actually do use wazero in this project. Just not for the purpose you might expect.

A WASM sandbox is too restrictive for plugins, so I solved that problem differently. Later, though, I needed to port the syntax highlighting engine and plugin from Far Manager that I've been using for years. That's where wazero turned out to be invaluable: I wrote a minimal C wrapper around the plugin, compiled it to WASM, and voilà! I have to say, the performance is more than sufficient for this use case.

The only thing I couldn't get working was propagating exceptions. Maybe you could point me in the right direction? I've described my unsuccessful attempts in the README:
https://github.com/unxed/colorer4go

I'll be opening a new PR soon with updated usage information. In the meantime, thanks—wazero is a fantastic piece of software, and I've really enjoyed working with it.

unxed:
Oops, new update: f4 now actually CAN use wasm plugins via wazero. . We can reopen this PR.

It might seem strange that I changed my mind so many times, but the other day I finally figured out a way to pass calls through to wazero FFI (using a purego-like solution based on goffi and a custom [broker](https://github.com/unxed/ffibridge)) + we anyway needed wazero for colorer4go, and that’s what tipped the scales in favor of wazero.

---

Also:

> I wrote a minimal C wrapper around the plugin, compiled it to WASM

See #2519

evacchi:
I think you'll have to open another PR :p

In the interest of completeness, if you only needed wazero for FFI a valid alternative is @ncruces's AoT translator https://github.com/ncruces/wasm2go

of course if you'd like to use wazero for plugins, then it makes sense to also use it for FFI!

unxed:
:) I already use wasm2go for another goal. To support torrentzip format I needed 1:1 zlib-compatible stream. go zip libraries can not do this, and adoption looked very time-consuming. I tried wazero for that, but overhead of 1.2~1.5 was too large for archiver. So I took wasm2go and optimized result via recent LLM. And got result even faster than C version! https://github.com/unxed/zlib4go
Another two reasons are 1) zlib has almost no updates and if one is released I can repeat all steps in an hour via "pls look at this commit and apply the same optimizations to this go code" and 2) zlib is small and feats in LLM context perfectly.

As for the colorer Far plugin, it is not a thing that requires cosmic speeds. It works in f4 even faster than in C/C++ far2l due to seriously optimized editor of f4. Also, colorer has updates from time to time, and this flow does not require LLM-optimize code generated by wasm2go for each release. So I finally used wazero for this and it solved the problem
perfectly!

And I thought: if we anyway have wazero on board, and its speed is enough for tasks like syntax highlighting, why not to try implementing wasm-based plugins support for the second time? I thought about this for months, but can not find the solution for native libraries calls (like original Far plugins can do). And finally I found the solution by translated go ffi API (purego-like) to wasm so any host api can be used (and of course I had to implement security model for that).
https://github.com/unxed/ffibridge

PR #2488, opened 2026-04-15
interpreter: ensure i32 operations are masked

evacchi:
while testing the Exception Handling implementation against https://github.com/klippa-app/go-pdfium we discovered an unrelated, latent issue in the interpreter. 

go-pdfium passes through unsafe pointer reads like `*(*uint64)(unsafe.Pointer(&uint32Var)))`. 

Normally, in the compiler (the default on arm64/amd64) this is irrelevant, because the native instructions truncate/mask/ignore the upper "garbage" 32 bits; however, the interpreter does not truncate such values, causing execution to fail.

This PR aligns the behavior to the compiler; notice that we are now casting when we pop operands, so we are adding a slight overhead in the 32-bit cases (although this being the interpreter, we should not care much); an alternative could be addressing this at the boundary, sanitizing host input parameters.


evacchi:
FYI @jerbob92 thanks for the help with this

anuraaga (approved):

PR #2489, opened 2026-04-18
feat: exception handling spec

evacchi:
Add experimental support to the Exception Handling spec.

Exception handling adds structured error propagation to WebAssembly through `tags` (typed exception signatures), `throw/throw_ref` (raising exceptions), and `try_table` (catching exceptions with typed `catch` clauses). A new value type, `exnref`, represents a reference to a caught exception.

Excluding tests and spec suites, the entire feature amounts to a little less than 2000 lines of code.

Feature flag: `experimental.CoreFeaturesExceptionHandling`

## What's the use for this?

In the spirit of always having a real use case for the specs we implement, I did check this implementation against the one large Wasm C++ codebase I am aware about, i.e. https://github.com/klippa-app/go-pdfium, and I verified that the entire suite, compiled with `emcc+wasm-opt` does work; incidentally, much of the existing support code, based on Emscripten's longjmp/setjmp, can be dropped (thx @jerbob92 for the help!)

Broadly speaking, I don't know how much EH-enabled Wasm code you will find in the wild. For instance, code compiled using Clang's C++ backend does not really throw new-style exceptions, unless you compile WASI-SDK with some flags, or you build with `emcc` and translate old-style exceptions using `wasm-opt` (see exceptions_test.go)

In short, this spec should be really seen as a step towards supporting the GC spec. The design decisions (esp. regarding the test suite) follow from this reasoning.

## Spec Suite

`wast2json` **does not** support the EH suite, but `wasm-tools` does, albeit with a tiny quirk in  the way negative numbers are rendered. In this PR _I am not_ moving the other suites to `wasm-tools` though.

The spec test suite passes entirely, except for two tests that are intentionally skipped.

These tests assume **a very small subset** of typed references is implemented; specifically, the distinction between nullable and non-nullable function references (`(ref null $t)` vs `(ref $t)`). I have decided to just flatten all non-nullable refs to nullable refs at decode time, so we cannot detect the type mismatch the spec expects in these cases.

Passing the **full suite is actually straightforward** because it only requires introducing a non-nullable variant for each ref type: strictly speaking the 2 failing tests are only for related nullable vs non-nullable **func refs** -- so I can special-case these and/or add non-nullable types for each known value; however, we will **need** to rethink the representation of Values for the GC proposal anyway (because this will introduce custom type indices, and these won't fit our currently byte-sized `ValueType` enum).

The subtype-checking functions (`isRefSubtypeOf`, `isStrictRefSubtypeOf`) are kept as abstraction points but currently they are just equalities.

Nevertheless, as proven by the CPP example and https://github.com/klippa-app/go-pdfium this PR provides a starting point that is already useful (even if it is for limited cases).

## Interpreter

Exception handling in the interpreter:

**Within a frame**, `try_table` blocks are compiled into a static exception table with PC ranges. When `throw` or `throw_ref` executes, `searchExceptionTable` scans backwards (inner handlers first) for a catch clause whose PC range covers the current instruction. If a match is found, `applyExceptionHandler` adjusts the operand stack depth and jumps to the catch target PC directly.

**Across frames**, if no handler is found in the current frame, `panic(&thrownException{...})` propagates up the Go call stack. Each cross-function call goes through `callWithUnwind()`, which uses `defer/recover` to intercept the panic. Its `canRestore` method unwinds `ce.frames` to the caller's depth and searches that frame's exception table. If a handler is found, `doRestore` applies it and `callWithUnwind` returns `true` (frame unwound -- caller refreshes locals and continues). If not, the panic re-propagates to the next outer `callWithUnwind`.

Note on the implementation: this mechanism is unified with the existing `snapshot/restore` API through the `restorable` interface: both `*thrownException` and `*snapshot` implement `canRestore`/`doRestore`, so `callWithUnwind` handles both with a single defer/recover path. The short-circuit check (`len(exceptionTable) == 0 && no snapshotter`) skips the defer/recover overhead entirely for frames that don't need it.

## Compiler

The compiler could not reuse the snapshot/restore mechanism directly, but it follows a similar pattern. One cool thing is that exception handling could be implemented entirely as an SSA-level lowering with new exit codes; there is no backend-specific code.

Throwing uses a two-phase ABI:

  1. `ExitCodeThrowAlloc`: exits to the Go runtime to allocate an `Exception` struct on the heap and writes a pointer to its params array into the execution context. Then we return control to compiled code, which stores the tag parameters directly into the struct.
  2. `ExitCodeThrow`: exits to the Go runtime again to search for a matching handler, restore the stack checkpoint, and branch to the catch target.

The idea is that we NEED to allocate the parameter slice dynamically, so we delegate to the Go runtime. Incidentally, this also allows us to unify `throw` and `throwRef`; in the latter case, we skip the `alloc` phase, and we exit straight to `ExitCodeThrow`.

### Trampolines

The compiler uses four trampolines to return to host to manage exception state:

| Trampoline | Exit Code | Purpose |
|---|---|---|
| **ThrowAlloc** | `ExitCodeThrowAlloc` | Allocates an `Exception` struct on the Go heap, writes `&exn.Params[0]` into `execCtx` so compiled code can store tag parameters directly. |
| **Throw** | `ExitCodeThrow` | Searches `tryHandlers` for a matching catch clause, restores the stack checkpoint on match, sets `caughtExceptionClauseIdx` for the compiled dispatch. Used for both `throw` (after ThrowAlloc) and `throw_ref` (directly). |
| **TryTableEnter** | `ExitCodeTryTableEnter` | Clones the current stack as a checkpoint, pushes a `tryHandler` with the catch clause table and saved module instance. Sets `caughtExceptionClauseIdx = -1` (no exception). |
| **TryTableLeave** | `ExitCodeTryTableLeave` | Pops the most recent `tryHandler`. Emitted at `try_table` block ends, and before any branch or return that exits a `try_table` scope. |

### Lowering

As mentioned earlier, this entirely resolved at the SSA-level. In particular:

- **Entry**: when a `try_table` is encountered, the compiler emits a call to the TryTableEnter trampoline. Control returns to compiled code, which reads `caughtExceptionClauseIdx` from the execution context — if it's -1 (no exception), execution continues into the `try_table` body.

- **Normal exit**: when execution reaches the `End` opcode of the `try_table` block, a TryTableLeave trampoline call is emitted.

However, every TryTableEnter must have a matching TryTableLeave on every control flow path that doesn't throw; so we also "TryTableLeaves" also for the following:

- **Early exit (branch)**: a `br`, `br_if`, or `br_table` can jump to a label outside the `try_table`, skipping its `End`. `emitTryTableLeaves(depth)` walks the control frame stack and emits one TryTableLeave for each `try_table` frame the branch crosses. For `br_if`, this happens in a trampoline basic block that only executes on the taken path.

- **Early exit (return)**: `return`, `return_call`, and `return_call_indirect` call `emitTryTableLeaves` with the full control stack depth, popping all active handlers before leaving the frame.

- **Throw**: on `throw`/`throw_ref`, compiled code exits to the dispatch loop via ExitCodeThrow. `doHandleException` searches `tryHandlers` innermost-to-outermost. On match, it restores the cloned stack checkpoint, sets `caughtExceptionClauseIdx`, and re-enters compiled code at the handler's return address. The compiled dispatch branches to the matching catch clause's handler block.

### Caveats

- As described above, we are skipping 2 spec tests (will be addressed when we implement typed references)
- Because of that, I am _not_ enabling by default the fuzzer: I would have to special-case for typed references and/or TODOs I have left behind; I did run the fuzzer with some special-casing while debugging the issue with pdfium (#2488) and did not find issues (the fuzzer did go frequently OOM though); in fact, the pdfium issue was actually unrelated (!)


anuraaga:
@evacchi Great to see progress on this! Didn't look at the code yet but figured I'd give it a try with setjmp

https://github.com/wasilibs/go-protoc-gen-builtins/compare/main...anuraaga:wazero-eh?expand=1

I get an invalid instruction for seemingly `i32.const`, though presumably that means the entire binary reader is getting misaligned or something. I'm a bit unsure on the caveats about EH listed out but I think this is supposed to be working, or is it related to the unimplemented bits?

```
❯ go test .
--- FAIL: TestBuf (11.72s)
    --- FAIL: TestBuf/gorun (5.42s)
        protoc_gen_builtins_test.go:90: failed to run buf: exit status 1
            2026/04/19 11:32:25 invalid function[1757]: invalid instruction 0x6
            exit status 1
            2026/04/19 11:32:27 invalid function[103]: invalid instruction 0x6
            2026/04/19 11:32:27 invalid function[1704]: invalid instruction 0x6
            Failure: plugin go run ../cmd/protoc-gen-upb: exit status 1
            plugin go run ../cmd/protoc-gen-cpp: signal: killed
            plugin go run ../cmd/protoc-gen-kotlin: signal: killed
            plugin go run ../cmd/protoc-gen-csharp: signal: killed
            plugin go run ../cmd/protoc-gen-php: signal: killed
            plugin go run ../cmd/protoc-gen-python: signal: killed
            plugin go run ../cmd/protoc-gen-ruby: signal: killed
            plugin go run ../cmd/protoc-gen-upbdefs: signal: killed
            plugin go run ../cmd/protoc-gen-objc: signal: killed
            plugin go run ../cmd/protoc-gen-upb_minitable: signal: killed
            exit status 1
    --- FAIL: TestBuf/installed (2.34s)
        protoc_gen_builtins_test.go:90: failed to run buf: exit status 1
            2026/04/19 11:32:29 invalid function[1757]: invalid instruction 0x6
            Failure: plugin protoc-gen-upb: exit status 1
            plugin protoc-gen-ruby: signal: killed
            plugin protoc-gen-python: signal: killed
            plugin protoc-gen-pyi: signal: killed
            plugin protoc-gen-upb_minitable: signal: killed
            plugin protoc-gen-csharp: signal: killed
            plugin protoc-gen-upbdefs: signal: killed
            plugin protoc-gen-cpp: signal: killed
            exit status 1
FAIL
FAIL    github.com/wasilibs/go-protoc-gen-builtins      11.898s
FAIL
```

evacchi:
interesting! I'll take a look! maybe I'm some missing `default` case handling?

ncruces:
Wild guess, @anuraaga, did you try with `CoreFeaturesExtendedConst`?
```sh
$ wasm-dis protoc-gen-cpp.wasm | tail
 ;; custom section ".debug_loc", size 560199
 ;; custom section ".debug_abbrev", size 100631
 ;; custom section ".debug_info", size 1471429
 ;; custom section ".debug_str", size 653358
 ;; custom section ".debug_line", size 152173
 ;; custom section ".debug_ranges", size 108472
 ;; custom section "producers", size 148
 ;; features section: threads, mutable-globals, nontrapping-float-to-int, bulk-memory, sign-ext, exception-handling, reference-types, multivalue, extended-const, bulk-memory-opt
)
```

anuraaga:
Yeah did give that a try, unless I messed something up it didn't affect the error.

evacchi:
ok, good news! as mentioned here (not really stressed though!)

> Broadly speaking, I don't know how much EH-enabled Wasm code you will find in the wild. For instance, code compiled using Clang's C++ backend does not really throw new-style exceptions, unless you compile WASI-SDK with some flags, or you build with emcc and translate old-style exceptions using wasm-opt (see exceptions_test.go)


the protoc wasm binaries use old-style EH (`try/catch/delegate`, opcodes `0x06/0x07/0x09`). This PR only implements the new EH proposal (`try_table` = `0x1f`)

you just need to post-process your Wasm binaries with `wasm-opt --translate-to-exnref`!

```
$ go test .
ok      github.com/wasilibs/go-protoc-gen-builtins      6.736s
```

Since this might be confusing, I am adding the old opcodes  as `OpcodeLegacyTry` etc... so we can print a helpful error instead.


```
$ go test . -count 1
--- FAIL: TestBuf (6.09s)
    --- FAIL: TestBuf/gorun (1.24s)
        protoc_gen_builtins_test.go:90: failed to run buf: exit status 1
            2026/04/19 13:39:38 invalid function[1757]: legacy exception handling instruction 0x6 not supported; recompile with wasm-opt --translate-to-exnref
            exit status 1
            2026/04/19 13:39:38 invalid function[1704]: legacy exception handling instruction 0x6 not supported; recompile with wasm-opt --translate-to-exnref
            2026/04/19 13:39:39 invalid function[103]: legacy exception handling instruction 0x6 not supported; recompile with wasm-opt --translate-to-exnref
```

internal/engine/wazevo/frontend/frontend.go
```diff
@@ -95,12 +110,70 @@ func NewFrontendCompiler(m *wasm.Module, ssaBuilder ssa.Builder, offset *wazevoa
...
+func (s *SharedCatchClauseTable) Append(clauses []wazevoapi.CatchClauseInstance) int {
+	s.mu.Lock()
+	id := len(s.table)
+	s.table = append(s.table, clauses)
+	s.mu.Unlock()
+	return id
+}
+
+func (s *SharedCatchClauseTable) Table() [][]wazevoapi.CatchClauseInstance {
+	return s.table
```
ncruces, line 161: From what I could understand, compilation workers race to add stuff to the table, then the result is collected at the end, which is why no mutex is needed here?

I'd like to somehow ensure this is how it's used. Maybe flip a boolean here, that causes `Append` to panic?

ncruces (approved):
It's hard to in good faith say I reviewed this, but I did go over most of the code, trying to catch stuff.

Found only a nit, but again, I honestly can't hold something this big in my head, so that's not worth much.

I guess I approve and assume we'll be here to pick up the pieces if something breaks.

Independently testing something else real was great @anuraaga!

internal/engine/wazevo/frontend/frontend.go
evacchi: see https://github.com/wazero/wazero/pull/2489/commits/26ee820557cfe2614445549591afe4aeb9a42707

anuraaga:
Thanks for finding that - I thought it was only problematic for C++ exceptions, but realized even normally needed the magic `-mllvm -wasm-use-legacy-eh=false`. Luckily that allows us to enjoy a `parse exception: bad node code 31` error during linking instead. Glad to be mostly out of Wasm, except for these moments when back in ;)

protoc ends up compiling setjmp/longjmp even though it's actually dead code. After getting other non-exception issues fixed when moving off of wasix, I could get pgquery to build and tests run fine, including the parse errors.

https://github.com/wasilibs/go-pgquery/compare/main...anuraaga:wazero-eh?expand=1

That being said, I must say the new EH do seem more suited to wazero, `try_table` fits the control flow of the snapshotting we had quite nicely. Just got that from a very quick skim, may be able to read the code more later but feel free to merge either way.

anuraaga (approved):

mathetake (approved):

mathetake:
maybe try enabling fuzzing?

evacchi:
> maybe try enabling fuzzing?

I did! However, as I mentioned in the Caveats section:

> * Because of that, I am _not_ enabling by default the fuzzer: I would have to special-case for typed references and/or TODOs I have left behind; I did run the fuzzer with some special-casing while debugging the issue with pdfium ([interpreter: ensure i32 operations are masked #2488](https://github.com/wazero/wazero/pull/2488)) and did not find issues (the fuzzer did go frequently OOM though); in fact, the pdfium issue was actually unrelated (!)

so basically, I have already typed references in the pipeline (rough branch), once that is done we should be able to enable fuzzing for good

PR #2492, opened 2026-04-29
arm64: fix relocation bug when number of args > 8

mpoindexter:
On ARM64, when there were more than 8 parameters, and conditional branch trampolines needed to be inserted in a method body, resolveAddressingMode would be called multiple times, resulting in a panic when an already resolved addressing mode was re-resolved.

Move the addressing mode resolution out of the conditional branch loop to resolve.


mpoindexter:
This should reproduce and resolve the previously reported bug https://github.com/wazero/wazero/issues/2401

achille-roussel (approved):
The change looks good to me!

Will leave it open for a bit in case @ncruces wants to take a look.

ncruces (approved):
LGTM. @evacchi worked on the rellocations and commented on the original bug, so maybe he wants to chime in.

evacchi (approved):
let's ship it 🛳️

PR #2494, opened 2026-05-11
feat: WebAssembly GC (wasm-gc) support behind experimental.CoreFeaturesGC

achille-roussel:
## Summary

Adds WebAssembly GC ("wasm-gc") support — the largest component of Wasm 3.0 — to the interpreter, gated behind `experimental.CoreFeaturesGC`.

What's covered:

- **Type system**: rec groups, sub / sub final, struct, array, packed i8/i16 storage; iso-recursive tying with a Cohen-style subtype display for O(1) `IsSubtype` lookups; canonical keys are rec-group-aware so two types that reduce to the same iso-recursive form get equal IDs even when declared in differently-shaped rec groups.
- **Heap representation** (`RATIONALE.md` documents the choice): piggyback on Go's GC. `WasmStruct` / `WasmArray` carry a `TypeID` first-field so runtime kind discrimination works without a parallel ref stack. `i31` uses a tagged-uintptr encoding; externref-as-anyref uses a separate tag bit.
- **Validator**: rich `(Nullable, HeapKind, TypeIdx)` info plumbed through `valueTypeStack`; `IsValueTypeSubtypeOf` over (byte, *rich) pairs handling nullability, abstract hierarchy, and concrete-by-TypeIdx; type-section structural checks reject `sub` of a `final` supertype, form mismatches, and field/param/result mismatches against the spec matching rule.
- **All 37 wasm-gc + typed-funcref instructions** in the interpreter: `ref.i31`, `i31.get_s/_u`, `ref.eq`, `ref.as_non_null`, `any.convert_extern`, `extern.convert_any`, `struct.new[_default]`, `struct.get[_s|_u]`, `struct.set`, `array.new[_default|_fixed|_data|_elem]`, `array.get[_s|_u]`, `array.set`, `array.len`, `array.fill`, `array.copy`, `array.init_data`, `array.init_elem`, `ref.test[_null]`, `ref.cast[_null]`, `br_on_cast`, `br_on_cast_fail`, `br_on_null`, `br_on_non_null`, `call_ref`, `return_call_ref`.
- **Import compatibility**: subtype-aware check using `Store.IsSubtype` in addition to byte-level `EqualsSignature`, so `sub` vs `sub final` and recursive-type import mismatches are caught.
- **Decoder finality**: plain `(func)` / `(struct)` / `(array)` shorthand correctly desugars to `(sub final () ...)` per spec.

## Wazevo

Out of scope for this PR. `wazevo.CompileModule` scans the type section up front and returns "wasm-gc modules ... are not supported by the optimizing compiler" if any entry has struct or array form. Modules using only GC instructions on funcref-only signatures still hit this guardrail by design — runtime selection should pick the interpreter for GC modules, or surface this error if the user requested the compiler.


experimental/table/lookup.go
```diff
@@ -22,7 +22,9 @@ func LookupFunction(
 	expectedParamTypes, expectedResultTypes []api.ValueType,
 ) api.Function {
 	m := module.(*wasm.ModuleInstance)
-	typ := &wasm.FunctionType{Params: wasm.FromApiValueType(expectedParamTypes), Results: wasm.FromApiValueType(expectedResultTypes)}
+	// Final mirrors the (sub final (func ...)) identity of binary-decoded
+	// shorthand func types so the looked-up FunctionTypeID matches.
+	typ := &wasm.FunctionType{Params: wasm.FromApiValueType(expectedParamTypes), Results: wasm.FromApiValueType(expectedResultTypes), Final: true}
```
evacchi: first thing that "pops" is this new field has to be set explicitly everywhere to maintain the old semantics, so I'd argue for flipping it (although it's semantically less straightforward); e.g. something like `NonFinal` so that its default `false` is actually the right thing for backwards compat. 

This should contribute to reduce the line count

internal/wasm/module.go
```diff
@@ -820,6 +923,32 @@ type FunctionType struct {
...
+	// its function-type semantics without change. (GC proposal.)
+	Form CompositeForm
+
+	// Fields lists the fields of a struct type, in declaration order.
+	// Used only when Form == CompositeFormStruct. (GC proposal.)
+	Fields []FieldType
+
+	// ArrayField is the single element field type of an array type.
+	// Used only when Form == CompositeFormArray. (GC proposal.)
+	ArrayField FieldType
```
evacchi, line 958: I wonder if we really want both `Fields []FieldType` and `ArrayField FieldType` discriminating over `Form`. 

My OOP-pilled mind obviously wants to use polymorphism and a `Composite` interface with 2 subtypes, but also `[]Fields` with `0` being the only index when `Form == CompositeFormStruct` would work. Of course that also means that we always go through a reference to find about the field of an array. 

Would that be an issue though?

internal/wasm/module.go
```diff
@@ -820,6 +923,32 @@ type FunctionType struct {
...
+	// supertype, or nil for top-level (no supertype) types. The MVP allows
+	// at most one supertype. (GC proposal.)
+	SuperTypeIndex *Index
+
+	// Final indicates whether this type can be a supertype of others.
+	// The shorthand 0x60 / 0x5F / 0x5E forms and the explicit 0x4F (sub
+	// final) form set this to true; the 0x50 (sub) form sets it to false.
+	// Default zero value (false) is benign for non-GC modules because they
+	// never declare subtypes. (GC proposal.)
+	Final bool
```
evacchi: note about Final above

internal/wasm/composite.go
```diff
@@ -0,0 +1,91 @@
...
+	case CompositeFormArray:
+		return "array"
+	}
+	return fmt.Sprintf("<unknown composite form %d>", f)
+}
+
+// PackedType is the storage type of a struct field or array element when the
+// field is not a full value type but a sub-byte / 2-byte packed integer.
+// Introduced by the GC proposal.
+type PackedType uint8
```
evacchi: I wonder if we may be able to encode all of this in `ValueType`. 

Possibly `FieldType` may be just a superset of valid `ValueType` values? 

essentially we only need i8+i16+1 bit mutable+1 packed

internal/wasm/store.go
```diff
@@ -86,6 +109,23 @@ type (
...
+		// Go-traceable reference to each object and is freed when the module
+		// is closed. Guarded by Store.gcMux.
+		//
+		// gcID is the module's id in Store.gcModules, assigned lazily on the
+		// first GC allocation (0 = unassigned). The table is append-only for
+		// the module's lifetime: a production wasm-gc engine would add
+		// per-object tracing/reclamation, but bounding it to the module
+		// already reclaims everything when the module closes.
+		gcObjects []any
+		gcID      uint32
```
evacchi: these 2 private fields are sort of a smell, as compared to the rest of the struct. 

the existing pattern for funcref/externref (raw pointer on stack, root kept alive elsewhere) should work for struct/array refs too, so we should be able to remove both private fields and the Store indirection

compare `uintptr(unsafe.Pointer(&function))` vs `functionFromUintptr.` -- we should be able to adopt the same pattern

evacchi (commented):
I'm leaving a few comments but I think I am going to create a branch off yours and add some commits on top with my suggestions

Issue #2483, opened 2026-03-14
Implement typed references


eliben:
**Describe the bug**
The `br_on_null` WASM instruction is not supported.

**To Reproduce**
This WAT code:

```
(module
  (func (export "test") (param externref) (result i32)
    (block
      local.get 0
      br_on_null 0
      drop
      i32.const 1
      return
    )
    i32.const 0
  )
)
```

When compiled to WASM (with `wasm-tools` or some other means), and loaded with Wazero reports this:

```
instantiate error: invalid function[0] export["test"]: invalid instruction 0xd5
```

**Environment (please complete the relevant information):**
- Go version: 1.24
- wazero Version: 929e4000d53c5dffb0f2c2f91c3e7366c2b2c3bb
- Host architecture: amd64
- Runtime mode: interpreter

**Additional context**

Sample Go code to load:

```
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/tetratelabs/wazero"
)

func main() {
	ctx := context.Background()

	wasm, err := os.ReadFile("br_on_null.wasm")
	if err != nil {
		fmt.Printf("read error: %v\n", err)
		os.Exit(1)
	}

	r := wazero.NewRuntime(ctx)
	defer r.Close(ctx)

	mod, err := r.Instantiate(ctx, wasm)
	if err != nil {
		fmt.Printf("instantiate error: %v\n", err)
		os.Exit(1)
	}
	defer mod.Close(ctx)
}
```

ncruces:
That instruction is part of the typed references proposal, which wazero doesn't support: https://github.com/wazero/wazero/issues/2426#issuecomment-3538706451

evacchi:
this will require some thinking; the new spec allows for new arbitrary types (typed function references).

For now I am landing on a single `uint64` encoding the same byte info for `wasm.ValueType` in the low byte, then use [8-15] for flags (concrete/nullable) and [16-31] for the index (which if I understand correctly is 32-bit).

however, this (1) touches a lot of places, (2) makes the type descriptor 1 -> 8 bytes. 

(also, I need to check if WasmGC makes things worse, i.e. if the 32-bit index is still enough)
EDIT: apparently it should  be fine, the GC specs adds more abstract heap types that we can encode in the low byte

evacchi:
one of the most annoying things is that, whatever we decide, it is likely that `wasm.ValueType` will no longer be an alias to `api.ValueType` (itself an alias for `byte`).

[...]

PR #2495, opened 2026-05-11
refactor: wasm.ValueType as type definition + widening (uint64)

evacchi:
Relates to #2483

This change makes sense mostly as a preparation for typed function references and GC, where the single `byte` repr is not sufficient anymore. This is also a follow up to #2489 where some references were introduced but the widening was not necessary.

Note: the widening is a distinct commit. We can also just introduce the type definition without the widening here. Widening could happen in the follow up PR for typed refs. I am doing it here because it doesn't change much in terms of lines of code.
 

## Details

- Migrates `wasm.ValueType` from a `byte` type alias to a distinct `uint64` type definition, changing every reference to it everywhere (that's why it's a "shotgun change", hitting everything: the codebase was using `api.ValueType` everywhere interchangeably with `wasm.ValueType`, because `wasm.ValueType` was a convenience alias to avoid imports.
- The wider type will allow encoding nullability, concrete type indices, and other flags in a single value, while maintaining backward compatibility through the public `api.ValueType` (`byte`) interface
- `api.ValueType` remains a `byte` so the public API is unchanged

## What changes in practice

**`ValueType` widened from `byte` to `uint64`** (`internal/wasm/module.go`):
- The low byte (`Kind()`) preserves the original WebAssembly type encoding (`0x7F` = i32, `0x7E` = i64, etc.)
- Upper bits reserved for flags (nullability, concrete ref index) used by the typed-refs work (will follow up with a PR for that too)
- All internal code now uses `wasm.ValueType` consistently instead of raw `byte` constants

**`api.ValueType` stays `byte`**:
- Public API surface is unchanged
- `function_definition.go` maps between internal `wasm.ValueType` and public `api.ValueType` via `Kind()`

**Binary decoding updated** (`internal/wasm/binary/`):
- Value type decoding returns `wasm.ValueType` instead of raw bytes
- Element segment, table, global, and import decoders adapted accordingly

**Test updates across the codebase**:
- All test files updated to use `wasm.ValueType*` constants instead of raw byte literals
- No behavioral changes, purely mechanical updates


internal/wasm/module.go
```diff
@@ -1158,36 +1198,78 @@ const (
...
 func isReferenceValueType(vt ValueType) bool {
-	return vt == ValueTypeExternref || vt == ValueTypeFuncref || vt == ValueTypeExnref
+	return vt.IsRef()
 }
 
 // isRefSubtypeOf returns true if actual is assignment-compatible with expected.
-// Currently, non-nullable ref types are desugared to nullable at decode time,
-// so this reduces to equality. When non-nullable ref types are properly supported,
-// this function should allow non-nullable to match nullable and vice versa.
 func isRefSubtypeOf(actual, expected ValueType) bool {
```
ncruces, line 1189: `isRefSubtypeOf` seems the same as `isStrictRefSubtypeOf`  which doesn't match the removed comments?

I'm not sure what's the correct fix here... but intuitively, `isRefSubtypeOf` doesn't need `expected.IsNullable()`?

ncruces (commented):
Thanks for splitting this off. Looks like 99% boilerplate, which if needed will make it easier to review future extensions. Left a comment almost to prove I read all through it. I'm not sure about the fix though.

internal/wasm/module.go
evacchi: actually, I think I ended up removing the strict check and just use isRefSubtypeOf; it was originally introduced with exception handling, but then while implementing typed refs I realized it was not necessary. I'll double-check :/

evacchi: uhm in fact I thought I had left this untouched in this PR

evacchi: heh! I did forget to push commit :D d9d4ccfab924648a03dc336e51e414d8d2e3d107 in this commit I am removing a bunch of stuff that is instead moved into the follow-up

internal/wasm/module.go
```diff
@@ -1167,7 +1171,11 @@ func ValueTypeName(t ValueType) string {
 	} else if t == ValueTypeExnref {
 		return "exnref"
 	}
-	return api.ValueTypeName(t)
+	return api.ValueTypeName(api.ValueType(t))
+}
+
+func (v ValueType) Kind() byte {
+	return byte(v)
```
achille-roussel, line 1178: Maybe to improve readability:

```suggestion
func (v ValueType) Kind() api.ValueType {
	return api.ValueType(v)
```

internal/wasm/module.go
```diff
@@ -1167,7 +1171,11 @@ func ValueTypeName(t ValueType) string {
 	} else if t == ValueTypeExnref {
 		return "exnref"
 	}
-	return api.ValueTypeName(t)
+	return api.ValueTypeName(api.ValueType(t))
```
achille-roussel, line 1174:
```suggestion
	return api.ValueTypeName(t.Kind())
```

ncruces (approved):

PR #2497, opened 2026-05-14
feat: typed function references spec

### LINKED ISSUE #2483: Implement typed references

evacchi:
Add experimental support for the [Typed Function References](https://github.com/WebAssembly/function-references) proposal. 

Closes https://github.com/wazero/wazero/issues/2483, follows up to the refactoring in https://github.com/wazero/wazero/pull/2495 and prepares to WasmGC.

Typed function references extend WebAssembly's type system with non-nullable reference types and concrete function type indices (`(ref $t)`, `(ref null $t)`), enabling direct calls through typed references (`call_ref`, `return_call_ref`) and null-aware branching (`br_on_null`, `br_on_non_null`, `ref.as_non_null`).

Excluding tests and spec suites, the feature amounts to roughly 1,900 lines of code.

Feature flag: `experimental.CoreFeaturesTypedFunctionReferences`

## What's the use for this?

Sadly, very little. This proposal is a pretty much just a prerequisite for the GC proposal.

On the flip side, it completes the exception handling spec (https://github.com/wazero/wazero/pull/2489): 
1. two EH spec tests were previously skipped because they required distinguishing nullable from non-nullable references. Those tests now pass.
2. The fuzzer is now enabled for both exception handling and typed function references. It could not be enabled for EH because the fuzzer would generate func refs.

So I guess, technically, it has a use :D

## Type System

`ValueType` (already `uint64` introduced in https://github.com/wazero/wazero/pull/2495) is now extended with bit flags to encode nullability and concrete type indices:

| Bits  | Purpose                         |
|-------|---------------------------------|
| 0-7   | Base type byte (same as before) |
| 8     | Non-nullable flag               |
| 9     | Concrete ref flag               |
| 32-63 | Type index (for `(ref $t)`)     |

This encoding should be fine for WasmGC too and should not require further changes in the near future.

Subtyping rules: non-nullable is a subtype of nullable (same kind/index); concrete function refs `(ref $t)` are subtypes of `funcref`.

## New Instructions

| Opcode            |  Hex   | Description                                              |
|-------------------|--------|----------------------------------------------------------|
| `call_ref`        | `0x14` | Indirect call through a typed function reference         |
| `return_call_ref` | `0x15` | Tail-call variant of `call_ref`                          |
| `ref.as_non_null` | `0xd4` | Assert ref is non-null, trap otherwise                   |
| `br_on_null`      | `0xd5` | Branch if null, push non-null ref on fall-through        |
| `br_on_non_null`  | `0xd6` | Branch if non-null (carrying the ref), fall-through on null |

## Validation

Tricky bits:

- Non-nullable local initialization tracking: `local.get` on a non-nullable ref local is rejected unless a `local.set`/`local.tee` has been executed in the same or enclosing block scope. State is saved/restored at block boundaries per the spec (needed an additional field to keep track of init'd values)
- `ref.func` now pushes `(ref $t)` (the concrete non-nullable type of the referenced function) onto the validation type stack instead of plain `funcref`, so that passing it to `call_ref` type-checks without an upcast.
- Block types, element segments, table types, and const expressions all support concrete ref types.
- The type section now validates forward references (standalone types can only reference previously defined types; rec group members can reference each other).

## Interpreter

The five new opcodes are compiled to new IR operations and executed in the interpreter loop. `call_ref` / `return_call_ref` load the function instance from the opaque reference pointer, null-check, and dispatch. `br_on_null` / `br_on_non_null` pop the reference, check nullity, and either branch or fall through with the appropriate stack state.

## Compiler (wazevo)

Implemented entirely as SSA-level lowering with no backend-specific code:

- `call_ref` / `return_call_ref`: load executable and module context pointers from the function instance, null-check via `ExitIfTrueWithCode(ExitCodeNullReference)`, then dispatch as an indirect call.
- `br_on_null` / `br_on_non_null`: compare against zero, branch with trampoline blocks for try-table exits and listener support.
- `ref.as_non_null`: null-check with trap.

## Binary Decoding

- `decodeRefType` helper extracted and shared across `value.go`, `element.go`, `table.go`, and `code.go` for consistent handling of `(ref null ht)` / `(ref ht)` prefixes.
- Tables support the `0x40 0x00` prefix for initializer expressions (required for non-nullable table element types).
- `DecodeBlockType` handles concrete ref types as block results.

## Cross-Module Linking

`call_indirect` uses `FunctionTypeID` for fast runtime type checks. The existing `FunctionType.key()` method builds the key from raw `ValueType` bytes, but with concrete refs `(ref $0)` in module A and `(ref $0)` in module B may refer to structurally identical types at different local indices. `structuralTypeKey` fixes this by replacing local type indices with the already-assigned `FunctionTypeID` of the referenced type, so two modules with the same structural signature share a single `FunctionTypeID`.

## Spec Suite

The spec test suite uses `wasm-tools json-from-wast` (same as exception handling). All 22 test files pass.

## Fuzzing

The fuzzer (`nodiff`) now enables both `CoreFeaturesExceptionHandling` and `CoreFeaturesTypedFunctionReferences`. Dummy import generation handles `exnref` and concrete ref types.


internal/wasm/binary/section.go
```diff
@@ -47,12 +53,33 @@ func decodeTypeSection(enabledFeatures api.CoreFeatures, r *bytes.Reader) ([]was
...
 }
 
+// validateTypeForwardRefs rejects concrete reference types (ref $t) whose type
+// index is not yet defined. For standalone types, maxTypeIndex is the count of
+// types decoded so far; for rec groups, it is the index after the last member,
+// allowing mutual references within the group.
+func validateTypeForwardRefs(ft *wasm.FunctionType, maxTypeIndex uint32) error {
+	for _, vt := range ft.Params {
+		if vt.IsConcreteRef() && vt.TypeIndex() >= maxTypeIndex {
+			return fmt.Errorf("unknown type")
```
achille-roussel: Looks like we could make this a sentinel value?

achille-roussel: Alternatively, inject more details about the unknown type for debugging?

achille-roussel (approved):
👏

internal/wasm/func_validation.go
```diff
@@ -531,7 +536,7 @@ func (m *Module) validateFunctionWithMaxStackValues(
 					return fmt.Errorf("inconsistent block type length for %s at %d; %v (ln=%d) != %v (l=%d)", OpcodeBrTableName, l, defaultLabelType, ln, tableLabelType, l)
 				}
 				for i := range defaultLabelType {
-					if defaultLabelType[i] != valueTypeUnknown && defaultLabelType[i] != tableLabelType[i] {
+					if defaultLabelType[i] != valueTypeUnknown && !areRefTypesCompatible(defaultLabelType[i], tableLabelType[i]) {
 						return fmt.Errorf("incosistent block type for %s at %d", OpcodeBrTableName, l)
```
ncruces: typo: inconsistent

ncruces (approved):
This is the most I could find. I guess that's good?

internal/wasm/binary/section.go
evacchi: went with the details 👍🙏

internal/wasm/func_validation.go
evacchi: ✅

ncruces (approved):

anuraaga (approved):

wwqgtxx:
While upgrading dependencies, we discovered that the following code in `internal/wasm/store.go` caused `fastschema/qjs@v0.0.6` to fail, returning a highly confusing error: `failed to instantiate module: import func[env.jsFunctionProxy]: signature mismatch: i32i64i32i32_i64 != i32i64i32i32_i64`. Everything returned to normal after removing that conditional check.
```golang
if m.TypeIDs != nil && importedModule.TypeIDs != nil {
	// Use structural type IDs for comparison (handles concrete ref types across modules).
	actualTypeIdx, ok := src.typeIndexOfFunction(imported.Index)
	matched = ok && importedModule.TypeIDs[actualTypeIdx] == m.TypeIDs[i.DescFunc]
}
```

evacchi:
Thanks I'll take a look!

evacchi:
@wwqgtxx PR #2511 should address your issue, let me know if that works!

wwqgtxx:
Thank you for the quick turnaround. I tested locally and confirmed that `fastschema/qjs` is working again after applying the patch.

PR #2500, opened 2026-05-18
feat: experimental Fuel API support

JairusSW:
I'm building a benchmarking framework for AssemblyScript running on top of wazero. Timing-based statistics work fine, but it'd be great to also retrieve the number of executed wasm operations i.e. fuel.

This PR adds a basic experimental fuel API. It only supports the `interpreter` engine for now so that the diff is managable (and easy to review/make sure i'm headed in the right direction). I'd appreciate some feedback on the API shape.

It adds a few methods:

`experimental.SetFuel(ctx, units)`

`experimental.GetFuel(ctx)`

`experimental.AddFuel(ctx, units)`

Usage is in the tests

Not sure what your policy is on LLMs, but I did use Claude to write comments


JairusSW:
Should we switch over to a weighted fuel system? Wasmtime uses a 1:1 ratio with fuel:op but perhaps some optional weights would make sense since `i32.add` would technically use the same fuel as `memory.copy`

IIRC I believe Wasmer uses a weighted system. I'll take a look at it when I get a moment

paralin:
The "fuel" concept is very interesting, thanks.

JairusSW:
closed due to the core team's laziness and unwillingness to respond

evacchi:
Sorry about that, unfortunately there is a long backlog and we're all volunteers at this point :)

JairusSW:
Ah, i probably spoke too harshly then. I thought some of y'all worked for Tetrate. Sorry about that, definitely understand the feeling of volunteering haha

evacchi:
Thanks. The project has been donated to the community hence the change in namespace

PR #2501, opened 2026-05-26
compiler: support branch target CFI + allocate and release stacks with MAP_STACK on OpenBSD

catap:
OpenBSD requires execution stacks to come from mappings marked with MAP_STACK, while wazevo switches to a private execution stack before running generated code. Allocate OpenBSD call stacks with mmap(MAP_STACK), release them on finalization and after stack growth, and keep heap backed stacks for other systems.

OpenBSD also enforces branch target CFI when hardware support is present: CET IBT on amd64 and BTI on arm64. Emit ENDBR64 and BTI JC landing pads for generated function entries, entry preambles, Go call trampolines, stack grow trampolines, Go reentry points, and br_table targets; adjust relative address handling where the leading landing pad changes offsets.

Keep README platform wording conservative, since OpenBSD arm64 is locally verified but not covered by CI.

### DIFF
--- a/internal/engine/wazevo/backend/isa/amd64/abi_entry_preamble.go
+++ b/internal/engine/wazevo/backend/isa/amd64/abi_entry_preamble.go
@@ -28,6 +28,7 @@ var (
 // CompileEntryPreamble implements backend.Machine.
 func (m *machine) CompileEntryPreamble(sig *ssa.Signature) []byte {
 	root := m.compileEntryPreamble(sig)
+	m.c.Emit4Bytes(endbr64Instruction)
 	m.encodeWithoutSSA(root)
 	buf := m.c.Buf()
 	return buf
--- a/internal/engine/wazevo/backend/isa/amd64/abi_entry_preamble.go
+++ b/internal/engine/wazevo/backend/isa/amd64/abi_entry_preamble.go
@@ -28,7 +28,6 @@ var (
 // CompileEntryPreamble implements backend.Machine.
 func (m *machine) CompileEntryPreamble(sig *ssa.Signature) []byte {
 	root := m.compileEntryPreamble(sig)
-	m.c.Emit4Bytes(endbr64Instruction)
 	m.encodeWithoutSSA(root)
 	buf := m.c.Buf()
 	return buf
@@ -42,9 +41,12 @@ func (m *machine) compileEntryPreamble(sig *ssa.Signature) *instruction {
 
 	//// ----------------------------------- prologue ----------------------------------- ////
 
+	// Landing pad for indirect branch tracking (Intel CET).
+	cur := linkInstr(root, m.allocateInstr().asEndbr64())
+
 	// First, we save executionContextPtrReg into a callee-saved register so that it can be used in epilogue as well.
 	// 		mov %executionContextPtrReg, %savedExecutionContextPtr
-	cur := m.move64(executionContextPtrReg, savedExecutionContextPtr, root)
+	cur = m.move64(executionContextPtrReg, savedExecutionContextPtr, cur)
 
 	// Next is to save the original RBP and RSP into the execution context.
 	cur = m.saveOriginalRSPRBP(cur)
--- a/internal/engine/wazevo/backend/isa/amd64/abi_go_call.go
+++ b/internal/engine/wazevo/backend/isa/amd64/abi_go_call.go
@@ -271,6 +271,7 @@ func (m *machine) CompileGoFunctionTrampoline(exitCode wazevoapi.ExitCode, sig *
 	cur = m.revertRBPRSP(cur)
 	linkInstr(cur, m.allocateInstr().asRet())
 
+	m.c.Emit4Bytes(endbr64Instruction)
 	m.encodeWithoutSSA(m.rootInstr)
 	return m.c.Buf()
 }
@@ -330,6 +331,8 @@ func (m *machine) storeReturnAddressAndExit(cur *instruction, execCtx regalloc.V
 
 	nop, l := m.allocateBrTarget()
 	cur = linkInstr(cur, nop)
+	endbr := m.allocateInstr().asEndbr64()
+	cur = linkInstr(cur, endbr)
 	readRip.asLEA(newOperandLabel(l), ripReg)
 	return cur
 }
@@ -376,6 +379,7 @@ func (m *machine) CompileStackGrowCallSequence() []byte {
 	cur = m.revertRBPRSP(cur)
 	linkInstr(cur, m.allocateInstr().asRet())
 
+	m.c.Emit4Bytes(endbr64Instruction)
 	m.encodeWithoutSSA(m.rootInstr)
 	return m.c.Buf()
 }
--- a/internal/engine/wazevo/backend/isa/amd64/abi_go_call.go
+++ b/internal/engine/wazevo/backend/isa/amd64/abi_go_call.go
@@ -26,6 +26,9 @@ func (m *machine) CompileGoFunctionTrampoline(exitCode wazevoapi.ExitCode, sig *
 	cur := m.allocateNop()
 	m.rootInstr = cur
 
+	// Landing pad for indirect branch tracking (Intel CET).
+	cur = linkInstr(cur, m.allocateInstr().asEndbr64())
+
 	// Execution context is always the first argument.
 	execCtrPtr := raxVReg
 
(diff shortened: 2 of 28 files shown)


catap:
Ooops, [added](https://github.com/wazero/wazero/compare/b885c3425a39e1a3cbad5cd1c0f1a49454f94fb9..cee7ed4ae6e01c1077eb744a86050711d453a78c) missed `bti jc` into expected output.

ncruces:
Nice! I never figured out what was "wrong" with OpenBSD.

Just to confirm: you verified OpenBSD on ARM64 locally?

I didn't work in the compiler internals, so I'd really rather have someone else review this too.

The only comment I have is that _maybe_ we should use `MAP_STACK` on all the platforms it exists on, not just OpenBSD? I know it exists of FreeBSD and (even if it's a NOOP there) on Linux. That might make any issues more visible.

catap:
Yes, I had tested OpenBSD on -current on amd64 with IBT and non-IBT cpu, and on arm64 which has BTI, and as far as I know all arm64 which works in openbsd has BTI.

and it was tested as go test that was enough I think, because before it fails as soon as I moved openbsd to compiler.

regarding use of map-stack: it is defeintly out of the scope of that PR 🙈

ncruces:
So the instructions are being added regardless of OpenBSD, but IIUC are no-ops on all processors/OSes that don't support them.

The benefit here is having a working OpenBSD compiler; other OSes won't benefit.

For there to be some benefit elsewhere we could:
- use `MAP_STACK` on systems that support it (FreeBSD seems to do something with it, Linux understands it but it's a no-op)
- use `PROT_BTI` on Linux ARM64

catap:
Yes, it should be harmless for all other OSes as far as I know.

catap:
@ncruces just a friendly ping!

ncruces:
From my side, this seems OK. But:
 > I didn't work in the compiler internals, so I'd really rather have someone else review this too.

There's also the question of if we want to change the compiled code for the exclusive benefit of OpenBSD. 

Which is why I wrote enabling this on Linux ARM64 would be interesting.

My vote is in, but I won't be deciding alone.

catap:
@ncruces got it, thanks.

The real root cause why I did it is this trhead: https://marc.info/?t=177971001600002&r=1&w=2

Long story short: a user complained that his site doesn't work in hugo anymore since upgrade from OpenBSD 7.8 to 7.9, and the cause is... that hugo switched to resize webp via wazero, and wazero in OpenBSD is very slow, and it hits timeout.

Regarding Linux on ARM64, I don't have where to test it nor idea does Linux supports it, and because it works without this, I think it does not :-)

internal/engine/wazevo/backend/isa/amd64/abi_entry_preamble.go
```diff
@@ -28,6 +28,7 @@ var (
 // CompileEntryPreamble implements backend.Machine.
 func (m *machine) CompileEntryPreamble(sig *ssa.Signature) []byte {
 	root := m.compileEntryPreamble(sig)
+	m.c.Emit4Bytes(endbr64Instruction)
```
evacchi: I don't think we need to emit the instruction explicitly, instead we can make this part of the preamble, allocate and link the instruction

internal/engine/wazevo/backend/isa/amd64/abi_go_call.go
```diff
@@ -271,6 +271,7 @@ func (m *machine) CompileGoFunctionTrampoline(exitCode wazevoapi.ExitCode, sig *
 	cur = m.revertRBPRSP(cur)
 	linkInstr(cur, m.allocateInstr().asRet())
 
+	m.c.Emit4Bytes(endbr64Instruction)
```
evacchi: same here, we can fold this into `encodeWithoutSSA()`

internal/engine/wazevo/backend/isa/amd64/abi_go_call.go
```diff
@@ -330,6 +331,8 @@ func (m *machine) storeReturnAddressAndExit(cur *instruction, execCtx regalloc.V
 
 	nop, l := m.allocateBrTarget()
 	cur = linkInstr(cur, nop)
+	endbr := m.allocateInstr().asEndbr64()
+	cur = linkInstr(cur, endbr)
```
evacchi, line 338: basically, similarly to how it's done here

internal/engine/wazevo/backend/isa/amd64/abi_go_call.go
```diff
@@ -376,6 +379,7 @@ func (m *machine) CompileStackGrowCallSequence() []byte {
 	cur = m.revertRBPRSP(cur)
 	linkInstr(cur, m.allocateInstr().asRet())
 
+	m.c.Emit4Bytes(endbr64Instruction)
```
evacchi: ditto

internal/engine/wazevo/backend/isa/amd64/instr_encoding.go
```diff
@@ -1419,6 +1421,11 @@ func encodeRet(c backend.Compiler) {
 	c.EmitByte(0xc3)
 }
 
+// endbr64Instruction encodes "ENDBR64", the landing pad required by
+// Intel CET indirect branch tracking. Emit4Bytes writes little endian bytes,
+// so this emits f3 0f 1e fa.
+const endbr64Instruction = 0xfa1e_0ff3
+
```
evacchi: we don't need this constant if the sequence it's only ever referenced in `encode(c backend.Compiler)` we can move the comment there though

internal/engine/wazevo/backend/isa/arm64/abi_go_call.go
```diff
@@ -243,6 +243,7 @@ func (m *machine) CompileGoFunctionTrampoline(exitCode wazevoapi.ExitCode, sig *
 	ret.asRet()
 	linkInstr(cur, ret)
 
+	m.compiler.Emit4Bytes(btiJCInstruction)
```
evacchi: same set of comments applies here, we do not need to encode explicitly, we can just wire the instructions as we do normally

evacchi (commented):
I think we can make this slightly cleaner, I'll send you a patch

catap:
@evacchi I'm waiting your patch / update which I'll gladly retest

internal/engine/wazevo/call_engine.go
```diff
@@ -155,14 +155,22 @@ func (c *callEngine) init() {
...
-	c.stack = make([]byte, stackSize)
+	c.stack = allocateStack(stackSize)
 	c.stackTop = alignedStackTop(c.stack)
 	if wazevoapi.StackGuardCheckEnabled {
 		c.execCtx.stackBottomPtr = &c.stack[wazevoapi.StackGuardCheckGuardPageSize]
 	} else {
 		c.execCtx.stackBottomPtr = &c.stack[0]
 	}
 	c.execCtxPtr = uintptr(unsafe.Pointer(&c.execCtx))
+	runtime.SetFinalizer(c, (*callEngine).finalize)
```
evacchi, line 166: I wonder if instead of the finalizer we should explicit close somewhere...

evacchi:
notice that, while MAP_STACK is only enabled for OpenBSD, this changes codegen on all platforms (BTW I sent a PR to your branch)

catap:
@evacchi I just added you to my repo, you should be able to push to it and the branch of pr directly

catap:
`475ab26315c01a53a8e3a94fd230bbe25a69813b`  commit passed `go test` on: OpenBSD/arm64 with IBT/BTI, OpenBSD/amd64 without IBT/BTI.

OpenBSD/amd64 with IBT/BTI will be made soon, but I think it will pass.

catap:
@evacchi it is passed on OpenBSD/amd64 with IBT/BTI

ncruces:
> notice that, while MAP_STACK is only enabled for OpenBSD, this changes codegen on all platforms (BTW I sent a PR to your branch)

Yeah, I mentioned that too. I think we can make some of this useful elsewhere: https://github.com/wazero/wazero/pull/2501#issuecomment-4554670874

evacchi:
Agreed! It would be nice to enable this on other platforms too

catap:
@evacchi but as separated step, right?

catap:
Just friendly ping!

catap:
Friendly ping!

evacchi:
sorry I promise I did not forget about this, I'm just kinda conflicted about having it enabled everywhere and (slightly) bloating the generated code :/

catap:
Just friendly ping!

PR #2504, opened 2026-06-07
compiler: try_table exception handling corrupts locals

### LINKED ISSUE #2503: Exception Handling support corrupts locals mutated after try table enter
**Describe the bug**
On entering a try_table, wazevo clones the stack, including all wasm locals. Then, when a throw happens inside of that try_table block, the cloned stack is restored. If locals are mutated after the try_table entry, those mutations are lost when throw occurs and the stack captured at try_table entry is restored.

**To Reproduce**
```wat
(module
  (tag $e)
  (func $thrower (throw $e))
  (func (export "f") (result i32)
    (local $flag i32)
    (local.set $flag (i32.const 1))        ;; flag = 1 at try_table entry
    (block $done (result i32)
      (block $catch (result exnref)
        (try_table (catch_all_ref $catch)
          (local.set $flag (i32.const 0))  ;; flag = 0 inside the try body, before the throw
          (call $thrower))                 ;; throws in a callee; caught by catch_all_ref
        (unreachable))
      (drop)                               ;; $catch: drop the exnref
      (br $done (local.get $flag)))         ;; return flag
  )
)
```

**Expected behavior**
`f()` must return **0** (the value of `$flag` at the throw point). 

| engine | `f()` |
|--------|-------|
| **wazevo (compiler)** | **1**  ← BUG |
| interpreter | 0  (correct) |
| V8 / browsers (cross-check) | 0  (correct) |

- Go version: 1.26
- wazero Version: 5500f5c2528897113e7ad2652a6d2e1ee436c541
- Host architecture: arm64
- Runtime mode: compiler

**Additional context**
I investigated fixing this and opening a PR, but the fix seems like it will involve either significant performance or simplicity tradeoffs - either maintaining a memory storage of mutated locals that can be applied back on top of the cloned stack prior to restore (costs allocating a buffer and a memory write for each local.set inside a try block), or some sort of unwind/stack map mechanism (significant complexity). I could also be completely wrong given my relative lack of familiarity with this code base, but I didn't see a simpler/performant way to fix this.

I encountered this while trying to get DuckDB running in wazero, which heavily uses c++ exceptions, and extracted the minimal reproducer above from my investigation of the failures.

evacchi:
Fixes #2503 

In `try_table` we clone the stack on entry and restore it when an exception is caught. This reverts locals to their values at entry of the block, instead of the values when an exception is thrown.

In the compiled code, the `try_table` trampoline returns and then a `br_table` instruction dispatches to either the body block (normal path) or a handler block (exception caught). The handler blocks branch off from that dispatch point, which is _before_ the try body; so the SSA sees them as successors of the "pre-body" block, thus it passes their "pre-body" values.

We introduce a heap-allocated struct (`localsSaveArea`) where we stash the values for locals in the `try_table` bodies (on write), so the handler can read the correct values.

- Inside a `try_table` body `local.set/local.tee` emit an extra store to the save area when `tryTableDepth > 0.`
- On `try_table` entry with catch clauses: stores all locals to the save area, increments depth.
- On `try_table` end with catch clauses: decrements depth, emits leave trampoline.
- Handlers call `reloadLocalsFromSaveArea()` to redefine SSA variables from the buffer.
- `try_table` with no catch clauses is treated as a plain block

We also introduce a new `TryTableInfo` struct, replacing the `[][]CatchClauseInstance` table. The new struct groups together `CatchClauses` with a couple new fields `NumLocals` and `ReuseLocals`; `ReuseLocals` is set for when we have nested `try_tables` within the same function (tryTableDepth > 0) so they share the enclosing handler's save area instead of allocating a fresh one.

At runtime:
- `TryTableEnter`: allocates the save area when `NumLocals > 0 && !ReuseLocals`.
- `TryTableLeave`: calls `restoreLocalsSaveAreaPtr` to walk the handler stack for the nearest enclosing save area.
- `doHandleException`: calls `restoreLocalsSaveAreaPtr` from the matched handler's index so the handler reads from the correct buffer.

Also, addressed cache serialization/deserialization, and added tests covering the issues.

---

additionally, added test cases and fixes for https://github.com/wazero/wazero/pull/2504#issuecomment-4655653689

- branch to a try_table's own label skips its leave
- catch clause jumps past an enclosing try_table without leaving it


evacchi:
@mpoindexter let me know if this actually fixes the bug for you, i.e. if DuckDB works now

mpoindexter:
yep, fixes the issue

ncruces (approved):

mpoindexter:
I tested further since this patch unblocked the immediate error, and found a couple of additional areas that are not working (not sure if they are introduced by this PR or preexisting). Both boil down to not correctly performing a try table exit. Minimal reproducers extracted from the real world code:

## Bug #1 — branch to a `try_table`'s own label skips its leave

### Reproducer (`bug1_br_target.wat`)

```wat
(module
  (tag $e)
  ;; T is exited by `br $T` (a branch to the try_table's own label). Its handler
  ;; must be popped. The subsequent `throw` must then propagate to the OUTER
  ;; handler (-> 1). If T's leave was skipped, the stale T catches it (-> 0).
  (func (export "run") (result i32)
    (block $ok (result exnref)
      (try_table $TO (catch_all_ref $ok)
        (block $h (result exnref)
          (try_table $T (catch_all_ref $h)
            (br $T))
          (throw $e))             ;; T popped -> TO catches; T stale -> T catches
        (drop)
        (return (i32.const 0)))   ;; INCORRECT: stale T caught it
      (unreachable))
    (drop)
    (i32.const 1)))               ;; CORRECT: throw propagated past popped T
```

- interpreter: `1`
- wazevo (unpatched): `0`

## Bug #2 — catch clause jumps past an enclosing `try_table` without leaving it

### Reproducer (`bug2_catch_outside.wat`)

```wat
(module
  (tag $e)
  ;; TI's catch (catch_all $L) jumps to $L, which is OUTSIDE the enclosing TO.
  ;; That jump must pop TO. The 2nd throw must then reach the OUTER handler
  ;; (-> 1). If TO was left stale, it catches the 2nd throw instead (-> 0).
  (func (export "run") (result i32)
    (block $ok (result exnref)
      (try_table $TOUTER (catch_all_ref $ok)
        (block $L
          (block $hO (result exnref)
            (try_table $TO (catch_all_ref $hO)
              (try_table $TI (catch_all $L)
                (throw $e))        ;; 1st throw -> TI catches -> after $L (exits TO)
              (unreachable))
            (unreachable))
          (drop)
          (return (i32.const 0)))  ;; INCORRECT: stale TO caught the 2nd throw
        (throw $e))                ;; 2nd throw -> TO popped: TOUTER; stale: TO
      (unreachable))
    (drop)
    (i32.const 1)))                ;; CORRECT
```

- interpreter: `1`
- wazevo (unpatched): `0`

evacchi:
nice! I pushed a commit that cleans up a bit the design of the original PR and also fixes this particular bug, let us know if you find more

mpoindexter:
With this latest commit all the cases I tested worked. 

I won't claim to have tested exhaustively, but the scenario I'm running is loading the DuckDB Python extension, and then using it to download data over http, then execute a query against the downloaded data, so it's hitting a pretty substantial portion of DuckDB (IO, query compilation, query execution).

mpoindexter:
As part of what I'm trying to do I noticed that with the deeply nested, try/catch heavy DuckDB code performance was...not great. Ran things through the profiler, and basically all the runtime was either garbage collection or cloneStack calls. Two things contributed to this:
- each try table entry clones the _whole_ stack
- each stack clone is separately allocated as a brand new byte slice, which creates a lot of garbage as we enter a leaf function with try/catch, exit, enter another leaf and so on

I think both are avoidable:
- we really only need to clone the current stack frame (vs. entire stack) for EH. Anything above this frame can't really be touched by the current function (as far as I can tell - it's quite possible I got this wrong)
- since the EH stack clones and local save areas grow and contract with the stack we can treat it like a sort of shadow stack, and grow/shrink from a single byte slice we grow as needed and truncate as we exit try tables

I did a pretty simple implementation of this at https://github.com/evacchi/wazero/compare/wasm-eh-bug...mpoindexter:wazero:optimize-stack-clones?expand=1 and got something like a 13x speedup on a simple benchmark, and my `time` stats on my DuckDB using program went from `45.16s user 8.12s system 128% cpu 41.404 total` to `1.76s user 0.89s system 11% cpu 23.907 total` (lots of network IO in there that accounts for the lesser difference in wall clock time). On arm64 it seems to work in practice for my workload, did not test amd64. Not opening as a PR since it was a quick hack and I did not fully verify it, but maybe there's some good ideas in there that would help.

Happy to wait for this to PR to merge and iterate further on the idea above if you think it's useful and like the approach, or if you wanted to you could take it and run with it yourself, or say "this isn't the right direction" I certainly won't claim it's necessarily the right approach, just an interesting data point.

evacchi:
Nice!! Definitely worth investigating, I'll take a look. For now I'd keep this PR scoped to bug fixing and keep perf for a follow up, but thanks nonetheless!

mpoindexter:
Makes total sense not to include it in this PR - just wanted to mention it here since the changes were done on top of this PR and depend on a lot in it, so didn't see a good way to surface these findings elsewhere. Thanks for all the work on this, EH support is exciting!

PR #2507, opened 2026-06-10
fix: allow fd_renumber onto stdio file descriptors

racerxdl:
### Problem

`FSContext.Renumber` returns `ENOTSUP` whenever the **target** fd is flagged as a preopen. Stdio fds 0-2 are registered with `IsPreopen: true`, so any guest that redirects its own stdio fails:

wasi-libc lowers `freopen(path, mode, stdin)` / `dup2(fd, STDIN_FILENO)` to `path_open` + `fd_renumber(fd, 0)`. Under wazero the renumber fails, the guest's stdin silently remains the original stream, and subsequent reads misbehave.

**Real-world reproduction:** PGlite (PostgreSQL 17.x compiled to WASI p1) runs `initdb`, which feeds its bootstrap script to itself by `freopen`ing stdin from `/tmp/initdb.boot.txt`. Under wazero it aborts with flex's `input in flex scanner failed`; the same module works under wasmtime, which permits renumbering onto stdio ([preview_1.rs fd_renumber](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-common/src/snapshots/preview_1.rs)).

### Change

Allow `fd_renumber` when the target is a stdio fd (0-2), matching wasmtime and POSIX `dup2` semantics. Directory preopens remain protected — replacing those would corrupt libc's preopen path-resolution table, which is the case the original guard exists for.

```go
if toFile.IsPreopen && to > FdStderr {
    return sys.ENOTSUP
}
```

### Testing

- New subtest in `TestFSContext_Renumber` covering renumber onto fds 0/1/2 (entry replaced, source fd invalidated), with the existing directory-preopen `ENOTSUP` cases unchanged.
- New `Test_fdRenumber` table cases (`file to stdin/stdout/stderr` → `ESUCCESS`).
- Full `go test ./...` passes.
- Verified end-to-end: PGlite's `pgl_initdb` completes under wazero with this change and the resulting Postgres answers queries over the wire protocol.

Automated by MisterMal

### DIFF
--- a/imports/wasi_snapshot_preview1/fs_test.go
+++ b/imports/wasi_snapshot_preview1/fs_test.go
@@ -2520,6 +2520,40 @@ func Test_fdRenumber(t *testing.T) {
 <== errno=ESUCCESS
 `,
 		},
+		// Stdio fds are flagged as preopens but, unlike directory preopens,
+		// they are valid renumber targets: libc emulates freopen/dup2 onto
+		// stdio as open + fd_renumber(fd, stdioFd), e.g. when a program
+		// redirects its own stdin from a file. wasmtime permits this.
+		{
+			name:          "file to stdin",
+			from:          fileFD,
+			to:            sys.FdStdin,
+			expectedErrno: wasip1.ErrnoSuccess,
+			expectedLog: `
+==> wasi_snapshot_preview1.fd_renumber(fd=4,to=0)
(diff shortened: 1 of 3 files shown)


evacchi (approved):

evacchi:
thanks for the contribution!

PR #2509, opened 2026-06-23
wazevo: add integer constant folding and arithmetic simplification pass

samyfodil:
Adds `passConstFoldingArithmeticOpt` to the wazevo SSA mid-end — one of the optimizations the `runPreBlockLayoutPasses` TODO calls out.

For the integer binary ops `iadd`/`isub`/`imul`/`band`/`bor`/`bxor` it:

- **folds `const op const`**, rewriting the instruction in place into an `Iconst`
- **applies algebraic identities** (`x+0 -> x`, `x*1 -> x`, `x*0 -> 0`, `x&x -> x`, ...) by aliasing the result to an existing value

Both rely on the existing alias + `passDeadCodeEliminationOpt` machinery to clean up the now-dead operands, so the pass adds no new infrastructure and is a single linear scan. Floats, shifts, and division are intentionally left out to keep it small and obviously correct; they can be added when a benchmark shows they pay off.

### Tests

- unit coverage for the fold arithmetic (`TestFoldBinaryIntConst`)
- golden end-to-end tests for folding and the identity rewrites (`TestPass_constFoldingArithmetic`)
- `BenchmarkConstFolding` + `testdata/const_fold.wat` as reproducible evidence (see below)

All existing `ssa`, `frontend`, `backend`, and wazevo e2e tests pass.

### Benchmarks

A/B with the pass toggled on/off, compiling four real modules from different toolchains.

**How often it fires + generated code size:**

| Module (producer) | folds | identities | code size delta |
|---|---|---|---|
| `case.wasm` (TinyGo/LLVM) | 0 | 0 | 0 |
| `greet.wasm` (TinyGo/LLVM) | 0 | 0 | 0 |
| `wasi.wasm` (Rust/cargo-wasi, release) | 11 | 0 | -48 B (-0.015%) |
| `wasi.wasm` (Zig/clang) | 38 | 3 | -160 B (-0.10%) |

**Compile time** (benchstat, n=10): no statistically significant cost — `case.wasm` 10.39 -> 10.43 ms (p=0.32), `wasi.wasm` Zig/clang 16.51 -> 16.61 ms (p=0.53); geomean +0.47%, both p well above 0.05.

**Execution — two regimes:**

*(a) Already-optimized wasm* (`case.wasm`, TinyGo/LLVM; benchstat n=12): the pass fires 0x, the generated machine code is byte-identical, and runtime is unchanged (`random_mat_mul`, `reverse_array`, `fibonacci` all within noise). No effect — as expected when the producer already folded everything.

*(b) When constant arithmetic survives into a hot loop* — exercised by the committed `BenchmarkConstFolding` over `testdata/const_fold.wat`. benchstat of `main` vs this branch (n=10):

| Workload | folding off (main) | folding on (branch) | speedup |
|---|---|---|---|
| `const_chain` | 942.3 µs | 638.1 µs | **-32.3%** (p=0.000) |
| `identities` | 1393.8 µs | 996.9 µs | **-28.5%** (p=0.000) |

The win is large because wazevo has no LICM: without folding, the per-iteration constant arithmetic (a 6-op chain, and the `x*1`/`x+0`/`x&x`/`x|0` identities) re-runs every iteration; folding collapses it to a single constant load. Reproduce with the steps in `BenchmarkConstFolding`'s doc comment.

**Takeaway:** correct and effectively free on compile time. On already-optimized producer output it rarely fires, so no runtime change there; but where constant arithmetic does reach wazevo in hot code, it is worth ~30%. It also lays groundwork toward the higher-value passes the same TODO lists (CSE, copy-propagation).


evacchi:
hey thanks for the contribution

so we tried this earlier https://github.com/wazero/wazero/pull/1851 and ended up _not_ doing it because any compiler worthy of the name is already doing it (e.g. running through wasm-opt) and this ends up being just extra work for the compiler, which takes already some time on large binaries.

the candidates for evaluation are under https://github.com/wazero/wazero/tree/main/internal/integration_test/stdlibs (I'd start with Zig's stdlib)

what I remember is that the compile times went up, but the speed improvements were marginal

e.g see [this thread about the Zig stdlib](https://github.com/wazero/wazero/pull/1851#discussion_r1409942945):

> in terms of hard numbers, on my machine it compiles the wasm binary in about 7.02 seconds with the extra pass and 6.94-6.99 without
> 
> To be fair, that might not be surprising: e.g., if the wasm binary went through `wasm-opt` it might be already minimized, and if it weren't, operations between constants may not naturally occur in a Wasm binary very often (after all, it's already a compilation target)

samyfodil:
@evacchi fair, my goal is to have a set of optimization mechanisms that can be turned on and off, this is just the first of many. we can drop it, though in my case not all wasm is guaranteed to be generated by a compiler or run through wasm-opt.

ncruces:
Yes, but the choice for us, is whether we want to be left maintaining something that can be better done with `wasm-opt`. Note that you can even use `wasm-opt.wasm` to cleanup your Wasm modules - in Go - using `wazero`. So I'd say, it's entirely fair to assume the Wasm code is decent to begin with.

That said: there are optimizations that definitely do make sense, e.g., everything that can lead to improved bounds checks elimination (or not just bounds checks, but traps in general).

You'll have much better luck if you can demonstrate that some of these passes lead to further optimization opportunities of already good code.

samyfodil:
@ncruces This PR is basic, so it will only improve non-compiler-generated code, or code deliberately compiled with something like -O0. #2510, which I’d like to land after this one, does show improvements over compiled code. 
image

samyfodil:
Wazero is amazing project. However I don't have an issue maintaining a lot of LOCs with AI & I prefer to move fast. Moved effort to https://github.com/samyfodil/wazy. Thanks!

PR #2510, opened 2026-06-24
wazevo: add common subexpression elimination pass

samyfodil:
> Stacks on #2509 (const-folding). Until that lands, the diff here also shows the
> const-folding commits; this branch will be rebased to CSE-only once #2509 merges.

## What

Adds **common subexpression elimination** (`passCommonSubexpressionEliminationOpt`) to the wazevo SSA mid-end — one of the optimizations the `runPreBlockLayoutPasses` TODO calls out.

Walks blocks in reverse post-order (so a dominating definition is always seen first) and value-numbers each pure, single-result instruction by `(opcode, alias-resolved args, immediates, type)`. When an earlier identical instruction dominates the current one, the result is aliased to it and `passDeadCodeEliminationOpt` deletes the redundant instruction — reusing the existing alias + DCE machinery, so no new infrastructure.

## Conservative by design

- **Loads excluded** — a store to an overlapping address between two identical loads changes the result, which needs alias analysis.
- **Constants excluded** — the backend stays free to rematerialize immediates.
- A single representative per value number is kept, which catches dominator-chain and straight-line redundancy without a full scoped table.

## Measured

Isolated A/B of the CSE commit (`HEAD` vs `HEAD~1`), wazevo compiler, `BenchmarkInvocation/compiler/fib_for`, `-count=10`, benchstat (i9-12900HK, linux/amd64). Short benchmarks with real run-to-run variance, so `p` is reported alongside.

| workload | sec/op Δ | p |
|---|---|---|
| `fib_for_5` | +2.5% | 0.002 |
| `fib_for_10` | **−6.6%** | 0.000 |
| `fib_for_20` | **−6.3%** | 0.000 |
| `fib_for_30` | **−5.2%** | 0.000 |
| geomean | **−4.0%** | |

A net speedup on compute-bound recursion. The +2.5% on `fib_for_5` is ~1.3ns (51.9n→53.2n) — call-overhead-dominated trivial input, in the noise. Array/matmul workloads are flat, as expected (CSE excludes loads).

## Tests

Golden tests for within-block and cross-block-dominated reuse, plus negative cases (swapped operands stay distinct, identical loads are kept). Full `./internal/engine/wazevo/...` suite passes locally; the WebAssembly spec suite runs in CI.


samyfodil:
Closing — opened prematurely. Not intended for upstream review at this time.

samyfodil:
This stacks on #2509. Closing for now to keep review focused there first — I'll reopen this once #2509 is merged.

PR #2511, opened 2026-06-28
fix: store should recompute type IDs at instantiation time

evacchi:
PR https://github.com/wazero/wazero/pull/2497 introduced a regression; `EqualsSignature` (a simple `ValueType` equality) was replaced by a more articulate logic, because `ValueType` now might embed concrete type information (i.e. a `ValueType` encodes extra info in some of its bits)

When we compare types across different stores (in the presence of a compilation cache), type IDs may not match because they are assigned in the order they are seen, so one short-circuit in the comparison logic caused import matching to fail even if the signatures were structurally identical.

We now recompute the typeIDs in the instantiating store, to ensure that they actually match. We might want to revisit this in the future and possibly design a more efficient procedure for equality.

### DIFF
--- a/runtime.go
+++ b/runtime.go
@@ -333,8 +333,16 @@ func (r *runtime) InstantiateModule(
 		name = code.module.NameSection.ModuleName
 	}
 
+	// Recompute typeIDs for the instantiating store. The compiled module may
+	// have been compiled in a different runtime/store (shared compilation
+	// cache pattern), so its cached typeIDs can differ from this store's.
+	typeIDs, err := r.store.GetFunctionTypeIDs(code.module.TypeSection)
+	if err != nil {
+		return nil, err
+	}
+
 	// Instantiate the module.
-	mod, err = r.store.Instantiate(ctx, code.module, name, sysCtx, code.typeIDs)
+	mod, err = r.store.Instantiate(ctx, code.module, name, sysCtx, typeIDs)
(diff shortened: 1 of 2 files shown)


evacchi:
confirmed fixed by user @wwqgtxx

wwqgtxx:
By the way, the test section of this PR seems to have accidentally placed the comment for `TestNewRuntime_concurrent` before the body of the newly added function.

evacchi:
ouch yes you're right

ncruces (approved):

samyfodil:
Follow-up on this one: re-resolving on every `InstantiateModule` costs a per-type map lookup under the store lock, and the only case that needs it is a `CompiledModule` reaching a *different* store (the shared-compilation-cache pattern). Recording which store assigned the IDs makes it skippable:

```go
typeIDs := code.typeIDs
if code.typeIDStore != r.store { // set in CompileModule
	if typeIDs, err = r.store.GetFunctionTypeIDs(code.module.TypeSection); err != nil {
		return nil, err
	}
}
```

Same-store instantiation then behaves exactly as it did before the regression; cross-store still gets fresh IDs. Types are interned per store, so re-resolving there returns the same IDs anyway.

Carrying this in a derivative (https://github.com/samyfodil/wazy/pull/37) — happy to send it as a PR here if it's wanted.

PR #2514, opened 2026-07-04
wazevo: elide bounds checks for constant addresses within memory minimum

mgilbir:
## Summary

A memory's declared minimum size is a static lower bound on its length at any point in time, since memories can only grow. Any access at a constant address whose end lies within that minimum can therefore never be out of bounds, and needs no runtime check.

Previously every such access emitted a length load, compare, conditional branch, and an inline exit sequence, even when the address was a compile-time constant guaranteed safe by the memory's minimum size. This pattern is common in interpreter-style modules whose working state lives in globals at fixed addresses, and known-safe bounds are invalidated at loop headers, so accesses inside loops paid the full check on every iteration.

On a TeX-engine (tectonic) wasm32-wasip1 module compiling a tabularray longtblr document on an M1 Pro, this reduces wall time from 16.4s to ~9.5-10.5s (~40%), reaching parity with wasmtime/Cranelift on the same module, and shrinks generated code for the hottest functions by ~2x.

Golden test expectations for constant-address accesses within the minimum are regenerated accordingly. Spec tests and the rest of the suite pass unchanged, including the memory out-of-bounds trap tests — elision only applies to accesses that can never trap.

Not stacked on #2515 (shared conditional-trap exit sequences) — the two touch disjoint files (this PR only touches the frontend; #2515 only touches the arm64/amd64 backends) and neither depends on the other to build or pass tests. They do both regenerate golden assembly listings for the same couple of test cases in `backend_test.go` (`memory_stores`, `AtomicRmwAdd`), so whichever lands second will need those golden values regenerated against the other's already-merged change — see the note on #2515 for details.

## Test plan

- [x] `go test ./...`
- [x] spec tests
- [x] cross-compile checks from `CONTRIBUTING.md` (plan9, wasip1, gojs, aix, s390x, ppc64le, arm, 386)
- [x] `make format`

---

Note on process: this change was developed with AI assistance (Claude Code); I've reviewed the code and the reasoning behind it to the best of my ability.

### DIFF
--- a/internal/engine/wazevo/backend/backend_test.go
+++ b/internal/engine/wazevo/backend/backend_test.go
@@ -1713,151 +1713,16 @@ L2:
 L0 (SSA Block: blk0):
 	stp x30, xzr, [sp, #-0x10]!
 	str xzr, [sp, #-0x10]!
-	mov x8, xzr
-	uxtw x8, w8
-	ldr w9, [x1, #0x10]
-	add x10, x8, #0x4
-	subs xzr, x9, x10
-	mov x10, x0
-	b.hs #0x34, (L10)
-	movz x11, #0x4, lsl 0
-	str w11, [x10]
-	mov x11, sp
-	str x11, [x10, #0x38]
-	adr x11, #0x0
-	str x11, [x10, #0x30]
-	exit_sequence x10
-L10:
-	ldr x10, [x1, #0x8]
-	add x8, x10, x8
+	ldr x8, [x1, #0x8]
 	str w2, [x8]
-	orr w8, wzr, #0x8
-	uxtw x8, w8
-	add x11, x8, #0x8
-	subs xzr, x9, x11
-	mov x11, x0
-	b.hs #0x34, (L9)
-	movz x12, #0x4, lsl 0
-	str w12, [x11]
-	mov x12, sp
-	str x12, [x11, #0x38]
-	adr x12, #0x0
-	str x12, [x11, #0x30]
-	exit_sequence x11
-L9:
-	add x8, x10, x8
-	str x3, [x8]
-	orr w8, wzr, #0x10
-	uxtw x8, w8
-	add x11, x8, #0x4
-	subs xzr, x9, x11
-	mov x11, x0
-	b.hs #0x34, (L8)
-	movz x12, #0x4, lsl 0
-	str w12, [x11]
-	mov x12, sp
-	str x12, [x11, #0x38]
-	adr x12, #0x0
-	str x12, [x11, #0x30]
-	exit_sequence x11
-L8:
-	add x8, x10, x8
-	str s0, [x8]
-	orr w8, wzr, #0x18
-	uxtw x8, w8
-	add x11, x8, #0x8
-	subs xzr, x9, x11
-	mov x11, x0
-	b.hs #0x34, (L7)
-	movz x12, #0x4, lsl 0
-	str w12, [x11]
-	mov x12, sp
-	str x12, [x11, #0x38]
-	adr x12, #0x0
-	str x12, [x11, #0x30]
-	exit_sequence x11
-L7:
-	add x8, x10, x8
-	str d1, [x8]
-	orr w8, wzr, #0x20
-	uxtw x8, w8
-	add x11, x8, #0x1
-	subs xzr, x9, x11
-	mov x11, x0
-	b.hs #0x34, (L6)
-	movz x12, #0x4, lsl 0
-	str w12, [x11]
-	mov x12, sp
-	str x12, [x11, #0x38]
-	adr x12, #0x0
-	str x12, [x11, #0x30]
-	exit_sequence x11
-L6:
-	add x8, x10, x8
-	strb w2, [x8]
-	movz w8, #0x28, lsl 0
-	uxtw x8, w8
-	add x11, x8, #0x2
-	subs xzr, x9, x11
-	mov x11, x0
-	b.hs #0x34, (L5)
-	movz x12, #0x4, lsl 0
-	str w12, [x11]
-	mov x12, sp
-	str x12, [x11, #0x38]
-	adr x12, #0x0
-	str x12, [x11, #0x30]
-	exit_sequence x11
-L5:
-	add x8, x10, x8
-	strh w2, [x8]
-	orr w8, wzr, #0x30
-	uxtw x8, w8
-	add x11, x8, #0x1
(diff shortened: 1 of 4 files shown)


ncruces:
Hi @mgilbir so this one seems relatively simple and maintainable (around 20 LoC), for an optimization that `wasm-opt` can't do.

Questions:
- what happens to existing benchmarks, did you try?
- is this common enough in Wasm modules (e.g. the ones already in the repo)?
- how are you compiling tectonic to get such an improvement?

IMO an optimization that is:
1. not too complex
2. can't be shifted elsewhere (to `wasm-opt`)
3. improves real world code

Is exactly the kind of stuff we should be targeting.

jerbob92:
## Benchmark report: this PR vs current `main` (real-world workload + wazero benches)

We benchmark wazero as the runtime for [go-pdfium](https://github.com/klippa-app/go-pdfium): PDFium compiled to a 5.3 MB wasm module (~6,400 functions, built with wasm exception handling / `try_table`), rendering and extracting text from real PDF documents. Hopefully useful as a large-real-module datapoint next to the micro benches.

**Setup:** this PR merged onto current `main` (`3ab42173`) vs plain `main`. Apple M5 (darwin/arm64), Go 1.26. go-pdfium numbers are medians of 3 interleaved A/B rounds (one `OpenDocument` + render at 200 DPI + close per op; text extraction is `GetPageTextStructured`, thousands of small guest calls per op). wazero numbers are `benchstat` over `-count=6` of `internal/integration_test/bench` (compiler engine + compilation).

### ✅ Correctness

The **full go-pdfium test suite passes on this PR** — ~660 specs, run on both the compiler and interpreter engines, including heavy `try_table` exception-handling usage and pixel-exact render comparisons.

### go-pdfium (large real module): neutral

| Workload | main | PR | Δ |
|---|---|---|---|
| render rect-wrong.pdf | 28.91 ms | 29.86 ms | +3.3% |
| render alpha_channel.pdf | 90.9 ms | 90.6 ms | −0.3% |
| render test.pdf | 1.182 ms | 1.183 ms | ~0% |
| render building_plan.pdf | 2.735 ms | 2.743 ms | +0.3% |
| render invoice A | 10.34 ms | 10.40 ms | +0.6% |
| render image_page.pdf | 108.7 ms | 109.0 ms | +0.3% |
| render invoice B | 5.77 ms | 5.75 ms | −0.3% |
| render mona.pdf | 0.623 ms | 0.620 ms | −0.5% |
| text extraction ×2 | 2.332 / 0.313 ms | 2.343 / 0.314 ms | ~0% |

Neutral overall, which matches the PR's scope: C/C++ compiled through emscripten addresses guest memory almost entirely through dynamic pointers, so the constant-address elision rarely fires in this module. The one visible mover (rect-wrong +3.3%, consistent across rounds) looks like code-layout shift rather than the optimization itself — every other document is flat.

### wazero's own benchmarks: geomean −8.0%

```
                                                     │      main       │              this PR                │
                                                     │      sec/op     │    sec/op     vs base               │
Invocation/compiler/fib_for_10-10                         153.6n ±  1%   147.7n ±  0%   -3.84% (p=0.002 n=6)
Invocation/compiler/string_manipulation_size_50-10        7.353µ ±  2%   6.503µ ± 14%  -11.55% (p=0.041 n=6)
Invocation/compiler/string_manipulation_size_100-10       23.34µ ±  3%   20.89µ ± 11%  -10.51% (p=0.002 n=6)
Invocation/compiler/string_manipulation_size_1000-10      1.655m ±  1%   1.526m ±  1%   -7.81% (p=0.002 n=6)
Invocation/compiler/reverse_array_size_500-10             1.615µ ±  1%   1.504µ ±  1%   -6.87% (p=0.002 n=6)
Invocation/compiler/reverse_array_size_10000-10           75.63µ ±  2%   69.03µ ±  0%   -8.72% (p=0.002 n=6)
Invocation/compiler/random_mat_mul_size_20-10             47.43µ ±  2%   43.94µ ±  1%   -7.36% (p=0.002 n=6)
Compilation/without_extern_cache-10                       5.135m ±  1%   4.957m ±  4%   -3.46% (p=0.026 n=6)
geomean                                                                                 -8.04%
```

(Omitted rows were statistically flat; `fib_for_30` read +4.9% but with ±19% variance, and `random_mat_mul_size_10` read −54% with ±53% variance on base — neither is a reliable signal.)

### Verdict

Solid win where it applies: wazero's own guest kernels (which do use constant addresses) improve by a **−8% geomean**, compilation gets slightly faster, and a large dynamic-address module like PDFium is unaffected with the full test suite green. 👍

evacchi:
this is very helpful @jerbob92 would you be able to run this against ./internal/integration_test/stdlibs as well? you can limit to the zig stdlib

jerbob92:
Here you go:

Follow-up: I also ran this PR against the Zig stdlib integration benchmarks in `internal/integration_test/stdlibs`.

Setup: base = current `main` (6a92c9b3), PR branch = base + this PR only. Apple M5, idle machine, interleaved passes, 10 samples per benchmark. Zig test binaries built with Zig 0.11.0 per the Makefile (`test.wasm` + wasm-opt'd `test-opt.wasm`); every `Run` iteration executes the full Zig stdlib test suite (~11.5s of guest execution) and must exit 0.

```
                             │      main      │             PR #2514              │
                             │     sec/op     │   sec/op    vs base               │
Zig/Compile/test-opt.wasm-10       1.773 ± 3%   1.756 ± 3%       ~ (p=0.579 n=10)
Zig/Run/test-opt.wasm-10           11.17 ± 1%   11.23 ± 1%  +0.54% (p=0.029 n=10)
Zig/Compile/test.wasm-10           2.139 ± 3%   2.071 ± 5%  -3.18% (p=0.029 n=10)
Zig/Run/test.wasm-10               11.55 ± 1%   11.59 ± 1%       ~ (p=0.579 n=10)
geomean                            4.703        4.664       -0.83%
```

Verdict: execution is neutral (the +0.54% on test-opt Run is marginal). The −3.2% compile time on the unoptimized binary is plausibly real: eliding bounds checks at compile time means fewer nodes to lower and less code to emit, and the unstripped binary has the most elidable constant-address accesses. That execution is flat here is consistent with the optimization's shape — constant addresses within the memory minimum are dense in emscripten-built C/C++ modules but rarer in Zig stdlib codegen.

evacchi (approved):
I think it's fine, maybe I'll run some fuzz later

ncruces (approved):

PR #2515, opened 2026-07-04
wazevo: share conditional-trap exit sequences per function (arm64 + amd64)

mgilbir:
## Summary

Each conditional trap (bounds checks, div-by-zero, etc.) inlined a full exit sequence at the check site, with the guard branch jumping over it — so the hot path took a branch on every successful check and the sequence was duplicated at every site.

Instead, this branches to a per-function trap island shared by all sites with the same exit code, and falls through on the hot path. Islands are materialized after register allocation using only real/dead registers so the change doesn't clobber live state on the never-returning trap path (arm64: reserved tmp register + x17 scratch; amd64: word below RSP + RBP scratch, since there's no reserved scratch register there). When source-offset info is tracked (debug info), the inline lowering is kept so recorded trap addresses still map to the individual site.

Two commits, one per backend:
- arm64: total generated code for a TeX-engine (tectonic) wasm32-wasip1 module shrinks from 16.3MB to 10.8MB (-34%); wall time on an M1 Pro is unchanged within noise, with the benefit expected on smaller instruction caches / cache contention.
- amd64: port of the same approach; measured on an i7-3770 (32KiB L1i) compiling a tabularray longtblr document with the tectonic module, wall time drops ~9% versus the bounds-check elision alone, confirming the benefit scales with instruction cache pressure.

This PR is **not stacked** on #2514 (the bounds-check elision PR) and has no code or build dependency on it: the two changes touch disjoint files (this PR only touches the arm64/amd64 backends; #2514 only touches the frontend), and this branch passes the full test suite — including amd64-specific backend codegen tests and spec tests executing the actual generated machine code — with #2514's commit entirely absent.

One overlap worth flagging for merge ordering: both PRs regenerate golden assembly-listing expectations in `backend_test.go` for the same couple of test cases (`memory_stores`, `AtomicRmwAdd`; this PR also touches `memory_load_basic`), since both change codegen for those sample modules. That's a golden-text overlap, not a functional dependency — whichever of the two lands second will hit a merge conflict there and need those golden values regenerated against the already-merged change, same as any two PRs touching the same generated fixture.

## Test plan

- [x] `go test ./...` (linux/amd64, darwin/arm64 — amd64 verified via QEMU emulation since I don't have amd64 hardware on hand)
- [x] spec tests (including under amd64 emulation, exercising the actual generated machine code at runtime)
- [x] cross-compile checks from `CONTRIBUTING.md` (plan9, wasip1, gojs, aix, s390x, ppc64le, arm, 386)
- [x] `make format`

---

Note on process: this change was developed with AI assistance (Claude Code); I've reviewed the code and the reasoning behind it to the best of my ability.

### DIFF
--- a/internal/engine/wazevo/backend/isa/amd64/machine.go
+++ b/internal/engine/wazevo/backend/isa/amd64/machine.go
@@ -89,7 +89,13 @@ type (
 		jmpTableTargets [][]uint32
 		// jmpTableTargetNext is the index to the jmpTableTargets slice to be used for the next jump table.
 		jmpTableTargetsNext int
-		consts              []_const
+
+		// trapIslands are this function's shared conditional-trap exit
+		// sequences, one per exit code, emitted once after register
+		// allocation (see lowerExitIfTrueWithCodeShared and emitTrapIslands).
+		trapIslands []trapIsland
+
+		consts []_const
 
 		constSwizzleMaskConstIndex, constSqmulRoundSatIndex,
 		constI8x16SHLMaskTableIndex, constI8x16LogicalSHRMaskTableIndex,
@@ -181,6 +187,34 @@ func (m *machine) getOrAllocateConstLabel(i *int, _var []byte) label {
 	return m.consts[index].label
 }
 
+// trapIsland is a shared per-function exit sequence for conditional traps
+// with the given exit code, reachable via the label.
+type trapIsland struct {
+	code wazevoapi.ExitCode
+	l    label
+}
+
+// trapIslandCtxOffsetFromRSP is where a conditional-trap site stores the
+// execution context pointer for the island to reload: the word just below
+// the stack pointer (disp32 -16), which holds no live data and cannot be
+// overwritten between the site and the island (no call or push intervenes,
+// and signal handlers run on their own stack).
+const trapIslandCtxOffsetFromRSP uint32 = 0xffff_fff0
+
+// getOrCreateTrapIsland returns the label of this function's shared trap
+// island for the given exit code, allocating it on first use. The island
+// itself is materialized by emitTrapIslands after register allocation.
+func (m *machine) getOrCreateTrapIsland(code wazevoapi.ExitCode) label {
+	for _, ti := range m.trapIslands {
+		if ti.code == code {
+			return ti.l
+		}
+	}
+	l, _ := m.allocateLabel()
+	m.trapIslands = append(m.trapIslands, trapIsland{code: code, l: l})
+	return l
+}
+
 // Reset implements backend.Machine.
 func (m *machine) Reset() {
 	m.consts = m.consts[:0]
@@ -204,6 +238,7 @@ func (m *machine) Reset() {
 	m.perBlockHead, m.perBlockEnd, m.rootInstr = nil, nil, nil
 	m.pendingInstructions = m.pendingInstructions[:0]
 	m.orderedSSABlockLabelPos = m.orderedSSABlockLabelPos[:0]
+	m.trapIslands = m.trapIslands[:0]
 
 	m.amodePool.Reset()
 	m.jmpTableTargetsNext = 0
@@ -1028,7 +1063,14 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
 		m.lowerExitWithCode(m.c.VRegOf(execCtx), code)
 	case ssa.OpcodeExitIfTrueWithCode:
 		execCtx, c, code := instr.ExitIfTrueWithCodeData()
-		m.lowerExitIfTrueWithCode(m.c.VRegOf(execCtx), c, code)
+		if instr.SourceOffset().Valid() {
+			// When source offset info is tracked (debug info), keep the exit
+			// sequence inline so the recorded trap address maps back to this
+			// specific site.
+			m.lowerExitIfTrueWithCode(m.c.VRegOf(execCtx), c, code)
+		} else {
+			m.lowerExitIfTrueWithCodeShared(m.c.VRegOf(execCtx), c, code)
+		}
 	case ssa.OpcodeLoad:
 		ptr, offset, typ := instr.LoadData()
 		dst := m.c.VRegOf(instr.Return())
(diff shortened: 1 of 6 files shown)


ncruces:
Compared to #2514 this one is more complex (~300 vs ~20 LoC).

It's also ... very dear to my heart, as I implemented something very similar for the original compiler.

One important thing that needs to be taken into account is: traps are about the most important use case we have for **line numbers** (so that they show up in stack traces).

We really don't wan't to merge traps when `WithDebugInfoEnabled`. Did you take that into account?

evacchi:
@mgilbir wanna a give this and #2514 another spin and address the comments so we land them?

jerbob92:
## Benchmark report: this PR vs current `main` (real-world workload + wazero benches)

We benchmark wazero as the runtime for [go-pdfium](https://github.com/klippa-app/go-pdfium): PDFium compiled to a 5.3 MB wasm module (~6,400 functions, built with wasm exception handling / `try_table`), rendering and extracting text from real PDF documents. A module of this size is exactly where per-function code-size reductions should pay off, so hopefully a useful datapoint next to the micro benches.

**Setup:** this PR merged onto current `main` (`3ab42173`) vs plain `main`. Apple M5 (darwin/arm64), Go 1.26. go-pdfium numbers are medians of 3 interleaved A/B rounds (one `OpenDocument` + render at 200 DPI + close per op; text extraction is `GetPageTextStructured`, thousands of small guest calls per op). wazero numbers are `benchstat` over `-count=6` of `internal/integration_test/bench` (compiler engine + compilation).

### ✅ Correctness

The **full go-pdfium test suite passes on this PR** — ~660 specs, run on both the compiler and interpreter engines, including heavy `try_table` exception-handling usage (PDFium uses setjmp/longjmp lowered to wasm EH throughout its error paths) and pixel-exact render comparisons.

### go-pdfium (large real module): −3% to −9% across the board

| Workload | main | PR | Δ |
|---|---|---|---|
| render test.pdf (text) | 1.182 ms | 1.072 ms | **−9.3%** |
| render mona.pdf | 0.623 ms | 0.581 ms | **−6.7%** |
| render building_plan.pdf (CAD) | 2.735 ms | 2.570 ms | **−6.0%** |
| render rect-wrong.pdf | 28.91 ms | 27.37 ms | **−5.3%** |
| render invoice A | 10.34 ms | 9.82 ms | **−5.0%** |
| render invoice B | 5.77 ms | 5.49 ms | **−4.9%** |
| render image_page.pdf (big JPEG) | 108.7 ms | 105.7 ms | −2.8% |
| render alpha_channel.pdf | 90.9 ms | 91.3 ms | ~0% |
| text extraction (2 documents) | 2.332 / 0.313 ms | 2.309 / 0.307 ms | −1.0% / −1.9% |

Every render workload improves, most by 5–7%. The two least-affected documents (image_page, alpha_channel) spend their time in a handful of very hot pixel loops where icache pressure is low; the documents that execute *broad* code paths (parsers, fonts, rasterizer, many mid-size functions — exactly the many-conditional-traps profile of a big C++ codebase) benefit the most. This is consistent with the PR's mechanism: PDFium's 6,400 functions are full of memory-access trap checks, and sharing the exit sequences shrinks essentially every function body.

### wazero's own benchmarks: geomean −3.5%, mixed on tiny kernels

```
                                                     │      main       │              this PR               │
                                                     │      sec/op     │   sec/op     vs base               │
Invocation/compiler/fib_for_10-10                         153.6n ±  1%   150.5n ± 2%   -2.02% (p=0.006 n=6)
Invocation/compiler/fib_for_20-10                         17.26µ ±  2%   16.40µ ± 1%   -4.99% (p=0.002 n=6)
Invocation/compiler/fib_for_30-10                         2.073m ±  2%   1.997m ± 2%   -3.64% (p=0.004 n=6)
Invocation/compiler/string_manipulation_size_50-10        7.353µ ±  2%   7.071µ ± 2%   -3.84% (p=0.002 n=6)
Invocation/compiler/string_manipulation_size_100-10       23.34µ ±  3%   21.54µ ± 0%   -7.70% (p=0.002 n=6)
Invocation/compiler/string_manipulation_size_1000-10      1.655m ±  1%   1.704m ± 8%   +2.91% (p=0.002 n=6)
Invocation/compiler/reverse_array_size_500-10             1.615µ ±  1%   1.690µ ± 1%   +4.64% (p=0.002 n=6)
Invocation/compiler/reverse_array_size_1000-10            3.392µ ±  1%   3.576µ ± 1%   +5.44% (p=0.002 n=6)
Invocation/compiler/reverse_array_size_10000-10           75.63µ ±  2%   78.42µ ± 1%   +3.69% (p=0.002 n=6)
Invocation/compiler/random_mat_mul_size_20-10             47.43µ ±  2%   62.77µ ± 3%  +32.35% (p=0.002 n=6)
Compilation/without_extern_cache-10                       5.135m ±  1%   4.631m ± 1%   -9.82% (p=0.002 n=6)
geomean                                                                                -3.51%
```

The micro-bench picture is what one would expect from a code-layout change: fib/string kernels improve, while a few *tiny* single-hot-loop kernels regress a little (`reverse_array` +4–5%) or notably (`random_mat_mul_size_20` +32% — worth a look; possibly the shared exit landing pad displacing that kernel's loop alignment). Kernels that small don't suffer icache pressure, so they only see the layout downside. Compilation itself is **−9.8%** (fewer emitted bytes).

### Verdict

For a large real-world module this PR is a clear, uniform win — **−3% to −9% on every PDFium workload** with the full downstream test suite green, plus ~10% faster compilation. The tiny-kernel regressions in the micro benches (notably `random_mat_mul_size_20`) may be worth investigating for loop-alignment interaction, but they don't reproduce in any real workload we have. Strong 👍 from us.

evacchi:
same Q as #2514, if you have time to spare ./internal/integration_test/stdlibs (zig stdlib in particular) gives a comparable data point. other than that thanks a lot @jerbob92 for the data!!

jerbob92:
Sure:

Follow-up: I also ran this PR against the Zig stdlib integration benchmarks in `internal/integration_test/stdlibs`.

Setup: base = current `main` (6a92c9b3), PR branch = base + this PR only. Apple M5, idle machine, interleaved passes, 10 samples per benchmark. Zig test binaries built with Zig 0.11.0 per the Makefile (`test.wasm` + wasm-opt'd `test-opt.wasm`); every `Run` iteration executes the full Zig stdlib test suite (~11.5s of guest execution) and must exit 0.

```
                             │      main      │             PR #2515              │
                             │     sec/op     │   sec/op    vs base               │
Zig/Compile/test-opt.wasm-10       1.773 ± 3%   1.715 ± 2%  -3.29% (p=0.001 n=10)
Zig/Run/test-opt.wasm-10           11.17 ± 1%   11.11 ± 1%       ~ (p=0.105 n=10)
Zig/Compile/test.wasm-10           2.139 ± 3%   2.074 ± 0%  -3.03% (p=0.023 n=10)
Zig/Run/test.wasm-10               11.55 ± 1%   11.65 ± 0%  +0.85% (p=0.001 n=10)
geomean                            4.703        4.632       -1.51%
```

Verdict: the consistent −3% on compilation of both binaries is the headline — sharing the conditional-trap exit sequences means less code to emit per function. Execution is neutral on the optimized binary and +0.85% on the unoptimized one (real but small; the Zig stdlib suite has far fewer conditional-trap sites than the bounds-check-dense C/C++ modules where this PR shows its execution wins).

evacchi (approved):
I'm good with this! @ncruces thoughts?

internal/engine/wazevo/backend/isa/amd64/machine.go
```diff
@@ -1028,7 +1063,14 @@ func (m *machine) LowerInstr(instr *ssa.Instruction) {
 		m.lowerExitWithCode(m.c.VRegOf(execCtx), code)
 	case ssa.OpcodeExitIfTrueWithCode:
 		execCtx, c, code := instr.ExitIfTrueWithCodeData()
-		m.lowerExitIfTrueWithCode(m.c.VRegOf(execCtx), c, code)
+		if instr.SourceOffset().Valid() {
+			// When source offset info is tracked (debug info), keep the exit
+			// sequence inline so the recorded trap address maps back to this
+			// specific site.
+			m.lowerExitIfTrueWithCode(m.c.VRegOf(execCtx), c, code)
```
ncruces, line 1070: Inline with my https://github.com/wazero/wazero/pull/2515#issuecomment-4916881784, if this does what it says on the tin, I'm OK to merge. That was my major concern.

ncruces (approved):

PR #2517, opened 2026-07-26
fix: capacity sizing when module maximum exceeds memory limit

levmv:
When `WithMemoryCapacityFromMax` is enabled, `newMemorySizer` uses the module-declared maximum before applying `WithMemoryLimitPages`. As a result, modules whose maximum exceeds the configured memory limit are rejected instead of using that limit as their effective maximum.

This change applies the configured limit first, then uses the resulting effective maximum as the capacity. This fixes the remaining `WithMemoryCapacityFromMax` case reported in #1153. Behavior is unchanged when `WithMemoryCapacityFromMax` is disabled or the module maximum is already within the configured limit.


evacchi (approved):
thanks for your contribution!

Issue #2522, opened 2026-08-14
Exception Handling support introduces use after free vulnerability


mpoindexter:
**Describe the bug**
The current version of the exception handling code assumes that only one exception can be live at any time. This conflicts with the spec's assumption that any number of exceptions can be live, and referred to via an exnref handle.

Based on this assumption, wazero currently uses a single field to root the current exception so GC does not collect it, and passes exnref to guest code as a pointer to the single current exception, cast to a uintptr. The guest can then retain that exnref, cause a new exception to be allocated and now write to an already freed location on the Go heap.

**To Reproduce**
A small wasm module and Go program reproduce this as a crash. Note that this could potentially have more serious side effects as it allows guest writes into the host heap, but this at least demonstrates the problem. Note that the call to a host GC function as shown in this POC is not necessary for the problem to occur, it just makes it more reproducible.

```go
package main

import (
        "context"
        _ "embed"
        "fmt"
        "os"
        "runtime"

        "github.com/tetratelabs/wazero"
        "github.com/tetratelabs/wazero/api"
        "github.com/tetratelabs/wazero/experimental"
)

//go:embed poc.wasm
var pocWasm []byte

func main() {
        ctx := context.Background()

        cfg := wazero.NewRuntimeConfigCompiler()
        if len(os.Args) > 1 && os.Args[1] == "interp" {
                cfg = wazero.NewRuntimeConfigInterpreter()
        }
        r := wazero.NewRuntimeWithConfig(ctx,
                cfg.WithCoreFeatures(api.CoreFeaturesV2|experimental.CoreFeaturesExceptionHandling))
        defer r.Close(ctx)

        // The only thing the host does is collect: runtime.GC() is what the guest's
        // (call $gc) runs.
        if _, err := r.NewHostModuleBuilder("host").
                NewFunctionBuilder().WithFunc(func() { runtime.GC() }).Export("gc").
                Instantiate(ctx); err != nil {
                panic(err)
        }

        mod, err := r.Instantiate(ctx, pocWasm)
        if err != nil {
                panic(err)
        }
        run := mod.ExportedFunction("run")

        // Whether the freed exception's span is fully released -- so the dangling pointer
        // the runtime installs lands in unallocated memory -- depends on heap state, so run
        // the case a handful of times. The process aborts mid-loop on the vulnerable runtime.
        const attempts = 10
        for i := 0; i < attempts; i++ {
                if _, err := run.Call(ctx); err != nil {
[...]

PR #2523, opened 2026-08-14
exceptions: support more than one active exception

mpoindexter:
An exnref was a pointer to a wasm.Exception, passed to guest code as an integer. The runtime kept the exception it was propagating in a single field, and that field was the only thing keeping the exception alive.

The spec allows more than one exception to be active at once. A handler catches one with catch_ref and stores the exnref. The guest throws a second exception, which overwrites the field. Nothing in Go references the first exception any more, and the integer the guest holds is invisible to the garbage collector, so the object can be freed or moved. A later throw_ref on that exnref dereferences the stale pointer and reads whatever is at that address as a tag and a params slice. This is memory corruption in the host process.

An exnref is now a handle: an opaque integer that guest code cannot dereference. The runtime maps handles to exceptions and drops the mapping when nothing holds the handle. Holders are counted in two places, because they have different costs.

Global and table slots are counted in wasm.ExceptionStore, under a mutex, since a store can be used from several goroutines at once. The engines pass the slot address to the store, which reads or writes the slot itself. Doing the access under the same lock as the count keeps two barriers on one slot from racing. This covers global.get and global.set, table.get and table.set, and the bulk table.fill, table.copy, table.init and table.grow.

Everything else that can hold a handle belongs to one call: operand stack slots, locals, the exception being raised, and the params of another exception. None of these is reachable from another goroutine, so the call engine counts them in a plain map with no locking. Exnrefs in globals and tables are rare in practice, so the locked path is rare, while the per-call path runs on every catch and every local.get. The compiler emits the per-call adjustments inline, guarded on the handle being non-zero, so an exnref local holding ref.null exn does not call into Go.

Compiled code cannot release what an unwind discards, because the frames a raise passes through do not run again. A try_table checkpoint already saves the operand stack and the locals, so it now saves the call's counts as well, and catching restores them. Locals need one correction: a handler reads them from the save area rather than from the restored stack, so the runtime takes a reference for the handle in the save area, and the handler releases the one the restore put back.

Since slots hold counted references, closing a module releases what its globals and tables named, through a new ModuleEngine.ModuleClosed hook. Instantiation needs no counterpart, because a constant expression can only produce ref.null exn. That relies on global.get in a constant expression naming an immutable global, which the spec requires and wazero did not check, so validation now rejects it.

A handle is only meaningful inside the call that created it, so it cannot cross the host boundary. A host function may not declare an exnref in its signature. A wasm function whose signature has one is no longer callable from the host: the export still exists and can be introspected, but calling it returns an error. Wasm-to-wasm calls of such types are unaffected.

throw_ref raises an exception from a different place than the original throw, so an uncaught exception now reports two stacks: the one it was propagating on, and, if it differs, the one it was first thrown on, under its own header. wasmdebug gained sections to print this.

A frame that an exception unwinds never returns, so a FunctionListener watching it never got an exit call. Each frame a raise removes is now reported to its listener as aborted, with
experimental.ErrUnwoundByException, whether or not a handler catches the exception; a call that succeeds can contain aborted frames. Reporting every frame means unwinding further than a stack trace prints, so unwindStack takes the frame limit as a parameter.

This is a possible fix for #2522


mpoindexter:
@evacchi - I did some further work on this to make it simpler. I got rid of the precise reference counting for exnrefs held on the native stack, and replaced with a conservative GC scan. This ends up being much smaller, and likely a better direction for eventual GC. That cuts the size of the change considerably. I also eliminated the changes to track the original location for rethrown exceptions - if we want it, can go in a separate PR. I kept the more precise listener notifications since that seems like a bug to me (without it, a function can be entered, but never exited from the listener's point of view on a caught and handled exception). 

While doing this, I also found another bug: if the catch handler of the top level try table of a function mutates locals, they are written to the wrong localSaveArea. See the second commit in this PR for details. This was an existing bug.

evacchi:
Feel free to take a look at this branch of mine where I'm trying to bridge with the future wasmgc. I swear I'll try to take a look at this soon 😄 https://github.com/evacchi/wazero/tree/fix-exnref-uaf

mpoindexter:
@evacchi that is super cool and definitely more elegant than what I was doing in the original version of this PR!

A couple of thoughts:

- keeping the use of pointers vs having some sort of handle reference table is quite nice since a host function now has a way to understand reference typed parameters (by casting them to a pointer). A couple of tradeoffs I think it implies: 
  1. I guess in this design a WASM function called from Go can never return a reference typed value, since there's no place to keep the reference alive that I can see. A handle based design would also have to deal with this, but at least it can guarantee that attempting to resolve a reference that's not held anywhere deterministically returns an error if you try to resolve it vs the common pattern being "cast a uintptr to a Go pointer", which is likely to work _sometimes_, and often in test code where the GC hasn't reclaimed the thing. 
  2. If Go ever switches to a moving GC, I guess it will require coordination with it using something like `runtime.Pinner` - not a concern now clearly.
  
- it's not immediately apparent to me that this will behave correctly in the face of multiple concurrent call engines manipulating globals/tables - maybe it will, I haven't analyzed it in enough detail to be sure, but the pattern of writing uintptrs to tables/globals and then after that calling a method to sync the uintptrs into actual pointers the GC can see seems like a place where concurrency can cause trouble. The boundary for where concurrent access can leak in is quite broad: a wazero Runtime instance, since module instances can import/export globals and tables. Obviously today writing globals/tables concurrently is racy, but at least it's not in a way that can crash the host process - the worst guest code can do is cause other guest code to get incorrect results. Trying ensure this was always safe is why in my code I just replaced the code path for interacting with ref typed globals/tables with one that synchronizes all access.

- Unless I misunderstood something, the shadow stack design implies that writing a local with a reference typed value incurs crossing the Go boundary, which is inherently pretty expensive. For exnrefs I don't think it matters really at all, but for the general GC case it might, where lots of ref typed values get passed around throughout the call stack.

PR #2527, opened 2026-08-18
Fix listener test failure on Go 1.27

### LINKED ISSUE #2526: TestEngine*/before_listener_stack_iterator_offsets fails with Go 1.27 (regression from Go 1.26)
## Summary

`before_listener_stack_iterator_offsets` fails deterministically under Go 1.27 (tested: go1.27rc2 and go1.27rc3) in **both** the compiler and the interpreter engines, while passing under go1.26.5 on the same machine. Also reproducible on wazero `main` (checked 2026-08-18) and on the v1.12.0 tag.

## Reproduction

```
git clone https://github.com/tetratelabs/wazero && cd wazero   # v1.12.0 or main
go test ./internal/integration_test/engine/ -run 'TestEngine/stack_iterator' -count=1 -v
```

With go1.26.5: PASS. With go1.27rc3:

```
--- FAIL: TestEngineCompiler (0.00s)
    --- FAIL: TestEngineCompiler/before_listener_stack_iterator_offsets (0.00s)
```

```
    adhoc_test.go:1620: f1 offset: 0x3
    adhoc_test.go:1621: f2 offset: 0x11
    adhoc_test.go:1622: f3 offset: 0x18
    adhoc_test.go:1640: Recorded stack 0 :
    adhoc_test.go:1645: 	0 -> f1 :: 0x0
    require.go:331: unexpected value
        expected: adhoc.frame{..., offset:0x3}
        was:      adhoc.frame{..., offset:0x0}
```

The recorded program counter for the f1 frame maps to source offset `0x0` instead of the expected `0x3`.

## Environment

- Linux amd64 (also fails on the Ubuntu buildd environment; PPA build logs available on request)
- go1.27rc2 / go1.27rc3 (fail) vs go1.26.5 (pass), identical checkout, same host
- Fails identically for `TestEngineCompiler` and `TestEngineInterpreter`

## Context

Found while test-rebuilding wazero's Ubuntu reverse-dependencies against Go 1.27 ahead of the Ubuntu golang-defaults 1.26 -> 1.27 transition. Since this involves the experimental listener StackIterator API, I can't tell whether Go 1.27 changed something the engine relies on (e.g. runtime/stack behavior the compiler backend's assembly glue depends on) or whether this is a latent wazero bug exposed by toolchain/runtime changes. Happy to help bisect Go commits if useful.

mpoindexter:
Prior to Go 1.27, `debug/Dwarf.New` would return an empty `dwarf.Data` if invalid input was given to `dwarf.New`. In go 1.27 this returns nil instead.  This change caused Module.DWARFLines to become nil, which prevented the recording of source offset information.

Change Module.DWARFLines presence to be entirely controlled by the dwarfEnabled flag: if there is no `dwarf.Data`, the methods in DWARFLines just become a noop.

Fixes #2526

### DIFF
--- a/internal/wasmdebug/dwarf.go
+++ b/internal/wasmdebug/dwarf.go
@@ -27,9 +27,6 @@ type line struct {
 
 // NewDWARFLines returns DWARFLines for the given *dwarf.Data.
 func NewDWARFLines(d *dwarf.Data) *DWARFLines {
-	if d == nil {
-		return nil
-	}
 	return &DWARFLines{d: d, linesPerEntry: map[dwarf.Offset][]line{}}
 }
 
@@ -49,7 +46,7 @@ func isTombstoneAddr(addr uint64) bool {
 // Line returns the line information for the given instructionOffset which is an offset in
 // the code section of the original Wasm binary. Returns empty string if the info is not found.
 func (d *DWARFLines) Line(instructionOffset uint64) (ret []string) {
-	if d == nil {
+	if d == nil || d.d == nil {
 		return
 	}
 


ncruces (approved):

internal/wasmdebug/dwarf.go
```diff
@@ -27,9 +27,6 @@ type line struct {
 
 // NewDWARFLines returns DWARFLines for the given *dwarf.Data.
 func NewDWARFLines(d *dwarf.Data) *DWARFLines {
-	if d == nil {
-		return nil
-	}
```
ncruces, line 32: Oh, now I get it, sorry. I was reading this diff backwards. You removed the `nil` check, you didn't add it.

PR #2529, opened 2026-08-19
wazevo: fix group IDs of split critical-edge branches, restoring compare→flag fusion

jerbob92:
Disclaimer: most of this work was done using Claude Fable with manual verification/guidance/planning by me.

> **Review note.** This PR changed shape during review. It was first posted as "sink pure comparisons into flag consumers across instruction groups" (`MatchPureInstrOneOf`). @evacchi's reproducer showed that version fusing a compare across basic blocks and crashing the register allocator; chasing that revealed that the fusion failures it set out to fix have a different cause, in `splitCriticalEdge`, and that fixing the cause recovers the fusions on its own. The pure matcher has since been removed (last commit); the description below covers the branch as it now stands. The earlier commits are kept so the review discussion still lines up.

## The bug

wazevo fuses an `Icmp`/`Fcmp` into the conditional branch or `select` that consumes it (`SUBS` + `B.cond` rather than `CSET` + `CBNZ`). The fusion is gated on `MatchInstr`, which requires the comparison to be in the same **instruction group** as its consumer. Group IDs are assigned by `passDeadCodeEliminationOpt` in one sweep over all blocks; every side-effecting instruction opens a new group, and since every block ends in a branch, no two blocks share a group.

`splitCriticalEdge` runs *after* that sweep. It allocates a new branch and swaps it into the predecessor in place of the original, which moves out to the trampoline. The new branch was never numbered: it kept group **0**. Two consequences:

1. **Fusion silently fails at every split conditional branch.** The branch is lowered with `currentGID == 0`, so no compare can match it. On PDFium (6,600 functions), **63,584 of 99,716 conditional-branch sites (64%) are lowered at group 0**. Loop backedges are the archetypal critical edge, so this hits exactly the hot loops: in the module's hottest function, a pixel-compositing loop, the backedge was

   ```
   CMP  W16, W12
   CSET X17, NE
   CBNZ W17, .+0x8
   B    .+0x8
   B    .-0x5b0        ; the actual backedge
   ```

   where a single fused `B.NE` suffices.

2. **A latent miscompile.** A compare that genuinely *is* in group 0 — the entry block's first group — passes the strict check at any of those branches, in any block, and is fused across the block boundary. `TestE2E/compare_fused_across_split_edge` is that shape (compare in the entry block; `br_if` as the first instruction of the merge block after an `if`, on an edge carrying a block argument into a block with two predecessors). On `main` it fuses across the boundary; it happens to produce correct code because nothing spills, but under register pressure this is a `definingBlk should not be nil` panic at compile time. It does not occur in PDFium, but it is reachable from plain wasm.

## The fix

`passDeadCodeEliminationOpt` leaves the next unused group ID on the builder. `splitCriticalEdge` gives the swapped-in branch the original's group (it takes its place as the predecessor's terminator) and the moved branch a fresh one for the trampoline. Every block's instructions now carry a contiguous run of group IDs unique to that block, which is what the `InstructionGroupID` doc has always claimed. `TestBuilder_splitCriticalEdge` checks the numbering. Five lines of code in `ssa`; the strict matcher is unchanged.

| PDFium, compare→flag fusions at branch + select sites (arm64) | count |
|---|---|
| `main` | 28,649 |
| this PR | **80,446** |
| for reference: the pure matcher as first posted | 80,669, of which 204 across block boundaries |

Also fixed: amd64 `lowerSelect` called `icmp.Lowered()` — a getter whose result was discarded — so a fused compare was never marked and got lowered a second time. Now `MarkLowered()`.

## Why the pure matcher was dropped

With the group IDs right, sinking a comparison across an instruction group added **19** fusions on PDFium (0.02%), and review found it a poor trade even for those: sunk past a call it keeps both compare operands live instead of one flag (arm64: 6 instructions / 1 spill slot / 2 memory ops → 7 / 2 / 4), and a compare sunk to the branch site can no longer fold its own operand (an extend, a shift, a masked test), since those folds are gated on the branch's group. It was also the source of both soundness holes found in review. So it is gone, and `icmp_in_different_group` is back to asserting the strict result.

## Correctness

- Full spectest matrix passes: v1, v2, exception-handling, threads, tail-call, typed-function-references, extended-const. Full repo suite passes.
- Regression tests: `TestE2E/compare_fused_across_blocks` (@evacchi's reproducer against the earlier version) and `TestE2E/compare_fused_across_split_edge`; `TestBuilder_splitCriticalEdge` asserts the group numbering.
- Instrumented ground truth (a walk of the current block's instruction list, independent of group IDs): compiling PDFium and running the v2 spectests performs **0** cross-block fusions and leaves **0** non-entry blocks with a group-0 root (`main`: 1,377 such blocks).
- Rendered output over the PDF corpus is byte-identical to `main`.

## Performance (Apple M5, arm64; go-pdfium, 200 PDFs × ≤3 pages at 100 DPI; benchstat, interleaved)

- `main` → this PR: −4.5% to −4.9% point estimate across runs; the cleanest run gave **−4.87% (p=0.007, n=10)** for a tree differing from this one by 19 fusions, and this tree vs that one is **~ (p=0.58, n=10)**. This machine's run-to-run noise is ±2–11%, so treat the fusion counts above as the reliable evidence and the timings as consistent with them.
- Earlier runs against the original version of the PR: geomean −1.24% on per-document go-pdfium benchmarks at 200 DPI (mona.pdf −4.69% p=0.002, alpha_channel.pdf −2.91% p=0.009); wazero's own suite geomean −0.8%; Zig stdlib suite neutral. Codegen for that version and this one differ only by the 19 fusions and the 204 unsound ones, so those numbers should carry over.

### DIFF
--- a/internal/engine/wazevo/backend/compiler.go
+++ b/internal/engine/wazevo/backend/compiler.go
@@ -87,6 +87,16 @@ type Compiler interface {
 	// Note: caller should be careful to avoid excessive allocation on opcodes slice.
 	MatchInstrOneOf(def SSAValueDefinition, opcodes []ssa.Opcode) ssa.Opcode
 
+	// MatchPureInstrOneOf is like MatchInstrOneOf but without the
+	// instruction-group restriction. It must only be used with pure
+	// (side-effect-free, non-trapping) opcodes whose evaluation can be sunk
+	// to the matching site regardless of intervening side effects, e.g. an
+	// integer comparison sunk into the conditional branch that consumes it.
+	// The not-yet-lowered check keeps this sound under any lowering order:
+	// if the definition was already materialized elsewhere, the match fails
+	// and the caller falls back to using the materialized value.
+	MatchPureInstrOneOf(def SSAValueDefinition, opcodes []ssa.Opcode) ssa.Opcode
+
 	// AddRelocationInfo appends the relocation information for the function reference at the current buffer offset.
 	AddRelocationInfo(funcRef ssa.FuncRef, isTailCall bool)
 
@@ -309,6 +319,27 @@ func (c *compiler) MatchInstr(def SSAValueDefinition, opcode ssa.Opcode) bool {
 		instr.Opcode() == opcode &&
 		instr.GroupID() == c.currentGID &&
 		def.RefCount < 2
+}
+
+// MatchPureInstrOneOf implements Compiler.MatchPureInstrOneOf.
+func (c *compiler) MatchPureInstrOneOf(def SSAValueDefinition, opcodes []ssa.Opcode) ssa.Opcode {
+	instr := def.Instr
+	if !def.IsFromInstr() {
+		return ssa.OpcodeInvalid
+	}
+	if def.RefCount >= 2 {
+		return ssa.OpcodeInvalid
+	}
+	if instr.Lowered() {
+		return ssa.OpcodeInvalid
+	}
+	opcode := instr.Opcode()
+	for _, op := range opcodes {
+		if opcode == op {
+			return opcode
+		}
+	}
+	return ssa.OpcodeInvalid
 }
 
 // MatchInstrOneOf implements Compiler.MatchInstrOneOf.
--- a/internal/engine/wazevo/backend/compiler.go
+++ b/internal/engine/wazevo/backend/compiler.go
@@ -92,13 +92,16 @@ type Compiler interface {
 	// must only be used with pure (side-effect-free, non-trapping) opcodes
 	// whose evaluation can be sunk to the matching site regardless of
 	// intervening side effects, e.g. an integer comparison sunk into the
-	// conditional branch that consumes it. A definition from an earlier
-	// block never matches, regardless of purity: fusing it away would leave
-	// it unmaterialized where other block-scoped backend state, such as the
-	// register allocator's liveness analysis, still expects to find it.
-	// The not-yet-lowered check keeps this sound under any lowering order:
-	// if the definition was already materialized elsewhere, the match fails
-	// and the caller falls back to using the materialized value.
+	// conditional branch that consumes it.
+	//
+	// A definition from another block never matches, regardless of purity.
+	// Blocks are lowered in reverse post-order, so a definition in another
+	// block has usually been lowered already; but Lowered() only records
+	// consumption by a fusion, not ordinary lowering, so it cannot tell. The
+	// definition would then be emitted twice, and anything folded into its
+	// first emission (an extend, say) never gets a register of its own, which
+	// the second emission then reads. Within one block the consumer is always
+	// lowered before the definition it fuses, so this cannot happen.
 	MatchPureInstrOneOf(def SSAValueDefinition, opcodes []ssa.Opcode) ssa.Opcode
 
 	// AddRelocationInfo appends the relocation information for the function reference at the current buffer offset.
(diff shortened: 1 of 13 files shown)


internal/engine/wazevo/backend/compiler.go
```diff
@@ -311,6 +321,27 @@ func (c *compiler) MatchInstr(def SSAValueDefinition, opcode ssa.Opcode) bool {
 		def.RefCount < 2
 }
 
+// MatchPureInstrOneOf implements Compiler.MatchPureInstrOneOf.
+func (c *compiler) MatchPureInstrOneOf(def SSAValueDefinition, opcodes []ssa.Opcode) ssa.Opcode {
```
evacchi: this reads no state from the compiler `c` so you could move it to a package-level func. It removes the duplication in `util_test.go` for both backends

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -82,7 +87,7 @@ func (m *machine) LowerConditionalBranch(b *ssa.Instruction) {
 	cvalDef := m.compiler.ValueDefinition(cval)
 
 	switch {
-	case m.compiler.MatchInstr(cvalDef, ssa.OpcodeIcmp): // This case, we can use the ALU flag set by SUBS instruction.
+	case m.compiler.MatchPureInstrOneOf(cvalDef, condBranchFusionMatchesIcmp[:]) == ssa.OpcodeIcmp: // This case, we can use the ALU flag set by SUBS instruction.
```
evacchi: awkward: we define 1-element arrays above just to use `MatchPureInstrOneOf` here, but it would be more natural to have a `MatchPureInstr` that takes 1 elem instead of an array, or align to amd64 where there is only 1 `var condBranchMatches = [...]ssa.Opcode{ssa.OpcodeIcmp, ssa.OpcodeFcmp}`

internal/engine/wazevo/backend/isa/arm64/lower_instr.go
```diff
@@ -2038,13 +2043,13 @@ func (m *machine) lowerSelect(c, x, y, result ssa.Value) {
 
 	var cc condFlag
 	switch {
-	case m.compiler.MatchInstr(cvalDef, ssa.OpcodeIcmp): // This case, we can use the ALU flag set by SUBS instruction.
+	case m.compiler.MatchPureInstrOneOf(cvalDef, condBranchFusionMatchesIcmp[:]) == ssa.OpcodeIcmp: // This case, we can use the ALU flag set by SUBS instruction.
```
evacchi: same as above

internal/engine/wazevo/backend/isa/amd64/machine.go
evacchi: maybe this could be turned into `MarkLowered()`

evacchi (commented):
mostly nits. However I'm investigating a possible issue that this patch might uncover. Namely, this test crashes the compiler

```diff
diff --git a/internal/engine/wazevo/e2e_test.go b/internal/engine/wazevo/e2e_test.go
index 16341ec4..420ef19f 100644
--- a/internal/engine/wazevo/e2e_test.go
+++ b/internal/engine/wazevo/e2e_test.go
@@ -959,6 +959,13 @@ func TestE2E(t *testing.T) {
 				expResults: []uint64{9},
 			}},
 		},
+		{
+			name: "compare_fused_across_blocks", m: testcases.CompareFusedAcrossBlocks.Module,
+			calls: []callCase{
+				{params: []uint64{5, 5}, expResults: []uint64{7}},
+				{params: []uint64{5, 6}, expResults: []uint64{9}},
+			},
+		},
 	} {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
diff --git a/internal/engine/wazevo/testcases/testcases.go b/internal/engine/wazevo/testcases/testcases.go
index 830ce3de..3ceb6b3e 100644
--- a/internal/engine/wazevo/testcases/testcases.go
+++ b/internal/engine/wazevo/testcases/testcases.go
@@ -3091,6 +3091,44 @@ var TryTableCatchWithReturnCall = TestCase{
 	},
 }
 
+// CompareFusedAcrossBlocks defines a compare whose operand folds into the
+// compare itself (i64.extend_i32_u on arm64), consumed by a branch in a
+// later block. The store in between forces a new instruction group, so the
+// compare must not be re-materialized at the branch.
+var CompareFusedAcrossBlocks = TestCase{
+	Name: "compare_fused_across_blocks",
+	Module: &wasm.Module{
+		TypeSection:     []wasm.FunctionType{{Params: []wasm.ValueType{i64, i32}, Results: []wasm.ValueType{i32}}},
+		ExportSection:   []wasm.Export{{Name: ExportedFunctionName, Type: wasm.ExternTypeFunc, Index: 0}},
+		MemorySection:   &wasm.Memory{Min: 1},
+		FunctionSection: []wasm.Index{0},
+		CodeSection: []wasm.Code{{
+			LocalTypes: []wasm.ValueType{i32},
+			Body: []byte{
+				wasm.OpcodeLocalGet, 0,
+				wasm.OpcodeLocalGet, 1,
+				wasm.OpcodeI64ExtendI32U,
+				wasm.OpcodeI64Eq,
+				wasm.OpcodeLocalSet, 2,
+				// Side effect between the compare and its use.
+				wasm.OpcodeLocalGet, 1,
+				wasm.OpcodeIf, blockSignature_vv,
+				wasm.OpcodeI32Const, 0,
+				wasm.OpcodeI32Const, 0,
+				wasm.OpcodeI32Store, 0x2, 0x0,
+				wasm.OpcodeEnd,
+				wasm.OpcodeLocalGet, 2,
+				wasm.OpcodeIf, blockSignature_vv,
+				wasm.OpcodeI32Const, 7,
+				wasm.OpcodeReturn,
+				wasm.OpcodeEnd,
+				wasm.OpcodeI32Const, 9,
+				wasm.OpcodeEnd,
+			},
+		}},
+	},
+}
+
 var LargeMethodBodyWithManyArgs = TestCase{
 	Name: "large_method_body_with_many_args",
 	Module: &wasm.Module{
```

PR #2530, opened 2026-08-19
wazevo: EH checkpoint optimizations

jerbob92:
Disclaimer: most of this work was done using Claude Fable with manual verification/guidance/planning by me.

### Problem

Entering a wasm `try_table` takes a checkpoint of the current execution state so a matching throw can unwind to it. The current implementation checkpoints by **cloning the entire call stack** (the same mechanism as `experimental.Snapshot`), allocating a fresh buffer every time. For exception-heavy modules — e.g. C++ compiled with wasm exceptions, where RAII-styled code enters `try_table` scopes constantly — this makes try_table entry both an allocator hot path and a memory-bandwidth hot path: rendering one alpha-compositing PDF page through an EH-enabled PDFium module allocated **49 MB across 4,731 allocations** in checkpoint clones alone.

### Change (two commits)

**Commit 1 — pool the checkpoint buffers.** Stack-clone buffers and locals save areas are recycled through per-call-engine free lists (`takePooledTryStack`, `trySaveAreaPool`) instead of being freshly allocated per `try_table` entry and dropped on exit. This removes steady-state allocation without changing what is stored.

**Commit 2 — stop cloning at all in the common case.** A checkpoint does not need a copy of the whole stack: as long as the live stack only *grows* deeper after the checkpoint (the normal case), the checkpoint state can be described by **offsets from the stack top** (`spOff`/`fpOff`) plus a small snapshot of the trampoline-owned window at the entry stack pointer (frame info + arg/ret area — bounded at 256 bytes, ISA-specific layout). On unwind, the offsets are re-anchored to the current stack top and the trampoline window is written back. Full clones remain in exactly one case: when the `experimental.Snapshot` API is enabled, since snapshot restores can rewind the stack to an earlier state, which would invalidate top-relative offsets — that case keeps the (now pooled) clone path.

### Fixes folded in during the rebase (the "failing tests")

**1. `checkptr` violations under `-race`.** The branch failed `-race` runs — not due to data races but **`checkptr` violations** (checkptr is enabled by `-race`): the new code read the trampoline window through bare `uintptr`→`unsafe.Pointer` conversions and a deprecated `reflect.SliceHeader` launder, at three sites (`trampolineWindowBytes` on both ISAs, the checkpoint snapshot, and the unwind restore). All three now derive their pointers from typed sources (`stackPointerBeforeGoCall *uint64`, `unsafe.Add` on the stack slice) — the idiom the surrounding upstream code already uses.

**2. amd64 was miscompiling every catch-after-a-call (CI: exception-handling spectests + `TestCppExceptions` failing with `unreachable`).** The offset-mode restore re-enters the try_table-enter trampoline's resume path, which on amd64 ends with `mov rsp, rbp; pop rbp; ret` — it reads **Caller_RBP from `[rbp]` and the return address from `[rbp+8]`, on the stack** (arm64 takes the resume return address from the execution context instead). The snapshot window was sized `8 + sizeInBytes` (slice-size slot + *unaligned* arg/ret area), which excludes the alignment padding and those two slots — and any call between the try_table entry and the throw overwrites exactly them (the next `call` pushes its return address at the same stack depth). The window is now computed as `fp + 16 − sp`, the exact trampoline-owned span from the Go-call stack pointer up to and including the Caller_RBP/Return-Addr pair (`trampolineWindowBytes` takes `framePointerBeforeGoCall` as a second argument).

**3. amd64 stale absolute frame pointer after stack growth.** The snapshotted Caller_RBP is an *absolute* pointer into the stack buffer; if `growStack` reallocates the stack between entry and throw, the restored value would dangle (this is the same amd64-only concern that makes upstream `adjustClonedStack` a no-op on arm64 but real on amd64). Offset-mode checkpoints now record the entry-time stack top, and a new per-ISA `rebaseTrampolineWindow` hook rebases the Caller_RBP slot by the stack-top delta on restore (no-op on arm64: its window holds no stack-absolute values).

**4. riscv64 (and every non-amd64/arm64 arch) failed to build**: the `isa_other.go` stub had kept the old `trampolineWindowBytes` signature. Fixed, and cross-builds of linux/riscv64, linux/386, and windows/amd64 verified.

## Correctness

- Full spectest matrix passes on **both arm64 (native) and amd64 (verified locally under Rosetta 2)**: v1, v2, **exception-handling**, threads, tail-call, typed-function-references, extended-const. The amd64 exception-handling suite and `experimental` `TestCppExceptions` failed before fix 2 above.
- `-race` (with checkptr): exception-handling spectests, `experimental`, wazevo unit tests, and `integration_test/engine` all pass (these failed before fix 1 above).
- `integration_test/engine` passes on amd64 as well.
- Cross-builds verified for linux/riscv64, linux/386, windows/amd64.
- A downstream golden-hash corpus (EH-enabled PDFium rendering 32 verified outputs) is unchanged.

### PDF rendering (go-pdfium: EH-enabled PDFium-as-wasm, open + render at 200 DPI + close)

**Allocations — the headline.** Per rendered page:

```
                          B/op (before)   B/op (after)              allocs/op
alpha_channel.pdf         50,285 Ki   →   1.7 Ki   (-100.0%)        4731 → 34  (-99.3%)
image_page.pdf            42,941 Ki   →   1.7 Ki   (-100.0%)        4040 → 34  (-99.2%)
invoice A                 15,516 Ki   →   1.7 Ki    (-99.99%)       1503 → 34  (-97.7%)
mona.pdf                   2,639 Ki   →   1.7 Ki    (-99.94%)        280 → 34  (-87.9%)
geomean                      433 Ki   →   1.7 Ki    (-99.6%)         255 → 34  (-86.7%)
```

Every document converges to the same fixed ~1.7 KiB / 34 allocations per render — the checkpoint traffic is gone entirely; what remains is per-call harness overhead.

**Time:** geomean **−2.54%**:

```
Wasm/mona.pdf-10          551.8µ ± 2%   523.2µ ± 6%  -5.19% (p=0.002)
Wasm/invoice-a.pdf-10     9.596m ± 2%   9.288m ± 3%  -3.21% (p=0.009)
Wasm/building_plan.pdf-10 3.212m ± 4%   3.120m ± 2%  -2.84% (p=0.041)
Wasm/rect-wrong.pdf-10    29.56m ±12%   28.75m ± 4%  -2.74% (p=0.026)
Wasm/image_page.pdf-10    102.3m ± 5%   100.2m ± 1%  -2.08% (p=0.002)
```

The time win is smaller than the allocation win because Go's allocator is fast; the deeper benefit is the removed GC pressure — 50 MB/render of garbage no longer competes with the application's own heap, which matters most in long-running services rendering many documents concurrently.

### wazero's own benchmark suite

Neutral (time geomean −0.4%, B/op geomean −0.3%; scattered sub-µs movers in both directions). Expected: the suite contains no exception-handling workloads, and the changed code only runs on `try_table` entry/unwind.

### Zig stdlib suite (`internal/integration_test/stdlibs`, Zig 0.11.0)

Neutral (all rows n.s., geomean −0.25%) — the Zig test binaries are built without wasm exception handling.

## Notes for review

- Behavior with `experimental.Snapshot` enabled is unchanged by design (full clones, now pooled): top-relative offsets cannot survive a snapshot restore that rewinds the stack.
- The trampoline-window layout knowledge is isolated in per-ISA `trampolineWindowBytes`/`rebaseTrampolineWindow` (arm64: frame info + arg/ret + return-address block, no absolute values; amd64: size slot + aligned arg/ret + Caller_RBP/Return-Addr, with the Caller_RBP slot rebased if the stack buffer moved), mirroring where `GoCallStackView`/`AdjustClonedStack` already encode the same per-ISA layout facts.
- The pool is per-call-engine and therefore goroutine-confined, same lifecycle as the stack itself; no synchronization is added.

### DIFF
--- a/internal/engine/wazevo/call_engine.go
+++ b/internal/engine/wazevo/call_engine.go
@@ -62,11 +62,31 @@ type (
 	// On match, we restore the stack to the checkpoint state and re-enter at returnAddress.
 	tryHandler struct {
 		// Cloned stack and state from the try_table entry checkpoint,
-		// using the same approach as experimental.Snapshot.
+		// using the same approach as experimental.Snapshot. The full clone
+		// is only taken when the snapshotter is enabled (stack != nil); in
+		// the common case the checkpoint is recorded as spOff/fpOff plus a
+		// small snapshot of the trampoline-owned stack window instead.
 		sp, fp, top    uintptr
 		returnAddress  *byte
 		savedRegisters [64][2]uint64
-		stack          []byte // cloned stack
+		stack          []byte // cloned stack, nil in offset mode
+		// spOff/fpOff record the checkpoint stack/frame pointers as offsets
+		// from the aligned stack top. The frames between a try_table entry
+		// and a throw remain intact in the live stack, so the checkpoint can
+		// be restored by recomputing the pointers against the current stack
+		// top; offsets survive stack growth because growStack copies the
+		// contents top-aligned into the new buffer.
+		spOff, fpOff uintptr
+		// trampWindow snapshots the trampoline-owned region at the entry
+		// stack pointer ([sp, sp+16+frame_size+16), per the Go-call stack
+		// layout: frame_size, sliceSize, the arg/ret slice, the return
+		// address and size_of_arg_ret). A later trampoline exit from the
+		// same frame (e.g. the throw itself) reuses this region, so it must
+		// be restored before resuming at the entry continuation — most
+		// importantly the trampoline return address, which would otherwise
+		// make the resumed code continue after the *throw* site.
+		trampWindow      [32]uint64
+		trampWindowBytes uintptr
 		// catchClauses describes what exceptions this handler catches.
 		catchClauses []wazevoapi.CatchClauseInstance
 		// localsSaveArea is a heap buffer where locals are mirrored inside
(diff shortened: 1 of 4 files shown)


unxed:
An independent data point for this PR from a real C++ workload: [Colorer](https://github.com/unxed/colorer4go), a syntax-highlighting library with RAII-heavy C++.

Setup: built with wasi-sdk 34.0-rc.2 using `-fwasm-exceptions -mllvm -wasm-use-legacy-eh=false` (linked with `-lunwind`) as a reactor, then `wasm-opt -O3` (binaryen 131). Run with `CoreFeaturesV2 | experimental.CoreFeaturesExceptionHandling` on linux/amd64 with Go 1.26.8, compiler engine and a warm compilation cache. The baseline is the same library built with `-fno-exceptions`. Parse times are the best of 5; each binary was run twice.

| | no-EH, v1.12.0 | EH, main 451613ca | EH, this PR 414570f5 |
|---|---|---|---|
| load catalog | 131–147 ms | 719–1076 ms | 204–213 ms |
| select file type (C++ grammar) | 124 ms | 3924 ms | 210 ms |
| parse 22.8k lines of C++ | 532 ms | 755 ms | 535 ms |
| parse 256.6k lines of Go | 2045 ms | 4616 ms | 2291 ms |

The regions produced are identical in all three columns. The no-EH module on this PR performs the same as on v1.12.0.

Correctness, for context: on v1.12.0 the compiler trapped with `invalid table access` whenever the module caught a C++ exception, while the interpreter was correct. Bisecting on this module puts the fix exactly at #2504. Since then, and with this PR, the caught-exception scenarios behave as in the interpreter.

Module, inputs and a standalone harness (bench + a correctness check for caught exceptions) are attached: wazero-eh-colorer.zip. Numbers in its README come from that harness.

[wazero-eh-colorer.zip](https://github.com/user-attachments/files/32284667/wazero-eh-colorer.zip)

PR #2533, opened 2026-08-20
Improve performance of WithCloseOnContextDone(true) 

### LINKED ISSUE #2466: Very large performance impact of WithCloseOnContextDone(true)

mpoindexter:
Currently, when WithCloseOnContextDone is true, each loop iteration calls into Go to check if the module is closed. This costs about 12ns per loop iteration on a MacBook M4. The call back in to Go is necessary to ensure that other goroutines run if no cores set according to GOMAXPROCS are available.

We replace this with two constructs: first, a call to runtime.entersyscall/runtime.exitsyscall surrounding the call to asm implemented entrypoint. This is the same mechanism used by the Go runtime for calls that exit the Go runtime like syscalls, or cgo calls. This indicates to the Go scheduler that the current goroutine is now executing external code, and other Go code can be scheduled.
    
Next, we replace the call to check exit code that returns to Go with a direct check of the field in the module (indirected through a pointer stored in executionContext).

Since Go code is now never blocked from running due to WASM code this cheap check is safe since the cancellation goroutine can ensure that the module is closed when the context associated with it is closed.

While working on this, I discovered an additional problem: two code shapes that effectively ignore termination since they execute for large amounts of time without ever looping. First, an exponential call pattern (e.g. func0 calls func1 twice, func1 calls func2 twice, ..., funcN calls funcNplus1 twice) - this will not be limited by stack depth before it gets to effectively infinite numbers. Second, tail calls can recurse without ever running a loop and without growing the stack. A new test demonstrates both of these cases.

To alleviate this problem, I added termination checks to function entries, which covers both cases. This adds a tiny amount of overhead to each function call, only when ensuring termination.

This should resolve #2466.

### DIFF
--- a/internal/engine/wazevo/call_engine.go
+++ b/internal/engine/wazevo/call_engine.go
@@ -49,6 +49,12 @@ type (
 		// pendingException holds the most recently caught exception, so handler
 		// code can read its params after re-entry.
 		pendingException *wasm.Exception
+
+		// entrypoint and afterGoFunctionCallEntrypoint are how this call enters compiled
+		// code. Which pair they hold depends on whether the module was compiled with
+		// ensureTermination: see entrypoints.
+		entrypoint                    entrypointFn
+		afterGoFunctionCallEntrypoint afterGoFunctionCallEntrypointFn
 	}
 
 	// tryHandler records the state at a try_table entry for exception handling.
@@ -147,6 +153,11 @@ type (
 		// memclrAddress holds the address of memclrNoHeapPointers implemented
 		// by the Go runtime. See memclr.go.
 		memclrAddress uintptr
+		// moduleClosedPtr is a pointer to the underlying uint64 of the
+		// owning ModuleInstance.Closed field. Compiled code reads through it
+		// at every loop back-edge (when ensureTermination is on) and calls
+		// back into Go when non-zero.
+		moduleClosedPtr *uint64
 	}
 )
 
@@ -342,7 +353,7 @@ func (c *callEngine) callWithStack(ctx context.Context, paramResultStack []uint6
 	if c.stackTop&(16-1) != 0 {
 		panic("BUG: stack must be aligned to 16 bytes")
 	}
-	entrypoint(c.preambleExecutable, c.executable, c.execCtxPtr, c.parent.opaquePtr, paramResultPtr, c.stackTop)
+	c.entrypoint(c.preambleExecutable, c.executable, c.execCtxPtr, c.parent.opaquePtr, paramResultPtr, c.stackTop)
 	for {
 		switch ec := c.execCtx.exitCode; ec & wazevoapi.ExitCodeMask {
 		case wazevoapi.ExitCodeOK:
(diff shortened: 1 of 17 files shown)


internal/engine/wazevo/call_engine.go
```diff
@@ -144,6 +150,11 @@ type (
 		// localsSaveAreaPtr points to the tryHandler's localsSaveArea slice
 		// backing array. Handlers load locals from this slice.
 		localsSaveAreaPtr uintptr
+		// moduleClosedPtr is a pointer to the underlying uint64 of the
+		// owning ModuleInstance.Closed field. Compiled code reads through it
+		// at every loop back-edge (when ensureTermination is on) and calls
+		// back into Go when non-zero.
+		moduleClosedPtr *uint64
```
evacchi, line 160: an idea I got from some other experiment. you might also be able to allocate a register a callee-saved register at the start of the function and store the pointer here and avoid the memory read on hot loops; tight loops would benefit more but at the cost of more register pressure.

mpoindexter: I don't think we can: the point here is to be able to interrupt when a concurrent goroutine sets a flag in memory. Hoisting to a register would make it pretty much impossible to set from Go I think?

The memory read is quite cheap: it's a predicted-not-taken branch (I took care to make sure the uncommon case is a forward jump so the branch predictor guesses "not taken"), so it's about as cheap as it can be while still allowing for a different goroutine to write to it. I guess the only way to theoretically speed this up at all would be to hoist some sort of counter to a register, and then only check memory 1/K times - but it's so cheap in my measurements that it's really not worth that complexity and code space IMO.

evacchi: Fair!

PR #2540, opened 2026-09-06
poll_oneoff: do not sleep when there is no clock subscription

NotAFlightRisk:
`wazero run` never comes back from a guest that calls `poll(fds, nfds, -1)` on a writable fd. The events do get written back and nevents is right. Then it parks in `Nanosleep` for 9223372036854775807ns, which is about 292 years.

`timeout` starts at max Duration so the loop can take the minimum, and only a clock subscription ever lowers it. wasi-libc adds that subscription inside `if (timeout)` ([ppoll.c](https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/sources/ppoll.c)), so a poll with an infinite timeout sends a list with no clock event in it, and the sentinel makes it all the way to the sleep.

The comment above that sleep already says the timeout is nonzero if there are clock subscriptions, so this is mostly just making that true.

The test passes its own `Nanosleep` in beacuse the default one is a no-op. Nothing in the suite ever saw the argument.

---

Somthing adjacent I've left alone: with a clock subscription and an fd_write subscription together, the fd_write event gets acked straight away but the whole clock timeout still gets slept. Could well be deliberate, and I didnt want to change it in here.

### DIFF
--- a/imports/wasi_snapshot_preview1/poll.go
+++ b/imports/wasi_snapshot_preview1/poll.go
@@ -89,8 +89,8 @@ func pollOneoffFn(_ context.Context, mod api.Module, params []uint64) sys.Errno 
 		file sys.File
 	}
 	// The timeout is initialized at max Duration, the loop will find the minimum.
-	var timeout time.Duration = 1<<63 - 1
-	hasClockSub := false
+	const noTimeout = time.Duration(1<<63 - 1)
+	timeout := noTimeout
 	// Count of all the subscriptions that have been already written back to outBuf.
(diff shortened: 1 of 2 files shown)


evacchi (commented):
I think you might simplify to something like

```
const noTimeout = time.Duration(1<<63 - 1)
...
timeout := noTimeout
...
if timeout != noTimeout {
    sysCtx.Nanosleep(int64(timeout))
}
```

evacchi (approved):

